PR #177, opened 2026-06-03
refactor(db): squash 27 migrations into one greenfield baseline

ishiland:
## Summary
Collapses the `0001..0027` Alembic chain into a single greenfield baseline (`0001_baseline.py`) plus the unchanged procrastinate queue migration (`0002_procrastinate.py`). No existing deployments, so no upgrade/stamp continuity is needed.

The baseline's table/column/index DDL is `alembic revision --autogenerate` output from the current ORM models; the non-autogenerable DDL is grafted on verbatim, in dependency order:
- extension / PostgreSQL-version presence guards
- `catalog.immutable_*` + `set_updated_at()` functions (before the tables — the Computed `search_vector` column and the functional indexes reference them)
- `records` `updated_at` trigger
- conditional pgvector HNSW index (from `0011`)
- `geolens_readonly` role + data-schema grants (from `0023`)

Enterprise SAML columns (`idp_*`) and the `'saml'` CHECK value are intentionally kept **out** of OSS core — autogenerate emitted them from the deferred model union, so they're stripped to match the existing core schema (the enterprise overlay's `e002_add_saml_columns` re-adds them).

## Verification
- `pg_dump --schema=catalog` of the old 27-migration chain vs the new baseline is an **identical statement set** (differs only in physical column order — a benign incremental-`ALTER TABLE ADD COLUMN` artifact; model order is more canonical).
- `alembic check` → **no model/migration drift**.
- Full backend suite: **2997 passed**, no migration/schema regressions.

## Also in this PR
- `env.py`: register the pgvector `vector` type for reflection + enable `compare_type` / `compare_server_default` so `alembic check` is a reliable drift gate.
- `test_alembic_upgrade_clean_db.sh`: run `alembic check` after the clean-DB upgrade (permanent CI drift gate).
- Remove 6 static-analysis tests pinned to now-deleted migration files (the schema they asserted is covered by the baseline + the alembic-check gate).

@codex please review.


ishiland:
@codex review

PR #180, opened 2026-06-04
chore(examples): remove orphaned postman + sdk examples

ishiland:
Audit of `examples/` for necessity and accuracy. Both removed items were **accurate but unreferenced + untested**.

- **`examples/postman/`** — OGC API Postman collection. All 4 routes still resolve, but linked from nowhere, no test/CI, duplicating the live OpenAPI / docs site.
- **`examples/sdk/python/upload_shapefile_publish_map.py`** — titled an SDK example but never uses the typed `geolens.api.*` surface; it hand-rolls raw httpx to literal route strings, so it misrepresents the SDK and would silently rot on any API rename. Orphaned + untested.
- `.env.example` — drop the now-dangling `examples/sdk/python` reference.

### Kept
`examples/manifests/` is unchanged — README-referenced **and** test-covered (`cli/tests/test_manifest_examples.py`, **3 passed** after deletion; deliberate http/s3/publication-intent coverage).


ishiland:
@codex review

Issue #186, opened 2026-06-05
Terrain from a small-extent DEM renders as a raised pedestal — add guard rails + mask raster-DEM nodata


ishiland:
## Summary

Enabling 3D terrain from a **small-extent DEM** produces a raised "pedestal/slab" with vertical cliffs at the DEM boundary, because MapLibre treats everywhere outside the DEM tiles as 0 m elevation. This is largely inherent MapLibre behavior, but GeoLens offers no guard rail, so it can look broken when the user zooms out past the DEM footprint.

## What happens

- A DEM covering only a small AOI (e.g. a ~3 km swissALTI3D patch around the Matterhorn) deforms the terrain mesh only where it has tiles; elsewhere elevation = 0 m.
- The patch's *lowest* point (~2000 m valley floor) is already ~2 km above the surrounding 0 m basemap, and terrain `exaggeration` amplifies it — so the whole patch sits on a ~2–3 km-tall pedestal with near-vertical cliffs at the tile edge.
- In tight framing (camera inside the DEM extent) it looks correct; zoomed out, the patch reads as a floating slab/wall.

This is **expected** MapLibre partial-DEM behavior, not a styling bug — filing as an enhancement for guard rails.

## Suggested improvements

- Warn when a terrain DEM covers only a small fraction of the current viewport (and/or optionally clamp the camera to the DEM bounds).
- Ensure raster-DEM `nodata` is masked in the terrainrgb encoding. swissALTI3D tiles use `nodata = -9999`; unmasked, tile-edge nodata could encode as extreme elevation.
- Docs: recommend draping a high-res DEM over a coarse global DEM (e.g. Copernicus GLO-30) for continuous elevation when the AOI is small.

_Found while building a swissALTI3D Matterhorn terrain map; see the pedestal effect in internal `matterhorn-dem-boundary.jpeg`._

ishiland:
Shipped in **v1.3.0**. 3D terrain now masks raster-DEM nodata in the terrainrgb encoding (no more boundary spikes from a `-9999` fill) and the builder warns when the active DEM covers only a small slice of the viewport. The camera-clamp was deferred (documented). See CHANGELOG [1.3.0].

PR #187, opened 2026-06-05
fix(vrt): flag DEM-tile mosaics as is_dem so terrain/hillshade work

### LINKED ISSUE #185: VRT mosaics of DEM tiles aren't flagged is_dem — terrain & hillshade impossible via the UI
## Summary

A **VRT mosaic of DEM tiles is not flagged `is_dem`**, so it cannot be used for 3D terrain or hillshade — and there is **no UI control to fix it**. Mosaicking DEM tiles into a VRT is the documented way to build a seamless high-resolution terrain surface, so this fully blocks terrain-from-mosaic for UI-only users.

## Root cause

- Direct raster ingest auto-detects DEMs: `cog.py:85` sets `is_dem_candidate = src.count == 1 and _is_float_dtype(...)`, and `tasks_raster.py:156` persists it (`is_dem=meta.get("is_dem_candidate", False)`).
- **The VRT path skips this.** `tasks_vrt.py:92` builds the VRT's `RasterAsset` without running DEM detection and without inheriting `is_dem` from its source datasets, so the mosaic lands with `is_dem = false` even though every source tile is a DEM (and the mosaic is itself single-band float).
- Terrain requires `is_dem === true` (`frontend/.../TerrainControls.tsx:30`), and so does the hillshade render mode (`renderAs.ts:333` → source resolves to `raster-dem` only when `is_dem === true`).
- There is **no frontend control that sets `is_dem`** — the only write path is `PATCH /api/datasets/{id}` `{ "is_dem": true }` (`service_metadata.py:214` `_apply_is_dem`). The layer-options menu offers only Rename/Duplicate/Delete/Add-to-group; Settings → Terrain shows "No terrain layer is active" with no picker.

## Reproduction

1. Ingest two single-band float DEM COGs (e.g. swissALTI3D 2 m tiles) → each gets `is_dem = true`.
2. `POST /api/ingest/vrt/create` with both as `source_dataset_ids` (`vrt_type: "mosaic"`).
3. Add both as layers to a map and `GET /api/maps/{id}`:

   | layer | record_type | is_dem | layer-list icon |
   |---|---|---|---|
   | tile | `raster_dataset` | `true` | ⛰ |
   | VRT | `vrt_dataset` | `false` | ▦ (plain raster) |

4. In the builder the VRT offers only plain raster render modes; Settings → Terrain stays "No terrain layer is active." Only `PATCH /api/datasets/{vrt}` `{"is_dem":true}` unblocks terrain + hillshade.

## Suggested fix

In `tasks_vrt.py`, set `is_dem` on the VRT's `RasterAsset` when the mosaic is single-band float (reuse the `cog.py` detection) **or** when all source datasets are `is_dem`. Optionally also expose a manual "Use as elevation / 3D terrain" toggle in the raster layer's style editor.

## Related (minor)

`GET /api/datasets/{id}` does not surface `is_dem` in its `raster` object — it read `None` even after the PATCH succeeded and the map *layer* reported `is_dem: true`. Clients can't read DEM status from the dataset API.

_Found while building a swissALTI3D Matterhorn terrain map._

ishiland:
Closes #185.

## Problem

A VRT mosaic of single-band float DEM tiles landed with `is_dem = false`: `tasks_vrt.py` built the VRT's `RasterAsset` without running the DEM detection that the raster ingest path uses (`tasks_raster.py:156` ← `cog.py:85` `is_dem_candidate`). Terrain (`TerrainControls.tsx:30`) and the hillshade render mode (`renderAs.ts:333`) both gate on `is_dem === true`, and there is no UI control to set it — so mosaicking DEM tiles (the documented way to build a seamless high-res terrain surface) produced a layer that could not be used as terrain at all.

## Fix

- **Create + regenerate paths** (`tasks_vrt.py`): set `is_dem` from `meta["is_dem_candidate"]` (computed by `extract_raster_metadata` on the assembled VRT), mirroring the raster ingest path. A single-band float mosaic → DEM; a multi-band band-stack → not a DEM. Regenerate recomputes it so add/remove-source flips it correctly.
- **Response** (`RasterMetadata` schema + `helpers.py` serializer): surface `is_dem` on `GET /api/datasets/{id}` so clients can read DEM status (previously it was never exposed, even when set).
- Regenerated `backend/openapi.json` + Python/TS SDKs (field-only diff).

## Testing

- New unit tests in `test_vrt_ingest_tasks.py`: a single-band float mosaic is flagged `is_dem=true`; a multi-band stack stays `is_dem=false`.
- `test_vrt_ingest_tasks.py`, `test_regenerate_vrt_integration.py`, `test_raster_ingest.py` → 24 passed.
- Manually reproduced before the fix (single swissALTI3D tile `is_dem=true`, mosaic `is_dem=false`); after the fix the mosaic is auto-flagged, so terrain + hillshade work without the manual `PATCH /datasets/{id} {"is_dem":true}` workaround.

Does not address the small-AOI terrain "pedestal" UX (#186), which is tracked separately.

@codex review

### DIFF
--- a/backend/app/modules/catalog/datasets/domain/helpers.py
+++ b/backend/app/modules/catalog/datasets/domain/helpers.py
@@ -95,6 +95,7 @@ def _build_raster_metadata(
         res_x=raster_asset.res_x,
         res_y=raster_asset.res_y,
         band_count=raster_asset.band_count,
+        is_dem=raster_asset.is_dem,
         nodata=raster_asset.nodata,
         compression=raster_asset.compression,
         width=raster_asset.width,
--- a/backend/app/modules/catalog/datasets/domain/schemas.py
+++ b/backend/app/modules/catalog/datasets/domain/schemas.py
@@ -118,6 +118,10 @@ class RasterMetadata(BaseModel):
         default=None, description="Pixel resolution in Y (CRS units)"
     )
     band_count: int | None = None
+    is_dem: bool | None = Field(
+        default=None,
+        description="True if this raster is a DEM (single-band float) usable for 3D terrain/hillshade",
+    )
     nodata: str | None = Field(default=None, description="Global nodata sentinel value")
     compression: str | None = Field(
         default=None, description="Internal compression, e.g. DEFLATE, LZW"
--- a/backend/app/processing/ingest/tasks_vrt.py
+++ b/backend/app/processing/ingest/tasks_vrt.py
@@ -101,6 +101,10 @@ async def create_vrt_dataset(
         epsg=meta.get("epsg"),
         band_count=meta.get("band_count"),
         dtype=meta.get("dtype"),
+        # A VRT mosaic of single-band float DEM tiles is itself a DEM. Mirror the
+        # raster ingest path (tasks_raster) so terrain + hillshade light up; without
+        # this the mosaic lands is_dem=false and is unusable as terrain (#185).
+        is_dem=meta.get("is_dem_candidate", False),
         nodata=nodata_str,
         res_x=meta.get("res_x"),
         res_y=meta.get("res_y"),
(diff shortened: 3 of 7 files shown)


ishiland:
@codex review — VRT DEM mosaic fix for #185. Focus areas: (1) `is_dem=meta.get('is_dem_candidate', False)` propagation on both the create and regenerate paths in `tasks_vrt.py`; (2) the `RasterMetadata` field + serializer addition; (3) the new VRT is_dem tests. SDKs + openapi.json regenerated.

PR #188, opened 2026-06-05
chore(showcase): seed-script + map-audit styling fixes

ishiland:
## What

Follow-up to a full quality audit of all 5 saved maps in the local instance. Adds the reproducible **showcase seeder** (`scripts/seed-showcase.py`) with three styling fixes baked in, and records the live-DB fixes applied to the already-seeded maps.

All maps already rendered correctly (0 console errors, terrain/3D/extrusion all bind). The audit's headline CRITICAL/HIGH candidates were **false positives** caught by live verification (legend renders by default on `plugins:null`; the ADK hillshade draws *above* the aerial, not occluded; the share-link "404" is an 8-char display-hint, not a broken token). What remained were polish/correctness items, addressed here.

## Changes in this PR (reviewable)

`scripts/seed-showcase.py` (new) — seeds Manhattan / Income / Matterhorn with corrected styling:

| Map | Fix | Why |
|---|---|---|
| Manhattan | `style_config.breaks` → `[60,120,250,450,750,1200]`, `colors` → full 7-stop Plasma, add `colorLabel:"Roof height (ft)"` | The legend reads `style_config` while the map paints a 7-stop `interpolate`. The old config had **6 colors and dropped `#fdb42f`**, so the legend diverged from the fill above 750 ft. Now they match exactly; the smooth skyline gradient is preserved. |
| Income | `fill-opacity` `0.9` → `1.0` | Layer-opacity `0.9` × fill-opacity `0.9` compounded to `0.81` and desynced from the outline (which tracks the `0.9` layer opacity). One opacity knob now. |
| Matterhorn | add `hillshade-illumination-anchor:"map"` | NW (315°) lighting was rotating with the camera on the bearing -150 / pitch 66 view; now geographically fixed. |

## Applied live (operational — not files; the 2 ADK maps are hand-built, no source)

- The 3 fixes above were also PATCHed onto the existing seeded maps (stable IDs + the public share token preserved; not re-seeded).
- **ADK 3D Relief** (`535310d2`): land-classification layer `show_in_legend:false` (it was ~0.014 effective alpha — invisible — but advertised in the legend); generated the missing thumbnail + OG image (was 404, now 200).

Verified live: Manhattan legend now shows 7 bins incl. "750 – 1.2K" / "≥ 1.2K"; income fill-opacity 1.0; Matterhorn anchor `map`; ADK thumbnail 200. `style_config` was re-included on every PATCH to avoid the null-on-omit gotcha. DB-verified 0 nulls in both styled columns.

## Deliberately deferred

- **Sidebar terrain badge (M-005, LOW, code).** `map-stack.ts` computes `sourceStatus:'disabled'` for a hillshade-mode DEM used as the terrain source, while the runtime/legend correctly render the mesh (verified live on `535310d2`). The narrow fix (treat enabled+DEM-present as `'active'`) **conflicts with an intentional, explicitly-tested assertion** (`map-stack.test.ts:116` "does not mark stale terrain config active when the source DEM is not in Terrain mode"). Flipping it overrides a deliberate design decision and a guard comment suggests the `render_mode` clause is load-bearing elsewhere — so this needs a maintainer call rather than a drive-by change. **Reviewer input welcome.**
- Null-data guards (defensive only — 0 nulls present; wrapping in a `case` would break the legend's `parseStepOrInterpolate`), never-expiring public share token (intentional for a marketing showcase), and minor cosmetic items (extrusion min-zoom, hidden basemap under the aerial).

## Test plan

- `python3 -m py_compile scripts/seed-showcase.py` ✅
- Live render of all 5 maps via Playwright: 0 console errors; 3D/terrain/extrusion bind; legends present; thumbnails 200.

### DIFF
--- a/scripts/seed-showcase.py
+++ b/scripts/seed-showcase.py
@@ -1,0 +1,379 @@
+#!/usr/bin/env python3
+"""Seed GeoLens with the marketing "showcase" maps.
+
+Builds three capability-showcase maps from public, openly-licensed data:
+
+  1. Manhattan Skyline   - 3D fill-extrusion by real building height + graduated color
+                           (NYC Open Data Building Footprints, Socrata 5zhs-2jue)
+  2. New York Income     - data-driven quantile choropleth
+                           (USDA ERS Atlas of Rural & Small-Town America)
+  3. The Matterhorn      - 3D terrain mesh + hillshade from a VRT mosaic of COG tiles
+                           (swisstopo swissALTI3D 2m lidar, OGD)   [--with-terrain]
+
+Everything here is reproducible against a fresh stack. The flows and the
+non-obvious gotchas they encode were verified live against the running API.
+
+Requires: pip install httpx
+
+GOTCHAS this script encodes (learned the hard way):
+  * A plain GeoJSON URL is NOT a "service" - the service connector (probe/preview)
+    only accepts WFS / ArcGIS Feature Service / OGC API Features. For Socrata or
+    ArcGIS-query GeoJSON, DOWNLOAD the file then use the /ingest/upload path.
+  * Socrata serialises numbers as STRINGS ("53.84"); coerce numeric columns to real
+    numbers before upload or GDAL ingests them as text (breaks graduated styling).
+  * NYC height_roof is in FEET -> heightScale = 0.3048 or the skyline triples.
+  * GeoLens LOWERCASES column names on ingest. Reference the lowercased name in
+    every paint/style expression (Median_HH_Inc_ACS -> median_hh_inc_acs).
+  * A job's terminal status is "complete" (not "completed").
+  * Map camera is set via PUT /maps/{id}; bearing must be within [-180, 180].
+  * A VRT mosaic does NOT inherit is_dem from its source tiles - PATCH the VRT
+    dataset {"is_dem": true} or map terrain silently refuses to engage.
+"""
+
+import argparse
+import io
+import json
+import os
+import sys
+import time
+
+try:
+    import httpx
+except ImportError:
+    print("Missing required package. Install with:\n  pip install httpx", file=sys.stderr)
+    sys.exit(1)
+
+DEFAULT_BASE_URL = "http://localhost:8080"
+
+# --- public data sources -----------------------------------------------------
+
+NYC_BUILDINGS = (
+    "https://data.cityofnewyork.us/resource/5zhs-2jue.geojson"
+    "?$where=height_roof>0 AND within_box(the_geom,40.770,-74.020,40.700,-73.965)"
+    "&$limit=50000"
+)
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review — showcase seeder + map-audit styling fixes. Focus areas:

1. **Manhattan reconciliation (`build_manhattan`)**: the fill-color is a 7-stop `interpolate`, and I changed `style_config.breaks`/`colors` to mirror those stops exactly (`breaks=[60,120,250,450,750,1200]`, 7 colors incl. `#fdb42f`) because the legend renders from `style_config`, not the paint. Is this the right reconciliation, or should the paint instead become a discrete `step` to match a graduated classifier? Note `method:"quantile"` is now a slight misnomer (breaks are manual round numbers).
2. **Income opacity (`build_income`)**: `fill-opacity` 0.9 → 1.0 so the 0.9 layer opacity is the single knob (was compounding to 0.81). Confirm this doesn't unintentionally change the intended look vs. just fixing the outline desync.
3. **Matterhorn**: added `hillshade-illumination-anchor:"map"` — sanity-check the value/placement.
4. **General**: any correctness/robustness issues in the new `scripts/seed-showcase.py` (error handling, the quantile-break math `vals[max(0, round(n*k/6)-1)]`, the `/api` path usage).

Context: the two Adirondack maps are hand-built (not in this script); their live-only fixes (land-class `show_in_legend:false`, regenerated thumbnail) were applied directly. A LOW sidebar terrain-badge fix (`map-stack.ts` `sourceStatus` for hillshade-mode DEM terrain sources) was deferred because it conflicts with an intentional assertion at `map-stack.test.ts:116` — opinions welcome on whether that's worth aligning.

ishiland:
Thanks @codex — addressed in 047fb91d.

The Matterhorn build now:
1. Raises immediately if any manifest entry is `action:"error"` (with the per-entry messages), instead of letting the `if r.get("job_id")` comprehension silently drop them.
2. Resolves each tile's `dataset_id` from its queued job **or** its existing `dataset_id` — the old `job_id`-only filter also dropped `action:"skip"` entries, so a re-run over already-ingested tiles would have built a partial VRT too.
3. Asserts the resolved tile count equals the expected AOI tile count before calling `vrt_create`, so an incomplete set can never reach the mosaic.

No changes to the three styling fixes you reviewed; the deferred sidebar terrain-badge item (`map-stack.test.ts:116` conflict) remains a separate call.

ishiland:
@codex review

ishiland:
Addressed in the latest commit, @codex 👍

Each map's final `set_view` PUT now sends `visibility:"public"` (after layers are added), and the three already-seeded live maps were flipped to public to match — all have `has_non_public:false`, and the Adirondack showcase map was already public. The previous round's manifest-error guard is in too.

PR #192, opened 2026-06-05
feat(report): user-menu entry point + error-only floating button

ishiland:
## Summary

Follow-up to #191. Adds a second, discoverable entry point to the in-app problem reporter and makes the idle UI truly zero-footprint.

The reporter now has **two entry points sharing one wizard**:
- **User menu → "Report a problem"** — always available and discoverable (sits with Settings/Admin/Logout). The primary home; works even when nothing has been captured.
- **Floating button — only when errors are captured.** Previously a persistent quiet button; now it renders *only* once `errorCount > 0`, so the idle builder/app has no corner element, while a real problem still surfaces the proactive red count badge the moment it happens.

This resolves the tension between "minimal / out of the way" and "discoverable + proactive": the menu item gives a permanent home with no clutter, and the badge preserves the feature's core value (a non-technical user notices a problem right when it occurs).

## Implementation
- `lib/report/use-report-dialog.ts` — small zustand store holding the wizard's open-state so both entry points can trigger it. Capture and the wizard itself are unchanged.
- `ReportProblemHost.tsx` — reads open-state from the store; renders the floating button only when there are captured errors.
- `Navbar.tsx` — "Report a problem" item in `UserMenu` calling `openReport()` (covers desktop + mobile, which share the avatar dropdown).
- Host tests rewritten for the new behavior: idle = no FAB, FAB appears on error with a badge, and the wizard opens via the shared store (the user-menu path).

## Verification
- typecheck + eslint + **2763 frontend tests** green.
- Live MCP check on the running app: idle shows **no floating button**; the user menu has "Report a problem" which opens the wizard (area auto-defaults to "Map Builder"); injecting a console error makes the floating button appear with `aria-label "Report a problem (1 errors detected)"` + badge `1`.

### DIFF
--- a/frontend/src/components/layout/Navbar.tsx
+++ b/frontend/src/components/layout/Navbar.tsx
@@ -1,8 +1,9 @@
 import { useState, useEffect } from 'react';
 import { Link, NavLink, useLocation } from 'react-router';
 import { useTranslation } from 'react-i18next';
-import { ChevronDown, User, LogOut, Settings, Shield, Plus, Database, FolderOpen, Map, Menu, Layers, Upload, LogIn } from 'lucide-react';
+import { ChevronDown, User, LogOut, Settings, Shield, Plus, Database, FolderOpen, Map, Menu, Layers, Upload, LogIn, LifeBuoy } from 'lucide-react';
 import { useAuth } from '@/hooks/use-auth';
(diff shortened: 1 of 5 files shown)


ishiland:
@codex review

ishiland:
@codex review

PR #233, opened 2026-06-09
fix(installer): check out release tags by fully-qualified ref to prevent branch shadowing

ishiland:
## Summary

Fixes a **high-severity supply-chain code-execution** finding (Codex security review) in `scripts/install.sh`, the script served at `raw.githubusercontent.com/geolens-io/geolens/main/scripts/install.sh` for `curl … | sh` installs.

The installer resolved the latest release from `refs/tags/*` but then passed the **bare** tag name to `git clone --depth 1 --branch "$ref"`. Git resolves `--branch <name>` as a **branch first** and only falls back to a tag, so an attacker who can push `refs/heads/v1.2.2` (shadowing the legitimate `refs/tags/v1.2.2`) would have their branch built and run by fresh installs — while the installer announces `Installing release v1.2.2`. The malicious checkout controls `docker-compose.yml` / `Dockerfile` / app code and reads the freshly generated `.env` secrets.

## Fix

Release tags are now checked out via an explicit, fully-qualified `refs/tags/` fetch into a detached HEAD, which **cannot** be shadowed by a same-named branch:

```sh
git init -q "$INSTALL_DIR"
git -C "$INSTALL_DIR" remote add origin "$REPO_URL"
git -C "$INSTALL_DIR" fetch -q --depth 1 origin "refs/tags/$ref:refs/tags/$ref"
git -C "$INSTALL_DIR" checkout -q --detach "refs/tags/$ref"
```

- The auto-resolved release path (the reported vector) now uses the safe checkout.
- An explicit `GEOLENS_REF` that names a **real tag** takes the same shadow-safe path (closes the same bug class for pinned-tag installs).
- A non-tag `GEOLENS_REF` (e.g. `main`) still uses `clone --branch` to **track** the branch — no behavior change.
- The "no release tag resolved → clone default branch" fallback is unchanged.

## Verification

Reproduced against a local remote where `refs/tags/v9.9.9` (release) is shadowed by `refs/heads/v9.9.9` (malicious), then ran the **real** `install.sh` with a docker shim (10/10 ground-truth assertions):

- **Before/repro:** bare `git clone --branch v9.9.9` checks out the malicious **branch**.
- **After (auto-resolve):** installs the release **tag**, exits 0, still prints `Installing release v9.9.9`.
- Explicit tag pin → release tag (shadow-safe); `GEOLENS_REF=<branch>` → branch tracked (no regression); no-tags → default-branch fallback.
- `sh -n` clean; `pre-commit run --all-files` + `gitleaks` (tracked tree) green.

## CI scope

`scripts/install.sh` matches none of the `changes` path filters, so the backend/frontend/CLI/security jobs are skipped for this diff on `pull_request`; the `pre-commit` job is the only blocking gate and passes locally.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,7 @@ jobs:
       e2e: ${{ steps.filter.outputs.e2e }}
       cli: ${{ steps.filter.outputs.cli }}
       alembic: ${{ steps.filter.outputs.alembic }}
+      installer: ${{ steps.filter.outputs.installer }}
     steps:
       - uses: actions/checkout@v6
       - uses: dorny/paths-filter@v4
@@ -66,6 +67,9 @@ jobs:
               - 'backend/scripts/test_alembic_upgrade_clean_db.sh'
               - 'backend/app/models/**'
               - 'db/**'
+            installer:
+              - 'scripts/install.sh'
+              - 'scripts/tests/test-install-tag-resolution.sh'
 
   # ---------- backend ----------
   backend-lint:
@@ -129,6 +133,22 @@ jobs:
 
       - name: Secret scan (tracked tree)
         run: gitleaks dir . -c .gitleaks.toml --redact --no-banner
+
+  # ---------- installer ----------
+  # Regression guard for the release-tag checkout in scripts/install.sh: a remote
+  # branch named identically to a release tag must not shadow the tag. Pure
+  # shell + git (no Docker/DB), so it is fast and only runs when the installer or
+  # its test changes.
+  installer-test:
+    name: Installer Tag Resolution
+    needs: changes
+    if: needs.changes.outputs.installer == 'true' || github.event_name == 'push'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Tag/branch shadowing regression test
+        run: sh scripts/tests/test-install-tag-resolution.sh
 
   openapi-snapshot:
     name: OpenAPI Snapshot Drift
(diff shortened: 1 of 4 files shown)


ishiland:
## Post-review QA (adversarial multi-lens) → one fix applied

Ran a 4-lens adversarial review + completeness critic over the fix, test, and CI. Consensus: **core security fix is correct and KISS-compliant** (verified `git ls-remote --tags --refs` strictly excludes branches; the fully-qualified `refs/tags/` fetch is unshadowable; the explicit-`GEOLENS_REF` hardening closes the same RCE on the supported pin path — not scope creep).

**One real regression found and fixed (commit `552d8def`):** the `git init`+`fetch` sequence was *non-atomic* — a failed/interrupted fetch left a half-initialized `INSTALL_DIR` that the next run mis-detected as an existing checkout and then died at `cp .env.example .env` (strictly worse than the old `git clone`, which cleans up on failure). Fixed by building in a temp dir and `mv`-ing into place only after checkout succeeds. Reproduced the wedge end-to-end with a fetch-failure fault injection, then confirmed the fix leaves no leftover dir and re-runs work.

**Validation (ground truth, run directly — not delegated):**
- Tree-equivalence: the tag checkout yields a **byte-identical working tree (2274 files)** to `git clone --branch v1.2.2`, same commit, both detached HEAD — i.e. the installed result equals the previously stack-validated path.
- Real installs: auto-resolve (`v1.5.9`) + `GEOLENS_REF=v1.2.2` (annotated, HEAD `52aec45`) both exit 0; idempotent re-run detects the existing checkout; **0 leftover temp dirs**.
- Regression test: 10/10 green; proven to **fail** against the pre-fix installer; robust to hostile global git config (gpgsign/default-branch).

**Deferred as low-priority follow-ups (pre-existing, not worse-than-before):**
- `awk -F/ '$NF'` basename parse lets a hypothetical `*/vX.Y.Z` tag mis-announce + fail-closed (DoS only, no RCE; needs repo push access; canonical repo has no such tags).
- A transient `ls-remote` failure during explicit-pin classification can fall back to `clone --branch` (narrow window; the auto-resolve path degrades safely to default-branch + warning).

Public installer (`getgeolens.com/install.sh`) re-sync prepared in getgeolens-io/getgeolens.com#23 (blocked on this PR).

Issue #203, opened 2026-06-07
Enforce metadata completeness for DCAT-US 3.0 conformance (publish gate + backfill)


ishiland:
## Summary

Add an option to enforce dataset metadata completeness at publish time so the DCAT-US 3.0 (and W3C DCAT 3) feeds are conformant for deployments that need it. Roadmap enhancement — not blocking current work.

## Background

DCAT-US 3.0 makes `description` and `contactPoint` mandatory on a Dataset. By design, GeoLens does **not** fabricate contact metadata (see `backend/app/standards/dcat_us/README.md`) — it serializes what's present and ships `/datasets/dcat-us/3.0/validation/` so operators can see which datasets are missing required fields. As a result, datasets published without a summary or a contact serialize to non-conformant DCAT-US.

The enforcement machinery already exists and is aligned: `validate_record` (`backend/app/modules/catalog/validation/service.py`) requires title, summary, license, spatial extent, lineage, CRS, ≥1 contact, and ≥1 keyword before publish — which is a superset of the DCAT-US required fields. It only runs when `REQUIRE_METADATA_FOR_PUBLISH` is enabled, and that flag defaults to `False` (`backend/app/core/persistent_config.py`).

## Decision needed (product fork)

Pick the posture before implementing:
1. **Enforce at publish** — default `REQUIRE_METADATA_FOR_PUBLISH=True`. Strongest conformance; adds onboarding friction (conflicts somewhat with low-friction early-release posture).
2. **Filter the feed** — emit only valid records in the DCAT/DCAT-US feeds, skip incomplete ones. No onboarding change; weaker catalog coverage.
3. **Accept incomplete** — keep current behavior; rely on the validation endpoint as an operator tool. Status quo.

## Scope (assuming option 1)

- [ ] Flip `REQUIRE_METADATA_FOR_PUBLISH` default (or make it deployment-configurable with a documented recommendation)
- [ ] Update test fixtures that publish datasets without full metadata
- [ ] Backfill existing published records (summary + a default org-level `pointOfContact`) — script + decision on what contact to use; do **not** invent federal emails
- [ ] Add an end-to-end round-trip test: ingest → PATCH editable fields → assert each surfaces in DCAT, DCAT-US, STAC, and OGC Record
- [ ] Add a W3C DCAT 3 validation guard (DCAT-US already has one; DCAT 3 has none)
- [ ] Docs: note the conformance posture in the DCAT-US README

## Notes

[...]

Issue #202, opened 2026-06-07
Add GeoDCAT-AP serialization profile (EU/INSPIRE discovery)


ishiland:
## Summary

Add a **GeoDCAT-AP** serialization profile alongside the existing DCAT 3 (W3C) and DCAT-US 3.0 outputs. This is a later/backlog task, not blocking current work.

## Motivation

Generic DCAT 3 is the shared core, but government data portals harvest *application profiles*, not raw DCAT. GeoLens already serves:

- DCAT 3 (W3C) — generic baseline
- DCAT-US 3.0 — US federal / data.gov

The gap for the EU / INSPIRE geospatial market is **GeoDCAT-AP**, the geospatial profile of DCAT-AP that bridges ISO 19115 / INSPIRE metadata into DCAT-AP. GeoLens already stores most of the required upstream fields (contact roles, lineage, constraints, maintenance frequency, spatial/temporal extent, CRS), so this is largely a mapping + serialization task rather than a model change.

## Scope

- [ ] Add a `backend/app/standards/geodcat_ap/` serializer (`record_to_geodcat_ap`, `catalog_to_geodcat_ap`) mirroring the structure of `standards/dcat/` and `standards/dcat_us/`
- [ ] Map ISO 19115 model fields → GeoDCAT-AP terms (lineage, constraints, contact roles via CI_RoleCode, maintenance frequency, spatial resolution/CRS)
- [ ] Expose catalog + per-dataset endpoints in `catalog/datasets/api/router_export.py` (e.g. `/datasets/geodcat-ap/` and per-dataset variant), following the existing DCAT-US route pattern
- [ ] Add schema validation (mirror `dcat_us/validation.py` → `validate_geodcat_ap`) and a `/validation/` endpoint
- [ ] Tests: serialization correctness, namespace/context, ISO field mapping, and a round-trip (ingest → edit → GeoDCAT-AP output passes validation)
- [ ] Docs: note the new profile in standards/discovery docs

## Notes

- Pin the targeted GeoDCAT-AP version + schema commit the way DCAT-US does (`DCAT_US_SCHEMA_VERSION` / `DCAT_US_SCHEMA_COMMIT`).
- Priority depends on whether EU/INSPIRE gov is a target market — defer until that's confirmed.

ishiland:
[...]

PR #257, opened 2026-06-18
v1.3.0 release candidate: v1039–v1043 + raster-ingest & public-viewer fixes (review only, do not release)

ishiland:
## v1.3.0 release candidate — for review, **do not release yet**

This is the `feat/v1043-self-hosted-release` branch (36 commits on top of `origin/main` @ `bb63bccb`). Opening it for a Codex review pass before any merge or release. **The v1.3.0 cut (merge + `v1.3.0` tag + publish) is deliberately deferred** — please review for correctness first.

### What's in here

Four already-audited internal milestones, framed for v1.3.0:

- **v1039 Tier-1 hardening** (38 findings) — correctness/security/perf/ops across paired GET/POST + commit/preview surfaces.
- **v1040 Tier-2 hardening** (40 findings).
- **v1041 map-builder authoring** — data-driven classification (Jenks/stddev/manual), CVD-safe + reversible ramps, per-layer legend customization (`0004_add_maps_legend_title`), layer search, raster/DEM authoring fidelity.
- **v1042 dormant tenancy substrate** — `single_tenant` is the only active mode; no behavior change (migrations `0005`–`0007`).

Plus the v1043 theme (self-hosted release/upgrade excellence): version coherence, prebuilt-image upgrade flow, migration safety, install/compose correctness, backup/DR honesty, docs accuracy. And 4 pre-release tweaks (Collections/Maps create-button dedup, GH #186 terrain guard-rails, GH #202 GeoDCAT-AP profile, GH #203 conformant-by-filtering DCAT feeds).

### ⭐ Newest work — please focus review here (2 bugs found by a clean-slate rebuild + smoke rehearsal)

A full `docker compose down --volumes` → `--no-cache` rebuild → reseed → extensive Playwright MCP smoke surfaced two real bugs that mock-heavy tests + warm QA had missed:

1. **`86eb7a9c` — `fix(ingest): resolve dataset_asset_orm_class via CatalogPort, not ProcessingPort`**
   The v1040 BUG-041 fix (`f279c9c4`) called `dataset_asset_orm_class()` on the **ProcessingPort**, but that ORM accessor lives on the **CatalogPort** → **every raster/COG/VRT ingest crashed** with `'DefaultProcessingPort' object has no attribute 'dataset_asset_orm_class'`. This has been broken on `main` since 2026-06-12. Fixed + a real-port (non-mocked) regression test that also fails-before/passes-after. Proven live: 9/9 COG ingests failed → all succeeded; `dataset_assets` 0 → 27 rows.

2. **`8bef9681` — `fix(viewer): idle-retry the data-layer sync on cold style load`**
   Public/shared map viewers opened via a hard page load (the normal shared-link path) could render with **no data layers** — only basemap + (for terrain maps) the 3D mesh. The Matterhorn showcase appeared blank to anonymous visitors while rendering fine in the builder. Root cause: the main data-sync effect early-returned on `!isStyleLoaded()` with no idle-retry; terrain widens that window. Fix mirrors the sibling visibility effect's BUG-037 idle-retry. **Verified live (Playwright MCP):** before = 0 data layers, after = 5 incl. the DEM hillshade, consistent across cold reloads; viewer suite 31/31 green; typecheck clean. *No unit test — the jsdom map mock can't reproduce the real style/effect timing race (the same reason the bug was subtle); the live before/after is the evidence. Reviewer scrutiny welcome on this one.*

### Verification done locally
- Clean-slate rebuild boots to all-healthy; migrations `0001`→`0007` from scratch.
- Reseeded 3 showcase maps (Manhattan 3D extrusion, NY income choropleth, Matterhorn 3D terrain).
- Extensive Playwright MCP smoke: login, catalog/search + pgvector similarity, dataset detail + signed tiles, 2D/3D/terrain viewers, builder load + reload durability, admin control plane, anonymous public viewer, mobile. 0 console errors (1 benign MapLibre terrain-maxzoom warning).

### NOT done (intentionally)
- No merge, no `v1.3.0` tag, no publish — all human-gated, pending this review.
- CI-on-push items (image-pull prod smoke, enterprise-heads migration, MinIO backup round-trip) run on push; not locally executed.


ishiland:
@codex review

ishiland:
@codex review — pushed fixes for all 4 review findings (P1 MinIO fail-closed via runtime entrypoint guard; P2 prod-smoke v-prefix + db-pull; P2 upgrade.sh containerized pg_dump) plus the 3 CI failures (alembic `heads`, valid minio image, layering CODE-08/CODE-01, and the -n4 shared-engine loop leak).

ishiland:
## Round 3 — CI failures + remaining Codex findings addressed

Pushed `3074cf5f`, `0a2b2215`, `2bdba41d`. Status of every inline finding:

**CI failures (both pre-existing release debt surfaced by this branch's first full CI run, not the session's product fixes):**

- **Backend Tests / `Alembic drift check (MIG-03)`** — the gate flagged *all* catalog tables as `add_table`. Root cause was **search_path ordering, not real drift**: the setup step sets the postgres DB default to `catalog, data, public` (catalog **first** = connection-default schema). The ORM models are explicitly `schema="catalog"` and `env.py` runs the check with `include_schemas=True`, so catalog-as-default makes alembic reflect every table under the unqualified/`None` key while the metadata is keyed under explicit `catalog` → wholesale mismatch. Added a step that resets the default to `data, public` **after** migrations and **before** the check. Migrations are untouched (each is schema-qualified; `0002_procrastinate` sets its own search_path). A real drift column still trips the gate (exit 255).
- **Backup Restore Round-trip / apt** — `awscli` was withdrawn from the Ubuntu 24.04 (noble) archive. The `ubuntu-latest` runner already ships AWS CLI v2 and the only `aws` use is `s3 mb/cp` against MinIO, so dropped `awscli` from the apt line.

**Codex inline findings:**

| Finding | Status |
|---|---|
| `docker-compose.prod.yml:422` — mirror MinIO fail-closed guard (P1) | ✅ **Fixed** (`0a2b2215`) — the dev fix (b0d39ab5) missed prod; added the identical runtime entrypoint cred-guard. |
| `prod-smoke.yml:22` — race image publication (P2) | ✅ **Fixed** (`2bdba41d`) — bounded pull retry (10×30s). |
| `.env.example:560` — bare semver in deploy docs (P2) | ✅ **Fixed** (`2bdba41d`) — `vX.Y.Z` → `X.Y.Z`. |
| `docker-compose.yml:553` — restore `${MINIO_ROOT_USER:?required}` (P1) | ⛔ **Declining** — see below. |
| `prod-smoke.yml:52` — strip `v` prefix (P2) | ✅ Already fixed (b2313327, `VER="${VER#v}"`). |
| `prod-smoke.yml:88` — skip build-only `db` in pull (P2) | ✅ Already fixed (b2313327, `db` omitted). |
| `scripts/upgrade.sh:45` — drop host `pg_dump` preflight (P2) | ✅ Already fixed (b2313327, backup runs in-container). |

**Declining `docker-compose.yml:553` (restore `${MINIO_ROOT_USER:?required}`):** `${VAR:?}` is evaluated at compose **parse time** for the entire file even when the `cloud-dev` profile is inactive — that aborts a default `docker compose up` / the install one-liner for every operator using a verbatim `.env.example` (MinIO creds are only set under `cloud-dev`). This is the INST-01 gotcha v1043 deliberately removed. The dev compose already fail-closes correctly via a **runtime entrypoint guard** (`docker-compose.yml:530-538`) that `exit 1`s on blank creds — and `:-` alone isn't enough because MinIO silently boots `minioadmin:minioadmin` on an empty var, which the guard closes. The prod compose now has the same guard.

@codex review

ishiland:
## Full CI green — two additional onion-layer fixes after the last review

CI run [`27785971096`](https://github.com/geolens-io/geolens/actions/runs/27785971096) is **green (25 pass / 2 skip / 0 fail)** at `80f696fa`. Two more pre-existing CI-debt failures surfaced and were fixed after the previous `@codex review` (which ran on `2bdba41d`):

- **`778cb555`** `fix(ci): install postgresql-client-17 via PGDG` — dropping the withdrawn `awscli` unblocked the Backup Restore apt step, revealing `pg_dump: server version mismatch` (host client v16 vs the `postgis/postgis:17-3.5` service). Installed the matching v17 client; the full S3/MinIO offset round-trip now passes.
- **`80f696fa`** `test(conftest): pre-dispose shared engine pool` — Pytest Parallel Isolation flaked on the known `-n4` "got Future attached to a different loop" race (`test_schema_skew_guard`). `5c1f6b12` disposed the shared async-engine pool only *after* in-scope tests, which can't protect the first DB-touching test from a connection left by a prior test (any module) on the same worker. Added a *before*-dispose so each in-scope test starts with a fresh pool bound to its own loop.

All seven of the earlier inline findings remain addressed (see the prior comment). Branch is push-only for review — not merging / not tagging a release.

@codex review

ishiland:
## Three additional re-review findings addressed (all P2)

- **`docker-compose.prod.yml:375` / `docker-compose.yml`** — the BKP-01 `upload_staging:/staging` mount existed only in the prod compose; mirrored it into the **source** compose backup service so `--profile backup` on a source-build install no longer silently produces DB-only backups (`backup-entrypoint.sh` skips the archive when `/staging` is unmounted). `ce7eaf9c`
- **`.env.example:568`** — the `GEOLENS_VERSION` knob still showed `v1.2.3` as valid; documented bare semver only (compose uses the value verbatim; the manual path bypasses install.sh's v-strip). `ce7eaf9c`
- **`scripts/upgrade.sh:160`** — the prebuilt upgrade now fetches the target tag and does a **path-restricted checkout** of the compose files + `scripts/lib` + the container-mounted helper scripts **before** `compose pull`, so new images get the new release's config. It deliberately **excludes** the running scripts (upgrade.sh/restore.sh/install.sh — no mid-run self-swap) and never touches `.env`; a non-git install or any git failure warns and continues pull-only. Covered by `test-upgrade-order.sh` (17/17, new assertion that the sync runs before the pull). `c8b3649f`

Branch is push-only for review — not merging / not tagging a release.

@codex review

ishiland:
## Two re-review findings on the upgrade sync addressed

- **P1 (data safety) `upgrade.sh`** — the migration ran while the old `api`/`worker` were still up. Now `api` + `worker` are stopped right after the backup (and after the rollback trap is armed) to quiesce writers before the schema migration; `db` stays up for migrate, and the final `compose up -d` restarts them on the new images. `f1183a64`
- **P2 `upgrade.sh:45`** — the unconditional `need_command git` aborted Docker-only explicit-target pull-only upgrades. git is now required only for latest-tag resolution; the best-effort sync is guarded by `command -v git` and continues pull-only without it. `f1183a64`

Test: `test-upgrade-order.sh` 18/18 — full order is `backup -> stop api/worker -> pull -> migrate -> app_up`.

Branch is push-only for review — not merging / not tagging a release.

@codex review

ishiland:
## Two re-review refinements on the writer-quiesce (upgrade.sh)

- **P1 (data safety)** — moved the `api`/`worker` quiesce to **before** the `pg_dump` (pg_dump is a start-time snapshot and doesn't block writers, so a post-dump write could be lost on rollback). Added `restart_writers()` to bring the app back if the backup itself fails (before the rollback trap is armed), so a dump failure never leaves the app down.
- **P2** — replaced the 120s migrate poll with `docker wait`, which blocks until the one-shot actually exits and returns its real code — so a long production migration is never falsely failed mid-flight.

`test-upgrade-order.sh` 18/18; order is `stop → backup → pull → migrate → app_up`. `e8ec954a`

Push-only for review — not merging / not tagging a release.

@codex review

ishiland:
## Final hardening pass — writer-quiesce now fail-closed

- **P1** — the pre-dump `compose stop api worker` now **aborts** the upgrade (after restarting whatever it stopped) if the stop fails, rather than warning and dumping under active writers. So the no-writers precondition is a hard guarantee before the rollback backup is taken. Covered by `test-upgrade-order.sh` CASE 6 (20/20).

This completes the upgrade-flow hardening from the review. All blocking and substantive findings across the review rounds are addressed; the PR is green. Any remaining suggestions are fine-grained edge-case hardening and are non-blocking for v1.3.0.

@codex review

PR #259, opened 2026-06-19
fix(quicklook): scale vector simplify tolerance to data extent

ishiland:
## Problem
Vector dataset quicklooks (the catalog thumbnails) rendered **blank** for any dataset made of small features — building footprints, parcels, POIs. The endpoint returned `200` but served a blank canvas.

## Root cause
`generate_vector_quicklook` simplified geometry with a **fixed** `ST_Simplify(geom_4326, 0.01)` tolerance (~1.1 km). Sub-kilometre features collapse to empty geometry at that tolerance, so nothing gets drawn. Large polygons (e.g. counties) are unaffected, which is why the bug only surfaced on fine-grained datasets.

Measured on a ~22k-building dataset (2000-feature sample):

| simplify tolerance | non-empty geometries |
|---|---|
| `0.01°` (previous) | 0 → blank |
| `0.0001°` | 868 |
| `0.0` (none) | 2000 |

## Fix
Scale the simplify tolerance to the data extent and render resolution (`extent / (size * 2)`) instead of hard-coding `0.01`. Small features now survive simplification; large, detailed polygons still get vertex decimation for render speed.

## Testing
- Existing quicklook tests pass: `test_vector_quicklook_pure`, `test_quicklook_predicate`, `test_quicklook_async_context`, `test_vector_quicklook_offload_perf007` (34 total).
- Verified end-to-end on a live instance: a Manhattan building-footprints dataset went from a blank 760-byte quicklook to a correctly rendered thumbnail.

### DIFF
--- a/backend/app/processing/vector/quicklook.py
+++ b/backend/app/processing/vector/quicklook.py
@@ -146,27 +146,29 @@ async def generate_vector_quicklook(
     est_rows = bounds_row.est_rows or 0
 
     # Simplify tolerance scaled to the render resolution (~half a pixel at `size`),
-    # capped at the original fixed 0.01 degrees so it is never coarser than the prior
-    # behaviour. A fixed 0.01 degrees (~1.1 km) collapsed sub-kilometre features —
-    # building footprints, parcels, POIs — to empty geometry, so those datasets
-    # rendered a blank thumbnail. Scaling to the extent keeps small features visible
-    # at city scale; the cap avoids over-simplifying small features scattered across a
-    # very wide (e.g. continental) extent, where extent/(size*2) would exceed 0.01.
+    # used only to decimate vertices on large, detailed polygons. The third
+    # ST_Simplify argument (preserveCollapsed=true) keeps sub-tolerance features —
+    # building footprints, parcels, POIs — as minimal geometry instead of letting
+    # them collapse to empty. Previously a fixed 0.01-degree tolerance (without
+    # preserveCollapsed) wiped such features out, so those datasets rendered a blank
+    # thumbnail. With preserveCollapsed the tolerance can scale freely with the extent
+    # without ever dropping small features — including small features scattered across
+    # a continental extent, where the tolerance is large.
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review

ishiland:
Good catch — addressed in 63e0951b.

The extent-scaled tolerance only exceeds the original `0.01°` once the bbox is wider than ~5.12° (`extent / (size*2) > 0.01`), which is exactly the regression you described for medium-small features scattered across a wide extent. Fixed by clamping:

```python
simplify_tol = min(extent / (size * 2), 0.01)
```

This makes the tolerance **never coarser than the prior fixed `0.01°`** — identical to the old behaviour for very wide extents (no regression), and strictly finer for city/parcel-scale extents (the original blank-thumbnail fix). I went with a clamp rather than a feature-scale derivation to avoid an extra per-dataset query; the cap guarantees we never do worse than `main` does today.

Quicklook tests still green (34 passing). @codex review

ishiland:
@codex review

ishiland:
Right again — the clamp only avoided being *worse* than `main`; at a continental extent it still pinned the tolerance at the known-collapse `0.01°`. Fixed at the root in 6a238d7b by dropping the clamp and using `ST_Simplify`'s **`preserveCollapsed=true`** flag, so sub-tolerance features are kept as minimal geometry instead of collapsing to empty — at *any* extent.

Verified through the full render pipeline (PostGIS `ST_Simplify` → `ST_AsGeoJSON` → Shapely `shape`/`make_valid`) on a ~22k-building dataset (2000-feature sample):

| tolerance | preserveCollapsed | non-empty features |
|---|---|---|
| `0.01°` | false (old) | 0 → blank |
| `0.01°` | **true** | 2000 |
| `0.117°` (continental-scale) | **true** | 2000 |

So small features now survive even when the extent-scaled tolerance is large, while large detailed polygons are still decimated for render speed. The clamp is no longer needed and was removed. Quicklook tests green (34). @codex review

ishiland:
@codex review

PR #260, opened 2026-06-19
docs(readme): add a "Try the live demo" call-to-action

ishiland:
Adds a prominent **▶ Try the live demo** call-to-action to the top of the README, between the value proposition and the badges (above the `curl … | sh` install line).

### Why
At 0 stars the biggest conversion gap for a self-hosted tool is that a visitor can't *try it* without installing Docker. A one-click hosted demo is the single most-cited friction-killer — it turns a 30-second skim into a star. The demo (https://demo.getgeolens.com) is now live, hardened (registration off, AI off, read-only-ish), and self-healing nightly.

### What
- A `for-the-badge` button linking to the demo
- A sub-caption noting it's zero-install and the shared guest login (`guest` / `GeoLensDemo1!` — an intentionally-public editor account on a nightly-reset instance)

Verified the demo URL returns 200 over valid TLS and the badge renders. Docs-only change.

### Follow-up (not in this PR)
A 5–10s GIF of the drag-to-style / "Ask AI" flow — best captured with a real screen recording rather than generated; happy to do that next.


ishiland:
@codex review

ishiland:
@codex review

PR #266, opened 2026-06-20
v1045 — Close the Loop: Notifications, Self-Serve Signup & v1.4.0 GA

ishiland:
Closes the loop v1044 opened: the lead funnel captured signups/leads to a DB row that notified no human (zero outbound send path). This milestone ships the outbound notification port, wires real events, generalizes signup with email verification, and cuts the deferred public **v1.4.0** GA (release-ready; the publish push stays human-gated). Connectors (the moat) are deliberately the NEXT milestone — descoped + demand-justified per an 8-agent analysis.

## What's in here (6 phases, 29/29 reqs)
- **NOTIF** — Outbound notification port: `NotificationSink` Protocol + no-op community default, stdlib SMTP + httpx webhook channels, admin test-send. Default-OFF, fail-safe, secret-safe.
- **EVENT** — Wired to real events (signup/lead, ingest complete/failed, health alerts), each opt-in. Sink registered in the shared `bootstrap()` so it works in API **and** the procrastinate worker.
- **SIGNUP** — Email-verified self-serve signup (migration 0009; single-use expiring sha256-at-rest tokens; `allow_signup` default-OFF, admin-gated). Google SSO preserved.
- **REL** — All version sites → 1.4.0, CHANGELOG `[1.4.0]`, release.yml prev-tag hardened to strict semver, version-coherence gate. No release triggered.
- **POLISH/HYG** — New-map thumbnail race fixed (deferred capture until layer sources sync) + backlog hygiene.

## Verification
- Local gates: **2988 frontend vitest + 143 backend notification/signup/worker tests**, OpenAPI no-drift, version 1.4.0 coherent, i18n 4-locale parity.
- Live Playwright MCP close-gate on a rebuilt stack: boot Healthy (migration 0009 applied), `edition=community` (open-core preserved), new-map thumbnail captured live (`PUT …/thumbnail/ → 204`), notification admin UI renders.
- Adversarial code-review ran each phase and caught a real bug every time — all fixed + regression-tested: unverified SMTP TLS, community→enterprise edition mis-detection, suspended-account reactivation via verify link, a capture-timing leak.

## Post-open changes (review + hardening)
After the PR opened, GitHub CI surfaced failures the local gate had masked, and a follow-up review pass hardened the signup path. All landed on this branch; full CI is green (23 ✅ / 0 ❌) and the branch is up-to-date with `main`.

- **CI fixes** — health-alert cooldown used `time.monotonic()` vs a `0.0` sentinel, suppressing the first alert during the first 5 min of process uptime (**real prod bug** — no alert when the DB is down right after a deploy); fixed with a `None` sentinel. Also: `notification_sinks` added to the additive-slot guard set, `email_verification_tokens` CASCADE FK whitelisted, SDK lockfile version synced, and a leaked-global test-isolation fix for the health-alert tests under xdist.
- **Signup outcome is now server-driven** — `RegisterResponse.next_step` (`verify_email` | `await_approval`) replaces client-side inference from a cached `/auth/config` snapshot, eliminating a post-submit race (Codex). `smtp_configured` was removed from `/auth/config` (superseded — avoids shipping an unused public field in GA).
- **Signup enumeration fix** — the register response previously diverged between a genuine new signup ("check your email") and a swallowed collision ("awaiting approval") whenever email-verification was on, leaking account existence (violating the SEC-012 enumeration-safety the code claims). The outcome is now computed once from `(config + submitted email)`, so new-user and collision return **byte-identical** responses; regression test added.
- **Accepted limitation** — a verification email is still delivered only on the new-user path, so a registrant using their own email can infer username existence out-of-band. No clean fix without a signup redesign (always-send leaks via content + enables spam). Documented in code + accepted for GA; tracked in #267.
- **Reviews** — 2 independent Codex passes + a 4-lens adversarial review on these changes (verdicts posted as PR comments). Final Codex HIGH = the accepted out-of-band limitation above.

## Human-gated (not done here)
- Cut v1.4.0: after merge, `git tag v1.4.0 && git push origin v1.4.0` (triggers GHCR/Release/SDK).
- Notifications need env config (SMTP/webhook secrets) + restart to actually send.
- Optional: bump demo `GEOLENS_VERSION` pin `v1044 → v1.4.0`.

## Follow-ups
- #267 — signup username-enumeration via verification-email delivery side channel (accepted for GA).

https://claude.ai/code/session_0172k2YPgRv8L2SeGw18kv6n


ishiland:
## 🤖 Codex review (v1045)

> Independent automated review of `feat/v1045-close-the-loop` — generated by Codex CLI, not a human reviewer.

---

### BLOCKERS (must fix before merge)

None observed.

---

### HIGH (strong recommendation to fix)

**[H1] Verification token redemption is not atomic** — `backend/app/modules/auth/verification.py:92,105`

`redeem_verification_token()` selects an unconsumed token and only later sets `consumed_at`; two concurrent requests with the same token can both pass the check before either commits. Replace the SELECT+UPDATE pattern with a conditional `UPDATE ... WHERE token_hash = ... AND consumed_at IS NULL AND expires_at > now() RETURNING user_id`, or add a `SELECT FOR UPDATE` row-lock on the token.

---

### MEDIUM (should fix)

**[M1] Register UI can show email-verification prompt when backend is in admin-approval mode** — `frontend/src/pages/RegisterPage.tsx:122`

The frontend branches on `config.email_verification_required`, but the backend silently falls back to admin approval when SMTP is unconfigured (`backend/app/modules/auth/router.py:333`). In a no-SMTP deployment the user sees "check your email / resend" even though no email was sent.

**[M2] Notification test-send audit event is never committed** — `backend/app/modules/settings/router.py:558–573`

`send_test_notification()` calls `audit_emit(...)` but returns before `db.commit()`, and the DB dependency does not auto-commit. The admin audit row for this action is lost on every call.

---

### LOW / NITPICKS

**[L1] `/resend-verification` lookup is case-sensitive despite its comment** — `backend/app/modules/auth/router.py:472–475`

The predicate is `User.email == body.email` (exact match). Users who registered with different casing get the generic 200 with no resend. Normalize both sides with `lower()` or use `ilike`.

---

### LOOKS GOOD

- SMTP credentials use `SecretStr`, never appear in logs or API responses, and TLS uses `ssl.create_default_context()` — no verification bypass observed.
- Verification tokens stored as SHA-256 hashes; expiry enforced server-side; invalid/expired/consumed attempts all return the same generic 400.
- `/resend-verification` returns the same 200 for known and unknown emails (enumeration-safe). Delivery failures log error type + user id only, not the token.
- Verifying an email does **not** reactivate a suspended account; activation is gated on `status == "pending"` and `is_active == false`.
- Notification dispatch is fail-safe: `emit_event_safe()` suppresses sink failures, and ingest terminal-status notifications fire after the status is committed — can never roll back ingest state.
- Open-core boundary looks preserved: `_overlay_extension_names()` filters the `notification_sinks` slot; built-in sinks are registered post edition-resolution, so an OSS deploy will not detect as enterprise.
- Built-in sink registration is idempotent per process; duplicate multi-worker registrations not observed as a risk.
- Release workflow strict-semver prev-tag filter looks correct; no `v1.4.0` tag exists locally or in the diff.
- Version bump is confined to expected files (pyproject.toml, package.json, openapi snapshot, SDK artifacts).
- Thumbnail deferred-capture is bounded by a 5s poll with a cancellation flag — no handle leak or double-capture race observed.

---

### VERDICT

**CONCERNS** — No blockers observed, but the concurrent token-consumption race [H1] and the two correctness issues [M1, M2] should be addressed before merge.

ishiland:
## Codex review — addressed (commit `1de9112e`)

- ✅ **H1 (high)** — `redeem_verification_token` now claims the token in one atomic `UPDATE … WHERE consumed_at IS NULL … RETURNING`, closing the concurrent double-redeem race. (32 verification/admin tests green.)
- ✅ **M2 (med)** — `send_test_notification` now `await db.commit()`s so the test-send audit row is persisted.
- ✅ **L1 (low)** — resend-verification email lookup is now case-insensitive (`func.lower`).
- 📋 **M1 (med) — follow-up:** on a no-SMTP deploy, `RegisterPage` shows "check your email" because it branches on `email_verification_required` while the backend silently falls back to admin-approval. Tracked as a fast follow (expose `smtp_configured` in `/auth/config` + branch the page) — backend already logs the fallback (WR-02). Not blocking; default deploys are admin-approval until SMTP is configured.

Thanks Codex — H1 was a real concurrency hole the prior review pass missed.

ishiland:
## 🤖 Codex review — CI fixes + M1 (`1de9112e..7a602439`)

> Independent automated review of the two commits added after the prior Codex pass — generated by Codex CLI (not the connector bot). Scope: `faff96de` (green-CI fixes) + `7a602439` (M1 SMTP-aware signup). The empty re-trigger commit `0439aa08` is excluded.

**Verdict: SHIP** — no blockers. The health-alert cooldown sentinel change (`0.0` → `None`), the security posture of exposing `smtp_configured` on the public unauthenticated `/auth/config`, the server↔client branch alignment, and the xdist test-isolation changes all passed without issue.

### Findings

**LOW — `frontend/src/pages/RegisterPage.tsx`** (config-fetch race)
The post-submit view (`VerificationPending` vs `PendingApproval`) is derived from the cached `/auth/config` snapshot. A config change or rolling deploy between the config fetch and the submit could make the UI disagree with the backend's actual register path (e.g. show "check your email" when the backend sent no mail).
- *Suggested fix (Codex):* have the server return the outcome in `RegisterResponse` (machine-readable), or refetch `/auth/config` after a successful submit before choosing the pending view.
- *Disposition:* **Accepted, deferred** — rare race, cosmetic-only (wrong pending message in a narrow window), and both Codex and the local 4-lens review rate it non-blocking for GA. The `RegisterResponse`-outcome fix is the cleaner long-term fix and is tracked as a fast-follow.

### Already green
CI on the head commit is fully green (23 ✅ / 4 skipped / 0 failures); the 5 original red checks + M1 are all passing.

ishiland:
## 🤖 Codex review #2 — server-driven signup outcome + enumeration fix (`7a602439..6628b115`)

> Independent Codex CLI review of the change that replaced `smtp_configured` with `RegisterResponse.next_step` and unified the register response. **Verdict: FIX-FIRST** — both findings addressed below.

### HIGH — username enumeration via verification-email delivery (resolved: accept + document + track)
The HTTP response is now enumeration-safe (byte-identical for new-user vs swallowed collision), **but a verification email is still sent only on the new-user path** — so a registrant supplying their own email can infer username existence out-of-band (email arrives ⇒ username was free).
- Pre-existing (email-only-for-new-user predates this PR); this PR fixed the HTTP-response vector that previously *also* leaked via differing messages.
- No clean code fix: "always send" leaks via differing content and opens an email-spam vector; fully closing it needs a signup/identity redesign.
- **Disposition:** accepted for v1.4.0 GA. Documented at the send site in `router.py` and in the regression test; tracked in **#267**. Mitigated by `/register` rate limiting + requiring `EMAIL_VERIFICATION_REQUIRED` + SMTP.

### LOW — regression test rigor (fixed)
Added `assert r1.content == r2.content` (raw bytes, not just parsed JSON) to the collision-vs-new-user test, catching serialization/field-order/size divergence.

### Clean
`wants_email_verification` computation, `next_step` correctness, `SignupResponse` typing, and SDK/OpenAPI consistency all checked out. Body+status unification verified correct.

---
Also fixed a full-suite FE type-drift guard (`check_fe_type_drift.py`) for the `RegisterResponse` name collision (auth vs dataset type; real mirror is `SignupResponse`).

PR #268, opened 2026-06-21
v1046 — Admin Identity Governance & Surface Hardening

ishiland:
## v1046 — Admin Identity Governance & Surface Hardening

Puts an org's sign-in perimeter under admin control: which providers, which email domains, password-or-SSO — from a hardened, organized admin surface. All in **core (OSS)**; SAML stays an enterprise-overlay feature (core carries only the existing seam).

Built on the existing auth plumbing (v1044 Google SSO + generic OAuth-provider CRUD, v1045 email-verified signup). Scoped from a 5-agent live codebase map, then 5 phases (1235–1239), each verified goal-backward; the security-critical enforcement phase also got an adversarial code review (2 blockers found + fixed) and the whole milestone closed with an orchestrator-driven **live Playwright close-gate** on a rebuilt stack.

### What's in it (13 requirements)

**SSO / login-mode**
- `password_login_enabled` setting (default on). When off ⇒ **SSO-only**: `POST /auth/login` 403s for non-admins; **admins keep password login as break-glass**.
- `GET /auth/config` surfaces the flag so the login page hides the password form without a flash.
- **Lockout guard**: refuses to disable password login (422) unless an enabled OAuth provider exists — extended to provider disable/delete too, so an admin can't lock the org out in two steps.
- **GitHub** OAuth provider (`provider_type='github'`): auto-populated endpoints, primary-**and**-verified email resolution via `/user/emails` (account-takeover guard), login button.

**Domain allowlist** (`allowed_email_domains`, empty = unrestricted; case-fold + `*.sub` wildcards)
- One shared `is_email_allowed` helper enforced at **every** auth path: signup, password login, SSO callback (JIT, before provisioning — TOCTOU-free), refresh, GitHub callback, and admin-create. `manage_settings` admins are exempt (break-glass).

**Admin surface hardening**
- Rate-limit every admin mutation (12 endpoints) + settings/oauth-provider mutations (previously **zero** `@limiter`).
- OAuth login/callback emit a correlation-id audit entry (success + failure); admin user-create checks role existence before flush (fail-fast).
- Regression tests lock the existing invariants: guard matrix, last-admin lockout, OAuth-secret redaction, `token_version` invalidation, SAML enterprise gate.

**Settings organization**
- New admin **Auth tab** controls: an "Allow Password Login" switch + an `allowed_email_domains` list widget (add/remove chips, jsonb-additive-safe save), 4-locale i18n.

### Live close-gate (Playwright, on a rebuilt v1046 stack)
Admin saves a domain allowlist → `PUT /settings 200` (no 422) · disallowed register → **403**, allowed → **201** · GitHub provider created (secret redacted) → GitHub button on login · SSO-only flip → **200** (lockout guard) · admin break-glass login → **200** · anonymous `/login` under SSO-only hides the password form + shows the GitHub button · migration `0010` constraint live. (Screenshot attached in the session.)

### Schema / migration
- Migration `0010` extends the `chk_oauth_providers_type` constraint to `('oidc','google','microsoft','saml','github')` — includes `'saml'` for enterprise-overlay parity, idempotent up/down, head-coupled migration tests green. *(Enterprise `e002` may later need `'github'` added — flagged in the migration header; OSS side complete.)*

### Testing
- 154 cross-phase backend tests green under `pytest -n 4`; frontend typecheck + vitest + 4-locale i18n parity green.
- `ruff`, `make openapi-check`, `make sdks-check` all pass (OpenAPI snapshot + Python/TS SDKs regenerated for the new schema fields).

### Out of scope (deliberate)
SAML config in core (seam only) · API-key scoping (deferred to backlog) · multi-org / tenancy.

---
27 feature/fix/test commits + 1 SDK-regen fix. No release tag in this PR.

https://claude.ai/code/session_01N1zmPNRxcyq7UyMJnzEbwC


ishiland:
@codex review

Please do an independent review of this branch (`feat/v1046-admin-identity-hardening`) — it's a security-sensitive identity/authz milestone. Focus areas:
- **Domain allowlist enforcement** at all 6 auth paths (signup, password login, SSO callback, refresh, GitHub callback, admin-create) via the shared `is_email_allowed` — any bypass / TOCTOU / path missed?
- **`manage_settings` break-glass** consistency — server-side capability check (not spoofable), correctly absent on no-principal paths (signup / SSO-JIT-new).
- **SSO-only** (`password_login_enabled`) enforced in the handler not just the UI; **lockout guard** (settings + provider disable/delete) — race or off-by-one.
- **GitHub email resolution** — only primary AND verified accepted; migration `0010` constraint composition with enterprise `e002`.
- **Rate-limit** coverage (no-trailing-slash alias bypass) + audit correlation not leaking secrets.

ishiland:
## Codex review — addressed (commits `5836633f`, `9fa26bd7`, `716f10b4`)

All 3 findings were real and are fixed:

- **P1 — Require an allowed email before SSO JIT provisioning** (`oauth/service.py`): the new-identity domain check was gated on `if email:`, so a new SSO identity whose IdP omits the email claim skipped the allowlist and provisioned with `email=None`. Now: when `allowed_email_domains` is non-empty and the IdP emits no email, the callback raises `OAuthDomainNotAllowedError` (a new JIT identity has no principal for break-glass → unverifiable ⇒ reject). Empty allowlist unchanged. +2 tests. (`5836633f`)
- **P2 — Serialize password-disable with provider changes** (`settings/router.py`): added a transaction-scoped `pg_advisory_xact_lock` with one shared key at all three lockout-guard sites (password-disable, provider disable, provider delete), so the two mutation classes can't both pass and leave 0 providers + password login off. +3 regression tests (a true two-session race test was rejected as flaky). (`9fa26bd7`)
- **P2 — Roll back OAuth state before committing failure audit** (`oauth/router.py`): the generic-error handler now `db.rollback()`s before `audit_emit`, so a partially-flushed JIT user/refresh-token from the failed flow is discarded and only the failure-audit row persists. +1 test (monkeypatched flush-then-raise). (`716f10b4`)

Verification: 109 affected-suite tests green under `pytest -n 4`; no API-schema change (openapi/sdks unaffected). Thanks @codex.

ishiland:
### CI fix — reverted the advisory lock (P2 #2)

The full CI suite (4288 tests) surfaced what my local v1046 subset didn't: the `pg_advisory_xact_lock` from `9fa26bd7` (Codex P2 — lockout-guard concurrency) is a **global, transaction-scoped lock that serializes every settings mutation across CI's parallel xdist workers on the shared Postgres**, hanging the lockout-guard tests (200 instead of 422 + 'I/O on closed file'). It's correct in production (admin settings writes are rare) but breaks the parallel test suite, and a global lock that breaks CI is worse than the narrow race it prevents.

**Reverted it (`898048b2`).** The race Codex flagged is **recoverable**: a `manage_settings` admin always retains password-login break-glass, so the org is never truly locked out — at worst a non-admin is temporarily SSO-gated until an admin re-enables a provider. Tracking a properly-scoped serialization (row-level lock on the settings+provider rows in a consistent order) as a follow-up rather than shipping the over-coarse global lock.

Also fixed `898048b2`'s sibling: the new defensive `except Exception:` sites got `# broad:` justifications for the CODE-08 invariant (`c972b463`).

FIX-A (P1, no-email JIT bypass) and FIX-C (P2, partial-side-effects rollback) stand. Re-running CI.

ishiland:
## Codex findings — final disposition (verified against HEAD `78e29f49`)

All three findings re-verified against current branch HEAD. Summary:

| # | Sev | Finding | Status |
|---|-----|---------|--------|
| 1 | P1 | Require allowed email before SSO JIT provisioning | ✅ Fixed — `5836633f` |
| 2 | P2 | Serialize password-disable vs provider changes | ⚠️ Accepted (rationale below) |
| 3 | P2 | Roll back OAuth state before committing failure audit | ✅ Fixed — `716f10b4` |

**#1 (P1) — resolved.** `service.find_or_create_oauth_user` now rejects a brand-new identity with no `email` claim while `allowed_email_domains` is non-empty (`raise OAuthDomainNotAllowedError` *before* any User row is created). Returning users and the empty-allowlist path are unchanged; enterprise SAML JIT routes through the same gate. Covered by tests asserting no user row is created in the reject case.

**#3 (P2) — resolved.** The generic-OAuth-error branch in `oauth/router.py` now `await db.rollback()` *before* `audit_emit` + `commit`, so no partial JIT user/identity rows survive while the failure-audit row still persists. The two remaining commit-without-rollback branches (`email_not_verified`, `domain_not_allowed`) were checked: both raise *before* any `db.add/flush`, so they carry no partial state.

**#2 (P2) — accepted as a non-security foot-gun, not reintroducing a lock.** The race is real (the lockout guards read the enabled-provider / password-login state without serialization), but it **cannot produce a true lockout**: the SSO-only login gate runs *after* password verification and exempts any holder of `manage_settings`, and `validate_permission_matrix` hard-guarantees the `admin` role can never lose `manage_settings` (removal/grant both raise). So an admin can always password-break-glass in, even with `password_login_enabled=false` and zero enabled providers. The constant-key `pg_advisory_xact_lock` (9fa26bd7) stays **reverted** (898048b2) — it serialized *all* settings writes across the parallel CI workers and hung the suite. The residual is an operational foot-gun (admin must re-enable a provider/password login), not a security defect. If we later decide to close the foot-gun, the correct fix is a row-scoped `SELECT … FOR UPDATE` on the single `app_settings` key + provider rows in a consistent order — that locks only those rows, avoiding the cross-worker contention that sank the advisory lock.

CI is green on `78e29f49`.

ishiland:
@codex review

Please re-review the two commits added since your last pass — they address your P2 concurrency finding and a test-isolation flake it surfaced:

- `c355d439` — row-lock the three SSO lockout guards (`oauth_service.lock_enabled_providers`: `SELECT id WHERE enabled=true ORDER BY id FOR UPDATE`), acquired before reading `password_login_enabled`, counting "enabled others" from the locked set. Replaces the reverted constant-key advisory lock. Focus: deadlock/lock-ordering, whether all three pairwise races (password-disable, provider-disable, provider-delete) are truly serialized, and CI-hang risk.
- `01ffd64e` — purge providers up-front in `test_validate_oidc_empty` (global-provider-state isolation under `pytest -n`).

CI is green on `01ffd64e`.

ishiland:
@codex review

Good catch on the cache window — that race is real and your row-lock alone didn't close it. Fixed in `a9525ff1`:

- Added `PersistentConfig.get_uncached(db)` — reads `app_settings` directly with no cache read/write.
- Both provider lockout guards (disable + delete) now read `password_login_enabled` via `get_uncached` **after** acquiring the provider `FOR UPDATE` lock. Since the guard acquires the lock only after a concurrent password-disable commits, a READ COMMITTED read observes the committed `false` — no stale-cache repopulation window. The password-disable guard itself never reads the flag (it counts the locked provider set), so it's unaffected.
- Regression test poisons the cache with a stale value and asserts `get_uncached` returns the committed DB value.

Please confirm the cache-bypass closes the window.

ishiland:
@codex review

Fixed the login-gate finding and preempted the rest of that cache-staleness class:

- `88bae27a` — SSO-only login gate now reads `password_login_enabled` via `get_uncached`; the adjacent DOMAIN-04 login gate gets the same treatment. +test that poisons the cache and asserts a non-admin login is still rejected.
- `587429b0` — every remaining `allowed_email_domains` **enforcement** read is now cache-bypassed: refresh re-check, signup gate, OAuth JIT (returning + new user), admin-create.

Left cached deliberately: `GET /auth/config` display reads (the repopulator — harmless now that enforcement bypasses the cache), and `registration_enabled` (feature toggle, not authn/authz; mild signup lag).

Please confirm the stale-cache enforcement class is fully closed.

ishiland:
@codex review

Fixed the verified-email allowlist bypass in `425c2c2a`: the non-empty-allowlist JIT block now requires `email_verified is True` (raising `OAuthEmailUnverifiedError`) before treating an allowed-domain email as satisfying the allowlist — using the same trust signal the H-30 auto-link gate already relies on, so a trusted IdP (incl. SAML JIT) that sets `email_verified=True` is unaffected. Empty-allowlist behavior is unchanged. Test covers `email_verified: false` and an omitted claim (both rejected, no user created).

Please confirm.

ishiland:
@codex review

Addressed both findings from the last pass in `ac734f6a`:

- **P1 (returning-user allowlist):** the returning-user domain re-check now trusts the IdP email claim only when `email_verified` is true; otherwise it falls back to the stored (already-trusted) email — so an unverified `attacker@allowed.example` can't bypass a disallowed stored email. +test.
- **P2 (GitHub Enterprise):** `_resolve_github_identity` now uses the provider's configured `userinfo_url` and derives `/emails` from it; the GHE access token is never sent to api.github.com. +test asserting the configured endpoints are used.

Please confirm.

ishiland:
@codex review

Both findings from the last pass addressed:

- **P3 (`461358fd`):** `validate_allowed_email_domains` now rejects non-string entries with a ValueError (→ 422) before `normalize_domains` calls `.strip()`, so `{"allowed_email_domains":[123]}` no longer 500s. +test.
- **P2 GitHub admin UI (`97d47e19`):** the admin OAuth form now offers GitHub. Selecting it sets sensible defaults + the GitHub scope and hides the discovery-URL field (GitHub is plain OAuth2); optional authorize/token/user-API URL fields are shown only for github for GitHub Enterprise (blank ⇒ backend GitHub.com defaults). `OAuthProviderType` + 4 i18n locales updated. typecheck/eslint/i18n-parity/vitest green.

Please confirm.

ishiland:
@codex review

Both addressed in `81eefa68`:

- **GitHub endpoints on update:** `update_provider` now applies the same GitHub.com endpoint defaulting as `create_provider`, so editing a provider into github or clearing the GHE overrides (blank ⇒ defaults, per the UI hint) can never persist null endpoints. +test.
- **Break-glass UI:** SSO-only mode now shows an explicit "Admin? Sign in with a password" disclosure that reveals the password form on demand (clean SSO-only default preserved; server still enforces the gate). en/de/es/fr + test.

Please confirm.

PR #274, opened 2026-06-22
feat(import): upload progress, larger batch limit, batch quota banner

ishiland:
Three import-flow improvements surfaced while importing rasters on the demo (the screenshot showed `.tif` files failing with "Dataset quota exceeded: 12 of 10 datasets used" and no upload feedback).

## What changed

**1. Upload progress for large files (rasters)**
`fetch()` can't report upload-byte progress, so `uploadFile` now uses `XMLHttpRequest` and emits a 0–1 progress fraction; presigned multipart reports per-chunk. `BulkUploadProgress` renders a per-file bar + percent. The XHR path mirrors the existing `authenticatedRawFetch` proactive-refresh + single 401-retry so a first-after-idle upload doesn't hard-fail on a stale JWT.

**2. Batch file limit raised 10 → 25**
The 10-file cap was client-only (`react-dropzone`). Bumped to 25 and parametrized the copy via `{{max}}` so the number lives in one place. Deliberately **not** size-based: the meaningful caps (per-file 500MB via `body_limit` + the per-user dataset quota) are already server-enforced, so a client size-limiter would just duplicate them.

**3. Batch-level quota banner**
A quota (422) error is identical across every file in a batch, so the old UI repeated the same red line on each row. Now it surfaces **once** as a banner with the "X of Y datasets used" detail; each affected row just shows a short "limit reached" label.

## Notes
- The "12 of 10" itself is working-as-designed (demo `max_datasets_per_user=10` DB override). This PR only improves how it's communicated — it does not change quota enforcement.
- i18n keys added to all four locales (en/de/fr/es); parity + source-key gates pass.

## Verification
- `npm run typecheck` — clean
- `npm run test:i18n` + `check:i18n:changed` — pass
- `eslint` on touched files — clean
- `vitest run src/components/import src/api` — 138 passed (incl. new uploadChunks progress test + updated use-ingest test)

@codex review

https://claude.ai/code/session_01N1zmPNRxcyq7UyMJnzEbwC

### DIFF
--- a/backend/app/processing/ingest/router.py
+++ b/backend/app/processing/ingest/router.py
@@ -84,7 +84,7 @@ from app.core.persistent_config import (
     UPLOAD_MAX_SIZE_MB,
     get_allowed_extensions_list,
 )
-from app.modules.quota.service import check_upload_quota
+from app.modules.quota.service import check_upload_quota, get_user_quota_usage
 from app.processing.raster.validation import validate_sources
 from app.platform.storage import get_storage
 from app.standards.ogc.errors import ERROR_RESPONSES_WRITE
@@ -136,6 +136,14 @@ async def get_upload_config(
     """Return upload configuration including presigned upload availability."""
     max_size_mb = await UPLOAD_MAX_SIZE_MB.get(db)
     allowed_exts = await UPLOAD_ALLOWED_EXTENSIONS.get(db)
+
+    # Advisory remaining-quota hint so the client can cap a batch at what the
+    # user can actually create. None when no count cap is set (unlimited).
+    usage = await get_user_quota_usage(db, user.id)
+    remaining = (
+        max(0, usage.count_cap - usage.dataset_count) if usage.count_cap > 0 else None
+    )
+
     return UploadConfigResponse(
         presigned_uploads=settings.storage_provider == "s3",
         presigned_threshold_bytes=settings.presigned_multipart_threshold_mb
@@ -143,6 +151,7 @@ async def get_upload_config(
         * 1024,
         max_file_size_bytes=max_size_mb * 1024 * 1024,
         allowed_extensions=allowed_exts,
+        remaining_dataset_quota=remaining,
     )
 
 
--- a/backend/app/processing/ingest/schemas.py
+++ b/backend/app/processing/ingest/schemas.py
@@ -436,6 +436,14 @@ class UploadConfigResponse(BaseModel):
     allowed_extensions: str = Field(
         description="Comma-separated list of allowed file extensions."
     )
+    remaining_dataset_quota: int | None = Field(
+        default=None,
+        description=(
+            "Datasets the caller may still create before hitting the per-user "
+            "count cap, or null when no count cap is configured (unlimited). "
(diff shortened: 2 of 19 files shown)


ishiland:
Addressed the P2 (batch size vs. enforced quota) in d54425e8.

Investigation: dataset rows are created by the **async worker** (`tasks_raster.py:109`, etc.), not at commit — so per-file upload checks all pass for a fresh user and the batch overshoots the cap regardless of the client number. True atomic count enforcement would need worker-side locking; that's a pre-existing architectural gap, deferred.

Took your first option — align the client cap to remaining quota:
- `GET /ingest/upload/config` now returns `remaining_dataset_quota` (`count_cap - dataset_count`, or `null` when no cap is set).
- `FileDropzone` clamps `maxFiles` to `min(25, remaining)`, floored at 1 so an at-cap user still gets the explicit server quota error + banner (`react-dropzone` treats `maxFiles: 0` as unlimited).

Regenerated `openapi.json` + SDKs; added config-endpoint quota tests. On the demo (cap 10) a fresh user now sees a 10-file cap; OSS default (no cap) keeps 25.

@codex review

ishiland:
Addressed the config-load-window P2 in 01b513f4.

`FileDropzone` is now disabled while `useUploadConfig` is pending, so an unresolved config (`uploadConfig === undefined → null`) is no longer treated as "unlimited" — it's "unknown". Once the query settles, success carries the real `remaining_dataset_quota`; an error degrades to permissive, consistent with how `allowedExtensions`/`maxSizeMb` already behave. `staleTime` is 5m, so this only gates the genuine first fetch.

@codex review

ishiland:
Addressed both findings in 5c9d4f2e.

- **P1 (comment ref):** anchored the source comment to `PR #274` per the AGENTS.md inline review-comment convention.
- **P2 (stale quota):** scoped the `upload-config` query key by user id (logout only clears `auth.me`, so the per-user `remaining_dataset_quota` could otherwise leak across an account switch), and `invalidate` it on `reset()` so "Upload More" after an import refetches the current dataset count instead of the cached pre-import value.

Typecheck, eslint, and the import/api suite (138) pass.

@codex review

ishiland:
Addressed in 515aa4b2 with a structural fix rather than more per-call-site invalidation.

`useUploadConfig` now uses `staleTime: 0` + `refetchOnMount: 'always'`, so the quota-bearing response is re-read on every Import mount — covering a delete elsewhere or an import-then-navigate-away-and-back, without needing to hook each mutation (`useDeleteDataset` etc.). The same-mount "Upload More" path stays covered by `invalidate` on `reset()`, and the key remains user-scoped for account switches.

Typecheck/eslint/138 tests pass.

@codex review

ishiland:
Addressed in e958e4f8 — the dropzone now gates on `isFetching` instead of `isPending`, so it's disabled during the `refetchOnMount` background refresh too (not just the cold load). The stale cached cap is no longer usable in the window before the live quota GET lands; once any fetch settles, the live `remaining_dataset_quota` applies.

Typecheck/eslint/138 tests pass.

@codex review

ishiland:
@codex review

PR #314, opened 2026-06-24
fix: demo bug-bash — dataset detail, builder, import & jobs hardening

ishiland:
## Summary

A batch of fixes from a live demo bug-bash, in 5 grouped commits — dataset-detail rendering/UX, the map builder, the import flow (URL + STAC), and the jobs API + console hygiene.

### UI / dataset detail
- **Dataset detail loaded scrolled to the bottom** — `AppLayout` resets window scroll on pathname change; a list→detail SPA click no longer carries the list's scroll offset onto the shorter detail page.
- **Overview tab disorganized** — *Similar Datasets* / *Used in Maps* moved up the sidebar, their headers normalized to the compact eyebrow style, related cards switched to a vertical stack (the horizontal scroll row didn't fit the 300px rail).
- **Builder map controls overlapped the coordinate HUD** — the coordinate pill now drops below the centered `MapToolbar` in the builder (scoped via `data-builder-canvas`; viewer untouched).

### Map builder
- **Redundant "Basemap" tab in the Add-Data dialog** — it duplicated the left-pane basemap controls (same `swapBasemapPreset` path). Removed the tab, routed the left-pane "Swap basemap" kebab to the existing `BasemapGroupEditorScene` flyout, and deleted the dead `BasemapPicker` component + dangling test mocks.

### Import
- **Large ArcGIS FeatureServer preview hung ~2 min then showed no attributes** — preview drove GDAL's `ESRIJSON` driver against a `/query` URL; that driver ignores `resultRecordCount` and paginates the **entire** layer (3.75M rows on the repro) until the subprocess timed out and silently returned empty columns. Preview now derives fields/CRS/geometry from the FeatureServer `?f=json` metadata + a bounded sample query — instant regardless of layer size. Preview timeout lowered to 30s and now raises instead of returning a fake-empty preview.
- **No import progress indicator** — `JobProgress` renders a determinate bar + stage label from the `progress`/`current_step`/`rows_processed` the worker already emits; added a `beforeunload` guard warning before abandoning an in-progress import.
- **STAC bulk import showed only "N failed"** — now surfaces the real per-item error reasons (grouped) instead of a bare count; SSRF rejections are logged. (The common cause is catalogs exposing `s3://` asset hrefs, which the SSRF guard blocks — intentionally **not** bypassed here.)

### Jobs API / console hygiene
- **`/jobs/by-dataset/{id}` 404 polluted the dataset-page console** — now returns `200 + null` for the "dataset visible but no ingest job" case (remote/STAC/registered datasets); the not-visible `404` is preserved so job existence still isn't leaked.
- **Public dataset page called admin-only `/settings/all/` → 403** — switched to the public feature-flags endpoint.
- **maplibre "Expected value to be of type number, but found null"** — null-safe `fill-extrusion-height` expression (`has`-guarded).

### Dataset preview hardening (conditional freeze)
- **Dataset detail page could freeze the browser tab** (server stays healthy). Hardening: the raster preview fires hero `loaded`/`error` transitions at most once per mount, removes its map event listeners on teardown, and treats HTTP `204` tiles as present (not errors) so a sparse remote COG resolves deterministically; `useUnsavedGuard` now blocks only real pathname changes, not the hash-based tab nav that was tripping the react-router blocker ("blocker on POP navigation" warning). **Note:** the freeze is conditional (remote-tile/titiler health) and could not be reproduced on demand — these are the evidence-backed, low-regression hardening changes, not a confirmed single fix.

## Also investigated — no code change
- **Demo SSO "broken"** — not a bug. 3 OAuth buttons render on `/login`; `auth_methods:[]` is hardcoded empty in OSS (the enterprise SAML slot) and never drove the OAuth buttons (those come from `/auth/oauth/providers`).
- **Sentinel non-percentile stretch "errors"** — could not reproduce; Min/Max + Std-Dev both return 200 in the live builder with 0 console errors. Likely a stale frontend bundle.
- **Editor authz asymmetry** (any editor can edit any dataset, but only their own maps) — intentional (shared org catalog vs owned artifacts). Flagged: dataset **reupload** lacks the owner/admin gate that **delete** has — left as-is pending a product decision.
- **Map thumbnails not generated on seed (THUMB-01)** — deferred; no server-side thumbnailer exists.

## Tests
- **frontend:** `tsc -b` clean; full `vitest` **2996 passed**.
- **backend:** `ruff` clean; affected suites (services/preview/jobs/stac/arcgis/ingest) **107 passed** (new tests added for the ArcGIS metadata path, the 30s-timeout-raises behavior, and the jobs 200+null contract).


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #315, opened 2026-06-24
fix(api): smoke-check backlog — OGC raster hardening, search pagination, column stats, param validation

ishiland:
## Smoke-check backlog — backend hardening

Fixes the backend defects surfaced by the post-#314 demo smoke-check. Every bug below was **reproduced live** against the local stack before fixing, and each fix ships with regression tests. Work was done in two passes: implement → adversarial multi-lens review (which caught one regression I introduced + three incomplete fixes) → round-2 fixes.

### Fixed

| # | Sev | Bug | Fix |
|---|-----|-----|-----|
| B1 | **CRITICAL** | Raster/DEM `GET /collections/{id}/items` returned **500** (`UndefinedTableError`) and held a DB connection ~20s on an unauth endpoint — a DoS-amplification vector. | `record_type` guard → fast **404** before any query, on both OGC handlers **and** the native `/datasets/{id}/features*` router (the same root cause lived on a second router). Missing-table *vector* datasets (cold-evicted/partial ingest) now return **503** instead of 500. |
| B2 | HIGH | Default search pagination **duplicated/dropped** records (unstable `ORDER BY`, no unique tiebreaker). | Append `Record.id` tiebreaker to every dataset search sort branch. |
| B3 | HIGH | `GET /datasets/{id}/columns/{col}/stats/` **500** on text columns (`::numeric` cast). | Detect column type; non-numeric columns return a categorical summary (`count` + `distinct_count` + `data_type`). New optional `ColumnStatsResponse` fields. |
| B4 | HIGH | On feature collections, CQL2 `filter=` was **silently ignored**; a malformed filter returned 200. | Reject `filter=` on feature collections with **400** (CQL2 is supported on the `datasets`/records collection, which has queryables); reserve `filter`/`filter-lang`. |
| B5b | LOW | 429 responses lacked **Retry-After**. | Emit `Retry-After` from the limiter window. |
| B5c | MED | `numberReturned > numberMatched` (appended search collections weren't counted). | `numberMatched` is now stable across pages and includes the full matching-collection count; next-link driven by the dataset total only (no phantom next-link to an empty page). |
| B5d | LOW | `/datasets/{id}/relationships/` returned `record_id`s that **404** on dereference. | Return dereferenceable `Dataset.id` values (list **and** create responses). |
| B5e | LOW | Malformed `datetime=` → **500**; `filter-lang` ignored on `/search/datasets`. | 400 on malformed datetime (records/search/facets/STAC) and on bad `filter-lang`. |

[...]

PR #317, opened 2026-06-24
fix(api): raster tile fmt sanitize (stretch/colormap 422) + search param validation + facet/header/login hygiene

ishiland:
Follow-up to #315/#316 — closes the residual smoke-check hygiene items plus a real, live-reproduced raster rendering bug.

## Raster stretch/colormap 422 (the big one)
On the live demo, `?stretch=minmax|percentile|stddev` and `?colormap_name=` on raster tiles returned **422** (base tiles were fine). Root-caused by isolating the layers: hitting the api **directly** → 200, but through the proxy chain → 422. A reverse-proxy rewrite (frontend nginx `/raster-tiles`: `rewrite ... $is_args$args` + a *variable* `proxy_pass`) URL-encodes the request query into the **path**, so the `{fmt}` path param arrives as `png?stretch=...` and the built Titiler URL becomes `.../771.png?stretch=...?url=...` (double `?`) → Titiler 422.

**Fix:** sanitize + allowlist `{fmt}` at the top of `raster_tile_proxy`. The real render params still arrive as proper query params, so stripping the path pollution lets the tile render **with** the stretch applied. Also a hardening win — `{fmt}` feeds an upstream URL. (Left the nginx rewrite alone: it works for base tiles and variable-`proxy_pass` query handling is fragile to change blind; the backend sanitize covers all deployments.)

## Hygiene (post-#315 assessment residuals)
- **A2** — `record_type` / `sort_by` were free-form: a bogus value silently returned 200 (empty result / silent `created_at` fallback). Now `400` via `SearchQueryParams.to_filters()` (the single chokepoint for both native + OGC search).
- **A4** — the `record_type` facet advertised a `collection` count, but `?record_type=collection` filters `Record.record_type` (no such rows) and always returned 0 — a dead facet value. Dropped it; collections are surfaced via the dedicated `collections` facet.
- **A3** — `/api` responses carried **two** conflicting `X-Frame-Options` (backend `DENY` + nginx-inherited `SAMEORIGIN`). nginx `proxy_hide_header X-Frame-Options` in the `/api/` block → exactly one survives. Backend still emits `DENY` for direct (non-proxied) access.
- **Login** — removed the redundant "Single sign-on is available when your organization has configured it" caption that only ever rendered *beneath already-configured SSO buttons* (the component returns null with zero providers, so it never showed in the no-provider case it described). Dropped the unused i18n key (en/de/fr/es).

## Tests
- `test_proxy_polluted_fmt_query_in_path_is_sanitized`, `test_proxy_rejects_unsupported_fmt`
- `test_search_rejects_bogus_enum_params[record_type|sort_by]`, `test_search_accepts_valid_enum_params`
- `test_facets_excludes_dead_collection_record_type`
- 84 backend (touched files) + 31 frontend auth tests green; frontend typecheck clean.

## Live-verified on the demo (`demo-2026-06-24d`)
- stretch minmax/percentile/stddev + custom pmin/pmax → **200** (differing PNG sizes confirm stretch is applied)
- bogus `record_type`/`sort_by` → **400**; facet no longer lists `collection`
- `/api` → single `X-Frame-Options: SAMEORIGIN`; login caption gone from the served bundle

### DIFF
--- a/backend/app/modules/catalog/search/router.py
+++ b/backend/app/modules/catalog/search/router.py
@@ -212,6 +212,21 @@ async def _build_raster_assets(
 # ---------------------------------------------------------------------------
 
 
+# Allowed values for record_type / sort_by query params (A2 validation). Kept in
+# sync with the Record.record_type CHECK constraint (models.py) and the sort
+# branches in service_datasets._resolve_sort_order.
+_ALLOWED_RECORD_TYPES = {
+    "vector_dataset",
+    "raster_dataset",
+    "vrt_dataset",
+    "map",
+    "service",
+    "collection",
+    "table",
+}
+_ALLOWED_SORT_BY = {"relevance", "date_added", "name", "title", "last_updated"}
+
+
 class SearchQueryParams(_BaseModel):
     """Query parameters for the search endpoints, injectable via FastAPI Depends().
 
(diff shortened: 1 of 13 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #320, opened 2026-06-25
fix(authz): owner-or-admin consistency across dataset & collection mutations (backend + UI)

ishiland:
## Problem

A user could unpublish (and edit, re-upload, delete, etc.) datasets they did not create. A deep authorization audit confirmed this was systemic: authz is hand-rolled per endpoint, and `check_dataset_access` — a **visibility** check that returns true for any authenticated user on a public+published dataset — was used as a **write** gate. So any `editor` could mutate other users' **public** datasets and mutate/delete **any** collection (collections had no ownership check at all). Maps were already correct via `check_map_ownership`.

The rule: **if a user created a resource, only they — or a global admin — may change it.**

## Backend (commit 1)

Shared owner-or-admin guards mirroring the map gold standard, applied to every mutating endpoint:

- **`check_dataset_write_access`** (`catalog/authorization.py`) — visibility check (404) → ownership (403). NULL `record.created_by` (seeded/import data, owner-deleted rows) is **admin-only**.
- **`check_collection_ownership`** (`catalog/collections/service.py`) — creator or admin; NULL owner admin-only.

Applied to: dataset metadata PATCH, `/status` + `/target-status` (publish/unpublish), single + bulk delete, reupload (all 6 handlers), attribute edit/reset, relationship create/delete (source-owner), VRT regenerate, `validate?refresh=true`; and collection update/delete/add/remove. Read endpoints keep `check_dataset_access`. Fixed the false "Admin only" docstring on collection delete; the `visibility-filter-coverage` pre-commit hook now recognizes the write guard.

## Frontend (commit 2)

The UI gated edit/publish/delete/membership controls on role alone, so after the backend tightened, a non-owner editor would see buttons that 403 on click. Added the canonical client helper the audit found missing — `canMutateResource` / `useCanMutate` (creator-or-admin; NULL owner admin-only) — and gated the dataset and collection mutation controls on it.

> Note: the maps list-card delete stays role-gated (cosmetic — backend already 403s; `MapSummaryResponse` carries no owner id). Tracked as a small follow-up.

## Behavior changes
- Non-owner editors can no longer mutate peers' public datasets/collections (this was the bug).
- Resources with no recorded owner are now admin-only to mutate (safe default).

## Tests
Regressions for the reported "unpublish a dataset you didn't publish" bug, plus public-dataset edit and collection delete/update — non-owner `403`, owner and admin allowed. Frontend unit test for `canMutateResource`. Updated bypass mocks in `test_workflow_extension` and `test_defer_orphan_guard` (the latter per Codex review) to patch the new write gate. Regenerated `openapi.json` + SDKs for the collection docstring changes.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -52,19 +52,17 @@ repos:
         name: "Visibility Rule 1 — route files calling get_dataset must reference check_dataset_access or apply_visibility_filter"
         description: >
           Any route handler that fetches a Record/Dataset/Map by ID must call
-          check_dataset_access_or_anonymous (read) or check_dataset_access (write)
-          or apply_visibility_filter. See AGENTS.md Security pre-commit checklist Rule 1.
-          One file (router_reupload.py) is excluded: it has a tracked,
-          separately-managed access-control gap and stays excluded until that
-          fix ships.
+          check_dataset_access_or_anonymous (read), check_dataset_access (read),
+          check_dataset_write_access (owner-or-admin mutation), or
+          apply_visibility_filter. See AGENTS.md Security pre-commit checklist Rule 1.
         language: system
         entry: >-
           bash -c '
           for f in "$@"; do
             if grep -qE "^@(stac_|ogc_|layers_|router)\.(get|post|patch|delete|put)" "$f" \
               && grep -qE "get_dataset\(" "$f" \
-              && ! grep -qE "check_dataset_access|apply_visibility_filter" "$f"; then
-              echo "FAIL: $f has route handlers calling get_dataset without check_dataset_access or apply_visibility_filter"
+              && ! grep -qE "check_dataset_access|check_dataset_write_access|apply_visibility_filter" "$f"; then
+              echo "FAIL: $f has route handlers calling get_dataset without an access/visibility check"
               exit 1
             fi
           done
--- a/backend/app/modules/catalog/authorization.py
+++ b/backend/app/modules/catalog/authorization.py
@@ -133,3 +133,40 @@ async def check_dataset_access(
         )
 
     return user_roles
+
+
+async def check_dataset_write_access(
+    db: AsyncSession,
+    dataset: Any,
+    dataset_id: uuid.UUID,
+    user: Identity,
+    *,
+    user_roles: set[str] | None = None,
+) -> set[str]:
+    """Enforce owner-or-admin for dataset MUTATIONS. Raises 404/403.
+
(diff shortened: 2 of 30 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

(Note: the reupload patch P2 was already addressed in 46e338c2 — `test_defer_orphan_guard.py` now patches `router_reupload.check_dataset_write_access`; the prior flag was on the pre-fix snapshot.)

ishiland:
@codex review

ishiland:
All three P2s from the prior review are addressed in the latest commits (the inline threads are carried over from earlier snapshots):

- `router_reupload.py` stale patch targets → fixed in `46e338c2` — `test_defer_orphan_guard.py` now patches `router_reupload.check_dataset_write_access` (3 sites). Verified at HEAD.
- `DatasetPage.tsx:420` keep delete permission gate → fixed in `8a074735` — delete now uses `canEdit` (`isEditor && canMutate` = edit_metadata capability **and** ownership).
- `CollectionDetailPage.tsx:108` keep manage permission gate → fixed in `8a074735` — actions now use `canManage` (`isEditor && canMutate`).

CI is green and the branch is mergeable.

ishiland:
@codex review

ishiland:
Addressed the P1 (`e292f122`) — and it pointed at a real gap the original audit missed:

**Code (the actual fix):**
- `features/router.py` — create/replace/patch/delete feature-row endpoints now use `check_dataset_write_access` (a non-owner editor could previously edit rows in a peer's public dataset).
- `ingest/router.py` — `add_vrt_source` now owner-gates the VRT being mutated, and `remove_vrt_source` (which had **no** ownership check) now does too. The source being *linked* still needs only read access.
- Two follow-up sweeps confirm these were the last gaps: every mutating handler that loads a dataset/map/collection now reaches a write guard, except read/create/suggestion paths (ai metadata suggestions, vrt-create) which correctly require read+capability, not source ownership.
- Regression test: non-owner-editor 403 on feature insert.

**On the hook itself:** I kept it a coarse file-level heuristic (it catches "a mutating route file with no access check at all"). A precise per-handler write-vs-read classification can't be done reliably in a grep — it would false-positive on legitimately read-gated mutations (vrt-create, collection link-targets, ai metadata suggestions). Per-handler correctness is enforced by review + the owner-or-admin regression tests instead.

@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #321, opened 2026-06-25
feat(dataset): Dataset Details Overview whitespace + role/edit affordances + label-tier cleanup

ishiland:
## Summary

UI/UX pass on the **Dataset Details page**, driven by a live, multi-lens review (whitespace, hierarchy, IA, role-UX, a11y). Functionally the page was already sound (every tab verified live across vector/raster/VRT, 0 console errors); this PR addresses the **quality** findings — chiefly the Overview tab's poor use of white space.

All changes are frontend-only and were verified live in the browser at 1440px and 852px, plus the role/permission states.

## What changed

**Whitespace / layout (the core complaint)**
- New vector/table **"Fields" card** so the Overview left column carries content (column name+type + "View all data →") instead of sitting ~500px empty while the rail ran much longer. Mirrors the existing Raster Properties card; the two branches are mutually exclusive by record type.
- Every sidebar card compacted (`gap-2 py-4`) to neutralize shadcn `Card`'s inherited `gap-6 py-6`; dashed dividers → solid.
- **Similar datasets** now flow 2-up below `lg` and 1-up in the rail, with badge + count + match% on one line — roughly halves the section height at tablet width.

**IA / clarity**
- Overview "Metadata" sidebar card renamed to **"Details"** with an "Edit in Metadata →" jump, ending the two-"Metadata" collision.
- Dropped the duplicated **BBox** row from the Overview rail (it lives in the stats strip + Metadata tab; this also removed the only bbox precision mismatch — 2dp rail vs 4dp card).

**Role / edit affordances**
- **Structure tab** now explains *"Attribute editing is disabled for this deployment"* to owners/admins when `enable_dataset_editing` is off — previously it showed nothing and no reason, while the Metadata tab dinged them on an Attribute-Completeness score they couldn't act on.
- **Editor-non-owners** get an ownership-specific denial reason on editable fields instead of the generic "Editors can make changes" viewer copy.
- Dropped the redundant third-person "Editors can update fields…" banner for editors (the pencil already signals it; also reclaims vertical space).
- Editable field shell reveals its tint on **hover/focus** instead of at rest, so the description box no longer reads as "active/error".

**Typography / a11y**
- Centralized the repeated mono-caps eyebrow class into a single `SECTION_EYEBROW` source of truth (7 call sites), and split the label system into two explicit tiers: mono-caps **section markers** vs quiet **sans field labels** in the detail lists — fixing the "flat hierarchy" finding at the source.
- Bumped sub-12px stat/label text to the 11px legibility floor; default the SideKV value tooltip so truncated `bbox`/`table_name` keep their full value.

## Verification
- `tsc -b` clean · `eslint` clean · **full frontend suite: 3004 tests pass** (updated one EditableFieldShell test that asserted the old always-on tint).
- Live-checked after a frontend restart: vector Overview @1440 + @852, Structure deployment-gate, raster Overview unaffected (no Fields card; Raster Properties + Band Details intact), anonymous read-only state. 0 console errors throughout.

## Notes on the two "held" P1 sub-items (done the long-term-sustainable way)
- **Typography:** rather than a one-off restyle, the eyebrow treatment is now a single reusable constant and the tiers are explicit, so the look can evolve (or gain heading semantics) in one place.
- **Spatial dedup:** the one genuinely redundant + precision-drifting display (the Overview-rail bbox) is removed. The remaining CRS appearances live on **different tabs** in contextually-correct roles — the always-visible stats strip (glance), the Raster Properties technical spec block, and the Metadata tab's Spatial Extent card (formal record). Gutting the Spatial Extent card's CRS row would leave a card titled "Spatial Extent" showing only a bbox, and a `formatEpsg()` util for a trivial `EPSG:${n}` string would be over-engineering — so this is intentionally left as-is.

## Still open (deliberately out of scope)
- Full page **heading-outline** a11y fix (CardTitle renders a `div`) — needs a whole-page heading-level audit; `SECTION_EYEBROW` is the natural home for it as a follow-up.
- P2 polish items from the review (stats-strip sparse-cell hugging, anon "Add to Map" CTA, license/governance grouping).

https://claude.ai/code/session_017QVGwV2NqDZEnh4zzA418o

### DIFF
--- a/frontend/src/components/dataset/AttributeTable.tsx
+++ b/frontend/src/components/dataset/AttributeTable.tsx
@@ -304,11 +304,11 @@ export function AttributeTable({ datasetId, canEdit = false, compact = false }: 
       {/*
         PERF-07: scroll container is `parentRef`. Virtualization is applied to
         the <tbody> only — the native <table>/<thead>/<tr>/<td> structure is
-        preserved so column auto-sizing, sticky headers, and the existing
-        `w-max min-w-full` layout continue to work. Vertical scroll is captured
-        here (max-h + overflow-auto); the inner shadcn Table container's
-        `overflow-x-auto` continues to handle wide-column horizontal overflow
-        within this scroll region.
+        preserved so column auto-sizing and sticky headers keep working. The
+        table is `w-max` (content-sized): wide tables overflow and scroll via
+        the inner shadcn Table container's `overflow-x-auto`; few-column tables
+        size to their content instead of stretching to fill (no empty columns).
+        Vertical scroll is captured here (max-h + overflow-auto).
 
         The recommended TanStack pattern for native-table virtualization
         (https://tanstack.com/virtual/latest/docs/framework/react/examples/table)
@@ -328,7 +328,7 @@ export function AttributeTable({ datasetId, canEdit = false, compact = false }: 
               : { height: `${virtualizer.getTotalSize()}px`, width: '100%' }
           }
         >
-          <Table className="w-max min-w-full">
+          <Table className="w-max">
             <TableHeader className="sticky top-0 z-10 bg-muted/80 backdrop-blur-sm">
               {table.getHeaderGroups().map((headerGroup) => (
                 <TableRow key={headerGroup.id}>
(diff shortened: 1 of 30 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #326, opened 2026-06-25
fix(auth): make seed_initial_admin concurrency-safe (+ CI :latest gate)

ishiland:
## Why

The `v1.4.0-rc.1` prod-smoke gate correctly withheld the GitHub Release — it caught a **real backend bug**: `seed_initial_admin` is not concurrency-safe. Prod runs `uvicorn --workers N`, so on a **fresh DB** every worker runs the lifespan and races the count-check + INSERT. Two workers both see `count==0` and both INSERT → one hits `UniqueViolationError` on `uq_users_username_global` → the admin row never commits → `admin/admin` login returns 401. This breaks **every fresh self-hosted prod install**. (Existing installs are unaffected — the admin was seeded once, long ago.)

## What

**`fix(auth)` — `backend/app/api/main.py`**
- Take a `pg_advisory_xact_lock` before the count-check + INSERT in `seed_initial_admin`, so exactly one worker seeds and the rest see `count>0` and no-op. The lock is xact-scoped (released on commit/rollback).

**CI — `.github/workflows/publish.yml`**
- Gate the GHCR `:latest` tag on a **GA (non-prerelease)** tag (`!contains(ref_name, '-')`). An `-rc`/`-beta` push no longer clobbers `:latest` with an unreleased candidate (same spirit as #325 gating `release.yml`'s `isPrerelease`).

**Test — `backend/tests/test_seed_admin_concurrency.py`**
- Deterministic regression test: holds the advisory key on a separate connection and asserts `seed_initial_admin` **blocks** until released, then completes with exactly one admin. Verified it fails without the lock and passes with it.

## Verification
- New test passes; confirmed it fails when the lock is removed (has teeth). `ruff` clean; pre-commit hooks pass.
- End-to-end proof is the re-cut **`v1.4.0-rc.2`** pipeline going green (the prod-smoke gate that caught this).

## Notes / not in this PR
- **Smoke-after-publish ordering**: already handled — `prod-smoke.yml` has a 40×30s pull-retry loop whose comment documents choosing that over a `workflow_run` dependency. No change.
- **GHCR `:latest` is currently broken** (points at rc.1) — restoring `:latest`→`1.3.0` is a registry op, done separately from this merge.

Also folds in a small **login spacing** tweak (`style(login)`) — tightens the sign-in panel's vertical rhythm (right panel only; brand panel untouched).

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -81,7 +81,11 @@ jobs:
             type=semver,pattern={{version}}
             type=semver,pattern={{major}}.{{minor}}
             type=semver,pattern={{major}}
-            type=raw,value=latest,enable=${{ github.event_name == 'push' && github.ref_type == 'tag' }}
+            # `latest` only on a real GA tag — exclude prereleases (a `-rc`/`-beta`
+            # suffix). Otherwise a `v1.4.0-rc.1` push clobbers `:latest` with an
+            # unreleased candidate (same spirit as #325 gating release.yml's
+            # isPrerelease). Fresh `:latest` pulls must track the last GA.
+            type=raw,value=latest,enable=${{ github.event_name == 'push' && github.ref_type == 'tag' && !contains(github.ref_name, '-') }}
           labels: |
             org.opencontainers.image.description=${{ matrix.description }}
             org.opencontainers.image.licenses=Apache-2.0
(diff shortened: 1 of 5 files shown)


ishiland:
@codex review

ishiland:
@codex review

PR #327, opened 2026-06-25
feat(admin): surface per-user storage & dataset quotas in Storage settings

ishiland:
## What

`max_storage_bytes_per_user` and `max_datasets_per_user` are registered, enforced persistent-config settings, but the **Storage** settings tab hardcoded only upload size / extensions / tile TTL — so these quotas could only be set via the settings API or a direct DB override, with **no admin UI control**. This adds them as fields, grouped under a **"Per-user limits"** subsection (0 = unlimited).

| | |
|---|---|
| `SettingsStorageTab.tsx` | Two new fields (`max_storage_bytes_per_user`, `max_datasets_per_user`) in a bordered "Per-user limits" subsection with source/reset badges |
| `en/de/fr/es admin.json` | Field + subsection i18n in all four locales (locale-parity gate) |

**No backend change** — the keys were already registered (`tab="storage"`) and the settings API already accepts them.

## Placement rationale

These are **global instance caps applied to every user** (not per-user overrides; enforced per-user at ingest). A 4-lens design panel (taxonomy, discoverability, YAGNI, industry convention) unanimously landed on: **keep them on Storage, grouped under a "Per-user limits" heading** — *not* in the per-user management list (a category error for global policy) and *not* a dedicated Quotas tab (over-engineering for two fields; revisit at ~4+ limit settings). This mirrors GitLab's "Account and limit" grouping, and keeps the resource-limit family (upload size, extensions) together.

> Note: these remain **global** caps — there is no individual per-user override (the backend reads one `MAX_STORAGE_BYTES_PER_USER` value). True per-user quotas would need a backend schema change (separate work).

## Verification
- Live, as admin: renders, **save + reload persists** server-side (proves the settings API round-trip), reset-to-0 works.
- `typecheck` ✓, `eslint` ✓, `check:i18n:changed` ✓, JSON valid + locale parity ✓.


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #328, opened 2026-06-26
feat(settings): show per-user storage & dataset usage on the Settings page

ishiland:
## What

Adds a **Storage Usage** card to the end-user Settings page (Preferences tab) so any signed-in user can see their own consumption against the configured per-user caps. Pairs with the admin per-user quota settings (#327).

- **File storage** — bytes used vs cap (0 = unlimited)
- **Datasets** — count vs cap (0 = unlimited)

Both rows show a native `<progress>` bar only when a cap is set (avoids `max=0`); otherwise an "Unlimited" label.

## How

- **Backend** — `GET /auth/me/usage/` (+ no-slash shadow, dual-decorator ROUTE-01), strictly self-scoped via `get_current_active_user`, reusing the existing `get_user_quota_usage(db, current_user.id)`. No new query, no new schema (`UserQuotaUsage` already exists).
- **Frontend** — `useMyUsage` hook (lazy, only on Settings; gated on auth; 60s staleTime) + `getMyUsage` API fn + a card on `SettingsPage`. Native `<progress>` styled for WebKit + Firefox. Counts use locale-aware `formatNumber`, bytes use `formatBytes`.
- **i18n** — `settings.storage.*` added to all 4 locales (de/en/es/fr).
- **Honest copy** — `bytes_used` tracks uploaded **raster file** bytes only (vector/table/VRT live in PostGIS and carry no `data` asset), so the row is labeled "File storage" rather than implying total storage; the **dataset count** is the accurate cross-type figure.

## Tests / gates

- New `backend/tests/test_me_usage.py`: 401-unauth, response shape, fresh-user-zero (COALESCE path), default caps = 0/unlimited.
- OpenAPI snapshot + SDKs (TS + Python) regenerated.
- Local: backend ruff + tests, frontend typecheck + eslint + i18n parity — all green.

Reviewed by an adversarial multi-lens pass before opening (0 P0/P1; one P2 copy-honesty fix + two P3 nits applied).


ishiland:
@codex review

ishiland:
@codex review

PR #329, opened 2026-06-26
chore(settings): storage-usage follow-ups (admin label + live usage refresh)

ishiland:
Follow-ups from the storage-usage feature (#328), both live-verified on localhost:8080.

## 1. Admin label consistency
The admin user-list column header was a hardcoded `"Storage Used"`. Now it's an i18n'd **"File storage"** (`users.table.storage`, all 4 locales) — matching the end-user card and honest about `bytes_used` tracking raster file bytes only. (Verified: header renders "File storage".)

## 2. Live usage refresh
`useMyUsage` now uses `staleTime: 0` + `refetchOnMount: 'always'`, so opening Settings re-reads `/auth/me/usage` and reflects an upload/delete immediately instead of serving a value stale by up to the old 60s window. This mirrors the existing per-user `remaining_dataset_quota` handling in `useUploadConfig` (`use-ingest.ts`), the established pattern for this class of per-user quota data. (Verified: a usage fetch fires on each Settings mount.)

No backend/API changes. Local: typecheck + eslint + i18n parity green.

Not addressed (by design): `bytes_used` raster-only is a documented backend limitation (true cross-type total deferred to metered/cloud quota work); the dataset **count** is the accurate cross-type fence.

### DIFF
--- a/frontend/src/components/admin/UserList.tsx
+++ b/frontend/src/components/admin/UserList.tsx
@@ -185,7 +185,7 @@ export function UserList() {
                 <TableHead>{t('users.table.email')}</TableHead>
                 <TableHead>{t('users.table.roles')}</TableHead>
                 <TableHead>{t('users.table.status')}</TableHead>
-                <TableHead>Storage Used</TableHead>
+                <TableHead>{t('users.table.storage')}</TableHead>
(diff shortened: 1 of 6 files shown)


ishiland:
@codex review

PR #333, opened 2026-06-26
fix(sdk): make the TypeScript SDK resolve under Node's native ESM loader

ishiland:
## Problem

`@geolens/sdk@1.4.0` (npm) crashes at runtime for pure-Node ESM consumers:

```
import('@geolens/sdk')
→ ERR_MODULE_NOT_FOUND: Cannot find module '.../dist/client/sdk.gen'
  imported from .../dist/client/index.js
```

The hey-api–generated client uses **extensionless** relative imports
(`export * from './sdk.gen'`). tsc under `moduleResolution: Bundler` (set in the
v1.4.0 recovery so the SDK would compile) emits those specifiers verbatim, which
Node's native ESM resolver rejects. Bundler consumers (vite/webpack) resolve them
fine, so it slipped through until `verify-published`'s clean-`node` import smoke
caught it — that workflow has been red for v1.4.0 because of this.

## Fix

A filesystem-resolved post-build step (`scripts/fix-esm-extensions.mjs`, wired as
`tsc && node scripts/fix-esm-extensions.mjs`) rewrites the **emitted dist**
imports to explicit specifiers:

- `./sdk.gen` → `./sdk.gen.js`, `../core/utils.gen` → `../core/utils.gen.js` (files)
- `./client` → `./client/index.js` (directory)

It resolves each specifier against the real files on disk (not a regex guess), so
both file and directory imports are handled, across `.js` and `.d.ts`. It does
**not** touch the generated `src`, so the SDK drift gate is unaffected.
`prepublishOnly` now also runs a Node-ESM import smoke (`verify:esm`) so a
non-resolvable build can never be published again.

## Verification

- `npm run build` rewrites specifiers in 18 files; `npm run verify:esm` → `esm import ok`; unit tests pass (3/3).
- Adversarial: `npm pack` → install the tarball into a clean dir → run the **exact** `verify-published` smoke → `createGeolensClient export ok` (exit 0).

## Versioning / release

No version bump here — `check_version_coherence.py` requires all version sites to
agree, so the fix lands at 1.4.0 and stays coherent. npm 1.4.0 is immutable, so
**actually shipping the fix needs a v1.4.1 cut** (full patch: bump all packages →
1.4.1, tag `v1.4.1`). PyPI `geolens`/`geolens-cli` 1.4.0 are already correct;
they'd republish identical content at 1.4.1.


ishiland:
@codex review

PR #334, opened 2026-06-26
ci(verify-published): stop verifying the unused private geolens-worker image

ishiland:
## Why

The re-run of `verify-published` for v1.4.0 failed its **Verify GHCR images** job on:

```
docker pull ghcr.io/geolens-io/geolens-worker:1.4.0
→ unauthorized
```

`geolens-worker` is published but **private and unused**: the shipped
`docker-compose.prod.yml` runs the worker service on the **geolens-api** image
(its baked ENTRYPOINT/CMD/HEALTHCHECK differ only), with an explicit comment that
this avoids any dependency on the geolens-worker package visibility. So no
self-hoster ever pulls `geolens-worker` — the gate was failing on an image nobody
consumes (api/frontend/backup pulled fine in the same job).

## Fix

Drop `geolens-worker` from the GHCR pull list; verify only what users actually
pull (api, frontend, backup), with a comment explaining the omission.

## Context

This was surfaced while closing out the v1.4.0 release tail. The other
`verify-published` failure (TypeScript SDK `ERR_MODULE_NOT_FOUND`) is fixed
separately in #333 and needs a v1.4.1 publish to clear. With both merged + 1.4.1
cut, verify-published goes fully green.

### DIFF
--- a/.github/workflows/verify-published.yml
+++ b/.github/workflows/verify-published.yml
@@ -148,13 +148,18 @@ jobs:
             IMAGE_TAG="${VERSION#v}"
           fi
 
+          # geolens-worker is intentionally NOT verified here: the shipped
+          # docker-compose.prod.yml runs the worker service on the geolens-api
+          # image (its baked ENTRYPOINT/CMD differs only), so no consumer pulls
+          # the geolens-worker package — it stays private and unused. Pulling it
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review

PR #335, opened 2026-06-26
test(a11y): seed a dataset for the dataset-detail accessibility check

ishiland:
## Problem

The push-only **Accessibility (axe + keyboard-nav)** CI job has **never passed** on `main` (failed on every run since it was added in v1048). The failure is a **test-fixture gap, not an axe violation**:

```
Error: Could not find a stable searchable dataset fixture   (e2e/helpers/catalog.ts:148)
  at e2e/accessibility.spec.ts:120  ("dataset detail page has no accessibility violations")
```

The job boots a fresh, **empty** stack and runs *only* `accessibility.spec.ts` + `keyboard-nav.spec.ts`. The dataset-detail test called `getSearchSeed()`, which searches the catalog for a pre-existing dataset — but nothing seeds one. (In the full E2E suite, `upload.spec.ts` creates data first; this job doesn't run it.) So `getSearchSeed` threw every time.

## Fix

- Add `seedDataset()` / `deleteDataset()` to `e2e/helpers/catalog.ts` — drives the same **upload → commit → poll** ingest sequence the import UI uses, returning `{id, title}`.
- `accessibility.spec.ts` `beforeAll` now seeds a real dataset; the dataset-detail test visits it **by id directly** (dropping the flaky search-heuristic dependency); `afterAll` deletes it (DELETE requires `confirm_title`).
- `getSearchSeed` is untouched — still used by `search.spec.ts` (which runs in the full suite where data exists).

## Verification (local, live stack)

- Full `accessibility.spec.ts` → **9/9 passed** (31.2s).
- `seedDataset` ingest completes in ~2s; the seeded title returns exactly one search result; `afterAll` leaves **0 orphan datasets**.

## Note

This job is **push-only** (gated off PRs for cost), so it does **not** run on this PR — it validates on merge to `main`. `keyboard-nav.spec.ts` has no catalog-data dependency and was unaffected.

### DIFF
--- a/e2e/accessibility.spec.ts
+++ b/e2e/accessibility.spec.ts
@@ -1,6 +1,6 @@
 import { test, expect } from '@playwright/test';
 import AxeBuilder from '@axe-core/playwright';
-import { getAuthToken, getSearchSeed } from './helpers/catalog';
+import { getAuthToken, seedDataset, deleteDataset } from './helpers/catalog';
 
 const wcagTags = ['wcag2a', 'wcag2aa', 'wcag21a', 'wcag21aa'];
 const BASE_URL = process.env.E2E_BASE_URL ?? 'http://localhost:8080';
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

PR #336, opened 2026-06-26
fix(deps): bump joserfc 1.6.4 -> 1.7.1 (CVE-2026-48990)

ishiland:
## Why

A fresh advisory **CVE-2026-48990** against `joserfc 1.6.4` began failing **both** pip-audit gates — **Security Scan** and **Python audit** — on every PR and on `main`. Because **Security Scan is a `needs` of the push-only Accessibility job**, this CVE was *also* causing that job to be **skipped**. So this one bump unblocks all three.

## Fix

`joserfc` is transitive via `authlib (>=1.7.1)`, whose constraint already permits the fixed release, so `uv lock --upgrade-package joserfc` (1.6.4 → **1.7.1**) is sufficient — no `pyproject.toml` change. Lockfile-only.

## Verification

- `pip-audit --strict` → **No known vulnerabilities found**.
- joserfc is used only in the OAuth `id_token` `iss`-validation path (`auth/oauth/router.py`, `service.py`). OAuth/SSO suite incl. the Microsoft-`iss` regression (`test_oauth_microsoft_iss.py`, `test_oauth.py`) → **44 passed** against 1.7.1.

## Context

Final blocker to a green `main` after the v1.4.0 release tail + the a11y CI gate fix (#335).

### DIFF
--- a/backend/uv.lock
+++ b/backend/uv.lock
@@ -1187,14 +1187,14 @@ wheels = [
 
 [[package]]
 name = "joserfc"
-version = "1.6.4"
+version = "1.7.1"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "cryptography" },
 ]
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review

PR #339, opened 2026-06-27
fix(ai): map viewers can use AI chat read-only; AI edits stay owner-only

ishiland:
## Summary

The map builder's **AI assistant now works for anyone who can view a map**, not just the owner. You can ask the assistant questions about a map's data on any map you can see; using the AI to **edit** a map stays limited to the owner, and AI-suggested changes still only persist when the owner saves.

## Root cause

AI builder chat (`/ai/chat/`, `/ai/chat/stream/`) was gated on map **ownership** — `_validate_chat_layers` raised `403 "You do not own this map"` for any map the caller didn't own. The SSE streaming path wraps every pre-flight `HTTPException` into a generic SSE `error` event (the body is always a 200 stream), so the frontend lost the status and rendered the misleading, **retryable** *"Something went wrong. Please try again."* — with a Retry button that could never succeed.

This surfaced on the demo: a `guest` (role `editor`, owns no maps) opening a showcase map in the builder and using "Ask AI" hit this every time.

## Why ownership was the wrong guard

AI chat is **read-only with respect to map state** — edit actions are applied client-side in the builder and only persist at **Save**, which is already owner-gated. And data access is independently bounded by `build_table_allowlist` (RBAC `apply_visibility_filter`), so a caller can only ever query datasets they can already see. The map-ownership check protected nothing the allowlist + Save guard didn't already cover.

## Change

- **`_validate_chat_layers`** now gates on **view access** (`_check_map_read_access` → 404 if not viewable, no existence oracle) and returns `can_edit` (`_can_edit_map`, owner-or-admin — the exact Save boundary) instead of raising a 403.
- **`can_edit` selects the AI tool set** via the new `select_chat_tools()`: owners get the full toolbox; viewers get **`query_data` only**, so the model literally cannot emit an edit action. `build_chat_system_prompt(can_edit=False)` adds a read-only directive so the assistant explains it can only answer questions on a map you don't own.
- **SSE error events now carry the HTTP `status`**, and the frontend (`ChatPanel.consumeStream`) surfaces a status-bearing pre-flight error as an `ApiError` — so a 403/503 classifies honestly (clear, non-retryable) instead of the generic retry message. Model-emitted mid-stream errors (tool-loop/deadline, no status) still map to the inline `StreamModelError` path.

## Verification

- **Live (local stack):** a non-owner editor on a public map gets a real streamed answer — *"…you're viewing it with read-only permissions… I'm happy to answer questions about the data…"* — with **`actions: []`** (no edits emitted). A non-viewer on a private map returns `{"message":"Map not found","status":404}`. Owner behavior unchanged.
- **Tests:** new cases for public read-only access, owner `can_edit`, the 404 view-access semantics, and the SSE-status → forbidden-banner mapping. `test_ai_chat.py` + `test_chat_streaming.py` (25 backend) and `ChatPanel.test.tsx` (41 frontend) green; `ruff`, `eslint`, and `tsc -b` clean.

## Notes / out of scope

- The builder still opens editable for non-owners with no guard (manual edits apply locally and Save 403s); this PR makes the **AI** path coherent (read-only) but does not add a builder-wide ownership gate — that's a separate UX decision.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,11 @@ and releases use semantic versioning.
   environment variables. Off-site S3 upload remains gated on
   `BACKUP_S3_ENABLED=true` and signs with AWS Signature V4 (Cloudflare R2,
   modern AWS S3, and MinIO compatible).
+- **The map builder's AI assistant is available to anyone who can view a map.**
+  Viewers — not just the map's owner — can now ask the assistant questions about
+  a map's data (counts, statistics, spatial analysis). Using the AI to *edit* a
+  map remains limited to the owner, and AI-suggested changes still only persist
+  when the owner saves the map.
 
 ### Added
 
@@ -38,6 +43,11 @@ and releases use semantic versioning.
 
 ### Fixed
 
+- **AI assistant no longer fails with a generic error on maps you don't own.**
+  Asking the builder's AI about a map you can view but not edit previously
+  surfaced "Something went wrong. Please try again." with a retry that never
+  worked; it now answers read-only questions, and genuine permission errors
+  show a clear, non-retryable message instead of a blind retry.
 - **Map Builder correctness fixes.** Bulk visibility/opacity on a multi-layer
   selection now matches single-layer behavior (companion outlines, labels,
   hypsometric relief, and clusters included); numeric-column filters now show a
--- a/backend/app/platform/extensions/protocols.py
+++ b/backend/app/platform/extensions/protocols.py
@@ -221,6 +221,11 @@ class AIProviderExtension(Protocol):
         history: list[dict] | None = None,
         port: Any,
         map_id: str | None = None,
+        # The advertised tool set for this turn. None means the provider's
+        # default chat tools; a restricted list (e.g. read-only) must be honored.
+        # Implementations with an explicit signature must accept this kwarg —
+        # stream_chat_edit always passes it.
+        tools: list[dict] | None = None,
     ) -> AsyncIterator[dict[str, object]]: ...
 
     async def structured_complete(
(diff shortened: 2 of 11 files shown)


ishiland:
@codex review

ishiland:
@codex review — addressed the P2: read-only tool set is now enforced at execution/collection (`_execute_and_yield_tools` allowed_tools guard covering native + XML-fallback paths) plus the non-streaming `chat_edit_map` tool_executor/action_collector, with a regression test.

ishiland:
@codex review — round 2 addressed in b9b7e036: (1) dropped read-only tool calls now append a refusal to `results_out` + emit `tool_result(success=False)`, keeping the OpenAI-native tool_call_id zip aligned; (2) `tools` is now declared on the `stream_chat_events` Protocol (community providers use `**kwargs`; enterprise has no custom provider impl). Regression test strengthened to assert the alignment.

ishiland:
@codex review — final commit a3cffdf9 raises the chat_service.py facade size-budget cap 400→440 (test-only; rationale in the diff) to admit the read-only access-model lines. CI green expected on re-run (the two failures were this one size-budget guard).

ishiland:
Triage of the latest Codex pass (commit `a3cffdf9`, which only changes a test-budget cap):

- **tools.py / streaming.py zip / protocols.py `tools`** — these three are re-posts of findings already fixed in `b9b7e036` (which Codex reviewed clean, 'Didn't find any major issues'). Verified intact: read-only enforcement is at execution/collection in `_execute_and_yield_tools` (allowed_tools guard), dropped calls append a refusal to `results_out` so the OpenAI-native zip stays aligned (`streaming.py:91`), and `tools` is declared on the `stream_chat_events` Protocol (`protocols.py:228`).
- **P1 — surface read-only chat in the viewer route** — valid and accepted. This PR fixes the reported bug (a non-owner **admin** in the builder, `can_edit=true`, was blocked by the old ownership check) and ships the read-only backend foundation. Surfacing a read-only 'Ask AI' on `PublicMapViewerPage` for `can_edit=false` viewers is a focused **follow-up PR** (it needs a dedicated lightweight viewer-chat component to respect the viewer/builder bundle split), tracked separately.

PR #340, opened 2026-06-27
feat(viewer): read-only "Ask AI" for public map viewers

ishiland:
## What

Follow-up to #339. That PR let the chat backend serve view-only callers a read-only toolbox (`query_data` only), but the path was **UI-unreachable**: the builder `ChatPanel` only mounts for `can_edit` users, while signed-in viewers land on `PublicMapViewerPage` with no AI affordance (the demo guest audience).

This adds a lightweight **`ViewerChatPanel`** that surfaces the read-only path.

## How

- **Self-gating** on `useAIAvailability` (token + `use_ai_chat` + AI configured) — anonymous and unpermitted viewers render nothing, no console noise.
- **Read-only by construction:** streams from `/ai/chat/stream/`; a `show_query_result` flies the map to the matched features via the shared `useEphemeralLayers` overlay and renders a compact result table. Non-query (edit) actions are ignored defensively (the backend already withholds edit tools from viewers).
- **Bundle-clean:** dedicated viewer component with a plain textarea (no builder mention/slash machinery); the builder `ChatPanel` is untouched and never pulled into the viewer chunk.
- Map instance captured via `ViewerMap`'s existing `onMapReady` to drive flyover.
- i18n keys added under `viewer.ai` across all four locales (parity gate green).

## Scope note

Read-only chat is reachable only to **signed-in viewers with `use_ai_chat`** — `/ai/chat` requires that capability. Anonymous share-link viewers see no panel.

## Tests

- New `ViewerChatPanel.test.tsx`: gating (renders nothing when unavailable), streamed answer, `show_query_result` → flyover + table, retry-able error.
- Full frontend gate green locally: `tsc -b` (0), `eslint .` (0), `vitest run` (3162 passed), i18n source-key + parity gates pass.
- No backend changes — the read-only chat path shipped in #339.

> Not live-verified end-to-end: the local dev stack's `LLM_MODEL` 404s the local key (pre-existing, unrelated), so the LLM round-trip can't run locally. Gating/streaming/flyover/error are covered by unit tests against mocked streams.

### DIFF
--- a/frontend/src/components/viewer/ViewerChatPanel.tsx
+++ b/frontend/src/components/viewer/ViewerChatPanel.tsx
@@ -1,0 +1,357 @@
+import { useCallback, useEffect, useRef, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import { AlertCircle, Loader2, RotateCcw, SendHorizontal, Sparkles, Square, X } from 'lucide-react';
+import type { Map as MaplibreMap } from 'maplibre-gl';
+import { Button } from '@/components/ui/button';
+import { ApiError } from '@/api/client';
+import { streamChatMessage } from '@/api/maps';
+import { useAIAvailability } from '@/hooks/use-ai-availability';
+import { useEphemeralLayers } from '@/components/builder/hooks/use-ephemeral-layers';
+import { cn } from '@/lib/utils';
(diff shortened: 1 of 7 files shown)


ishiland:
@codex review

PR #342, opened 2026-06-27
fix(builder): legend raster icon, copy-badge noise, demo-banner scroll

ishiland:
Three small, independent Map Builder fixes.

## 1. "Copy N of M" badge noise
The duplicate-disambiguation badge fired on shared `dataset_id` **or** display name. Two differently-named layers off one dataset (a line + its casing, a halo, any multi-style-of-one-source pattern) got tagged `Copy 1 of 2` / `Copy 2 of 2` even though their names already distinguish them — and they aren't duplicates.

Now badges only on genuine **name collisions** (the only case where rows are actually indistinguishable). `dataset_*` fields stay in the metadata, unused, for informational consumers. Live stack badge and legend badge share the same function, so they can't drift.

## 2. Raster legend "purple polygon"
On-map legend (`LegendPlugin`) rendered raster/VRT layers as a polygon swatch in the default color. Root cause: it passed the raw `layer.layer_type` (`'raster_geolens'`) to `ColorizedGeometryIcon`, whose raster/vrt branch keys off the capability **kind** (`'raster'`/`'vrt'`), so the check never matched and it fell through to `ShapeIcon`. `StackRow` was immune because it passes `caps.kind`.

Fix: pass `getLayerCapabilities(layer).kind`, aligning the call site with the layer list. Raster → grid icon, VRT → layers icon.

## 3. Map page scrolls under the demo banner
`MapBuilderPage` sized itself `h-[calc(100vh - 3.5rem)]`, subtracting only the navbar — not the demo banner that sits between the navbar and `<main>`. Result: the page was `banner-height` too tall and scrolled.

Fix: `flex-1 min-h-0` so the map fills exactly what `<main>` leaves after navbar + banner + footer. Banner- and footer-agnostic, no magic numbers. Matches `PublicMapViewerPage`, which was already immune.

## Tests
- `map-stack.test.ts` — guards name-only badging (badges same-name duplicates, does **not** badge differently-named layers off one dataset)
- `layer-icons.test.tsx` — guards the `ColorizedGeometryIcon` raster/vrt contract
- Existing builder/page suites green; typecheck + lint clean.

### DIFF
--- a/frontend/src/components/builder/__tests__/map-stack.test.ts
+++ b/frontend/src/components/builder/__tests__/map-stack.test.ts
@@ -2,6 +2,7 @@ import { describe, expect, it } from 'vitest';
 import {
   MAP_STACK_GROUP_ORDER,
   buildMapStack,
+  computeDisambiguationLabels,
   flattenMapStack,
   type MapStackGroup,
   type MapStackMapInput,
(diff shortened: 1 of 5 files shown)


ishiland:
@codex review

PR #344, opened 2026-06-27
fix(dataset): lead header with visibility, badge status only when draft

ishiland:
## Problem
The dataset header showed `Published · Private`, which reads as a contradiction. They're **two orthogonal fields** — `record_status` (lifecycle: `draft` → `published`) and `visibility` (`public`/`restricted`/`private`) — but "Published" colloquially implies "public," so the pair looks self-contradictory. The data is correct; the labeling isn't legible.

Compounding it: "Published" is the steady state and carries no information once a record is live — *of course* it's published, you're looking at it. The status badge only earns its place when the state is the **exception that needs action** (`draft`).

## Fix
- Lead with the **visibility** badge (the access question users actually have).
- Show the record-status badge **only** when `record_status !== 'published'` (i.e. draft).

A published+private dataset now reads simply **Private**; a draft shows a **Draft** badge. This mirrors the pattern `SearchResultCard` already uses (`showStatusBadge = record_status !== 'published'`), so the detail header and search cards are now consistent.

No backend, API, or i18n change — purely how the existing fields are displayed.

## Tests
`DatasetPage.hero.test.tsx` — published hides the status badge; draft shows "Draft". Existing suites green; typecheck + lint clean.

### DIFF
--- a/frontend/src/pages/DatasetPage.tsx
+++ b/frontend/src/pages/DatasetPage.tsx
@@ -72,7 +72,7 @@ import { useUnsavedGuard } from '@/hooks/use-unsaved-guard';
 import { Badge } from '@/components/ui/badge';
 import { Skeleton } from '@/components/ui/skeleton';
 import { cn } from '@/lib/utils';
-import { visibilityColors } from '@/lib/status-colors';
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

ishiland:
@codex review

PR #347, opened 2026-06-28
v1049 Polish & Trust Pass: builder, admin & UX bug-bash

ishiland:
## v1049 Polish & Trust Pass — Builder, Admin & UX bug-bash

A post-GA **fix-and-harden** pass (no new product surface) filing down the rough edges a first-time evaluator hits on the live demo, plus one security-correctness verification and a regression-coverage pass. Sourced from a founder bug list; each fix carries a regression guard.

### Map Builder
- **BLDR-01** — Expanding a layer-editor panel no longer scrolls the *entire* builder. The app shell used `min-h-screen` and `<main>` hit the flexbox `min-height:auto` trap; scoped a definite viewport height (`h-dvh`/`overflow-hidden`) + `min-h-0` to authenticated map routes so the editor panel scrolls internally. Public map routes keep their footer + document scroll.
- **BLDR-02** — The cluster editor can now color clusters **by `point_count`** with a MapLibre `step` ramp (parity with the "create and style clusters" example). A "Color by cluster size" switch seeds the example tiers (base / 100 / 750) with add/remove/threshold edits; toggling off reverts to the existing flat color (no change for existing maps). New `clusterColorRamp` builder key with camel↔snake parity on both sides (covered by the existing alias-drift guard).

### Admin Console
- **ADM-01** — "Published Maps" was wired to `map_share_tokens`, so it only showed maps with a share link (its own empty-state said "No maps have been published yet"). Re-keyed the listing on `Map` filtered to `visibility=public`, LEFT JOINing the latest share token per map (`DISTINCT ON`, preferring active). Every published map now appears; token fields are nullable; the UI shows a "No link" status for unshared maps. OpenAPI regenerated.
- **ADM-02** — Count badges for **Users / Published Maps / Audit Log** in the Operations sidebar, capped at **999+**. (The Users badge now shows total users; the pending-approval workflow remains on the Users page.)
- **ADM-03** — Removed the "Users by Status" card from the Overview.
- **ADM-04** — User-deactivation failures now surface the real reason (e.g. "Cannot deactivate the last admin user") instead of a generic toast. Backend already returned the specific 400; the frontend was discarding it.
- **ADM-05** — Embedding-Coverage "Regenerate All" showed two spinners (both buttons keyed off one `isPending`). Each spinner now keys off the in-flight mutation variable.
- **ADM-06** — Honest storage display: a "Counts uploaded file size only" note when file storage is 0 B with vector datasets, and an "Over limit" label when the dataset count sits above a later-lowered cap (e.g. 18 / 10). Display-only; cross-type byte accounting stays a deferred backend item.

### Login-mode security
- **SEC-01** — Investigated the report that disabling "Allow Password Login" is bypassable via a login-page link. **It is not**: the link only reveals the form (the intentional, clearly-labeled "Admin? Sign in with a password" break-glass), and the server returns **403** for a non-admin password login even with a correct password. Tightened the existing regression test to pin the exact 403 detail. No server/UI change — the gate is correct by design.

### Search & theme
- **SRCH-01** — Catalog search placeholder → "Search the catalog" (all 4 locales).
- **UX-01** — Theme toggle consolidated to the navbar user dropdown; removed the duplicate Settings "Appearance" card.

### Testing
- Frontend: full vitest green (3191), typecheck, eslint, i18n parity. New/updated regression tests for BLDR-01, BLDR-02, ADM-04, ADM-05, ADM-02, ADM-01, SRCH-01, UX-01.
- Backend: touched suites green `-n 4` (maps, admin invariants, SSO login-mode, builder alias drift); ruff clean; OpenAPI no-drift.

### Out of scope / deferred
- Moving the basemap selector into Map Settings (decided to keep it as a layer-stack row).
- Cross-type (vector/PostGIS) storage-byte accounting (real backend fix; deferred).
- Seeding more demo collections (a demo-content/ops task, not code).

### DIFF
--- a/backend/app/modules/catalog/maps/schemas.py
+++ b/backend/app/modules/catalog/maps/schemas.py
@@ -88,6 +88,7 @@ _BUILDER_CAMEL_TO_SNAKE_KEYS = {
     "clusterColor": "cluster_color",
     "clusterTextColor": "cluster_text_color",
     "clusterTextSize": "cluster_text_size",
+    "clusterColorRamp": "cluster_color_ramp",
     "folderGroupId": "folder_group_id",
     "folderGroupName": "folder_group_name",
     "folderGroupExpanded": "folder_group_expanded",
--- a/backend/app/modules/catalog/maps/schemas.py
+++ b/backend/app/modules/catalog/maps/schemas.py
@@ -1154,12 +1154,15 @@ class OgImageUploadRequest(BaseModel):
 
 
 class AdminShareTokenResponse(BaseModel):
-    id: uuid.UUID
+    # ADM-01: the admin "Published Maps" listing includes public maps that have
+    # no share link, so the token-specific fields are nullable. `created_at` is
+    # the map's creation time (always present).
+    id: uuid.UUID | None = None
     map_id: uuid.UUID
     map_name: str
-    token: str
-    is_active: bool
-    expires_at: datetime | None
+    token: str | None = None
+    is_active: bool | None = None
+    expires_at: datetime | None = None
     created_at: datetime
     created_by: str | None
     embed_token_count: int = 0
--- a/backend/app/modules/catalog/maps/schemas.py
+++ b/backend/app/modules/catalog/maps/schemas.py
@@ -1154,7 +1154,7 @@ class OgImageUploadRequest(BaseModel):
 
 
 class AdminShareTokenResponse(BaseModel):
-    # ADM-01: the admin "Published Maps" listing includes public maps that have
+    # #347 (ADM-01): the admin "Published Maps" listing includes public maps that have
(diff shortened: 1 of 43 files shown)


ishiland:
@codex review

ishiland:
@codex review — addressed both findings: regenerated the Python+TS SDKs for the nullable AdminShareTokenResponse fields (b824f42c), and anchored the milestone finding IDs to #347 in source comments per AGENTS.md (343b41cf).

ishiland:
@codex review — also pushed 02f6a30e fixing the service_public.py line-budget guard bumped by the ADM-01 query (the only red check; documented cap 625→660 per the file's convention).

ishiland:
@codex review

PR #352, opened 2026-06-28
fix(ai): light LLM model defaults to openai_model, not hardcoded gpt-4o-mini

ishiland:
## Problem

On the live demo, "Ask AI" replied *"I was unable to query the data directly"* for a trivial question (top 10 cities by population) — even with the layer referenced explicitly. The data was fine (the Cities layer has `pop_max` + `name`, 1251 rows).

Live API logs showed every `query_data` attempt failing identically:

```
Generating SQL  model=gpt-4o-mini  provider=openai_compatible
query_data.failed  error_type=NotFoundError
  "Error code: 404 - DeploymentNotFound: The API deployment for this resource does not exist."
```

## Root cause

Chat uses **two** models: a main model (`LLM_MODEL`) for conversation and a **separate "light" model** (`LLM_MODEL_LIGHT`) for `query_data` SQL generation and AI metadata.

`LLM_MODEL_LIGHT`'s env-default **hardcoded `"gpt-4o-mini"`** for any non-Anthropic provider — it never read the configured `OPENAI_MODEL`. So with `OPENAI_MODEL=gpt-4.1-mini` and an Azure OpenAI base URL, the main model resolved to a real deployment (chat worked) but the light model stayed `gpt-4o-mini`, which has no Azure deployment → `404 DeploymentNotFound` on every SQL-gen call.

This is a silent split-brain footgun for **any** self-hoster pointing `OPENAI_BASE_URL` at Azure OpenAI, a gateway, or Ollama where the model name must match a real deployment: chat works, `query_data` + metadata silently 404.

## Fix

- `LLM_MODEL_LIGHT` now falls back to `settings.openai_model` (the user's known-good model) instead of a hardcoded name.
- New `OPENAI_MODEL_LIGHT` env / `openai_model_light` setting to opt into a separate cheaper light model; a blank value normalizes to `None` → falls back.
- Anthropic path unchanged (`claude-haiku-4-5-20251001`).

## Tests

`backend/tests/test_llm_model_light_default.py` — covers all three resolution paths (Anthropic → haiku, OpenAI no-override → `openai_model`, explicit override). `test_persistent_config.py` (51) green; ruff clean.

## Demo

The live demo was already fixed out-of-band by setting `llm_model_light` to the working deployment; the exact query now returns Tokyo / New York / Mexico City / … (10 rows). This PR makes the safe behavior the default so it can't recur.

PR #353, opened 2026-06-29
ci: bump enterprise pin to e002 oauth-constraint fix + doc OPENAI_MODEL_LIGHT

ishiland:
Two small follow-ups from the AI light-model session.

## 1. Bump `ENTERPRISE_PINNED_REF` → `be40d109` (consistency fix, not a flake fix)

Pulls in geolens-enterprise **PR #22**, which aligns `e002`'s `chk_oauth_providers_type` with the canonical declaration in core `oauth/models.py` by adding `'github'` to the union. The overlay's `e002` runs on a branch parallel to OSS `0010`; both recreate that constraint, so making e002 write the same full union ensures the overlay can never drop `'github'` whichever branch sorts last.

**Correction:** this was originally filed as a fix for the recurring Pytest Parallel Isolation flake on `test_oauth_github_provider`. Investigation disproved that: PPI runs OSS-only (no overlay), and the failures were the documented **v1020 per-worker-DB-lifecycle race** (3 failures clustered on one worker that lost its DB setup under connection contention), not a constraint issue. So this is a correctness/consistency improvement, **not** a flake fix. `be40d109` = the prior pinned tree + only PR #22; no `EXTENSION_API_VERSION` change.

## 2. Document `OPENAI_MODEL_LIGHT` in `.env.example`

The light-model override added in #352, documented next to `OPENAI_MODEL`. Blank reuses `OPENAI_MODEL`.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -265,6 +265,12 @@ FRONTEND_PORT=8080
 # Type: string | Default: gpt-4o
 # OPENAI_MODEL=gpt-4o
 
+# [OPTIONAL] Lighter LLM model for SQL generation (Ask AI "query_data") + AI metadata.
+# Type: string | Default: reuses OPENAI_MODEL (Anthropic provider uses claude-haiku)
+# Must be a model your provider actually serves; for Azure OpenAI it must match a
+# deployment name. Leave blank to reuse OPENAI_MODEL (recommended).
+# OPENAI_MODEL_LIGHT=
+
 # [OPTIONAL] Custom base URL for OpenAI-compatible providers (Ollama, Groq, Together, etc.)
 # Type: string | No default
 # Example: http://host.docker.internal:11434/v1 (Ollama)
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

ishiland:
Good catch, and it goes deeper — you're right that the PPI job is OSS-only, so this pin bump can't be the PPI-flake fix. Investigating further: the actual PPI flake is OSS-side. `test_oauth_github_migration.py` (the round-trip that downgrades the constraint, removing 'github') and the `test_oauth_github_provider` create tests both hit the *shared* `settings.test_database_url` via autocommit/subprocess connections. Under `-n 4` the round-trip's downgrade window on that shared DB races against github INSERTs from *other* xdist workers — the atomic-method mitigation only covers same-worker ordering. Backend Tests (overlay, serial) is green, which confirms e002 doesn't actually drop github in practice. So: the merged enterprise e002 change is a defensive correctness fix, not the flake fix; the real fix is isolating the round-trip test's DB. Reframing this PR accordingly.

PR #366, opened 2026-06-29
[codex] Fix Service URL progress and large ArcGIS imports

ishiland:
## Summary

Fixes the Service URL import path for large remote layers in two parts:

- publishes bounded progress during the long service ogr2ogr loading phase so the UI no longer appears stalled at the initial value
- chunks large ArcGIS FeatureServer and MapServer imports into paged resultOffset / resultRecordCount queries, respecting the service maxRecordCount, and appends later chunks into the same target table
- keeps WFS and OGC API Features on the existing single-GDAL import path; the progress heartbeat still applies to all service imports

## Root Cause

Service imports only wrote progress at the start of ogr2ogr and then again during finalization. Large ArcGIS layers also ran as one unbounded ESRIJSON import, so a slow upstream service could hit the 1800 second subprocess timeout.

## Validation

- POSTGRES_PORT=5434 uv run pytest tests/test_ingest_progress.py tests/test_arcgis_auth.py::test_build_gdal_source_arcgis_result_offset tests/test_ogr_subprocess_env.py tests/test_ingest_service_geometry_type.py -q
- uv run ruff check app/processing/ingest/tasks_vector.py app/processing/ingest/ogr.py app/modules/catalog/sources/preview.py app/modules/catalog/sources/adapters/arcgis.py app/platform/extensions/defaults.py app/core/processing_port.py tests/test_ingest_progress.py tests/test_arcgis_auth.py
- uv run ruff format --check app/processing/ingest/tasks_vector.py app/processing/ingest/ogr.py app/modules/catalog/sources/preview.py app/modules/catalog/sources/adapters/arcgis.py app/platform/extensions/defaults.py app/core/processing_port.py tests/test_ingest_progress.py tests/test_arcgis_auth.py

## Follow-up

OGC API Features chunking is intentionally out of scope for this PR because that needs provider-specific handling of next links or pagination rather than ArcGIS REST offsets.

### DIFF
--- a/backend/app/core/processing_port.py
+++ b/backend/app/core/processing_port.py
@@ -317,6 +317,7 @@ class ProcessingPort(Protocol):
         token: str | None = None,
         order_field: str | None = None,
         result_limit: int | None = None,
+        result_offset: int | None = None,
     ) -> tuple[str, str]: ...
 
     # -------------------------------------------------------------------------
--- a/backend/app/modules/catalog/sources/adapters/arcgis.py
+++ b/backend/app/modules/catalog/sources/adapters/arcgis.py
@@ -62,6 +62,31 @@ def _normalize_esri_geom_type(esri_type: str | None) -> str | None:
     return _ESRI_GEOM_TYPE_MAP.get(esri_type, esri_type)
 
 
+def _extract_arcgis_object_id_field(data: dict) -> str | None:
+    value = data.get("objectIdField")
+    if isinstance(value, str) and value.strip():
+        return value.strip()
+
+    fields = data.get("fields")
+    if isinstance(fields, list):
+        for field in fields:
+            if not isinstance(field, dict):
+                continue
+            if field.get("type") != "esriFieldTypeOID":
+                continue
+            name = field.get("name")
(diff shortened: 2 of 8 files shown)


ishiland:
@codex review

ishiland:
@codex review

Updated the PR to require `advancedQueryCapabilities.supportsPagination` plus a positive `maxRecordCount` before using ArcGIS offset pagination. Unsupported or unknown pagination falls back to the original single-GDAL import path.

ishiland:
@codex review

Addressed the ordering review comment: ArcGIS chunked imports now require the layer metadata endpoint to provide a stable order field (`objectIdField`, an `esriFieldTypeOID` field, or `uniqueIdField`) before using `resultOffset`/`resultRecordCount`. If the order field is unavailable, the worker falls back to the original single GDAL import path.

ishiland:
@codex review

Pushed one additional CI fix: the service-ingest heartbeat broad `except Exception` now carries the required same-line `# broad:` justification. Local verification: the layering invariant passes with `POSTGRES_PORT=5434`, plus the changed ingest/ArcGIS test set, ruff check, and ruff format check.

PR #371, opened 2026-07-01
fix(api): hide enterprise ops from public schema, bump 1.4.2, widen deployed gate

ishiland:
## Summary

Closes the source-repo half of the announcement re-audit blockers so the OSS launch surface is clean. Paired with `getgeolens.com` PR (docs/marketing redeploy + contract sync).

## Changes

**Hide enterprise/edition operations from the public API surface (APIDOCS)**
- The trailing-slash `convert-saml-to-local` route and the `audit-logs/export/{format}` route were still advertised in the public OpenAPI schema (and the generated docs API reference). Added `include_in_schema=False` to both, matching the already-hidden non-slash SAML route, `Edition Info`, and `Get Enterprise Only Tabs`.
- The routes stay registered and functional at runtime for admin/enterprise-overlay callers (verified: both paths remain in the app route table). `List Audit Logs` (OSS viewing) is retained.
- Regenerated `backend/openapi.json` and the Python/TypeScript SDKs. `make openapi-check` and `make sdks-check` are green.

**Patch release prep (REL-01)**
- `make bump VERSION=1.4.2` across all 10 version sites; `make version-check` green. CHANGELOG `[1.4.2]` added. Publishing (PyPI/npm/GHCR + GitHub Release) happens on the `v1.4.2` tag push after merge; the `geolens-cli` long description regenerates from the already-clean `cli/README.md`, clearing the immutable stale "community or enterprise" wording on PyPI.

**Widen the deployed-surface gate (GATE)**
- `scripts/deployed_surface_gate.json` gains eight pages the prior gate did not cover (which is why the home page could pass while stale copy survived elsewhere): `/features/`, `/quickstart/`, localized `/es|fr|de/`, `docs …/guides/api/`, and the two PyPI project pages.
- Added `pypi.org` to `ALLOWED_DEPLOYED_HOSTS`. Patterns use precise operation/leak labels rather than bare `Enterprise`/`SAML` to avoid false-positives on legitimate copy. Offline unit test updated (32 tests green).

## Schema/API impact

Two admin/enterprise operations leave the public OpenAPI schema and the generated SDKs (they remain callable at runtime). `openapi.json` `info.version` and `docs-contract.json` move to `1.4.2`.

## Verification

- `make openapi-check` / `make sdks-check` / `make version-check` — all exit 0
- Route-registration check: both hidden routes still in `app.api.main:app` route table
- `python3 -m unittest tests.test_check_public_surface tests.test_check_deployed_surface` — 32 OK
[...]

PR #372, opened 2026-07-01
fix(gates): check PyPI description via JSON API, not the HTML project page

ishiland:
## Summary

Follow-up to #371. The `pypi_geolens_cli` / `pypi_geolens` deployed-surface checks were vacuously passing, which is a false-confidence gap in the launch gate.

## The bug

The two PyPI checks pointed at the `/project/<name>/` HTML pages. `pypi.org` serves a ~3KB minimal response to non-browser fetches (the checker's `urllib` UA), so the long description is absent from the fetched HTML and the `community or enterprise` forbidden pattern never matched — the gate passed even though the live `geolens-cli` 1.4.1 description still contains "community or enterprise".

## The fix

Point both checks at the PyPI JSON API (`/pypi/<name>/json`), whose `info.description` carries the full rendered long description and is served in full to programmatic clients (host stays `pypi.org`, already allowlisted).

## Verification

- `python3 -m unittest tests.test_check_public_surface tests.test_check_deployed_surface` — 32 OK
- Live gate now correctly **FAILS** on the stale `geolens-cli` 1.4.1 description (`community or enterprise` matched via JSON) and **passes** on `geolens`. It clears once `1.4.2` publishes a clean description from `cli/README.md`.

This makes GATE-04 actually detect the PyPI blocker instead of silently passing.

### DIFF
--- a/scripts/check_deployed_surface.py
+++ b/scripts/check_deployed_surface.py
@@ -16,7 +16,7 @@ import sys
 from dataclasses import dataclass
 from html.parser import HTMLParser
 from pathlib import Path
-from typing import Callable, Sequence
+from typing import Callable, Iterator, Sequence
 from urllib.parse import urljoin, urlparse
 from urllib.error import HTTPError, URLError
 from urllib.request import HTTPRedirectHandler, Request, build_opener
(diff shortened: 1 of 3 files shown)


ishiland:
@codex review

ishiland:
Thanks @codex — fixed in the latest commit. `fetch_url` is now content-type aware: for JSON responses it parses and joins the decoded string values, so escapes like `community\nor enterprise` are resolved before the assertions run (non-JSON and malformed-JSON fall back to raw text). Added unit coverage for the escaped-newline case and the non-JSON passthrough. Verified the live gate still catches the current stale `geolens-cli` 1.4.1 description via the decoded JSON.

@codex review

ishiland:
@codex the fix is in `scripts/check_deployed_surface.py` (`fetch_url` → `decode_scannable_text`), not the config line this comment re-anchored to. For a JSON content-type, the response is now `json.loads`-parsed and its string values joined, so `community\nor enterprise` is decoded to a real newline before scanning. Proven by `test_json_response_is_decoded_before_scanning` (escaped phrase → matches) and `test_non_json_response_is_passed_through_unchanged` (raw/malformed fallback). CI is green. Merging.

PR #373, opened 2026-07-01
fix(collections): invalidate catalog cache on create

ishiland:
## Problem

Creating a collection in the demo showed the success toast but the collection appeared to never be created.

## Root cause

The collections **list** endpoint serves admins a 60-second Valkey cache keyed `catalog:collections:admin:{skip}:{limit}`. The **create** handler was the only mutation that never called `invalidate_catalog_cache()` — `update`, `delete`, and both dataset-membership handlers already do.

So for an admin, a freshly created collection stayed hidden behind the stale cache for up to 60s: the `201` and success toast were real and the row **was** persisted, but the list kept returning the cached page without it — looking like the create silently failed. Non-admins hit no server cache, so their creates showed instantly (which is why this only reproduces on the demo's admin account).

## Fix

One line: `await invalidate_catalog_cache()` after the create commits, matching the sibling handlers.

## Test

`test_create_collection_invalidates_admin_list_cache` populates the admin list cache, creates a collection, and asserts the next list reflects it. Verified it **fails without the fix and passes with it**. Full `TestCreateCollection` suite green; ruff clean.

### DIFF
--- a/backend/app/modules/catalog/collections/router.py
+++ b/backend/app/modules/catalog/collections/router.py
@@ -104,6 +104,10 @@ async def create_collection_endpoint(
         )
         await db.commit()
         await db.refresh(collection)
+        # Without this, an admin's newly created collection stays hidden behind
+        # the 60s catalog:collections:admin cache until it expires. Matches the
+        # invalidate call in the update/delete/dataset-membership handlers.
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

Issue #380, opened 2026-07-01
Document the &cols= opt-in for attribute columns on low-zoom vector tiles


ishiland:
Vector tiles strip attribute columns below zoom 10 to keep low-zoom tiles small. Callers who need attributes at low zoom can opt in per column by appending `&cols=<column>` to the tile URL.

This behavior isn't documented anywhere user-facing. Add a short note (README API section or a docs-site pointer) describing the default and the opt-in.

**Acceptance:** the note is accurate against the tiles behavior in `backend/app/processing/tiles/`.

**Start here:** `ARCHITECTURE.md`, "where do I change X?" table, *tiles* row.

Issue #379, opened 2026-07-01
Surface the edition badge in the admin overview


ishiland:
Show a small badge on the admin overview page reflecting `EditionInfo.edition` ("Community" or "Enterprise"), fetched via the existing `fetchEdition()` helper.

**Files:** `frontend/src/pages/admin/AdminOverviewPage.tsx`, `frontend/src/api/edition.ts`

**Acceptance:** the badge renders the current edition; frontend tests pass (`cd frontend && npx vitest run`). New user-visible strings need entries in all four locales (en/es/fr/de).

**Start here:** `ARCHITECTURE.md`, "Frontend" section, plus the "where do I change X?" table's *admin setting* row.

Issue #378, opened 2026-07-01
Document the migrate container's expected exit in the development setup guide


ishiland:
The `migrate` service in the compose stack is a one-shot container: it runs `alembic upgrade heads` and then exits, and the `api`/`worker` services wait on its successful completion. New contributors sometimes read that exit as a crash.

Add a short troubleshooting subsection to `.github/CONTRIBUTING.md` (Development Setup) explaining that the exit is expected and showing how to read its logs with `docker compose logs migrate`.

**Acceptance:** the subsection renders correctly and reads clearly.

**Start here:** docs only, `.github/CONTRIBUTING.md` under "Development Setup".

PR #381, opened 2026-07-01
docs(contributing): match the real good-first-issue label name

ishiland:
One-line follow-up from the announcement-readiness audit (P2-1).

CONTRIBUTING told newcomers to search for a hyphenated `good-first-issue` label, but the repo's label is the spaced `good first issue`, so the documented search returned an empty list. The text now links the label query directly. Three starter issues (#378, #379, #380) are filed under the label, so the on-ramp is no longer empty.


ishiland:
@codex review

PR #387, opened 2026-07-02
fix(ai): drop temperature on Anthropic path, Decimal-safe serialization, plain-text chat output

ishiland:
## What

Three small, self-contained fixes to the AI chat stack, all verified live against the local dev stack (end-to-end Ask AI spatial query returning Decimal-carrying SQL results):

1. **Drop `temperature` at the three Anthropic call sites** (`DefaultAnthropicProvider.complete`, its structured-output method, and `_stream_anthropic_chat`). Claude 4.6+ models reject a non-default `temperature` with a 400. The provider-agnostic `complete()`/`structured_complete()` signatures are unchanged — the Anthropic provider now discards the kwarg, while the OpenAI-compatible path (`DefaultOpenAICompatibleProvider`, `_stream_openai_chat`) still sends it, which gpt-4.1-mini accepts.
2. **`json.dumps(..., default=str)` at tool-result and SSE sites** (both stream emitters in `processing/ai/router.py`, the Anthropic `tool_result` content and OpenAI `tool` message content in `streaming.py`). `query_data` rows can carry Postgres `numeric` (Decimal) and `datetime` values straight from PostGIS, which crashed serialization on every provider path.
3. **Plain-text rule in the chat system prompt** (`build_chat_system_prompt`). ChatPanel renders message content raw, so markdown arrived as literal `**bold**`/backticks on every provider.

## Impact

- No schema, API-contract, or config changes. No OpenAPI/SDK regeneration needed.
- Fixes two bugs that are currently live on the demo's Azure OpenAI path (Decimal crash, markdown artifacts) — should be deployed to the demo VM after merge.

## Verification

- `cd backend && uv run ruff check .` and `uv run ruff format --check .` — green
- Pre-commit hooks (SSRF rule, secrets, ruff) — green
- Live: restarted `api` on the dev stack; AI chat answers in plain text and a spatial `query_data` question returning Decimal columns streams a results table without error; Anthropic provider (`claude-sonnet-5`) no longer 400s

### DIFF
--- a/backend/app/platform/extensions/defaults.py
+++ b/backend/app/platform/extensions/defaults.py
@@ -1004,6 +1004,7 @@ class DefaultAnthropicProvider:
         base_url=None,
         temperature=0.5,
     ):
+        del temperature  # rejected by Claude 4.6+; kept in signature for callers
         # Deferred imports (Phase 214 discipline)
         import json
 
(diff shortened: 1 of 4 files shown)


ishiland:
@codex review

PR #394, opened 2026-07-03
fix(builder): remediate 2026-07-03 audit findings (B-019..B-032)

ishiland:
Remediates every action item from the 2026-07-03 builder deep-dive audit (the follow-up to #392's 2026-07-02 remediation): the P1 set **B-019..B-023**, the P2 set **B-024..B-031**, and the **B-032** LOW rollup. Prior deferred items B-010..B-018 remain tracked in their existing todos per the audit's own guidance, except the pieces the audit explicitly bundled here (UX-02 desktop Escape and ST-02 editor typing — closing out **B-015** — plus LM-02 from B-016 and UX-03 from B-017).

## P1

| ID | Sev | Fix |
|----|-----|-----|
| **B-019** (VT-01) | **HIGH** | Reupload (both file + service tasks) now purges the MVT tile cache **post-commit** via a shared helper — it was the one write path that never invalidated, so stale tiles kept 304-serving for up to `tile_cache_ttl` after every reupload. |
| B-020 (FL-01) | MED | Data filters are no longer ANDed into cluster/cluster-count layers (adapter + live handler) — cluster features carry only `point_count`/`cluster_id`, so any predicate blanked every bubble. Documented ceiling: cluster counts include filtered-out points. |
| B-021 (LM-01) | MED | "Add to group…" kebab splices the row after the group's last child + renumbers `sort_order` (mirror of #392's drag-path fix); Test E now asserts position and a new Test E2 covers the kebab path. |
| B-022 (SH-02) | MED | `X-Embed-Token` is only attached to first-party requests (own origin / configured tile CDN) — third-party basemap sprite/glyph/tile CDNs no longer receive the credential. |
| B-023 (SH-01) | MED | **Decision: honor the token** (not soften the copy). `GET /maps/shared/{token}` accepts `X-Embed-Token` and unions the token's scoped dataset layers into the payload — embed tokens are a private-dataset **capability** (SEC-022 posture, EMBED-02/04) that the tile path has always honored; the metadata payload now matches, so a dataset downgraded to private after publish renders in embeds exactly as the SharePanel copy promises. Fail-closed: active + map-bound + origin-checked, and the join still requires current map membership + map publicity. |

## P2

- **B-024** (ST-01, both sides): zero surviving icon pairs → flat fallback instead of a spec-invalid length-3 `match` that made `addLayer` throw (swallowed) and the symbol layer silently vanish. Plus **ST-04**: `to-string` match input + stringified labels so numeric category columns actually match (frontend adapter + `style_json.py` export, byte-parity).
- **B-025** (VT-05): `ST_MakeValid` guards the simplify input — one invalid geometry can no longer 503 a whole tile.
- **B-026** (VT-02/03/04): `tile_version` (`Dataset.current_version`) wired into `MapLayerResponse` + `SharedLayerResponse` — the builder's existing `_v=` cache-buster read finally has a producer, and the viewer threads it too. Viewer now derives its MVT source-layer via `getMvtSourceLayerName()` (every frontend hardcode of `` `data.${table}` `` converted), with a parity test pinning the contract against the server emission.
- **B-027** (UX-01): builder sidebar wrapped in a recoverable `PanelErrorBoundary` — a stack-panel render error no longer takes down the whole builder.
- **B-028** (LM-02): ephemeral query overlay re-adds itself on every `style.load` (basemap switches wiped it while the badge stayed); zooms only on first show.
- **B-029** (UX-04): shortcuts sheet documents the full existing keyboard model (`?`, reorder mode, Escape, ⌘/Ctrl+Click, Shift+Click, Shift+↑/↓).
- **B-030** (CH-01/02/03): `fill-extrusion-*` survives chat paint validation (the extrusion companion consumes it; mirrors the backend allowlist); `set_label` validates column against the layer schema + sanitizes `textColor` on **both** sides (backend feeds the model a retryable error); `set_data_driven_style` requires the target layer + a real mutation before reporting "applied".
- **B-031** (ST-02, closes half of B-015): Advanced JSON editor validates against the **rendered** adapter type (heatmap/symbol; cluster-as-circle) instead of raw geometry. Bundled per the audit: **UX-02** Escape-to-close on the desktop editor flyout (closes B-015's other half) and rail panels (UX-07).

## B-032 LOW rollup — all addressed

LM-03 (announce real drop position) · LM-04 (group eye = children aggregate, render + toggle) · LM-05 (keyboard reorder over the rendered stack) · FL-02 (empty-state copy says "in view") · FL-03 (numeric `in_list` per-entry NaN guard — the whole-string check also rejected *valid* numeric lists) · LB-10 (restore branch re-asserts label filter) · ST-03 (expression opacity × master slider) · ST-05 (validator crash blocks apply) · ST-06 (cluster export limitation documented) · SH-04 (batch token endpoint accepts `X-Embed-Token` → embed terrain gets the real DEM bounds/maxzoom descriptor instead of empty defaults) · CH-11 (staging no longer kills undo; downgrade moved to the accept path) · CV-01 (`og_image_url` added to the `MapResponse` mirror) · PF-05 (lazy-seed bounds ref) · VT-06 (NULL post-clip geometries filtered) · VT-07 (positive, offset cluster feature ids) · VT-08 (tenant-prefixed cache keys invalidated) · UX-05/UX-08 (decision records: fixed panel widths + layer-owned terrain documented as intentional) · UX-06 (spinner aria-label localized) · exported vector-source `maxzoom` aligned with the live builder (14 plain / 22 cluster, per the audit's orchestrator observation).

## API / schema impacts

- `GET /maps/shared/{token}` and `POST /tiles/tokens/` accept an optional `X-Embed-Token` header (fail-closed capability fallback; anonymous behavior unchanged without it).
- `MapLayerResponse.tile_version` + `SharedLayerResponse.tile_version` (additive, nullable).
- `backend/openapi.json` + generated SDKs regenerated accordingly. No DB migrations (reads the existing `datasets.current_version`).
- New i18n keys in all four locales (en/es/fr/de).

## Verification

- Backend: `uv run ruff check .` + `format --check` clean; targeted suites green (`test_shared_map_embed_scope.py` **new**, `test_mvt_audit_fixes.py`, `test_tile_cache.py`, `test_collect_chat_action.py`, `test_maps_style_json.py`, `test_phase_274_tile_cache.py` = 131 passed) plus regression sweep over changed areas (`test_maps.py`, `test_embed_tokens.py`, `test_vector_tile_auth.py`, `test_sec024_terrain_config_privacy.py`, `test_embed_framing_csp.py`, `test_ai_chat_actions_phase1135.py`, `test_reupload*.py` = 252 passed).
- Frontend: `npm run typecheck` (tsc -b), `npm run lint`, full `vitest` **3275/3275**, `npm run test:i18n` locale parity.
- `make openapi-check` + `make sdks-check` pass.
- E2E: `npm run e2e:smoke:builder` **26/26** against the restarted dev stack (one spec updated for the new radio-role semantics).

Audit source: `docs-internal/audits/builder-audit-20260703.md` (internal, not in repo) — finding IDs in code comments are anchored to this PR number per the inline review-comment convention.

### DIFF
--- a/backend/app/modules/catalog/features/router.py
+++ b/backend/app/modules/catalog/features/router.py
@@ -4,19 +4,33 @@ import uuid
 from urllib.parse import urlencode
 
 import structlog
-from fastapi import APIRouter, Depends, HTTPException, Query, Request, Response, status
+from fastapi import (
+    APIRouter,
+    Depends,
+    Header,
+    HTTPException,
+    Query,
+    Request,
+    Response,
+    status,
+)
 from fastapi.responses import JSONResponse
 from sqlalchemy.exc import OperationalError, ProgrammingError
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from app.modules.audit.service import AuditEvent, audit_emit
 from app.core.identity import Identity
-from app.modules.auth.dependencies import get_current_active_user, require_permission
+from app.modules.auth.dependencies import (
+    get_current_active_user,
+    get_optional_user,
+    require_permission,
+)
 from app.modules.catalog.authorization import (
     check_dataset_access,
     check_dataset_write_access,
 )
 from app.modules.catalog.datasets.domain.service import get_dataset
+from app.modules.embed_tokens.service import validate_embed_token_access
 from app.core.dependencies import get_db
 from app.modules.catalog.features.schemas import (
     FeatureCreate,
@@ -58,10 +72,28 @@ features_router = APIRouter(prefix="/datasets", tags=["Features"])
 )
 async def get_features_geojson_z_endpoint(
     dataset_id: uuid.UUID,
-    user: Identity = Depends(get_current_active_user),
+    request: Request,
+    user: Identity | None = Depends(get_optional_user),
+    embed_token: str | None = Header(
+        default=None,
+        alias="X-Embed-Token",
+        description=(
+            "Optional embed token. Datasets in the token's scope are "
+            "authorized even without user credentials (embed viewers)."
+        ),
+    ),
     db: AsyncSession = Depends(get_db),
 ) -> JSONResponse:
-    """Return up to 5,000 features as RFC 7946 GeoJSON with Z coordinates."""
+    """Return up to 5,000 features as RFC 7946 GeoJSON with Z coordinates.
+
+    fix(#394) codex P2: the viewer's bounded-GeoJSON path (small 3D layers,
+    eligible cluster layers) already sends ``X-Embed-Token``, and the B-023
+    shared-map union now exposes embed-scoped private layers to embeds — so
+    this endpoint accepts the token as fallback authorization via the SAME
+    ``validate_embed_token_access`` capability check as tile serving.
+    Credentialed callers (JWT / API key) keep the exact prior RBAC path;
+    anonymous callers without a valid scoped token still get 401.
+    """
     dataset = await get_dataset(db, dataset_id)
     if dataset is None:
         raise HTTPException(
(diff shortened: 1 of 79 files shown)


ishiland:
@codex review

ishiland:
@codex review

Both P2s addressed in 2e68003b:
- **Bounded GeoJSON**: `/datasets/{id}/features.geojson` now accepts `X-Embed-Token` via the same `validate_embed_token_access` capability check as tile serving (the viewer already sent the header) — regression test added.
- **tile_version**: the `_v=` cache-buster is threaded through every vector URL build path now — builder `toSyncInput()`, the builder token-refresh/retry rebuild, the viewer token-refresh rebuild, and the render-mode fallback.

Also fixed the two CI failures (the LOC-budget layering guard): router additions compressed back under their hard ceilings; reviewed cap bumps recorded for `service_public.py` and `chat_actions.py` per the guard's escalation convention.

ishiland:
@codex review

Round-2 P2 addressed in 6d0aef46: label colors now validate against the real CSS named-color set (hex restricted to 3/4/6/8 digits, functional args to numeric/separator chars) — `notacolor` / `rgb(foo)` fall back to the default instead of reaching MapLibre. Shared backend module `processing/ai/colors.py` + byte-parity ChatPanel mirror, with negative-case tests.

PR #397, opened 2026-07-04
feat(frontend): unify the UI on the survey-instrument design language

ishiland:
## Summary

One-pass UI redesign that unifies the app on the "survey instrument" language the dataset page and import flow already used: mono-caps eyebrow labels, hairline survey grids, mono readouts for measured values (counts, EPSG codes, table names), squared chips, and a tick-rule "neatline" divider as the signature element. The older soft-pill generation (rounded-22px cards, translucent surfaces) and the stock-shadcn shell are gone.

This is not a reskin. The pass also surfaces dataset provenance as a first-class concept and makes the admin area read like part of the product.

## What changed

**Tokens and primitives** (`index.css`, `ui/badge`, `ui/table`)
- New `.eyebrow`, `.readout`, and `.tick-rule` utilities; base radius 10px → 8px.
- Badge goes from pill to squared chip. Table column headers go mono-caps. Both propagate app-wide, including the builder and admin, with no per-page edits.

**Shell**
- Navbar is a full-bleed control bar with baseline-marker active tabs (same underline vocabulary as the import mode tabs). Admins get a visible Admin entry in the top nav and mobile sheet instead of only the user dropdown.
- PageHeader closes with the tick-rule neatline; this lifts Collections, Maps, Settings, and all eight admin pages at once. Footer shows the instance version (`v1.4.2`) as a readout.

**Dataset origin, end to end** (the PostGIS-native ask)
- New `source_format` field on OGC search records (additive; `openapi.json` and both SDKs regenerated).
- `datasetOrigin()` derives Uploaded / PostGIS / Service / STAC / Created from persisted fields. Registered PostGIS tables store no `source_format`, VRTs are composed so they show the type badge alone.
- `OriginBadge` renders on catalog cards and the dataset header, with explanatory tooltips ("Registered from an existing PostGIS table. The data stays in your database").
- Dataset headers also show the live PostGIS table name as a copyable mono readout for vector and table datasets, so QGIS/psql/ogr2ogr users can grab it without digging.

**Catalog search**
- Filter rail and search card drop the soft-pill styling for crisp bordered cards; rail section labels use the eyebrow, facet counts are right-aligned mono readouts.

**Admin control plane**
- Overview stats become a bordered survey grid (same pattern as the dataset stats bar). Sidebar group labels use the eyebrow. Health rows keep their latency readouts and status dots.

**States and a11y**
[...]

PR #399, opened 2026-07-04
fix(frontend): restore AA contrast on footer version readout

ishiland:
## What

The footer version badge (`AppFooter.tsx`) was set to `text-muted-foreground/80` in the #397 redesign. At `/80` opacity it composites to **4.35:1** against the footer background — under the 4.5:1 WCAG AA threshold. The axe suite (now enforced in CI, not `if:false`) fails all four public-page a11y checks on this one global element, which has left `main` red since #397.

## Fix

Drop the `/80` opacity. The full `--muted-foreground` token measures **7.13:1** against the same background.

Only this element was flagged by axe; the remaining `text-muted-foreground/80` uses on non-public-page components are left to the scheduled GLUX-012 contrast sweep.

## Verify
- Contrast computed from `--muted-foreground: oklch(0.45 0.005 250)` over the axe-measured `#fbfaf8` bg → 7.13:1.
- CI axe run is the gate; this PR turns `main` green.

https://claude.ai/code/session_01TRCpoeiQ6bpLZEHr5ZBj2T

### DIFF
--- a/frontend/src/components/layout/AppFooter.tsx
+++ b/frontend/src/components/layout/AppFooter.tsx
@@ -54,7 +54,7 @@ export function AppFooter({
         ))}
         {/* Instance version readout — self-hosters see at a glance what they run. */}
         <span aria-hidden="true">·</span>
-        <span className="readout text-muted-foreground/80">v{__APP_VERSION__}</span>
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review

PR #405, opened 2026-07-05
docs(editions): document Community/Enterprise split and open-source commitment

ishiland:
## What

Expands `EDITIONS.md` (the open-core boundary doc, REL-01) from a one-line distribution note into an explicit edition split plus a written open-source commitment:

- **Community** — this repo, Apache-2.0, complete and self-hostable indefinitely.
- **Enterprise** — a separate, optional commercial overlay (support/SLA, SAML SSO, audit-log export, branding removal, single-tenant hardening). Nothing in Community is disabled without it.
- **Commitment** — the open standards (OGC API Features/Records, STAC, DCAT) and data portability stay in the Apache-2.0 core; features that are free today will not be moved behind the paid edition.

## Why

Makes the free/paid boundary legible to prospective buyers so it is never a surprise.

## Impact

Docs only. No schema, API, or config changes.


ishiland:
Good catch — reproduced locally (`scripts/check_public_surface.py` failed with 5 violations). Reworded so the open-source commitment stays (standards stay free, free features stay free, no lock-in, portable) but without the public-surface-gate-banned positioning terms. Gate now passes: `public-surface gate passed (67 tracked file(s) scanned)`. The paid side is still disclosed, just in OSS-positive language.

@codex review

PR #423, opened 2026-07-06
build(docker): align python version references with the shipped 3.14 image

ishiland:
## What & why

Bump #255 moved the runtime base to `python:3.14.6-slim` but left several mirrors of that version behind, so the repo advertised three different Python versions for the same image (README `3.13`, Dockerfile/pyproject comments `3.14.3`, CVE scan `3.14.5`). This is the concrete runtime/version drift flagged in a recent external audit.

Root cause: Dependabot bumps the Dockerfile `FROM` line but not the prose comments or the workflow scan tag that mirror it. This fixes the values and makes the prose drift-proof (name only the minor; point at `FROM` as the source of truth).

## Changes

- **Dockerfile / `backend/pyproject.toml`** — drop the stale `3.14.3` patch pins from comments; name only `3.14` and point at the `FROM` line as the source of truth.
- **README (en + es/fr/de)** — "Python 3.13 bundled" → `3.14` (the image ships 3.14, not 3.13); badge `backend 3.13` → `3.13+` so it reads as the `requires-python>=3.13` floor rather than a contradiction of the shipped 3.14.
- **`.github/workflows/dep-audit.yml`** — the *enforcing* Trivy scan targeted `python:3.14.5-slim`, a base we no longer ship. Bumped to `3.14.6-slim` (matches the Dockerfile) and added a note that the node/python/nginx pins must track the Dockerfile `FROM` tags. node + nginx were verified already in sync.

## Impact

- No product version change — still `1.4.2`; `make version-check` unaffected.
- Config/CI: the CVE gate now scans the base image we actually ship (small behavior fix).
- The `FROM python:3.14.6-slim` lines — the actual tested runtime — are unchanged.

## Verification

- No `3.14.3` / `3.14.5` leftovers remain in Dockerfile, `pyproject.toml`, or `dep-audit.yml`.
- All `3.14.6` references (both Dockerfile `FROM`s + the Trivy scan) now agree.
- Version surfaces still `1.4.2` across frontend/backend/cli/sdks.

https://claude.ai/code/session_01N161FvzP863aYUDGCAboBy

### DIFF
--- a/.github/workflows/dep-audit.yml
+++ b/.github/workflows/dep-audit.yml
@@ -87,7 +87,9 @@ jobs:
         # `enforce: "1"` fails the job on a fixable CRITICAL CVE; `enforce: "0"`
         # is report-only. The slim/alpine images we pin (node/python/nginx) are
         # controllable — a CRITICAL there is actionable (bump the tag), so they
-        # ENFORCE. postgis/postgis:17-3.5 is an uncontrollable third-party fat
+        # ENFORCE — keep these three pins matched to the Dockerfile `FROM` tags
+        # they mirror (bump together). postgis/postgis:17-3.5 is an
+        # uncontrollable third-party fat
         # Debian-11 image whose CRITICAL set is dominated by CVEs in baked-in Go
         # binaries that only the upstream maintainer can rebuild — and that set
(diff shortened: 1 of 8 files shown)


ishiland:
@codex review

ishiland:
@codex review

PR #432, opened 2026-07-09
fix(import): queue drops staged during the upload-config fetch instead of swallowing them

ishiland:
## What

The /import dropzone could **silently swallow files** (no toast, no request, no recovery without a page reload): `useUploadConfig` runs with `staleTime: 0` + `refetchOnMount: 'always'` (quota freshness, PR #274), and the dropzone was `disabled` during that fetch — react-dropzone ignores drops while disabled. Any file staged in the boot-time refetch window was dropped on the floor. This is the product fix for the race the PR #430 e2e suite had to work around with reload-retry.

## How

- `FileDropzone` is no longer disabled during the config fetch (the `disabled` prop is deleted — this was its only user).
- `UploadForm` queues files accepted while `isFetching` and flushes them once the query settles; multiple drops in the window **merge** rather than overwrite.
- The flush re-applies the batch cap against the **fresh** quota via a shared `effectiveBatchLimit` helper; over-cap batches are rejected whole with a toast, matching react-dropzone's own `maxFiles` semantics. So the original Codex P2 guarantee on #274 — never act on an unresolved/stale quota — still holds; the difference is queued-then-validated instead of silently discarded.

No schema/API/i18n impact (the toast reuses the existing `dropzone.batchLimit` key).

## Verification

- `npm run typecheck` / `npm run lint` / `npm run build` — clean
- `npm run test:coverage` — full suite passes
- New regression suite `UploadForm.pendingConfigQueue.test.tsx` (4 tests): queue-then-process, merge of same-window drops, whole-batch rejection when the fresh quota is exceeded, immediate processing when already settled

@codex review

https://claude.ai/code/session_01HSodEnENJNBi95hS2xqATn

### DIFF
--- a/frontend/src/components/import/FileDropzone.tsx
+++ b/frontend/src/components/import/FileDropzone.tsx
@@ -11,9 +11,21 @@ import type { DataKind } from '@/types/api';
 /** Client-side batch guard. Real caps (per-file size, dataset quota) are server-enforced. */
 const MAX_BATCH_FILES = 25;
 
+/**
+ * Batch cap = the smaller of the UX limit and remaining quota. Floor at 1
+ * (not 0 — react-dropzone treats maxFiles:0 as unlimited) so an at-cap user
+ * can still drop one file and get the explicit server quota error + banner.
+ * Shared with UploadForm's queued-drop flush (PR #274 follow-up) so both
+ * enforce the same limit.
+ */
+export function effectiveBatchLimit(remainingQuota: number | null | undefined): number {
+  return remainingQuota != null
+    ? Math.min(MAX_BATCH_FILES, Math.max(1, remainingQuota))
+    : MAX_BATCH_FILES;
+}
+
 interface FileDropzoneProps {
   onFilesAccepted: (files: File[]) => void;
-  disabled?: boolean;
   allowedExtensions?: string[];
   maxSizeMb?: number;
   /** Remaining per-user dataset quota; null = unlimited. Caps the batch so the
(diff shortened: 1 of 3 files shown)


ishiland:
@codex review

ishiland:
Addressed the P2 in 0bf70d94: while the quota query fetches, the dropzone gets a permissive cap (null → 25-file UX limit) instead of the cached quota, so a stale-LOW value can't reject a multi-file drop at the react-dropzone layer before the pending queue re-validates it against the fresh quota at flush. Added a prop-wiring regression test (import suite now 81 passing; typecheck/eslint clean).

@codex review

ishiland:
Addressed the round-2 P2 in 312c8499: the flush now re-validates queued files per-file against the settled config's extension list and per-file size cap (same rejection toast as react-dropzone), before the existing whole-batch quota clamp — so every gate the dropzone couldn't enforce during the fetch window is applied against fresh values at flush. Regression test added with a mixed batch (bad extension + oversized + valid); import suite 82 passing, typecheck/eslint clean.

@codex review

ishiland:
Addressed the round-3 P2 in d39939ed: allowedExtensions/maxSizeMb are now also withheld from the dropzone while the config query fetches (matching the quota prop), so no stale cached gate can reject a drop before it queues — the flush is the single client-side enforcement point for fetch-window drops, validating extension + size + quota against the settled config. Prop-wiring test extended to pin all three gates in both states; import suite 82 passing, typecheck/eslint clean.

@codex review

PR #436, opened 2026-07-09
docs(agents): Rule 3 — describe the MinIO fail-closed entrypoint guard, not the retired :?required syntax

ishiland:
## What

One-bullet fix in AGENTS.md's security pre-commit checklist (Rule 3, "Two distinct enforcement layers"): the second enforcement layer for MinIO credentials is described as the Docker Compose `MINIO_ROOT_USER:?required` required-variable syntax. That syntax no longer exists in `docker-compose.yml`.

## Why

The compose file replaced it (INST-01 + Codex P1) because `${VAR:?}` aborts `compose config`/`up` at parse time even when the cloud-dev profile is inactive, breaking a verbatim-`.env.example` install. The shipped mechanism is parse-safe `${MINIO_ROOT_USER:-}` references plus a fail-closed minio entrypoint that refuses to start on blank credentials (MinIO treats an empty `MINIO_ROOT_USER` as unset and would otherwise silently boot as `minioadmin:minioadmin`).

Surfaced by a doc-drift audit: a checklist that greps for `:?` in `docker-compose.yml` per the current AGENTS.md text always comes back empty, false-failing the Rule 3 compose layer.

## Impact

Docs only — no schema/API/config changes.

## Verification

- `grep -c ':?' docker-compose.yml` → 0 (syntax is gone; only the comment explaining its removal mentions it)
- `grep -n 'refusing to start with default/blank credentials' docker-compose.yml` → hits the minio entrypoint guard the new text describes
- Pre-commit hooks pass on the commit (secrets scan included)


ishiland:
@codex review

PR #447, opened 2026-07-10
fix(security): close post-merge review gaps

ishiland:
## What

- Reject recursive CTEs anywhere in the parsed SQL tree, including nested `WITH RECURSIVE` clauses behind a harmless outer `WITH`.
- Preserve explicit `null` values for OAuth endpoint fields during merge-mode config imports so discovery and explicit endpoint modes can switch atomically.
- Add regressions for nested recursion, both OAuth mode-switch directions, required secret rotation, and unrelated-null merge semantics.

## Why

A Codex review triggered when #445 was marked ready landed after that PR had already merged and identified two actionable P2 gaps:

- [Reject every recursive CTE](https://github.com/geolens-io/geolens/pull/445#discussion_r3562510748)
- [Keep null endpoint clears in merge imports](https://github.com/geolens-io/geolens/pull/445#discussion_r3562510752)

## Impact

AI-generated SQL now fails closed on every recursive CTE node. Config imports can apply valid OAuth endpoint-mode transitions while credential destination changes still require a replacement secret.

## Root cause

The sandbox inspected only the first `exp.With` node. Merge-mode config import also treated every `None` value as “leave unchanged,” even though endpoint nulls are explicit mode-clearing values in exported configurations.

## Checks

- Confirmed both new regressions failed before the fixes and passed afterward.
- 183 sandbox/config/OAuth regression tests passed.
- 111 focused sandbox and OAuth destination tests passed.
- Ruff check and format checks passed.
- Pre-commit hooks and `git diff --check` passed.
- Independent review found no blocking issues.

### DIFF
--- a/backend/app/platform/config_ops/service.py
+++ b/backend/app/platform/config_ops/service.py
@@ -70,6 +70,10 @@ _PROVIDER_COMPARE_FIELDS = [
     "enabled",
 ]
 
+_OAUTH_ENDPOINT_FIELDS = frozenset(
+    {"discovery_url", "authorize_url", "token_url", "userinfo_url"}
+)
+
 
 def _diff_provider(existing_dict: dict, imported: dict) -> list[str]:
     """Return list of fields that differ between existing and imported provider."""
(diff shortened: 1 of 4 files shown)


ishiland:
@codex review

Issue #448, opened 2026-07-11
Perf profile 2026-07-10: remediation of audit findings


ishiland:
Tracking issue for the 2026-07-10 performance profile remediation (internal report: perf-profile-20260710).

Key findings being fixed:
- P0: staging-swap re-ingest silently loses the geom_4326 GIST index on second and later re-ingests (CREATE INDEX IF NOT EXISTS name collision with the previous swap's index) — meteorite hero map and manhattan_building_heights_3 currently full-scan per tile.
- P0: prod compose memory caps sum to 9GB with no parametrization; demo VM has no swap (blocks D2s_v3 downsize).
- P1: catalog.record_embeddings.embedding is unbounded `vector` (atttypmod=-1) so the intended HNSW index cannot be built; semantic search is O(N) forever.
- P1: query-embedding call in the search hot path carries a 130s timeout.
- P1: worker runs with Procrastinate concurrency=1, unconfigurable (head-of-line blocking).
- P1: SQL-gen prompt defeats Anthropic prompt caching (static reference + schema + question all in system prompt; question double-billed).
- P1: chroma-js (color-vendor) leaks into the frontend entry graph via layer-icons/LegendEntries -> color-ramps.
- P2: embedding backfill unbatched; RRF exact cosine COUNT unbounded at scale; non-streaming complete() loops lack token/deadline guards; COG temps land on the 512MB /tmp tmpfs; map-vendor chunk not prefetched on map routes.

ishiland:
Remediation merged in #449 and shipped in **v1.4.5**. Closing.

PR #449, opened 2026-07-11
perf: close the 2026-07-10 performance-profile audit findings

ishiland:
Closes the in-repo action items from the 2026-07-10 performance profile (internal report `perf-profile-20260710`). Tracking issue: #448.

## What changed, by audit item

| # | Item | Change |
|---|------|--------|
| 1 | Missing GIST indexes | **Applied live** (not in this diff): `CREATE INDEX CONCURRENTLY` + `ANALYZE` on dev (`meteorite_landings_meteoritical_society`, `manhattan_building_heights_3` — EXPLAIN now Index Scan, 25 buffers vs 1,136) **and demo** (`recent_earthquakes_heatmap_source`, `recent_earthquakes_m4_5_last_30_days`); both DBs re-audited clean |
| 2 | Staging-swap index regression | Root cause found: `CREATE INDEX IF NOT EXISTS idx_<staging>_geom_4326` collides **by name, schema-wide** with the previous swap's index, so the 2nd+ re-ingest silently loses the index when the old table is dropped. `add_4326_column` now does a per-table `pg_indexes` check + unnamed `CREATE INDEX`; `_apply_reupload_swap` self-heals post-swap. Regression tests added |
| 3 | D2s downsize bundle | `mem_limit`s are env-overridable in both composes (fixed caps summed 9GB > 8GB hosts); `frontend`/`backup` gain caps. Demo-side values, swap, and `UVICORN_WORKERS=1` are encoded in the private demo provisioning recipe + runbook (local `infra/`, excluded from the public repo by design) |
| 4 | Untyped embedding column blocks HNSW | Alembic `0012`: types `embedding` to `vector(N)` (N from stored data → `embedding_dims` config → default 1536) and builds `ix_record_embeddings_hnsw` (`m=16, ef_construction=64`). Applied to the dev DB: 36/36 embeddings preserved, index built. Handles the `{"v": ...}`-wrapped app_settings shape and takes the exclusive lock up front |
| 5 | 130s embed timeout in search hot path | `asyncio.wait_for(..., 8s)` around the query-embedding call (existing silent FTS fallback catches it); no port-signature change so enterprise overlays stay compatible |
| 7 | Worker concurrency = 1, unconfigurable | `WORKER_CONCURRENCY` + `WORKER_QUEUES` settings → `run_worker_async(concurrency=..., queues=...)`; a second worker service can pin `WORKER_QUEUES=raster` |
| 8 | SQL-gen prompt defeats caching | Split: static ~3K-token PostGIS reference → system prompt (stable, `cache_control`-cacheable); schema + question → user message (question no longer sent twice) |
| 9 | color-vendor leaks into entry graph | `color-ramps.ts` drops chroma-js: static brewer stops + local interpolation, **bit-exact** with `chroma.scale().colors()` (pinned by a 27-ramp × 14-count parity test so legend swatches keep matching colors stored in saved maps). `color-vendor` no longer referenced from `dist/index.html`; it loads only with builder chunks |
| 10 | Unbatched embedding backfill | `generate_embeddings_batch()` (single-text path delegates to it); backfill embeds 128 texts per provider call |
| 12 | Unbounded exact cosine COUNT in RRF | Exact count gated to ≤5,000 embeddings (cheap row-count probe); past that, approximated from the vetted top-k ranks (lower bound; pagination `next` links survive) |
| 13 | No guards on non-streaming complete() | Both provider loops get the streaming path's PERF-009 wall-clock (90s) + cumulative-token (200K) guards; unit test asserts the abort |
| 14 | COG temps on 512MB tmpfs | Temp copies (`cog.py`) **and** the conversion output dir (`tasks_raster.py` — whose disk-space check was measuring the wrong filesystem) land on the staging volume |
| 15 | map-vendor serialized behind route chunk | Map routes `Promise.all` the page chunk with `import('maplibre-gl')` so the 306KB gz vendor downloads in parallel |

## Deliberately not in this PR
- **#6 (remote Sentinel-2 COGs, 0.9–2s/tile):** content decision — localize the 9 remote assets in-region vs de-feature from hero maps; VM-size-independent. Options + steps documented in the (private) demo runbook.
- **#11 (`max_parallel_workers_per_gather=0`):** demo-DB `ALTER SYSTEM` at downsize time; in the downsize runbook.
- **#16–18 (P3: FTS OR→UNION, generated geom_4326, raster proxy ETag/cache) + single-pass `-of COG`:** the audit's own priority matrix marks these skip-for-now/deferred.

## Schema/API/config impacts
- New Alembic revision `0012_type_embedding_vector` (idempotent; skips already-typed columns; deletes only dimension-mismatched embedding rows, which are regenerable cache).
- New env knobs (all default to previous behavior): `WORKER_CONCURRENCY`, `WORKER_QUEUES`, `DB_MEM_LIMIT`, `API_MEM_LIMIT`, `WORKER_MEM_LIMIT`, `TITILER_MEM_LIMIT`, `FRONTEND_MEM_LIMIT`, `BACKUP_MEM_LIMIT`. (`.env.example` couldn't be edited in this session — follow-up if wanted.)
- No OpenAPI surface changes.

## Verification
- Backend: full suite `pytest -n 4` → 4,841 passed (1 unrelated auth flake, passes in isolation); `ruff check` + `format` clean; `make alembic-check` clean; migration round-tripped (down/up) on the test DB and applied to dev + test DBs.
- Frontend: `typecheck`, `lint`, full `test:coverage` (3,476 passed incl. new chroma-parity suite), production build — `color-vendor` absent from `dist/index.html`, map-vendor still isolated.
- Live: dev EXPLAIN flipped to Index Scan on the meteorite table; demo DB re-audit returns zero unindexed tables; https://demo.getgeolens.com returns 200.
- Compose: `docker compose config` clean for both files; pre-commit security hooks pass.

https://claude.ai/code/session_01171DzVEr4GLkrygD6fuc13

### DIFF
--- a/backend/alembic/versions/0012_type_embedding_vector.py
+++ b/backend/alembic/versions/0012_type_embedding_vector.py
@@ -93,10 +93,23 @@ def upgrade() -> None:
           END IF;
 
           -- Same DDL the baseline and rebuild_embedding_column intend.
-          EXECUTE 'CREATE INDEX IF NOT EXISTS ix_record_embeddings_hnsw '
-                  'ON catalog.record_embeddings USING hnsw '
-                  '(embedding vector_cosine_ops) '
-                  'WITH (m = 16, ef_construction = 64)';
+          -- fix(#449, codex P1): pgvector rejects HNSW on vector columns over
+          -- 2000 dims (3072-dim models are legal config, cap is 4096) — skip
+          -- the index rather than fail the upgrade; semantic search falls
+          -- back to exact scans. Re-probe: the ALTER above may have typed it.
+          SELECT atttypmod INTO cur_typmod
+          FROM pg_attribute
+          WHERE attrelid = 'catalog.record_embeddings'::regclass
+            AND attname = 'embedding';
+
+          IF cur_typmod BETWEEN 1 AND 2000 THEN
+            EXECUTE 'CREATE INDEX IF NOT EXISTS ix_record_embeddings_hnsw '
+                    'ON catalog.record_embeddings USING hnsw '
+                    '(embedding vector_cosine_ops) '
+                    'WITH (m = 16, ef_construction = 64)';
+          ELSE
+            RAISE NOTICE 'Skipping ix_record_embeddings_hnsw: % dims exceeds pgvector''s 2000-dim HNSW limit; semantic search stays on exact scans.', cur_typmod;
+          END IF;
         END $$;
         """
     )
--- a/backend/alembic/versions/0012_type_embedding_vector.py
+++ b/backend/alembic/versions/0012_type_embedding_vector.py
@@ -8,17 +8,24 @@ UI, so a deployment running on defaults never gets a typed column — and an
 HNSW/ivfflat index can never be built (pgvector requires a typed column).
 Semantic search then does an O(N) x dims cosine scan per query, twice.
 
-This migration types the column in place, preferring (in order) the
-dimension of vectors already stored (only when they all agree — mixed
-dimensions mean a model switch, where the active config must win), the
-``embedding_dims`` persistent
-config override (skipped when ``ENV_ONLY_CONFIG`` is set, mirroring
-runtime resolution in persistent_config.py), and finally the configured
-``EMBEDDING_DIMS`` env value / Settings default (1536,
-text-embedding-3-small). Rows whose dimension disagrees with the chosen
-value are deleted — embeddings are a regenerable cache (content-hash
-backfill restores them). Deployments whose column is already typed (the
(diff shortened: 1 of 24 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #451, opened 2026-07-11
fix(builder): composable hillshade + terrain, and stop partial patches erasing style metadata

ishiland:
## Summary

Addresses every finding in `docs-internal/bugs/hillshade-terrain-render-options-root-cause.md`. The root cause was a **partially completed state-model migration**: the renderer, viewer, and style exporter already treated a DEM as a composable source of *both* a visible hillshade/color-relief overlay *and* an additive 3D terrain mesh, but the editor, stack, legend, and persistence layers still assumed the two were mutually exclusive. This PR completes the composable model and fixes the release-blocking `MapLayerPatch` data-integrity bug independently.

The Matterhorn demo map (`62727875-…`) persists the exact hybrid trigger state; no app code assumed it was corrupt data, and the fix converges on the composable model the codebase already favored.

## Data-integrity fix (release-blocking, independent of the UI work)

- **HT-14** — `MapLayerPatch._normalize_paint_boundary` was an `after`-validator that assigned `paint`/`style_config` unconditionally, adding both to `model_fields_set` and defeating the router's `exclude_unset=True`. Any partial patch — including the builder's own eye-toggle Save, which sends exactly `{id, visible}` — carried an explicit `style_config=None` that wiped DEM hypsometric metadata (`hypso_enabled`/`hypso_ramp`), flipped `render_mode:"terrain"` DEMs back to hillshade, and falsified history `changed_fields`. Reworked to `before`-mode so omitted fields stay omitted; `style_config` is materialized only when the client sent it or the legacy paint split produced builder keys. Covered by a unit test on `model_fields_set` and a DB-backed end-to-end test proving a `{id, visible}` PATCH preserves DEM style metadata.
- **HT-15** — `TerrainConfig` now enforces `enabled ⇒ source_dataset_id`. Because the same model validates stored JSONB on the read path (`MapResponse`), it **coerces** an inconsistent value to disabled rather than raising, so a legacy/corrupt row self-heals instead of 500-ing on read.

## Composable state model (UI + persistence)

| Finding | Fix |
|---|---|
| **HT-01/02/06** | DEM editor replaces the exclusive Hillshade/Terrain radio with two independent switches — **Hillshade shading** (`style_config.render_mode`) and **Use as 3D terrain** (`terrain_config`) — with the terrain exaggeration slider under the terrain switch. Each switch writes only its own authority, so a round trip of one never mutates the other. |
| **HT-03/11** | The stack no longer suppresses `render_mode:"terrain"` DEM rows — they stay visible, selectable, and editable, so the DEM is always reachable and rebinding can't orphan hidden layers. (The map still paints nothing for them; render suppression is unchanged.) |
| **HT-04** | Settings terrain panel gains an **Open terrain layer** action that opens the bound DEM's editor. |
| **HT-05** | Deleted the false "source conflict" advisory note and the unreachable `shouldSkipHillshadeForTerrain` tile-size guard + `terrainSourceTileSize` plumbing (same-dataset binding ⇒ tile sizes always match, so the skip branch was dead code). |
| **HT-07** | Opacity/zoom controls render only while the overlay exists (the mesh ignores both); raster/hillshade layers now reapply saved `_minzoom/_maxzoom` on sync (previously only vector paths called `syncLayerZoomRange`). |
| **HT-08** | The synthetic terrain legend entry (builder + viewer) keeps the bound DEM's display name instead of degrading to a generic "3D terrain" row. |
| **HT-10** | `resolveTerrainSourceLayer` has deterministic duplicate semantics (prefer a visible rendering) and is shared by builder, viewer, and embeds. |
| **HT-12** | `useViewerTerrain` honors the bound DEM's saved visibility the same way the builder does, so builder and viewer (and embeds) no longer disagree. |
| **HT-13** | A save that touches both a layer diff and the map-level terrain binding persists through the single full-replacement PUT (reconciled by id, per #430 V-14) instead of two sequential requests that could half-commit. |
| **HT-18** | Editor-close focus restoration falls back to the first stack row when the originating row was removed. |
| **HT-19** | `StyleConfig`'s data-driven `mode`/`column`/`ramp` fields are optional so DEM-only configs typecheck without `as unknown as` casts. |

## Documented deferrals (not fixed here)

These three are noted in the report as investigations rather than UI defects, and per the root-cause analysis they aren't quick fixes:

- **HT-09** (color-relief no-data footprint) — can't distinguish no-data from sea level client-side; needs backend tile work. Left as a separate investigation.
- **HT-16** (duplicate DEM source decode) — performance/lifecycle; the exported-style single-source model is the eventual target but it's out of scope for this correctness pass.
- **HT-17** (`elevation maxzoom > source.maxzoom` console warning) — diagnostic only; no functional impact.

## Test coverage

The report noted several unit tests that *locked in* individual pieces of the bug. Those were rewritten to the composable model, and new coverage was added:

- DEM editor: two-switch model + a **Matterhorn hybrid-state regression** (overlay + terrain on one DEM) proving the terrain round trip never rewrites `render_mode`.
- Stack/legend: terrain-mode DEM rows render; a hillshade-bound DEM reads as an active terrain source; the synthetic legend entry carries the DEM name.
- `map-sync`: dropped the dead `shouldSkipHillshadeForTerrain` suite; assert raster/hillshade zoom-range reapplication.
- Save: combined layer + terrain change persists via one PUT.
- Backend: unit test on `model_fields_set` + DB-backed `{id, visible}`-preserves-style_config regression.

## Verification

- Frontend: `npm run typecheck`, `eslint` (changed files), `npm run test:i18n`, and the full `builder`/`viewer`/`map-plugins`/`pages` suites — **1992 passed**. New `demEditor` i18n keys added to all four locales (en/es/fr/de); removed keys pruned.
- Backend: `ruff check`/`format`, targeted `pytest` for schemas, style-json, and DB-backed maps terrain/patch tests — all green.

@codex review

https://claude.ai/code/session_019Km6yR58rC79qbXYXyJEzy


ishiland:
@codex review

ishiland:
Addressed all three Codex P2 findings in f2ac574e:

- **use-viewer-terrain / ViewerMap reveal gate** — the reveal condition now computes the same effective terrain expectation the hook uses (`resolveTerrainSourceLayer` exists AND `visible !== false`), so a terrain map whose bound DEM is saved-hidden reveals immediately instead of waiting on the 4s safety timer.
- **map-stack terrain status** — a hidden bound DEM now reports `sourceStatus: 'disabled'` (matching BuilderMap's no-mesh path and Settings' `isTerrainActive`), and the source resolves through `resolveTerrainSourceLayer` so a duplicate rendering can't flip the status.
- **terrain-legend** — `deriveTerrainLegendEntry` resolves the backing DEM the same way and returns null when it's saved-hidden, so no phantom terrain legend entry lists a hidden layer's name.

Added regression tests for the hidden-source cases (terrain-legend + map-stack) and re-ran the full builder/viewer/plugin suite (1900 passing) plus typecheck/lint.

@codex review

ishiland:
@codex review

ishiland:
## Self-review (in lieu of Codex — quota-blocked)

Ran an adversarial multi-dimension review of the full source diff (correctness / regression / edge-cases), with every candidate finding independently verified before acting. Three were confirmed real; all fixed in a17bb3f8. (Two candidate findings were verified and **refuted** — the HT-12 viewer-visibility change is the intended fix for a pre-existing builder↔viewer divergence, not a new defect.)

**1. HT-13 was over-engineered — reverted.** The composable model this PR introduces makes the hillshade overlay (`style_config.render_mode`) and the 3D terrain binding (`terrain_config`) *independent* authorities. So a partial save (layer PATCH commits, metadata PUT fails) is no longer a contradiction — it's an incomplete save of two independent settings, coherent on reload and recoverable by re-toggling. Routing every combined save through the full-replacement PUT added a real risk (it re-serializes every layer via `toLayerInput`, which can null server-only layer fields the session never loaded — the #430 V-01 class) with no re-sync warning. Dropped it; the normal split PATCH+PUT path is safe here and the existing failed-save toast prompts a retry.

**2. HT-07 zoom fix was too broad — scoped down.** The raster/hillshade `syncLayerZoomRange` call applied an unconditional `(0, 22)` fallback, so a plain raster/imagery layer (which has no zoom editor and never carries `_minzoom/_maxzoom`) got force-hidden at the max zoom stop. Now it only reapplies a range when a saved one is actually present, leaving maplibre's default untouched otherwise.

**3. New 'ghost DEM' state — added an honest badge.** Decoupling the overlay toggle from terrain binding made a new state reachable: a DEM with its overlay off (`render_mode: 'terrain'`) that isn't the active terrain source paints nothing, yet showed a normal eye-on stack row. Added a 'Not shown — enable Hillshade or 3D terrain' badge (mirroring the existing audience-hidden badge pattern) so the row stays honest. The in-editor hint already covered the moment of creation; this covers the closed-editor row.

Added regression tests for all three (raster no-custom-range → no zoom cap; split-save path with terrain change; draws-nothing badge). Full builder/viewer/plugin/pages suite green (**1997 passing**), plus typecheck, lint, and i18n parity (new `stackRow.drawsNothing` key in all four locales).

ishiland:
Second review pass (all 7 CI-green commits) came back READY WITH NITS — the three earlier codex P2s (reveal gate, stack status, synthetic legend entry) are confirmed fixed. The four remaining P2 findings are addressed in 99975166:

- Dropped the obsolete D-04 duplicate-rendering guard (+ its toast key in all four locales) — terrain is a single map-level pointer now, duplicates can't accumulate it.
- Viewer terrain now stays off when the payload has rows for the bound dataset but none is a visible terrain-capable DEM (legacy/corrupt binding), while keeping token-seeding for private-DEM share-token embeds where the row is filtered out entirely.
- Extracted the reveal-gate predicate as `isViewerTerrainExpected` with pinning unit tests (a regression there was an unobserved 4s veil).
- Corrected the stale TEST-ONLY header in map-stack.ts (only `buildMapStack` is test-only).
- New e2e spec `builder-dem-editor.spec.ts` + single-band Float32 GeoTIFF fixture: two-switch composability, the drawsNothing badge, terrain_config binding, and the HT-14 visibility-only-save guarantee. Passes locally against the dev stack.

PR #452, opened 2026-07-11
fix(viewer): live DEM eye-toggle parity — hide the tint, disable terrain, align legend icons

ishiland:
## What

Two viewer/legend fixes found while smoke-testing #451 on the live demo. **Stacked on #451** (base = `fix/hillshade-terrain-composable`); GitHub will retarget to `main` when #451 merges — the delta here is only the two commits above #451's head.

### 1. Viewer live DEM eye-toggle left the tint painting and the mesh extruded (pre-existing, not a #451 regression)

Hiding the bound DEM via the **viewer** legend's eye only hid the hillshade MapLibre layer:

- the `-colorrelief` companion kept painting an unshaded hypsometric tint (`hillshadeAdapter.syncVisibility` never touched it; the full sync path that recreates it doesn't run on visibility-only diffs),
- the 3D terrain mesh stayed extruded (`useViewerTerrain` reads the saved `visible` flag, never the live `visibleLayers` override), and
- a phantom standalone "3D terrain" legend entry appeared for the just-hidden source.

Fix: the adapter now toggles (and declares via `getLayerIds`) the color-relief companion; `useViewerTerrain` takes the bound DEM's live visibility and clears terrain when it's false (mirroring BuilderMap's `effectiveTerrainEnabled`); the viewer legend drops the synthetic terrain entry for a live-hidden DEM (the DEM's own row remains as the re-show affordance). The `deriveCompanionIds` hillshade special-case is gone — the adapter owns its companion now.

### 2. Legend icons didn't match the layer-stack icons

The stack showed the DEM glyph chip (⛰/◬/▦) while the legends rendered unrelated lucide icons (Grid3x3/Layers rows, Mountain for the synthetic terrain entry). StackRow's TypeIcon logic is now a shared `LayerTypeIcon` in `layer-icons.tsx`, rendered by StackRow, the builder LegendPlugin, and the viewer LayerLegend; the synthetic 3D-terrain entries use the stack's terrain-mode ◬ chip.

## Behavior change / inverted pin

`LayerLegend.test.tsx` had a pin that the synthetic terrain entry **stays** when the DEM is live-toggled off — correct back when the mesh kept rendering, a phantom now that it doesn't. The pin is inverted with a comment. `StackRow.guard04.test.tsx` is removed: its spy seam (cross-module `extractStyleHints` call) no longer exists after consolidation; the `useMemo` it pinned moved verbatim into `LayerTypeIcon`.

## Verification

- `npm run typecheck`, `npm run lint`, full vitest suite **3503 passing** (new regression tests for the adapter companion visibility, the live-toggle terrain gate, the phantom legend entry, and glyph parity).
- Fresh local stack (`down -v`, reseeded showcase incl. Matterhorn terrain + Sentinel-2 imagery), Playwright smoke:
  - Viewer live eye-toggle now flattens the mesh, removes the tint, and shows no phantom entry; re-show restores everything.
  - Builder DEM permutations (terrain-only / shading-only / tint-off), ramp change, eye-toggle ⇄ Settings "No terrain layer is active" ⇄ legend, save→reload round-trips (visibility-only save changes ONLY the `visible` flag — HT-14 stays fixed), zoom-range editor, "Not shown" ghost-DEM badge.
  - Sentinel-2 raster editor: stretch Min/Max→Percentile→Min/Max with save→reload persistence, zero failed tile requests (no #317 regression), eye toggles, ▦ legend parity, sibling layers' `style_config` untouched.
  - Zero application console errors across the session.

## Screenshots

Legend/stack icon parity and the fixed viewer hide state are visible in the local smoke captures (available on request — not committed).


ishiland:
Review came back READY WITH NITS (typecheck + ~1,800 targeted tests run locally since no CI fires on a non-main base). The three P2s are addressed in 0a2566e9, and the branch is rebased onto the updated #451 head:

- `isTerrainBackingLiveVisible` in layer-identity is now the ONE definition of the bound DEM's live visibility, shared by the mesh gate (ViewerMap) and the synthetic legend entry (LayerLegend).
- Both synthetic terrain entries derive their chip via `demChipGlyph('terrain')` instead of hardcoding ◬.
- The deleted guard04 memoization coverage is replaced by a spy-free test in layer-icons.test.tsx that counts paint/layout property reads — including the null-paint fallback-object trap — and was verified to fail under that exact mutation.

PR #461, opened 2026-07-12
fix(builder): stop the style editor clobbering saved categorical symbology

ishiland:
## Summary

Two builder bugs, surfaced by the **Hurricane Alley** showcase map, that silently destroy hand-authored **categorical** symbology. The map has a categorical line layer whose saved `style_config.categories` lists fewer classes (6) than the data has distinct values (7 — the data includes a `TD` segment the config never listed).

### 1. Opening the style editor clobbered the saved symbology
Merely selecting the layer regenerated every per-category color from the ramp, dropped the category labels, and marked the map dirty — with **no user gesture**. The loop-skip guard (`styleConfigAlreadyMatches`) required the saved categories to equal the data's distinct values *exactly in count and order*, so any categorical layer with an unlisted value regenerated on mount.

**Fix:** relax the categorical guard to skip regeneration when every styled category still exists in the data (a subset), even if the data has additional values. Unlisted values keep falling through to the paint expression's own fallback, exactly as before the editor opened. A styled category that no longer exists in the data still counts as drift and regenerates.

### 2. The "Pending style preview" banner's "Reset" applied library defaults
The banner promises to reflect the layer *"before save"*, but its action flattened hand-authored expressions (e.g. the wind-speed `line-width` step expression) to defaults.

**Fix:** rename it **Revert** and restore the server baseline (paint/layout/style_config/opacity) instead. The data-driven editor is remounted on revert so its local ramp/mode/column re-seed from the restored config rather than immediately re-applying the just-discarded selection. The section-header **Reset** keeps its distinct reset-to-defaults behavior.

## Impact / scope
- Frontend-only; no schema, API, or config changes.
- Affects **every instance** — any user's hand-styled categorical map was one editor-open away from silent loss.
- New i18n keys `style.preview.revert` / `style.preview.revertTitle` added to all four locales (es/fr/de machine-authored — flagged for the batched native-speaker pass).

## Verification
- `npm run typecheck`, `npm run lint`, `npm run test:i18n` — green.
- `npx vitest run src/components/builder` — 1802 tests pass, incl. new guard subset-skip cases and split Revert/Reset tests.
- **Live** against the seeded map (standalone Vite → same DB), console watched throughout (0 errors / 0 warnings): opening the editor preserves authored colors, 0.765 opacity and the `wind_kt` width expression and stays *Saved*; a genuine ramp change still regenerates; **Revert** restores the saved style with no re-clobber and clears the banner.

## Notes
- Not included here: the Hurricane map's own seed-content polish (ramp mislabel, missing `TD` legend entry, double-dimmed opacity, arrow contrast, label density, duplicate popups) — that's demo content and will be a separate PR gated on a demo re-seed.
- After Revert the map-level "Unsaved changes" flag stays set (any layer mutation flips it); the per-layer banner correctly reads clean. Pre-existing behavior, out of scope here.

https://claude.ai/code/session_01QmxibW97dSfvC1m4S6xLXU

### DIFF
--- a/frontend/src/components/builder/DataDrivenStyleEditor.tsx
+++ b/frontend/src/components/builder/DataDrivenStyleEditor.tsx
@@ -223,14 +223,25 @@ export function styleConfigAlreadyMatches(p: StyleGuardParams): boolean {
     // through to the paint expression's own fallback, exactly as before the
     // editor opened. A stale category value no longer present in the data still
     // fails the check, so a genuinely drifted config regenerates as before.
+    //
+    // fix(#461, codex P2): when the column has NO distinct values the effect
+    // writes `categories: []`; that empty config must count as a match or the
+    // effect re-writes every render (an empty dataset / all-null column loops).
+    // So for empty data, match iff the saved categories are also empty; for
+    // non-empty data, require a non-empty subset (empty config still regenerates
+    // to style the values).
+    const categoriesSettled =
+      values.length === 0
+        ? ec.categories?.length === 0
+        : (ec.categories?.length ?? 0) > 0 &&
+          ec.categories!.every((c) => dataValues.has(c.value));
     return Boolean(
       ec.mode === 'categorical' &&
       ec.column === p.column &&
       ec.ramp === p.ramp &&
       (ec.reversed ?? false) === p.reversed &&
       ec.categories &&
-      ec.categories.length > 0 &&
-      ec.categories.every((c) => dataValues.has(c.value)),
+      categoriesSettled,
     );
   }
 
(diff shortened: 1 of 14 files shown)


ishiland:
@codex review

Addressed both P2 findings in c280d928:
- **Empty categorical loop**: the guard now special-cases empty data (matches an empty saved config; non-empty data keeps the subset rule).
- **Builder-only edits surviving Revert**: added an opt-in `{ replace }` flag threaded through the `set_style_config` action so Revert restores the saved config verbatim, bypassing the builder-merge.

Please re-review the current head.

PR #462, opened 2026-07-12
feat(catalog): launch-window catalog easy wins — GeoParquet, DCAT paging, ingest AI metadata, llms.txt

ishiland:
Four small, high-leverage wins from the 2026-07-11 catalog-enhancement research (`docs-internal/GTM/catalog-enhancement-research-2026-07-11/`), scoped to ship before the outreach wave without opening any long-lived project. Each is deliberately small; the compounding-but-heavy Tier-1 items (scheduled harvesting, white-label portal, subscriptions) are intentionally **not** here.

## What's in

**1. GeoParquet export** — the marquee differentiator ("your catalog is directly queryable").
- The Debian GDAL build ships **no Arrow/Parquet driver** (verified: `gdal-plugins` doesn't include it at any version), so ogr2ogr can't emit Parquet. New `processing/export/parquet.py` writes a spec-valid **GeoParquet 1.1** file directly via `pyarrow` — WKB geometry + `geo` file metadata, **EPSG:4326 / OGC:CRS84**.
- Wired into `GET /datasets/{id}/export?format=parquet`; rejects a non-4326 `target_crs` (no reprojection on that path). Non-spatial datasets are blocked like the other spatial formats.
- Dataset UI adds the **GeoParquet** format plus a copy-paste **DuckDB snippet** so the download is queryable, not just archived.
- New dependency: `pyarrow` (pure wheel, ~40 MB; no system libs, no base-image change). Chosen over swapping to `ghcr.io/osgeo/gdal` (large blast radius) — see PR discussion.

**2. DCAT feed paging** — `dcat`, `dcat-us/3.0`, `geodcat-ap` catalog feeds gain `limit`/`offset` so a catalog larger than 10k datasets (e.g. a federal `data.json` harvester) can crawl the whole feed instead of silently truncating. Defaults preserve the current single-document behavior and the BA-28 memory guard.

**3. Ingest AI metadata** — the bulk-import completion card now surfaces the **existing** AI metadata drafter (gated on `useAIAvailability`) the moment a dataset lands. Placement, not new capability.

**4. `/llms.txt`** — static file pointing agents at this instance's live discovery, search, and OGC/STAC/DCAT endpoints.

## API / schema impact
- New export format enum value `parquet`; new `limit`/`offset` query params on three DCAT feeds.
- Regenerated `backend/openapi.json` + Python/TS SDKs (drift-gate clean; diff is limited to these additions).

## Verification
- **GeoParquet writer** (DB-free unit test + a manual round-trip): `geo` metadata, WKB decode, null-geometry, and mixed-type→string fallback all pass.
- Added `test_export_parquet.py` (DB-backed endpoint test) for CI.
- Frontend: `typecheck`, `lint`, `test:i18n` (all four locales), and the ExportButton/BulkTrackingList component tests all green.
- Backend `ruff check` + `format` clean; pre-commit hooks (SSRF/visibility/secrets) passed.

## Notes for review
- GeoParquet is written in EPSG:4326 only (CRS84) — the lakehouse/DuckDB convention; a projected `target_crs` is rejected rather than silently ignored.
- The parquet writer builds the selection in memory before writing (bounded by the existing 5M-feature export cap); marked with a `ponytail:` upgrade path to a batched `ParquetWriter` if large-catalog exports OOM.

https://claude.ai/code/session_015cpiMW7X8MNmG4nN4dfzqB

### DIFF
--- a/backend/app/modules/catalog/datasets/api/router_export.py
+++ b/backend/app/modules/catalog/datasets/api/router_export.py
@@ -8,6 +8,7 @@ from fastapi import (
     APIRouter,
     Depends,
     HTTPException,
+    Query,
     Request,
     Response,
     status,
@@ -72,21 +73,32 @@ def _dcat_relationship_options():
 
 
 # fix(#430 BA-28): these anonymous feeds materialize every visible dataset (+ keywords/
-# contacts/distributions) into one in-memory JSON-LD doc with no pagination or
-# cache — a cheap repeatable memory/CPU amplifier on a large catalog. Bound the
-# row count (OGC/STAC cap at ≤200) and log when the bound truncates the feed.
+# contacts/distributions) into one in-memory JSON-LD doc with no cache — a cheap
+# repeatable memory/CPU amplifier on a large catalog. A single page is bounded at
+# _DCAT_FEED_MAX_DATASETS; the catalog #catalog-enhancement fix adds limit/offset
+# so a large-catalog operator (e.g. a federal data.json harvester) can crawl the
+# whole feed instead of silently losing everything past the first 10k.
+#
+# ponytail: offset paging keeps the memory guard while unblocking >10k catalogs;
+# a spec-correct single-document streaming data.json is the real fix if a
+# harvester that can't page ever needs >10k in one request.
 _DCAT_FEED_MAX_DATASETS = 10_000
 
 
 async def _get_visible_dcat_datasets(
-    db: AsyncSession, user: Identity | None
+    db: AsyncSession,
+    user: Identity | None,
+    *,
+    limit: int = _DCAT_FEED_MAX_DATASETS,
+    offset: int = 0,
 ) -> list[DatasetModel]:
     stmt = (
         select(DatasetModel)
         .join(Record, DatasetModel.record_id == Record.id)
         .options(_dcat_relationship_options())
         .order_by(Record.created_at.desc(), Record.id.desc())
-        .limit(_DCAT_FEED_MAX_DATASETS)
+        .offset(offset)
+        .limit(limit)
     )
 
     if user is not None:
@@ -98,13 +110,26 @@ async def _get_visible_dcat_datasets(
 
     result = await db.execute(stmt)
     datasets = list(result.unique().scalars().all())
-    if len(datasets) >= _DCAT_FEED_MAX_DATASETS:
+    if len(datasets) >= limit:
         logger.warning(
             "dcat_feed_truncated",
-            limit=_DCAT_FEED_MAX_DATASETS,
+            limit=limit,
+            offset=offset,
             authenticated=user is not None,
         )
     return datasets
+
+
+# Shared query params for the paginated catalog feed handlers.
+_FEED_LIMIT_Q = Query(
+    _DCAT_FEED_MAX_DATASETS,
+    ge=1,
+    le=_DCAT_FEED_MAX_DATASETS,
+    description="Max datasets in this page (default = max).",
+)
+_FEED_OFFSET_Q = Query(
+    0, ge=0, description="Datasets to skip — page a catalog larger than one page."
+)
 
 
 async def _get_dcat_dataset_for_export(
@@ -138,9 +163,11 @@ async def get_dcat_catalog(
     request: Request,
     user: Identity | None = Depends(get_optional_user),
     db: AsyncSession = Depends(get_db),
+    limit: int = _FEED_LIMIT_Q,
+    offset: int = _FEED_OFFSET_Q,
 ) -> JSONResponse:
     """DCAT 3 JSON-LD catalog feed. Respects dataset visibility."""
-    datasets = await _get_visible_dcat_datasets(db, user)
+    datasets = await _get_visible_dcat_datasets(db, user, limit=limit, offset=offset)
 
     base_url = await get_public_api_url(db)
     catalog = catalog_to_dcat(datasets, base_url)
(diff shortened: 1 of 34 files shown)


ishiland:
@codex review

ishiland:
Addressed both P2s in 22ecd4b9:
- **bbox semantics** (`parquet.py`): now mirrors the features-query predicate — envelope `&&` prefilter + exact `ST_Intersects`, with the antimeridian split when `minx > maxx` (since `parse_bbox` allows those). Added a DB-backed exact-intersection test.
- **llms.txt STAC link**: now `/api/stac/` with the trailing slash.

@codex review

ishiland:
Addressed both new P2s in the latest commit:
- **null-geometry rows**: dropped the blanket `geom_4326 IS NOT NULL` — full exports now keep null-geom rows (null geometry cell), matching the feature read path; bbox filters still exclude them naturally. Added a DB-backed test.
- **frontend types**: regenerated `src/types/api.generated.ts` (ExportFormat now includes `parquet`, plus the feed paging params).

@codex review

ishiland:
Addressed both new P2s:
- **user `geometry` column**: `_attr_names` now skips only the true internal cols (gid/geom/geom_4326); the WKB output column renames to `geom_wkb` when a user column already owns `geometry`, and the GeoParquet `primary_column` metadata follows. DB-free writer tests cover the collision.
- **discovery feeds**: GeoParquet added to `_FORMAT_MEDIA` (STAC/OGC `formats`) and `_DISTRIBUTION_TEMPLATES` (DCAT distributions); updated the three vector count assertions. Non-spatial datasets still correctly exclude it.

Note: the *regenerate frontend types* comment on d1b44888 was a stale re-flag — `api.generated.ts` already has `parquet` (regenerated in that same commit).

@codex review

ishiland:
Round 4:
- **event-loop offload** (real): Parquet build+write now run via `run_in_thread_draining`, same as the shapefile zip path — large exports won't block the loop or job heartbeats.
- **DCAT distribution / backfill** (your #3): scoped this deliberately. GeoParquet stays advertised **live for all datasets** via OGC API Records + STAC (`_FORMAT_MEDIA`). I reverted the materialized `_DISTRIBUTION_TEMPLATES` entry — since those rows are generated per-dataset at creation, it only covered *new* datasets and needs a data-migration backfill for existing ones. Deferring template+migration together to a focused follow-up beats shipping a new-vs-old half-state.
- The *regenerate frontend types* and *advertise in feeds* comments on the last commit were stale re-flags — `api.generated.ts` already has `parquet` and `_FORMAT_MEDIA` already advertises it.

@codex review

ishiland:
Round 5 — one real bug, fixed:
- **literal colons in `where`** (your parquet.py finding): the validated clause is interpolated into a SQLAlchemy `text()`, so a colon after a non-word char inside a string literal (e.g. `name = ' :name'`) was misparsed as an unbound bind and 500'd. Now escaped to `text()`'s literal-colon form (`\\:`); the bbox `:minx/:miny` binds are separate and stay real. Proved the bind-parse fix directly + added a DB-backed regression.

The other two comments are prior items: `api.generated.ts` already has `parquet` (regenerated), and DCAT materialized-distribution advertising is the deferred follow-up (OGC API Records + STAC advertise parquet live for all datasets in the meantime).

@codex review

ishiland:
Round 6 — one real bug, fixed:
- **DuckDB snippet quoting** (`ExportButton.tsx`): a title like `Bob's Roads` produced invalid DuckDB SQL. Single quotes are now doubled (`Bob''s Roads.parquet`), which is also consistent with the saved filename (`anchor.download` uses `datasetName`). Added a component test.

The other three comments are already-resolved/scoped prior items: the literal-colon escape is in HEAD (`parquet.py:189`), `api.generated.ts` already includes `parquet`, and the DCAT materialized-distribution listing is the deferred follow-up (OGC API Records + STAC advertise parquet live for all datasets now).

@codex review

ishiland:
Round 7 — fixed the snippet path-separator edge: a title like `Roads/2026` is now sanitized (`/`,`\\` → `_`) so the DuckDB snippet matches the browser-saved filename; test covers quote + separator together.

The other three comments are the recurring resolved/deferred items (colon escape in HEAD at parquet.py:189, `api.generated.ts` has `parquet`, DCAT materialized-distribution advertising deferred with OGC/STAC covering discovery live). I believe the actionable findings are now exhausted — the remaining loop is the full-diff re-review re-surfacing the same three.

@codex review

ishiland:
Round 8 — two real bugs, fixed by one rewrite of the parquet query:
- **attribute loss with absent metadata**: `Dataset.column_info` is nullable, so a dataset without metadata exported geometry-only. The parquet path now **introspects the live table** (`get_column_info`) for attribute names — matching the table ogr2ogr actually reads.
- **type fidelity**: `to_jsonb(t.*)` stringified dates/timestamps/UUIDs/numerics. Now attribute columns are **selected directly** so the driver returns native values and Arrow infers real types (geometry selected last + read positionally to avoid alias shadowing; exotic types fall back to string).

DB-backed test covers both (column_info=[] still exports attributes; a `date` column round-trips as an Arrow date, not a string).

@codex review

ishiland:
Round 9 — two real edges on the metadata-less path, fixed:
- **filterable without metadata**: `export_parquet` now introspects the live columns once and validates the `where` against them (not the nullable `column_info`), so a metadata-less dataset is both exportable and filterable. Dropped the now-unused `column_info` param.
- **bounded in-memory build**: the router's cap is skipped when `feature_count` is NULL, so `export_parquet` runs a LIMIT-bounded COUNT and raises `ExportTooLargeError` → 413 before streaming millions of rows into memory.

DB-backed tests cover both (filter with `column_info=[]`; lowered cap → 413).

@codex review

ishiland:
Round 10 — one genuine residual, fixed (the fix simplifies): the router's step-6b cap guard validated the filter against `dataset.column_info` before `export_parquet` ran, so an oversized metadata-less dataset was unfilterable. **Parquet is now exempt from step 6b** — `export_parquet` already self-caps via a bounded count against the live-introspected columns, so the router guard was redundant *and* wrong for parquet. Regression test added (feature_count=5,000,001 + column_info=[] + filter → 200, 1 row).

The other four comments this pass are re-flags of resolved/deferred items (the cap and colon fixes are in HEAD; `api.generated.ts` has `parquet`; DCAT advertising is the deferred follow-up with OGC/STAC live).

@codex review

Issue #458, opened 2026-07-11
Editing audit 2026-07-11: P1 findings (E-01..E-07)


ishiland:
Tracking issue for the P1 findings of the 2026-07-11 editing audit (full report in docs-internal/audits/editing-audit-20260711.md).

- **E-01** Feature geometry writes hard-code SRID 4326 into \`geom\`; every geometry edit on a projected-SRID file-ingested dataset 500s (\`features/service.py\` \`_geometry_sql\`). File ingest keeps the source CRS (\`run_ogr2ogr\` passes no \`-t_srs\`); only service ingest forces 4326.
- **E-02** No \`ST_IsValid\` enforcement: self-intersecting polygons persist and 500 later reads (incl. anonymous tile/bbox); degenerate-but-schema-valid input 500s at write.
- **E-03** AttributeTable inline cell edits send every value as a string; non-text column edits always fail with a generic toast (\`AttributeTable.tsx\`; drawing \`AttributeForm\` coerces per type, table path does not).
- **E-04** Metadata fields cannot be cleared to null — dropped in both \`InlineEdit.save()\` (truthy gate) and \`_apply_simple_field_assignments\` (\`is not None\` gate) — while the UI toasts "Changes saved".
- **E-05** Column DDL (\`layers/service.py\` add/rename/type-change/drop) never invalidates the Valkey tile cache (feature writes and reupload both do).
- **E-06** Column rename/drop does not warn about or reconcile saved maps whose style_config/filters/labels reference the column.
- **E-07** E2E feature-editing coverage was removed (\`dataset-detail.spec.ts\`, "moved to vitest") and never replaced; no edit→reload persistence round-trip exists.

ishiland:
## P1 remediation complete

All 7 P1 findings are merged to main:

- **#459** (ec5b070a) — E-01 geometry SRID transform, E-03 typed cell coercion (+ bigint precision), E-05 DDL tile-cache purge, E-12 attribute-metadata revive
- **#460** (f5af57ec) — E-02 geometry validity gate, E-04 metadata clear-to-null (+ re-embed/audit on clear), E-06 saved-map dependency warning (new `GET /layers/{id}/columns/{name}/references`), E-07 feature-editing e2e persistence round-trips

**Remaining (P2/P3), still open for a future pass** — full detail in `docs-internal/audits/editing-audit-20260711.md`:
- E-08 write verbs 500 on raster/tabular datasets (read guards exist, write don't)
[...]

PR #463, opened 2026-07-12
fix(editing): editing-audit P2/P3 tail — write-path contracts, flag enforcement, draft UX

ishiland:
Clears the actionable P2/P3 findings from the 2026-07-11 editing audit (tracking issue #458). The P1s shipped earlier in #459/#460; this is the tail.

## Backend — write-path contracts, flag enforcement, drift (`bdb7fc5a`)
- **E-08** feature writes to raster/VRT datasets now 404 (mirror the read side) instead of 500 on a missing data table.
- **E-09 / E-26** feature-write DB errors classify by SQLSTATE (reusing `core/db/sqlstate.py`) → 400 for bad input (type mismatch 22P02, NOT NULL 23502, tabular no-column 42703), 503 for an outage. No more 500s.
- **E-11** enforce the `enable_dataset_editing` flag server-side on feature-write + column-DDL handlers (`require_dataset_editing_enabled` → 403 when off). It was UI-only; owners/admins could drive around the disabled toggle via the API.
- **E-13** add/rename/drop column now map DB errors to 400/503 like `alter_column_type` already did (dependent-object, lock timeout no longer 500).
- **E-14** record sub-resource writes gate on read-visibility first → a private, non-owned record 404s (no existence leak) instead of 403; a visible-but-non-owned record still 403s.
- **E-15** contacts/keywords/distributions writes bust the catalog cache (DCAT/STAC), and keyword writes re-embed (keywords are embedding text). Previously these siblings did neither, unlike the top-level metadata PATCH.
- **E-25** feature writers reject unknown property keys with 400 instead of silently dropping them (on PUT a typo also nulled the intended column).

## Frontend — draft + inline-edit UX (`1e84a4ad`)
- **E-16** disable Discard while a draft save is in flight (no misleading "discarded" toast after the PATCH commits).
- **E-17** `savePendingDrafts` reads pending drafts from a ref kept in lockstep with state, so the field staged by the pre-save blur is no longer dropped-then-cleared.
- **E-21** inline attribute-cell save failures surface the backend's specific reason (`formatMutationError`).
- **E-22 (partial)** keep the cell in edit mode through the save await so the saving spinner renders, and keep the editor open on failure for correction.

## Tests
- Backend: `TestEditingEnforcementAndContracts` (E-08/E-09/E-11/E-25), `TestRecordWriteContracts` (E-14/E-15), and an autouse conftest fixture that turns the editing flag on suite-wide (the guard's real branch still runs).
- Frontend: `use-draft-editing` stale-closure regression guard + discard-resets-ref guard; `PendingEditsBar` asserts Discard is disabled while saving.

## Deferred (with rationale, tracked in #458)
E-10 (gid semantics — larger design), **E-18 skip** (LWW acceptable — decided), **E-19 keep API-only** (removing drops a real capability; a UI carries unresolved lossy-cast UX — decided), E-20 (4-layer plumbing for a cosmetic diff gap), E-22 remainder (Escape/blur race — fiddly), E-23 (TerraDraw keyboard — upstream), E-24, E-27 (self-healing), E-28, E-29, E-30, E-31 (`extra="forbid"` changes the SDK contract for a robustness-only P3).

## Verification
- `uv run pytest` — features/records/layers/DDL-idor/metadata/settings suites green under the new flag fixture.
- `npm run typecheck`, `eslint`, and the touched vitest suites green.
- No OpenAPI/SDK regen: runtime status-code behavior only; response models unchanged.

https://claude.ai/code/session_01Mtp1seEXHjAu4vciRXebkd

### DIFF
--- a/backend/app/modules/catalog/authorization.py
+++ b/backend/app/modules/catalog/authorization.py
@@ -170,3 +170,23 @@ async def check_dataset_write_access(
         status_code=status.HTTP_403_FORBIDDEN,
         detail="Only the dataset owner or an admin may modify this dataset.",
     )
+
+
+async def require_dataset_editing_enabled(db: AsyncSession) -> None:
+    """Enforce the `enable_dataset_editing` admin flag. Raises 403 when off.
+
+    fix(#458 E-11): the flag gated only the UI (StructureTab), so an owner/admin
+    could still edit features and run column DDL through the API with editing
+    switched off. Enforce it on those write paths server-side. Metadata edits are
+    deliberately *not* gated — the UI keeps only structure/feature editing behind
+    this toggle, and the backend mirrors that boundary.
+    """
+    # Local import mirrors the other persistent_config call sites (e.g.
+    # service_metadata's REQUIRE_METADATA_FOR_PUBLISH) and avoids any import cycle.
+    from app.core.persistent_config import ENABLE_DATASET_EDITING
+
+    if not await ENABLE_DATASET_EDITING.get(db):
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail="Dataset editing is disabled by the administrator.",
+        )
--- a/backend/app/modules/catalog/features/router.py
+++ b/backend/app/modules/catalog/features/router.py
@@ -15,10 +15,11 @@ from fastapi import (
     status,
 )
 from fastapi.responses import JSONResponse
-from sqlalchemy.exc import OperationalError, ProgrammingError
+from sqlalchemy.exc import DBAPIError, OperationalError, ProgrammingError
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from app.modules.audit.service import AuditEvent, audit_emit
+from app.core.db.sqlstate import is_operational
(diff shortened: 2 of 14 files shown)


ishiland:
Both Codex P2s addressed in `1dde0f43`:

1. **Tabular row delete → 500** (`features/router.py`): `_require_feature_table` now also rejects non-spatial datasets (`geometry_type is None`) with 400. A DELETE touches no geometry, so it slipped past the geom-write DBAPIError handler and 500'd in `refresh_dataset_metadata`'s unconditional `geom_4326` read (an insert also crashed earlier in `effective_geometry_type`'s `None.strip()`). Created layers always carry a concrete `geometry_type`, so spatial layers are never blocked. New test: tabular POST/DELETE → 400.

2. **Keyword delete re-embed target** (`records/service.py`): `delete_keyword` is now scoped by `record_id`, so a keyword id addressed through a different record's path 404s instead of deleting the row and re-embedding the wrong record. New test: mismatched-path delete → 404, row survives.

Full backend records + features suites green.

ishiland:
@codex review

PR #465, opened 2026-07-12
feat(catalog): advertise GeoParquet in DCAT distributions + backfill existing datasets

ishiland:
**Stacked on #462** (`feat/catalog-easy-wins`) — completes the one item deliberately deferred there. Retarget this PR to `main` once #462 merges.

In #462, GeoParquet is advertised **live** for all datasets via OGC API Records + STAC (`_FORMAT_MEDIA`), but the **materialized DCAT distribution rows** were left alone because extending them cleanly needs a backfill for existing datasets. This PR does that.

## Changes
- **`_DISTRIBUTION_TEMPLATES`** (`records/service.py`): re-add the GeoParquet download so a fresh spatial ingest materializes a parquet distribution. Spatial-only — non-spatial datasets still get just csv + ogc_features.
- **Alembic `0013`** (data migration): backfill the parquet distribution for existing spatial datasets. Idempotent (`NOT EXISTS` + the `uq_record_distribution` constraint), reversible (downgrade deletes the auto-generated parquet rows), **no schema change** (so `alembic-check` drift is unaffected).
- **Count tests**: spatial download `4→5`, system total `6→7` (and the "6 system + 1 manual" case `7→8`); `generate_distributions` docstring `6→7`. Non-spatial (`2`) and the OGC `formats` list are unchanged.

## Verification
Ruff clean; migration revision chain verified (`0013 → 0012`, single head); full sweep confirms no other spatial distribution-count assertion. **No local test DB**, so the migration + updated counts are validated by CI's **Alembic Clean-DB Upgrade** + **Backend Tests** gates.

## Note for review
The migration writes to `catalog.record_distributions` for every spatial dataset without a parquet row — bounded, idempotent, and reversible, but it's a data migration, so worth a look at the SQL and the downgrade.

https://claude.ai/code/session_015cpiMW7X8MNmG4nN4dfzqB

### DIFF
--- a/backend/alembic/versions/0013_backfill_geoparquet_distributions.py
+++ b/backend/alembic/versions/0013_backfill_geoparquet_distributions.py
@@ -1,0 +1,61 @@
+"""Backfill GeoParquet download distributions for existing spatial datasets.
+
+fix(#462 follow-up): _DISTRIBUTION_TEMPLATES now generates a GeoParquet download
+distribution at dataset creation (records/service.py), but those rows are
+materialized — existing spatial datasets created before this deploy would never
+advertise the parquet download in DCAT feeds. This backfills one row per spatial
+dataset that lacks it, so DCAT distributions match what a fresh ingest produces.
+
+Idempotent: the NOT EXISTS guard (plus the uq_record_distribution unique
+constraint on record_id/distribution_type/format/url) makes re-running a no-op.
+Downgrade removes only the auto-generated parquet download rows.
+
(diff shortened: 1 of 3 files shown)


ishiland:
@codex review — note this is a data migration (0013) backfilling GeoParquet DCAT distributions; the SQL + downgrade are the main thing to scrutinize. No local test DB, so counts/migration are CI-gated.

ishiland:
Rebased onto main (only the 3 DCAT files now; #462 merged). @codex review — data migration 0013 + the distribution-count updates get their first CI run here.

PR #466, opened 2026-07-12
feat(observability): ship reference Prometheus/Grafana monitoring config

ishiland:
## Summary

Ship reference Prometheus/Grafana monitoring config. The backend already exports metrics but shipped nothing to consume them — this closes M1's "instrumented but unusable out of the box" gap.

## What's included

- `infra/monitoring/prometheus.yml` — scrapes **both** metrics endpoints (the API at `api:8000`, and the worker's own server at `worker:8001`) and loads the alert rules
- `infra/monitoring/alerts.yml` — 6 alerts: target-down, API 5xx rate, p95 latency, job-queue backlog, job failures, DB-pool saturation. Thresholds are tunable defaults.
- `infra/monitoring/grafana-dashboard.json` — 10-panel, import-portable dashboard (uses a datasource variable so import prompts for Prometheus)
- `RUNBOOK.md` §4 — setup + quickstart; `README.md` — `### Monitoring` pointer

## Design

Reference config only — **no bundled Prometheus/Grafana stack**. Operators point their own monitoring at the files (a compose profile can wrap them later if there's demand). This keeps two CVE-prone images and a default-credential surface out of the maintained footprint while still closing the observability gap.

## Verification

- `promtool check config` + `check rules` — config valid, `rule_files` resolves, all 6 rules parse
- Live Prometheus scrape against the running stack: **both targets `up=1`**, all 6 rules loaded `[ok]`, custom metrics queryable end-to-end

None. No app code, no compose services, no migrations — additive config files + docs only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -358,6 +358,13 @@ For day-2 operations, restore procedures, and incident response, see
 [RUNBOOK.md](RUNBOOK.md). For provider-specific configuration options, see
 [Backups & Restore](https://docs.getgeolens.com/guides/admin/backups/#backup-destinations).
 
+### Monitoring
+
+The API and worker export Prometheus metrics out of the box (HTTP rate/latency/
+errors, job-queue depth, DB pool, tile-cache). Reference scrape config, alert
+rules, and a Grafana dashboard ship in [`infra/monitoring/`](infra/monitoring/);
+see [RUNBOOK.md §4](RUNBOOK.md#4-monitoring) for the setup steps.
+
 ## Reference
 
 | Guide | Description |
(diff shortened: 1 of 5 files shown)


ishiland:
Addressed the P2 in e10428f4: swapped `http_request_duration_seconds_bucket` (buckets cap at `le=1.0`, so `histogram_quantile(...) > 1` could never fire) for the high-resolution `http_request_duration_highr_seconds_bucket` (buckets to 60s) — in both the alert and the dashboard p95/p99 panel. Verified against live `/metrics`.

@codex review

PR #467, opened 2026-07-12
docs(compose): clarify FRONTEND_TILE_BASE_URL vs CDN_BASE_URL for tile routing

ishiland:
One-line clarification prompted by live demo setup: the frontend `TILE_BASE_URL` (`FRONTEND_TILE_BASE_URL`) outranks the backend `CDN_BASE_URL` in `resolveTileBaseUrl`, so setting `CDN_BASE_URL` alone only flips the admin **Tile Cache** badge — it does not route tiles. The existing "Override for a CDN" comment implied otherwise.

Fronting the whole domain with a CDN (as the demo now does via Cloudflare) needs neither var — tiles route through the edge same-origin.

Doc-only; no functional change.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -399,8 +399,13 @@ FRONTEND_PORT=8080
 # ==============================================================================
 
 # --- CDN ---
-# [OPTIONAL] CDN origin URL for tile delivery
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

PR #473, opened 2026-07-12
fix(worker): scope enterprise port assertion to each overlay tier

ishiland:
## Problem

`assert_enterprise_ports_resolved()` (worker startup, after `bootstrap()`) demanded **all five** single-slot ports resolve to non-`Default*` implementations whenever `GEOLENS_EDITION=enterprise`:

```
processing_port, catalog_port, permission, identity, workflow
```

But the enterprise overlay only registers `permission` / `identity` / `workflow`. `processing_port` and `catalog_port` are provided **only** by the cloud (multi-tenant) overlay. So a **bare enterprise deployment** — enterprise overlay, no cloud/multi-tenant — crash-looped the worker on ports it was never meant to have, while the API process (which does not run this assertion) served fine. Net effect: a split-brain where the API is up but background jobs never process.

## Fix

Split the port checks into two tiers:

| Set | Ports | Required when |
|-----|-------|---------------|
| `_ENTERPRISE_PORT_CHECKS` | `permission`, `identity`, `workflow` | resolved edition is `enterprise` |
| `_CLOUD_PORT_CHECKS` | `processing_port`, `catalog_port` | `GEOLENS_TENANCY_MODE=multi_tenant` |

`multi_tenant` is the same signal that already mandates the cloud overlay (`check_tenancy_mode_supported`), so the cloud ports are asserted exactly where they are actually guaranteed to exist.

Also switches the trigger from the raw `GEOLENS_EDITION` env var to the **resolved** edition (`get_edition()`), so:
- a license-key activation that never sets `GEOLENS_EDITION` is still verified, and
- a `GEOLENS_LICENSE_ENFORCE` downgrade to community correctly stays silent.

All five ports are still logged at startup for observability regardless of tier.

## Impact

- **No schema / API / config changes.** Pure worker-startup assertion logic.
- Removes a false-positive boot failure for bare-enterprise workers; tightens coverage for license-key-activated deployments.

## Verification

```
uv run pytest tests/test_worker_overlay_resolution.py tests/test_worker.py \
  tests/test_layering.py::test_decomposed_service_modules_stay_within_size_budgets -q
uv run ruff check app/platform/extensions/bootstrap.py tests/test_worker_overlay_resolution.py
```

New regression tests:
- `test_no_raise_on_enterprise_without_cloud_ports` — bare enterprise with cloud ports still `Default*` must **not** raise.
- `test_licensed_enterprise_without_env_var_is_checked` — resolved-enterprise-without-env-var is still asserted.
- `test_raises_on_multi_tenant_missing_cloud_ports` — cloud ports are enforced under `multi_tenant`.

### DIFF
--- a/backend/app/api/main.py
+++ b/backend/app/api/main.py
@@ -26,7 +26,10 @@ from app.platform.cache.provider import init_tile_cache
 from app.core.db import async_session, engine
 from app.core.logging_config import setup_logging
 from app.core.runtime.staging import ensure_staging_ready, sweep_orphaned_exports
-from app.platform.extensions.bootstrap import bootstrap
+from app.platform.extensions.bootstrap import (
+    assert_enterprise_ports_resolved,
+    bootstrap,
+)
 from app.modules.auth.models import Role, User, UserRole
 from app.modules.auth.providers.local import hash_password
 from app.modules.auth.router import limiter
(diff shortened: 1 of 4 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
Addressed in `e4251efef`:
- **P2 (API/worker split-brain):** `assert_enterprise_ports_resolved()` now runs in the api/main.py lifespan too, so both entrypoints fail closed together. Structural parity tests added.
- **P1 (entitlement):** already resolved in `802cbd4e` — `("entitlement", "DefaultEntitlementPort")` is in `_CLOUD_PORT_CHECKS`, resolved+logged, with a dedicated fail-closed regression test. The re-report anchored to the new comment block.

@codex review

ishiland:
@codex review

PR #506, opened 2026-07-13
fix(api): align runtime and generated contracts

ishiland:
## Summary

- align OpenAPI authentication alternatives, RFC 7807 media types, global failures, rate limits, and route-specific runtime errors
- publish and enforce typed STAC and SSE contracts, including exact bbox dimensionality and public settings routes used by the frontend
- regenerate the OpenAPI snapshot and Python/TypeScript clients, adopt generated frontend status types, refresh affected caches, and add CI drift coverage
- make CLI round-trip tests consume the local CLI and generated SDK instead of unrelated installed packages

## Compatibility notes

- generated Python operations now return typed ProblemDetail values for documented 429, 500, and 503 responses instead of treating them as UnexpectedStatus; ProblemDetail.detail also accepts structured runtime details
- schema-free public STAC operations remain usable with Client or AuthenticatedClient
- skip is now the preferred pagination parameter on the two internal feeds, while deprecated offset aliases remain supported

## Audit disposition

Addressed H01-H05, M01-M07, L01, L02, L05, and L06 from the 2026-07-13 API contract audit.

- L03 is deferred because renaming the path parameter would break generated SDK method parameter names even though the URL is unchanged
- L04 required no code change: user and admin API-key list envelopes already consistently use items

## Verification

- backend Ruff check and format check
- backend full suite: 5,031 passed; the four cases affected by live source regeneration and router ratchets were rerun cleanly after stabilization; 59 skipped, 14 deselected
- focused API contract and STAC suites: 72 passed
- frontend lint, typecheck, production build, coverage, i18n, and 59 focused regression tests
- make openapi-check
- make sdks-check
- make cli-test: 214 unit tests plus 8 round-trip passes and 2 expected skips
- frontend npm run types:check


ishiland:
@codex review

ishiland (commented):
## Review: contract alignment — approve ✅

Reviewed the full hand-written surface (~42 files) and spot-checked the generated artifacts against the committed `openapi.json`. Every finding from the 2026-07-13 audit is addressed or deferred with a stated reason, and the runtime → OpenAPI → SDK → frontend chain stays consistent by construction. **No blocking issues and no correctness bugs.**

### Verification
- All 18 findings map to concrete fixes (H-01…H-05, M-01…M-07, L-01/02/05/06; L-03 deferred, L-04 no-op), and each has a matching assertion in `test_api_contract_openapi.py`.
- Confirmed the two fragile assumptions the OpenAPI post-processing rests on:
  - SlowAPI stores exempt routes as `"%s.%s" % (module, name)` and its decorators use `@functools.wraps`, so `_document_rate_limits`' `f"{endpoint.__module__}.{endpoint.__name__}"` exempt check matches exactly.
  - The AST drift-guard's `inspect.getsource(route.endpoint)` unwraps `@wraps` wrappers (`getsourcelines` calls `unwrap`), so it inspects the real handler body **through** `@limiter.limit`/`@limiter.exempt` — verified empirically, not blind to rate-limited routes.
- Spot-checked `openapi.json`: `ProblemDetail.detail` = `[array, object, string]`; `ApiKeyHeader`/`ApiKeyQuery` present; **248/252** ops carry 429 (the 4 without are exactly the exempt raster/token routes); SSE + STAC components registered; no `DatasetMetaUpdate` leak; no duplicate `StacItemCollection`.
- CI drift gates (OpenAPI Snapshot Drift, SDKs Drift Gate, `types:check`, CLI round-trip) are green and **non-skipped**, so the 277 generated files are consistent with source.

### Nicely done
- Systematic schema post-processing over ~250 manual `responses=` annotations — the right altitude, and it can't drift.
- Guards against **over**-documentation too: `test_upload_quota_and_upstream_failures_are_route_scoped` asserts 502 is absent where unwarranted, and the exempt raster-proxy correctly lacks 429.
- `/health` 503 keeps `model=HealthResponse` rather than blindly applying the problem+json pattern.
- `make cli-test` now exercises the real local CLI+SDK via `--with-editable` (better than the audit's keyring-only suggestion).

### Low-severity notes (non-blocking)
1. **New 500 surface on STAC items.** `_build_item_response` / `StacItemCollection` construction now run `StacItemResponse.model_validate(...)` where the old path passed the dict straight through. Real serializer output is safe (`properties.datetime` is always set; bbox is a 4-float PostGIS envelope), but a future serializer change that drops a required field or emits a non-4/6 bbox would now 500 instead of passing through. A one-line comment at the `model_validate` call marking it as a validation boundary would help the next reader.
2. **ChatPanel: a streamed 500 now renders inline instead of offering retry.** Deliberate and tested, tied to the SSE `status` unification (map-generation errors now carry `status` too). Flagging as an intentional UX change, not a regression.
3. **Drift-guard scope.** `test_direct_route_http_exceptions_are_documented` only catches `HTTPException` raised *directly* in handler bodies (literal statuses), not those raised in called helpers/services. The docstring is honest about this; noting it so the net isn't over-trusted later.

Nit: the exception handler builds the problem body two ways — a manual dict for the dict/list branch, `ProblemDetail(...).model_dump()` for strings. Harmless now that `detail` accepts dicts; not worth churning.

ishiland:
## Follow-up: adversarial deep-dive verification

Ran a second, independent pass (parallel sweeps for security completeness, 429/503 enforcement accuracy, full H-04 appendix coverage, STAC runtime behavior, and SDK/frontend consumer breakage). **Net: the PR holds up — no blocking issues, no security gaps, no breaking changes, zero uncovered appendix items.** Independently confirmed:

- **H-01 airtight.** Every auth-*enforcing* dependency routes through `OAuth2PasswordBearer` (so the normalizer catches it); the only bearer-less auth dep is optional and used solely on the 3 public STAC reads; no manual credential checks exist outside `dependencies.py`. All 30 no-`security` operations verified genuinely public — no protected route is advertised as anonymous.
- **429/503 accuracy is real, not aspirational.** `SlowAPIMiddleware` + the `60/second` global default (fires on undecorated non-exempt routes) and the global `DBAPIError→503` handler are pre-existing runtime behavior, so the blanket 429/503 documentation is accurate. The 4 ops lacking 503 are genuinely DB-free; the 4 lacking 429 are the exempt tile routes with no 429 path.
- **H-04 appendix: full coverage.** All 38 direct-raise ops + every middleware/dependency group (429 on the 32 rate-limited routes, 502/503 on the six non-streaming AI ops, raster-proxy 503-despite-being-exempt) resolve to the right status in the committed spec.
- **SDK/frontend clean.** Python SDK 429/500/503→`ProblemDetail` branch counts match the spec exactly (248/252/248, zero `UnexpectedStatus` fall-through); every frontend caller of the retyped status functions discards the return value.
- **Schema surgery is valid** — no new dangling `$ref`s; the manually-registered `ProblemDetail`, SSE event, and STAC components all resolve.

Two things worth a look (both minor, non-blocking):

1. **`Retry-After` is documented but dropped on the 2 handler-raised tile 429s.** `tile_endpoint`/`cluster_tile_endpoint` raise `HTTPException(429, headers={"Retry-After": "2"})` (`tiles/router.py:1701,1726`), but the global `http_exception_handler` rebuilds the response and does **not** forward `exc.headers`, so the `Retry-After` header that `RATE_LIMIT_RESPONSE` advertises isn't actually emitted for those two ops. (SlowAPI-enforced 429s set it via `_rate_limit_handler`, so those are fine.) Slightly ironic given the PR's runtime↔contract-alignment theme. One-line fix if you want parity: forward `headers=exc.headers` in that handler — it would also preserve `WWW-Authenticate`/`Allow` on other raises.

2. **Augmenting my STAC note above:** the item/item-collection change isn't only a new 500-surface — because `StacItemResponse.model_dump(mode="json")` has no `exclude_none`, item bodies now emit explicit `null`/`[]` for optional fields `main` previously omitted (`bbox`, `collection`, `stac_extensions`, `properties.title/…`, per-asset `type/roles/…`). It's additive and backward-compatible for STAC clients (absent ≡ null-optional) and drops no extension fields — but it's a real wire-format change worth knowing if any snapshot test or strict consumer diffs exact bodies.

Out of scope, noted for later: `backend/openapi.json` carries 20 pre-existing dangling `#/$defs/…` refs on the OGC-Features / `features` geo+json bodies (identical on `main`, untouched here). The typed-component pattern this PR establishes for STAC is exactly how to fix them, and a `$ref`-resolution assertion in `test_api_contract_openapi.py` would catch that whole class going forward.

ishiland:
@codex review

PR #510, opened 2026-07-14
fix(admin): harden control-plane governance and lifecycle

ishiland:
## Summary

- enforce capability-based admin access across backend routes, navigation, settings, AI operations, job retry, and tenant-safe permission matrices
- serialize identity lifecycle decisions, preserve the final active administrator, add explicit active/suspended/deactivated states, and fingerprint API keys
- make configuration preview/import/reset atomic and exact with signed preview tokens, overwrite locking, OAuth-account cascade reporting, and safe lockout checks
- make audit exports, user exports, embedding backfills, stale cleanup, and retry telemetry durable, correlated, cancellation-safe, and privacy-safe
- expose the new lifecycle, audit, cleanup, and config contracts in the admin UI and regenerate OpenAPI plus both SDKs

- adds Alembic revision `0016_admin_identity_hardening`
  - adds API-key fingerprints
  - normalizes user lifecycle state and installs a consistency check constraint
- adds explicit user lifecycle status mutations and additive API response fields for fingerprints, cleanup outcomes, and configuration/OAuth deletion counts
- overwrite imports now require the matching signed dry-run preview token and apply the exact reviewed plan under a database lock
- no new environment variables or secrets are introduced

## Verification

- backend: `5115 passed, 59 skipped` (`pytest -n 4 -m 'not perf'`)
- frontend coverage: `3601 tests passed`; 70.98% statements / 66.38% branches / 65.67% functions / 72.92% lines
- architecture: layering guard passed
- migrations: clean PostgreSQL upgrade through head plus `alembic check` passed
- contracts: `make openapi-check`, `make sdks-check`, and `npm run types:check` passed
- SDK: TypeScript build and 4 tests passed
- CLI: 214 unit tests passed; 8 round-trip tests passed, 2 skipped
- frontend: lint, typecheck, production build, and all i18n gates passed
- repository: Ruff check/format, version coherence, diff checks, and pre-commit security hooks passed

## UI evidence

The UI changes are permission/state behavior updates rather than a visual redesign. Focused component tests cover capability redirects, navigation visibility, lifecycle controls, config preview warnings, and invalid audit-filter handling.


ishiland:
@codex review

ishiland (commented):
## Code review

Thorough pass over the backend governance/lifecycle/config-ops changes plus the frontend capability gating. **Verdict: strong.** No correctness or security defects found — three trivial nits only. Below is what I actively verified (against the branch), so the confidence is auditable.

### Correctness — verified, not assumed

- **The new `is_active = (status = 'active')` CHECK constraint is safe across every write path.** Enumerated all `User(...)` creations and `status`/`is_active` mutations: registration (`pending`/`False`), OAuth JIT (`active`/`True`), email-verify (`is_active=True, status="active"`), admin `create_user` (`active`/`True`), and the boot seed in `api/main.py` (sets `is_active=True`, lets `status` fall to its `server_default='active'`). The migration normalizes legacy `active+is_active=false` rows *before* installing the constraint. No path can violate it.
- **Last-admin serialization is race- and deadlock-free.** `_get_lifecycle_user` takes advisory-xact-lock → row `FOR UPDATE`; `_get_other_admin_count` counts only *viable* admins (`status='active' AND is_active AND role=admin`). Advisory→row ordering is consistent across deactivate/update/delete; approve/reject use row-lock only (they can't reduce viable admins), so no lock cycle. Self-demotion/self-disable is blocked at the router **and** caught by `_ensure_not_last_admin`.
- **Preview-token flow is sound.** Apply re-runs preflight under the `EXCLUSIVE` table lock and re-verifies the token's `state`/`payload` digests against freshly-computed values → any concurrent change or payload edit yields 409, and a post-apply replay can't re-match (provider IDs/`updated_at` change). The `oauth_accounts_deleted != plan.oauth_accounts_deleted` guard fails closed before commit.
- **Streaming exports** use their own `async_session()` (never the request session after the handler returns); outcome bookkeeping is shielded and capped at 5s.
- **Overwrite dry-run rejects secret-less non-SAML providers symmetrically with apply** (`OAuthProviderCreate._validate_per_type` requires `client_secret`), so an export→overwrite round-trip fails at *preview* with a clear per-index message and never issues a doomed token.

### Findings (all minor)

1. **`snapshot_at` is dead parameter surface** — `backend/app/modules/audit/service.py`. Threaded through `_apply_filters` / `query_audit_logs` / `stream_audit_logs` (and applies a `created_at <= snapshot_at` filter), but **no caller ever passes it** — export relies on `exclude_resource_id` + the MVCC snapshot per its own comment. Wire it or drop it.
2. **`assert api_key.fingerprint is not None`** — `admin/router.py` and `auth/router.py`. Stripped under `python -O`, so it's not a runtime guarantee. Harmless (the fingerprint is always set, and `ApiKeyCreateResponse.fingerprint` is a required `str`), but don't treat `assert` as an invariant check.
3. **`me_permissions` issues one `check_permission` await per capability, sequentially** — correct and cheap for the in-memory default extension, but an enterprise `PermissionExtension` that consults the DB per call turns this into O(capabilities) queries on a hot endpoint. A batch seam would future-proof it.

Non-issues checked and cleared: `UserEditDialog` dropping the `?? fallback` on `userStatusColors[user.status]` (safe — closed 4-value enum, all mapped); no new `httpx.AsyncClient(follow_redirects=True)` (SSRF Rule 2 N/A); new handlers read `User`/`AuditLog`/`IngestJob`, not the Rule-1 dataset entities, and are permission-gated; `StaleCleanupResponse(**details)` field set matches `as_dict()` exactly; jobs cleanup does not double-reap staged paths.

### Risk / deploy notes
- Migration 0016 is **not losslessly reversible**: downgrade collapses `suspended`→`active+is_active=false`. Worth calling out in the release notes.
- Contract surface changed (new `status` field, new `SettingChange`/`OAuthProviderChange` actions, `X-Config-Preview-Token` header, extra response fields). SDKs/OpenAPI are regenerated in-PR — just ensure `make openapi-check`/`sdks-check` are green on the final commit.

Nothing blocks merge; I'd land it after (optionally) resolving nit #1.

ishiland:
@codex review

PR #528, opened 2026-07-16
fix(editing): remediate editing-audit 2026-07-16 findings (E-32..E-51)

ishiland:
Remediates all 20 findings from the 2026-07-16 editing audit (`docs-internal/audits/editing-audit-20260716.md`). Prior-audit remediations (E-01..E-31, #458) were regression-checked intact; this PR covers the new tail.

## P1
- **E-32** — multiline metadata drafts were effectively unsaveable: `InlineEdit`'s unmount cleanup re-ran on every parent re-render (inline-arrow `onDirtyChange` props), instantly self-reverting the dirty flag, so the pending-edits bar never appeared; multiline blur cancelled, and Ctrl+Enter (hover-tooltip-only) was the sole save path. Fixed with a ref-based unmount-only cleanup + visible Save/Cancel buttons (mousedown-prevented so blur can't cancel first). Live-verified pre-fix in the audit.

## P2
- **E-33** — column identifiers now quoted in all layer DDL + `get_distinct_values`; reserved-word columns (`desc`, `order` — routine ogr2ogr output) no longer break every schema op.
- **E-34** — `quality_detail` recomputed after column DDL (parity with reupload).
- **E-35** — unchanged cell values commit as cancel; blur no longer fires no-op PATCHes that bumped `updated_at`, rolled the tile `_v=`, purged Valkey, and wrote `attribute.edit` audit events for non-changes (observed live).
- **E-36** — map-path feature-edit errors surface backend detail via `formatMutationError` (E-21 parity).
- **E-37** — column-DDL audit feed gates on `check_dataset_write_access`; any authenticated user could previously enumerate editor usernames/user_ids on public datasets. (New `check_dataset_write_access` on ProcessingPort; enterprise overlay doesn't override this port.)
- **E-38/E-39** — validation errors announced + field-associated (`aria-invalid`/`aria-describedby`/`role="alert"`); the inline cell editor has an accessible name and numeric `inputMode`.

## P3
E-40 (offset-pagination documented), E-41/E-42 (SQLSTATE-granular error mapping: `42P01`→503, unknown→500), E-43 (`sample_values` reconciled on rename/drop), E-44 (`user_modified_fields` honored on column revive), E-45 (`DistributionUpdate.url` HTTP(S)-validated), E-46 (explicit-null clears on contact/distribution PATCH; non-nullable fields 422), E-47 (record sub-resource writes emit `metadata.edit` into dataset history), E-48 (no-op `tile_columns` PATCH no longer purges tiles), E-49/E-50/E-51 (SchemaEditor labels, drop-confirm waits for the map-references warning, open cell editor closes when the row set changes).

## API/schema impact
- `backend/openapi.json` + SDKs regenerated (two endpoint description updates only; no shape changes).
- Behavior changes: column-DDL audit feed now 404s non-owners on public datasets (E-37); contact/distribution PATCH treats explicit null as clear and 422s nulls on required fields (E-46); distribution PATCH validates URLs (E-45).

## Tests
- Backend: reserved-word DDL lifecycle, quality-recompute, public-dataset audit-feed denial, URL validation, clear-to-null + 422 contracts, history audit events, no-op tile_columns guard — all passing (`test_layer_column_ops`, `test_audit_column_ddl_feed`, `test_record_subresource_edits` (new), `test_dataset_tile_columns_metadata`), plus adjacent suites (`test_features_crud`, `test_dataset_metadata_idor`, `test_metadata_roundtrip`, `test_layering`, `test_records_authz_sec001`) — 110 passed.
- Frontend: E-32 dirty-persistence + Save/Cancel button suite (`InlineEdit.test.tsx`), E-35/39/51 contracts (`AttributeTable.test.tsx`), new `SchemaEditor.test.tsx` (first direct coverage of the destructive-DDL surface). 204 tests across the dataset surface passing; `npm run typecheck`, eslint, and `test:i18n` green (one new key, `attributes.cellEditorLabel`, added to all four locales).

## Verification
- `make openapi-check` green; `make sdks` regenerated (sdks-check drift was the uncommitted doc updates).
- Browser smoke check of E-32/E-35/E-38/E-39/E-50 against the live dev stack to follow on this PR.

https://claude.ai/code/session_01Xd64KjyK8rmgEfvWemkP5w

### DIFF
--- a/backend/app/core/processing_port.py
+++ b/backend/app/core/processing_port.py
@@ -216,6 +216,16 @@ class ProcessingPort(Protocol):
         user_roles: set[str] | None = None,
     ) -> set[str]: ...
 
+    async def check_dataset_write_access(
+        self,
+        session: AsyncSession,
+        dataset: Any,
+        dataset_id: uuid.UUID,
+        user: Identity,
+        *,
+        user_roles: set[str] | None = None,
+    ) -> set[str]: ...
+
     async def get_user_roles(
         self, session: AsyncSession, user: Identity
     ) -> set[str]: ...
--- a/backend/app/modules/audit/router.py
+++ b/backend/app/modules/audit/router.py
@@ -413,10 +413,13 @@ async def get_column_ddl_feed(
     to dataset owners so they can detect editor-initiated schema changes.
 
     Access control (AGENTS.md Pre-Commit Checklist Rule 1):
-    - Owner + granted roles: 200 with their own dataset's DDL history
-    - Non-owner editor (no grant): 404 (check_dataset_access raises 404 for
-      private datasets)
+    - Owner: 200 with their own dataset's DDL history
     - Admin: 200 (admin access is always allowed)
+    - Anyone else — including authenticated readers of a PUBLIC dataset: 404
+      via check_dataset_write_access. fix(#458 E-37): the feed previously used
+      check_dataset_access (read visibility), which let any logged-in user
+      enumerate editor usernames/user_ids on public datasets, contradicting
+      this owner-facing contract.
     - Anonymous: 401 (get_current_active_user dependency)
 
     The dataset 404-before-auth-query ordering ensures non-existent datasets
@@ -433,10 +436,9 @@ async def get_column_ddl_feed(
             status_code=status.HTTP_404_NOT_FOUND, detail="Dataset not found"
         )
 
-    # Step 2: enforce visibility / ownership gate
-    # check_dataset_access raises HTTPException(404) for non-owners of private datasets,
-    # consistent with the column-DDL write endpoints from Phase 1061 Plan 02.
-    await port.check_dataset_access(db, dataset, dataset_id, user)
+    # Step 2: enforce ownership gate (owner-or-admin, 404 on denial), matching
+    # the column-DDL write endpoints this feed reports on (fix(#458 E-37)).
+    await port.check_dataset_write_access(db, dataset, dataset_id, user)
 
     # Step 3: fetch DDL history. Preserve the old offset parameter while new
     # clients converge on the repository-wide skip/limit convention.
--- a/backend/app/modules/catalog/datasets/api/router.py
+++ b/backend/app/modules/catalog/datasets/api/router.py
@@ -293,6 +293,10 @@ async def update_dataset_metadata(
         )
     await check_dataset_write_access(db, dataset, dataset_id, user)
 
+    # fix(#458 E-48): capture the pre-update value so a PATCH that echoes the
+    # same tile_columns doesn't roll the tile version / purge the tile cache.
+    tile_columns_before = list(dataset.tile_columns or [])
+
     try:
         dataset = await update_user_metadata(
             db,
@@ -330,13 +334,19 @@ async def update_dataset_metadata(
     # fix(#525 B-038): a tile_columns change alters the attribute set embedded
     # in vector tiles — roll the _v= URL cache-buster in the same transaction
     # (the post-commit Valkey purge cannot reach CDN/browser caches).
-    if "tile_columns" in meta.model_fields_set:
+    # fix(#458 E-48): only when the value actually changed — a no-op echo used
+    # to purge every cached tile for the table.
+    tile_columns_changed = (
+        "tile_columns" in meta.model_fields_set
+        and list(dataset.tile_columns or []) != tile_columns_before
+    )
+    if tile_columns_changed:
         dataset.bump_tile_cache_version()
     await db.commit()
     await db.refresh(dataset)
     await db.refresh(dataset.record)
     await invalidate_catalog_cache()
-    if "tile_columns" in meta.model_fields_set:
+    if tile_columns_changed:
         tile_cache = get_tile_cache()
         if tile_cache is not None:
             await tile_cache.invalidate_table(dataset.table_name)
--- a/backend/app/modules/catalog/datasets/api/router.py
+++ b/backend/app/modules/catalog/datasets/api/router.py
@@ -295,7 +295,12 @@ async def update_dataset_metadata(
 
     # fix(#458 E-48): capture the pre-update value so a PATCH that echoes the
     # same tile_columns doesn't roll the tile version / purge the tile cache.
-    tile_columns_before = list(dataset.tile_columns or [])
+    # None and [] are captured distinctly (fix(#528) codex r1): None means
+    # per-zoom defaults while [] means "never project attributes"
+    # (_select_tile_columns), so a None↔[] flip IS a tile-content change.
+    tile_columns_before = (
+        list(dataset.tile_columns) if dataset.tile_columns is not None else None
+    )
 
     try:
         dataset = await update_user_metadata(
@@ -338,7 +343,7 @@ async def update_dataset_metadata(
     # to purge every cached tile for the table.
     tile_columns_changed = (
         "tile_columns" in meta.model_fields_set
-        and list(dataset.tile_columns or []) != tile_columns_before
+        and dataset.tile_columns != tile_columns_before
     )
     if tile_columns_changed:
         dataset.bump_tile_cache_version()
--- a/backend/app/modules/catalog/records/router.py
+++ b/backend/app/modules/catalog/records/router.py
@@ -10,6 +10,7 @@ from sqlalchemy.ext.asyncio import AsyncSession
 from sqlalchemy.orm import joinedload
 
 from app.core.identity import Identity
+from app.modules.audit.service import AuditEvent, audit_emit
 from app.modules.auth.dependencies import get_optional_user, require_permission
 from app.modules.catalog.authorization import (
     check_dataset_access_or_anonymous,
@@ -72,7 +73,14 @@ _LANGUAGE_PATH = Path(
 )
 
 
-async def _propagate_record_write(record_id: uuid.UUID, *, reembed: bool) -> None:
+async def _propagate_record_write(
+    record_id: uuid.UUID,
+    *,
+    reembed: bool,
+    db: AsyncSession | None = None,
+    user: Identity | None = None,
+    details: dict | None = None,
+) -> None:
     """Keep downstream surfaces coherent after a record sub-resource write.
 
     fix(#458 E-15): contacts/keywords/distributions feed the DCAT-US/STAC feeds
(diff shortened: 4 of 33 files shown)


ishiland:
Browser smoke check completed against the live dev stack (branch code, flag temporarily enabled then restored to false):

- **E-32** ✅ pending-edits bar now appears while typing in a multiline field; visible Save/Cancel buttons render under the textarea; the Save button commits despite the blur (mousedown-prevented); Discard reverts cleanly.
- **E-35** ✅ opening a cell editor and clicking away with an unchanged value closes the editor with **zero** PATCH requests (api access log clean, `updated_at` untouched).
- **E-38/E-49** ✅ reserved-name validation error carries `role="alert"`, `aria-invalid="true"`, and a matching `aria-describedby`; the name input is label-associated and the type select has an accessible name.
- **E-39** ✅ the cell editor announces "Edit ada for feature 1" and gets `inputMode="decimal"` on integer columns.
- **E-50** ✅ drop-column Confirm is disabled while the map-references query loads (+30ms probe) and enables once the "Used by 1 saved map(s)" warning renders (+730ms).

Backend-only findings (E-33 reserved-word DDL, E-34 quality recompute, E-37 audit-feed gate, E-41..E-48) are covered by the integration tests in this PR, which run against real Postgres.

frontend/src/components/dataset/InlineEdit.tsx
```diff
@@ -148,19 +158,44 @@ export function InlineEdit({
...
-          className={cn(inputClasses, 'resize-none min-h-[3rem]')}
-          rows={3}
-          title={t('common:inlineEdit.hint', { defaultValue: 'Ctrl+Enter to save, Escape to cancel' })}
-        />
+        <div>
+          <textarea
+            ref={inputRef as React.RefObject<HTMLTextAreaElement>}
+            value={draft}
+            onChange={(e) => handleDraftChange(e.target.value)}
+            onBlur={cancel}
```
ishiland: **E-32 follow-up (a11y, medium):** the new Save/Cancel buttons are pointer-only. Tabbing from the textarea fires blur → `cancel()` → `setEditing(false)`, which unmounts the buttons before a keyboard user can activate them — `onMouseDown preventDefault` only protects pointer clicks. Keyboard users are left with Ctrl+Enter, hinted only in a hover `title`, which is the discoverability gap E-32 was about.

Not a regression (blur cancelled before too), but the fix is small: skip `cancel()` here when `e.relatedTarget` is inside the editor wrapper, e.g.

```tsx
onBlur={(e) => {
  if (e.currentTarget.parentElement?.contains(e.relatedTarget as Node)) return;
  cancel();
}}
```

(or hoist blur handling to the wrapping `<div>` and do the `contains(relatedTarget)` check there). Worth doing before merge or as an immediate follow-up.

backend/app/core/processing_port.py
```diff
@@ -216,6 +216,16 @@ async def check_dataset_access(
         user_roles: set[str] | None = None,
     ) -> set[str]: ...
 
+    async def check_dataset_write_access(
```
ishiland, line 219: **Verify before merge:** in this repo only `DefaultProcessingExtension` and the test fake implement `ProcessingPort` (both updated here), but the enterprise overlay is out-of-repo. If it ships its own `ProcessingPort` implementation without `check_dataset_write_access`, the audit-feed endpoint 500s (AttributeError) on enterprise installs. The PR body asserts the overlay doesn't override this port — a quick grep in the overlay repo makes that claim load-bearing instead of asserted.

backend/app/modules/audit/router.py
```diff
@@ -433,10 +436,9 @@ async def get_column_ddl_feed(
             status_code=status.HTTP_404_NOT_FOUND, detail="Dataset not found"
         )
 
-    # Step 2: enforce visibility / ownership gate
-    # check_dataset_access raises HTTPException(404) for non-owners of private datasets,
-    # consistent with the column-DDL write endpoints from Phase 1061 Plan 02.
-    await port.check_dataset_access(db, dataset, dataset_id, user)
+    # Step 2: enforce ownership gate (owner-or-admin, 404 on denial), matching
+    # the column-DDL write endpoints this feed reports on (fix(#458 E-37)).
+    await port.check_dataset_write_access(db, dataset, dataset_id, user)
```
ishiland, line 441: **Note (deliberate, just confirming):** this tightens beyond the public-dataset leak itself — grant-holding editors previously got 200 from the DDL feed and now get 404. The docstring frames the feed as owner-facing (detect *editor*-initiated changes), so denying editors is coherent, and the PR body calls it out. Flagging only so the grantee-facing behavior change is a conscious choice, not a side effect.

backend/app/modules/catalog/records/router.py
```diff
@@ -93,6 +107,29 @@ async def _propagate_record_write(record_id: uuid.UUID, *, reembed: bool) -> Non
...
+                    AuditEvent(
+                        user_id=user.id,
+                        action="metadata.edit",
+                        resource_type="dataset",
+                        resource_id=dataset_id,
+                        details=details,
+                    ),
+                )
+                await db.commit()
+        except Exception:  # broad: audit is best-effort here; the write is committed
```
ishiland, line 129: **Minor:** this broad `except Exception` also swallows genuine programming errors — a missing import or bad kwarg in this block would ship silently (the new history test is what actually proves the path works). Best-effort audit is a fine design and it's documented, but consider narrowing to `SQLAlchemyError` (or `(SQLAlchemyError, ConnectionError)`) later. Also worth knowing: the audit event commits separately from the main write, so a crash between the two commits drops the event — acceptable for best-effort.

ishiland (commented):
Reviewed all 20 remediations. Verified against the checkout: `_sql_quote_ident` exists (column_stats.py:51), the SQLSTATE constants exist, `_raise_ddl_db_error` rolls back before classifying, `select`/`Dataset` were already imported in the records router, the session factory uses `expire_on_commit=False` (session.py:26) so the second commit in `_propagate_record_write` is safe, and the E-46 null-rejection lists exactly match the NOT NULL columns on `RecordContact` and `RecordDistribution`. E-45 mirrors the create-path `HttpUrl(v)` pattern verbatim.

Four inline comments: one a11y fix worth making before merge (E-32 buttons are pointer-only), one pre-merge verification (enterprise overlay vs. the new `ProcessingPort` method), two notes.

Nits, no action required:
- `_qcol` duplicates `_sql_quote_ident`; a cross-domain import would fight `test_layering`, so the duplication is the right call unless a third copy appears.
- The E-51 effect keys on undebounced `activeFilters`, so the open cell editor closes on the first filter keystroke, before the row set actually changes. Conservative and safe.
- The E-48 comparison normalizes `None` and `[]` to the same value, so flipping between explicit `[]` and `null` won't bump the tile version — both mean "default", so this looks correct.
- The AttributeTable source-regex contract tests are brittle by nature, but jsdom can't render the virtualized rows and the file already established that pattern for PERF-07, with e2e covering the live flow.

Otherwise ship-shape: E-37 closes a real username/user_id enumeration, E-45 blocks `javascript:`/`file://` re-entering the DCAT/STAC feeds, E-48/E-35 remove pointless tile purges and audit writes, and the test additions are strong (reserved-word DDL lifecycle, public-dataset feed denial, first direct SchemaEditor suite, and an E-32 test that reproduces the inline-arrow re-render).

backend/app/modules/catalog/datasets/api/router.py
```diff
@@ -330,13 +334,19 @@ async def update_dataset_metadata(
     # fix(#525 B-038): a tile_columns change alters the attribute set embedded
     # in vector tiles — roll the _v= URL cache-buster in the same transaction
     # (the post-commit Valkey purge cannot reach CDN/browser caches).
-    if "tile_columns" in meta.model_fields_set:
+    # fix(#458 E-48): only when the value actually changed — a no-op echo used
+    # to purge every cached tile for the table.
+    tile_columns_changed = (
+        "tile_columns" in meta.model_fields_set
+        and list(dataset.tile_columns or []) != tile_columns_before
+    )
```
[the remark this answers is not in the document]
ishiland: Fixed in bbde1460f — the guard now preserves the None-vs-[] distinction (None = per-zoom defaults, [] = never project attributes per `_select_tile_columns`, so the flip IS a tile-content change). `test_tile_columns_noop_patch_does_not_bump_tile_version` extended with the None→[] case.

frontend/src/components/dataset/InlineEdit.tsx
```diff
@@ -148,19 +158,44 @@ export function InlineEdit({
...
+            onKeyDown={handleKeyDown}
```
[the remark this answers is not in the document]
ishiland: Fixed in bbde1460f — the textarea blur now checks `relatedTarget` containment against the editor container, so Tab to the editor's own Save/Cancel keeps it mounted; keyboard-path tests added (Tab-to-Save commits, leaving the editor entirely still cancels).

frontend/src/components/dataset/InlineEdit.tsx
ishiland: Fixed in bbde1460f (same fix as the Codex thread above): blur ignores focus moves into the editor's own container, so keyboard users can Tab to Save/Cancel; pointer path keeps the mousedown-preventDefault. Both paths covered in InlineEdit.test.tsx.

backend/app/core/processing_port.py
ishiland: Verified before this PR was opened: the enterprise overlay never registers a `processing_port` extension (searched the overlay repo's sources — no ProcessingPort implementation and no `registry[\"processing_port\"]` key), so the community `DefaultProcessingPort` (updated here) serves in both editions and the audit-feed endpoint cannot hit a missing method.

backend/app/modules/audit/router.py
ishiland: Confirmed deliberate: the feed exists so owners can detect *editor*-initiated schema changes — grant-holding editors are the audited population, not the audience, so denying them is coherent with the docstring's contract. Pinned by the new public-dataset denial test.

backend/app/modules/catalog/records/router.py
ishiland: Agreed it's broad — kept for shape-consistency with the sibling best-effort steps in `_propagate_record_write` (cache bust and embed defer catch broadly for the same committed-write reason), and the new history test proves the path executes. Will narrow to `SQLAlchemyError` if we revisit the helper.

PR #532, opened 2026-07-16
fix(quicklook): call hooks unconditionally so a cached 404 can't crash consumers

ishiland:
## Summary

`useQuicklook` violated the Rules of Hooks: once a quicklook 404 populated the SP-07 session negative-cache, the next re-render of the **same** component instance early-returned before any hooks ran, throwing `Rendered fewer hooks than expected` and crashing `SearchResultCard` (search page) / `DatasetSearchPanel` (builder) to the page error boundary. Reachable in production whenever a quicklook file is missing from storage — exactly the scenario the negative-cache exists for.

## Fix

- Hooks (`useQueryClient`, `useEffect`, `useQuery`) now run on every render; the fetch is gated with the query's `enabled` flag (`datasetId != null && !knownMissing`) instead of early returns.
- Removes the `eslint-disable react-hooks/rules-of-hooks` suppression that documented the misconception.
- No behavior change otherwise: same statuses, same negative-caching, same blob-URL lifecycle.

## Verification

- New regression test re-renders the same hook instance after a 404 — it reproduces the exact crash against the old implementation and passes with the fix.
- `npx vitest run` on use-quicklook, SearchResultCard, and quicklook-cache suites: 52 passed.
- `npx eslint` on changed files and `npm run typecheck`: clean.

### DIFF
--- a/frontend/src/components/maps/hooks/__tests__/use-quicklook.test.ts
+++ b/frontend/src/components/maps/hooks/__tests__/use-quicklook.test.ts
@@ -113,6 +113,26 @@ describe('useQuicklook', () => {
     expect(mockApiFetchBlob).toHaveBeenCalledTimes(1);
   });
 
+  // Test 4b: regression — after a 404 populates the negative cache, the SAME
+  // component instance must survive a re-render. The old implementation
+  // early-returned before its hooks once isQuicklookKnownMissing flipped true,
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

ishiland:
@codex review

PR #531, opened 2026-07-16
feat(ai): dataset-scoped AI chat on the dataset page

ishiland:
## What

Adds **dataset-scoped AI chat** (dataset-chat v1): a read-only "Ask AI" panel on the dataset detail page that answers natural-language questions about one dataset's data, plus an "Open in builder" action on query results.

## How

**Backend**
- New `POST /ai/chat/dataset/stream/` SSE endpoint (`processing/ai/router.py`). It reuses the entire hardened map-chat pipeline — light-model SQL generation, sqlglot fail-closed validation, RBAC table allowlist, read-only sandboxed execution, per-user token budget — with one synthetic server-built layer.
- `_validate_chat_dataset()` resolves + authorizes the attacker-controlled `dataset_id` via the port (`check_dataset_access`, 404 on denial — no existence oracle; mirrors `_authorize_metadata_dataset` SEC-D). Raster/VRT datasets are rejected with 422 (no queryable `data.*` table). All layer context (table, columns, samples, counts) comes from the DB — the client only names the dataset. Sample values respect `AI_SEND_SAMPLE_VALUES`.
- `can_edit=False` selects the read-only tool set (`query_data` only), and a dataset-framed system prompt (`build_dataset_chat_system_prompt`, new `chat_dataset.py` sub-module) replaces the map-editing prompt via a `system_prompt_override` parameter on `stream_chat_edit`.
- The prompt-injection sanitizer + prompt sizing caps moved from `chat_service.py` to `chat_constants.py` so both prompt builders share them (keeps the facade under its `test_layering` size budget).

**Frontend**
- `DatasetChatPanel` (mirrors `ViewerChatPanel`): floating "Ask AI" launcher on `DatasetPage`, streaming answers, compact query-result table (shared `QueryResultTable`), and "Open in builder" — creates a map and stages the dataset via the existing `?add_dataset=` flow. Self-gates on `useAIAvailability` (token + `use_ai_chat` + AI configured); rendered for vector/table datasets only, matching the backend gate.
- Shared SSE parsing extracted in `api/maps.ts` (`parseSSEBody` / `streamChatSSE`); `streamChatMessage` and `streamGenerateMap` now reuse it instead of carrying duplicate parsers.

## API / schema / config impacts

- New endpoint `POST /ai/chat/dataset/stream/` + `DatasetChatRequest` schema → `backend/openapi.json`, SDKs (`sdks/python`, `sdks/typescript`), and `frontend/src/types/api.generated.ts` regenerated.
[...]

PR #533, opened 2026-07-16
feat(ai): carry the dataset-chat query result into the builder

ishiland:
## Summary

Dataset-chat v1 (#531) deferred one piece of the chosen results UX: "Open in builder" created a map and staged the dataset, but dropped the query result the user was looking at. This completes it.

- **`frontend/src/lib/chat-result-handoff.ts`** (new): one-shot sessionStorage handoff (`geolens-chat-result`). `toChatResultHandoff()` validates the `show_query_result` spatial payload with the same guard as the viewer flyover (finite, WGS84-bounded, non-inverted bbox — the #527 B-054/C-06 rule); `stashChatResult`/`takeChatResult` are exception-safe, so quota overflow on a large FeatureCollection or private-mode storage degrades to opening the builder without the result.
- **`DatasetChatPanel`**: keeps the spatial payload per assistant message; "Open in builder" stashes it and appends `&chat_result=1` only when the stash succeeded.
- **`MapBuilderPage`**: a pickup effect gated on the `mapInstance` state (not `mapData`, so it never races map creation) takes the stash once, deletes the URL param, and renders through the existing `use-ephemeral-layers` path — same dashed-outline overlay + fitBounds the viewer chat uses, including its style.load re-add behavior.

No backend, i18n, or API-contract changes.

## Verification

- `npx vitest run` — handoff lib (12 tests incl. bbox rejection matrix, one-shot take, quota fallback), DatasetChatPanel (7 tests incl. carried vs non-spatial navigation), full `src/components/builder` suite (1832), MapBuilderPage header-actions (5): all green.
- `npx eslint` on changed files, `npm run typecheck`: clean.

### DIFF
--- a/frontend/src/components/dataset/DatasetChatPanel.tsx
+++ b/frontend/src/components/dataset/DatasetChatPanel.tsx
@@ -9,6 +9,7 @@ import { streamDatasetChatMessage } from '@/api/maps';
 import { useCreateMap } from '@/hooks/use-maps';
 import { useAIAvailability } from '@/hooks/use-ai-availability';
 import { QueryResultTable, type QueryResult } from '@/components/viewer/ViewerChatPanel';
+import { stashChatResult, toChatResultHandoff, type ChatResultHandoff } from '@/lib/chat-result-handoff';
 import { cn } from '@/lib/utils';
 import type { ChatAction, ChatHistoryMessage } from '@/types/api';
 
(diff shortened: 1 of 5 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #534, opened 2026-07-16
feat(ai): add a result sanity-check stage to the chat system prompts

ishiland:
## Summary

Adopts the strongest pattern from the geosql evaluation (dataset-chat follow-up, #531): the model self-validates a `query_data` result before presenting it.

New `QUERY_RESULT_SANITY_PROMPT` block in `chat_constants.py`, injected into **both** chat surfaces (map chat in `chat_service.py`, dataset chat in `chat_dataset.py`) after their shared "Query Data Responses" section:

- Suspicious **empty result** → retry once with case-insensitive/fuzzy matching before reporting "no results" (catches misspelled/miscased filter values).
- **Magnitude check** → a count above the layer's feature count, an implausible area/distance, or identical values in every row → retry once spelling out the expected units (catches degree-vs-meter and wrong-column errors).
- Retry still wrong → present the result **with the doubt stated**, never silently.

Bounded at one retry per failure mode so a confused model can't burn the user's token budget in a loop.

Prompt-only change — no tool, schema, or API surface changes. The facade stays under its `test_layering` line cap (410 → 412 of 446).

## Verification

- `pytest tests/test_chat_narrative.py tests/test_ai_chat_dataset.py tests/test_layering.py` — 51 passed (includes two new prompt assertions, one per surface).
- `ruff check` / `ruff format --check` clean.

Note: assertion-based AI evals for behaviors like this are a separate queued follow-up — this PR is the prompt half of that geosql takeaway.

https://claude.ai/code/session_01NXiEUdCwnuPRDKsbbkPWbg

PR #537, opened 2026-07-16
test(ai): assertion-based NL→SQL eval harness

ishiland:
## Summary

The remaining half of the geosql takeaway (#534 shipped the prompt half): the repo's **first AI regression evals**. `backend/tests/evals/test_sql_generation_evals.py` runs five golden questions through the REAL pipeline — `generate_sql()` against the configured LLM provider, then `validate_and_execute()` through the production sandbox — with two assertion layers per case:

1. **SQL structure** — e.g. an acres question must cast `::geography` (the degrees/meters trap), a top-N question must `LIMIT 3`.
2. **Executed results vs ground truth** — computed from the seeded rows by direct SQL (exact count, expected name sets, 10% tolerance on area magnitude, correct superlative/ordering).

Every case also implicitly asserts the sandbox accepts the generated SQL — model drift into disallowed constructs fails loudly.

## Mechanics

- **Opt-in**: skipped unless `RUN_AI_EVALS=1` (5 live light-model calls ≈ cents). Normal `pytest`/CI runs show 5 skips.
- **`make ai-evals`** target added (sources `.env.test` when present; needs the dev DB and the provider key). AGENTS.md commands list updated.
- **Fixture hygiene**: per-test unique `data.eval_parks_*` table; teardown removes the data table AND the catalog Dataset/Record rows, so the shared dev DB stays clean.
- Deterministic-leaning: `generate_sql` runs at temperature 0.0; assertions leave the model freedom where it legitimately has it (aliases, column order, SQL shape).

## Verification

- `make ai-evals` against live Anthropic: **5 passed** (twice).
- Without the flag: 5 skipped in 0.2s.
- `ruff check` / `ruff format --check`: clean.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -11,6 +11,7 @@ The React/Vite frontend is in `frontend/src/`: `components/`, `pages/`, `hooks/`
 - `make dev` / `make down`: start or stop the Docker Compose development stack.
 - `make migrate`: run Alembic migrations in the API container. `make alembic-check` fails if the ORM models have drifted from the migration scripts (run it for schema-adjacent changes).
 - `make test` / `make test-cov`: run backend pytest and coverage.
+- `make ai-evals`: live-provider NL→SQL regression evals in `backend/tests/evals/` (skipped in normal test runs; costs provider tokens, needs `ANTHROPIC_API_KEY` and the dev DB).
 - `npm run e2e` or `npm run e2e:smoke`: run Playwright suites.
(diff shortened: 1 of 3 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #538, opened 2026-07-16
fix(sandbox): allow AND/OR in compound conditions; add spatial eval cases

ishiland:
## Summary

Extending the #537 eval harness immediately paid off: the new distance case generated a `JOIN ... ON a AND b` and exposed a **production sandbox bug** — sqlglot models `AND`/`OR` (`exp.Connector`) as `Func` subclasses, so the SEC-025 fail-closed function walk rejected **every compound condition** (`WHERE x AND y`, `JOIN ... ON ... AND ...`, `HAVING ... AND`, `CASE WHEN ... AND`) as "Query uses a disallowed function". Any AI chat question needing more than one condition failed at validation.

### Sandbox fix (`validator.py`)
- `exp.Connector` nodes are skipped in the function walk — they are boolean operators, not callables. Their operands are still visited by `find_all`, so a blocked/unlisted function inside either side of an AND/OR is still rejected (fail-closed posture unchanged; red-team cases prove it).
- The walk moved to `_check_function_allowlist()` to stay under ruff's C901 cap; logic otherwise byte-identical.

### Tests
- `test_sec025_sandbox_allowlist.py`: new `TestBooleanOperatorsAllowed` — 5 allowed operator shapes + 3 red-team rejections (`pg_sleep`/`dblink`/unlisted `ST_*` inside AND/OR operands). Pure, no LLM.
- `tests/evals/`: two new live cases — **distance-in-miles** (cross-table, geography cast + magnitude vs ground truth; the regression eval for this fix) and **spatial containment** (point-in-polygon join must resolve to exactly the one station seeded inside the park; requires an `ST_*` predicate). Fixture gains a stations point table, second catalog dataset, and full teardown.

## Verification

- `pytest tests/test_sec025_sandbox_allowlist.py tests/test_sandbox.py tests/test_ai_chat_dataset.py`: 157 passed.
- `RUN_AI_EVALS=1` live vs Anthropic: **7/7 passed** (was 6/7 — distance failed on the sandbox rejection before the fix).
- `ruff check` / `format --check`: clean.

### DIFF
--- a/backend/app/platform/sandbox/validator.py
+++ b/backend/app/platform/sandbox/validator.py
@@ -307,6 +307,54 @@ def _reject_recursive_cte(stmt: exp.Expression, sql: str) -> None:
         raise SandboxError("invalid_query", "Recursive queries are not allowed")
 
 
+def _check_function_allowlist(stmt: exp.Expression, sql: str) -> None:
+    """Fail-closed function check (SEC-025).
+
+    For each Func node in the AST, extract its canonical lowercase name:
+      • Anonymous nodes  → fn.name.lower()  (the raw SQL identifier)
+      • Named Func nodes → fn.sql_name().lower()  (sqlglot's canonical name,
+        which may differ from the SQL keyword — see _ALLOWED_FUNCTIONS docs)
+
+    Then apply fail-closed logic (order matters):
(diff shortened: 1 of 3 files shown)


ishiland:
@codex review

PR #542, opened 2026-07-16
fix(ai): dataset-chat smoke follow-ups — overlay idle retry, viewer result badge, e2e carryover

ishiland:
## What

Post-merge browser smoke of the dataset-chat work (#532–#540) surfaced three issues on `main`; this PR closes all of them plus a UI-consistency cleanup.

### 1. Chat-result overlay lost to a style-load race (commit 1)

The `?chat_result=1` cold-load pickup shipped in #533 races the builder's dataset-layer sync. At pickup time `useEphemeralLayers` can observe `isStyleLoaded() === false` while the initial `style.load` event has already fired — so neither the immediate-add path nor the `style.load` subscription ever runs. The stash is consumed and the "Query result · N features" badge renders, but the map keeps no overlay and never fits the result bbox. Reproduced live (map probed: no `ephemeral-result` source/layers despite the badge).

**Fix:** one-shot `idle` retry when the style isn't loaded at apply time — the same idiom as BuilderMap's SP-03 layer-sync gate. The existing `zoomed` flag keeps the bbox fit single-shot; cleanup unsubscribes both listeners. Verified live after the fix (overlay + camera fit present).

### 2. Viewer overlay had no badge, legend entry, or dismissal (commit 2)

`ViewerChatPanel` shares `useEphemeralLayers` but only wired `handleQueryResult` — a viewer-role user got orange highlights with nothing naming them and no way to clear them. Reuse the builder's `EphemeralBadge` (new optional `className`; the viewer's bottom-left column is occupied by BasemapToggle at `bottom-8` and the Map data button at `bottom-20`, so the badge sits at `bottom-32` — clearance measured live, no overlap). Dismiss verified live: badge and overlay both clear.

### 3. E2E carryover coverage (commit 2)

`e2e/dataset-chat.spec.ts` gains the open-in-builder carryover flow: canned SSE with a spatial `show_query_result`, click through, then assert **both** the result badge (React state) and the camera fit via the coordinate readout (map state). Honest caveat, also noted in-spec: the race in item 1 is timing-dependent and does not reproduce in a fast local run (validated by swapping the pre-fix hook back in — the e2e still passed), so the deterministic race regression lives in the unit test; the e2e pins the stash → pickup → overlay integration that jsdom cannot.

### 4. Stale "Experimental" badge removed (commit 3)

The builder rail header and map-create AI tab still carried an Experimental badge while the two newer chat surfaces (dataset page, viewer) shipped without one. Removed for consistency, along with the now-unused `experimentalBadgeColor` token and `chat.experimental` keys in all four locales (`npm run test:i18n` green).

## Verification

- `use-ephemeral-layers` 9/9 (2 new: idle-retry race repro, idle unsubscribe on dismiss)
- `ViewerChatPanel` 6/6 (new: badge render + dismiss wiring)
- Full builder + maps suites: 87 files, 1296 tests green
- `npm run test:i18n` green after locale-key removal
- `npx playwright test e2e/dataset-chat.spec.ts` 3/3 against the live worktree stack
- eslint + `npm run typecheck` clean
- Live QA on all three chat surfaces (dataset panel, builder, viewer) with real Anthropic

### DIFF
--- a/e2e/dataset-chat.spec.ts
+++ b/e2e/dataset-chat.spec.ts
@@ -63,6 +63,41 @@ const SSE_FRAMES = [
   '',
 ].join('\n');
 
+// Spatial result for the builder-carryover test (#533/#542). The bbox centers
+// at exactly (40.75, -73.95) — values chosen to round cleanly in the builder's
+// coordinate readout so the camera-fit assertion is deterministic.
+const CARRYOVER_BBOX = [-74.0, 40.7, -73.9, 40.8];
+const CARRYOVER_GEOJSON = {
+  type: 'FeatureCollection',
+  features: [
+    { type: 'Feature', geometry: { type: 'Point', coordinates: [-73.98, 40.72] }, properties: { name: 'Alpha' } },
+    { type: 'Feature', geometry: { type: 'Point', coordinates: [-73.92, 40.78] }, properties: { name: 'Beta' } },
+  ],
+};
+const CARRYOVER_SSE_FRAMES = [
+  'event: token',
+  'data: {"text": "Found 2 stations."}',
+  '',
+  'event: actions',
+  `data: ${JSON.stringify({
+    actions: [
+      {
+        type: 'show_query_result',
+        rows: [['Alpha'], ['Beta']],
+        columns: ['name'],
+        row_count: 2,
+        geojson: CARRYOVER_GEOJSON,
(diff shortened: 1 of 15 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #552, opened 2026-07-17
ci: run the live NL→SQL evals on the nightly cron

### LINKED ISSUE #548: CI: run the live NL→SQL evals (make ai-evals) on the nightly cron
## What

`backend/tests/evals/` (added in #537, extended in #538) runs 7 assertion-based NL→SQL evals against the live provider and the production sandbox. They are gated behind `RUN_AI_EVALS=1` and cost provider tokens, so they don't run in PR CI — but nothing runs them on a schedule either, so prompt/model drift would go unnoticed until a user hits it.

`ci.yml` already has a nightly `cron: "0 7 * * *"` trigger; adding a job that starts the db service, seeds via the eval fixtures (self-contained — they create and drop their own tables), and runs `make ai-evals` is a small change.

## Blocker

Requires an `ANTHROPIC_API_KEY` repository secret (Actions) — not yet provisioned.

## Related

Chat-loop behavioral evals (asserting on full chat-stream behavior rather than `generate_sql`) are deferred separately: the Anthropic main-loop path omits temperature, so behavioral assertions would flake. The eval value proved itself immediately — the harness's distance case exposed the sandbox AND/OR bug fixed in #538.

ishiland:
## What

Adds an `ai-evals` job to `ci.yml`, gated to `schedule`/`workflow_dispatch` only — the same trigger pattern as `e2e-test`. Closes #548.

The eval suite (`backend/tests/evals/`, #537/#538) calls the **real Anthropic API** and executes the generated SQL through the production sandbox, so it stays out of push/PR CI (token cost; fork PRs get no secrets anyway). Nightly it gives a drift signal within a day — this suite is what caught the sandbox AND/OR bug (#538), so the alarm has already proven it earns its keep.

## How

- Setup mirrors the `backend-test` job exactly (custom PostGIS+pgvector container, extensions/roles/schemas, `alembic upgrade heads`), then runs `uv run pytest tests/evals/ -v` with `RUN_AI_EVALS=1`.
- Provider config needs nothing beyond the `ANTHROPIC_API_KEY` secret (provisioned): `LLM_PROVIDER`/`LLM_MODEL_LIGHT` resolve to anthropic/haiku via their `env_default` factories.
- `timeout-minutes: 30` so a hung provider call can't burn Actions minutes.
- No `needs:` — independent nightly signal, ~7 live cases ≈ a few thousand Haiku tokens per run.

## Verification

- YAML validated (`yaml.safe_load` — job parses with 9 steps).
- Post-merge: I'll trigger a `workflow_dispatch` run to prove the job green end-to-end without waiting for the 07:00 UTC cron.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1320,6 +1320,107 @@ jobs:
         if: always()
         run: docker compose down
 
+  # ---------- nightly live AI evals ----------
+  # fix(#548): the assertion-based NL→SQL evals (backend/tests/evals/, added in
+  # #537/#538) call the REAL Anthropic API and execute the generated SQL through
+  # the production sandbox. They cost provider tokens, so they run ONLY on
+  # schedule/workflow_dispatch — never on push/PR (fork PRs receive no secrets
+  # anyway, so they could never run there). A failure here signals prompt/model/
+  # sandbox drift, not ordinary flake: generate_sql pins temperature 0.0 and the
+  # assertions leave the model freedom where it legitimately has it. Investigate
+  # before rerunning.
+  ai-evals:
+    name: AI Evals (live provider)
+    if: github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
+    runs-on: ubuntu-latest
+    # Live-API job: cap it so a hung provider call can't burn Actions minutes.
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review

PR #555, opened 2026-07-17
ci: detect changes via git diff, not the GitHub API

### LINKED ISSUE #546: CI: 'Detect Changes' required check fails hard on transient GitHub API errors
## Problem

`Detect Changes` (dorny/paths-filter) is a **required** check that every other job gates on, and it calls the GitHub API to list PR changed files. When the API has a bad moment it dies in seconds with the raw 503 HTML page as the error — and because it is required, the whole PR goes red and auto-merge stalls.

On 2026-07-16 (evening) this happened **three times** on a single PR (#542) during a GitHub instability window; each needed a manual `gh run rerun <run-id> --failed`.

## Proposal

Give the step flake tolerance, e.g.:
- wrap `dorny/paths-filter` in a retry (a `nick-fields/retry` step or a small retry loop), or
- fall back to `git diff --name-only origin/${{ github.base_ref }}...HEAD` computed locally from the checkout when the API call fails — the checkout is already present in the job.

Either keeps a genuine failure loud while absorbing 5-second infra 503s.

ishiland:
## Change

Passes `token: ''` to `dorny/paths-filter` in the required **Detect Changes** job. With an empty token the action (verified in the pinned v4 source) detects PR changes locally via `git diff base.sha..merge-commit` from the existing checkout instead of calling the GitHub API to list changed files — the call whose transient 503s failed the check hard three times on #542 and stalled auto-merge.

Semantics are identical: on `pull_request` events the checked-out merge commit is built on `base.sha`, so the two-dot diff yields exactly the PR's changed files. Push/schedule paths were already git-based. Also drops the job's now-unneeded `pull-requests: read` permission.

## Verification

`pull_request` workflows run the PR's own ci.yml, so the Detect Changes run on this PR exercises git-mode detection directly — confirm its log shows "changes will be detected using git diff" and the filter outputs fire (this PR should trigger no backend/frontend jobs, only the always-on ones).

Closes #546

https://claude.ai/code/session_01XPiD2rhh1eNmYnPRojE39j

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,8 +17,8 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
-# Least-privilege default for every job. The `changes` job widens to add
-# pull-requests:read; no job needs write (publishing lives in publish*.yml).
+# Least-privilege default for every job; no job needs write (publishing lives
+# in publish*.yml).
 permissions:
   contents: read
 
@@ -27,9 +27,6 @@ jobs:
   changes:
     name: Detect Changes
     runs-on: ubuntu-latest
-    permissions:
-      contents: read
-      pull-requests: read
     outputs:
       backend: ${{ steps.filter.outputs.backend }}
       frontend: ${{ steps.filter.outputs.frontend }}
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review

PR #556, opened 2026-07-17
fix(ai): carry geometry into chat query results deterministically

### LINKED ISSUE #544: AI chat: geometry should reach query results deterministically (and raw WKB should not render in result tables)
## Problem (two symptoms, one root)

**1. Spatial questions often produce no map overlay.** When a user asks a location-shaped question ("list the names and locations of X"), the SQL model is free to select attribute columns (e.g. `longitude`/`latitude`) instead of `geom_4326`. Without a geometry column in the result set, `_extract_geojson` returns nothing, the `show_query_result` action carries no geojson, and every geometry-dependent surface silently degrades: no highlight overlay in builder/viewer chat, no result-layer carryover from the dataset panel's "Open in builder" (#533). The system prompt in `backend/app/processing/ai/sql_generator.py` has no rule about when to include `geom_4326` in the SELECT.

**2. When geometry IS selected, the chat result table renders raw WKB hex** (`0104000020E6...`) as a column on all three chat surfaces — noise for the user and wasted panel width.

## Proposal

- Prefer a **deterministic server-side** fix over a prompt rule: ensure geometry is available to `_extract_geojson` regardless of the model's column choice (e.g. append/queryize the geometry column server-side for feature-listing results), so behavior doesn't vary by provider/model.
- Frontend: exclude geometry columns from the rendered result table (they remain in the geojson payload).
- Add an NL→SQL eval case (`backend/tests/evals/`) pinning the "list the locations of X" shape.

## Context

Found during the post-merge browser smoke of the dataset-chat work (#531, #533, #542). The alternative (a system-prompt rule instructing the model to include `geom_4326` for feature/location questions) is cheaper but model-dependent and needs the eval gate either way.

ishiland:
Fixes #544.

## Problem

Two symptoms, one root (found in the post-merge browser smoke of the dataset-chat work):

1. **Spatial questions often produced no map overlay.** The SQL model could answer "list the names and locations of X" with attribute columns only (e.g. `longitude`/`latitude`), so `_extract_geojson` found nothing, `show_query_result` carried no geojson, and every geometry-dependent surface silently degraded (builder/viewer highlight overlay, dataset-panel "Open in builder" result-layer carryover).
2. **When geometry WAS selected, raw WKB hex rendered** in the chat result tables on all three surfaces.

## Approach (server-side, deterministic — no prompt rule)

- **`ensure_geometry_selected()`** (`chat_geojson.py`): a sqlglot pass over the generated SQL that appends the layer's `geom_4326` when the query is a plain single-table row-level SELECT from a geometry layer. Conservative by design — aggregates (incl. anonymous ones like `EVERY`), `GROUP BY`, `DISTINCT`, joins, CTEs, set operations, `SELECT *`, and already-selected geometry all leave the SQL unchanged. Runs before sandbox validation, so the appended column goes through the normal validate/execute path.
- **`strip_geometry_columns()`**: once the geojson payload is extracted, the geometry column is dropped from the tabular tool result. WKB never reaches the result tables (all three chat surfaces get this for free — no frontend change needed) and stops wasting model-context tokens.
- **Eval**: new live NL→SQL eval case pins the "list the name and location of every station" shape end-to-end through `generate_sql` → append pass → sandbox → `_extract_geojson` (verified locally against the real provider: passed).

## Notes

- `chat_actions.py` line-cap 400 → 425 in `test_layering.py` (reviewed bump, +14 lines of logic in `_handle_query_data`).
- No OpenAPI/SDK impact — `show_query_result` action shape is unchanged.
- Skipped the issue's optional system-prompt rule: the server-side rewrite covers the single-table case deterministically; joins/aggregates keep today's behavior.

## Verification

- `uv run pytest tests/test_ephemeral_geojson.py tests/test_sql_engine.py tests/test_layering.py` — 120 passed (18 new unit tests + 1 mocked integration test).
- `RUN_AI_EVALS=1 uv run pytest tests/evals/...::test_locations_reach_geojson_deterministically` — passed against the live provider.
- `ruff check` / `ruff format --check` clean; pre-commit hooks passed.

https://claude.ai/code/session_01XPiD2rhh1eNmYnPRojE39j

### DIFF
--- a/backend/app/processing/ai/chat_actions.py
+++ b/backend/app/processing/ai/chat_actions.py
@@ -44,6 +44,13 @@ logger = structlog.stdlib.get_logger(__name__)
 
 _DEFAULT_LABEL_TEXT_COLOR = "#333333"
 
+# Overlay/table render budget: only this many rows reach the chat result table
+# and the ephemeral map overlay. fix(#556 review P2): when we append geom_4326
+# to an otherwise attribute-only query, cap the fetch to this budget so a
+# large polygon/line layer doesn't transfer up to 1000 full geometries just to
+# discard all but these.
+_OVERLAY_ROW_BUDGET = 50
+
 
 def _safe_label_text_color(value: object) -> str:
     # fix(#394) CH-02: hex / real CSS named color / numeric-arg functional form
@@ -139,17 +146,27 @@ async def _handle_query_data(
 
     # fix(#544): geometry must reach _extract_geojson regardless of which
     # columns the model chose, or every map surface silently loses its overlay.
-    sql = ensure_geometry_selected(sql, layers)
+    sql_with_geom = ensure_geometry_selected(sql, layers)
+    geom_appended = sql_with_geom != sql
+    sql = sql_with_geom
 
     if stage_callback:
         stage_callback("Running query...")
 
-    result = await chat_service.validate_and_execute(
-        sql, session, user, restrict_tables=restrict_tables
-    )
+    # fix(#556 review P2): when we appended geometry, cap the fetch to the
+    # overlay render budget. Only _OVERLAY_ROW_BUDGET rows reach the table and
+    # overlay anyway, so fetching (and holding) up to 1000 full geometries for
+    # a large polygon/line layer is a transfer/memory regression. row_count
+    # then reflects the rendered budget with the truncated flag signalling more
+    # rows exist. Queries where the model selected geometry itself keep the
+    # default limit (pre-existing behavior, out of this rewrite's scope).
+    exec_kwargs: dict = {"restrict_tables": restrict_tables}
+    if geom_appended:
+        exec_kwargs["row_limit"] = _OVERLAY_ROW_BUDGET
+    result = await chat_service.validate_and_execute(sql, session, user, **exec_kwargs)
 
     # Extract GeoJSON for ephemeral result layers
-    columns, rows = result.columns, result.rows[:50]
+    columns, rows = result.columns, result.rows[:_OVERLAY_ROW_BUDGET]
     geojson_result = _extract_geojson(columns, rows)
     if geojson_result is not None:
         # fix(#544): geometry now travels via geojson only — raw WKB in the
--- a/backend/app/processing/ai/chat_actions.py
+++ b/backend/app/processing/ai/chat_actions.py
@@ -21,7 +21,11 @@ from sqlalchemy.ext.asyncio import AsyncSession
 from app.core.identity import Identity
 from app.platform.sandbox import SandboxError
 from app.processing.ai.chat_constants import _EDIT_TOOLS, ERROR_MESSAGES
-from app.processing.ai.chat_geojson import _extract_geojson
+from app.processing.ai.chat_geojson import (
+    _extract_geojson,
+    ensure_geometry_selected,
+    strip_geometry_columns,
+)
 from app.processing.ai.colors import is_css_colorish, label_halo_color
 from app.processing.ai.chat_styles import _build_data_driven_style
 from app.processing.ai.schemas import (
@@ -133,17 +137,31 @@ async def _handle_query_data(
         error_msg = sql.strip().removeprefix("-- ERROR:").strip()
         return {"error": error_msg, "category": "llm_cannot_answer"}
 
+    # fix(#544): geometry must reach _extract_geojson regardless of which
+    # columns the model chose, or every map surface silently loses its overlay.
+    sql = ensure_geometry_selected(sql, layers)
+
     if stage_callback:
         stage_callback("Running query...")
 
     result = await chat_service.validate_and_execute(
         sql, session, user, restrict_tables=restrict_tables
     )
+
+    # Extract GeoJSON for ephemeral result layers
+    columns, rows = result.columns, result.rows[:50]
+    geojson_result = _extract_geojson(columns, rows)
+    if geojson_result is not None:
+        # fix(#544): geometry now travels via geojson only — raw WKB in the
+        # tabular payload is token noise for the model and renders as hex in
+        # the chat result tables.
+        columns, rows = strip_geometry_columns(columns, rows)
+
     # Limit rows in tool result for token economy
     # Note: raw SQL intentionally excluded to prevent info disclosure via LLM leakage
     out: dict = {
-        "columns": result.columns,
-        "rows": result.rows[:50],
+        "columns": columns,
+        "rows": rows,
         "row_count": result.row_count,
         "truncated": result.truncated,
     }
@@ -151,9 +169,6 @@ async def _handle_query_data(
         out["note"] = (
             "No matching results found. The user may want to try different criteria."
         )
-
-    # Extract GeoJSON for ephemeral result layers
-    geojson_result = _extract_geojson(result.columns, result.rows[:50])
     if geojson_result is not None:
         out["geojson"] = geojson_result[0]
         out["bbox"] = geojson_result[1]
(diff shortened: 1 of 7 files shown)


ishiland:
@codex review

ishiland:
@codex review — pushed a second commit (114645dd0): live browser smoke caught an aliased `ST_AsGeoJSON(geom_4326) AS location` column surviving the name-based strip; stripping is now value-based.

ishiland:
@codex review — both P2s addressed in d0c4a169c (shared root: name-only, rows[0]-only detection). `_detect_geom_column` gained a value-based fallback so aliased computed geometry (`ST_Buffer(...) AS buffer`) drives the overlay itself, `ensure_geometry_selected` skips the append for aliased geometry expressions, and detection/stripping probe the first non-null value per column so NULL leading geometries neither kill the overlay nor leak WKB into the table. Live-verified: a 500m-buffer question now renders the buffer polygons and a clean table.

ishiland:
@codex review — round-2 P2s addressed in c873706da: alias unwrap now handles `exp.Paren` alongside `exp.Cast`, and the value-based fallback requires candidates to actually parse (shapely WKB / GeoJSON shape) so hex-like attributes such as `md5(name) AS id` can't shadow the real geometry column.

ishiland:
@codex review — round-3 P2s addressed in 3f8b3d7d6:

1. **Parse-verified detection.** Consolidated to a single strict `_is_geom_value` used by *both* the name-preferred and value-fallback phases (deleted the redundant `_is_geometry_cell`). A geometry-*named* column holding a non-WKB hex (or an LLM alias like `md5(x) AS geometry`) no longer shadows the real `geom_4326` — it must actually parse as WKB/GeoJSON.
2. **Bounded fetch.** When (and only when) we append `geom_4326`, the fetch is capped to the overlay render budget (`_OVERLAY_ROW_BUDGET = 50`), so a large polygon/line layer no longer transfers up to 1000 full geometries to discard all but 50; `row_count`/`truncated` reflect the budget. Model-selected-geometry queries keep the default limit (pre-existing behavior, out of the rewrite's scope).

Live-verified: 'list every station with its location' (496 rows) appends geometry, caps at '50 rows (showing a sample)', strips WKB from the table, and the builder overlay renders 50 features with camera fit. 192 unit tests + the live NL→SQL eval pass.

ishiland:
@codex review — round-4 P2 addressed in 7af89629e. `_selects_geometry` (the append gate) no longer trusts the alias name: it unwraps the alias (and Cast/Paren) and decides purely on the underlying expression — an aliased `geom_4326` column or a geometry-returning function. So `md5(name) AS geometry` and `ST_X(geom_4326) AS st_x` now correctly trigger the append (scalar-under-geometry-name no longer masquerades as selected geometry), while `ST_Buffer(...)::geometry AS buffer` still suppresses it. Added tests for both scalar-alias cases; 136 unit + live eval green.

ishiland:
@codex review — round-5 P2 addressed in 150cbaedf. The Cast/Paren unwrap in `_selects_geometry` now runs for every select item, not just aliased ones: unwrap `exp.Alias`, then `exp.Cast`/`exp.Paren`, before the Column / geometry-function check. So an unaliased `ST_Buffer(geom_4326::geography, 1000)::geometry` correctly suppresses the append (no redundant second geometry column, no source WKB leaking into feature properties), matching the aliased-cast behavior. Added 3 tests (unaliased cast, unaliased paren+cast, unaliased cast-of-geom-column); 139 unit + live eval green.

ishiland:
@codex review — round-6 P2 addressed in 13ab0133f. The appended geometry qualifier is now built from the alias/table AST identifier (copied so its `quoted` flag survives) and passed as a real `exp.Column` to `stmt.select()`, instead of an f-string. Quoted aliases round-trip correctly now — `FROM data.parks AS "P"` → `"P".geom_4326` (not lowercase-folding `P`), and a spaced alias no longer raises a ParseError inside `stmt.select()`. Unquoted aliases render identically (`s.geom_4326`, `parks.geom_4326`). Added tests for quoted-uppercase and spaced aliases; 141 unit + live eval green.

ishiland:
@codex review — round-7, two P2s addressed in 6d9c1b2b8:

**(A) row_count preserved under the transfer cap.** The cap corrupted the documented total, so when it truncates I now recover the true total via a geometry-free `SELECT COUNT(*) FROM (<sql>) _geolens_total` (best-effort; bounded by the query's own LIMIT). Transfer size and reported total are decoupled — 'list every station' reports 496, not 50, while still transferring only 50 geometries.

**(B) window aggregates append geometry.** `COUNT(*) OVER ()` / `RANK() OVER (...)` are row-level; the guard now skips only *non-windowed* aggregates (`find_ancestor(exp.Window) is None`). Exposed a latent bug: the `SELECT *` bail used `item.find(exp.Star)`, which also matched the star inside `COUNT(*)` — narrowed to a genuine all-columns star (bare `*` / `table.*`).

Added tests for window count/rank, count-recovery (fires only when truncated), and the star narrowing. 172 unit + live eval green.

ishiland:
@codex review — round-8 P2 addressed in 63f77c8a9. The `_ANON_AGG_NAMES` check is now gated to `exp.Anonymous` nodes, where `fn.name` is the function name — so `CAST(mode AS TEXT)` (exp.Cast, `name="mode"`) no longer trips the aggregate guard and a row-level query casting a column named `mode`/`every`/`percentile_*` keeps its overlay. Verified via probe that only `EVERY` parses as Anonymous; `MODE`/`PERCENTILE_CONT`/`PERCENTILE_DISC` are `exp.AggFunc` subclasses caught by isinstance, so all four remain blocked. Added tests for the CAST(column-named-mode) case and the ordered-set aggregate. 147 unit + live eval green.

ishiland:
@codex review — round-9 P2 addressed in 5af1eaa9c. `ensure_geometry_selected` now bails when `stmt.args.get("having")` is set: a HAVING without GROUP BY is an implicit single-group aggregate, so appending `geom_4326` would make Postgres reject the non-grouped column. `args.get` scopes to the outer SELECT, so a subquery's HAVING still allows the append on a row-level outer query (covered by test). 149 unit + live eval green.

ishiland:
@codex review — round-10 P2 addressed in b83c58320. Confirmed real: a SRID-less ingest keeps `geometry_type` but exposes only native `geom` (documented at `ensure_geom_4326_gist_index`), so the append 42703'd an otherwise-valid attribute query.

`ChatMapLayer` carries no geometry-column name and a schema probe would add multi-tenant complexity to every call, so I guard at execution instead: if the appended query raises `SandboxError`, retry the model's original SQL once — attribute rows still return (overlay dropped), and if the original also fails its error propagates unmasked. Zero overhead on the happy path (fallback runs only on actual failure), and it's tenancy-agnostic. Added tests for the fallback and for the no-retry-on-non-appended-failure case. 151 unit + live eval green.

ishiland:
@codex review — round-11 P2 addressed in ebc996dc0. Confirmed via probe: sqlglot's postgres dialect re-serializes `<=>` (pgvector cosine) as `IS NOT DISTINCT FROM` on round-trip, corrupting nearest-neighbor ranking; `<->` round-trips faithfully; `<#>` fails parse_one (already caught by the try/except → append skipped).

Since re-rendering only happens on the append path, I added a preservation guard: after rendering, if any `<=>`/`<->`/`<#>` operator the original had was dropped, return the untouched SQL (overlay sacrificed to keep the query's meaning). Count-based rather than an operator denylist, so it catches future round-trip regressions too. Added tests: cosine query preserved (append skipped, no `IS NOT DISTINCT FROM`), and `<->` KNN still gains its overlay. 153 unit + live eval green.

Issue #558, opened 2026-07-17
fix(ai): SQL prompt advertises pgvector <#> operator the sandbox validator can't parse


ishiland:
## Problem

The NL→SQL system prompt advertises the pgvector negative-inner-product operator `<#>` as a valid operator the model may emit:

```
# backend/app/processing/ai/sql_generator.py
  embedding <#> '[...]'::vector   -- negative inner product        (line ~220)
...
- Vector columns (type vector(N)) are queried with the <->/<=>/<#> operators … (line ~379)
```

But `sqlglot.parse` — which `validate_sql` (`backend/app/platform/sandbox/validator.py`) uses as its first stage — **raises `ParseError` on `<#>`**. So if the model follows the prompt and generates an inner-product query, the sandbox rejects it as `invalid_query` before it ever runs, and the user gets a generic "couldn't generate a valid query."

`<=>` (cosine) and `<->` (L2) both parse fine; only `<#>` is affected.

### Evidence

```python
>>> import sqlglot
>>> sqlglot.parse("SELECT id FROM data.t ORDER BY embedding <#> '[1,2,3]'::vector LIMIT 5", dialect="postgres")
sqlglot.errors.ParseError: Required keyword: 'expression' missing for <class 'sqlglot.expressions.core.LT'> ...
>>> # <=> and <-> both parse OK
```

## Impact

Self-inflicted prompt/validator contract mismatch. Low incidence (needs a `vector(N)` column and the model choosing inner-product over cosine/L2), but when it happens the query is silently unusable.

## Suggested fix

Simplest: **drop `<#>` from the prompt** (remove the operator line + the mention in the constraints list). Cosine (`<=>`) and L2 (`<->`) cover the practical similarity use cases. (Note `<=>` has its own multi-tenant re-render issue — see the sibling sandbox issue.)

Alternatively, teach the validator/executor to handle `<#>` if inner-product ranking is genuinely wanted, but that's more work than the operator warrants.

_Surfaced during the #556 (#544) review. Pre-existing._

ishiland:
Fixed by #559 (merged as 1c0bd58b6): `<#>` removed from the SQL prompt, plus a prompt/validator contract test that fails if an unparseable vector operator is re-advertised. (Auto-close missed this one — the PR body's "Fixes #557 and #558" only closes the issue right after the keyword.)

PR #559, opened 2026-07-17
fix(sandbox/ai): preserve pgvector <=> in multi-tenant rewrite; drop unparseable <#> from prompt (#557, #558)

### LINKED ISSUE #557: fix(sandbox): multi-tenant schema rewrite corrupts pgvector <=> operator via sqlglot re-render
## Problem

In multi-tenant mode, `execute_safe` rewrites the logical `data.*` schema to the per-tenant physical schema by **re-serializing the parsed AST**:

```python
# backend/app/platform/sandbox/executor.py
def _rewrite_logical_data_schema(sql, physical_schema):
    statement = sqlglot.parse_one(sql, dialect="postgres")
    ...
    return statement.sql(dialect="postgres")   # <-- full re-render
```

This runs on **every** sandbox query when `is_multi_tenant()` is true (`execute_safe`, ~line 84–92).

sqlglot's postgres dialect does **not** faithfully round-trip the pgvector cosine operator: `embedding <=> '[...]'::vector` re-renders as `embedding IS NOT DISTINCT FROM CAST('[...]' AS VECTOR)` — silently turning nearest-neighbor ranking into a boolean equality test. So in a **multi-tenant deployment, any semantic-similarity query returns the wrong rows**, with no error.

This is the same sqlglot infidelity fixed for the chat geometry-append path in #556 (round-11 review), but here it lives in the general sandbox execution path and fires independently of chat.

### Evidence

```python
>>> import sqlglot
>>> sqlglot.parse_one("SELECT id FROM data.t ORDER BY embedding <=> '[1,2,3]'::vector LIMIT 5", "postgres").sql("postgres")
"SELECT id FROM data.t ORDER BY embedding IS NOT DISTINCT FROM CAST('[1,2,3]' AS VECTOR) LIMIT 5"
```

`<->` (L2) round-trips fine; `<#>` fails `parse_one` outright (see the sibling issue on the prompt/validator mismatch).

## Scope / demand-gate

Requires **multi-tenant + pgvector/semantic-search both active**. Single-tenant is unaffected — `execute_safe` runs the original SQL string verbatim there (no rewrite). Narrow, but a silent-wrong-data correctness bug when it hits.

## Suggested fix

Either:
- preserve distance operators through the rewrite (mirror the count-based guard added in `chat_geojson.ensure_geometry_selected` in #556 — bail/rewrite-schema-only if a `<=>`/`<->`/`<#>` the original had is dropped), **or**
- narrow the schema rewrite so it does not re-serialize the whole statement (targeted qualifier substitution that leaves operators untouched).

## Related, lower-priority (fold-in)

geom-only datasets (SRID-less ingest keeps `geometry_type` but exposes only native `geom`, no `geom_4326` — see `ensure_geom_4326_gist_index` docstring): #556 added an execution fallback so **attribute** chat queries survive on these, but **geometry/overlay** chat is still broken because the SQL prompt and `build_sql_schema_context` hardcode `geom_4326`. Proper fix = expose the real geometry-column name through ingest metadata → prompt/schema-context. Rare; track as a known limitation.

_Surfaced during the #556 (#544) review. Not launch-blocking; demand-gate against real multi-tenant + pgvector reports._

ishiland:
Fixes #557 and #558 — two latent pgvector correctness bugs from the same root cause (sqlglot's postgres dialect imperfectly round-trips pgvector distance operators), surfaced during the #556 review.

## #558 — prompt advertised `<#>` the validator can't parse
`SQL_SYSTEM_PROMPT` told the model `embedding <#> '[...]'::vector` was valid, but `sqlglot.parse` (the sandbox validator's first stage) raises `ParseError` on `<#>`, so any model-generated inner-product query was rejected as `invalid_query` before running. Cosine (`<=>`) and L2 (`<->`) cover the practical similarity cases, so `<#>` is dropped from the operator reference and the constraints list.

## #557 — multi-tenant schema rewrite corrupted `<=>`
`_rewrite_logical_data_schema` re-serializes the parsed AST to map the logical `data.*` schema to a per-tenant physical schema, on **every** sandbox query in multi-tenant mode. sqlglot mis-parses `<=>` (cosine) as `NullSafeEQ`, so the re-render rewrote it to `IS NOT DISTINCT FROM` — silently turning nearest-neighbor ranking into a boolean equality test and returning the **wrong rows**. Single-tenant runs the original SQL verbatim and is unaffected.

The schema rewrite is mandatory (the logical `data.` schema has no physical table), so unlike the optional geometry-append guard in #556 it can't bail to the original SQL. Instead it swaps `<=>` for a sentinel operator that round-trips (`&&`, which the validator rejects as `array_overlaps`, so validated input can never contain it), rewrites the schema on the AST, then restores the operator. Fails closed if the sentinel is somehow already present.

## Tests
- `test_advertised_vector_operators_pass_the_validator` — scans the prompt and asserts every advertised vector operator survives `validate_sql` (would have caught #558; re-adding `<#>` re-breaks it).
- `test_schema_rewrite_preserves_pgvector_cosine_operator` / `..._l2_and_cosine_together` — DB-free unit tests proving `<=>` (and `<->` alongside) survive the rewrite while the logical schema is still translated.

## Deferred
The geom-only-dataset overlay gap noted in #557 (SRID-less ingests expose native `geom`, not `geom_4326`, which the prompt + `build_sql_schema_context` hardcode) is **not** addressed here — it needs ingest-metadata plumbing, not an operator fix. Tracking it separately; leaving #557 to close on the `<=>` fix.

## Verification
- `pytest tests/test_sandbox_tenant_schema.py tests/test_sql_engine.py tests/test_sandbox.py tests/test_sql_safety.py tests/test_sec025_sandbox_allowlist.py tests/test_ephemeral_geojson.py tests/test_layering.py` — all green
- `ruff check` + `ruff format --check` — clean
- No OpenAPI/SDK impact (prompt text + internal executor only)

### DIFF
--- a/backend/app/platform/sandbox/executor.py
+++ b/backend/app/platform/sandbox/executor.py
@@ -7,6 +7,8 @@ for end users while full details are logged server-side.
 
 from __future__ import annotations
 
+import re
+
 import structlog
 import sqlglot
 from sqlglot import exp
@@ -29,14 +31,34 @@ DEFAULT_TIMEOUT_MS = 10_000
 # boolean equality test. The multi-tenant schema rewrite MUST re-render (the
 # logical `data.` schema has no physical table), so unlike the optional
 # geometry-append guard (#556) it cannot bail to the original SQL. Instead we
-# swap `<=>` for a sentinel operator that DOES round-trip, rewrite the schema on
-# the parsed AST, then restore. `&&` is chosen because the sandbox validator
-# rejects it (unlisted_function: array_overlaps), so validated SQL can never
-# contain it and the restore can't collide with real input. `<->` (L2) already
-# round-trips and `<#>` fails parse upstream, so `<=>` is the only operator that
-# reaches here valid-but-corrupted.
+# swap the `<=>` OPERATOR for a sentinel that DOES round-trip (`&&`, which the
+# validator rejects as array_overlaps so it never appears as a real operator in
+# validated SQL), rewrite the schema on the parsed AST, then restore. Both swaps
+# skip single-quoted string literals, so a literal such as `note = '&&'` or
+# `note = '<=>'` is preserved verbatim (fix(#559) review: a raw substring swap
+# corrupted or rejected such literals). `<->` (L2) already round-trips and `<#>`
+# fails parse upstream, so `<=>` is the only operator that reaches here
+# valid-but-corrupted.
 _COSINE_OP = "<=>"
 _COSINE_SENTINEL = "&&"
+_STRING_LITERAL_RE = re.compile(r"'(?:[^']|'')*'")
+
+
+def _swap_outside_string_literals(sql: str, old: str, new: str) -> str:
+    """Replace ``old`` with ``new`` only outside single-quoted string literals.
+
+    Splitting on the literal pattern keeps operator tokens in the code segments
+    replaceable while literal contents (which may legitimately contain the same
+    characters) pass through untouched.
+    """
+    segments = _STRING_LITERAL_RE.split(sql)
+    literals = _STRING_LITERAL_RE.findall(sql)
+    out: list[str] = []
+    for i, segment in enumerate(segments):
+        out.append(segment.replace(old, new))
+        if i < len(literals):
+            out.append(literals[i])
+    return "".join(out)
 
 
 def _rewrite_logical_data_schema(sql: str, physical_schema: str) -> str:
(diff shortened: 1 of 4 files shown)


ishiland:
@codex review

ishiland:
Addressed the P3: the sentinel swap is now literal-aware (splits on single-quoted string literals and only swaps operator tokens in code segments, on both swap-in and restore), so `WHERE note = '&&' ORDER BY embedding <=> ...` and literal `'<=>'` are preserved verbatim. Dropped the fail-closed collision branch. Added a regression test. @codex review

ishiland:
Addressed round-2 P3: the operator swap is now driven by sqlglot's tokenizer rather than a regex, so `<=>` inside a literal of any quoting form — `'...'`, `E'...'`, `$$...$$` — is classified as a string token and left untouched, while `IS NOT DISTINCT FROM` (keywords, not NULLSAFE_EQ) is never rewritten. Only genuine NULLSAFE_EQ/DAMP operator tokens are swapped, by source span. Regression test added for dollar-quoted literals. @codex review

PR #562, opened 2026-07-17
ci: raise coverage floor to 80 (measured 85.49%)

ishiland:
## Why
The `-n4` Backend Tests run in #561 measured **85.49%** actual backend coverage, but the CI floor was `--cov-fail-under=58.5` and pyproject's `fail_under=60`. The gate tolerated a ~25-point regression before failing — effectively toothless.

## Change
Raise both floors to **80** (aligned):
- `.github/workflows/ci.yml` — `--cov-fail-under=80`
- `backend/pyproject.toml` — `fail_under = 80` (used by `make test-cov`)

80 gates a real regression while leaving a conservative **~5.5-point buffer** below the measured 85.49%, so normal PR churn won't false-RED. Follows the pyproject ratchet note (raise, never lower without a CHANGELOG rationale).

## Self-validating
This PR edits `ci.yml`, which the backend paths-filter (added in #561) now includes — so **this PR's own Backend Tests job runs the full suite under the new 80 floor**. If coverage were below 80, this PR would go red. Watch that job.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -845,10 +845,10 @@ jobs:
             tests/test_layering.py
 
       - name: Run tests with coverage
-        # --cov-fail-under=58.5 is the CI floor. backend/pyproject.toml declares
-        # fail_under=60; the two are intentionally NOT yet aligned here — raising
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

PR #563, opened 2026-07-17
fix(ai): degrade chat gracefully when geom_4326 column is absent (#560)

### LINKED ISSUE #560: fix(ai): geometry/overlay chat broken on geom_4326-less datasets (prompt + schema-context hardcode geom_4326)
Split out from #557's "Related, lower-priority (fold-in)" section so it survives #557 closing on the `<=>` operator fix (#559).

## Problem
SRID-less ingests keep `geometry_type` but expose only the native `geom` column — no `geom_4326` (see the `ensure_geom_4326_gist_index` docstring). #556 added an execution fallback so **attribute** chat queries survive on these datasets, but **geometry/overlay** chat is still broken: `SQL_SYSTEM_PROMPT` and `build_sql_schema_context` (`backend/app/processing/ai/sql_generator.py`) hardcode `geom_4326`, so the model emits `geom_4326` and the query fails / returns no overlay.

## Proper fix
Expose the real geometry-column name through ingest metadata → prompt / schema-context, instead of hardcoding `geom_4326`.

## Scope
Rare (only SRID-less ingests). Not launch-blocking. Demand-gate against real reports.

_Surfaced during the #556 (#544) review; originally filed as a rider on #557._

ishiland:
## What

AI chat geometry/overlay queries against a table that exposes only native `geom` (no reprojected `geom_4326` column) previously failed with a generic *"Something went wrong. Try rephrasing your question."* This makes `_handle_query_data` detect that specific case and return a clean, actionable note instead, so the model explains and offers attribute answers.

The SQL prompt + schema-context always name the geometry column `geom_4326`, so the model emits it. The #556 fallback only rescued *appended-geometry* attribute queries; a query where the **model itself** wrote `geom_4326` (a pure overlay/intersect question, or `geom_4326` in a `WHERE` / `ORDER BY`) hit the `raise` and surfaced the generic failure.

Detection keys on the specific Postgres error `column "geom_4326" does not exist`, which lives on `SandboxError.__cause__` (`SandboxError.category` is the generic `query_failed`). Applied at **both** failure points: the model-written path and the #556 append-retry path. Any other error still propagates unmasked.

## Why this is intentionally minimal (not a full "thread the real column name through" fix)

Research finding: **no current ingest path can produce a `geometry_type`-set-but-no-`geom_4326` dataset.** Unknown-CRS uploads are *rejected* at the missing-CRS gate (not stored geom-less), and every spatial path — ogr2ogr files, GeoParquet, WFS/ArcGIS, x/y & WKT overrides, `created`, reuploads — runs `add_4326_column`. The only real exposure is **legacy pre-convention tables** (ingested before the `geom_4326` convention existed; no backfill migration retrofits them).

Even a full fix couldn't make metric spatial math correct on such tables — the native `geom` keeps a non-4326 SRID (`::geography` distances / 4326-literal joins would still be wrong). So this ships defensive graceful-degradation and the full overlay-rendering fix stays demand-gated per the issue's own note.

For a future full fix, the clean signal is `dataset.srid IS NOT NULL` ⟺ `geom_4326` exists (`dataset.srid` = `Find_SRID(..., 'geom')`; never `0` due to `chk_datasets_srid_positive`; do **not** AND with `geometry_type`, which is `NULL` for an empty/all-NULL-geometry table that still has `geom_4326`).

## Testing

- 2 new unit tests in `tests/test_sql_engine.py::TestQueryDataTool`: model-written `geom_4326` miss → clean note, no retry; `geom_4326` in `ORDER BY` → append + retry both miss → clean note.
- Local: **324 passed** across the AI/chat/sandbox suites (`test_sql_engine`, `test_ephemeral_geojson`, `test_chat_narrative`, `test_ai_chat_dataset`, `test_sandbox`, `test_sec025_sandbox_allowlist`, `test_sandbox_tenant_schema`, `test_phase_274_caches_and_pool`); `ruff format` + `ruff check` clean; pre-commit hooks pass.

Fixes #560.

### DIFF
--- a/backend/app/processing/ai/chat_actions.py
+++ b/backend/app/processing/ai/chat_actions.py
@@ -52,6 +52,32 @@ _DEFAULT_LABEL_TEXT_COLOR = "#333333"
 _OVERLAY_ROW_BUDGET = 50
 
 
+def _geom_4326_missing_note(err: SandboxError) -> dict | None:
+    """Clean degrade note when ``err`` was caused by a missing ``geom_4326`` column.
+
+    fix(#560): the SQL prompt + schema-context always name the geometry column
+    ``geom_4326`` (sql_generator.SQL_SYSTEM_PROMPT / build_sql_schema_context),
+    so the model emits it. A table without that column — a legacy dataset
+    ingested before the geom_4326 convention (no current ingest path produces
+    one: the missing-CRS gate rejects unknown-CRS uploads and every spatial path
+    builds geom_4326) — fails with ``column "geom_4326" does not exist``.
+    ``SandboxError.category`` is the generic ``query_failed``; the Postgres text
+    lives on ``__cause__`` (executor sets it via ``raise SandboxError(...) from
+    exc``). Returns None for any other error so it propagates unmasked.
+    """
+    detail = f"{err} {err.__cause__}".lower()
+    if "geom_4326" not in detail or "does not exist" not in detail:
+        return None
+    logger.info("query_data.geom_4326_missing")
+    return {
+        "error": (
+            "This dataset's geometry can't be used in spatial queries. "
+            "I can still answer questions about its attribute columns."
+        ),
+        "category": "llm_cannot_answer",
+    }
+
+
 def _safe_label_text_color(value: object) -> str:
     # fix(#394) CH-02: hex / real CSS named color / numeric-arg functional form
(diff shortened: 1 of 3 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #521, opened 2026-07-15
fix(admin): keep the fixed admin sidebar below the sticky navbar

ishiland:
## Summary

Follow-up to the design-audit remediation (#520), from a user-reported observation: at the top of an admin page, scrolling up makes the header visibly nudge downward over the sidebar.

Root cause (measured live): the navbar is `sticky top-0` (in document flow) while the shadcn sidebar container is `fixed inset-y-0 h-svh` (viewport-pinned at y=0, behind the navbar). Two consequences:
- the sidebar's own "Admin" header was permanently buried under the navbar (~57px of the sidebar was unreachable), and
- during macOS elastic overscroll the document (navbar included) translates while the fixed sidebar stays put — the navbar slides down over the sidebar.

Fix: one line on `AdminSidebar` — `className="top-14 h-[calc(100svh-3.5rem)]"` on the `<Sidebar>` (the primitive forwards it to the fixed container; tailwind-merge swaps out `h-svh`). Verified live: sidebar spans y=56→viewport bottom, the "Admin" eyebrow header is visible, "Back to App" sits at the true bottom, icon-collapse unaffected.

Scope note: this fixes the buried header and confines the overscroll artifact to the 1px seam; *fully* eliminating relative motion during rubber-band requires making the navbar `fixed` app-wide (spacer, DemoBanner ordering, builder viewport math) — filed separately.

## Verification
- `npx eslint src/components/admin/AdminSidebar.tsx` ✅ `npm run typecheck` ✅
- `npx vitest run src/components/admin` — 56 passed ✅
- Live geometry check via Playwright (navbar bottom 57 / sidebar top 56 / eyebrow visible at y=68) ✅

https://claude.ai/code/session_01Xd64KjyK8rmgEfvWemkP5w

PR #564, opened 2026-07-17
fix(shell): pin the top navbar with position:fixed (#522)

### LINKED ISSUE #522: App shell: make the top navbar fixed so it shares positioning context with viewport-fixed surfaces
Follow-up to #521 (admin sidebar offset fix).

The navbar is `sticky top-0` (in document flow) while the admin sidebar is `position: fixed`. Mixed positioning means macOS elastic overscroll translates the navbar but not the sidebar, producing visible relative motion at the seam. #521 confines the artifact (1px seam) but cannot eliminate it.

The durable fix is to make the navbar `fixed top-0 inset-x-0` app-wide so both chrome surfaces are viewport-pinned and nothing moves relative to anything during rubber-band. That is a small refactor with real blast radius, needing its own QA pass:

- a spacer / `pt-14` on the layout column to replace the navbar's flow height
- DemoBanner ordering/stacking above or below the fixed navbar
- builder + full-viewport map routes (`h-dvh overflow-hidden` viewport math assumes the navbar consumes flow space)
- skip-link / anchor offsets (`scroll-mt-16` already exists, re-verify)
- mobile safe-area padding (`pt-[env(safe-area-inset-top)]` currently on the sticky header)

Nice-to-have alongside: consider whether `overscroll-behavior-y` should be tuned for app surfaces.

https://claude.ai/code/session_01Xd64KjyK8rmgEfvWemkP5w

ishiland:
Closes #522. Follow-up to #521 (admin sidebar offset).

## Problem
The navbar was `sticky top-0` (in document flow) while the admin sidebar is `position: fixed`. Mixed positioning means macOS elastic overscroll translates the sticky navbar but not the fixed sidebar, sliding one over the other at the seam. #521 confined the artifact to a 1px seam but couldn't eliminate it.

## Fix
Make the navbar `fixed top-0 inset-x-0` so both chrome surfaces are viewport-pinned and nothing moves relative to anything during rubber-band.

The only compensation needed is reserving the navbar's height on the shell, because a `fixed` navbar leaves document flow:

- `AppLayout` shell gains `pt-[calc(3.5rem+env(safe-area-inset-top))]` (navbar `h-14` + notch inset). This applies to **both** layout branches:
  - `min-h-screen` pages → content pushed below the navbar.
  - `h-dvh overflow-hidden` map routes → `flex-1` loses exactly this height, same as when the navbar consumed flow height (so #553/#347 behavior is preserved: no page overflow, panels still scroll internally).
- The two conditional banners (`SiteBanner`, `DemoBanner`) now render below the fixed navbar — no reorder needed since the fixed navbar is out of flow.

## What did *not* need changing (already assumed a fixed-at-56px navbar)
- Admin sidebar offset `top-14 h-[calc(100svh-3.5rem)]` (#521) — now shares the navbar's fixed context.
- Skip-link target `scroll-mt-16` (64px clears the 56px navbar).
- z-order: navbar `z-40` sits above the `z-10` sidebar and all map overlays, below `z-50` modals and the `z-100` skip link.

The `overscroll-behavior-y` nice-to-have from the issue is intentionally skipped — the fixed navbar alone removes the relative motion, which is the actual artifact.

## Verification
- `npm run typecheck`, `npm run lint`, `vitest run src/components/layout` (54 tests) — all green. Added an `AppLayout.test.tsx` guard asserting the navbar is `fixed` (not `sticky`) and the shell reserves top padding, so a regression to `sticky` fails CI.
- In-browser (1440×900) computed-layout checks:
  - **Search** (`min-h-screen`): navbar `position:fixed` at y=0 (h=57, z=40); `#main-content` top = 56 (flush below, not hidden).
  - **Admin**: navbar and sidebar both `position:fixed`, seam aligned (navbar bottom 57 / sidebar top 56), sidebar header visible.
  - **Builder** (`h-dvh`): shell height 900 = viewport, main fills 844px below the navbar, map fills to the bottom edge, `document` **not** scrollable (no overflow).

### DIFF
--- a/frontend/src/components/layout/AppLayout.tsx
+++ b/frontend/src/components/layout/AppLayout.tsx
@@ -34,7 +34,17 @@ export function AppLayout() {
   const showFooterBranding = !isEnterprise || branding?.show_badge !== false;
 
   return (
-    <div className={cn('flex flex-col', isAuthenticatedMapRoute ? 'h-dvh overflow-hidden' : 'min-h-screen')}>
+    // fix(#522): the navbar is fixed (out of flow), so reserve its height
+    // (h-14 + safe-area) as top padding on the shell. Applies to both branches:
+    // min-h-screen pages get their content pushed below the navbar, and h-dvh
+    // map routes correctly lose this height to flex-1 — exactly as they did when
+    // the navbar consumed flow height. Banners render below the navbar as a result.
+    <div
+      className={cn(
+        'flex flex-col pt-[calc(3.5rem+env(safe-area-inset-top))]',
+        isAuthenticatedMapRoute ? 'h-dvh overflow-hidden' : 'min-h-screen',
+      )}
+    >
       <SkipToContent />
       {/* fix(#553): inside the flex shell so h-dvh map routes lose height to
           the banner instead of overflowing the viewport */}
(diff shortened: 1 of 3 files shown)


ishiland:
@codex review

PR #572, opened 2026-07-18
docs+install: launch-audit polish (badge links, MCP in SECURITY scope, /api/health note, ready-banner password hint)

ishiland:
Pre-announcement polish from the 2026-07-18 launch-readiness audit (6-surface sweep ahead of the r/gis + Show HN posts). Four small fixes, no schema/API/config impact:

- **Localized READMEs (es/fr/de):** the Python badge had an empty link target (`]()`) — now links to python.org like the English README.
- **`.github/SECURITY.md`:** the supported-packages scope now includes `geolens-mcp` (first published to PyPI with v1.4.8); security researchers previously couldn't tell it was in scope.
- **`RUNBOOK.md` §4:** documents that uptime/liveness checks must target `/api/health` — a bare `/health` is answered by the frontend SPA catch-all with HTML 200, which uptime monitors happily misread as healthy (observed on the live demo).
- **`scripts/install.sh`:** the final "GeoLens is ready" banner now repeats the generated-admin-password retrieval hint (`grep '^GEOLENS_ADMIN_PASSWORD=' .env`). The existing hint prints once mid-install and scrolls away; on the piped `curl | sh` path (no TTY, password auto-generated) a first-timer otherwise dead-ends at the login screen. Guarded with `${...:-}` defaults because the configure block is skipped on idempotent re-runs, and mirrors the existing `PROJECT_HINT` cd-hint pattern.

**Verification**
- `sh -n scripts/install.sh` and `bash -n scripts/install.sh` clean.
- All four localized badge lines now byte-match the English README's target.
- Live demo probe: `GET /health` returns `text/html` (SPA shell), `GET /api/health` returns JSON with `version`/`build` — matching the new RUNBOOK note.

Note for review: the installer change intentionally does **not** echo the password itself (same posture as the mid-run hint — secret stays in `.env`).

### DIFF
--- a/.github/SECURITY.md
+++ b/.github/SECURITY.md
@@ -33,7 +33,7 @@ This policy covers:
 
 - The `geolens` source repository (this repo)
 - Official container images published to GHCR (`ghcr.io/geolens-io/*`)
-- Official packages published to PyPI (`geolens`, `geolens-cli`) and npm (`@geolens/sdk`)
+- Official packages published to PyPI (`geolens`, `geolens-cli`, `geolens-mcp`) and npm (`@geolens/sdk`)
 - The public demo at demo.getgeolens.com
 
 Out of scope here (report to the respective repo or contact): the marketing/docs website (getgeolens.com), and Helm/deployment packaging ([geolens-deployments](https://github.com/geolens-io/geolens-deployments)).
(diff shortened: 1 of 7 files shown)


ishiland:
@codex review

ishiland:
@codex review — pushed: the four READMEs' password-retrieval grep now points at geolens/.env for the piped-install path (your catch on the paired site PR applied here too).

PR #574, opened 2026-07-18
test(e2e): add E2E_SKIP_SEED opt-out for host-backend stacks

### LINKED ISSUE #547: E2E: allow skipping the catalog seeding step for host-backend stacks
## Problem

`e2e/auth.setup.ts` always seeds a vector dataset through a real ingest after authenticating. Against a host-run backend (uvicorn on the host + docker Postgres — the standard worktree QA rig), ingest cannot work: the local storage adapter rejects the absolute staging path (`Storage key must be relative, got absolute path: ...`), so setup fails **after** `storageState` is written.

Today's workaround is two-step: run once to let auth save, then rerun the target spec with `--no-deps`. Works, but it's a trap for anyone else.

## Proposal

An env opt-out, e.g. `E2E_SKIP_SEED=1`, that makes `auth.setup.ts` authenticate + save storage state but skip `seedDataset` (specs that need the shared fixture would skip themselves when the manifest is absent). CI behavior unchanged.

## Context

Hit during the dataset-chat close-out (#542) while running `e2e/dataset-chat.spec.ts` against a worktree Vite + host API stack with `E2E_BASE_URL`.

ishiland:
Closes #547.

## What

`E2E_SKIP_SEED=1` makes `e2e/auth.setup.ts` authenticate and save storage state, then return before the shared catalog fixture is seeded. Documented in `e2e/README.md`.

## Why

Against a host-run backend (uvicorn on the host + docker Postgres — the standard worktree QA rig), the seeding ingest cannot work: the local storage adapter rejects the absolute staging path, so setup fails **after** `storageState` is written and every run needs the two-step `--no-deps` workaround (#547).

## Impact

- CI behavior unchanged (flag unset → identical code path).
- `catalog.teardown.ts` already no-ops when the manifest is absent, so cleanup is unaffected.
- No API/schema/config impact.

## Verification

- `E2E_SKIP_SEED=1 npx playwright test --project=setup` → 2 passed; `user.json` written, no `catalog-fixture.json` created, teardown no-oped.
- `npx playwright test --project=setup` (flag unset) → 2 passed; fixture seeded and removed by teardown as before.

https://claude.ai/code/session_013Q1gyuRm3QXKZL73eQ9UUV

### DIFF
--- a/e2e/README.md
+++ b/e2e/README.md
@@ -29,6 +29,14 @@ catalog-dependent flows. The cleanup project removes it after the browser tests
 finish, including failed runs. The API-only export suite manages its own fixture
 and does not require a browser install.
 
+Against a host-run backend (uvicorn on the host + docker Postgres), the seeding
+ingest cannot work — set `E2E_SKIP_SEED=1` to make setup authenticate and save
+storage state without creating the shared fixture:
+
+```bash
+E2E_SKIP_SEED=1 E2E_BASE_URL=http://localhost:5173 npx playwright test e2e/foo.spec.ts --project=chromium
+```
+
 ## Smoke groups
 
 The `e2e:smoke:*` scripts in the root `package.json` group specs by area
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

Issue #569, opened 2026-07-18
post-1.4.8 audit follow-ups (P3 batch)


ishiland:
Non-blocking P3 findings from the 2026-07-18 pre-release audit (4-agent diff review of v1.4.7..98628eb30 + live smoke). None affect correctness or security in current use.

**Backend**
- `tasks_common.py` `rename_pkey_to_match_table`: constraint identifiers interpolated with bare quotes instead of a quote_ident helper (inputs are server-derived; hygiene only).
- `features/schemas.py` `inline_json_schema`: no cycle guard — a future self-referential GeoJSON model would loop at import time (fail-fast; docstring warns).
- `ingest/parquet.py`: geometry-only parquet ingested with `include_geometry=False` silently inserts zero rows (extreme edge case).
- `ingest/parquet.py`: no explicit total-row/cell-size cap during ingest (parity with the ogr2ogr path, bounded by the 500 MB upload cap; noted for a future ingest-budget pass).

**Frontend**
- `AdminSidebar.tsx`: `top-14`/`h-[calc(100svh-3.5rem)]` ignores `env(safe-area-inset-top)` — sidebar header sits under the navbar by the inset on notched devices.
- Dataset-page Ask-AI FAB (z-30) can be partially covered by `PendingEditsBar` (z-40) when pending edits and AI coexist.
- `PublicViewerPage` error/expired branch: `SiteBanner` + `min-h-screen` slightly overflows the viewport (loading/main branches use `h-dvh` correctly).
- es locale: new `ai.chat` keys say "dataset" while the rest of the es namespace uses "conjunto de datos".

**Tests/CI**
- `test_ai_send_sample_values.py::test_search_tool_omits_samples_when_disabled` errored once under host `pytest -n4` (passes solo and in CI) — suspected settings-override parallel-isolation leak; worth a PPI-class look.

**Ops/docs**
- `infra/demo/cloud-init.yaml` pins `GEOLENS_VERSION demo-2026-06-24e` (stale through the 1.4.x line; live demo redeploys via a separate path).
- EDITIONS.md now lists identity-provider role mapping, ABAC, publication approval workflows, and custom embed-token lifetimes as Enterprise — verify all four actually ship in enterprise overlay v0.3.1 (REL-01 accuracy).

ishiland:
**EDITIONS.md REL-01 accuracy item verified 2026-07-18** — all four Enterprise claims ship (checked against enterprise-overlay tag `v0.3.1` + current core main):

[...]

PR #575, opened 2026-07-18
fix: address post-1.4.8 audit P3 batch (#569)

ishiland:
Addresses the tractable items from #569 (post-1.4.8 audit P3 batch). Does not fully close the issue — see "Not addressed" below.

## Backend
- **`tasks_common.py` / `rename_pkey_to_match_table`**: constraint identifiers now quoted via the existing `_sql_quote_ident` helper instead of bare f-string quotes (hygiene; inputs are server-derived).
- **`features/schemas.py` / `inline_json_schema`**: cycle guard — a future self-referential model now raises `ValueError` with the model and `$def` name instead of looping forever at import time. Test added.
- **`ingest/parquet.py`**: a geometry-only parquet loaded with `include_geometry=False` now raises `IngestionError` (before touching the DB) instead of silently publishing an empty dataset. Test added.

## Frontend
- **`AdminSidebar.tsx`**: top offset and height calc now include `env(safe-area-inset-top)`, matching the `AppLayout` shell formula, so the sidebar header clears the navbar on notched devices. Desktop unchanged (inset resolves to 0; live-verified).
- **Dataset Ask-AI FAB** (`DatasetChatPanel.tsx`): `z-30` → `z-40` so the sticky `z-40` `PendingEditsBar` can no longer cover it; the panel renders later in the DOM so it wins the tie, and it still has no spatial overlap with the z-40 reporter lifebuoy.
- **`PublicViewerPage.tsx`** error/expired branch: now an `h-dvh` flex column like the loading/main branches, so `SiteBanner` no longer pushes the page past the viewport; the card uses `m-auto` inside an `overflow-y-auto` region so short viewports scroll instead of clipping the card top.
- **es locale**: `ai.chat` title/emptyHint now say "conjunto de datos" like the rest of the namespace.

## Ops
- **`infra/demo/cloud-init.yaml`**: stale `GEOLENS_VERSION demo-2026-06-24e` pin bumped to `v1.4.8` (the "bump to GA tag once cut" TODO in the old comment; the live demo redeploys via a separate path).

## Verified, no change needed
- **EDITIONS.md Enterprise claims** — all four ship: identity-provider role mapping (`group_role_mapping` gated as paid in core `auth/oauth/service.py`), ABAC (enterprise v0.3.1 `permission/policy.py` + `predicates.py`), publication approval workflows (enterprise v0.3.1 `workflow/policy.py` + `notify.py`), custom embed-token lifetimes (core `embed_tokens/service.py` locks Community to the 30-day default).

## Not addressed (stay open on #569)
- Parquet ingest total-row/cell-size cap — the issue itself defers this to a future ingest-budget pass.
- `test_search_tool_omits_samples_when_disabled` `-n4` flake — needs a PPI-class investigation, not a batch fix.
- docs-site connection-pool page sync — lives in the docs repo.

## Verification
- `cd backend && uv run ruff check . && uv run ruff format --check .` on touched files — clean.
- `uv run pytest tests/test_ingest_parquet.py tests/test_api_contract_openapi.py` — 36 passed (includes 2 new tests).
- `cd frontend && npm run lint && npm run typecheck` — clean; targeted vitest (26 files / 109 tests) and `npm run test:i18n` — pass.
- Browser smoke (dev stack, frontend restarted first): invalid share link renders the not-found card with zero viewport overflow (`scrollHeight === innerHeight`); admin sidebar sits flush below the navbar; dataset-page Ask-AI FAB renders at computed `z-index: 40`.

https://claude.ai/code/session_013Q1gyuRm3QXKZL73eQ9UUV

### DIFF
--- a/backend/app/modules/catalog/features/schemas.py
+++ b/backend/app/modules/catalog/features/schemas.py
@@ -19,22 +19,32 @@ def inline_json_schema(model: type[BaseModel]) -> dict:
     the exported OpenAPI document, where pydantic's ``#/$defs/...`` pointers
     resolve against the document root and dangle — strict consumers (docs
     generators, ref bundlers) reject the whole document. Only safe for
-    non-recursive models; the GeoJSON models here are deliberately acyclic.
+    non-recursive models; a recursive model raises instead of looping
+    (fix(#569): cycle guard so a future self-referential model fails fast
+    with a clear error at import time rather than hanging).
     """
     schema = model.model_json_schema()
     defs = schema.pop("$defs", {})
 
-    def resolve(node: Any) -> Any:
+    def resolve(node: Any, expanding: frozenset[str] = frozenset()) -> Any:
         if isinstance(node, dict):
             ref = node.get("$ref")
             if isinstance(ref, str) and ref.startswith("#/$defs/"):
-                target = resolve(defs[ref.rsplit("/", 1)[-1]])
+                name = ref.rsplit("/", 1)[-1]
+                if name in expanding:
+                    raise ValueError(
+                        f"inline_json_schema({model.__name__}): recursive "
+                        f"$ref to {name!r}; only acyclic models can be inlined"
+                    )
+                target = resolve(defs[name], expanding | {name})
                 # Keep sibling keys (description, default) over the target's.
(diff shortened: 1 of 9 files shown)


ishiland:
@codex review

PR #613, opened 2026-07-20
docs(readme): make no-lock-in an explicit position and drop the production claim

ishiland:
## What

Two changes to all four README locales.

**1. Removes an inaccurate claim.** The Early release note said *"The core has run in production"*. It is not accurate, and it had shipped translated into `es`, `fr`, and `de`. Removed from all four.

**2. Makes no-lock-in an explicit position.** The README already contained every portability fact (export formats, OGC/STAC endpoints, PostGIS storage, no telemetry) but scattered across sections and always phrased as a capability, never as a promise. Adds one bullet to the `Why GeoLens?` list stating it directly.

Localized to each locale's established term rather than transliterating the English:

| Locale | Term |
|---|---|
| en | No lock-in |
| es | Sin dependencia de proveedor |
| fr | Sans verrouillage |
| de | Kein Lock-in |

## Why now

Part of making data portability a consistent position across every GeoLens surface. The companion change to the marketing site and docs is geolens-io/getgeolens.com#68.

## Scope note

No new claims. Everything the new bullet asserts is already documented behavior and already stated elsewhere in the same README.

Deliberately **not** touching `frontend/src/i18n/**`: `common.json` already ships *"The data stays in your database"* in all four locales, so the in-app surface is already on-message, and adding keys there would risk the parity gate for no positional gain.

## Verification

```
python3 scripts/check_public_surface.py    # passed, 68 tracked files scanned
python3 scripts/check_docs_contract.py     # passed, no forbidden drift
```

## Drift spotted, not fixed here

The English README is **behind** its translations in two places. `es`/`fr`/`de` list `XLSX` as an upload format and `DCAT 3/DCAT-US/GeoDCAT-AP` catalogs in the "works with your tools" bullet; English lists neither. Out of scope for a copy-position change, worth a follow-up.

https://claude.ai/code/session_01GkA7MLHfoRxF9RU7Srs5rk

### DIFF
--- a/README.de.md
+++ b/README.de.md
@@ -66,7 +66,7 @@ GeoLens ersetzt diesen Ablauf:
 
 - **Ein Katalog:** Laden Sie Shapefiles, GeoPackages, GeoTIFFs, CSVs oder XLSX hoch; innerhalb weniger Minuten sind sie durchsuchbar, als Vorschau verfügbar und exportierbar
 - **Funktioniert mit Ihren Werkzeugen:** OGC API Features/Records, STAC API 1.0, DCAT-3-/DCAT-US-/GeoDCAT-AP-Kataloge und direkte Kachel-URLs für QGIS, ArcGIS und MapLibre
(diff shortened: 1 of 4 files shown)


ishiland:
@codex review

ishiland:
Good catch on the raster scoping (P3). Fixed in fd1107668: the bullet now says data stays on infrastructure you control and leaves through open formats, and names vector export formats, raster COG download, and direct OGC API access separately instead of implying everything leaves via the vector formats. Same fix applied to the marketing PR (getgeolens.com#68).

@codex review

PR #614, opened 2026-07-21
fix(seed): backfill license + keywords on the demo showcase datasets

ishiland:
Fixes the "demo catalog reads as a proprietary raster dump" cluster from the 2026-07-20 pre-launch demo audit (findings **1006 / 2317 / 976**).

## Problem
The showcase ingest flow (`ingest_geojson` / manifest) only sets `title` + `summary`, so all 16 hand-seeded showcase datasets default to license **`proprietary`** with **zero keywords** — contradicting the open sources their own summaries cite, and leaving the demo's faceted-search sidebar empty. Verified live: `demo.getgeolens.com/api/search/datasets/` → 27/27 `proprietary`, 17/27 zero-keyword (the only keyworded ones are the Sentinel-2 STAC scenes).

## Change
An idempotent post-build pass, `enrich_showcase_metadata`, plus three small `Api` helpers:
- **License** → a plain `DatasetUpdate` PATCH (`patch_dataset(license=…)`).
- **Keywords** → the records keyword endpoint (`POST /api/records/{record_id}/keywords/`), which is keyed by **`record_id`, not `dataset_id`** (resolved from the dataset detail) and takes **one keyword per POST**, added only when absent.
- Each dataset is isolated like the builders (one flaky PATCH can't skip the rest); the pass only touches datasets that exist, so it composes with `--only` and runs on a plain (no `--force`) re-seed.

Licenses are each dataset's real upstream terms (public-domain USGS / NOAA / US Census / MTA / NYC Open Data / NASA, Natural Earth, © OpenStreetMap contributors ODbL, swisstopo OGD).

Also (**fix 976**) rewrites the `Recent Earthquakes - Heatmap source` summary — which read as an internal rendering workaround — to describe it as the map's density layer. It stays **public** on purpose: a private dataset can't back a public map layer, so privatizing it would break the Restless Earth heatmap.

`World States & Provinces` is intentionally **left bare** — it is the summary-less canvas for the AI metadata-generation demo.

## Scope / gates
`scripts/` is not linted by CI (`working-directory: backend`) or pre-commit (`files: ^backend/`), so there is no ruff gate here; nonetheless the additions are `py_compile`-clean and `ruff format`-clean (the one remaining format hunk is pre-existing code left untouched).

## Validation
Not runtime-tested — there is no seed target but the production demo, so **the demo re-seed is the validation**. Endpoints/schemas were verified against the backend: `DatasetResponse.record_id`, `KeywordCreate`, and `POST /records/{id}/keywords/` (requires `edit_metadata`, which the admin seeder has).

## Deploy note
This pass lands on a plain (`--force`-less) re-seed. The **ADA map color fix is separate** — it lives in the map's *layer* definition and a no-`--force` re-seed skips existing maps, so ADA needs `--only restless --force` (recreates the map → new map_id) or a targeted layer PATCH. This metadata pass has no such constraint.

### DIFF
--- a/scripts/seed-showcase.py
+++ b/scripts/seed-showcase.py
@@ -2982,7 +2982,8 @@ def build_collections(api: Api, force: bool = False) -> str:
 # ingest flow (ingest_geojson / manifest) only sets title + summary, so every
 # one of these defaulted to license "proprietary" with zero keywords - which
 # reads as a proprietary raster dump and contradicts the open sources their
-# own summaries cite (2026-07-20 pre-launch demo audit, findings 1006/2317/976).
+# own summaries cite (fix(#614): proprietary licenses + empty keyword facets on
+# the demo, flagged in the 2026-07-20 pre-launch audit).
 # Licenses are each dataset's real upstream terms; keywords power the faceted-
 # search sidebar. "World States & Provinces" is intentionally omitted - it is
 # the summary-less canvas for the AI metadata-generation demo and must stay bare.
@@ -3026,7 +3027,7 @@ SHOWCASE_METADATA: dict[str, dict] = {
     "Recent Earthquakes - Heatmap source": {
         "license": "USGS Earthquake Hazards Program (US public domain)",
         "keywords": ["earthquakes", "seismic", "usgs", "density", "heatmap"],
-        # fix(976): the old summary read as an internal rendering workaround;
+        # fix(#614): the old summary read as an internal rendering workaround;
         # describe it as the map's density layer instead.
         "summary": (
             "USGS M4.5+ earthquakes from the last 30 days, styled as the "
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review

ishiland:
Both addressed in e293a2b02:

- **P2 (duplicates):** the pass now iterates `list_datasets_full()` and enriches *every* dataset whose title is in the metadata table, instead of `datasets_by_title()`'s title→newest-id map — so on a `--force`-reseeded instance the older same-titled copies get patched too, not just the newest.
- **P1 (finding ids):** replaced the bare private audit ids (`1006/2317/976`, `fix(976)`) with the `#614` PR anchor + one-line context, per the AGENTS.md inline-comment convention.

(The failing `JS audit` / Trivy checks are a repo-wide `brace-expansion <1.1.16` npm advisory, unrelated to this Python-only change.)

@codex review

PR #617, opened 2026-07-21
fix: repair the two dead /api/docs links (footer API entry + OGC service-doc)

ishiland:
Two launch-day links point at `/api/docs` (Swagger UI), which returns `404 {"detail":"Not Found"}` whenever interactive docs are disabled — i.e. **every `ENVIRONMENT=production` deployment** (the recommended self-host setting) and the **public demo**. Same root cause, two surfaces; fixed together.

## 1. Footer "API" link (user-visible) — `frontend/src/lib/external-links.ts`
The app footer renders for anonymous visitors on every non-authenticated route. Its **API** entry linked to the instance-relative `/api/docs`, so one of only five footer links — the one an API-minded evaluator is most likely to click — was dead on the demo and every production instance.

**Fix:** repoint `GEOLENS_API_DOCS_URL` at the canonical public API guide `https://docs.getgeolens.com/guides/api/` (verified `200`), matching how every other external footer link already works. The const is used only in `AppFooter.tsx`. No i18n change — `footer.api` is already `"API"` in all four locales.

## 2. OGC landing `service-doc` link (machine-visible) — `backend/app/standards/ogc/router.py`
`GET /api/` advertised `rel="service-doc" → /api/docs` **unconditionally**, so the one HTML entry point a standards client discovers by walking the landing links 404s in production. Walking the API root is exactly what a skeptical GIS professional does to check whether the OGC claims are real.

**Fix:** emit `service-doc` only when Swagger actually resolves (`not settings.is_production` — the same signal that drives `docs_url` in `api/main.py`). `service-doc` is optional in OGC API Common, so omitting it when `/docs` is off is spec-compliant. `service-desc → /openapi.json` (still served `200` in production) is unchanged.

## Verification
- `docs.getgeolens.com/guides/api/` → `200`; live demo: `/api/docs` → `404`, `/api/openapi.json` → `200` (confirms only `service-doc` needed gating).
- **Frontend:** ESLint clean, `tsc -b --noEmit` clean, 20/20 AppFooter-touching tests pass; no test hardcoded the old URL.
- **Backend:** `ruff check` + `ruff format --check` clean; `tests/test_ogc_discovery.py` 19/19 pass, including a new `test_landing_page_omits_service_doc_in_production` covering the production branch (service-doc absent, service-desc retained).
- No OpenAPI snapshot drift — the `LandingPage` response schema is unchanged (only runtime link content varies).

### DIFF
--- a/backend/app/standards/ogc/router.py
+++ b/backend/app/standards/ogc/router.py
@@ -8,6 +8,7 @@ from sqlalchemy import select
 from sqlalchemy.exc import OperationalError, ProgrammingError
 from sqlalchemy.ext.asyncio import AsyncSession
 
+from app.core.config import settings
 from app.core.db.tenant_session import current_tenant_var
 from app.core.dependencies import get_db
 from app.core.geo import extent_to_bbox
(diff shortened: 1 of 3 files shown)


ishiland:
@codex review

PR #629, opened 2026-07-22
fix: revalidate the SPA shell (#622) + seed water/hydrology showcase data (#626)

### LINKED ISSUE #622: SPA shell served with no Cache-Control — stale index.html references deleted hashed assets after redeploys
## What happened

On 2026-07-21 (launch day), 3 distinct real visitors (mobile carriers among them) received **404s on `/assets/index-QvqZ23KQ.css`** — a hashed asset from a build that was replaced on redeploy (12 requests over the day). Their browsers held a stale `index.html` referencing assets that no longer exist in the current image, so they got an unstyled/broken page.

## Root cause

`frontend/nginx.conf` sets cache headers for exactly two things:

- `location = /env-config.js` → `no-cache, no-store, must-revalidate` (line ~96)
- `location /assets/` → `expires 1y; Cache-Control: public, immutable` (correct)

The SPA shell itself (`/index.html`, served via the `rewrite ^ /index.html break;` and `try_files ... /index.html` locations) sends **no Cache-Control header at all** — confirmed live: `curl -I https://demo.getgeolens.com/` returns no `cache-control` (CF shows `DYNAMIC`, so the edge isn't caching it; this is browser heuristic caching). With no header, browsers may heuristically cache the shell and re-serve it days later, pointing at hashed assets that were deleted by the redeploy.

`LazyLoadErrorBoundary` covers lazy JS chunk failures with a reload, but a 404 on the root stylesheet at initial load isn't caught by it.

## Fix

Add `add_header Cache-Control "no-cache";` to every location that serves `index.html` in `frontend/nginx.conf`:

1. The SPA fallback location (`try_files $uri $uri/ /index.html`, ~line 321)
2. The `/m/` embed-shell location (`rewrite ^ /index.html break;`, ~line 275) — note this block was recently touched by #618 (per-token frame-ancestors), so coordinate the header with what's already emitted there.

`no-cache` (revalidate every time) is preferred over `no-store` — the shell is tiny and ETag/304 still works.

## Reproduction

1. `docker build --target frontend -t geolens-frontend-test .` then run it and `curl -sI http://localhost:<port>/` — observe no `Cache-Control` header.
2. After the fix: `Cache-Control: no-cache` present on `/`, `/maps/anything` (SPA route), and `/m/<token>`; `/assets/*` still `public, immutable`.

## Acceptance criteria

- [ ] `curl -sI` on `/`, a deep SPA route, and an `/m/` embed URL all return `Cache-Control: no-cache`.
- [ ] `/assets/*` responses unchanged (`public, immutable`, 1y).
- [ ] Existing embed-shell headers from #618 (frame-ancestors) still present on `/m/`.

### LINKED ISSUE #626: Demo/showcase seeds have no data for the most common organic search terms (water, rivers, lakes, hydrology)
## What happened

Real visitors on the public demo (2026-07-22) searched `water`, `water bodies`, `rivers`, `lakes`, `hydrology` — and got **zero results** for all of them.

This is not a search bug. Semantic search was verified healthy the same day via direct vector probes (synonym queries like "skyscrapers" → Manhattan Building Heights at cosine distance 0.56 resolve correctly). The catalog simply contains no water/hydrology datasets: the showcase seeds cover countries/states, earthquakes, hurricanes, volcanoes, tectonic plates, NYC subway/buildings, and DEMs — nothing aquatic. Nearest-neighbor distance for water queries against the whole corpus is ≥ 0.72, past the 0.7 relevance cutoff, so the semantic path correctly reports no matches and FTS finds nothing either.

Water-related terms are the classic first thing GIS people type into a catalog (and the README's own example curls use `q=hydrology`, which returns empty on every install seeded with the current showcase set — a long-standing papercut).

## Fix

Add water/hydrology showcase data to the seed script (`scripts/` — the `seed-showcase.py` used by installs and the demo):

- **Natural Earth 1:50m rivers + lake centerlines** (`ne_50m_rivers_lake_centerlines`) and **1:50m lakes** (`ne_50m_lakes`) — small (few MB), public domain, instantly recognizable on a world map.
- Give them real titles/summaries/keywords ("hydrology", "rivers", "lakes", "water bodies") so both FTS and embeddings hit.
- Regenerate embeddings for the new records as part of seeding (the existing backfill path).

## Acceptance criteria

- [ ] Fresh install + seed: `GET /api/search/datasets/?q=rivers` and `?q=hydrology` return ≥ 1 result each (`numberMatched` ≥ 1 — check the result set, not just a 200).
- [ ] README example curls return data on a seeded install.
- [ ] Demo re-seeded after release so the live search terms resolve (ops step, note in the PR).

ishiland:
Two small demo-visible fixes that share one redeploy/re-seed ops step.

Closes #622
Closes #626

## #622 — SPA shell served with no Cache-Control

`frontend/nginx.conf` set cache headers for `/env-config.js` and `/assets/` only. `index.html` went out with **no** `Cache-Control`, so browsers cached it heuristically and, after a redeploy, asked for hashed `/assets/*` that the new image no longer contains — 3 real visitors hit 404s on the root stylesheet on launch day (`LazyLoadErrorBoundary` only covers lazy JS chunks).

Fix: `add_header Cache-Control "no-cache"` on both locations that serve the shell — the SPA fallback (`location /`) and the `/m/` embed shell. `no-cache` revalidates but keeps ETag/304, so the cost is one conditional GET on a tiny document. `location /` also serves the other unhashed root files (favicon, og-image), which want the same rule.

**Verified** against the built `frontend` image running on the compose network:

| request | result |
|---|---|
| `/` | `Cache-Control: no-cache` |
| `/maps/anything` (deep SPA route) | `Cache-Control: no-cache` |
| `/m/<token>?embed=true` | `Cache-Control: no-cache` |
| `/assets/index-*.css` | `Expires: +1y`, `Cache-Control: public, immutable` (unchanged) |
| `/m/x?embed=true&et=bogus` | CSP still ends `frame-ancestors 'none'` — #618 fail-closed intact |

## #626 — no data for the most common organic searches

Demo visitors searched `water`, `water bodies`, `rivers`, `lakes`, `hydrology` → zero results for all five. Not a search bug (semantic was verified healthy by direct vector probe the same day): the catalog simply had nothing aquatic, so every water query landed past the 0.7 relevance cutoff. The README's own `q=hydrology` curl returned empty on every seeded install.

Fix: seed Natural Earth 1:50m **rivers + lake centerlines** and **1:50m lakes** in `build_catalog` (public domain, ~800 KB each, same pinned `v5.1.2` tag as the other NE sources), with summaries and keywords covering the terms people actually type. Also trims one fictional dataset (`subwatersheds`) out of the README's semantic-search example comment.

- Embeddings need no extra seeder step: ingest defers one, and the keyword/license pass re-defers on change.
- Titles deliberately differ from the retired first-generation ones in `RETIRED_DATASETS` (`… (Natural Earth 10m)`, `World Lakes (Natural Earth 1:50m)`), so `--prune` does not delete what this seeds.

**Verified** on the dev stack — anonymous, checking result sets rather than status codes:

```
water          2 ['World Lakes & Reservoirs …', 'World Rivers & Lake Centerlines …']
water bodies   2 [same]
rivers         2 [same]
lakes          2 [same]
hydrology      2 [same]
countries      4 (unchanged)
```

## Ops follow-up (after merge + release)

1. Redeploy the demo so nginx picks up the shell header (prune first, per the redeploy procedure).
2. Re-run the seeder on the demo: `python3 scripts/seed-showcase.py --only catalog` — it reuses everything existing and only ingests the two new datasets.
3. Spot-check `https://demo.getgeolens.com/api/search/datasets/?q=hydrology`.

## Note for reviewers

On a **fresh** seed the license/keyword enrichment pass can skip the just-ingested datasets: the admin dataset list is Valkey-cached for 60s and the worker's `invalidate_catalog_cache()` only reaches the API process when a shared Valkey is running. A second `--only catalog` run fills them in. That is pre-existing (#614's pass has the same exposure) and does not affect this PR's acceptance criteria — `hydrology`/`water bodies` are in the summaries, not only the keywords — so it is left alone here rather than widened into a backend change.

## Verification commands

- `docker build --target frontend -t geolens-frontend-test .` then run with `API_UPSTREAM=http://api:8000` on the compose network + `curl -sI`
- `python3 scripts/seed-showcase.py --only catalog`
- `curl -sG "http://localhost:8080/api/search/datasets/" --data-urlencode "q=hydrology"`

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,8 +84,8 @@ TOKEN=$(curl -s -X POST http://localhost:8080/api/auth/login/ \
 Semantic search takes a one-time admin setup: an embedding provider and the AI + Semantic Search toggles in the admin AI settings, plus an embedding backfill for data ingested before setup (the [search guide](https://docs.getgeolens.com/guides/user/search/) walks through it). Once that's on, search datasets by meaning instead of exact keyword matches:
 
 ```bash
-# Semantic search ranks by meaning: "hydrology" surfaces subwatersheds, lakes,
-# and river networks whose titles never mention the word
+# Semantic search ranks by meaning: "hydrology" surfaces the lake and river
+# network datasets whose titles never mention the word
 curl "http://localhost:8080/api/search/datasets/?q=hydrology&limit=3" \
   -H "Authorization: Bearer $TOKEN" | jq '.features[].properties.title'
 ```
--- a/frontend/nginx.conf
+++ b/frontend/nginx.conf
@@ -272,6 +272,8 @@ server {
         # `rewrite ^ ... break` rewrites the URI in place and STOPS location
         # re-matching, so this block's headers are the ones served. /m/* is never
         # a real static file, so the old `$uri` existence check was a no-op.
+        # fix(#622): same shell-revalidation rule as `location /` below.
+        add_header Cache-Control "no-cache" always;
         rewrite ^ /index.html break;
     }
 
@@ -318,6 +320,14 @@ server {
         add_header X-Content-Type-Options "nosniff" always;
         add_header Referrer-Policy "strict-origin-when-cross-origin" always;
         add_header Content-Security-Policy "default-src 'self'; script-src 'self' 'wasm-unsafe-eval' https://static.cloudflareinsights.com; style-src 'self' 'unsafe-inline'; img-src 'self' data: blob: https:; font-src 'self' data:; connect-src 'self' https: wss:; worker-src 'self' blob:; child-src 'self' blob:; object-src 'none'; base-uri 'self'; frame-ancestors 'self'" always;
+        # fix(#622): the SPA shell carried NO Cache-Control, so browsers cached it
+        # heuristically and, after a redeploy, asked for hashed /assets/* that the
+        # new image no longer contains (real 404s from 3 visitors on launch day).
+        # `no-cache` = revalidate every time, but ETag/304 still applies, so the
+        # cost is one conditional GET. Everything else served from here is
+        # unhashed too (favicon, og images), so it wants the same rule.
+        # /assets/* keeps `public, immutable` from its own location above.
+        add_header Cache-Control "no-cache" always;
         try_files $uri $uri/ /index.html;
     }
 }
--- a/scripts/seed-showcase.py
+++ b/scripts/seed-showcase.py
@@ -169,6 +169,8 @@ NE_BASE = (
 NE_COUNTRIES = NE_BASE + "ne_50m_admin_0_countries.geojson"
 NE_PLACES = NE_BASE + "ne_50m_populated_places.geojson"
 NE_ADMIN1 = NE_BASE + "ne_50m_admin_1_states_provinces.geojson"
+NE_RIVERS = NE_BASE + "ne_50m_rivers_lake_centerlines.geojson"
+NE_LAKES = NE_BASE + "ne_50m_lakes.geojson"
 # PB2002 plate-boundary steps (Peter Bird 2003, via Hugo Ahlenius/Nordpil; the
 # *steps* file - not boundaries - carries per-segment STEPCLASS + velocity).
 PB2002_STEPS = (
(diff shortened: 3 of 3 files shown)


ishiland:
@codex review

PR #630, opened 2026-07-22
fix(search): skip the embedding call for typeahead-length queries (#625)

### LINKED ISSUE #625: Semantic search embeds every search-as-you-type keystroke — add a minimum query length gate
## What happened

Demo API logs (48h) show the RRF semantic path generating **provider embedding calls for 1–3 character typeahead prefixes**: `u`, `us`, `usa`, `wa`, `tu`, `tun`, `tuni`… Each keystroke pause fires a full embedding API round-trip (Azure OpenAI), which then virtually always yields `empty_vector_ranks` (garbage-prefix embeddings don't land within the 0.7 cosine-distance cutoff) and falls back to FTS anyway.

The 300s embedding cache (`service_semantic.py`) doesn't help because each prefix is a distinct cache key.

## Cost/latency

Every short prefix burns a paid provider call and adds its latency to the search response, for a result that is deterministically FTS-only. With search-as-you-type in the catalog UI this is the dominant source of embedding calls in live traffic.

## Fix

In `_run_rrf_merge` (`backend/app/modules/catalog/search/service_semantic.py`, right after `q_stripped = filters.q.strip()` at ~line 266), return `None` (the existing "RRF doesn't apply → standard FTS sort path" contract) when `len(q_stripped) < 4`. One `if`, no config knob.

Note the docstring contract in that function: `None` means "fall through to the standard sort path" — that is exactly the wanted behavior for short queries, and it skips `_get_vector_ranks` (and therefore `generate_embedding`) entirely.

## Acceptance criteria

- [ ] A search with `q` shorter than the threshold performs **zero** calls to `generate_embedding` (assert via mock/spy in a unit test) and still returns FTS results.
- [ ] A search at/above the threshold behaves exactly as today.
- [ ] Threshold documented in the function docstring.

For context: the semantic path itself was verified healthy on the demo (direct vector probes: "skyscrapers" → Manhattan Building Heights at 0.56, "seismic activity" → Earthquakes at 0.48, "subway system" → NYC Subway Lines at 0.45 — all well within the 0.7 cutoff).

ishiland:
Closes #625

## Problem

Demo API logs (48h) show the RRF semantic path firing a **paid provider embedding round-trip for every search-as-you-type prefix** — `u`, `us`, `usa`, `wa`, `tu`, `tun`, `tuni`… Each one then misses the 0.7 cosine-distance cutoff (garbage-prefix embeddings land nowhere near a real record) and falls back to FTS anyway.

The 300s embedding cache in `service_semantic.py` can't absorb this: every prefix is a distinct cache key. With search-as-you-type in the catalog UI, these prefixes are the dominant source of embedding calls in live traffic, and each adds provider latency to a response that was deterministically FTS-only.

## Fix

One `if` in `_run_rrf_merge`: return `None` when the stripped query is shorter than `_MIN_SEMANTIC_QUERY_LEN` (4). `None` is that function's existing "RRF doesn't apply → caller falls through to the standard sort path" contract, so `_get_vector_ranks` — and therefore `generate_embedding` — is never reached. No config knob.

Threshold is documented in the function docstring and on the constant.

## Verification

`backend/tests/test_hybrid_search.py::test_short_query_skips_embedding_call` asserts the provider call count directly rather than inferring it from the response:

- `q=map` (3 chars) → `mock_embed.assert_not_called()`, still returns its FTS match (`River Systems Map`)
- `q=rrrr` (at the threshold) → `mock_embed.assert_awaited_once()`, vector path unchanged

Confirmed the test has teeth: it fails with the guard stashed out, passes with it.

## Verification commands

```
cd backend && set -a && source ../.env.test && set +a && uv run pytest tests/test_hybrid_search.py -q
# 13 passed

uv run pytest tests/test_search.py tests/test_search_embedding_cache.py \
  tests/test_search_pagination.py tests/test_search_cache.py tests/test_attribute_search.py -q
# 73 passed
```

## Note on the threshold

At 4, deliberate 3-char searches (`nyc`, `dem`, `gis`) lose semantic *ranking* but still get FTS, which matches them lexically anyway — those words appear in the titles/keywords people are searching for. The prefixes this actually cuts (`u`, `wa`, `tun`) had no semantic value to lose.

### DIFF
--- a/backend/app/modules/catalog/search/service_semantic.py
+++ b/backend/app/modules/catalog/search/service_semantic.py
@@ -40,6 +40,13 @@ _EMBEDDING_CACHE_MAX_SIZE = 512
 # the exact vector-only match COUNT (a full cosine scan) and approximates
 # from the already-fetched top-k ranks instead.
 _EXACT_SEMANTIC_COUNT_MAX_ROWS = 5000
+
+# fix(#625): search-as-you-type sent every keystroke prefix ("u", "us", "usa")
+# through a paid provider embedding call that then virtually always missed the
+# 0.7 cosine cutoff and fell back to FTS anyway. Below this length the query is
+# treated as a typeahead prefix and skips the vector path entirely. The 300 s
+# embedding cache cannot help here — each prefix is a distinct cache key.
+_MIN_SEMANTIC_QUERY_LEN = 4
 _embedding_cache: "OrderedDict[tuple[str, str], tuple[float, list[float]]]" = (
     OrderedDict()
 )
@@ -253,8 +260,10 @@ async def _run_rrf_merge(
     that violate an active filter (e.g. a polygon under ``geometry_type=Point``),
     AND a nearer non-visible neighbour displacing a valid match out of the top-k.
 
-    Returns ``None`` when RRF doesn't apply (vector backend empty/failed).
-    Caller falls through to the standard sort path on None.
+    Returns ``None`` when RRF doesn't apply (vector backend empty/failed, or the
+    query is shorter than ``_MIN_SEMANTIC_QUERY_LEN`` = 4 characters and is
+    therefore treated as a search-as-you-type prefix). Caller falls through to
+    the standard sort path on None.
 
     Returns ``([], total)`` rather than ``None`` when the FTS-cap query
     yields zero ids -- caller returns that tuple as-is. Preserved from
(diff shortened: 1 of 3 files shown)


ishiland:
@codex review

ishiland:
Pushed `9ee4951c4`: the gate addition pushed `service_semantic.py` from 386 to 395 LOC, past the 390 cap enforced by `test_layering.py::test_decomposed_service_modules_stay_within_size_budgets` — that was the Backend Tests failure. Trimmed the rationale comment and recorded the carve-out in the allowlist (cap 390 → 395, exact, so the next addition needs its own review).

Ran the exact failing CI step locally this time: `pytest tests/test_manifest_apply_{api,service,vrt,roundtrip}.py tests/test_layering.py` → 90 passed.

@codex review

PR #631, opened 2026-07-22
fix(oauth): persist the IdP's email_verified assertion on JIT provisioning (#623)

### LINKED ISSUE #623: OAuth JIT provisioning drops email_verified — all SSO users persist as unverified
## What happened

On the public demo, **27 of 27 OAuth-provisioned users have `email_verified = false`** in `catalog.users`, even though every OAuth path guarantees a verified email before provisioning.

## Root cause

The claims dicts guarantee verification:
- GitHub: `_fetch_github_claims` raises unless a primary+verified email exists, then returns `"email_verified": True` (`backend/app/modules/auth/oauth/service.py:649`)
- OIDC providers: the `email_verified` claim is checked for linking trust (`service.py:899`)

…but the JIT `User(...)` construction (`backend/app/modules/auth/oauth/service.py:971`) never sets `email_verified`, so the model default (`false`) is persisted:

```python
new_user = User(
    username=username,
    email=email,
    password_hash=None,
    auth_provider="oauth",
    status="active",
    is_active=True,
)   # email_verified missing
```

## Impact

- Admin user list shows every SSO user as unverified.
- Anything that gates on `email_verified` in the future (or the SIGNUP-03 verification flow semantics) misfires for the entire SSO population.

## Fix

1. In the JIT provisioning path, pass the claim through: `email_verified=bool(userinfo.get("email_verified"))`. Do **not** hardcode `True` — the OIDC path can provision from a claim where `email_verified` is absent/false (only the *linking* path requires it true), so persist what the provider asserted.
2. Alembic data migration backfilling existing rows: for GitHub-linked users `True` is safe (the guard makes it unconditional); for OIDC-linked users backfill `True` only if you accept the H-30 trust argument, otherwise leave them and let next-login refresh it (in which case also add a login-time update).
3. Unit test: JIT-provision with `email_verified: True` claims → row persists `True`.

## Acceptance criteria

- [ ] New OAuth signups persist the provider's `email_verified` assertion.
- [ ] Migration (or documented backfill) for existing `auth_provider='oauth'` rows.
- [ ] Test covering the provisioning path.

ishiland:
Closes #623

## Problem

**27 of 27 OAuth-provisioned users on the public demo have `email_verified = false`**, though every OAuth path guarantees a verified email before provisioning. The JIT `User(...)` construction simply never set the field, so the model default persisted.

## Fix

```python
email_verified=email is not None and email_verified,
```

Both halves are load-bearing: reaching provisioning with a non-`None` email means the IdP verified it (the H-30 branch nulls an unverified one), and an account with no email has nothing verified to record. Deliberately **not** hardcoded `True` — the OIDC path can provision from a claim that is absent or false; only the *linking* path requires it true.

## Existing rows — two complementary paths

**Migration `0028` backfills only GitHub-linked accounts.** That's the one provider whose assertion is unconditional: `_fetch_github_claims` raises unless the account has a primary+verified email, so a GitHub-linked row with an email cannot have come from an unverified address. OIDC/Google/Microsoft rows are left alone — their claim is per-provider and older rows predate the H-30 gate, so flipping them would be a guess on a security-adjacent flag.

**Those converge on next login instead.** A returning user whose IdP asserts verification for the *same* address we store gets the flag set. Guards: a verified claim for a *different* address changes nothing, and only `auth_provider='oauth'` accounts are touched, so a local account's own verification state is never flipped by an OAuth link.

Between them, the demo's 27 (all GitHub) fix on deploy, and any OIDC population catches up without a bulk assertion nobody can justify.

## Tests

| test | covers |
|---|---|
| `test_jit_persists_email_verified_claim` | both halves — verified claim → `True`; absent claim → email dropped, `False` |
| `test_returning_user_email_verified_refreshes` | legacy row converges on next login |
| `test_returning_user_email_verified_not_flipped_for_other_email` | verified claim for a different address verifies nothing |

The first two fail with the fix stashed out; the third is a guard test (asserts absence of change) and passes either way by design.

## Verification commands

```
uv run pytest tests/test_oauth.py tests/test_oauth_github_provider.py \
  tests/test_oauth_github_migration.py tests/test_oauth_account_tenant_isolation.py \
  tests/test_oauth_audit_correlation.py tests/test_oauth_destination_security.py \
  tests/test_oauth_microsoft_iss.py tests/test_layering.py -q
# 143 passed

docker compose exec api ... alembic upgrade heads   # applied clean on the dev stack
make alembic-check                                   # No new upgrade operations detected
```

## Not included

A login-time refresh for **local** accounts linking an OAuth identity — `email_verified` there belongs to the SIGNUP-03 verification flow, and an OAuth link shouldn't satisfy it. Say the word if that's wanted; it's a different question from this one.

### DIFF
--- a/backend/alembic/versions/0028_oauth_email_verified_backfill.py
+++ b/backend/alembic/versions/0028_oauth_email_verified_backfill.py
@@ -1,0 +1,54 @@
+"""Backfill email_verified for GitHub-provisioned OAuth users.
+
+fix(#623): the JIT provisioning path built ``User(...)`` without
+``email_verified``, so every OAuth-provisioned account persisted the model
+default (false) even though the claim guaranteed verification — 27 of 27 SSO
+users on the public demo.
+
+Backfill is deliberately narrow. Only GitHub-linked accounts are provably
+verified: ``_fetch_github_claims`` raises unless the account has a
+primary+verified email, so a GitHub-linked row with an email cannot have been
+created from an unverified address. OIDC/Google/Microsoft rows are left alone —
+their ``email_verified`` claim is per-provider and older rows predate the H-30
+gate, so asserting verification for them would be a guess. Those converge on
+next login instead (the returning-user refresh added alongside this migration).
+
+Revision ID: 0028_oauth_email_verified_backfill
+Revises: 0027_source_format_parquet
+Create Date: 2026-07-22
+"""
+
+from typing import Sequence, Union
+
+from alembic import op
+
+revision: str = "0028_oauth_email_verified_backfill"
+down_revision: Union[str, None] = "0027_source_format_parquet"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.execute(
+        """
+        UPDATE catalog.users u
+        SET email_verified = true
+        WHERE u.auth_provider = 'oauth'
+          AND u.email IS NOT NULL
+          AND u.email_verified = false
+          AND EXISTS (
+              SELECT 1
+              FROM catalog.oauth_accounts a
+              JOIN catalog.oauth_providers p ON p.id = a.provider_id
+              WHERE a.user_id = u.id
+                AND p.provider_type = 'github'
+          )
+        """
+    )
+
+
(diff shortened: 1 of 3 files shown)


ishiland:
@codex review

PR #632, opened 2026-07-22
fix(jobs): fail queue rows whose worker died mid-job (#624)

### LINKED ISSUE #624: Procrastinate jobs stranded in 'doing' forever after a worker restart mid-job
## What happened

The public demo has a `catalog.procrastinate_jobs` row (an `ingest_service` task, job id 181) that has been in status **`doing` since 2026-06-29 23:04** — three weeks. Its worker died/restarted mid-job and nothing ever reaped the queue row.

## Current behavior

- The user-facing side is handled: the stale-job reaper in `backend/app/platform/jobs/worker.py` marks `catalog.ingest_jobs` rows as failed ("Stale: running for over 60 minutes" / "pending for over 1 hour"), so the UI recovered.
- The **procrastinate row leaks**: nothing transitions `doing` jobs whose worker is gone. Queue depth/metrics and any admin queue views count phantom in-flight jobs indefinitely, and each worker restart mid-job adds another.

## Fix direction

Procrastinate 3.x tracks worker heartbeats (the `procrastinate_workers` table is present in our schema). Use its stalled-job facility (`app.job_manager.retrieve_stalled_jobs()` or the periodic stalled-job sweep — check the procrastinate docs for the pinned version) to detect `doing` jobs whose worker heartbeat is dead. **Fail them rather than requeue** — ingest tasks are not idempotency-audited, and failing matches the verdict the ingest_jobs reaper already delivered to the user.

Natural home: the existing reaper loop in `backend/app/platform/jobs/worker.py`, which already runs periodically and owns the "stale" concept.

## Reproduction

1. `make dev`, start a slow ingest (large file or a service ingest against a slow endpoint).
2. `docker compose kill worker` mid-job, then `docker compose up -d worker`.
3. Observe: `select id, status from catalog.procrastinate_jobs where status='doing'` — the old row stays `doing` forever while the new worker ignores it.

## Acceptance criteria

- [ ] A `doing` job whose worker heartbeat is stale is transitioned to `failed` within one reaper cycle.
- [ ] The corresponding `catalog.ingest_jobs` row (if still pending/running) gets the same stale treatment it does today.
- [ ] Test simulating a dead worker (insert a `doing` row with a nonexistent/stale worker id, run the sweep, assert transition).
- [ ] One-off cleanup note for operators (the demo's job 181) — the sweep itself should handle it on first deploy.

ishiland:
Closes #624

## Problem

A worker killed mid-job leaves its `catalog.procrastinate_jobs` row in **`doing` forever**. The demo carried one (job 181, `ingest_service`) from 2026-06-29 — three weeks.

The user-facing half was already handled: the ingest_jobs reaper fails the row with *"Stale: running for over 60 minutes"*, so the UI recovered. But nothing ever transitioned the **queue** row, so queue depth and any admin queue view count phantom in-flight work — one more per worker kill, accumulating forever.

## Fix

Procrastinate 3.x tracks worker heartbeats, so *"the worker is gone"* is a fact we can read rather than a timeout we have to guess at. `fail_stalled_queue_jobs()` sweeps jobs whose worker hasn't heartbeat in `STALLED_WORKER_SECONDS` (300s — 30 missed beats at the default 10s interval) and prunes the dead worker rows behind them.

Deliberate choices:

- **Fail, don't requeue.** Ingest tasks are not idempotency-audited, and failing matches the verdict the ingest_jobs reaper already delivered to the user.
- **Keep the row** (`delete_job=False`) — the post-mortem trail is the reason anyone noticed this.
- **300s, not the 30s default.** The cost of waiting is a stale metric; the cost of being wrong is failing live work. A graceful rolling restart keeps heartbeating and is never swept.
- **Inside the existing connector context, before the worker registers itself**, so a process can never sweep its own heartbeat.
- **Once per process, not per tenant** — procrastinate's queue tables are shared infrastructure, not RLS-partitioned like `ingest_jobs`.
- **Sweep failure is caught and logged.** It must never block a worker from starting.

## Verification — live, against a real database

Before writing any test, with a `doing` row pinned to a worker whose heartbeat was 5 days old:

```
warning  Failed stalled queue job — its worker stopped heartbeating
         procrastinate_job_id=328 task_name=...tasks_vector.ingest_file
info     Stalled queue sweep complete  jobs_failed=1 workers_pruned=1
```

```
 id  | status | worker_id        id |        last_heartbeat
-----+--------+-----------      ----+-------------------------------
 328 | failed |                   2 | 2026-07-22 18:07:30.954648+00
```

The row flipped `doing` → `failed`, the dead worker row was pruned, and **the concurrently-heartbeating live worker (#2) was left untouched** — the safety property that matters.

The unit tests then pin the decision logic: fail not requeue, keep the row, skip unpersisted jobs, no-op on a healthy queue.

## Ops note

The demo's job 181 needs no manual cleanup — the sweep transitions it on the first worker start after deploy.

## Verification commands

```
uv run pytest tests/test_stalled_queue_sweep_624.py tests/test_defer_orphan_guard.py \
  tests/test_layering.py tests/test_ingest.py -q          # 85 passed
uv run pytest tests/test_worker*.py tests/test_vrt_stale_sweep_gap002.py -q  # 60 passed
```

## Not included

Moving the sweep into a periodic loop. It runs at worker startup, which is exactly when a killed worker comes back — the scenario that creates these rows. A periodic sweep would add a timer for a condition that only arises at process start.

### DIFF
--- a/backend/app/platform/jobs/worker.py
+++ b/backend/app/platform/jobs/worker.py
@@ -395,6 +395,18 @@ async def main() -> None:
                     install_signal_handlers=True,
                     delete_jobs="successful",
                     shutdown_graceful_timeout=shutdown_timeout,
+                    # fix(#624 codex P1): MUST match STALLED_WORKER_SECONDS.
+                    # Procrastinate's own Worker.run() prunes workers silent for
+                    # longer than this. procrastinate_jobs.worker_id is ON DELETE
+                    # SET NULL, and select_stalled_jobs_by_heartbeat treats a
+                    # `doing` job with a NULL worker_id as stalled OUTRIGHT — no
+                    # heartbeat comparison, because there is no longer a heartbeat
+                    # to compare. So at the 30s default, a live worker that merely
+                    # stalls (DB blip, long GC) gets pruned by another worker's
+                    # startup, its in-flight jobs go worker_id=NULL, and the sweep
+                    # below fails them as stalled despite the 300s cushion. Equal
+                    # windows mean a NULL worker_id can only mean 300s of silence.
+                    stalled_worker_timeout=STALLED_WORKER_SECONDS,
                 )
             finally:
                 # Cancel inside the connector context — a sweep mid-query when
--- a/backend/app/platform/jobs/worker.py
+++ b/backend/app/platform/jobs/worker.py
@@ -173,6 +173,63 @@ async def _recover_stale_jobs_for_current_scope() -> None:
             )
 
 
+# fix(#624): a worker killed mid-job leaves its queue row in `doing` forever —
+# the demo carried one from 2026-06-29 for three weeks. Procrastinate 3.x tracks
+# worker heartbeats (default: every 10s), so a worker that has missed 30 of them
+# is gone, not slow; a graceful rolling restart keeps heartbeating and is never
+# swept. Deliberately generous — the cost of waiting is a stale metric, the cost
+# of being wrong is failing live work.
+STALLED_WORKER_SECONDS = 300
+
+
+async def fail_stalled_queue_jobs() -> int:
+    """Fail procrastinate rows whose worker died mid-job. Returns the count.
+
+    The ingest_jobs reaper above already gives the USER a verdict ("Stale:
+    running for over 60 minutes"), but nothing ever transitioned the queue row,
+    so queue depth and admin views counted phantom in-flight work — one more per
+    worker kill, accumulating forever.
+
+    Fail rather than requeue: ingest tasks are not idempotency-audited, and
+    failing matches the verdict the user already got.
+
+    Caller must hold an open connector (``task_app.open_async()``). Runs once per
+    process rather than once per tenant — procrastinate's queue tables are shared
+    infrastructure, not RLS-partitioned like ingest_jobs.
+    """
+    from procrastinate.jobs import Status
+
+    from app.processing.ingest.tasks import task_app
+
+    manager = task_app.job_manager
+    stalled = list(
+        await manager.get_stalled_jobs(seconds_since_heartbeat=STALLED_WORKER_SECONDS)
+    )
+    for job in stalled:
+        if job.id is None:  # unpersisted job — nothing to transition
+            continue
+        await manager.finish_job_by_id_async(
+            job_id=job.id, status=Status.FAILED, delete_job=False
+        )
+        log.warning(
+            "Failed stalled queue job — its worker stopped heartbeating",
+            procrastinate_job_id=job.id,
+            task_name=job.task_name,
+        )
+    # Drop the dead workers' rows too, so the heartbeat table doesn't grow one
+    # tombstone per killed worker.
+    pruned = await manager.prune_stalled_workers(
+        seconds_since_heartbeat=STALLED_WORKER_SECONDS
+    )
+    if stalled or pruned:
+        log.info(
+            "Stalled queue sweep complete",
+            jobs_failed=len(stalled),
+            workers_pruned=len(pruned),
+        )
+    return len(stalled)
+
+
 async def _registered_tenant_ids_for_recovery() -> list[str]:
     """Read the global tenant registry without touching an RLS child table."""
     from app.core.db import async_session
@@ -297,6 +354,13 @@ async def main() -> None:
         # worker service can pin WORKER_QUEUES=raster on multi-core hosts.
         queues = [q.strip() for q in settings.worker_queues.split(",") if q.strip()]
         async with task_app.open_async():
+            # fix(#624): inside the connector context (it needs an open pool) and
+            # before the worker registers itself, so this process's own heartbeat
+            # can never be in the window it sweeps.
+            try:
+                await fail_stalled_queue_jobs()
+            except Exception:  # broad: a sweep failure must not block the worker
+                log.warning("Stalled queue sweep failed", exc_info=True)
             await task_app.run_worker_async(
                 queues=queues,
                 concurrency=settings.worker_concurrency,
--- a/backend/app/platform/jobs/worker.py
+++ b/backend/app/platform/jobs/worker.py
@@ -14,7 +14,7 @@ from pathlib import Path
 
 import structlog
 import uvicorn
-from sqlalchemy import func, text, update
+from sqlalchemy import func, select, text, update
 
 from app.core.config import settings
 from app.core.logging_config import setup_logging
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

ishiland:
Good catch — confirmed and fixed in \`77c2d3fd3\` (also rebased onto main).

You're right that the startup-only sweep is always one restart behind: under \`restart: unless-stopped\` the replacement worker is up seconds after the crash, while the dead worker's heartbeat is still fresh, so the startup pass skips the very row it exists to reap — and `run_worker_async` only fetches `todo` jobs afterwards, so nothing looks again once that heartbeat crosses 300s. My "not included: a periodic loop" note in the PR description had exactly the wrong reasoning.

Keeps the startup pass (it clears rows stranded before this process existed) and adds a 60s sweep loop for the life of the worker. The loop is created inside the connector context and cancelled there too — cancelling after the pool closed would surface a sweep mid-query as shutdown noise.

**Verified against your scenario specifically, no restart involved.** With the worker running and heartbeating, I stranded a `doing` row on a worker whose heartbeat was 10 minutes old:

```
18:19:14  insert job 329 -> status=doing, worker_id=4 (heartbeat 10m old)
18:19:32  warning  Failed stalled queue job — its worker stopped heartbeating
                   procrastinate_job_id=329 task_name=...ingest_file
          info     Stalled queue sweep complete  jobs_failed=1 workers_pruned=1
18:19:34  job329=failed
```

The live worker (id 3, heartbeat current) was left alone — only the dead row was pruned. That reap came from the periodic loop; the old code would have left it until the next restart.

Also added a test that drives a raising first sweep and asserts the loop keeps going: now that the loop owns the reap, "one bad sweep kills the loop" would be a silent regression back to startup-only behavior. It's bounded, so that regression fails in 5s instead of hanging.

@codex review

ishiland:
Confirmed against the procrastinate source and fixed in `64957d679`.

The chain is exactly as you describe, and the SQL is the part I had not read:

```sql
-- select_stalled_jobs_by_heartbeat
WHERE job.status = 'doing'
  AND (job.worker_id IS NULL OR sw.id IS NOT NULL)
```

A `doing` job whose worker row is gone is stalled **outright** — no heartbeat comparison, because there is no heartbeat left to compare against. And `Worker.run()` opens with `prune_stalled_workers(self.stalled_worker_timeout)`, which we were leaving at the 30s default. So the 300s cushion was defeated by a window I never set: a live worker that merely stalls past 30s gets pruned by another worker's startup, its in-flight jobs go `worker_id=NULL`, and my sweep fails live work.

Worth noting this is not hypothetical for this deployment — `WORKER_QUEUES` exists precisely so a raster worker can run beside the default one, which is the two-worker shape the race needs.

Fixed by passing `stalled_worker_timeout=STALLED_WORKER_SECONDS` so both windows agree: a NULL `worker_id` can then only mean 300s of silence, which is already this module's definition of dead. The NULL branch stays load-bearing — it is what reaps rows whose worker row is long gone, i.e. the original three-week-old demo case.

Guarded with a source assertion rather than a mock, since the call sits in `main()` behind bootstrap and what must not regress is the kwarg staying bound to that same constant — precisely what a future "drop the redundant argument" tidy-up would undo. Worker restarted clean on the dev stack with the kwarg in place (procrastinate accepts it; a bad name would fail at runtime, not in tests).

@codex review

ishiland:
Confirmed in the source and fixed in `ddefb70eb`. `Worker._shutdown()` opens with `cancel_and_capture_errors(side_tasks)` — the heartbeat is one of those side tasks — and only *then* does `asyncio.wait(self._running_jobs, timeout=self.shutdown_graceful_timeout)`. So a worker deliberately finishing a long job during a rolling restart is silent for the entire graceful window, exactly as you describe.

Our default `WORKER_SHUTDOWN_TIMEOUT` is 30s, well under the 300s floor, so the default deployment was never exposed — but the knob is env-configurable and an instance doing long COG conversions is precisely the one that would raise it.

Fixed by deriving the threshold instead of hardcoding it:

```python
def stalled_worker_seconds() -> int:
    return max(
        _STALLED_WORKER_FLOOR_SECONDS,                                    # 300
        settings.worker_shutdown_timeout + _STALLED_SHUTDOWN_MARGIN_SECONDS,  # +60
    )
```

All three consumers now read from it — the sweep's `get_stalled_jobs`, its `prune_stalled_workers`, and `run_worker_async`'s `stalled_worker_timeout` — so your round-2 P1 invariant (procrastinate's prune window must equal ours) survives this change rather than being re-broken by it. Tests cover both ends: default stays at the floor, and a 600s graceful window pushes the threshold past it.

Thanks for the three rounds on this one — the first two were genuine bugs in my fix (startup-only was one restart behind; the NULL-worker_id branch bypassed the cushion entirely), and this one closed the last window where the sweep could outvote an operator's explicit configuration.

@codex review

ishiland:
Right about the mechanics again — a global sweep cannot derive its threshold from one process's config, and nothing in procrastinate's schema exposes another worker's shutdown window, so no amount of local arithmetic reaches the fleet.

I went a different way than the suggested remedy, though. A fleet-wide `WORKER_STALLED_TIMEOUT` knob moves the failure rather than removing it: a fleet where someone forgets to set it identically breaks exactly the same way, and now there's a knob to get wrong.

Fixed in `fab30f96e` by flooring the window at `JOB_TIMEOUT_SECONDS` — the 60 minutes this module *already* calls stale for `ingest_jobs`:

```python
return max(
    JOB_TIMEOUT_SECONDS,                                              # 3600
    settings.worker_shutdown_timeout + _STALLED_SHUTDOWN_MARGIN_SECONDS,
)
```

Two things fall out of that:

1. **The split-queue race disappears.** A raster worker's 600s graceful window sits far under an hour, so a general worker running the global sweep cannot classify it as stalled — regardless of which pool's config either process holds, and with no coordination required.
2. **Past an hour, failing is correct, not racy.** The ingest_jobs reaper has already failed the user-facing row with "Stale: running for over 60 minutes". Failing the queue row then is the *consistent* verdict — which is what this sweep exists to deliver. That's the principled stopping point: beyond the reaper's own window there is no live work left to protect.

Cost is a phantom queue row lingering up to an hour instead of five minutes — against a row that sat for three weeks. The local graceful window still raises the threshold past the floor (your r3 point), for the operator who configures a shutdown longer than the reaper's hour.

@codex review

ishiland:
Correct, and it invalidates the justification I gave you in round 4. I claimed that past the reaper's hour there is no live work left to protect. But the two heartbeats are independent signals: `Worker._shutdown` cancels the *worker* heartbeat and only then waits out `shutdown_graceful_timeout`, while `maintain_ingest_job_heartbeat` keeps renewing `ingest_jobs.heartbeat_at` for the whole window. The ingest row is therefore *not* stale, the reaper leaves it alone, and my sweep fails a job that is still running. My "consistent verdict" argument was simply wrong.

Rounds 3, 4 and 5 are all the same shape — a live worker looking silent — so I stopped tuning the timeout and read the signal that actually tracks the work (`b03e099dc`):

```python
alive = await _ingest_jobs_still_leasing(stalled)
...
if str(kwargs.get("job_id")) in alive:
    continue   # a timeout infers death; a lease observes life
```

The threshold stays as the backstop for jobs with no lease to read — non-ingest tasks, and hosted deployments where `ingest_jobs` is FORCE-RLS and this un-tenanted query returns nothing. Both documented at the call site.

**Verified on the dev stack, both directions.** A `doing` row pinned to a worker silent for three hours, ingest row leasing:

```
18:57:14  Skipping stalled queue job — its ingest job is still leasing  job=330
18:57:14  Stalled queue sweep complete  jobs_failed=0 jobs_skipped_alive=1
18:58:14  Skipping stalled queue job ...
18:59:14  Skipping stalled queue job ...
```

Then ageing that lease to two hours old:

```
18:59:33  lease aged
19:00:33  job330=failed
```

So it protects live work without becoming a blanket amnesty. And the pre-fix worker, still running during the container swap, failed that same row out from under its live lease — your reported bug, reproduced by accident.

One more thing the live run caught that the mocks could not: procrastinate's `Job` exposes the payload as **`task_kwargs`**, not `args` (the DB column is `args`; `Job.from_row` renames it). My first real `Job` raised `AttributeError` while every mocked test passed. Test helper now models the real attribute — worth flagging since it is the kind of thing a mock-only suite will happily bless forever.

Thanks for the five rounds. Three of them were real bugs in my own fix, and this last one taught me the reasoning error behind two of them.

@codex review

ishiland:
Fixed in `dd7d99e22` — and you're right that this shouldn't have been an accepted limitation. I'd documented the un-tenanted SELECT as a known gap, but `defer_async_with_tenant` already threads `tenant_id` into the job kwargs (see `tenant_task`), so the context was sitting in the row I'd already fetched.

Worse than a gap, as you note: under FORCE-RLS the guard *inverts*. Every hosted lease reads as dead, so the check I added to prevent failing live work would cause it.

Now grouped by tenant, each group's lookup inside `tenant_job_context`. Single-tenant jobs carry no `tenant_id` and take the plain path, where `tenant_job_context` is a hard no-op anyway.

On the fail-closed half: a failed lookup now means liveness is **unknown**, and unknown is treated as **alive** — an unreaped row is a stale metric, a wrongly reaped one is lost work. Caught per tenant, mirroring `recover_stale_jobs`' own per-tenant isolation in this same module, so one tenant's broken lookup can't strand the rest of the fleet (previously it would have aborted the entire sweep, which was your point).

Re-verified live after the restructure, both directions:

```
19:10:52  Skipping stalled queue job — its ingest job is still leasing  job=331
19:10:52  Stalled queue sweep complete  jobs_failed=0 jobs_skipped_alive=1
          ... survives every cycle while the lease is fresh ...
19:11:47  lease aged to 2h
19:12:07  job331=failed
```

@codex review

Issue #621, opened 2026-07-22
Map surfaces never recover after tile-token/session expiry — endless 403 tile retries with no re-auth or error UI


ishiland:
## What happened (observed on the public demo, 2026-07-21)

A real demo user (SSO signup) uploaded two private vector datasets and viewed them on a map. Tiles rendered fine for ~45 minutes, then **every tile request returned 403 for the next ~7 hours** (~290 x 403s across two devices), while the user repeatedly panned/returned to the page. No recovery, no visible error — the map just silently stopped rendering data.

Access-log timeline (times UTC, user identifiers omitted):

- `17:02–17:12` — datasets created (visibility `private`, status `published`)
- `17:03–17:44` — `/api/tiles/data.<table>/{z}/{x}/{y}.pbf` served 200/204 (valid HMAC sigs; log format strips query strings)
- `17:29` — **last successful `POST /api/tiles/tokens/` mint from this client**
- `17:49` — first 403 (`Signature required` / `Invalid or expired signature` branch)
- `17:49 → 00:30` — continuous 403s: MapLibre revalidates on the ~5-min tile TTL cadence, plus large bursts whenever the user interacted (84, 68, 48, 46 per hour). Second device (same user) hit the same wall.
- Meanwhile `POST /api/auth/refresh/` shows **3 simultaneous calls every ~14 min with mixed results, e.g. `200 + 401 + 401`** — the pattern of multiple tabs/contexts racing a rotating refresh token (`service.rotate_refresh_token`, `backend/app/modules/auth/router.py:238`): one caller wins, losers are stranded with dead tokens.

An authenticated admin session fetching the same tiles at the same time got 200s, so server-side access control is behaving as designed.

## Current behavior per surface (verified in code)

- **BuilderMap** (`frontend/src/components/builder/BuilderMap.tsx:129` GUARD-03): on a vector tile 401/403, re-signs the source once from the current token cache and retries; falls through to an auth-error toast. Covered by `BuilderMap.tile-retry.test.tsx`. But the re-sign uses the *cached* token — if the session is stranded (refresh 401), the retry re-uses a stale sig and the toast is the end of the road.
- **ViewerMap** (`frontend/src/components/viewer/ViewerMap.tsx:441-464`): deduped error toast ("Map tile error — some layers may not display correctly") — no re-mint, no stop; MapLibre keeps revalidating on the tile-TTL cadence.
[...]

Issue #401, opened 2026-07-04
Anonymous-capable read endpoints: model optional bearer + preserve 401 on stale creds (systemic)


ishiland:
Surfaced during codex review of #400. Two related, systemic characteristics of the anonymous-capable read path (`get_optional_user` → `check_dataset_access_or_anonymous`). Both are pre-existing and span multiple endpoints — deliberately left out of #400 (a clustering bugfix) to keep its SDK surface small.

### 1. OpenAPI models optional bearer as *required*
Every `get_optional_user` read op emits `security: [{OAuth2PasswordBearer: []}]`, so generated SDK methods require `AuthenticatedClient` even though the endpoint serves public data anonymously. Verified on `/datasets/{id}/features.geojson`, OGC `/collections/datasets/items`, and `/datasets/{id}`.

**Fix:** model these as `security: [{}, {OAuth2PasswordBearer: []}]` (anonymous allowed) at the source and regenerate SDKs, as one cross-cutting change. A generated `AuthenticatedClient` with a dummy token otherwise can't reach the anonymous path (hits the 401 branch).

### 2. 401-vs-404 on supplied-but-invalid credentials is inconsistent
#400 taught `features.geojson` to return 401 (not the anonymous 404) when credentials were *supplied* but failed to resolve — so the frontend's refresh-on-401 retry fires instead of a private layer permanently 404ing after a stale/revoked token. The OGC and STAC read routers have the identical latent issue (they go straight to the anonymous path, so stale-token private access 404s and never refreshes).

**Fix:** apply `request_carries_credentials()` (added in #400, `auth/dependencies.py`) to the OGC/STAC read handlers too, so the whole read path behaves consistently.

_Not launch-blocking; demand-gate against real SDK/stale-token reports._

ishiland:
Re-verified against current code — both parts still reproduce, keeping demand-gated. Notes for future-me on which half to reach for first and how to fix each.

**Priority: Part 1 is the launch-relevant, higher-impact half.** "Generate an SDK, hit the anonymous demo" is the pitch, and that path 401s with a dummy token because the generated client requires `AuthenticatedClient`. If a report ever fires, it's most likely this one. Part 2 only bites authenticated users with a stale token hitting OGC/STAC specifically, and it's frontend-mediated.

**Evidence (current `openapi.json` / code):**
[...]

PR #637, opened 2026-07-22
fix(standards): preserve 401 on stale credentials in OGC/STAC read paths

ishiland:
Closes the remaining half of #401. (Part 1 — the anonymous-alternative OpenAPI security modeling — already shipped in #506 via `_normalize_security_contract`; 61 ops in the committed snapshot model `security: [{}, ...]`.)

## Problem

A request that *supplied* credentials which failed to resolve (expired/revoked JWT, bad API key) fell through to the anonymous path on the OGC and STAC read handlers. A stale-token caller's private dataset then 404'd permanently — the frontend's refresh-on-401 retry never fired. `features.geojson` got this guard in #400; the OGC/STAC routers had the identical latent issue, and the demand-gate fired with the real stale-token report in #621.

## Change

- New `get_optional_user_or_401` dependency in `auth/dependencies.py`: wraps `get_optional_user`, raises 401 (`WWW-Authenticate: Bearer`) when `request_carries_credentials()` is true but resolution returned `None`. Truly credentialless requests still resolve to `None` and keep the public path.
- Swapped into the 3 OGC + 5 STAC anonymous-capable read handlers (`/collections/{id}`, `/collections/{id}/items`, `/collections/{id}/items/{fid}`; STAC collection items, item×2, search GET/POST).
- The STAC no-schema discovery trio (`get_optional_user_no_security_schema`: landing page, collections, collection) deliberately keeps the lenient public path — those are top-level discovery surfaces for STAC browsers.
- `features.geojson` keeps its inline guard: its 401 check must stay *behind* the embed-token fallback, so the shared dependency can't replace it.

## API/schema impact

None. The wrapper keeps `get_optional_user` in the dependant tree, so `_normalize_security_contract` stamps the same security contract — `make openapi-check` passes with zero snapshot diff, no SDK regen. Behavioral change only: supplied-but-invalid credentials on these 8 read ops now get 401 instead of the anonymous outcome (404 or public-subset 200), per RFC 6750.

## Verification

- `uv run pytest tests/test_ogc_public_access.py tests/test_stac_visibility.py` — 35 passed (9 new: parametrized stale-bearer 401s, invalid `X-Api-Key`, STAC search POST)
- `uv run pytest tests/test_features_geojson_z.py tests/test_layering.py tests/test_ogc_features.py tests/test_stac_api.py` — 86 passed
- `make openapi-check` — clean
- Live smoke on the dev stack (rebuilt `api` container): stale bearer → 401 on OGC items and STAC search; anonymous → 200 unchanged; valid admin token → 200.

Part of #401.

### DIFF
--- a/backend/app/modules/auth/dependencies.py
+++ b/backend/app/modules/auth/dependencies.py
@@ -202,6 +202,27 @@ def request_carries_credentials(request: Request) -> bool:
     )
 
 
+async def get_optional_user_or_401(
+    request: Request,
+    user: Annotated[Identity | None, Depends(get_optional_user)],
+) -> Identity | None:
+    """``get_optional_user``, but supplied-yet-unresolvable credentials get 401.
+
+    fix(#401): the OGC/STAC read handlers resolved a stale/revoked token to the
+    anonymous path, so a credentialed caller's private dataset 404'd instead of
+    401ing and the client's refresh-on-401 retry never fired. Use this on
+    anonymous-capable read endpoints; truly credentialless requests still
+    resolve to ``None`` and keep the public path.
+    """
+    if user is None and request_carries_credentials(request):
+        raise HTTPException(
+            status_code=status.HTTP_401_UNAUTHORIZED,
+            detail="Not authenticated",
+            headers={"WWW-Authenticate": "Bearer"},
+        )
+    return user
+
+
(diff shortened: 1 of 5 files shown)


ishiland:
@codex review

PR #638, opened 2026-07-23
feat(admin): resolve resource names in the audit log viewer

### LINKED ISSUE #620: Add dataset name in the audit logs
### Problem or motivation

Currently, the audit logs only show the first 8 characters of the resource ID. Hovering on it does not show the full resource ID. Searching datasets using the first 8-character resource ID on the Search page does not return any results. Therefore, it is diffcult to see which dataets were accessed. 

Image

### Proposed solution

I suggest adding a Resource NAME column so that it is easier to see which datastets are accessed. Additionally, having hyperlinks for RESOURCE NAME or RESOURCE ID columns would be very helpful. 

### Alternatives considered

_No response_

### Area

Admin Panel

### Additional context

_No response_

ishiland:
Closes #620 (reported by @giswqs — thanks!)

## Problem

Audit log rows only showed the first 8 characters of the resource UUID, with no way to see the full ID or tell which dataset a row referred to. Searching the catalog by the truncated ID returned nothing, so correlating access events with datasets was effectively manual.

## Changes

**Backend**
- `GET /admin/audit-logs/` now returns a `resource_name` per row, batch-resolved for the current page only (max 4 small `IN` queries per request): dataset → record title, map → map name, collection → name, user → username. Deleted/unknown resources resolve to `null`.
- Map names are fetched via the stable `get_map_names` sharing API — the open-core layering guard (`test_open_core_decomposition_boundaries_stay_clean`) forbids the audit module importing catalog map ORM internals.

**Frontend**
- New **Name** column in the admin audit log table; the name links to the dataset/map/collection detail page when a route exists (users render as plain text — there's no per-user detail route).
- The truncated Resource ID cell now shows the full UUID on hover (native `title` tooltip).
- `audit.table.resourceName` added to all four locales (en/es/fr/de).

**Regenerated artifacts:** `backend/openapi.json`, `frontend/src/types/api.generated.ts`, Python/TS SDK models.

## Scope notes

- CSV/JSON exports now include `resource_name` too (second commit): rows buffer in 500-row chunks and names resolve through the same batched helper as the list endpoint, on a second session (the stream's session holds a server-side cursor and cannot run statements mid-iteration) — the 100k-row streaming design is preserved. CSV appends the column last so positional consumers of the old layout keep working.
- Searching audit logs *by* resource name is out of scope; the visible name column addresses the discoverability gap in the issue.

## Verification

- `backend`: `uv run pytest tests/test_audit.py tests/test_layering.py` — 49 passed (includes a new test asserting dataset title resolution and `null` for deleted resources)
- `frontend`: `npm run typecheck`, `npm run test:i18n`, `npx vitest run src/components/admin src/pages/admin` (85 passed), eslint clean
- `make openapi` + `make sdks` regenerated; drift is committed
- Live smoke on the dev stack: API returns resolved names (`dataset.view` → "World Rivers & Lake Centerlines (Natural Earth 1:50m)", deleted map → `null`); admin UI renders the linked Name column

## UI check

Verified in the local admin UI — the Name column renders linked dataset titles (screenshot available on request; the audit page smoke-run is described above).


ishiland:
@codex review

ishiland:
@codex review

giswqs:
Thank you for the quick fix.

Issue #585, opened 2026-07-19
Builder contextual-menu review: consistency and discoverability findings


ishiland:
Audit of every contextual/dropdown menu in the map builder (layer-row kebab `StackRow.tsx`, folder-group kebab `FolderGroupRow.tsx`, basemap kebab `BasemapGroupRow.tsx`, title-bar overflow `MapTitleBar.tsx`, bulk-action overflow `BulkActionBar.tsx`). i18n is clean (all strings through t(), all four locales). Findings, ranked:

1. **No right-click context menu anywhere** — right-clicking a layer row or the canvas does nothing; every action hides behind a hover-revealed kebab (`opacity-0 group-hover`, StackRow.tsx:379-384). No Shift+F10/Menu-key handler either. Biggest discoverability gap.
2. **Single vs bulk grouping rules disagree** — StackRow offers "Add to group…" for any non-grouped layer incl. raster/DEM (StackRow.tsx:516-536); BulkActionBar.canGroup requires `dataset_record_type==='vector_dataset'` (BulkActionBar.tsx:93-103).
3. **Disabled items give no reason** — bulk Group/Ungroup/Apply-style (BulkActionBar.tsx:345-366) and Paste style on geometry mismatch (StackRow.tsx:491-500) disable silently.
4. **Ragged icons in the layer kebab** — Zoom/Copy/Paste have Lucide icons; Rename/Duplicate/Delete none; Add-to-group uses hardcoded `▸`/`＋` glyphs.
5. **"Add to group" is a flat inline list** rather than a `DropdownMenuSub` — grows unbounded with many groups.
6. **Basemap kebab asymmetric** — "Reset appearance" is never disabled and has no dirty-check/confirm (BasemapGroupRow.tsx:213).
7. **Three different delete-confirmation patterns** across StackRow / FolderGroupRow / BulkActionBar.
8. **"Duplicate map" has no in-flight feedback** (MapTitleBar.tsx:219) — disabled only, no spinner.

Suggested first slice: #1 (row-level `onContextMenu` opening the existing kebab menu) + #2 (unify the grouping predicate) + #3 (tooltip reasons on disabled items).

PR #639, opened 2026-07-23
feat(builder): right-click context menus, unified grouping predicate, disabled-item reasons (#585)

ishiland:
First slice of #585 (findings 1–3, the issue's suggested first slice).

## What changed

**1. Right-click context menu on stack rows** (finding 1)
`useKebabContextMenu` in `row-chrome.tsx` makes each row's existing kebab `DropdownMenu` controlled: right-click, Shift+F10, or the ContextMenu key on a `StackRow`, `FolderGroupRow`, or `BasemapGroupRow` opens it. The basemap row honors the existing multi-selection boundary (Phase 1051 CR-02). The kebab trigger gains `data-[state=open]:opacity-100` so the anchor is visible while the menu is open.

**2. Unified grouping predicate** (finding 2)
`canJoinFolderGroup` in `folder-groups.ts` is now the single rule: not a group row, not already in a group. The `dataset_record_type === 'vector_dataset'` requirement existed in **two** places — `BulkActionBar.canGroup` and the defense-in-depth filter in `use-bulk-layer-actions.ts` `handleBulkGroup` (the issue only caught the first) — and disagreed with StackRow's "Add to group…" and drag-and-drop membership (`resolveDropGroupMembership`), which have always accepted raster/DEM rows with no record-type guard. The now-unreachable `toasts.bulkGroupSkippedType` key is removed from all four locales.

**3. Inline reasons on disabled menu items** (finding 3)
Bulk Group / Ungroup / Apply-style and StackRow Paste-style show a muted one-line reason under the label when disabled. Inline text instead of the issue's suggested tooltips because tooltips don't fire on disabled elements.

4 new i18n keys added to all four locales (en/es/fr/de).

## Verification

- `npx vitest run` — 339 files / 3713 tests pass (incl. new right-click + Shift+F10 StackRow tests; flipped `BulkActionBar` Test 6 and bulk-ops Test 11 from "raster blocks grouping" to "raster groups successfully")
- `npm run typecheck`, `npm run lint`, `npm run test:i18n`, `npm run build` — clean
- Live smoke on the dev stack ("Restless Earth"): right-click opens the menu on layer/group/basemap rows; disabled Paste-style and Ungroup show their reasons; bulk-grouped a vector heatmap + the ETOPO raster, saved, reloaded — the group with the raster persists (confirming the record-type restriction was the outlier, not a backend constraint); ungrouped and re-saved to restore the map.

## Not in this slice (remaining #585 findings)

Ragged kebab icons (4), `DropdownMenuSub` for the group list (5), basemap Reset-appearance dirty-check (6), unified delete-confirm pattern (7), duplicate-map spinner (8).

Closes nothing — #585 stays open for the remaining findings.

### DIFF
--- a/frontend/src/components/builder/BasemapGroupRow.tsx
+++ b/frontend/src/components/builder/BasemapGroupRow.tsx
@@ -13,6 +13,7 @@ import {
   DragGripButton,
   STACK_ROW_GRID,
   rowStateClasses,
+  useKebabContextMenu,
   type DragHandleProps,
 } from '@/components/builder/row-chrome';
 
@@ -51,6 +52,7 @@ export const BasemapGroupRow = memo(function BasemapGroupRow({
   isMultiSelectionActive = false,
 }: BasemapGroupRowProps) {
   const { t } = useTranslation('builder');
+  const kebabMenu = useKebabContextMenu();
 
   // Phase 1051 IN-01: single source for the display name string. Used by
   // aria-label (visibility toggle), aria-label (kebab trigger), and the visible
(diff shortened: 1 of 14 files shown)


ishiland:
@codex review

PR #648, opened 2026-07-23
fix(ai,datasets): repair schema-invalid map specs, surface 3D metadata (#642, #647)

### LINKED ISSUE #642: AI map generation: LLMMapSpec pydantic validation failures are not repaired/retried — parse errors get a retry, shape errors don't
**Priority: P2** — AI map generation fails hard on schema-invalid LLM output; a real demo user hit it twice in a row and gave up.

## What happened

Demo VM, 2026-07-23 (36h window): two `ValidationError: 2 validation errors for LLMMapSpec` tracebacks in the api logs, both from one user's non-English session ("Yeongwol-gun" boundary map, Korean admin region):

```
Field required [type=missing, input_value={'title': 'Yeongwol-gun i...n boundary': '#FF0000'}}, ...]
Field required [type=missing, input_value={'name': 'Yeongwol-gun bo...
```

The LLM returned parseable JSON that was missing required `LLMMapSpec` fields, and the user got the raw error event both times.

## Root cause

`backend/app/processing/ai/service.py` (map generation, ~line 683–700): there is a recovery path for **parse** failures — `_parse_map_spec` → `_retry_parse_map_spec` ("retry extraction only") — but the subsequent pydantic construction is unguarded:

```python
spec = LLMMapSpec(**spec_dict)   # ValidationError propagates straight to the user
```

So "invalid JSON" gets a second chance, but "valid JSON, wrong shape" — the more common LLM failure mode — does not. Both `generate_map_from_prompt` and `stream_generate_map` share this.

## Suggested fix

- Catch `ValidationError` and run one repair round: feed the validation errors + the original spec back to the model ("fix these fields, return JSON only"), mirroring the existing parse retry (bounded, one attempt).
- On final failure, emit a friendly message ("The AI couldn't produce a valid map for this prompt — try rephrasing") instead of surfacing the pydantic error.
- Consider logging which fields go missing to spot systematic prompt/schema drift (both failures here were `Field required` misses, possibly correlated with non-English prompts).

### LINKED ISSUE #647: Dataset detail API returns null for is_3d/n_dims/z_min/z_max despite persisted values
**Priority: P3** — metadata is persisted correctly but never surfaces; found while smoke-testing #646.

## What happens

`DatasetDetailResponse` declares `is_3d`, `n_dims`, `z_min`, `z_max` (`backend/app/modules/catalog/datasets/domain/schemas.py:233` area), and the ingest pipeline persists them correctly on `catalog.datasets` (verified: earthquakes seed dataset and a fresh 3D GeoParquet ingest both have `is_3d=t, n_dims=3` in the DB). But the detail endpoint returns `null` for all four fields, for every dataset — nothing in the response assembly maps them from the `Dataset` row (`grep -rn "is_3d=" backend/app/modules/catalog/datasets/` has no hits; only the maps services map it, which is why 3D map rendering still works).

## Impact

Low — the 3D rendering path reads `is_3d` via the maps queries, so no user-visible 3D feature breaks. But any client, SDK consumer, or MCP reader relying on the documented dataset-detail contract gets `null` for fields the OpenAPI schema promises ("True if geometry has Z dimension"), and the frontend cannot show 3D metadata on the dataset page.

## Fix

Map `is_3d` / `n_dims` / `z_min` / `z_max` from the `Dataset` ORM row into the detail response where the other `datasets`-table fields (`srid`, `feature_count`, `source_format`, …) are populated. One-line-per-field plus a response assertion in the detail endpoint tests.

ishiland:
Fixes #642, fixes #647 — the backend half of the 2026-07-23 demo health review remediation.

## #642 — LLMMapSpec validation failures now get a repair round
`generate_map_from_prompt` and `stream_generate_map` had a recovery path for **parse** failures (`_retry_parse_map_spec`) but `LLMMapSpec(**spec_dict)` was unguarded — valid JSON with the wrong shape (the failure a real demo user hit twice: `Field required` for `name`/`layers`) propagated the raw pydantic error to the user.

- New `_repair_map_spec`: feeds the validation errors + original spec back to the model (bounded, one attempt, `tools=[] / max_rounds=1`, mirroring the parse retry).
- Final failure raises a friendly "try rephrasing" message instead of a pydantic traceback; missing field paths are logged (`fields=[...]`) to spot systematic prompt/schema drift.

## #647 — dataset detail API returns persisted 3D metadata
`DatasetResponse` documents `is_3d`/`n_dims`/`z_min`/`z_max` and ingest persists them, but `dataset_to_response` never mapped them — every dataset returned `null`. Now mapped from the Dataset row alongside the other datasets-table fields. No schema change, so no OpenAPI regen.

## Verification
- `pytest tests/test_ai_map_spec_repair.py tests/test_datasets.py::TestGetDataset` — 6 passed (3 new repair tests incl. second-failure + unparseable-output; new detail-endpoint 3D assertion).
- `pytest tests/test_ai_chat.py` — 9 passed; `tests/test_layering.py` — 29 passed.
- `ruff check` + `ruff format --check` clean.


ishiland:
@codex review

ishiland:
@codex review

Addressed the P2: `_repair_map_spec` now records its provider call via `record_token_usage` (subsystem `map_generation`) before accepting or rejecting the repaired spec, so repair tokens count toward the daily cap. The pre-existing `_retry_parse_map_spec` had the identical accounting gap, so it records too. Covered by two new tests (success and failed-repair both record).

PR #649, opened 2026-07-23
fix(frontend): gate owner-only VRT queries, self-heal stale shells (#644, #645)

### LINKED ISSUE #644: Dataset page fires owner-only VRT requests for anonymous viewers — guaranteed 401s on every public VRT dataset visit
**Priority: P3** — no data exposure (endpoints correctly 401), but every anonymous visit to a public VRT dataset page fires guaranteed-failing requests: console noise for visitors, log noise that camouflages real auth problems.

## What happened

Demo VM, 36h window: 21× `401 GET /api/datasets/{id}/vrt/generations/` plus a handful of 401s on `/vrt/status/` and `/vrt-sources/` — **all for the public showcase "swissALTI3D Matterhorn DEM" dataset**, i.e. anonymous (or expired-session) visitors simply viewing the dataset page. The pattern repeats in pairs roughly every 30 min as visitors browse.

## Root cause

`frontend/src/components/dataset/tabs/OverviewTab.tsx:182`:

```tsx
const { data: generationsData } = useVrtGenerations(isVrt ? dataset.id : '', { limit: 1 });
```

The fetch is gated only on `isVrt`, not on whether the current viewer could ever be authorized (the VRT management endpoints are owner/admin-scoped). `SourcesTab.tsx:66` has the same call plus `useVrtStatus`/`useVrtSources` — worth auditing whether that tab can render for non-owners too.

## Suggested fix

Gate the VRT queries on ownership/permission (e.g. `enabled: isVrt && canManageDataset`), the same way other owner-only UI is gated — or, if generation info is meant to be public for visible datasets ("last updated from source X"), make the read endpoint visibility-scoped instead (per AGENTS.md Rule 1, `check_dataset_access_or_anonymous`). Either resolution is fine; the current half-state (public page, owner-only fetch, unconditional call) is the bug.

### LINKED ISSUE #645: Stale SPA shells still 404 on hashed assets after deploy (15+ distinct IPs, growing) — add one-shot asset-failure auto-reload
**Priority: P2** — returning launch-window visitors are still landing on a broken app, and the count is *growing* day-over-day. #622's `no-cache` fix stops new shells from going stale, but it cannot heal the cohort that cached the old shell before the fix, and it does nothing for long-lived open tabs.

## What happened

Demo VM, 36h window: 20× `404 GET /assets/index-QvqZ23KQ.css` (a hashed asset from the pre-2026-07-22 build) from **15+ distinct client IPs** across several countries — 4 hits on 07-22 vs **16 on 07-23**, i.e. increasing a full day after the `Cache-Control: no-cache` fix went live. Live headers verified correct: `/` and deep routes both send `cache-control: no-cache`, `cf-cache-status: DYNAMIC`.

These are (a) visitors from the launch window whose browsers heuristically cached the old `index.html` and haven't revalidated yet, and (b) tabs left open across a deploy that lazy-load a route chunk/CSS that no longer exists. In both cases the page renders unstyled/blank with no recovery path — the worst possible second impression for a returning launch visitor.

## Why #622 can't cover this

`no-cache` on the shell only affects responses served *after* the fix. Heuristically-cached copies from before it remain "fresh" client-side until their heuristic TTL expires, and an already-open tab never refetches the shell at all. Nothing in the frontend detects a failed hashed-asset load.

## Suggested fix

One-shot self-healing on asset load failure:

- For lazy route chunks: catch dynamic-import failure (the classic `ChunkLoadError` pattern) and `location.reload()` once, latched via `sessionStorage` to prevent loops.
- For the initial CSS/JS `<link>`/`<script>`: an `onerror` handler in the shell that does the same latched reload — the reload goes through `no-cache` revalidation and picks up the current shell.
- Optionally keep the previous build's `assets/` alongside the current one in the image (last-1 grace) so mid-session navigations survive a deploy without a reload; the latched-reload alone is probably sufficient for the demo.

Complements #622 (cache-control on the shell) and closes the remaining stale-shell failure mode observed live.

ishiland:
Fixes #644, fixes #645 — the frontend half of the 2026-07-23 demo health review remediation.

## #644 — no more guaranteed 401s on public VRT dataset pages
The VRT management endpoints are owner/admin-scoped, but the dataset page fired them for every viewer (21× anon 401s on the public Matterhorn DEM in a 36h window):
- `OverviewTab` now gates `useVrtGenerations` on `canEdit` (the existing `isEditor && useCanMutate` gate that mirrors backend `check_dataset_write_access`).
- The Sources tab is hidden for non-owners (`showSources = isVrt && canEdit`): all three of its queries (`useVrtSources`/`useVrtStatus`/`useVrtGenerations`) are owner-scoped, so for a non-owner the tab could only ever render 401 noise.

## #645 — stale shells self-heal with a time-latched one-shot reload
#622's `no-cache` stops new shells going stale but can't heal pre-fix heuristic caches or tabs left open across a deploy (observed live: 16 asset-404s on 07-23, 15+ IPs, growing). Two complementary guards, both reloading at most once per 30s (sessionStorage latch, so a genuinely broken server can't loop):
- `public/asset-guard.js` — capture-phase `error` listener in the shell for the initial hashed `<script>`/`<link>` loads. External file, not inline, because the page ships a strict `script-src 'self'` CSP (SEC-020).
- `src/lib/stale-asset-reload.ts` — `vite:preloadError` handler for lazy route chunks, installed in `main.tsx`; suppresses the import throw when it reloads.

Forward-looking by nature: already-cached pre-fix shells can't contain the guard, but every shell served from this build on self-heals, which closes the growing cohort.

## Verification
- `npx vitest run` on the new latch tests + all dataset tab/page suites — 32 files, 219 passed.
- `npm run typecheck` + eslint on changed files — clean.
- No new `t()` keys (silent reload), so no locale-parity impact.

### DIFF
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -11,6 +11,11 @@
          the page carries a strict script-src 'self' CSP with no inline script.
          Render-blocking, so it still runs before first paint. -->
     <script src="/theme-init.js"></script>
+    <!-- fix(#645): one-shot reload when a stale shell's hashed assets 404 after a
+         deploy. External (not inline) for the SEC-020 script-src 'self' CSP; must
+         be in <head> before the hashed asset tags so the capture-phase listener
+         sees their error events. -->
+    <script src="/asset-guard.js"></script>
     <link rel="icon" type="image/svg+xml" sizes="any" href="/favicon.svg" />
     <link rel="icon" type="image/png" sizes="16x16" href="/favicon-16.png" />
     <link rel="icon" type="image/png" sizes="32x32" href="/favicon-32.png" />
--- a/frontend/public/asset-guard.js
+++ b/frontend/public/asset-guard.js
@@ -1,0 +1,27 @@
+// fix(#645): if the shell is stale (heuristically cached pre-#622, or a tab
+// left open across a deploy), its hashed /assets/ script/CSS 404 and the app
+// renders blank/unstyled with no recovery. Reload once so the no-cache shell
+// revalidates to the current build. External file, not inline — the page
+// ships a strict script-src 'self' CSP (SEC-020). Keep the latch key +
+// window in sync with src/lib/stale-asset-reload.ts (the same guard for
+// lazy route chunks).
+(function () {
+  window.addEventListener(
+    'error',
+    function (e) {
+      var el = e.target;
+      if (!el || !el.tagName) return;
+      var tag = el.tagName;
+      if (tag !== 'SCRIPT' && tag !== 'LINK') return;
+      var url = el.src || el.href || '';
+      if (url.indexOf('/assets/') === -1) return;
+      try {
+        var last = Number(sessionStorage.getItem('geolens-asset-reload-at') || 0);
+        if (Date.now() - last < 30000) return;
+        sessionStorage.setItem('geolens-asset-reload-at', String(Date.now()));
+      } catch (err) {}
+      window.location.reload();
+    },
+    true
+  );
+})();
--- a/frontend/src/components/dataset/panels/DetailPanel.tsx
+++ b/frontend/src/components/dataset/panels/DetailPanel.tsx
@@ -51,7 +51,9 @@ export function DetailPanel(props: DetailPanelProps) {
 
   const showData = isVector;
   const showStructure = isVector;
-  const showSources = isVrt;
+  // fix(#644): every query inside SourcesTab hits owner/admin-scoped VRT
+  // endpoints, so for non-owners the tab could only ever render 401 noise.
+  const showSources = isVrt && canEdit;
 
   const draftValues = useMemo(() => ({
     lineage_summary: resolveDraftValue('lineage_summary'),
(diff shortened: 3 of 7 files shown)


ishiland:
@codex review

ishiland:
@codex review

Addressed the P2: the gate is now `isAuthenticated` (`useAuthStore` token) rather than `canEdit` in both `OverviewTab` and `DetailPanel.showSources` — signed-in non-owners keep the read-only Sources view and generation info; only anonymous viewers (the observed 401 source) skip the queries. Mutation controls inside SourcesTab remain gated by `canEdit`.

ishiland:
@codex review

Addressed the r2 P2: `DetailPanel` now clamps `activeTab` to `overview` whenever the selected tab's trigger is hidden (`sources` for anonymous viewers, plus the pre-existing `data`/`structure`-on-raster deep-link variants), so a `#sources` deep link or a sign-out mid-Sources falls back to Overview instead of a blank panel.

Issue #643, opened 2026-07-23
API uvicorn worker OOM-killed at the container memory cap (~1.9 GB RSS) during a raster-proxy session — no recycling or memory observability


ishiland:
**Priority: P2 (investigation)** — first observed API OOM kill in production-like load; invisible to users this time, but it means one uvicorn worker can eat the whole container cap.

> **Re-scoped 2026-07-31.** Suggestions 2 and 3 have shipped (see "Already shipped"). What remains is suggestion 1 alone: reproduce and profile the raster-proxy path to find out what actually grew. This is open-ended investigation, not a scoped change, so it carries no milestone.

## What happened

Demo VM (D4s_v3, api container `mem_limit` 2 GiB, `UVICORN_WORKERS=2`), 2026-07-23 09:16:10 UTC:

```
Memory cgroup out of memory: Killed process (python, uid 1001, geolens-api-1)
total-vm: 6187228kB, anon-rss: 1838968kB, file-rss: 87168kB
```

- One uvicorn worker had grown to ~1.9 GB RSS (~6 GB VSZ); the master respawned it 4s later ("Application startup complete" 09:16:14).
- Zero user-visible impact this time: nginx logged no 5xx in the surrounding window (in-flight requests on the killed worker would have been dropped, but traffic was light).
- Container had been up ~15h; the only OOM event in the dmesg buffer.
- Activity in the preceding ~30 min: normal catalog browsing plus sustained raster-proxy traffic — `/api/tiles/raster-proxy/...` tiles for the swissALTI3D Matterhorn 2m VRT mosaic and ETOPO at 2.2–2.6 s/tile.
- Steady-state api memory is ~680 MB, so this was ~1.2 GB of growth (leak or unbounded buffering), not a baseline problem.

## Why it matters

With 2 workers under one 2 GiB cgroup, a single leaking/ballooning worker starves the sibling; an OOM kill during a burst drops every in-flight request on that worker with no error handling, and a repeat under launch-level load turns into visible 502s.

## Already shipped

[...]

PR #650, opened 2026-07-23
feat(observability): per-worker RSS gauge + log watermark (#643)

ishiland:
Part of #643 (does not close it — root cause still open; findings posted on the issue).

## What
Each api worker now samples its own RSS from `/proc/self/status` every 60s:
- **Prometheus gauge** `geolens_worker_rss_bytes` (mirrors the existing `pool.py` background-task pattern; task started/cancelled in the lifespan).
- **Hourly INFO heartbeat + startup baseline** in structured logs, so a growth curve exists in `docker logs` even when `/metrics` is never scraped — the demo OOM was invisible outside dmesg.
- **Rate-limited WARNING** (≤1 per 5 min) when a worker crosses **60% of the container cgroup limit** (v2 `memory.max`, v1 fallback; 1200 MB default when unlimited) — with two workers per cgroup, one worker at 60% is already starving its sibling.

No new dependencies; quietly no-ops on platforms without `/proc` (macOS dev).

## Verification
- `pytest tests/test_memory_metrics.py` — 5 passed (gauge set, watermark derivation from cgroup limit, no-limit fallback, warn + rate-limit + re-warn after window, no-/proc no-op). The rate-limit test caught a real bug (initial `_last_warned_at=0.0` suppressed the first warning near clock zero — now `-inf`).
- `tests/test_layering.py` — 29 passed; ruff clean.
- **Live in the real api container**: startup baseline `API worker memory ... rss_mb=257` logged per worker, and `/metrics` serves `geolens_worker_rss_bytes 2.6968064e+08` matching the log.

### DIFF
--- a/backend/app/api/main.py
+++ b/backend/app/api/main.py
@@ -314,9 +314,13 @@ async def lifespan(app: FastAPI):
     await init_tile_pool()
     await task_app.open_async()
 
+    from app.observability.metrics.memory import update_memory_metrics
     from app.observability.metrics.pool import update_pool_metrics
 
     pool_metrics_task = asyncio.create_task(update_pool_metrics())
+    # fix(#643): per-worker RSS gauge + log watermark so an OOM-bound worker
+    # is visible in normal logs before the kernel kills it.
+    memory_metrics_task = asyncio.create_task(update_memory_metrics())
 
     async def _stale_jobs_sweeper() -> None:
         """Periodically fail jobs whose worker crashed mid-run.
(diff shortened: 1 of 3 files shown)


ishiland:
@codex review

ishiland:
@codex review

Addressed the P2 within this PR's scope: the gauge is now **labeled by `pid`**, so scrapes landing on different workers accumulate as distinct per-worker series in the TSDB instead of silently overwriting a single series — verified live: `geolens_worker_rss_bytes{pid="142"} 2.697e+08`. Full multiprocess-collector adoption is a repo-wide, deployment-affecting change (every existing metric — HTTP histograms, pool, jobs — shares the per-registry property; needs `PROMETHEUS_MULTIPROC_DIR` in compose/Helm + boot cleanup of stale pid files), so it's split out as #651 rather than folded in here. The module docstring documents the caveat and points at #651; the structured-log heartbeat/watermark remains the authoritative per-worker signal, which is what the #643 incident actually lacked.

PR #656, opened 2026-07-23
fix(observability): zero stale job-queue gauges when a queue drains (#655)

### LINKED ISSUE #655: fix(observability): stale job-queue gauges — seen_todo/seen_doing reset logic was never implemented
## Symptom

`geolens_jobs_queue_depth` / `geolens_jobs_active` gauge children keep their last non-zero value forever once a queue's `todo`/`doing` rows disappear from the `GROUP BY` (e.g. a bulk cancel/delete drains a queue between refresh cycles). The comment in `_refresh_job_metrics` says *"Reset gauges to zero before setting"* and the code builds `seen_todo`/`seen_doing` sets — but nothing ever uses them (`backend/app/observability/metrics/jobs.py:63-75`). The reset was clearly intended and never written.

## Failure scenario

A queue backs up past 100 (`GeoLensJobQueueBacklog` threshold), the operator bulk-cancels it, the `status='todo'` rows vanish in one cycle → the gauge freezes at the last value → the reference alert keeps firing (or fires later) on a queue that is actually empty, and dashboards show a phantom backlog until the worker restarts.

## Fix sketch

Track the set of queue labels ever set; after processing rows, `set(0)` any tracked queue absent from `seen_todo`/`seen_doing` this cycle (that's what the dead sets were for).

## Related observation (may be intended)

On the first refresh after worker boot, `_prev_counts` is empty so `jobs_failed_total`/`jobs_completed_total` seed themselves with the full historical row counts from `procrastinate_jobs` (observed live: a fresh worker exports `failed_total{ingest}=5` from months-old rows). `increase()`-based alerting is immune (the seed precedes the first scrape), but anyone reading raw counter values will misread history as recent failures. Worth a code comment if it stays.

Found while dogfooding the reference monitoring config (`infra/monitoring/`) against the demo VM.

ishiland:
## Summary

Fixes #655. `_refresh_job_metrics` built `seen_todo`/`seen_doing` sets for a gauge reset that was never implemented — a queue whose `todo`/`doing` rows vanished between refresh cycles (bulk cancel, row deletion) froze at its last non-zero value forever, which can false-fire the `GeoLensJobQueueBacklog` reference alert and show phantom backlog on dashboards until the worker restarts.

The fix tracks queue labels ever set (`_known_queues`) and zeroes any queue absent from the current cycle's rows. Zero, not remove: the series stays exported so `or vector(0)` dashboard fallbacks and `increase()` alerts keep working.

Also adds the code comment requested in the issue's related observation: on the first post-boot cycle `_prev_counts` is empty, so the completed/failed counters seed with full historical row counts — raw values include pre-boot history; `increase()`/`rate()` are unaffected.

No schema/API/config impact.

## Verification

- `cd backend && uv run pytest tests/test_metrics.py -v` — 6 passed, including the new `test_refresh_job_metrics_zeroes_drained_queues` (populated cycle → drained cycle → gauges read 0)
- `uv run pytest tests/test_layering.py -q` — 29 passed
- `uv run ruff check . && uv run ruff format --check .` — clean

### DIFF
--- a/backend/app/observability/metrics/jobs.py
+++ b/backend/app/observability/metrics/jobs.py
@@ -37,8 +37,15 @@ jobs_failed_total = Counter(
     ["queue"],
 )
 
-# Track previous snapshot for counter delta computation
+# Track previous snapshot for counter delta computation.
+# NOTE(#655): on the first cycle after boot this is empty, so the counters seed
+# with the full historical procrastinate_jobs row counts — raw counter values
+# include pre-boot history. increase()/rate() queries are unaffected.
 _prev_counts: dict[tuple[str, str], int] = {}
+
+# Queues whose gauge children have been set at least once — zeroed (not
+# removed) when their todo/doing rows disappear from a cycle. fix(#655)
+_known_queues: set[str] = set()
 
 
 async def _refresh_job_metrics() -> None:
@@ -60,8 +67,8 @@ async def _refresh_job_metrics() -> None:
             )
             rows = result.fetchall()
 
-        # Reset gauges to zero before setting — handles queues that disappear
-        # We track which (queue, status) combos we see this cycle
+        # Queues absent from this cycle get zeroed after the loop via
+        # _known_queues, so a drained queue reads 0 instead of its last value
         seen_todo: set[str] = set()
         seen_doing: set[str] = set()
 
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

ishiland:
The SDKs Drift Gate failure is pre-existing and unrelated: an unpinned ruff 0.16.0 resolve inside the SDK generator env broke the gate repo-wide today. Fix is #657; this branch will be rebased once it merges.

PR #658, opened 2026-07-23
fix(observability): split the p95 latency alert into interactive vs tile rules

ishiland:
## Summary

Dogfooding the reference monitoring config on the demo VM: within 35 minutes of wiring alert delivery, `GeoLensApiHighLatencyP95` fired during plain map browsing. A visitor panning the NYC showcase map at ~2.2 tile-req/s pushed **global** p95 to 1.4s while non-tile p95 sat at 0.35s and host load at 0.13 — the alert paged on the product working as intended, because the label-less `highr` histogram can't exclude the legitimately slow tile tail.

Replaced with two rules in `infra/monitoring/alerts.yml`:

- **`GeoLensApiInteractiveLatencyP95`** — p95 of non-`/tiles/` handlers `>= 1` for 10m. Excluding a handler forces the handler-labeled histogram, whose buckets top out at `le=1.0`, so `histogram_quantile` clips at 1.0 and `>= 1` is the only expressible "worse than 1s" predicate (same bucket-ceiling reasoning as the #466 fix this replaces).
- **`GeoLensApiTileLatencyMean`** — exact mean from `_sum`/`_count` over `/tiles/*` `> 2s` for 10m. Means don't clip, active browsing keeps the mean well under 1s, and 2s+ sustained is genuine grinding (#643's raster-proxy incident ran 2.2–2.6s/tile). No traffic → 0/0 = NaN → never fires.

All tile serving (vector `.pbf` + raster proxy) lives under the single `/tiles` router prefix, so one regex covers it. RUNBOOK §4 alert table updated.

No schema/API impact; config + docs only.

## Verification

- `promtool check rules` — 7 rules OK
- Deployed live to the demo VM (which dogfoods this file verbatim): all 12 rules `health=ok`, the old alert cleared, neither new rule fires under the active tile-browsing session that false-fired the old one
- prom-guard healthchecks integration recovers on the next clean 15-min cycle

### DIFF
--- a/RUNBOOK.md
+++ b/RUNBOOK.md
@@ -312,7 +312,8 @@ targets just need network reach to `api:8000` and `worker:8001`.
 |---|---|---|
 | `GeoLensTargetDown` | API or worker unscrapeable for >2m | critical |
 | `GeoLensApiHigh5xxRate` | 5xx share of requests >5% over 5m | critical |
-| `GeoLensApiHighLatencyP95` | p95 request latency >1s, sustained 10m | warning |
+| `GeoLensApiInteractiveLatencyP95` | non-tile p95 at the 1s histogram ceiling for 10m | warning |
+| `GeoLensApiTileLatencyMean` | mean `/tiles/*` latency >2s for 10m | warning |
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

PR #665, opened 2026-07-24
feat(ogc): make the items page-size ceiling admin-configurable and clamp over-limit requests

### LINKED ISSUE #666: Make the OGC Features items page-size ceiling configurable from the admin dashboard
### Problem or motivation

The maximum `limit` accepted by the OGC API Features items endpoint (`/api/collections/{dataset_id}/items`) is a hardcoded FastAPI query constraint (`Query(10, ge=1, le=...)` in `backend/app/standards/ogc/router.py`). An operator who wants to allow larger bulk GeoJSON pages (or clamp them down for a resource-constrained deployment) has to edit source code and redeploy. I went looking for this setting in the admin dashboard and it does not exist there, which is where an operator would naturally expect a server-tuning knob like this to live.

Context: #665 raises the hardcoded ceiling from 200 to 10,000 now that the route pages via the constant-time `after_gid` keyset cursor, so the cost of a large page is response size rather than query time. The right ceiling is still deployment-dependent (memory, proxy body limits, tunnel bandwidth), which makes it a good candidate for runtime configuration rather than a compile-time constant.

### Proposed solution

Expose the ceiling as a `PersistentConfig[int]` entry (the same mechanism that backs the existing admin settings tabs, e.g. `GLOBAL_RATE_LIMIT` on the network tab or `UPLOAD_MAX_SIZE_MB` on storage), something like:

```python
OGC_ITEMS_MAX_PAGE_SIZE = PersistentConfig[int](
    key="ogc_items_max_page_size",
    type_=int,
    env_default=10_000,
    tab="network",
    label="OGC Features Max Page Size (items limit ceiling)",
)
```

One implementation note: a static `Query(le=...)` constraint is evaluated at route-definition time, so making the ceiling dynamic means moving the upper-bound check into the handler (or a dependency) and raising the same OGC 400 Problem Detail when the configured ceiling is exceeded. `ge=1` can stay declarative. The OpenAPI schema would document the default ceiling.

If the runtime value should also be discoverable by clients, the standard place is the OGC `maxLimit` in the API definition, but that could be a follow-up.

### Alternatives considered

- Editing `router.py` in a checkout and relying on `--reload` (what I do today) — works for a dev deployment, but is invisible to other admins and lost on redeploy from images.
- An environment variable only — better than hardcoding, but the project already has the admin-dashboard `PersistentConfig` pattern for exactly this kind of operator tuning, and env-only settings require a container restart.

### Area

Admin Panel

giswqs:
## Summary

The OGC API Features items endpoint (`/collections/{dataset_id}/items`) capped `limit` at 200 and rejected anything above it with HTTP 400. That was sized for offset pagination; Phase 269 H-24 moved this route to the constant-time `after_gid` keyset cursor, so a large page now costs response size, not query time.

Following review, this PR does **not** hardcode a new higher constant. Instead it makes the ceiling an **admin-configurable knob** (resolving #666) and **clamps** over-ceiling requests instead of rejecting them:

- **`OGC_ITEMS_MAX_PAGE_SIZE`** — a `PersistentConfig[int]` on the Network settings tab (same mechanism as `global_rate_limit` / `upload_max_size_mb`), default **1,000**. Operators raise it for bulk GeoJSON export or lower it on memory-constrained deployments, from the dashboard, no redeploy.
- **Clamp, don't reject** — the static `Query(le=...)` bound moves into the handler (`limit = min(limit, max(1, ceiling))`), satisfying OGC API Features Core `/req/core/fc-limit-response-1(C)` ("a value above the maximum SHALL NOT error"). Mirrors the STAC side (#664). The offset-paged list endpoints keep their 200 cap.
- Conservative default addresses the memory concern (the offset path is O(N) and the response is built fully in memory on an anonymous endpoint; cf. the #643 OOM) while leaving bulk export a dashboard opt-in.

Closes #666

## Changes

- `backend/app/core/persistent_config.py` — `OGC_ITEMS_MAX_PAGE_SIZE` knob (default 1,000, `network` tab).
- `backend/app/standards/ogc/router.py` — drop `le=` from the `limit` query; clamp to the configured ceiling in the handler.
- `backend/app/modules/settings/schemas.py` — value validator (`1..100000`).
- `frontend/.../SettingsNetworkTab.tsx` + all four locale bundles — admin dashboard field.
- `backend/openapi.json`, SDKs, frontend API types — regenerated.

## Verification

- `tests/test_ogc_pagination.py`: at-ceiling accepted, above-ceiling clamped (200, not 400/422), and a non-default configured ceiling honored. Full OGC + settings backend suites pass.
- Frontend i18n locale-parity + settings-tab suites pass; `make openapi-check` and `pre-commit run --all-files` clean.

### DIFF
--- a/backend/app/core/persistent_config.py
+++ b/backend/app/core/persistent_config.py
@@ -70,6 +70,11 @@ _DEFAULT_LOGIN_RATE_LIMIT = 5
 _DEFAULT_GLOBAL_RATE_LIMIT = 60
 _DEFAULT_SEMANTIC_SEARCH_RATE_LIMIT = 30
 _DEFAULT_BASEMAP_PROXY_RATE_LIMIT = 120
+# Ceiling for the OGC API Features items page size (`limit`). Conservative by
+# default (#665 review): the offset path is O(N) and the response is built fully
+# in memory on an anonymous endpoint, so a low default protects resource-limited
+# deployments while operators who need bulk export raise it from the dashboard.
+_DEFAULT_OGC_ITEMS_MAX_PAGE_SIZE = 1000
 
 
 # ---------------------------------------------------------------------------
@@ -723,6 +728,14 @@ CORS_ALLOWED_ORIGINS = PersistentConfig[str](
     env_default_factory=lambda: settings.cors_allowed_origins,
     tab="network",
     label="CORS Allowed Origins",
+)
+
+OGC_ITEMS_MAX_PAGE_SIZE = PersistentConfig[int](
+    key="ogc_items_max_page_size",
+    type_=int,
+    env_default=_DEFAULT_OGC_ITEMS_MAX_PAGE_SIZE,
+    tab="network",
+    label="OGC Features Max Page Size (items limit ceiling)",
 )
 
 # -- Storage tab --
--- a/backend/app/modules/settings/schemas.py
+++ b/backend/app/modules/settings/schemas.py
@@ -331,6 +331,15 @@ def validate_global_rate_limit(v: Any) -> int:
     return v
 
 
+def validate_ogc_items_max_page_size(v: Any) -> int:
+    # Upper bound guards against a mis-set value that would let an anonymous
+    # request build an unbounded FeatureCollection in memory (#665 review).
+    v = int(v)
+    if v < 1 or v > 100_000:
+        raise ValueError("ogc_items_max_page_size must be between 1 and 100000")
+    return v
+
+
 def validate_semantic_search_rate_limit(v: Any) -> int:
     v = int(v)
     if v < 1 or v > 1000:
@@ -549,6 +558,7 @@ SETTING_VALIDATORS: dict[str, Any] = {
     "log_level": validate_log_level,
     "login_rate_limit": validate_login_rate_limit,
     "global_rate_limit": validate_global_rate_limit,
+    "ogc_items_max_page_size": validate_ogc_items_max_page_size,
     "semantic_search_rate_limit": validate_semantic_search_rate_limit,
     "basemap_proxy_rate_limit": validate_basemap_proxy_rate_limit,
     "upload_max_size_mb": validate_upload_max_size,
--- a/backend/app/standards/ogc/router.py
+++ b/backend/app/standards/ogc/router.py
@@ -13,6 +13,7 @@ from app.core.db.tenant_session import current_tenant_var
 from app.core.dependencies import get_db
 from app.core.geo import extent_to_bbox
 from app.core.identity import Identity
+from app.core.persistent_config import OGC_ITEMS_MAX_PAGE_SIZE
 from app.core.public_urls import get_public_api_url, get_public_app_url
 from app.core.tenancy import is_multi_tenant
 from app.modules.auth.dependencies import get_optional_user_or_401
@@ -444,7 +445,19 @@ async def get_dataset_collection(
 async def get_collection_items(
     request: Request,
     dataset_id: uuid.UUID,
-    limit: int = Query(10, ge=1, le=200),
+    limit: int = Query(
+        10,
+        ge=1,
+        description=(
+            "Page size for bulk GeoJSON export. The maximum is an "
+            "admin-configurable ceiling (default 1000; Network settings tab, "
+            "`ogc_items_max_page_size`) rather than the 200 used by offset-paged "
+            "list endpoints, because this route pages via the constant-time "
+            "`after_gid` keyset cursor. Per OGC API Features Core "
+            "/req/core/fc-limit-response-1(C) a value above the ceiling is "
+            "clamped to it, not rejected."
+        ),
+    ),
     offset: int = Query(
         0,
         ge=0,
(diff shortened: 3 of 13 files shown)


giswqs:
If 10,000 is too much, can also consider 1,000. I feel that 200 is too conservative. 

image

ishiland:
Good catch on this and thanks for raising it @giswqs ! Its great to see you here. 

Four things before this can land, please push back if they seem unreasonable:

1. `Closes #666` doesn't match what #666 asks for. The issue proposes an admin-configurable ceiling via `PersistentConfig`; this PR raises a hardcoded constant. Merging as-is would auto-close the feature request without implementing it. Either switch to `Refs #666` or, my preference, implement the knob here, since it also resolves point 3. Your sketch in #666 is basically right, including the note that the bound check moves out of the static `Query(le=...)` into the handler.

2. Clamp instead of rejecting. This turns out to be a hard spec requirement, not a style preference. OGC API Features Core, Requirement `/req/core/fc-limit-response-1(C)`: "If the value of the `limit` parameter is larger than the maximum value, this SHALL NOT result in an error (instead use the maximum as the parameter value)." The current 400-on-over-limit behavior on main is a pre-existing conformance bug, and raising the ceiling keeps it. Since point 1 moves the bound check into the handler anyway, clamping is the same line of code. It also matches what #664 just did on the STAC side for the same reason.

3. The performance justification only covers the keyset path, and memory is the binding constraint. The legacy `offset` path was retained and is O(N); the old 200 cap partly protected against `offset=90000&limit=10000` style requests, which this raise makes 50x heavier. On response size: ~5 MB per 10k features is footprint-shaped. Heavy-polygon datasets (parcels, admin boundaries) run an order of magnitude more, the response is built fully in memory, and the endpoint is anonymous. We had an API worker OOM at its container cap earlier this week (#643), so a hardcoded 10,000 on day one makes me nervous. The configurable ceiling from #666 with a conservative default (1,000 to 2,000) resolves this: bulk-export deployments opt in from the dashboard, and resource-constrained ones aren't exposed by default.

4. Please add a test pinning the bound. Nothing in the suite asserts this route's ceiling; one test in `tests/test_ogc_pagination.py` covering "at the ceiling accepted, above the ceiling clamped" keeps the next refactor honest (plus a non-default value if it becomes configurable).

Thanks again for digging into the pagination internals before proposing this.

giswqs:
Thanks for the thorough review @ishiland, this all made sense. Pushed a revision that pivots the PR from "raise the constant" to "make the ceiling the configurable knob from #666". Point by point:

**1. Implement the knob (and keep `Closes #666`).** Done. Added `OGC_ITEMS_MAX_PAGE_SIZE = PersistentConfig[int]` on the `network` tab (same mechanism as `global_rate_limit` / `upload_max_size_mb`), wired into the admin Network settings tab with i18n in all four locales and a `1..100000` value validator. Since the dashboard knob is now actually implemented, `Closes #666` is accurate rather than a premature auto-close.

**2. Clamp instead of reject.** Done, and you're right that it fell out of point 1 for free. The static `Query(le=...)` bound is gone; the handler now does `limit = min(limit, max(1, await OGC_ITEMS_MAX_PAGE_SIZE.get(db)))`, so an over-ceiling `limit` is clamped to the maximum rather than 400'd, per OGC API Features Core `/req/core/fc-limit-response-1(C)`. Matches the `min(body.limit, ...)` shape #664 used on the STAC side.

**3. Memory is the binding constraint.** Agreed, and this is why the default is now **1,000**, not 10,000. The ceiling is per-deployment tunable from the dashboard: bulk-export deployments raise it, resource-constrained ones keep the conservative default and aren't exposed to `offset=90000&limit=10000`-style requests by default. The performance justification in the original description (keyset path) only covered query time; the configurable conservative default is what actually addresses the in-memory response-size risk on this anonymous endpoint.

**4. Test pinning the bound.** Added three to `tests/test_ogc_pagination.py`:
- `test_items_limit_at_default_ceiling_accepted` — at the ceiling, accepted.
- `test_items_limit_above_ceiling_clamped_not_rejected` — above the ceiling, clamped (200, not 400/422), and the echoed self link carries the clamped value.
- `test_items_limit_clamped_to_configured_ceiling` — a non-default configured ceiling (3) is honored: a `limit=100` request returns a 3-feature page with a `next` link.

`backend/openapi.json`, the SDKs, and the frontend API types are regenerated (the `limit` schema no longer advertises a static `maximum`; the description documents the default ceiling and the clamp behavior). `make openapi-check`, `pre-commit run --all-files`, the OGC + settings backend suites, and the frontend i18n/settings suites all pass locally.

ishiland:
This covers all four points, thanks for taking the rewrite rather than just bumping the constant. The knob plus clamp is exactly the shape #666 was asking for, and 1,000 is a sensible default to ship.

Two notes, neither blocking.

**Scope of the clamp.** It covers `/collections/{dataset_id}/items`. `/collections/datasets/items` is served by a different router (`modules/catalog/search/router.py`) that takes its bound from the shared `SearchQueryParams.limit = Query(10, ge=1, le=200)` in `catalog/search/query_params.py:85`, so it still returns 400 on an over-cap `limit`. I checked against the demo after your revision to be sure.

I am explicitly not asking you to fix that here. That dep is shared with `/api/search/datasets/`, which is not an OGC route and where the current reject is defensible, so changing it needs a deliberate call about the second route rather than a repeat of what you did. I have re-scoped #687 to cover just that, with the tradeoff written up.

**Merge order.** #689 edits the same four `admin.json` locale bundles in the CORS block, one line above the `corsAllowedOriginsPlaceholder` line you touched, so whichever of the two lands second gets a conflict there. `git merge-tree` confirms it is only those four files. Yours goes first and I will rebase mine on top. Nothing for you to do.

Issue #549, opened 2026-07-16
Builder chat: 'show me X' phrasing routes unpredictably between map edits and data queries


ishiland:
## Observation

In builder chat, "Show the ADA accessible stations served by the A line" applied a **layer filter** (map-edit path), while "Run a query selecting ... ADA accessible stations on the A line" produced a **query result** (table + highlight overlay). Both are reasonable, but a user can't predict which they'll get from natural phrasing, and the two outcomes look very different (persistent styled filter vs dismissible ephemeral overlay).

## The premise this issue started from was wrong

The original write-up proposed adding question-vs-change routing prose to the tool descriptions, on the theory that no such guidance existed. It does exist, in the system prompt. `backend/app/processing/ai/chat_service.py:268-273` already reads:

```
- When the user asks a QUESTION about their data (counts, statistics, spatial
  relationships, distances, finding features), use the query_data tool.
- When the user asks to CHANGE the map (colors, filters, labels, visibility,
  add/remove layers), use the map editing tools.
```

The misroute happens anyway. So the fix is not "add the guidance"; it is to work out why the guidance already there does not catch this phrasing, and to settle which site owns the answer.

## What actually goes wrong

"Show" is not classified anywhere. The prompt's two buckets are QUESTION verbs (counts, statistics, spatial relationships, distances, finding features) and CHANGE verbs (colors, filters, labels, visibility, add/remove layers). Neither list contains "show", and the request "show me the ADA accessible stations" reads as plausible under both.

Worse, the one place "show" *is* classified points the other way. `run_analysis`'s description (`backend/app/processing/ai/tools.py:333-345`) uses it as an example of a geometry-transform request:

> Use this for requests that TRANSFORM geometry -- 'buffer the schools by 500 metres', **'show the centre point of each parcel'**. Use `query_data` instead for questions ANSWERED from the data (counts, sums, which/where/how many).

Meanwhile `set_filter` (`tools.py:95-119`) documents expression syntax only and carries no routing guidance at all.

So the model is reading verb-class guidance from two competing sites, one of which teaches "show" as a transform verb, while `set_filter` sits unguarded.

## The fix: pick one owner for verb classification

The change is a prompt/description edit, but it needs a rule to edit against, not more prose bolted on:

[...]

PR #671, opened 2026-07-24
feat(analysis): buffer/centroid/clip/dissolve tools with map preview and dataset materialization

ishiland:
## Summary

Adds the first slice of PostGIS analysis tools (todo M4): a **Analysis** rail panel in the map builder where you pick a vector layer, run **buffer / centroid / clip** as a live map preview, and materialize any op (plus **dissolve**) into a new catalog dataset via the worker.

All SQL is server-built from validated parameters — no raw user or LLM SQL touches this path.

## Backend

- `POST /api/datasets/{id}/analysis/preview/` — renders the op through the existing read-only sandbox (`execute_safe`: READ ONLY txn, `geolens_reader` role, 10s timeout, per-user advisory lock shared with AI chat). Returns a GeoJSON FeatureCollection capped at 500 features with a `truncated` flag and bbox.
- `POST /api/datasets/{id}/analysis/materialize/` — quota-checked, enqueues a Procrastinate job (orphan-guarded); the worker CTASes the result table, normalizes SRID/PK/`geom_4326`, and registers it through `register_existing_table` so the output is a full private dataset (tiles, metadata, grants).
- Shared SQL rendering lives in `app/platform/analysis_sql.py` to satisfy the catalog↔processing layering guards; the worker is reached via the catalog port.
- Injection posture: numeric params are bounded Pydantic floats; clip masks are parsed and re-serialized by shapely (Polygon/MultiPolygon only, ≤5000 vertices, `make_valid`) before ever entering SQL; identifiers go through `_safe_table_ref` / regex-validated column names checked against `column_info`.
- Access control: both endpoints use `check_dataset_access` (private → 404).

## Frontend

- New Analysis rail panel (`AnalysisPanel.tsx`): layer picker, op select, buffer distance, clip-mask drawing via terra-draw directly on the builder map, preview as the existing ephemeral dashed overlay with fit-to-bounds, truncation toast, materialize with inline job polling and one-click "Add to map".
- i18n: full `analysisTools` key group in all four locales.
- **Bug fix that reaches beyond this feature** (`fix(builder)` commit): BuilderMap's cleanup effect depended on `[onMapRef]`, so parent-callback identity churn nulled the map ref mid-session and detached the error/idle handlers with no re-attach. This silently broke AI-chat ephemeral overlays too. Cleanup is now genuinely unmount-only.

## Schema / API / config impacts

- No DB migrations. Output tables use the standard per-dataset physical-table shape.
- OpenAPI + both SDKs regenerated (`backend/openapi.json`, `sdks/python`, `sdks/typescript`) — committed because the source change requires it.
- No new config, env vars, or dependencies.

## Verification

- `make test` — full backend suite green: 5770 passed, 76 skipped (new: 16 preview, 7 materialize, layering updated). New materialize tests join the `tenancy_global_state` xdist group (cluster-global `geolens_reader` GRANT contention).
- `make openapi-check && make sdks-check && make cli-test`
- `cd frontend && npm run typecheck && npm run lint && npm run test:i18n && npm run test:coverage && npm run build` (3738 vitest tests)
- Live smoke on the dev stack: buffer preview 412 features in 0.7s, truncation at 500, clip via drawn mask, worker materialize <3s with attributes carried and SRID 4326, dissolve, add-to-map renders real vector tiles.
- Browser UI audit: draw→preview→materialize→add-to-map flow, dark mode, <800px mobile sheet, Escape-to-close, Spanish locale, clean console on fresh load. Screenshots available — will attach below.

## Follow-ups (out of scope here)

- AI-chat `run_analysis` tool reusing the same service/schema — separate PR; should also help the routing ambiguity in #549.
- Persistent visual of the drawn clip mask after capture; optional `ST_SimplifyPreserveTopology` on large previews; dissolve `by_field` exposed in the panel (API/SDK support it already).


ishiland:
UI audit screenshots — in order: buffer preview (50 km on world lakes, 412-feature ephemeral overlay), clip preview from a drawn mask, materialized output added back to the map as vector tiles, dark mode, mobile (<800px) sheet.

qa-01-buffer-preview
qa-02-clip-preview
qa-04-added-layer
qa-05-dark-panel
qa-06-mobile-sheet

PR #673, opened 2026-07-24
perf(ci): drop the duplicate parallel-isolation job, benchmark -n 6

ishiland:
## Summary

`Pytest Parallel Isolation` has been a full duplicate of `Backend Tests` since #561 made the latter run `-n 4`: both ran the same ~5,850-test suite with identical xdist parallelization (`-n 4 --dist loadgroup`), differing only in coverage flags and the `lifecycle` marker — which selects **4 tests**. Any fixture-isolation regression the job guarded against fails Backend Tests identically. Measured on a representative run (30116234385): Backend Tests 11m19s + PPI 10m35s, both ~9.6 min of pytest.

## Changes

- **Delete the `pytest-parallel-isolation` job** (−112 lines) and remove it from the three `needs:` lists (none referenced its `.result` explicitly). Saves ~10.5 runner-minutes on every backend-touching PR push — roughly half the backend path's Actions spend.
- **Fold the 4 `lifecycle` tests into Backend Tests** (`-m 'not perf and not lifecycle'` → `-m 'not perf'`), so nothing PPI ran is lost.
- **Absorb PPI's trigger**: Backend Tests' `if:` now also fires on `changes.outputs.alembic` — previously an alembic-only PR ran the suite *only* via PPI, so deleting it without this would have dropped pytest coverage there.
- **Benchmark `-n 6`** (was `-n 4`): the suite is Postgres-wait-bound, so oversubscribing the 4-vCPU runner should fill DB-wait gaps. Baseline pytest step: 585s at `-n 4`. This PR's own Backend Tests run is the measurement; if it doesn't clearly win I'll revert to `-n 4` before merge.
- **Add `--durations=25`** so every run reports its slowest-test tail — data for any future sharding decision.

## ⚠️ Required before merge (repo settings — maintainer-only)

`Pytest Parallel Isolation` is a **required status check** on `main`. This PR's workflow no longer reports it, so the branch protection entry must be removed first or the PR (and every PR after it) can never merge:

Settings → Branches → `main` rule → Require status checks → uncheck "Pytest Parallel Isolation" — or:

```
gh api -X DELETE repos/geolens-io/geolens/branches/main/protection/required_status_checks/contexts \
  -f "contexts[]=Pytest Parallel Isolation"
```

## Verification

- `yaml.safe_load` parses; pre-commit yaml check passes.
- No remaining references to the job id or name (grep), no `needs.pytest-parallel-isolation.result` usages existed.
- This PR's own CI run doubles as the `-n 6` benchmark; result will be posted as a comment.

Note: `JS audit (frontend)` will show red here until #672 (npm-audit allowlist for the new react-router advisory) merges — unrelated to this change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -751,7 +751,9 @@ jobs:
   backend-test:
     name: Backend Tests
     needs: changes
-    if: needs.changes.outputs.backend == 'true' || github.event_name == 'push'
+    # Alembic-only changes must run the suite too — this condition absorbed the
+    # deleted pytest-parallel-isolation job's trigger when that job folded in.
+    if: needs.changes.outputs.backend == 'true' || needs.changes.outputs.alembic == 'true' || github.event_name == 'push'
     runs-on: ubuntu-latest
     defaults:
       run:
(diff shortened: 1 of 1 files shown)


ishiland:
**Benchmark result: `-n 6` loses — reverted to `-n 4`** (commit pending push, held for a rebase once #671 lands so this PR takes one CI cycle, not two).

| config | pytest step | suite |
|---|---|---|
| `-n 4` (baseline, run 30116234385) | **585s** | 5,770 passed |
| `-n 6` (this PR, run 30117552877) | **605s** | 5,733 passed, 81 skipped |

The suite keeps the 4 vCPUs saturated, so oversubscribing past core count only adds scheduler overhead. Two useful side-findings from the run:

- The suite is green at `-n 6`, so the parallel-isolation story holds even oversubscribed — more evidence the deleted job guarded nothing Backend Tests doesn't.
- Skip count went 76 → 81: the folded-in `lifecycle` tests **skip** without the enterprise overlay, i.e. the deleted job's only unique selection wasn't even executing in OSS CI.

`--durations=25` tail (now visible every run): flat profile, slowest is 28s of fixture setup, then a band of 8–13s migration-replay tests (`test_embedding_vector_migration` ×7, `test_oauth_github_migration`, backup-schedule validation) and sleep-based rate-limit tests. No single hot spot to fix — but a flat tail means a 2-way shard would split near-perfectly if we ever want wall-clock halved.

PR #674, opened 2026-07-24
feat(ai): add a run_analysis chat tool for buffer/centroid previews

ishiland:
Closes out the M4 spatial-analysis work (#671) with Phase 5: the AI chat tool.

## What

The map-chat assistant can now run the same parameterized PostGIS operations the Analysis rail panel offers. "Buffer the schools by 500 m" draws a preview on the map instead of being routed to `query_data`, which can answer questions about geometry but cannot transform it. This also helps #549 routing by giving analysis its own tool rather than overloading `query_data`.

## How

`run_analysis` reuses the catalog analysis service through `ProcessingPort` — the same server-built SQL, the same `AnalysisPreviewRequest` validation, and the same read-only sandbox rails as `POST /datasets/{id}/analysis/preview/`. No LLM-authored SQL touches this path: the model picks an operation enum and a bounded number, nothing else.

**Access control.** The layer list is client-supplied, so the handler re-resolves the dataset from `layer.dataset_id` and re-checks it with `check_dataset_access` (AGENTS.md Rule 1) rather than trusting the client's `dataset_id` or table name. There is an IDOR regression test for this.

**Tool scoping.**
- Offered to view-only callers: it only SELECTs, and its result is an ephemeral overlay, never a persisted map change.
- Withheld from dataset chat via `select_chat_tools(has_map=False)` — that surface has no map, so the model must not promise a preview it cannot render.
- Enum is deliberately narrower than the endpoint's: `clip` needs a drawn mask only the rail panel can supply, and `dissolve` is materialize-only.

**No frontend change needed.** Results ride the existing `show_query_result` action carrying `geojson`+`bbox` only. Every chat surface already draws that pair and skips the inline data table when `rows` is absent (pinned by the existing `ChatPanel.test.tsx` case), so a gid-only table never appears.

## Drive-by fix: stop echoing map payload into the model's context

Tool results are re-fed to the provider verbatim at four `json.dumps` sites. A 412-feature buffer preview serialized to ~1.3 MB there — **measured at 661,925 input tokens** on the follow-up round, against a 200k `MAX_REQUEST_TOKEN_BUDGET`, for coordinates the model cannot read anyway.

`model_safe_tool_result()` drops `geojson` at those sites only; the action collector still receives the untrimmed dict. The same live request now costs **354 input tokens**. `query_data`'s 50-feature overlay was quietly paying a smaller version of the same tax and benefits too.

## Layout

The handler lives in a new `chat_analysis.py` sibling (following `chat_styles` / `chat_geojson`) rather than growing `chat_actions.py`. The residual +11 lines of wiring take that module's reviewed cap from 530 to 550.

## Verification

- Full backend suite: 5786 passed. (Two unrelated modules errored in one parallel run and a third in another — different tests each time, all pass standalone and under `-n 4`; the known xdist session-acquisition flake, untouched by this diff.)
- New `test_ai_chat_run_analysis.py`: 18 tests — tool scoping, action shape, DB-backed buffer/centroid, unknown layer, foreign-private IDOR, hostile/hallucinated params, and the payload trim.
- Drift gates clean: `make openapi-check`, `make sdks-check`. **No API surface change, so no OpenAPI/SDK regeneration.**
- Frontend `npm run typecheck`, ChatPanel / ViewerChatPanel / DatasetChatPanel suites.
- Live end-to-end against the dev stack: model routed to `run_analysis`, `Running analysis...` progress event, 412-feature overlay with a valid bbox, correct narration.

## Follow-ups (unchanged from #671)

Persistent visual for the captured clip mask; `ST_SimplifyPreserveTopology` on large previews; dissolve `by_field` UI.

### DIFF
--- a/backend/app/core/processing_port.py
+++ b/backend/app/core/processing_port.py
@@ -230,6 +230,17 @@ class ProcessingPort(Protocol):
         self, session: AsyncSession, user: Identity
     ) -> set[str]: ...
 
+    async def run_analysis_preview(
+        self,
+        session: AsyncSession,
+        dataset: Any,
+        operation: str,
+        *,
+        user_id: uuid.UUID,
+        distance_meters: float | None = None,
+        mask: dict[str, Any] | None = None,
+    ) -> Any: ...  # -> AnalysisPreviewResponse
+
     async def get_column_stats(
         self,
         session: AsyncSession,
--- a/backend/app/platform/ai_tool_payloads.py
+++ b/backend/app/platform/ai_tool_payloads.py
@@ -1,0 +1,36 @@
+"""Trim AI tool results before they are echoed back into the model's context.
+
+A chat tool result serves two consumers with very different needs:
+
+- the **action collector**, which forwards map payload (GeoJSON overlays) to
+  the browser, and
+- the **model**, which is re-fed the tool result as conversation context so it
+  can narrate the outcome.
+
+Geometry is exclusively the first consumer's business — the model narrates from
+``feature_count`` / ``row_count`` / ``rows``, never from raw coordinates. Left
+in, a single ``run_analysis`` buffer preview of a 412-feature layer serialized
+to ~1.3 MB (~325k tokens), blowing MAX_REQUEST_TOKEN_BUDGET in one round and
+billing the user for coordinates the model cannot use. ``query_data``'s overlay
+(capped at 50 features) was quietly paying a smaller version of the same tax.
+
+Call this at every point where a tool result is serialized *for the provider*;
+the action collector must keep receiving the untrimmed dict.
+"""
+
+from __future__ import annotations
+
+# Keys whose whole purpose is client-side map rendering. `bbox` deliberately
+# stays: it is four numbers and gives the model useful spatial context.
+_MAP_ONLY_KEYS = frozenset({"geojson"})
+
+
+def model_safe_tool_result(result: dict) -> dict:
+    """Return ``result`` without map-only payload, for provider serialization.
+
+    Returns the original object when there is nothing to strip, so the common
+    path allocates nothing.
+    """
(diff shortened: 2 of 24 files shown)


ishiland:
Thanks — the P2 is a real bug and is now fixed in 51d53799e.

You're right that the tool schema invites it: it literally says `distance_meters` is "ignored otherwise", so a model sending a placeholder `0` on a `centroid` call is expected behavior, not a malformed call. But the value was forwarded verbatim into `AnalysisPreviewRequest`, whose `gt=0`/`le=100000` bounds validate the optional field regardless of operation — so a valid centroid preview came back as a tool error instead of drawing. Confirmed directly:

```
AnalysisPreviewRequest(operation='centroid', distance_meters=0)
  -> Input should be greater than 0 [type=greater_than, input_value=0]
```

Fix takes your first suggestion — only forward the distance for `buffer`. Added a regression test over `0 / -5 / 1000000 / 500`; it fails on 3 of the 4 without the fix.

The review also prompted a self-review pass that turned up two more issues, fixed in the same commit:

- `check_dataset_access` was wrapped in a bare `except Exception`, so a transient DB error would have been reported to the user as "You don't have access to that layer's dataset." Narrowed to `HTTPException` (the 404 that helper actually raises on denial); anything else now falls through to the generic handler.
- The caller's already-resolved `user_roles` is now passed into `check_dataset_access` instead of letting it re-query per call — the parameter existed and was simply unused.

Also tightened the UUID parse guard to cover only the parse rather than the surrounding DB call, and dropped a defensive `getattr()` since `DatasetProtocol` declares both fields. Tests now pass real user roles instead of an empty set, so the admin cases exercise the real authorization path rather than passing merely because the fixture datasets are public — mutation-checked by deleting the access call and confirming the IDOR test fails.

ishiland:
@codex review

Issue #681, opened 2026-07-25
feat(analysis): phase 2 umbrella — spatial join, measurement, select-by-location, overlay


ishiland:
> **Decomposed 2026-07-30.** The four operations below now have their own issues, each carrying its own API-shape decision, file pointers, and acceptance criteria. This issue is now the umbrella: priority order, shared constraints, and sequencing. Do not implement from this body; pick up a child.
>
> **Updated 2026-07-31.** The landing order and the shared-site collision that forces it now live in each child body under "Serialization", so a reader who picks up one child gets both without opening the others.

## Context

M4 shipped four analysis operations (buffer, centroid, clip, dissolve — #671/#674/#679). All four are **single-layer** transforms: GeoLens currently can't answer any question that relates two datasets to each other, and that's the shape of most real analysis work.

The canonical workflow across planning, utilities, journalism, and site selection is a chain: **buffer → select what falls inside → export the list/count**. We ship step one; the ring on the map can't yet answer the question that motivated drawing it.

A market pass over Felt, Atlas.co, CARTO, ArcGIS Online, QGIS, and Turf.js ranks the table-stakes gaps in this order:

## The four operations, in priority order

1. **Spatial join / point-in-polygon count** — #953. The most universal op in the survey (ships in every comparable product). Two variants: count/aggregate into the polygon layer, and attribute-transfer onto the point layer.
2. **Measurement (area / length)** — #954. Frequently the deliverable itself (acreage for permits, main length by zone), not an intermediate. Cheap: `ST_Area(geography)` / `ST_Length(geography)` as computed columns.
3. **Select-by-location + export** — #955. Completes the buffer chain: select features of layer B intersecting a geometry (a drawn mask or the features of layer A), then export the selected records. The record list is the deliverable (e.g. abutter notification lists).
4. **Intersect / overlay** — #956. Distinct from join: produces new geometry whose area you then measure (impact acres, overlap analysis).

Deliberately out of scope for the map-builder wedge: isochrones (needs a routing engine), clustering/hotspots, voronoi/hull (near-zero usage signal).

## Shared constraints

These bind all four children and are restated on each so none has to be read alongside the others.

[...]

PR #702, opened 2026-07-25
perf(analysis): per-row subdivided mask joins for clip previews

### LINKED ISSUE #693: perf(analysis): clip-by-layer preview re-unions the entire mask layer on every request — size guard + preview semi-join
From the 2026-07-25 analysis scalability audit. Clip-by-layer preview is effectively unusable against realistic mask layers today, independent of source-dataset size.

**Findings**
- `render_mask_cte` builds `WITH _mask AS MATERIALIZED (SELECT ST_Union(ST_CollectionExtract(ST_MakeValid(geom_4326),3)) …)` over the whole mask layer (`backend/app/platform/analysis_sql.py:56-60`). MATERIALIZED guarantees single evaluation (correct) but also full evaluation per request, uncached — every preview click repeats the union inside the 10s sandbox budget. A few thousand mask polygons is enough to time out; the user just sees "Query timed out" with nothing naming the mask.
- `_load_mask_dataset` checks polygonality only — no size guard anywhere (`backend/app/modules/catalog/datasets/api/router_analysis.py:75-87`).
- The layer-mask reaches the outer query as an InitPlan Param (not a Const), so the PostGIS selectivity estimator is blind regardless of table statistics; the drawn-mask variant inlines a literal and gets real stats.
- The correctness filter sits above the geometry expression, so `ST_Intersection` runs for every bbox-surviving row before `ORDER BY gid` — the LIMIT cannot bound the work.

**Recommended fix, two steps**
1. (S, ship first) Mask-size guard in `_load_mask_dataset` on `mask_dataset.feature_count` with an error that names the mask layer.
2. (M) Preview-only strategy change: per-row semi-join / per-row intersection against the mask's own GIST index (bounded by the 500-row candidate set) instead of one giant unioned geometry. Keep the single union for materialize, where it amortizes over millions of rows. An `ST_Subdivide`d mask (≈256-vertex chunks, GIST-indexed) wins for both when masks are reused.

**Coordinate with #681** — phase-2 spatial join faces the identical union-vs-semi-join choice; whatever primitive lands here should be shared.

ishiland:
## Summary

Closes #693 (step 2; the step-1 mask-size guard shipped in #701).

Clip-by-layer **previews** re-unioned the entire mask layer on every request (`render_mask_cte`) and then ran `ST_Intersection` against that single giant geometry per candidate row — all inside the 10s sandbox budget. Realistic masks made the preview time out, independent of source-dataset size.

The preview now uses a three-part shape rendered by the new `render_clip_layer_preview` in `app/platform/analysis_sql.py` (placed there so #681's spatial-join work can share it):

- a `_mask_pieces` MATERIALIZED CTE: the mask's polygonal parts, `ST_Subdivide`d into ≤256-vertex chunks once per statement;
- a per-row LATERAL aggregate that unions the intersections of only the pieces whose boxes touch the row (intersection distributes over union, so the result equals clipping against the unioned mask — one output row per gid however many mask rows overlap it);
- an EXISTS row filter probing the **raw** mask table, keeping the filter index-drivable with real join statistics in either direction (the old CTE reached the outer query as an InitPlan Param, which blinds the selectivity estimator).

**The materialize path is untouched** — the whole-layer union CTE is right for a batch job, where one union amortizes over every row.

## Benchmarks (100k-row source, first 500 preview rows, dev DB)

| mask layer | old (union CTE) | raw semi-join | subdivided CTE only | **shipped hybrid** |
|---|---|---|---|---|
| 1,000 × 257-vertex polygons | **>120s (timeout)** | 0.35s | 0.90s | **0.55s** |
| 3 small polygons | 0.03s | 0.05s | 0.06s | **0.05s** |
| single 100k-vertex polygon | 88s | 125s | 2.0s | **0.36s** |
| 100 polygons at the high-gid end | (≥ dense) | 0.23s | 2.4s | **0.25s** |

The two intermediate shapes each fail a realistic scenario, which is why the hybrid ships: pieces bound the per-pair intersection cost; the raw-table EXISTS keeps the row filter off the un-indexed CTE.

## Correctness

Old vs new result sets compared row-by-row (gid sets + `ST_Equals`/Hausdorff ≤1e-5 on geometries) on four mask layers — scattered dense, sparse, single-huge (182 rows straddling subdivision seams), and a degenerate layer with a LINESTRING row and a NULL-geometry row: identical everywhere. New live test covers overlapping mask rows (one merged feature per gid, no duplicate rows or slivers); new unit test pins the SQL shape.

## Verification

```
uv run pytest tests/test_analysis_preview.py tests/test_analysis_materialize.py tests/test_layering.py
```

No API, schema, or config changes.

### DIFF
--- a/backend/app/modules/catalog/datasets/domain/service_analysis.py
+++ b/backend/app/modules/catalog/datasets/domain/service_analysis.py
@@ -29,7 +29,7 @@ from app.modules.catalog.datasets.domain.schemas import (
     AnalysisPreviewRequest,
     AnalysisPreviewResponse,
 )
-from app.platform.analysis_sql import render_geometry_expr, render_mask_cte
+from app.platform.analysis_sql import render_clip_layer_preview, render_geometry_expr
 from app.platform.sandbox.executor import execute_safe
 
 PREVIEW_FEATURE_CAP = 500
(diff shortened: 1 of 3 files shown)


ishiland:
@codex review

Issue #698, opened 2026-07-25
ux(analysis): job-lifecycle polish — cancel pending jobs, queued step, retry copy, mask picker filter, i18n strays, multi-tab toasts


ishiland:
From the 2026-07-25 analysis audit; sibling of #686 (which covers post-preview continuity and buffer ergonomics — no overlap).

1. **Cancel for pending jobs**: no cancel exists anywhere; `status='cancelled'` is legal per the CHECK constraint (`backend/app/platform/jobs/models.py:26-29`) but never written, retry rejects non-failed rows, and the ops cleanup honors the same 1-hour cutoff — so a job stuck behind a down/backlogged worker holds the user's only analysis slot for a full hour with no way out. Add `POST /jobs/{id}/cancel` restricted to `status='pending'`, fenced (`UPDATE … WHERE status='pending'`, rowcount=0 ⇒ "already started"). Do NOT extend to running jobs until the CancelledError handling from #692 lands — cancelling a running job today would strand the row.
2. **`current_step="queued"`** stamped at job creation (`router_analysis.py`) so a pending job reads as "queued", not indistinguishable from broken. The column is deliberately free-form `String(32)`.
3. **Retry copy is wrong for analysis jobs**: they carry `file_path=""`, so `_retry_capability` answers "The source is no longer available. Start the import again." (`backend/app/platform/jobs/router.py:758-759`) for a job that is not an import. Add a branch keyed on the `analysis` metadata marker.
4. **Mask picker offers non-polygon layers** and defers to the server 422 (`frontend/src/components/builder/AnalysisPanel.tsx:116`) even though `dataset_geometry_type` is already client-side — filter to POLYGON/MULTIPOLYGON.
5. **Two untranslatable strings**: `throw new Error('No layer selected')` reaches users verbatim via `error.message || t(...)` (AnalysisPanel.tsx:189/253 → 244/277), and the panel status line concatenates the raw server error instead of the `analysisTools.jobFailedDetail` template the watcher already uses (AnalysisPanel.tsx:585).
6. **Multi-tab duplicate completion toasts**: the persisted job store has no cross-tab listener and the watcher mounts per tab — both tabs toast and invalidate. The auth store's cross-tab pattern is the reference.
7. **Title input lacks `maxLength`** against the server's 500 (422 only after clicking Create).

Issue #696, opened 2026-07-25
ops(analysis): resource-governance knobs — scoped work_mem, configurable timeouts, global preview bound, per-tenant cap


ishiland:
From the 2026-07-25 analysis scalability audit. Related governance gaps around the analysis workload.

> **Split 2026-07-30.** Item 1 (`temp_file_limit`) moved to **#959** and should ship on its own. It was the only item here that prevents an outage rather than tuning a soft limit, and it is one line plus a RUNBOOK note. It does not need any of the design decisions below.

1. ~~**`temp_file_limit` is unset**~~ → **#959**. A grouped dissolve over millions of rows spills the entire geometry column to `pgsql_tmp` on the same volume as `pgdata`, and disk exhaustion stops Postgres hard. Setting a limit turns that outage into a query error.
2. **Scoped `work_mem` for materialize**: the materialize CTAS is a single statement on a single session, so a `SET LOCAL work_mem` bump inside its transaction is safe in a way a global bump is not. Needs a ceiling so concurrent jobs can't stack (the current cap is per-user, not global). Nothing sets `work_mem` today — the materialize transaction sets only `statement_timeout` (`backend/app/processing/analysis/tasks.py:697` pre-CTAS and `:761` for registration), so this is a new `SET LOCAL` alongside those.
3. **Timeouts are module constants**: `MATERIALIZE_TIMEOUT = "300s"` (`backend/app/processing/analysis/tasks.py:58`) and the preview's `DEFAULT_TIMEOUT_MS = 10_000` (`backend/app/platform/sandbox/executor.py:26`). 300s covers roughly 150k–600k buffered polygon rows; an ordinary 1M-parcel buffer fails. Promote to Settings (per-operation budgets are the follow-up: centroid needs seconds, buffer/dissolve far more). Raising ceilings pairs with #691's lease and the pre-flight gates from #701 — a longer budget without those just holds the slot longer.

   **Correction (2026-07-30):** an earlier version of this item claimed both constants "already carry ponytail comments flagging the hardcoding." They do not, and `grep -rn ponytail backend/app/` returns zero matches repo-wide, so anyone grepping for that marker to find the work will find nothing. `MATERIALIZE_TIMEOUT` carries an ordinary comment at `tasks.py:56-57` ("Hardcoded ceiling; promote to persistent-config if operators hit it"); `DEFAULT_TIMEOUT_MS` carries none.

[...]

Issue #695, opened 2026-07-25
perf(jobs): dedicated analysis queue when a second worker exists


ishiland:
From the 2026-07-25 analysis scalability audit (verified against Procrastinate 3.9 sources).

> **This issue is a decision before it is a change.** A dedicated `analysis` queue only pays once a second worker process exists to listen on it. On the single-worker default it changes nothing that #703 has not already achieved. So the first question is whether a second worker service is wanted at all; if the answer is no, close this rather than shipping a queue nothing listens to differently. The code below is small and well understood once that call is made.

**Finding.** `procrastinate_fetch_job_v2` selects `ORDER BY jobs.priority DESC, jobs.id ASC` — the queue *name* is a filter, not a rank, and no defer site in the backend set a per-job priority. With the default single worker (`WORKER_CONCURRENCY=1`) listening on `"priority,ingest,raster"`, dispatch was strictly FIFO by job id across all three queues, so the `"priority"` queue used for small vector uploads and reuploads conferred no ordering advantage, and one 300-second analysis CTAS head-of-line blocked every upload, reupload, and embedding job for every tenant. The user-visible symptom: a 2 MB GeoJSON upload pending for many minutes with nothing identifying analysis as the cause.

**Shipped:** recommendation 1 landed in #703 (`7ca623b08`). `ANALYSIS_JOB_PRIORITY = -10` (`backend/app/modules/catalog/datasets/api/router_analysis.py:68`) is applied via `.configure(priority=ANALYSIS_JOB_PRIORITY)` at the defer site (`:367`), so queued uploads, reuploads, and embeddings always fetch ahead of a queued analysis CTAS on the unchanged `ingest` queue.

**Moved out:** recommendation 3 (per-tenant active cap) now lives on #696 alongside the other resource-governance knobs.

## The change, once the decision is made

Move `materialize_analysis` from `queue="ingest"` to `queue="analysis"` (`backend/app/processing/analysis/tasks.py:834`) and add the queue to the default worker configuration: `worker_queues` at `backend/app/core/config.py:207`, and the compose `WORKER_QUEUES` default, which is duplicated in **both** compose files (`docker-compose.yml:423` and `docker-compose.prod.yml:344`). Changing one and not the other leaves prod and dev subscribing to different queue sets.

One correction to the original recommendation: the **validator needs no change**. `validate_worker_queues` (`config.py:484-492`) only strips, rejects empty, and rejects duplicates; there is no allowlist of known queue names to extend.

[...]

PR #703, opened 2026-07-25
perf(jobs): defer analysis below default priority so uploads fetch first

ishiland:
## Summary

Addresses #695, recommendation 1 (the issue stays open for the conditional follow-ons).

`procrastinate_fetch_job_v2` orders by `jobs.priority DESC, jobs.id ASC` — queue **names** only filter, they never rank. With the default single worker (`WORKER_CONCURRENCY=1`) listening on `priority,ingest,raster`, dispatch is strictly FIFO by job id, so one queued 300-second analysis CTAS head-of-line blocked every upload, reupload, and embedding job fleet-wide. The user-visible symptom was a 2 MB GeoJSON upload pending for minutes with nothing naming analysis as the cause.

The materialize defer now rides `.configure(priority=ANALYSIS_JOB_PRIORITY)` (`-10`) on the unchanged `ingest` queue, so interactive ingest wins the fetch whenever both are queued. The supporting index (`procrastinate_jobs_priority_idx_v1`) already exists; no schema, config, or deploy-order implications — this is one argument at the defer site.

**Accepted trade-off** (called out in the issue): a steady upload stream can starve queued analysis indefinitely. Right for background work; surfacing queue wait in the UI is #698's `current_step="queued"` item, and per-op knobs are #696.

**Not done here, by design:** the dedicated `analysis` queue (only pays with a separate worker, and carries the workers-must-listen-before-API-defers deploy trap) and the per-tenant active cap — both remain tracked in #695.

## Verification

- `configure(priority=-10)` verified against installed procrastinate 3.9.0: produces a `JobDeferrer` with `job.priority == -10`, queue preserved; `defer_async_with_tenant` accepts configured tasks by documented design.
- New asserts in the happy-path defer test pin `priority == ANALYSIS_JOB_PRIORITY < 0` and `queue == "ingest"`; negative-checked (test fails with the `.configure` call removed).

```
uv run pytest tests/test_analysis_materialize.py   # 38 passed
```

### DIFF
--- a/backend/app/modules/catalog/datasets/api/router_analysis.py
+++ b/backend/app/modules/catalog/datasets/api/router_analysis.py
@@ -44,6 +44,14 @@ router = APIRouter(
 
 _SAFE_COLUMN_RE = re.compile(r"^[a-zA-Z_][a-zA-Z0-9_]*$")
 
+# fix(#695): Procrastinate ranks by per-job priority (DESC, default 0), not
+# by queue name — so a 300-second analysis CTAS enqueued first would
+# head-of-line block every upload on the shared single worker. Below-default
+# priority lets interactive ingest win the fetch whenever both are queued.
+# ponytail: a steady upload stream can starve queued analysis indefinitely —
+# acceptable for background work; a per-op budget knob is #696's scope.
+ANALYSIS_JOB_PRIORITY = -10
+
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

Issue #704, opened 2026-07-25
chore(db): move db image to postgis/postgis:18-3.6 — bullseye base EOLs 2026-08-31; GEOS 3.9→3.13 speeds analysis ops 1.5–2.3×


ishiland:
## Why now

`db/Dockerfile` builds `FROM postgis/postgis:17-3.5`, and that upstream tag is effectively frozen:

- Base is Debian **bullseye** (`pgdg110`), whose LTS ends **2026-08-31** (next month).
- The tag still ships **PG 17.5** (May 2025) — no rebuild in over a year, so we're missing a year of PostgreSQL minor/security releases. (PG 17 itself is supported upstream until Nov 2029; the forcing function is the image line, not PG EOL.)
- It bundles **GEOS 3.9.0 (Dec 2020)** and **PROJ 7.2.1**, which sit directly under the analysis hot paths (`ST_Buffer` ×7 call sites, `ST_Union` ×5, `ST_Intersection` ×6, `ST_MakeValid` ×18, `ST_Subdivide` ×1).
- There is **no `17-3.6` tag** upstream. Getting a current PostGIS/GEOS from the official image means the 18 line: `postgis/postgis:18-3.6` = PG 18.4 + PostGIS 3.6.4 on **trixie** (`pgdg13`), with GEOS 3.13.1 + PROJ 9.6.0.

## Measured speedups (17-3.5 vs 18-3.6, identical seeded workload)

| Benchmark (mirrors analysis ops) | 17-3.5 / GEOS 3.9 | 18-3.6 / GEOS 3.13 | Speedup |
|---|---|---|---|
| `ST_Union` 10k overlapping polys (dissolve) | 2299 ms | 1060 ms | **2.2×** |
| `ST_Union` 50k | 9495 ms | 4269 ms | **2.2×** |
| `ST_Union` 100k | 16023 ms | 6833 ms | **2.3×** |
| `ST_Buffer` 200k points | 1666 ms | 1114 ms | 1.5× |
| `ST_MakeValid` 100k bowties (default) | 5883 ms | 3393 ms | 1.7× |
| `ST_MakeValid` 100k (`method=structure`) | 5639 ms¹ | 1047 ms | **5.4×** |
| `ST_Intersection` 50k pairs (clip) | 1147 ms | 703 ms | 1.6× |
| `ST_Transform` 200k points (3857→4326) | 770 ms | 766 ms | 1.0× |

[...]

PR #714, opened 2026-07-26
fix(runbook): stop the PG 18 upgrade from overwriting real backup history

ishiland:

## The problem

RUNBOOK § 6 step 3 starts the whole stack, and the `backup` service runs a dump the moment it starts — `scripts/backup-entrypoint.sh:354` calls `run_backup` unconditionally on container start, with no skip-if-recent guard. At that point the new cluster is freshly migrated and **empty**.

The chain:

1. Step 3's `docker compose up -d --wait` starts `backup` (it has no `profiles:` key — "Backup runs by default", `docker-compose.yml:576`).
2. It writes a schema-only dump into `/backups/daily` as the newest file.
3. `prune_old_backups` globs `*.dump` and deletes oldest-by-mtime to stay inside `BACKUP_RETENTION_DAILY` (default 7) — so a real generation is dropped to make room.
4. `docker compose down` at step 2 carries no `-v`, so `backup_data` survives every retry and accumulates one more empty dump per pass.

The empty dumps use the same `geolens_<timestamp>.dump` naming as real ones and are the newest on disk. § 5 tells an incident responder to `ls -lt` and take the latest. That dump passes the `pg_restore -f /dev/null` gate — a schema-only archive is a perfectly valid archive, and `restore.sh` has no row-count or size sanity check — and is then applied with `--clean --if-exists`. **An incident on upgrade evening restores an empty database over production.**

`pre_pg18_manual.dump` itself is not the loss: step 1 mandates copying it to the host and step 4 restores from that copy. What is destroyed is the operator's retained backup history.

## The fix

Step 3 runs with `--scale backup=0`; step 5 brings the service up once the cluster holds real data. Nothing `depends_on` backup, so scaling it to 0 is safe.

This is an omission rather than a design choice — § 2 already treats stopping `backup` during recovery as expected practice ("Re-enable the backup service if it was stopped during recovery: `docker compose start backup`"). § 6 just never said to.

## Three further doc defects on the same paths

- **Step 0's pre-upgrade dump was verified with `ls -l`.** It is the only rollback artifact once step 2 removes the volume. A truncated archive looks fine to `ls` and also passes `--list`; corruption would not surface until `restore.sh` validates at step 4 — after the source of truth is gone. Now read back with `pg_restore -f /dev/null` while the old cluster still exists to re-dump from.
- **§ 2 and § 5 pass a literal `-` for stdin.** That alias was never documented for `pg_restore` and PG 18 rejects it. Verified against the bundled image:
  ```
  $ docker compose exec -T db pg_restore --version
  pg_restore (PostgreSQL) 18.4 (Debian 18.4-1.pgdg13+1)
  $ ... | pg_restore --list -
  pg_restore: error: could not open input file "-": No such file or directory
  ```
  That reads as "every backup is corrupt" when the dumps are fine. Omitting the filename reads stdin correctly. `chore(#704)` already fixed this in `restore.sh`; the docs lagged.
- **§ 2 described `restore.sh`'s gate as `pg_restore --list`.** It has been `-f /dev/null` since #710, and the distinction is the whole point of that change.

## Verification

- `pg_restore --list -` vs omitted-filename behavior confirmed live against the running PG 18.4 `db` container (above).
- `backup` has no dependents: confirmed via `docker compose config`.
- `run_backup` on start (`backup-entrypoint.sh:353-354`) and the `*.dump` glob in `prune_old_backups` (`:230-242`) read directly.
- Docs-only change; `pre-commit run --files RUNBOOK.md` passes.

Release impact: the PG 17 → 18 upgrade is the breaking change in 1.5.0, so this procedure is about to get its first real-world use.

### DIFF
--- a/RUNBOOK.md
+++ b/RUNBOOK.md
@@ -115,7 +115,9 @@ self-hosted Docker Compose deployment).
 
 `scripts/restore.sh` is the **canonical operator-facing restore entry point**. It:
 
-1. Validates the dump with `pg_restore --list` — aborts if the file is corrupt.
+1. Validates the dump with `pg_restore -f /dev/null` — reads every data block
+   and aborts if the file is corrupt or truncated. (`--list` would only read the
+   table of contents, which a truncated archive still passes.)
 2. Creates required extensions and schemas in the database.
 3. Stops `api` and `worker` to prevent write conflicts.
 4. Runs `pg_restore --clean --if-exists --no-owner` against the bundled `db` container.
@@ -197,9 +199,12 @@ output rather than hand-editing it.
 docker run --rm -v <project>_backup_data:/backups:ro alpine \
   ls -lt /backups/daily
 
-# Validate a specific dump after copying it out of the volume (step 0 above)
+# Validate a specific dump after copying it out of the volume (step 0 above).
+# Omit the filename to read stdin — the literal `-` alias was never documented
+# for pg_restore and PG 18 rejects it with `could not open input file "-"`,
+# which reads like a corrupt backup when the dump is fine (chore(#704)).
 docker compose exec -T db \
-  pg_restore --list - < ./restore/geolens_<YYYYmmdd_HHMMSS>.dump | head -20
+  pg_restore --list < ./restore/geolens_<YYYYmmdd_HHMMSS>.dump | head -20
 ```
 
 ---
@@ -457,9 +462,14 @@ docker run --rm \
   alpine sh -c 'cp /backups/daily/geolens_<YYYYmmdd_HHMMSS>.dump /out/ && \
                 cp /backups/daily/staging-<YYYYmmdd_HHMMSS>.tar.gz /out/ 2>/dev/null; ls -lh /out'
 
-# Validate the candidate dump before restoring
+# Validate the candidate dump before restoring. Omit the filename to read
+# stdin: PG 18 rejects the literal `-` with `could not open input file "-"`,
+# which reads like a corrupt backup when the dump is fine (chore(#704)).
+# `-f /dev/null` reads every data block; `--list` alone would pass a dump
+# truncated after its table of contents.
 docker compose exec -T db \
-  pg_restore --list - < ./restore/geolens_<YYYYmmdd_HHMMSS>.dump | head -20
+  pg_restore -f /dev/null < ./restore/geolens_<YYYYmmdd_HHMMSS>.dump \
+  && echo "candidate dump reads end-to-end"
 ```
 
 If the daily dump is corrupt, fall back to a weekly backup under
@@ -533,12 +543,18 @@ set of server binaries.
 
 ```bash
 # 0. Take a fresh pre-upgrade dump WITH THE OLD STACK STILL RUNNING, then
-#    verify it exists. (The scheduled backup service works too — this just
-#    guarantees a dump from the moment of the upgrade.)
+#    verify it reads back end-to-end. (The scheduled backup service works too —
+#    this just guarantees a dump from the moment of the upgrade.)
 docker compose exec backup sh -c \
   'PGPASSWORD="$POSTGRES_PASSWORD" pg_dump -Fc -h "$POSTGRES_HOST" -U "$POSTGRES_USER" "$POSTGRES_DB" \
    > /backups/daily/pre_pg18_manual.dump'
-docker compose exec backup sh -c 'ls -l /backups/daily/pre_pg18_manual.dump'
+#    `-f /dev/null`, NOT `ls -l` and NOT `--list`: this dump is the only
+#    rollback artifact once step 2 removes the volume, and a truncated archive
+#    both looks fine to `ls` and passes `--list` (the -Fc table of contents sits
+#    at the front). Read every data block now, while the old cluster still
+#    exists to re-dump from.
+docker compose exec backup sh -c \
+  'pg_restore -f /dev/null /backups/daily/pre_pg18_manual.dump && echo "pre-upgrade dump OK"'
 
 # 1. Copy the dump out of the backup volume to the host (survives volume ops).
 #    Replace <project> with your Compose project name (see `docker volume ls`).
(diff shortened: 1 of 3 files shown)


ishiland:
@codex review

ishiland:
Good catch — fixed in e0390523e, and it turned out to point at a better fix than the one suggested.

Rather than telling the operator to remember to delete the volume copy, the dump now never enters the retention pool: it is written straight to the **host** at `./backups/pre-upgrade/pre_pg18.dump`. `scripts/upgrade.sh:179` already established that convention for its own rollback dump, and streams it out the same way (`exec -T` at `upgrade.sh:207`). `prune_old_backups` only ever runs against `DAILY_DIR` and `WEEKLY_DIR`, so nothing under `pre-upgrade/` is ever pruned, and `backups/` is already gitignored.

That also **removes** the old step 1, which copied the dump out of the volume with a bare `docker run -v <project>_backup_data:...` and required the operator to look up their Compose project name.

Two related changes in the same commit:

- Connection variables are still read inside the `backup` container where they are set — only the destination moved. `-T` is required so Docker does not allocate a TTY and corrupt the binary stream.
- `docker compose stop api worker` before the dump. `pg_dump` snapshots when it *begins* and does not block writers, so a write acknowledged after it starts would be missing from the rollback dump — `upgrade.sh` guards this for the same reason. They come back with the stack at step 3. Flagging it explicitly since it is slightly beyond the reported issue; happy to drop it if you'd rather keep this PR narrow.

Verified against the running stack rather than reasoned about:

```
$ docker compose exec -T backup sh -c 'PGPASSWORD=... pg_dump -Fc ...' > ./backups/pre-upgrade/pre_pg18.dump
$ file pre_pg18.dump
PostgreSQL custom database dump - v1.16-0        (20M)
$ docker compose exec -T backup pg_restore -f /dev/null < pre_pg18.dump
pre-upgrade dump OK
```

And truncating that same archive to 40% confirms why the documented check is `-f /dev/null` rather than `--list`:

```
--list:        PASSED   (false green)
-f /dev/null:  correctly REJECTED
```

---

One correction to my own PR description, unrelated to your comment: I wrote that the step-0 dump "is the only rollback artifact once step 2 removes the volume." That overstates it — `docker compose down` carries no `-v`, so `backup_data` and its scheduled dumps survive step 2. The step-0 verification is hardening, not the last line of defence. The interaction is still worth noting: the surviving `backup_data` history is exactly what the empty-cluster dumps in this PR's main finding were poisoning.

ishiland:
@codex review

ishiland:
Fixed — the P1 is real and it was the worst kind: a data-loss path in the document people read when they have nothing else.

**What was wrong.** The block is written to be pasted. A failed `pg_dump` or a failed read-back only suppressed step 1's `echo`; the shell then ran on into `docker compose down` and `docker volume rm <project>_pgdata`. Past that point there is neither a dump nor a cluster.

**The fix.** Step 1 dumps to `.partial` and renames only after `pg_restore -f /dev/null` passes, so `pre_pg18.dump` exists **iff** a verified dump exists; step 2 is gated on `test -f` of exactly that file and chained with `&&`.

One case your suggestion doesn't cover, which I added: without a leading `rm -f`, an already-verified dump from an *earlier* attempt green-lights step 2 after a re-dump fails — and that older dump predates the step-0 quiesce. Simulated all four paths; only all-succeed reaches the volume removal:

```
A) pg_dump fails:                          step2: blocked, PG17 volume intact
B) dump ok, verify fails:                  step2: blocked, PG17 volume intact
C) both succeed:                           step2: *** pgdata VOLUME DESTROYED ***
D) stale verified dump, then dump fails:   step2: blocked, PG17 volume intact
```

**Same defect, second surface.** Chasing this found `scripts/upgrade.sh` — the path most self-hosters will actually take to 1.5.0 — one notch weaker: it accepted any non-empty file as its rollback artifact. A dump truncated by a disk that filled mid-write is non-empty, passes `--list` (the -Fc TOC is at the front), and the script would then migrate the schema behind an unrestorable backup. It now reads the archive back before the first irreversible step and discards the file if it fails.

`test-upgrade-order.sh` grew an ordering assertion (`backup < verify < pull`) and CASE 6b for the corrupt-dump abort. Removing the check fails 4 assertions. One note: my first version of the discard assertion was timestamp-flaky — the dump filename carries a whole-second stamp, so it also matched CASE 1's artifact and passed or failed depending on whether the two cases landed in the same second. Scoped it and confirmed stable over 8 consecutive runs.

@codex review

ishiland:
Fixed. This is worse than your report says, and the extra bit is the reason it matters.

Retrying from the top doesn't just delete the last recovery artifact and then fail. The fresh PG 18 cluster is **up and answering**, so the re-dump *succeeds*, passes `pg_restore -f /dev/null`, and step 2 destroys the volume again. The operator ends up holding a perfectly valid, fully verified dump — of an empty database. Every signal says the backup is good.

My `rm -f` caused this. It was there for a real reason (a stale verified dump green-lighting step 2 after a re-dump failed), but it traded a narrow failure for a total one.

Dumps are now timestamped and this attempt's file is carried in `$DUMP`, so nothing is ever deleted or overwritten. `$DUMP` also preserves what the `rm -f` was protecting: the gate is scoped to the dump *this* attempt verified.

Simulated three sequential attempts, including the empty-cluster retry:

```
Attempt 1 — succeeds, volume destroyed:       surviving verified dumps: 1
Attempt 2 — retry, PG18 up and EMPTY:         surviving verified dumps: 2
Attempt 3 — retry again:                      surviving verified dumps: 3
```

Attempt 1's dump — the only one holding real data — survives all of them.

And step 2 pasted alone into a fresh shell, with an old verified dump sitting on disk:

```
step2 blocked: DUMP unset in this shell (old dump on disk did NOT green-light it)
```

That's the intended answer rather than a limitation: re-run step 1 so the dump you're betting the cluster on is one you verified in this session, against the cluster still standing. Step 4 says how to pick a dump by hand if the shell is genuinely gone, and the rollback section notes the directory is deliberately never pruned.

32/32 upgrade-order and 15/15 doc-consistency still pass.

@codex review

PR #715, opened 2026-07-26
docs(changelog): announce the analysis toolkit, and gate the release header

ishiland:

## The release notes omit the release

`[Unreleased]` had `### Changed` and `### Fixed` but **no `### Added`** — so the headline feature of 1.5.0, the entire analysis toolkit, appeared nowhere in the release notes. This adds it, plus two unannounced operator-facing changes.

## The header is load-bearing and unguarded

`.github/workflows/release.yml:59` builds the release body by matching the header literally:

```
NOTES=$(awk "/^## \[${VERSION}\]/{found=1; next} /^## \[/{if(found) exit} found{print}" CHANGELOG.md)
```

If it matches nothing, lines 61-64 fall back — with no warning and exit 0 — to:

```
git log --format="- %s" "${PREV}..${TAG}" --no-merges | grep -v "^- docs(" | grep -v "^- chore(" | head -30
```

In the actual `v1.4.13..HEAD` range those two greps remove exactly two commits, and one of them is:

- `chore(db): upgrade bundled database to PostgreSQL 18 + PostGIS 3.6 (#705)` — dropped by `grep -v "^- chore("`
- `docs(runbook): §6 must rebuild the backup image too, not just db (#706)` — dropped by `grep -v "^- docs("`

So tagging `v1.5.0` with the header still reading `## [Unreleased]` would publish a release page that **never mentions the breaking PostgreSQL 17→18 upgrade or RUNBOOK § 6** — on the surface a self-hoster reads before running `docker compose pull && docker compose up -d`. The `upgrade.sh` major-boundary guard from #707 only covers the `upgrade.sh` path, not a direct compose pull.

**Nothing in CI read CHANGELOG.md.** `bump_version.py`'s site list excludes it, `check_version_coherence.py` never opened it, no pre-commit hook references it, and the three tests that do read it (`test_phase_275_changelog_routes.py`) only validate a legacy `### Added — Map Builder API surface` block. The last release got the header right purely by hand-discipline.

`make version-check` now asserts a `## [<canonical>]` section exists — green today on 1.4.13, red at bump time until the header is renamed.

## Two unannounced operator-facing changes

Both are reachable without touching a new feature:

- **CORS wildcard rejection (#689).** An operator upgrading with `cors_allowed_origins = "*"` already stored gets a 422 on the Network settings page for a value they never touched. Includes the upgrade note that an existing stored `*` keeps working until the next save.
- **TiTiler `MALLOC_ARENA_MAX=2` (#678)**, with the `TITILER_MALLOC_ARENA_MAX` override, stated as an arena-growth bound rather than a raster memory fix.

## Verification

- `uv run python scripts/check_version_coherence.py` → passes, and now reports the CHANGELOG section.
- Confirmed the gate is real: `## [1.4.13]` matches, `## [1.5.0]` does not — so it goes red at bump until the header is written.
- `pytest tests/test_phase_275_changelog_routes.py` → 3 passed (the new plain `### Added` does not collide with the legacy `### Added — Map Builder API surface` block those tests parse).
- CORS and TiTiler claims read from `backend/app/modules/settings/schemas.py:557-575` and `docker-compose.yml:496` / `.env.example:306`.

## Still to do at release time

This PR does **not** rename the header — that belongs in the `make bump VERSION=1.5.0` commit. The gate now forces it. The bump also needs the link reference:

```
[Unreleased]: https://github.com/geolens-io/geolens/compare/v1.5.0...HEAD
[1.5.0]: https://github.com/geolens-io/geolens/compare/v1.4.13...v1.5.0
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -583,6 +583,13 @@ jobs:
       - name: Check all version sites agree
         run: python scripts/check_version_coherence.py
 
+      # The CHANGELOG half of that gate only helps if it asks the same question
+      # release.yml asks. This asserts the two extractions agree case for case,
+      # including the empty-section and prerelease-tag paths that both fall back
+      # to a filtered git log (fix(#715)).
+      - name: Check the CHANGELOG gate tracks release.yml
+        run: python scripts/tests/test-changelog-gate.py
+
   # ---------- docs contract ----------
   # Enforce docs-contract.json: validates the contract against install.sh +
   # pyproject (so it can't drift from the binaries) and scans the READMEs for
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -54,9 +54,18 @@ jobs:
         run: |
           PREV="${{ steps.prev.outputs.tag }}"
 
-          # Extract section from CHANGELOG.md for this version
+          # Extract section from CHANGELOG.md for this version.
+          #
+          # fix(#715): strip any prerelease suffix first. A prerelease tag has no
+          # CHANGELOG section of its own and never will — bump_version.py's
+          # SEMVER_RE only accepts plain X.Y.Z, so `make bump` cannot create a
+          # `## [1.5.0-rc.1]` header, and check_version_coherence.py only ever
+          # validates the canonical version. Without this, every RC published the
+          # git-log fallback below, whose filter drops every docs(/chore( subject.
+          # An RC's notes are the notes for the release it is a candidate for.
           VERSION="${TAG#v}"
-          NOTES=$(awk "/^## \[${VERSION}\]/{found=1; next} /^## \[/{if(found) exit} found{print}" CHANGELOG.md)
+          CHANGELOG_VERSION="${VERSION%%-*}"
+          NOTES=$(awk "/^## \[${CHANGELOG_VERSION}\]/{found=1; next} /^## \[/{if(found) exit} found{print}" CHANGELOG.md)
 
           if [ -z "$NOTES" ]; then
             # Fallback: generate from git log
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -67,7 +67,12 @@ jobs:
           CHANGELOG_VERSION="${VERSION%%-*}"
           NOTES=$(awk "/^## \[${CHANGELOG_VERSION}\]/{found=1; next} /^## \[/{if(found) exit} found{print}" CHANGELOG.md)
 
-          if [ -z "$NOTES" ]; then
+          # fix(#715 review): strip whitespace before the emptiness test. `$(...)`
+          # removes trailing newlines but keeps spaces and tabs, so a section
+          # containing only whitespace made `[ -z ]` false and published
+          # effectively blank release notes — passing neither the fallback nor
+          # the version-coherence gate, which rejects exactly that input.
+          if [ -z "$(printf '%s' "$NOTES" | tr -d '[:space:]')" ]; then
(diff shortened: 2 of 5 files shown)


ishiland:
@codex review

ishiland:
Correct, and my wording had it backwards. Fixed in the latest commit.

Verified in the middleware:

```python
# backend/app/api/middleware/cors.py:84-87
origins = {o.strip() for o in raw.split(",") if o.strip()}
if "*" in origins:
    _origins_cache = (now, set())
    return False
```

That predates this release and is unchanged by it — the new validator guards the **write** path only. So an operator upgrading with `*` already stored is *already* denying every cross-origin request, and my note invited them to leave it in place.

The note now reads:

> **Upgrade note:** if `*` is already stored, cross-origin requests are *already* being denied — the new validator guards the write path, not the middleware, which has always turned an allowlist containing `*` into an empty set. Replace it with explicit origins to restore cross-origin access; Admin → Settings → Network will refuse to save the page until you do.

ishiland:
@codex review

ishiland:
Both fixed — and both reproduce.

**Empty section.** Confirmed against release.yml's actual awk:

```
VERSION=1.5.0    (header present, no body)  -> EMPTY -> git-log fallback
VERSION=1.4.13   (header + notes)           -> notes: ### Fixed - something real
```

The gate now runs the same extraction and requires non-whitespace content. This one mattered for the bump about to happen: adding the header and postponing the notes is exactly what release prep does.

**Prerelease.** Also real, and it has precedent — `v1.4.0-rc.1` is already in the tag list, and the workflow fires on any v-prefixed tag. I took your second option (normalize in release.yml) rather than validating the tag, because an RC's notes *are* the notes for the release it is a candidate for, and `## [1.5.0-rc.1]` can never exist anyway: `SEMVER_RE` rejects the suffix, so `make bump` cannot write that header. One `${VERSION%%-*}`.

Worth noting the adjacent thing is already right: an RC is marked as a GitHub pre-release, so it never becomes the `latest` pointer `install.sh` follows. Only the notes lookup was broken.

**Test.** `scripts/tests/test-changelog-gate.py` reads the awk program *back out of release.yml* and diffs it against the gate's Python across seven section shapes — normal, empty, whitespace-only, missing, last-in-file, followed by link refs, with nested `###` headers — asserting both agree on which inputs trigger the fallback. Editing the workflow's extraction without updating the gate now fails CI. 19/19, wired into the version-coherence job.

(My first draft of that test unescaped `\[` when reading the awk out of the YAML, which built an invalid program that matched nothing and reported 12 false failures. Inside shell double quotes the backslash reaches awk intact — fixed and noted in the source, since it is an easy trap for the next person.)

@codex review

ishiland:
Good catch — and the test was wrong in the specific way that matters: it modelled the *gate's* verdict instead of the *shell's*, which is exactly the drift this test exists to catch.

I fixed release.yml rather than relaxing the gate. Publishing whitespace as release notes isn't a better outcome than the git-log fallback, and the fallback exists for precisely this case:

```sh
if [ -z "$(printf '%s' "$NOTES" | tr -d '[:space:]')" ]; then
```

Now the two agree on whitespace-only, which makes `.strip()` the correct model here rather than a convenient one — noted in the test so the next reader doesn't have to re-derive it.

The test also asserts the emptiness test itself now, not just the extraction. Reverting release.yml to a plain `[ -z "$NOTES" ]` fails it:

```
not ok 16 - release.yml no longer whitespace-strips before its emptiness test
```

20/20.

@codex review

PR #716, opened 2026-07-26
fix(analysis): release the request session before the sandbox preview query

ishiland:

## The defect

`execute_safe` deliberately opens its **own** connection rather than using the caller's session — it needs `SET TRANSACTION READ ONLY` and `SET LOCAL ROLE`, which it cannot apply to a session the handler is still using (`executor.py:121-122` says so explicitly). But the preview handler was still holding its own connection at that moment:

- `router_analysis.py:126` injects `Depends(get_db)`.
- `get_db` (`core/dependencies.py:9-15`) yields a session that autobegins and pins a pooled connection at the first `get_dataset` execute in `_load_vector_dataset`.
- `service_analysis.py` then calls `execute_safe`, which does `engine.connect()` on the **same** engine (`core/db/session.py:25`, `pool_size=10 / max_overflow=3 / pool_timeout=30`).

**Two of 13 slots per in-flight preview.** Degradation starts at 7 concurrent previews (demand 14 > 13). Past that, waiters block for the full 30s `pool_timeout` with the pool empty — so **every endpoint on that worker stalls, not just analysis**. Tile routers take `Depends(get_db)` for auth/metadata, so tile requests stall too even though tile SQL runs on a separate asyncpg pool.

The waiter then raises SQLAlchemy `TimeoutError`, which `_handle_execution_error` matches as neither a cancellation nor a statement timeout nor a `DataError` — so it falls through to `query_failed`, which `_SANDBOX_STATUS` does not map. The user gets **HTTP 500**, not a 429 or 503.

Nothing upstream bounds it: no `--limit-concurrency` on uvicorn, no `limit_conn` in nginx, and SlowAPI's default limit is a per-IP request *rate*, which does not bound in-flight concurrency of slow requests. The advisory lock inside `execute_safe` does not help — it is taken *after* the second connection is checked out, so N concurrent requests from a single principal still demand 2N checkouts.

**Not new in kind** — `validate_and_execute` had the same shape pre-1.5. What 1.5 changed is exposure: from an LLM-gated chat turn to an always-on authenticated HTTP endpoint, reachable from the SDK, CLI, and the builder's Preview button.

## The fix

Snapshot `feature_count` (the only Dataset attribute read after the query), then `await db.rollback()` before `execute_safe`. Three lines.

The snapshot is required, not cosmetic: `rollback()` expires the ORM objects, so a later `dataset.feature_count` would silently re-query and re-check-out.

Measured on the running stack:

```
after query (session holds conn): 1
after rollback:                   0
sandbox conn open, total:         1     <- was 2
```

## Why this is safe for the other caller

`run_analysis_preview` is also called from the chat `run_analysis` tool. Checked both ways it could bite:

- No session writes in `chat_analysis.py` / `chat_actions.py` — the chat tools are read-only by design, and nothing is pending on the session when the tool runs.
- The one `await db.commit()` in the AI module (`ai/router.py:419`) is on the **generate-map** path, and `run_analysis` is not reachable from it: `generate_map_from_prompt` passes `ANTHROPIC_TOOLS`, whose tool set is `['search_datasets', 'get_dataset_details']`. `run_analysis` only reaches the map-surface chat loop via `select_chat_tools`.
- `enforce_ai_token_budget` (`ai/router.py:253-263`) is read-only.

## Test

Asserts `in_transaction()` at the moment `execute_safe` is entered — **not** `pool.checkedout()`. The test engine uses NullPool, so checkout counts there say nothing about the production QueuePool; an open transaction is what pins the connection, so it is the pool-independent form of the same invariant.

Verified load-bearing — with the `rollback()` removed:

```
FAILED tests/test_analysis_preview.py::TestAnalysisPreviewEndpoint::test_preview_releases_request_session_before_sandbox_query
```

## Verification

- `pytest tests/test_analysis_preview.py tests/test_analysis_materialize.py tests/test_ai_chat_run_analysis.py tests/test_layering.py tests/test_sandbox.py` → **181 passed**
- `ruff check` + `ruff format --check` → clean; full pre-commit hook set passes (including the Rule 1 visibility hook).

### DIFF
--- a/backend/app/modules/catalog/datasets/api/router_analysis.py
+++ b/backend/app/modules/catalog/datasets/api/router_analysis.py
@@ -138,7 +138,14 @@ async def analysis_preview_endpoint(
     )
     try:
         return await run_analysis_preview(
-            db, dataset, body, user.id, mask_dataset=mask_dataset
+            db,
+            dataset,
+            body,
+            user.id,
+            mask_dataset=mask_dataset,
+            # fix(#716): safe here — `user.id` is evaluated above, and neither
+            # this handler nor any middleware reads ORM state afterwards.
+            release_session=True,
         )
     except ValueError as exc:
         raise HTTPException(
--- a/backend/app/modules/catalog/datasets/domain/service_analysis.py
+++ b/backend/app/modules/catalog/datasets/domain/service_analysis.py
@@ -137,6 +137,7 @@ async def run_analysis_preview(
     user_id: uuid.UUID,
     *,
     mask_dataset: Dataset | None = None,
+    release_session: bool = False,
 ) -> AnalysisPreviewResponse:
     """Execute a preview operation and assemble a GeoJSON FeatureCollection.
 
@@ -147,6 +148,17 @@ async def run_analysis_preview(
     ``mask_dataset`` (clip only) sources the mask from another dataset's
     unioned geometries; the CALLER owns its visibility check, exactly as it
     owns the source dataset's.
+
+    ``release_session`` returns the caller's pooled connection before the
+    sandbox query (see below). OPT-IN, because the rollback that releases it
+    expires EVERY ORM instance on the session, not just ``dataset`` — including
+    the authenticated ``User``, whose next attribute read would then attempt a
+    sync refresh and raise ``MissingGreenlet`` (verified: after a rollback even
+    ``user.id`` raises). Only pass it from a caller that owns the session and
+    reads nothing off it afterwards. The REST endpoint qualifies — it evaluates
+    ``user.id`` before the call and touches no ORM state after, and no
+    middleware reads the ORM user post-handler. The AI chat tool does NOT: both
+    chat paths read ``user.id`` again after the tool returns.
     """
     table_ref = _safe_table_ref(dataset.table_name)
     mask_table_ref = (
@@ -165,7 +177,8 @@ async def run_analysis_preview(
     # → HTTP 500. `rollback()` returns the connection (measured: checkedout
     # 1 → 0), so a preview costs one slot instead of two.
     source_feature_count = dataset.feature_count
-    await db.rollback()
+    if release_session:
+        await db.rollback()
     result = await execute_safe(
         db,
         sql,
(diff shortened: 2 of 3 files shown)


ishiland:
@codex review

ishiland:
Confirmed and fixed in ffd515d9c. You were right, and my "safe for the other caller" analysis was wrong in a specific way worth naming: I checked for pending **writes** on the shared session and never considered expired-instance **reads**.

Reproduced directly before fixing:

```
>>> user = (await s.execute(select(User).limit(1))).scalars().first()
>>> await s.rollback()
>>> user.id
sqlalchemy.exc.MissingGreenlet: greenlet_spawn has not been called; can't call await_only() here
>>> user.username
sqlalchemy.exc.MissingGreenlet: ...
```

Note even the **primary key** raises — it is not served from the identity key once the instance is expired, which is worse than I'd have guessed.

## The fix

`release_session` is now keyword-only, defaulting to `False`. Only the REST endpoint opts in.

That is safe there and only there:
- `user.id` is evaluated in the argument list, before the call.
- The handler reads no ORM state afterwards.
- No middleware reads the ORM user post-handler — nothing in the tree references `request.state.user`, and none of the nine registered middlewares touch it.

The chat path keeps its two connections. That is unchanged from today's behaviour, so no regression — just a smaller win than the original commit claimed.

## Test

`test_preview_does_not_release_the_session_by_default` loads a `User`, calls the service through the default path, then reads `user.id` and `user.username`. Verified load-bearing — restoring the unconditional rollback gives exactly your predicted failure:

```
E  sqlalchemy.exc.MissingGreenlet: greenlet_spawn has not been called; can't call await_only() here
```

Worth flagging: the existing `test_ai_chat_run_analysis.py` suite passed against the broken version, so chat-path ORM-state survival had no coverage at all. This test closes that.

`pytest tests/test_analysis_preview.py tests/test_analysis_materialize.py tests/test_ai_chat_run_analysis.py tests/test_processing_port.py tests/test_layering.py tests/test_sandbox.py` → **184 passed**.

ishiland:
@codex review

Issue #694, opened 2026-07-25
feat(analysis): pre-flight per-operation feature_count gates (dissolve OOM guard, buffer output ceiling)


ishiland:
From the 2026-07-25 analysis scalability audit. Two failure modes converge on one mechanism: an at-enqueue, per-operation gate on `dataset.feature_count` (already loaded on the ORM row in the materialize route).

**Dissolve can OOM-kill the database container.** `ST_Union` as an aggregate accumulates every input geometry before the cascaded union runs (~1 GB practical wall), against `shared_buffers=512MB` and the 2 GB `mem_limit` on the db service. Memory grows faster than the 300s statement timeout elapses, so the kernel OOM-kills a backend and the postmaster resets **every connection in the cluster** — the whole app blips, all tenants. `ST_UnaryUnion(ST_Collect(...))` does not change the ceiling (identical accumulation). Grouped dissolve is worse under HashAggregate (all groups' arrays held simultaneously); GroupAggregate holds one at a time, and without stats on `by_field` the planner's choice is effectively random.
- Guard: reject above a `feature_count` threshold before the CTAS (`backend/app/processing/analysis/tasks.py`, pre-CTAS) or at enqueue with a clear 422.
- Plus `SET LOCAL enable_hashagg = off` for the grouped-dissolve CTAS to force the bounded sorted path.
- Ceiling-raiser later (L): hierarchical union (union per grid cell, then union the cells).

**Buffer is the only amplifying operation and has zero byte accounting.** Vector/PostGIS-resident datasets contribute 0 bytes to storage quota by design (`backend/app/modules/quota/service.py:34-42` defers to cloud metering), and `check_upload_quota` is called with `incoming_bytes=0` at enqueue. Buffer turns one API call that uploads nothing into an arbitrarily large table (500k inputs × ~32 output vertices at the 100 km distance ceiling), bounded only by the dataset-count quota, on shared infrastructure.
- Cap `feature_count` for buffer specifically (S), or estimate `feature_count × vertices-per-feature` against a configurable ceiling next to `MAX_BUFFER_METERS` (`backend/app/platform/analysis_sql.py:31`) (M).
- Belt-and-braces: post-CTAS `pg_total_relation_size` check before registration, failing into the existing cleanup path (which already drops the table correctly).

Caveat both gates honestly: `feature_count` is an ingest-time snapshot — a warning threshold, not a correctness gate.

PR #718, opened 2026-07-26
fix(builder): let the analysis size limits reach the person who hit them

ishiland:
Every blast-radius refusal added in this release rendered as **"The submitted values are invalid."**

`error-map.ts` drops unrecognised backend prose by design and falls back to the generic per-status string. No mapping existed for the analysis gates, so someone who tried to dissolve too large a layer was told their input was invalid — with no hint that size was the problem, and no number to filter down to. The server's prose carries the whole point of those changes: the limit, and the way out.

## What now resolves to a locale key

Three messages, following the storage- and dataset-quota patterns already in the file:

- the per-operation source caps (#694, re-checked in the worker by #701)
- the mask-layer cap (#693)
- the CORS wildcard guard (#689), where the explanation *was* the change — it exists to replace a silent, 30s-delayed deny-all with something a reader can act on, which the generic string undid

## Details

- Limits arrive comma-grouped from Python's `{n:,}`, so they are stripped and re-formatted for the active locale (250,000 / 250.000 / 250 000) rather than carrying en-US grouping into every language.
- The operation name in the source-cap message is deliberately dropped rather than interpolated: translating it would mean reaching into the builder namespace from the errors namespace, and the user picked the operation in the panel one click earlier. The limit is the part they do not already know.
- The same message already rendered verbatim on the async job-failure path, so this only affects up-front rejection — which is the common case, since the router gate fires first whenever the cached feature count is accurate.

## Verification

- `vitest src/lib/__tests__/error-map.test.ts` → 15 passed.
- Verified in the browser on all three paths and in German, confirming the sentence and the number grouping both follow the locale.
- Keys added to all four locales (en/es/fr/de) — `npm run test:i18n` passes.

## Known limitation

This maps three specific sentences by regex against backend prose. If the backend wording changes, the mapping silently reverts to the generic string — the exact failure this PR fixes — and nothing couples the two. A durable fix would be an error *code* on the response rather than prose matching; out of scope here, worth tracking separately.


ishiland:
@codex review

ishiland:
Fair, and my own 1.5.0 audit flagged the same thing independently. Fixed in f9bd2f2c7 — swapped `ponytail:` for `fix(#718):`, which is where the reasoning actually lives. The explanation itself is unchanged.

Worth flagging for scope: this is a pre-existing repo pattern, not something this PR invented.

```
$ grep -rn 'ponytail:' backend/app frontend/src scripts | wc -l
15
```

They mark deliberate simplifications and usually name their ceiling and upgrade path (e.g. `router_analysis.py:51` names the analysis-starvation risk and points at #696), so they carry real information — but they are unscoped labels, and AGENTS.md's convention says a lookup-able reference is required. Converting the other 14 is a separate mechanical pass, not scope for this PR; I've left it out deliberately rather than by oversight.

`vitest src/lib/__tests__/error-map.test.ts` → 15 passed.

ishiland:
@codex review

Issue #692, opened 2026-07-25
fix(analysis): correctness & lifecycle hardening batch from the 07-25 audit


ishiland:
From the 2026-07-25 analysis scalability/tenancy/scenario audit (four independent review passes over the merged #680/#682/#690 state). This batch collects the small correctness/security fixes; scale-ceiling work is tracked separately. PR to follow shortly.

1. **Materialize skips the `upload` permission.** The materialize route depends on `get_current_active_user` (`backend/app/modules/catalog/datasets/api/router_analysis.py:133`), while every dataset-creating ingest endpoint requires `require_permission("upload")` (`backend/app/processing/ingest/router.py`). A viewer-role account can therefore create datasets through analysis. Gate materialize; leave preview on the plain active-user dependency (read-only; the chat tool depends on it). Deployments that issued view-only accounts should pick this up in the next patch release.
2. **NULL-geometry rows survive materialize but not preview.** Preview filters `geom_out IS NOT NULL AND NOT ST_IsEmpty(geom_out)`; the worker's cleanup DELETE is clip-only and `ST_IsEmpty`-only (`backend/app/processing/analysis/tasks.py:235-240`), so the saved dataset disagrees with the preview (and dissolve keeps all-NULL groups). Widen to `geom IS NULL OR ST_IsEmpty(geom)` for every operation.
3. **Worker status writes are unfenced.** The task stamps pending→running directly instead of claiming via `claim_ingest_job_attempt` (`tasks.py:130-166` vs the ingest pattern at `backend/app/processing/ingest/tasks_vector.py:300-338`). A job the pending-sweeper already failed (worker down/backlogged >1h), or one the defer orphan-guard failed after the queue INSERT committed, later resurrects to running/complete — the one-job cap is defeated and a dataset appears after the user was shown a failure. Claim on the row's own `attempt_id` (it defaults at creation) and return early when the claim misses. This is the fencing half of #691 and lands independently of the cap-predicate change.
[...]

PR #719, opened 2026-07-26
perf(analysis): use the subdivided mask join for clip materialize too

ishiland:

## The defect

#702 subdivided the clip mask for **previews** and deliberately left **materialize** on a single whole-layer `ST_Union`, on the reasoning that one union amortizes over every row in a batch job.

Measured, that is backwards. Per-row `ST_Intersection` against one giant union geometry is superlinear in mask complexity, so the union shape loses at every size — not just at preview scale.

Benchmarked on the dev stack, 22,324 Manhattan building polygons against a 972-polygon / 249,804-vertex mask (just under `MAX_MASK_LAYER_FEATURES`):

| shape | time | rows |
|---|---|---|
| whole-layer union (old) | **33.2 s** | 12,291 |
| subdivided join (new) | **3.4 s** | 12,291 |

At the union shape's rate a ~250k-row source exceeds the 300 s `MATERIALIZE_TIMEOUT`, and **clip has no source-size cap** — `MAX_SOURCE_FEATURES` covers only `dissolve` and `buffer`. That is how a clip whose preview renders in under a second could fail on "Create dataset": the two paths disagreed about what was feasible.

## Equivalence

Intersection distributes over union, so the two shapes are equal in theory. Verified in practice as well:

- On a 300-polygon / 38.7k-vertex mask: both shapes returned **386 rows**, maximum Hausdorff distance **7.1e-15** (float noise, not a geometric difference).
- The new test pins it on a **multi-row overlapping mask** — the case where the distinction actually bites, since a single-row mask subdivides into pieces that trivially reassemble. It compares the materialized output against the whole-layer union's own result: row count plus symmetric-difference area, with a non-empty guard so it cannot pass vacuously.

One behavioural note: the subdivided shape emits rows whose intersection is NULL or EMPTY where the union shape filtered them in the `WHERE` (190 such rows in the 22k benchmark). Those already fall out in the post-CTAS `DELETE`, which has dropped NULL/EMPTY geometry for **every** operation since fix(#692) — hence the identical final counts.

## Net deletion

With materialize on the join, three things had no callers left and are gone:

- `render_mask_cte`
- `MASK_CTE_NAME`
- `render_geometry_expr`'s `layer_mask` parameter and branch

`render_clip_layer_preview` is renamed `render_clip_layer_join` — it now serves both paths, and the old name asserted a split that no longer exists. Net **-51 lines** of production code against +21.

## Verification

- `pytest tests/test_analysis_materialize.py tests/test_analysis_preview.py tests/test_ai_chat_run_analysis.py tests/test_layering.py` → **132 passed**
- The pre-existing end-to-end `test_clip_by_layer_materialize` (which carries a non-geometry column through and asserts the surviving row and geometry type) passes unchanged — good independent evidence the join shape preserves `carry_cols`, `gid`, and typing.
- `ruff check` + `ruff format --check` clean; full pre-commit set passes including the Rule 1 visibility hook.


ishiland:
@codex review

ishiland:
@codex review

ishiland:
Both fixed. The first one is more serious than the report suggests — my seam repair *introduced* a regression, it didn't just fail to cover a case.

I ran all four candidate expressions against PostGIS 3.6 over two touching mask polygons rather than reasoning about it:

| source | whole-mask (old) | new, no merge | `ST_Dimension = 1` (shipped) | `GeometryType = 'LINESTRING'` |
|---|---|---|---|---|
| LineString across seam | LINESTRING/1 | MULTILINESTRING/2 ✗ | LINESTRING/1 ✓ | LINESTRING/1 ✓ |
| MultiLineString, parts touching at a point | MULTILINESTRING/2 | MULTILINESTRING/2 ✓ | **LINESTRING/1 ✗** | MULTILINESTRING/2 ✓ |
| MultiLineString, disjoint | MULTILINESTRING/2 | MULTILINESTRING/2 ✓ | MULTILINESTRING/2 ✓ | MULTILINESTRING/2 ✓ |
| Polygon across seam | POLYGON/1 | POLYGON/1 ✓ | POLYGON/1 ✓ | POLYGON/1 ✓ |
| Point | POINT/1 | POINT/1 ✓ | POINT/1 ✓ | POINT/1 ✓ |

So the dimension test traded one wrong case for another: it fixed single-part lines and broke multi-part ones, in a case where the mask introduces no seam at all. `GeometryType` matches the old output in all five. The table is now in the docstring, because the next person will reach for `ST_Dimension` too.

(I went in expecting to find `ST_CollectionExtract` always returns a MULTI type in PostGIS 3.x, which would have made the whole premise wrong. It doesn't — row 1 shows `LINESTRING/1` from the old path. Worth testing rather than trusting.)

**Second finding also confirmed.** `_enforce_output_size` runs against the CTAS at `tasks.py:472`, the NULL/EMPTY `DELETE` at `:483`. And clip is the one operation with no entry in `MAX_SOURCE_FEATURES`, so nothing bounds how many bbox-only rows there are. Filter moved into the SELECT. Since the preview already had this predicate and the two must agree, it is now a single named constant in `analysis_sql` rather than a string in both files.

Both tests fail against the previous code — multi-part returns `LINESTRING/1`, and the 2025-decoy case fails as oversized under a 64 KB cap the single real feature clears (measured: 1 row = 16 KB, 2025 rows = 270 KB, so it is not a knife-edge). Full analysis + layering suites: 107 passed.

@codex review

ishiland:
Confirmed and fixed — reproduced against PostGIS directly:

```sql
CREATE TABLE tmpamb.src (gid int, geom_out text, geom_4326 geometry(Point,4326));
SELECT _src.gid, _src."geom_out", _op.geom_out AS geom
FROM tmpamb.src AS _src CROSS JOIN LATERAL (...) AS _op
WHERE geom_out IS NOT NULL;
-- ERROR:  column reference "geom_out" is ambiguous
```

And you're right that it's a regression rather than a pre-existing gap: the whole-mask shape had no lateral, so no collision was possible.

One thing your report didn't cover — the **preview** has the same latent fault. It selects no carry columns, so it never collided in practice, but it joins `_src` all the same, so any dataset with a `geom_out` attribute would have broken preview too the moment anything selected from `_src`. Both are now qualified via a shared `LATERAL_ALIAS`, since the two paths must keep agreeing.

I qualified **only** `geom_out`. My first pass also did `gid` and `ORDER BY gid`, which broke three existing SQL-shape assertions — correctly, as it turns out: `gid` is not ambiguous (the lateral exposes no such column) and leaving `ORDER BY gid` alone preserves the fix(#700 review) property that it can ride the pkey index.

New test builds a source table with a `geom_out TEXT` attribute and materializes a clip through it; it fails with the ambiguity error against the previous code. 79 analysis tests pass.

@codex review

PR #338, opened 2026-06-27
Map Builder code-quality & spec-conformance audit remediation

ishiland:
## Summary

Remediates the GeoLens Map Builder **Code Quality & Mapbox/MapLibre Spec Conformance** audit (`docs-internal/audits/builder-code-quality-review-20260626.md`): 2 P0 security blockers, 14 P1 correctness findings, 12 P2 follow-ups, and the 119-finding subsystem catalog (DRY, complexity, correctness, spec-conformance, dead-code). Work followed the audit's prescribed order — security first, then runtime/export correctness, then shared-contract extraction over local patches, with the god-files actually split.

167 files, ~+11.7k/−4.8k. Each phase was adversarially re-verified and committed atomically.

## What changed

**Security (P0)**
- Embed tokens are now revoked on share-revoke, public→private downgrade, and dataset/layer removal (with Redis cache invalidation + a fail-closed live check). `/m/{token}` now serves a per-token `Content-Security-Policy: frame-ancestors` from the token's allowed origins.

**Runtime & export correctness (P1)**
- Exported MapLibre styles now use the matching `source-layer` (`data.{table}`), emit `cols=` for low-zoom data-driven/label/filter columns, a valid `raster-dem` terrain source, and a DEM color-relief layer.
- Shared filter-grammar validator/normalizer wired into the layer schema, the raw-JSON editor, AI `set_filter`, and export/import.
- Bulk visibility/opacity now matches single-layer behavior (companions + stroke-disabled gate + clusters); folder-group visibility toggles all children; adding a layer while dirty appears immediately; line→data-driven clears stale gradient; catalog drag handle works on touch; non-admin editors with `use_ai_chat` can use builder chat; admin AI status reflects the selected provider; custom share expiration is edition-gated; AI-chat Undo no longer reverts an earlier edit.
- MVT: tenant-agnostic layer name, content-hash **ETag**/304, continuous z6–9 simplification tolerance, shared serve/validator helpers.

**Shared contracts & god-file splits**
- One source of truth for: builder paint-key allowlist + camel/snake alias maps (FE↔BE **parity drift-guards** added), per-render-mode default paint + magic constants, geometry classifier, companion-layer IDs, filter/expression grammar, MVT source-layer/`cols=`.
[...]

PR #720, opened 2026-07-26
fix(builder): stop offering raster layers as analysis sources

ishiland:

## The defect

The Analysis panel's eligibility filter was:

```ts
const datasetLayers = layers.filter((l) => !!l.dataset_id && !l.is_dem);
```

That excludes DEMs but not **ordinary rasters**. A COG orthophoto or a Sentinel scene carries a `dataset_id` and is not a DEM, so the panel listed it — and on a map whose first layer is a raster, it was also **auto-selected**.

The result was a fully enabled form pre-filled with a layer the server cannot process. Preview 422s, `error-map.ts` has no mapping for that message, so the user gets **"The submitted values are invalid."** on a form they never filled in.

Worse on a raster-only map: the panel has an `analysisTools.noLayers` empty state for exactly this situation ("Add a dataset layer to use analysis tools"), and it never fired, because the raster counted as a dataset layer.

## The fix

Analysis needs a **vector** dataset. Rasters have no `geometry_type` — verified against the running stack:

```
$ curl .../api/datasets/6a64ba12-…   # Sentinel-2 TCI
geometry_type: None   table_name: 'stac_50b1bb496be04459'
```

— and that is precisely the precondition `_load_vector_dataset` enforces server-side (`router_analysis.py:77`, 422 "Analysis requires a vector dataset"). So the filter keys on `dataset_geometry_type` rather than guessing from layer flags, the same approach the clip-mask picker already takes for polygonality (ux(#698)).

The predicate was written out three times — default selection, prefill validation, and the list itself — so it is now one `isAnalysableLayer` helper. The empty state keys off the same list and therefore starts working on a raster-only map for free.

## Tests

Two new cases, both verified load-bearing — against the old predicate:

```
× treats a raster-only map as having no analysable layers (#720)
× does not offer a raster layer alongside vector ones (#720)
Tests  2 failed | 17 passed (19)
```

## Verification

- `npx vitest run src/components/builder/__tests__/AnalysisPanel.test.tsx` → **19 passed**
- `npm run typecheck` → clean (the real type gate; `npx tsc --noEmit` is a no-op in this repo)
- `npx eslint` on both changed files → clean


ishiland:
@codex review

ishiland:
Fixed, and you were right to push on this — I checked the premise and my comment was wrong.

I claimed "rasters have no geometry_type." True in practice: all 68 raster/vrt datasets in a dev catalog store NULL.

```
 record_type    | geometry_type | count
----------------+---------------+-------
 raster_dataset |               |    67
 vrt_dataset    |               |     1
```

But that's a property of the ingest path, not an invariant — nothing enforces it, and the fixture you cited says so out loud ("geometry_type stays 'POINT' (as real raster layers report)"). So the predicate worked by accident, and any raster that acquired a geometry type would be offered for analysis again and 422 on Preview — the exact bug this PR exists to fix.

Now composed from `getLayerCapabilities`, which keys on `layer_type` + `dataset_record_type`. That also honours the convention already documented in that module for this situation (builder-audit #338 ADAPT-02/DRY-05: consume the canonical classifier rather than a local ladder) — which I should have followed the first time.

I kept the geometry_type test alongside it rather than replacing it: it is the precondition `_load_vector_dataset` enforces server-side, so a *vector* dataset missing one would 422 just the same. Neither condition covers the other.

New test uses a raster reporting `POINT`; it fails without the capability check. 20/20, lint + typecheck clean.

@codex review

ishiland:
You're right, and this reverses my last change. Verified there is no cross-validation anywhere: `layer_type` defaults to `'vector_geolens'` in `MapLayerInput`, `add_layer` persists whatever it's given, and nothing checks it against the record type.

So `getLayerCapabilities` was wrong in **both** directions, not just the one:

| case | capability-based | record_type-based |
|---|---|---|
| raster dataset, default `vector_geolens` layer_type | offered → 422 on Preview (the original bug, back) | excluded ✓ |
| vector dataset overridden to `raster_geolens` | hidden, though the endpoint accepts it | offered ✓ |

Both new tests fail against the capability version:

```
× excludes a raster carrying the default vector layer_type (#720 review)
× still offers a vector dataset rendered as raster (#720 review)
```

Now keyed on `dataset_record_type`, with `geometry_type` retained alongside it (still the server precondition — but testing it *alone* was the original bug).

Worth naming what I got wrong: I reached for "reuse the canonical classifier" as a maintainability principle without checking whether it classified the right *thing*. It answers "how do I render this", and I needed "what is this". The DRY instinct was right for a rendering question and wrong here — and it dropped the dependency again, which is the simpler code anyway.

22/22, lint + typecheck clean.

@codex review

PR #721, opened 2026-07-26
docs(readme): document the analysis toolkit in all four READMEs

ishiland:

1.5.0 ships buffer, centroid, clip, and dissolve — and the README's `## Features` section never mentioned any of it. Someone evaluating self-hosted catalogs scans that list, sees ingest/export and OGC/STAC standards, and reasonably concludes GeoLens cannot do spatial operations. On the same day the release adds them.

Adds a four-bullet `### Analysis` block between "Data ingestion and export" and "Standards and interop", in **en/de/es/fr** at the matching position so the four stay in sync.

## Scoped to what actually ships

The last bullet says the chat assistant runs **buffer and centroid** previews. That is deliberate, not an omission — clip and dissolve are excluded from the assistant's tool enum in `tools.py`, and `AnalysisPreviewRequest` rejects them on that path server-side. Claiming the assistant drives the whole toolkit would be wrong.

Likewise "capped at 500 features" and "Create dataset runs the operation over every feature as a background job" describe the real preview cap and the real materialize path, rather than implying previews are complete results.

## Verification

- All four files: `### Analysis` / `### Analyse` / `### Análisis` present, 4 bullets each, positioned identically (second `###` in the Features section, immediately before standards/interop).
- `pre-commit run --files README.md README.de.md README.es.md README.fr.md` passes.


ishiland:
@codex review

ishiland:
Both fixed in all four locales — both claims were wrong.

**STAC.** Confirmed at the source: `service_create.py:208` picks `record_type = "table" if ing.geometry_type is None else "vector_dataset"`, and `stac/router.py:81` restricts to `_STAC_RECORD_TYPES = ("raster_dataset", "vrt_dataset")`, enforced at both `:113` and `:415`. No analysis output has ever been reachable through a STAC endpoint. Claim dropped; OGC is accurate and stays.

**Dissolve preview.** Also confirmed: `AnalysisPreviewRequest.operation` is `Literal["buffer", "centroid", "clip"]` (schemas.py:887) against materialize's four (:943), and `AnalysisPanel.tsx:621` gates the button on `operation !== 'dissolve'`. Since the bullet directly above lists all four operations, "results preview on the map" read as covering dissolve. Each locale now names the three that preview and calls dissolve materialize-only.

docs-contract gate passes.

Worth saying plainly: this is a public README, these are the two sentences a GIS evaluator would actually test, and I wrote both without checking either. The STAC one in particular would have been found by someone else, in public.

@codex review

ishiland:
Fixed in all four locales. Confirmed the ceilings: `MAX_SOURCE_FEATURES = {"dissolve": 250_000, "buffer": 500_000}`, enforced at enqueue with a 422 before the job exists.

You identified the reason it matters more than a generic overclaim: that sentence is the one presenting **Create dataset** as the escape hatch from the 500-feature preview cap. A reader with a large dataset lands on exactly that line, and for them it was false in both halves — not every feature, and not even queued.

I put the actual numbers inline rather than "subject to size limits", since someone with a 2M-feature layer needs to know it won't work before they upload, not that a limit exists somewhere.

That's three false claims in this one README section — STAC exposure, dissolve preview, and now this. All three were mine, all three were checkable in under a minute, and none of them were.

docs-contract passes.

@codex review

PR #722, opened 2026-07-26
chore: remove the unscoped `ponytail:` markers and guard against new ones

ishiland:
Codex rated one of these a P1 on #718. There were 14 more.

## What

`ponytail:` appeared in 14 source comments across backend and frontend. It reads like a defined term and is not one — nothing in the repo explains it, it resolves to no issue, PR, or doc, and it is exactly the "bare, unscoped finding id that only resolves in a private tracker" AGENTS.md's inline review-comment convention rules out.

The comments themselves are the good part. Each names a real ceiling and its upgrade path:

| Site | Ceiling documented |
|---|---|
| `ingest/parquet.py` | executemany inserts, not COPY — switch at a few million rows |
| `export/parquet.py` | whole selection assembled in memory before writing |
| `router_analysis.py` | soft per-user materialize cap; TOCTOU can admit two |
| `router_export.py` | offset paging for >10k DCAT feeds |
| `runtime/staging.py` | no cross-process lock; safe because the sweep is idempotent |
| `analysis/tasks.py` | hardcoded 300s CTAS ceiling; analysis-only metrics counter |
| + 8 more | client-side guards, single-PUT progress, decorative canvas |

All of it is kept verbatim. Only the tag is gone. Where a tracked follow-up already existed the reference stays inline — the analysis-priority starvation note still points at #696.

## The guard

A pre-commit hook, scoped to the literal marker rather than a family of tags.

My first draft also matched `nit:` and `fixme-later:`, and it tripped on two existing comments — one of them `builder-audit #338 Applied-N nit:`, which is *correctly anchored* and precisely what the convention asks for, and the other a test name. A guard that punishes compliant comments is a guard someone disables, so I narrowed it.

## Verification

- Hook passes the clean tree; fails a deliberately reintroduced marker
- `ruff check` + `ruff format` clean over 345 backend files
- `npm run lint`, `npm run typecheck`, and 213 frontend tests across the touched areas pass

No behaviour changes — comments and one hook.

@codex review

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -81,6 +81,35 @@ repos:
         types: [python]
         files: '^backend/app/(modules/catalog|standards)/.*\.py$'
 
+      - id: no-unscoped-finding-markers
+        name: "AGENTS.md — no bare, unscoped finding markers in source comments"
+        description: >
+          AGENTS.md ("Inline review-comment convention") requires a comment that
+          references a review finding to carry a lookup-able anchor — a PR or
+          issue number — not a bare tag that resolves only in someone's private
+          context. 14 `ponytail:` markers reached main before anyone noticed;
+          they read as a defined term and are not one.
+
+          Write `fix(#1234): <one line>`, or just drop the tag and let the
+          sentence stand — a note about a deliberate limitation does not need a
+          prefix to be understood.
+        language: system
+        entry: >-
+          bash -c '
+          rc=0;
+          for f in "$@"; do
+            if grep -nEi "(^|[[:space:]#/*])ponytail:" "$f"; then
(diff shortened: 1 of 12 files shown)


ishiland:
Fixed. You're right, and the bypass forms you listed are the realistic ones — someone quoting the marker in prose is precisely how it would return.

Switched to `grep -Fi` on the literal. A boundary class was solving a problem I didn't have: nothing legitimate in this repo contains `ponytail:`, so there is no false-positive risk to guard against. Verified all three forms:

```
caught=1  <- // `ponytail:` backticked
caught=1  <- // (ponytail: parenthesised)
caught=1  <- // [ponytail: bracketed]
```

Clean tree still passes.

@codex review

PR #723, opened 2026-07-26
fix(builder): explain a rejected buffer distance, and stop persisting result geojson

ishiland:
Two independent medium findings from the 1.5 audit.

## 1. An out-of-range distance disabled both buttons in silence

Type `250000` metres and Preview + Create dataset both go dead with no explanation. The field carried `aria-invalid`, which marks it for a screen reader without telling anyone *what* is wrong — and `aria-invalid` with no `aria-describedby` is the a11y anti-pattern specifically: it announces "invalid" and stops.

Now a `role="alert"` message naming the ceiling **in the unit currently selected** (100,000 m / 100 km / 328,084 ft / 62.14 mi), reusing the existing `analysisTools.unit*` keys so no second set of unit strings appears. Four locales; parity gate green.

## 2. Result geojson blew the sessionStorage quota

A `show_query_result` action from `query_data` or `run_analysis` carries its whole `FeatureCollection`, up to the 500-feature preview cap. Fifty of those overflow the ~5 MB quota, the write throws `QuotaExceededError`, and the existing `catch` swallows it with a DEV-only warning.

So the failure is fully silent in production: chat history stops persisting from the first large result onward, and the user loses the whole conversation on reload with no indication why.

A `JSON.stringify` replacer drops `geojson` on the way out. Nothing rehydrates it — the geometry lives in the single-slot map overlay that a reload clears regardless — and the metadata the inline card actually renders (`row_count`, `truncated`, `operation`, `bbox`) is small and kept.

## Verification

Both tests fail against the previous code. The persistence one uses a realistic 500-polygon × 40-vertex payload, which serializes to megabytes unstripped:

```
AssertionError: expected '[{"id":"9f27d0c9-de9f-4921-a6df-d23ab…' not to contain 'FeatureCollection'
```

81 tests across both suites, lint, typecheck, and `test:i18n` all pass.

@codex review


ishiland:
Fixed. Entering exactly `0` disables both actions (`distanceValid` requires `> 0`) while the message named 0 as in range — so the one input a user is most likely to try when clearing the field got told it was fine.

Reworded to "greater than 0 and no more than {{max}} {{unit}}" across en/es/fr/de. I kept 0 invalid rather than making it valid: a zero-metre buffer returns the input geometry, which is a no-op job, not a result.

The test drove only an over-max value before; it now drives `0` explicitly and asserts the message says `greater than 0`.

18/18, i18n parity 4/4, lint + typecheck clean.

@codex review

PR #791, opened 2026-07-27
fix(analysis): enforce export capability on materialize and harden column handling

### LINKED ISSUE #761: test(analysis): pin the materialize clip-mask access check with a private-mask regression test
**Where:** `backend/app/modules/catalog/datasets/api/router_analysis.py:203-206` (the only write-path authz gate on the mask; the worker deliberately trusts enqueue per `tasks.py:428-441`)

The check is present and correct today. But preview has `test_clip_by_layer_mask_access_checked` while `test_analysis_materialize.py` tests private-*source* and the upload permission only — never a private *mask*. The router branch sits between two size gates that have been rewritten twice in two weeks (#701, #693), and the Rule 1 pre-commit grep is satisfied by the source-dataset call in the same file, so a regression here ships with every existing check green.

**Smallest test:** in `TestMaterializeEndpoint`, seed a private polygon dataset owned by another user, POST a clip materialize with that `mask_dataset_id`, assert 404.
---

### LINKED ISSUE #763: bug(analysis): attribute columns with non-identifier names are silently dropped from analysis outputs
**Where:** `backend/app/processing/analysis/tasks.py:273` (regex at `:45`)

`_list_carry_columns` keeps only columns matching `^[a-zA-Z_][a-zA-Z0-9_]*$` — no error, no warning, no log. Verified that such names reach physical tables: GDAL 3.13 with the ingest path's exact `-lco` set emits `"größe"`, `"Área"`, `"2020_pop"`, `"my field"`, `"名前"` (LAUNDER only lowercases and swaps `-`/`#`). The repo's own `unicode_attrs.geojson` fixture is affected. A buffered census/European dataset silently loses attributes; nothing in the job row or API response says so, and no test asserts attribute survival.

**Fix:** the names come from `information_schema`, not the request — quote-escape (`'"' + c.replace('"','""') + '"'`) and stop filtering. If the filter stays, record dropped columns via `_append_job_warning` (`tasks_common.py:266`), the way `rename_reserved_columns` already reports Socrata renames.
---

### LINKED ISSUE #766: bug(analysis): dissolve by a json column passes enqueue and fails opaquely after the queue wait
**Where:** `backend/app/modules/catalog/datasets/api/router_analysis.py:214-231`; `tasks.py:85-94`, `:299`

`by_field` is validated by name against `column_info` but never by type. GDAL maps nested GeoJSON objects to PostgreSQL `json` columns, and PG has no equality operator for `json` (verified against the running PG 18: `could not identify an equality operator for type json`). The resulting 42883 `ProgrammingError` is not matched by `_user_error_message` (only timeouts / DataError / InternalError), so the user burns the queue wait and their per-user job slot for "The analysis failed due to a database error" — the same class of foreseeable failure the adjacent `source_count` guard exists to pre-empt.

**Fix:** reject non-groupable `col["type"]` values at enqueue with a specific message; additionally classify SQLSTATE 42xxx in `_user_error_message` ("that column can't be used for this operation") to cover future schema-shaped failures.
---

ishiland:
## What

Backend batch for the v1.5.1 milestone (analysis materialize path):

- **Materialize now requires the `export` capability in addition to `upload`.** The output dataset is a caller-owned copy of the source's attributes — the same outcome the download endpoints gate on `export` — so the two paths now agree under a customized role matrix. Default-matrix installs are unaffected (editor/admin hold both capabilities). OpenAPI + generated SDK descriptions regenerated.
- **#763 — analysis outputs keep every attribute column.** `_list_carry_columns` filtered names against an identifier regex, silently dropping columns GDAL legitimately produces (`Área`, `2020_pop`, `my field`). Identifiers are now rendered with quote-escaping (names come from `information_schema`, never the request).
- **#766 — dissolve `by_field` is validated by type at enqueue.** PG has no equality operator for `json`/`xml`, so those columns burned the queue wait + the per-user job slot on an opaque 42883. Also classifies SQLSTATE 42xxx in `_user_error_message` as a column problem instead of a generic database error.
- **#761 — the materialize clip-mask visibility check is now pinned by a test** (private mask, non-owner caller, 404), mirroring the preview suite.

## Impact

- API: materialize 403s for roles granted `upload` but not `export` (non-default matrices only). Documented in CHANGELOG under Changed.
- No schema changes.

## Verification

- Full backend suite: **5878 passed, 75 skipped** (`pytest -n 4`)
- `make openapi` + `make sdks` regenerated; snapshots committed

Closes #761. Closes #763. Closes #766.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,24 @@ and releases use semantic versioning.
   was a dead end whose hint named an invisible button. The clip mask picker
   also says "No polygon layers on this map" instead of showing a lone "None"
   entry. (#779)
+- **Analysis outputs now keep every attribute column.** Columns whose names
+  are not identifier-shaped (`Área`, `2020_pop`, `my field`) were silently
+  dropped from buffer, centroid, and clip outputs; they are now carried
+  through intact. (#763)
+- **Dissolving by a non-groupable column is refused up front.** A `json`
+  column (how nested GeoJSON attributes ingest) used to enqueue, wait out
+  the queue, and fail with a generic database error; the request is now
+  rejected immediately with the column named, and schema-shaped failures
+  that do reach the worker report "a column can't be used for this
+  operation" instead of "database error". (#766)
+
+### Changed
+
+- **Creating a dataset from an analysis now requires the `export`
+  capability in addition to `upload`,** matching the download endpoints —
+  the result is a caller-owned copy of the source's attributes, so the two
+  paths now agree under a customized role matrix. Installations using the
+  default role matrix are unaffected (editor and admin hold both).
 
 ## [1.5.0] - 2026-07-26
 
--- a/backend/app/modules/catalog/datasets/api/router_analysis.py
+++ b/backend/app/modules/catalog/datasets/api/router_analysis.py
@@ -44,6 +44,12 @@ router = APIRouter(
 
 _SAFE_COLUMN_RE = re.compile(r"^[a-zA-Z_][a-zA-Z0-9_]*$")
 
+# fix(#766): PostgreSQL has no equality operator for these types, so a
+# dissolve GROUP BY on such a column fails the CTAS with an opaque 42883
+# after the queue wait. GDAL maps nested GeoJSON objects to `json`, so
+# real uploads hit this. Rejected at enqueue with the column named.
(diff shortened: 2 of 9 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #792, opened 2026-07-27
fix(builder): keep mid-save edits dirty instead of silently discarding them

### LINKED ISSUE #756: bug(builder): layer edits made while a save is in flight are silently discarded
**Where:** `frontend/src/components/builder/hooks/use-builder-save.ts:576-589`, `:684-686`, `:554-559`

`handleSave` destructures `localLayers` from state once at call time, awaits the network, then unconditionally clears `hasUnsavedChanges` and resets the baseline from post-await state. Any edit made during the await is (1) absorbed into the baseline so no future diff emits it, (2) overwritten in the UI by the `queryKeys.maps.detail` invalidation resync (`use-builder-layers.ts:239-249`, gated only on `!hasUnsavedChanges`), and (3) invisible to `useUnsavedGuard`. Nothing blocks editing during a save — only the Save button disables (`MapTitleBar.tsx:168`); the stack and editor flyout stay interactive. Same snapshot risk for `localName`, `basemapConfig`, `terrainConfig`.

**Fix:** save-generation counter (or identity compare) — only clear the dirty flag when current state still matches what was sent, and set the baseline to the *sent* payload. Alternatively disable the stack/editor while `isSaving` (smaller change, closes it at the UI boundary).
---

ishiland:
## What

Fixes the P1 save-path data-loss window from the 2026-07-27 builder audit: `handleSave` snapshots state at call time, awaits the network, then unconditionally cleared `hasUnsavedChanges`. Any edit made during the await was (1) absorbed into the save baseline so no future diff emitted it, (2) overwritten on screen by the `queryKeys.maps.detail` invalidation resync (gated only on `!hasUnsavedChanges`), and (3) invisible to the unsaved-navigation guard.

## How

- The save baseline is now explicitly the **sent** payload (it already was, structurally — now guaranteed against the flag).
- The dirty flag is only cleared when none of the snapshotted fields (`localLayers`, `groupMeta`, name/description/legend/notes, basemap, basemap/terrain config) changed identity during the round-trip; a `latestStateRef` tracks the last rendered state for the post-await comparison.
- Applies to both the normal save path and the #430 full-resync fallback.

If an edit did land mid-save, the map stays dirty: the mid-save edit remains diffable against the sent baseline, the resync guard keeps it on screen, and Save re-enables.

## Verification

- `npx vitest run src/components/builder/hooks/__tests__/use-builder-save.test.ts src/components/builder/hooks/__tests__/use-builder-layers-save-integration.test.ts` — 70 passed (2 new: mid-save edit keeps the flag; clean save still clears it)
- `npm run lint` + `npm run typecheck` — clean

Closes #756.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ and releases use semantic versioning.
 
 ### Fixed
 
+- **Edits made while a save is in flight are no longer silently discarded.**
+  Changing a layer, name, basemap, plugin, or any other map property during
+  the save's network round-trip used to be absorbed into the saved baseline,
+  overwritten on screen by the post-save refresh, and dropped from the
+  unsaved-changes guard. The map now stays marked unsaved until those edits
+  are actually saved. (#756)
 - **The analysis completion notice no longer covers the panel's own "Add to
   map" button.** In the builder it now appears top-center, clear of the right
   rail; elsewhere it keeps the usual corner. (#725)
--- a/frontend/src/components/builder/hooks/__tests__/use-builder-save.test.ts
+++ b/frontend/src/components/builder/hooks/__tests__/use-builder-save.test.ts
@@ -482,6 +482,38 @@ describe('useBuilderSave', () => {
     expect(setHasUnsavedChanges).not.toHaveBeenCalledWith(false);
   });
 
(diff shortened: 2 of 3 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #794, opened 2026-07-27
fix(builder): restore keyboard reorder on stack rows and gate the job poll on auth

### LINKED ISSUE #759: bug(a11y): dnd-kit keyboard reorder is dead on every stack row, and the row-level reorder mode is silent
**Where:** `frontend/src/components/builder/row-chrome.tsx:112`; `StackRow.tsx:143-193`; `FolderGroupRow.tsx:177`; `BasemapGroupRow.tsx:142`; `MapBuilderPage.tsx:1628`

Two halves:

1. `DragGripButton` spreads dnd-kit `listeners` and then unconditionally sets `onKeyDown={onKeyDown}` **after** the spread. Later JSX props win even when `undefined`, so the KeyboardSensor activator is destroyed for every caller (verified with a render probe — 0 invocations). Folder groups and the basemap group pass no custom `onKeyDown` and have no move-up/down menu items, so they are pointer-only — a WCAG 2.1.1 failure. This is the second occurrence of the props-after-listeners-spread trap in this codebase.
2. StackRow's fallback reorder mode (`keyboardReorderActive`) drives no `aria-pressed` and never calls `announce()`, while the pointer path announces pickup/position/drop. Space-arm + arrow-move is completely silent, and the shortcuts sheet advertises both bindings.

**Fix:** spread `listeners` after explicit handlers (or compose), reflect the mode in `aria-pressed`, and call `announce()` from the reorder handler (the live region and all four locales' keys already exist). Consider a grep/lint hook for JSX props set after a `listeners` spread — it has now bitten twice.
---

### LINKED ISSUE #762: bug(analysis): a stale persisted analysis job logs anonymous visitors out on public pages
**Where:** `frontend/src/components/analysis/AnalysisJobWatcher.tsx` (mounted in RootLayout, `App.tsx:63`); `hooks/use-ingest.ts:15-28`

`useJobStatus` has no auth-token guard — its sibling `useDatasetJobStatus` added `enabled: !!datasetId && hasToken` (`:39-44`) for exactly the "anonymous public pages should not hit protected job endpoints" reason. `analysis-job-store` is persist-backed, so a stale tracked job makes an anonymous visitor on the login page or a public viewer route fire an authenticated-only poll → 401 → `logout()`.

**Fix:** gate the watcher's query on `hasToken` like its sibling, and clear the persisted job on 401/404 instead of looping.
---

ishiland:
## What

Two audit fixes on independent surfaces:

- **#759 — keyboard reorder was dead on every stack row (WCAG 2.1.1).** `DragGripButton` spread the dnd-kit `listeners` and then set `onKeyDown={onKeyDown}` unconditionally after the spread — a later JSX prop wins even when `undefined`, so the KeyboardSensor activator was destroyed for every caller (verified in the audit by render probe: 0 invocations). This is the second occurrence of the props-after-listeners-spread trap (first: #525).
  - The grip now **composes**: the row's own handler runs first, and the dnd-kit activator only sees keys the row didn't claim. Folder groups and the basemap group (which pass no custom handler) get native dnd-kit keyboard drag back; StackRow's fallback reorder mode keeps claiming Space/arrows/Escape, so there is no double handling.
  - The fallback mode now drives `aria-pressed` on the grip and announces arm / each move / drop / cancel through the builder's shared live region (three new `a11y.reorder*` keys in all four locales; pickup and cancel reuse the existing drag keys). `aria-pressed` is only overridden when a row actually drives the mode, so dnd-kit's own managed value is preserved on group rows.
  - The fix lives in the one shared grip component, which is also where both occurrences of the trap lived — there is no remaining listeners-spread call site to hook a grep on.
- **#762 — a stale persisted analysis job logged anonymous visitors out.** `AnalysisJobWatcher` mounts in RootLayout over a persist-backed job id; `useJobStatus` had no token gate, so an anonymous session on a public page fired an authenticated-only poll whose 401 ended the session. The hook now gates on `hasToken` exactly like its sibling `useDatasetJobStatus`; the watcher already clears the tracked job on 401/403/404 once a real response arrives.

## Verification

- `npx vitest run src/components/builder/__tests__/ src/components/import/hooks/__tests__/use-ingest.test.ts` — 1348 passed (new: FolderGroupRow activator regression test, 3 StackRow reorder-mode tests, anonymous-poll gate test)
- `npm run lint`, `npm run typecheck`, `npm run test:i18n` — clean

Closes #759. Closes #762.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and releases use semantic versioning.
   destroyed the keyboard drag activator on all rows — folder groups and the
   basemap group, which have no fallback, were pointer-only. The handlers are
   now composed, the row-level reorder mode exposes its armed state via
-  `aria-pressed`, and arming, each move, dropping, and cancelling are
+  `aria-pressed`, and arming, each successful move, and finishing are
   announced to screen readers like the pointer drag path already was. (#759)
 - **A stale tracked analysis job no longer logs anonymous visitors out.** The
   job-status poll is now gated on having an auth token, so a persisted job id
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,17 @@ and releases use semantic versioning.
 
 ### Fixed
 
+- **Keyboard reorder works again on every stack row.** The shared drag grip
+  set a key handler after spreading the dnd-kit listeners, which silently
+  destroyed the keyboard drag activator on all rows — folder groups and the
+  basemap group, which have no fallback, were pointer-only. The handlers are
+  now composed, the row-level reorder mode exposes its armed state via
+  `aria-pressed`, and arming, each move, dropping, and cancelling are
+  announced to screen readers like the pointer drag path already was. (#759)
+- **A stale tracked analysis job no longer logs anonymous visitors out.** The
+  job-status poll is now gated on having an auth token, so a persisted job id
+  can't make a public page fire an authenticated request whose 401 ended the
+  session. (#762)
 - **Edits made while a save is in flight are no longer silently discarded.**
   Changing a layer, name, basemap, plugin, or any other map property during
   the save's network round-trip used to be absorbed into the saved baseline,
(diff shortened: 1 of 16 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #793, opened 2026-07-27
fix(analysis): make the panel's form, preview, and job state survive its lifecycle

ishiland:
## What

The AnalysisPanel state-lifecycle batch from the 2026-07-27 builder/analysis audit — four issues, one surface:

- **#757 — the form survives the panel's conditional unmount.** BuilderRail conditionally mounts the panel, so a click on Notes/History/Ask AI, Escape, or crossing the 800px breakpoint destroyed every field, including a hand-drawn clip mask. A new session-scoped `analysis-form-store` (mirroring `analysis-job-store`'s singular per-map design, deliberately not persisted) snapshots the form on unmount and restores it on mount for the same map. A remembered form whose source layer left the map is ignored wholesale; a chat-handoff prefill takes precedence.
- **#758 — stale/superseded previews.** [live] A preview overlay + "Result · N features" badge survived every input change and two materializes. Every input change now clears the overlay and bumps a request sequence; preview (and materialize) responses stamped with an older sequence are dropped — including the in-flight variant that used to draw *and fitBounds-zoom to* the previous layer's result.
- **#760 — job-state amnesia.** `jobId` now rehydrates from the persisted job store when the tracked job belongs to this map, so a reopened/reloaded panel shows the status line and completion actions. When the one-job-per-user cap disables Create for any other reason, the panel says so ("Another analysis is still running…"). The mobile rail buttons get the desktop rail's #682 running-job spinner + `aria-busy`.
- **#764 — post-run state never reset.** [live] After a successful run, the completed state and the old name survived layer/operation switches; one click created an identically-named dataset from different parameters. Run state + name now reset on any input change, and "Add to map" names the dataset it will add.

Two new translation keys (`analysisTools.anotherJobRunning`, `analysisTools.addToMapNamed`) added to all four locales.

Not included (called out in the audit as "ideally"): a catalog-wide duplicate-display-name lookup at create time — names aren't unique by design and the root cause here was the stale state, which this removes.

## Verification

- `npx vitest run src/components/builder/__tests__/AnalysisPanel.test.tsx` — 31 passed (8 new lifecycle tests)
- `npm run lint`, `npm run typecheck`, `npm run test:i18n` — clean
- `npm run test:coverage` — full suite green (Statements 71.39%)

Closes #757. Closes #758. Closes #760. Closes #764.

PR #795, opened 2026-07-27
fix(analysis): polish batch — toast placement, legend cap, error prose, viewer dead-end

### LINKED ISSUE #725: bug(builder): the analysis completion toast covers the panel's own "Add to map" button
Found while filming the 1.5 analysis demo, and it cost a retake: the take clicked the button, nothing happened, and the layer never landed.

## What happens

After a materialize job completes, `AnalysisPanel` renders an **Add to map** button under the job status line. `AnalysisJobWatcher` raises its completion toast at the same moment. The toast lands on top of the button, so the button is visible, enabled, and unclickable.

Measured on a 1512x945 viewport against the local stack:

```
panel Add to map box:  { x: 1207, y: 899, width: 291, height: 32 }
completion toast box:  { x: 1132, y: 848, width: 356, height:  73 }
OVERLAP: true
elementFromPoint at the button centre -> <LI> inside [data-sonner-toast]
plain click on the panel button       -> locator.click: Timeout 4000ms exceeded
```

The Playwright timeout is the interesting line. Actionability never settles because the element is obscured, which is exactly what a real pointer runs into.

## Why the two collide

- `main.tsx:77` mounts `<Toaster theme={resolvedTheme} richColors closeButton />` with no `position`, so sonner defaults to bottom-right.
- `AnalysisJobWatcher` raises the completion toast with `duration: Infinity`, deliberately, so a job that lands after the user's attention moved on is not missed. It stays until dismissed.
- `AnalysisPanel`'s creation block is the last thing in the right-hand rail panel, so **Add to map** sits in the bottom-right corner too.

The overlap is therefore permanent for as long as the toast is up, not a momentary animation race.

## Severity

Low, because it is not a dead end. The toast carries its own **Add to map** action wired to the same `analysisAddToMap.current`, and that is the path a user is most likely to take. But a control that renders enabled and cannot be clicked is still wrong, and the panel button is the one a user returns to after dismissing the toast for some other reason.

## Options

1. Give the `Toaster` `position="bottom-center"`. One line, but it moves every toast in the app.
2. Move the panel's **Add to map** above the job status line, or into the button row at the top of the creation block.
3. Drop the panel button entirely, now that the toast action covers the case. Fewest moving parts, but it loses the affordance once the toast is dismissed.

(2) looks like the smallest correct change.

## Repro

1. Open a map with a vector layer, open the Analysis panel, run any materialize (buffer at 50 m on a small layer is fastest).
2. Wait for the job to complete.
3. Try to click the panel's **Add to map** without dismissing the toast first.

### LINKED ISSUE #731: bug(viewer): the layer legend overflows the map container and collides with the Map data button and basemap toggle
## What happens

On the saved-map viewer, the Layers legend panel, the "Map data" button, and the basemap-toggle thumbnail all pile up in the bottom-left corner. The legend runs down behind the other two controls, which paint over it and occlude legend rows (observed live on the demo's "Restless Earth" map: the "Major cities (by population)" row is half-hidden behind the Map data button and the basemap thumbnail).

## Repro conditions

Window height × legend length — not auth state. A short browser window plus a map whose legend is tall (3+ layers with size/color ramps, e.g. the seeded Restless Earth showcase) reproduces it for anonymous *and* logged-in users. It only *looks* anonymous-specific because the authenticated builder uses a different sidebar layout; the viewer routes are where these three widgets coexist. The `/m/` embed page has the identical structure and is affected too.

Observed on the live demo (v1.4.13); the three components and both page files are unchanged on current main, so 1.5.0 ships with it as-is.

## Root cause

All three widgets are absolutely positioned inside the map container, but the legend's height cap is measured against the **viewport**, not its container:

- Legend — `frontend/src/components/viewer/LayerLegend.tsx:263`: `absolute start-3 top-14 z-10 w-64 max-h-[calc(100vh-5rem)] overflow-y-auto`
- Map data button — `frontend/src/components/viewer/AccessibleMapDataPanel.tsx:70`: `absolute bottom-20 left-3 z-30`
- Basemap toggle — `frontend/src/pages/PublicMapViewerPage.tsx:188` (and `PublicViewerPage.tsx:199`): `absolute bottom-8 start-3 z-10`

The map container is shorter than the viewport by the app header, title bar, and footer, so `max-h-[calc(100vh-5rem)]` lets the legend grow taller than the box it sits in. It then extends through the bottom-left corner where the other two controls are anchored, and z-order (button z-30 vs legend z-10; the toggle renders later in the DOM) paints them over the legend content.

## Fix shape

One line in `LayerLegend.tsx:263`: replace the viewport-relative cap with a container-relative one that reserves clearance for the bottom-left cluster, e.g. `max-h-[calc(100%-11rem)]` (the panel is absolutely positioned, so `%` resolves against the map container; 3.5rem top offset + ~7.5rem clears the Map data button at `bottom-20` plus its height).

Caveat for the fix: `LayerLegend` is also consumed by `frontend/src/components/map-plugins/builtin/LegendPlugin.tsx` — verify that consumer's container is a positioned context (so `100%` resolves sensibly there) before shipping the class change.

## Triage

Pre-existing (≤ 1.4.13), cosmetic but on the anonymous showcase path. Not a 1.5.0 blocker — candidate for the same polish batch as #725 and #727.

### LINKED ISSUE #774: bug(frontend): error-map drops the backend prose for four analysis 4xx responses
**Where:** `frontend/src/lib/error-map.ts:153-179`; backend literals at `router_analysis.py:81`, `:105`, `:222`, `:273`

#718 mapped the two size-limit messages; these four fall through to generic fallbacks:
- `An analysis job is already running; wait for it to finish` (429) → "Too many requests. Wait a moment and try again." — wrong advice; the correct advice is that the user's own job is still running. Reachable routinely: the persisted job store has no cross-tab sync, so a second tab passes `canSave` and earns the 429.
- `Analysis requires a vector dataset`, `mask_dataset_id must reference a polygon dataset`, `Unknown dissolve column: 'x'` (422) → "The submitted values are invalid."

**Fix:** add the four literals to `EXACT_ERROR_KEYS` with dedicated strings in all four locales; consider a storage-event sync for `analysis-job-store` so the second tab disables Create instead of erroring. Related: the preview shares the `geolens:ai-sql:{user_id}` advisory lock with AI chat (`service_analysis.py:190-195`), so chat activity can 429 the panel — either split the namespace or name the other feature in the message.
---

### LINKED ISSUE #779: ux(analysis): a viewer who picks Dissolve gets a form with no actions and a hint naming a hidden button
**Where:** `frontend/src/components/builder/AnalysisPanel.tsx:473-480`, `:687-693`, `:702`

Preview is hidden for dissolve (no server preview path) and the whole materialize block is hidden without `can('upload')`. A viewer selecting Dissolve sees layer/operation/group-by selects and nothing to press — while the helper text reads "…Run it with Create dataset", a control that is not rendered for them.

**Fix:** when `!canCreateDataset`, hide Dissolve from the operation list or swap the hint for one explaining dissolve needs upload permission. Related empty-state gap in the same panel: with no analysable layers it renders a bare sentence with no CTA (`:389-397`) — every sibling empty state offers an action; wire an "Add data" button to the stack header's `onAddDataClick`. Also: the clip mask-layer dropdown shows nothing when no polygon layers exist — add a "no polygon layers on this map" empty state.
---

ishiland:
## What

The four v1.5.1 polish issues:

- **#725 — the completion toast covered the panel's "Add to map" button.** The toast is deliberately `duration: Infinity`, so the default bottom-right placement parked it permanently on the rail panel's own button (measured overlap in the issue; `elementFromPoint` resolved to the toast). Reordering inside the panel stays viewport-height-dependent, so instead the completion toast renders **top-center while a builder for the job's map is mounted** (sonner 2 per-toast `position`) and keeps the default corner everywhere else.
- **#731 — viewer legend ran behind the bottom-left controls.** The legend capped its height against the *viewport* while being absolutely positioned in the *map container*, colliding with the Map data button (bottom-20) and basemap toggle (bottom-8). The cap is now `max-h-[calc(100%-11rem)]` (container-relative: 3.5rem top offset + clearance over the bottom-left stack); the legend already scrolls internally. Embed `/m/` pages share the component and inherit the fix.
- **#774 — four analysis 4xx literals fell through to generic fallbacks.** The one-job-per-user 429 rendered as "Too many requests" (wrong advice — it's the user's own job); the vector-required / polygon-mask / unknown-dissolve-column 422s collapsed to "values are invalid". All four are mapped with translated strings in the four locales; the unknown-column message interpolates the column name. (The issue's "consider" items — cross-tab job-store sync and splitting the preview's advisory-lock namespace from AI chat — are not taken here.)
- **#779 — viewer + Dissolve was a dead end.** Dissolve is materialize-only and the creation block is hidden without `upload`, so a viewer picking it saw a form with no actions and a hint naming an invisible button. Dissolve is no longer offered without the capability, and the clip mask picker gets a "No polygon layers on this map" empty state. (The panel-level "Add data" CTA suggested as related is left out — it needs new prop threading through BuilderRail and is not part of the filed dead-end.)

## Verification

- `npx vitest run src/lib/__tests__/error-map.test.ts src/components/builder/__tests__/AnalysisPanel.test.tsx src/components/analysis/__tests__/AnalysisJobWatcher.test.tsx` — green (new: 4-literal mapping test, viewer-dissolve test)
- `npm run lint`, `npm run typecheck`, `npm run test:i18n` — clean

Note: this PR touches `AnalysisPanel.tsx`/`AnalysisJobWatcher.tsx`, which #793 also modifies — whichever merges second gets a trivial rebase.

Closes #725. Closes #731. Closes #774. Closes #779.


ishiland:
@codex review

PR #797, opened 2026-07-27
docs(changelog): move the 1.5.1 export-capability fix under Security with its advisory link

ishiland:
GHSA-8gvc-94m7-m462 is published, so the changelog entry for the 1.5.1 fix can now say what it was: the materialize path bypassed the `export` capability. This moves the bullet from **Changed** to a new **Security** section at the top of the 1.5.1 entry and links the advisory (inline URL, so the awk-extracted release body renders it).

The v1.5.1 GitHub Release body has been re-extracted from this version of the entry and already updated to match.

No code changes. Verification: `awk '/^## \[1.5.1\]/{f=1;next} /^## \[/{if(f)exit} f{print}' CHANGELOG.md` renders the Security section first (83 lines).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,19 @@ and releases use semantic versioning.
 
 ## [1.5.1] - 2026-07-27
 
+### Security
+
+- **Creating a dataset from an analysis now requires the `export`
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review

Issue #798, opened 2026-07-27
ops(backup): warn at startup when offsite upload is disabled


ishiland:
From the 2026-07-26 demo-VM operations audit. **Re-scoped 2026-07-31:** most of what this issue asked for already exists. What is left is one line of shell and one thing to check on the demo VM.

**Size: S.**

## What already shipped

Both remedies the original body asked for are in the tree, and both predate or shortly follow the filing:

- **A configurable offsite copy target.** `upload_to_s3()` (`scripts/backup-entrypoint.sh:250`) uploads the verified dump, and the paired staging archive, to `daily/` and on Sundays `weekly/` (`:145-161`). It signs with AWS SigV4, so it works against AWS S3, Cloudflare R2, and MinIO. A failed upload fails the cycle after local retention pruning has run, so a transient S3 outage cannot let local dumps accumulate. It also refuses loudly if `BACKUP_S3_ENABLED=true` is set without the bucket and credentials (`:254-259`), rather than promising an offsite copy it cannot make.
- **The RUNBOOK section.** `RUNBOOK.md:75-105` documents the offsite upload, opens with an explicit single-disk warning (`:77-82`: "Losing that disk loses the data and every backup of it in one event"), and lists the required `.env` keys.

So the risk is documented and the mechanism exists. The original framing, that a self-hoster has no supported path to get dumps off the host, is no longer true.

## What is actually left

**The default is silent.** `BACKUP_S3_ENABLED` defaults to `false` (`scripts/backup-entrypoint.sh:45`). On startup the service logs `S3 upload: false` (`:413`) as a plain info line among the schedule and retention lines. An operator who never opened RUNBOOK sees nothing that reads as a warning, and the deployment that most needs the warning is exactly the one that never configured anything.

Add a startup line, in the same block at `:410-413`, that says plainly what the disabled default means: backups are being written to the same host, and usually the same physical disk, as the database, so a single disk loss takes the data and every backup of it together. It should read as a warning rather than as configuration output, and it should point at the RUNBOOK section for how to turn the offsite copy on. It costs nothing when offsite upload is enabled, because it does not print then.

Keep it to the startup path, not per cycle. A nightly warning in the logs of a deliberately local-only install is noise that trains operators to ignore the log.

## Checklist

- [ ] `scripts/backup-entrypoint.sh` warns at startup when `BACKUP_S3_ENABLED` is not `true`, naming the single-disk exposure and pointing at RUNBOOK § "Offsite (S3) upload".
[...]

Issue #711, opened 2026-07-26
feat(backup): opt-in PITR profile — physical backup + WAL archiving for sub-24h RPO


ishiland:
> **Status: blocked on demand confirmation.** This is a product decision, not ready work. See the Prerequisite section at the bottom before picking it up.

Split out of the backup review that produced #710. That PR fixed the integrity gate and made the current guarantee honest; this issue is the guarantee itself.

## The gap

The bundled backup is a nightly `pg_dump`, so the default RPO is **up to 24 hours**. That is now stated in RUNBOOK § 1 rather than left to be inferred, but stating it does not shrink it.

It cannot be shrunk with this tooling. `pg_dump` produces a *logical* dump, which does not contain the information WAL replay needs — so no schedule change buys point-in-time recovery. Sub-24h RPO requires physical backup plus continuous WAL archiving. There is no middle setting.

Managed-Postgres deployments already have the answer: provider-native PITR, which RUNBOOK correctly points at. **This issue is specifically about bundled-Postgres self-hosters.**

## Why this is opt-in and not the new default

The failure modes are not symmetric.

- `pg_dump` fails → last night's backup is missing. Visible, recoverable, nothing stops.
- `archive_command` fails (unreachable destination, full disk, bad permissions) → PostgreSQL retains WAL until `pg_wal` exhausts the filesystem and **the database stops accepting writes**.

Making continuous archiving the default trades a bounded data-loss risk for an unbounded availability risk, on single-host deployments that often have no DBA and no alerting. That is the wrong default for this product. It should be a deliberate choice by an operator who wants it and can monitor it.

## Scope

- A compose profile (e.g. `--profile pitr`) wiring `wal_level = replica`, `archive_mode = on`, and a supported archive command. Note `archive_mode` requires a **database restart**, not a reload.
- A durable archive destination — dedicated volume and/or S3, reusing the existing `BACKUP_S3_*` configuration rather than inventing a second S3 surface.
[...]

PR #800, opened 2026-07-27
fix(backup): go unhealthy when backups stop succeeding, not just when the loop dies

### LINKED ISSUE #712: ops(backup): nothing detects that backups stopped succeeding — add a freshness signal
Split out of the backup review (#710, #711). This one is arguably higher priority than either.

## Problem

The `backup` service healthcheck is:

```
pgrep -f backup-entrypoint || pgrep -f sleep
```

That is a **liveness** check. The container reports `healthy` while every single `pg_dump` fails, because the entrypoint process is alive and sleeping until the next cycle. RUNBOOK § 4 is honest about this ("confirms the entrypoint process is running"), but honesty is not detection.

There is no freshness signal anywhere: no metric, no alert, no age check on the newest artifact. Discovering that backups stopped requires a human to run `docker compose logs -f backup` and read them.

#710 makes this quieter rather than louder. A dump that now fails end-to-end verification is discarded, so that failure mode produces **no file** instead of a bad one — equally invisible. We have made the dump trustworthy while nothing checks that it exists.

The classic backup disaster is not a 24-hour RPO. It is "backups stopped six months ago and nobody noticed."

## Suggested work

Either or both:

1. **Freshness metric + reference alert.** Export the age of the newest artifact in `/backups/daily` alongside the existing metrics (the #466 monitoring reference config is the natural home), and ship an alert rule for `age > 1.5 × schedule interval`. This is the signal an operator with monitoring actually wants.

2. **Make the healthcheck mean something.** Go `unhealthy` when the newest dump is older than ~1.5× the configured `BACKUP_SCHEDULE` interval, so `docker compose ps` tells the truth for operators with no monitoring stack. Needs care: the interval is a cron expression, and a fresh install has no dump yet, so the check must tolerate a grace window on first start (the existing `start_period` is 30s, which is not enough for this).

Also worth surfacing the last-cycle outcome rather than just file age — a cycle can fail at the S3 upload while the local dump lands fine (`run_backup` already distinguishes these and returns non-zero).

## Notes

- The S3 offload path does not need extra verification: `aws s3 cp` checksums on upload and the exit code is already checked.
- Keep this independent of #711 (opt-in PITR). This applies to the default nightly-dump configuration that every install runs today, which is exactly why it should not wait behind a feature that most deployments will never enable.

ishiland:
## Problem (#712)

The `backup` service healthcheck was pure liveness — `pgrep -f backup-entrypoint || pgrep -f sleep` — so the container reported **healthy** while every single `pg_dump` failed. Nothing anywhere detected that backups stopped succeeding.

## Design: success marker + max-age freshness healthcheck

- **Marker.** `scripts/backup-entrypoint.sh` now touches `/backups/.last-success` at the very end of `run_backup()`. Every failure path (pg_dump failure, `pg_restore -f /dev/null` verification failure — that end-to-end verify from #710 is untouched — and S3 upload failure when `BACKUP_S3_ENABLED=true`) returns before the touch, so the marker's mtime records the last **fully successful** cycle. It lives in the `backup_data` volume, so it survives container restarts and is invisible to the `*.dump` / `staging-*.tar.gz` retention globs.
- **Healthcheck.** Both compose files replace the pgrep probe with: `find /backups/.last-success -mmin -${BACKUP_MAX_AGE_MINUTES:-1560} | grep -q . || exit 1` — healthy iff the marker exists and is younger than the knob. Deliberately parse-safe `${VAR:-default}` interpolation (never `:?`, per INST-01).
- **Knob.** `BACKUP_MAX_AGE_MINUTES`, default **1560 min = 26 h** — the default daily `BACKUP_SCHEDULE` plus slack. Documented in `.env.example` (passes `scripts/check_env_doc_drift.py`) and RUNBOOK §4 with tuning guidance (~1.5× the schedule interval for non-daily schedules). Parsing the cron expression to derive the interval was considered and rejected as over-engineering.

## First-install behavior

No marker exists on a fresh volume, so `start_period` grows 30 s → **10 m**: the container shows `starting` until the initial on-startup backup succeeds (seconds on a fresh DB — the 5 s `start_interval` keeps the installer's wait-for-healthy budget comfortably cleared), and turns `unhealthy` ~90 s after `start_period` if no backup ever succeeds. A very large DB whose initial dump outlasts 10 m shows a transient `unhealthy` that clears on the first probe after the cycle completes. All of this is documented honestly in RUNBOOK §4.

## Scope notes

- `docker-compose.prod.yml` is included beyond the original file list because its header mandates keeping the `backup` block in sync with `docker-compose.yml`, and prod installs are the ones that need this signal most. Compose file and image version travel together (`GEOLENS_VERSION`), so the new probe never meets an old entrypoint.
- **Alerting follow-up (not wired here):** the reference `infra/monitoring/alerts.yml` has no backup-freshness rule; exporting newest-artifact age as a metric + a reference alert (per the issue's option 1) is a separate follow-up. This PR is the no-monitoring-stack signal (`docker compose ps` tells the truth).
- **#711 (opt-in PITR) and #798 (single-disk DR) remain separate** — this applies to the default nightly-dump configuration every install runs today.

## Validation

- `bash -n scripts/backup-entrypoint.sh`
- `cp .env.example .env` (local only, gitignored), then `docker compose config -q` ✔, `docker compose --profile cloud-dev config -q` ✔, `docker compose -f docker-compose.prod.yml config -q` ✔, `docker compose -f docker-compose.prod.yml --profile cloud-dev config -q` ✔
- Rendered probe verified via `docker compose config backup`: default interpolates to `-mmin -1560`; `BACKUP_MAX_AGE_MINUTES=99` override renders `-mmin -99`
- Probe semantics tested standalone: missing marker → exit 1, fresh marker → exit 0, stale marker → exit 1
- `python3 scripts/check_env_doc_drift.py` → OK (130 documented keys aligned)

Closes #712


ishiland:
@codex review

Issue #691, opened 2026-07-25
Analysis materialize: hold the per-user job slot on a heartbeat lease, not job status alone


ishiland:
## Problem

The per-user materialize cap in `router_analysis.py` blocks while the user has any `pending` or `running` analysis job. That is correct — it never lets a second expensive CTAS through — but it has one bad failure mode: **a worker that dies mid-job holds the user's slot until the platform job timeout fails the row** (`JOB_TIMEOUT_SECONDS`, 60 min). The Analysis panel's Create button mirrors the same rule, so it stays disabled for that whole period.

Since #700, a routine worker deploy no longer strands the slot at all, so the residual window this issue closes is a hard-killed worker (SIGKILL/OOM) between heartbeat renewal and the sweep.

## Why not a staleness window

Tried and reverted in #682. Elapsed time is not a liveness signal here:

- The 300s `statement_timeout` bounds each **statement**, not the job. The task runs a CTAS, a DELETE, an EXISTS probe, two ALTERs, a primary key, `add_4326_column` and registration in sequence, so a legitimate materialize over a large dataset can outlive any window short enough to be useful — and releasing the slot then lets a second CTAS through, defeating the cap.
- Measuring from enqueue is wrong the other way: a job can sit queued behind a backlog and only just have started.

Both directions were caught in review on #682 (`3650230574`, `3650230577`).

## Proposed fix

Give materialize a real lease, the way the ingest tasks already do:

1. ~~Stamp `attempt_id` when the job row is created, and pass it to the task.~~ **Not needed**: the column already defaults to `uuid4` at creation, and `retry=0` means there is no second delivery to fence against.
2. ~~In the task, `claim_ingest_job_attempt` then `maintain_ingest_job_heartbeat`, cancelling it at terminal state.~~ **Shipped** in #682 (`23bbb72e4`) and hardened in #700. Verified on `main` at 28e205ef2: the worker claims via `claim_ingest_job_attempt` on the row's own `attempt_id` (`backend/app/processing/analysis/tasks.py:592-594`), starts the heartbeat at `:618-623`, handles `CancelledError` from graceful shutdown at `:783-806`, and stops it with cancel+await in `finally` at `:830`.
3. **Change the cap predicate**, the only remaining work. Details below.
4. Once the server can distinguish dead from slow, the client can follow the same signal. It currently has no staleness rule on purpose (`AnalysisJobWatcher.tsx:48-56` spells out why), so the UI never disagrees with the API.

[...]

PR #802, opened 2026-07-27
fix(analysis): fence terminal writes, surface renames, size-check real output

### LINKED ISSUE #786: chore(analysis): backend job hygiene batch — unfenced terminal writes, discarded collision warning, premature size check
Three P3s from the same file, one PR's worth:

1. **Terminal job-status writes are not attempt-fenced** (`tasks.py:542-544`, `:252-257`) — plain ORM overwrites, unlike `_fail_cancelled_job` which routes through `update_ingest_job_for_attempt` ("Fenced on the attempt token FIRST"). If the heartbeat stalls a full lease and the sweeper fails the row, the worker later overwrites `failed` → `complete` — the exact resurrection the claim-fence comment (`:365-381`) guards against, reintroduced at the back. Related to #691's lease work.
2. **Table-name collision warning discarded** (`tasks.py:450` — `out_table, _warning =`) while the upload path persists it to `job.user_metadata` (`tasks_vector.py:413-422`). An analysis output can silently land in `parcels_buffered_3`.
3. **Pre-DELETE size check counts rows the DELETE removes** for buffer/centroid/dissolve (`tasks.py:478-493`) — the clip-by-layer branch filters inside the CTAS for exactly this reason (comment at `:313-319`); the other three shapes are an unfinished application of the same #719 review. Move the early check after the DELETE or extend the CTAS filter.

Umbrella fix worth doing alongside: give analysis jobs a warning channel (reuse `_append_job_warning`) so collisions, dropped columns, and geometry reclassification share one surface.
---

ishiland:
Three P3s from the 2026-07-27 builder/analysis audit, all in `backend/app/processing/analysis/tasks.py`.

## 1. Terminal job-state writes are now attempt-fenced

The success write (`job.status = "complete"`) and `_mark_job_failed`'s failure write were plain ORM overwrites, unlike `_fail_cancelled_job`, which routes through `update_ingest_job_for_attempt`. If the heartbeat stalled a full lease and the stale-job sweep failed the row, the worker later resurrected `failed` → `complete` — the exact clobber the claim-fence comment guards against, reintroduced at the back of the task.

- The success write moves into `_complete_job_for_attempt`: the fence shares the registration transaction, so a miss rolls the `Dataset` row back with it, and the already-committed output table (provably an unregistered orphan at that point — only this worker registers or completes, `retry=0`) is dropped instead of leaked.
- `_mark_job_failed` now fences FIRST, mirroring `_fail_cancelled_job`: a successful fence proves the row was still `running`, making the partial table safe to drop; on a miss both the row and the table are left alone (the failure path cannot distinguish a swept row from a `complete` row whose final commit reached the server just before the connection dropped).
- Scope: minimal fencing only, per the issue — the #691 heartbeat-lease design is untouched.

## 2. Table-name collision warning surfaced

`out_table, _warning = await generate_table_name(...)` threw the rename away, so an analysis output could silently land in `parcels_buffered_3`. It now persists to `user_metadata['collision_warning']` — the same key the upload path (`tasks_vector.py`) writes and `_job_to_status_response` already surfaces as `JobStatusResponse.warning_message`, so the existing job-status surface picks it up with no schema change.

## 3. Output-size check measures real output

The early `_enforce_output_size` probe runs on the CTAS before the NULL/EMPTY cleanup DELETE — and a DELETE cannot shrink `pg_total_relation_size` (dead tuples keep their pages until a rewrite), so "move the check after the DELETE" would have been a no-op. Fixed via the issue's other option: NULL/EMPTY-geometry rows are excluded **inside the CTAS** (`_wrap_not_empty`, with an `OFFSET 0` fence so the geometry expression isn't re-evaluated in the outer WHERE), extending the fix(#719 review) clip-by-layer rule to buffer/centroid/dissolve and the drawn-mask clip. The DELETE stays as the backstop that keeps the saved dataset agreeing with the preview.

## Tests

Four new regression tests in `backend/tests/test_analysis_materialize.py`, each verified to fail against the pre-fix code and pass with it:

- `test_sweeper_failed_job_is_not_resurrected_at_completion` (item 1, success path — no resurrection, registration rolled back, orphan table dropped)
- `test_sweeper_failed_job_error_is_not_overwritten` (item 1, failure path — sweeper's message preserved, table left alone)
- `test_name_collision_warning_is_surfaced` (item 2)
- `test_cleanup_rows_do_not_count_toward_size_cap` (item 3 — 2,025 NULL-geometry decoys under a 64 KB cap, same shape as the #719 clip test)

## Verification

- `cd backend && uv run ruff check . && uv run ruff format --check .` — clean.
- Against local Postgres (localhost:5434): `tests/test_analysis_materialize.py` + `test_analysis_preview.py` + `test_jobs_retry_analysis_copy.py` + `test_ai_chat_run_analysis.py` — **122 passed**; `test_ingest_job_attempt_fencing.py` + `test_stalled_queue_sweep_624.py` + `test_layering.py` — **45 passed**.

No API/schema/config impact.

Closes #786

### DIFF
--- a/backend/app/processing/analysis/tasks.py
+++ b/backend/app/processing/analysis/tasks.py
@@ -248,19 +248,83 @@ async def _enforce_output_size(
         )
 
 
+async def _complete_job_for_attempt(
+    session: AsyncSession,
+    *,
+    job_id: str,
+    attempt_id: uuid.UUID,
+    dataset_id: uuid.UUID,
+    schema: str,
+    out_table: str,
+    operation: str,
+) -> None:
+    """Commit registration together with the fenced terminal 'complete' write.
+
+    fix(#786): the terminal write is fenced on the attempt token, like the
+    claim and ``_fail_cancelled_job`` — a plain ``job.status = "complete"``
+    would overwrite a row the stale-job sweep already failed (a stalled
+    heartbeat expires the lease), resurrecting failed → complete and handing
+    the user a dataset for a job they were told failed. The fence shares the
+    registration transaction, so a miss rolls the Dataset row back with it.
+    """
+    if await update_ingest_job_for_attempt(
+        session,
+        uuid.UUID(job_id),
+        attempt_id,
+        values={"status": "complete", "dataset_id": dataset_id},
+    ):
+        await session.commit()
+        ANALYSIS_JOBS.labels(operation=operation, status="complete").inc()
+        return
+    await session.rollback()
+    logger.warning("analysis.complete_write_superseded", job_id=job_id)
+    # The build commit made the output table durable, and the rollback above
+    # discarded its only registration. Only this worker registers or
+    # completes the job (retry=0), so the table is a provable orphan — drop
+    # it instead of leaking it.
+    try:
+        await session.execute(text(f'DROP TABLE IF EXISTS "{schema}"."{out_table}"'))
+        await session.commit()
+    except Exception:  # broad: best-effort cleanup of the orphaned output
+        await session.rollback()
+
+
 async def _mark_job_failed(
     session: AsyncSession,
     *,
     job_id: str,
+    attempt_id: uuid.UUID,
     exc: Exception,
     schema: str,
     out_table: str | None,
     operation: str,
 ) -> None:
-    """Roll back, drop this job's own partial table, and record the failure."""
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

Issue #686, opened 2026-07-25
ux(analysis): post-preview continuity and buffer ergonomics


ishiland:
Polish items from the 2026-07-25 analysis audit, none blocking:

1. **Truncated preview → materialize continuity**: after "Showing the first 500 of N source features", nudge toward Create dataset (the full result is exactly what materialize is for) — e.g. a toast action or inline hint.
2. **Buffer units**: meters only today; US-centric users think in feet/miles. A unit select (m / km / ft / mi) converting to meters client-side keeps the API unchanged.
3. **Negative buffers** (shrink/inset): PostGIS supports negative distances on polygons; needs its own bounds and a polygon-only guard.
4. **Draw-mask accessibility**: the drawn clip mask is pointer-only. Clip-by-layer (#682) is the keyboard-accessible alternative; a docs note or in-panel hint should say so.

PR #803, opened 2026-07-27
fix(analysis): honor the stack selection, convert buffer units, and keep Add Data open

### LINKED ISSUE #772: ux(analysis): the panel ignores the layer selected in the stack and defaults to the first eligible one
**Where:** `frontend/src/components/builder/AnalysisPanel.tsx:111-118`; `BuilderRail.tsx:294-304`; `MapBuilderPage.tsx:823-841`

`layerId` initializes to `prefillLayerId ?? firstEligibleId` and the stack's `selectedLayerId` is never threaded into `railProps` — selecting a layer and opening Analysis lands on whatever sorts first, and later selection changes never propagate. The common intent ("analyze *this* layer") costs two clicks every time, or silently runs on the wrong dataset.

**Fix:** pass `selectedLayerId` through `railProps` and prefer it when analysable. Companion discoverability win: an "Analyze this layer" entry in the stack-row kebab (next to Zoom / Copy style / Open dataset) opening the panel prefilled. Related hardening: reconcile `layerId`/`maskLayerId` when the referenced layer is deleted while the panel is open — today both selects go blank with every button disabled and no message (`:155-168`).
---

### LINKED ISSUE #773: ux(analysis): switching buffer units reinterprets the number instead of converting it
**Where:** `frontend/src/components/builder/AnalysisPanel.tsx:240-241`, `:530-556`

`distanceValue = Number(distance) * BUFFER_UNIT_METERS[unit]`; changing the unit never touches the number. [live] 100 m → miles produced "100 miles", caught only because it exceeded the cap (good message: "no more than 62.14 miles"). The dangerous case is silent: 500 m → feet is a 3.28× semantic change with no visible difference and no validation trip. Combined with the stale-preview issue, the on-map preview still shows the old distance while the form means something else.

**Fix:** convert the displayed value on unit change (round sensibly) — `BUFFER_UNIT_METERS` already makes it trivial.
---

### LINKED ISSUE #776: ux(builder): the Add Data dialog closes after every single add
**Where:** `frontend/src/pages/MapBuilderPage.tsx:1908-1914`; `components/builder/DatasetSearchPanel.tsx:353-375`

`onAddDataset` calls `setShowAddData(false)` in the add callback, so building a five-layer map means five full open-search-add round trips. The panel is clearly designed for repeated adds — `renderDatasetAction` renders "Added ✓" and "another rendering" states that are almost unreachable because the dialog closes first — and the EmptyStackState direct-add path already supports repeated adds, so the two entry points disagree.

**Fix:** keep the dialog open on add and let the "Added" state do its job; close on Escape/backdrop/explicit button. Bonus in the same file: drop `pointer-events-none opacity-50` from the results container during debounced refetches (`:572-577`) — the pulse band at `:522-524` already signals it, and the freeze makes visible rows non-clickable on every keystroke pause.
---

ishiland:
Three analysis-panel UX fixes from the 2026-07-27 builder/analysis audit, batched because #772 threads a new source into the same layer/form precedence chain #773's converted value has to survive.

## #773 — unit switch converts instead of reinterpreting
Switching the buffer unit now converts the displayed number so the physical distance is preserved: 100 m → **0.1 km**, not "100 km". Conversion goes through the existing `BUFFER_UNIT_METERS` table and is rounded to 6 significant digits to avoid float tails (`0.3 km → 300`, not `300.00000000000006`). An empty/unparseable field just switches the unit. The existing input-change invalidation (preview clear, seq bump) is kept — the rounded value can differ from the previewed one by a hair, and the conservative path is the existing one.

## #772 — panel targets the stack-selected layer + kebab entry
`selectedLayerId` (the builder's `layers.expandedLayerId`) is now threaded through `railProps` → `BuilderRail` → `AnalysisPanel`.

**Chosen precedence (mount-time):**
1. explicit chat prefill (`prefill.layerId`)
2. stack-selected layer, when analysable (raster/DEM/group/'settings' ids validate false and change nothing)
3. remembered form's layer (#793 store)
4. first eligible layer

Honoring the #793 store semantics: when the stack selection **displaces** the remembered layer, the mount is treated as a different draft — the layer-coupled fields reset exactly as if the layer select had been edited (group-by → none per #680, a remembered mask layer that would now mask itself → none), any tracked run stays ambient instead of being adopted (`runDisowned` path), and a panel-owned preview overlay is cleared as stale. The layer-agnostic fields (operation, distance+unit, drawn mask, typed dataset name) still restore. Selection changes while the panel is already mounted deliberately do NOT re-target it — a live override would clobber an explicit in-panel choice and wipe a just-run preview; the kebab entry below covers that intent explicitly.

Also adds an **"Analyze this layer"** entry to the stack-row kebab (next to Zoom to layer), which opens the panel via the same prefill path the chat handoff uses — so it works even when the row isn't selected, and remounts an already-open panel via the prefill nonce key. The entry is gated on the panel's own eligibility predicate, extracted to `frontend/src/components/builder/analysis-eligibility.ts` so `StackRow` doesn't statically import the lazy-loaded panel chunk. New `stackRow.kebabAnalyzeLayer` key added to all four locales (en/es/fr/de).

## #776 — Add Data dialog stays open for multi-add
The add callback no longer calls `setShowAddData(false)` — the dialog stays open so the row's "Added ✓ / another rendering" states finally do their job, matching the POL-05 drag-drop path (which already kept the modal open) and the EmptyStackState direct-add path. Escape / backdrop / the dialog's X still close it.

**Editor-behind-dialog decision:** the old callback also ran `handleSelectLayer(newLayerId)`, which would now open the layer editor *behind* the still-open dialog. Instead, the last added layer id is parked in a ref and the editor opens **when the dialog closes** — the single-add flow ends in the same place it used to (editor open on the new layer), multi-add ends on the most recently added layer, and closing without adding selects nothing. The pending id is cleared on dialog open so a mid-flight add finishing after an instant close can't leak into a later session.

Bonus from the same issue: dropped the `pointer-events-none opacity-50` freeze on the results list during debounced refetches — the pulse band already signals them, and the freeze made visible rows unclickable on every keystroke pause.

## Tests
- `AnalysisPanel.test.tsx`: 13 new tests — unit-conversion behavior (convert, round, empty-field, wire-value preservation) and the full #772 precedence matrix (stack beats first-eligible, ineligible selection ignored, prefill wins, remembered-form displacement: field restore, group-by reset, self-mask reset, run stays ambient vs. still rehydrates on match, stale overlay clear). The #686 wire-conversion test now picks the unit before typing, since switching afterwards now converts the number.
- `StackRow.test.tsx`: kebab entry fires `onAnalyzeLayer`, hidden for non-analysable layers, hidden when unwired.
- `DatasetSearchPanel.test.tsx`: Test 2 now asserts the refetch band shows with **no** interaction freeze.
- `analysis-form-store.test.ts`: unchanged (no store changes) — still passes.

## Verification
- `cd frontend && npm ci`
- `npm run lint` — clean
- `npm run typecheck` — clean
- `npm run test:i18n` — 4/4 pass (locale parity incl. the new key)
- `npx vitest run` on: AnalysisPanel (62), AnalysisPanel.draw-suppression, analysis-form-store, StackRow, DatasetSearchPanel ×2, BuilderRail, UnifiedStackPanel, MapBuilderPage.{notes-ai-rail,a11y,settings-wiring,sheet-close-button,header-actions} — 208 tests, all passing

Skipped: the full `npm run test:coverage` suite and Playwright e2e (CI runs them). Checked e2e specs for reliance on the old auto-close: none — the existing "another rendering" flow already closes via Escape.

Note for the concurrent a11y PR: AnalysisPanel hunks stay clear of the clip-action label (~:942) and the status region (~:1118).

Closes #773
Closes #772
Closes #776

### DIFF
--- a/frontend/src/components/builder/AnalysisPanel.tsx
+++ b/frontend/src/components/builder/AnalysisPanel.tsx
@@ -29,6 +29,7 @@ import { useAuthStore } from '@/stores/auth-store';
 import { useAnalysisJobStore } from '@/stores/analysis-job-store';
 import { useMapDrawStore } from '@/stores/map-draw-store';
 import type { LayerActions } from '@/components/builder/ChatPanel';
+import { isAnalysableLayer } from '@/components/builder/analysis-eligibility';
 import type { EphemeralAnalysisHandoff } from '@/components/builder/hooks/use-ephemeral-layers';
 import type { AnalysisOperation, MapLayerResponse } from '@/types/api';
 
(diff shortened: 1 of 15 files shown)


ishiland:
@codex review

Issue #788, opened 2026-07-27
bug(a11y): a11y/i18n polish batch from the 2026-07-27 audit (P3)


ishiland:
Batched:

1. Hardcoded visible literal `Basemap · {name ?? 'Untitled'}` in the sublayer breadcrumb (`LayerEditorPanel.tsx:325`) — the aria-label one line above IS translated, so visible text and accessible name disagree in non-English locales; invisible to `npm run test:i18n` (key-existence checker only).
2. Two inline confirms claim `role="alertdialog"` without `aria-modal`, focus trap, or Escape (`row-chrome.tsx:147`, `LayerEditorPanel.tsx:479`) — Escape bubbles and closes the whole panel with the confirm pending. Add dialog semantics or drop to `role="group"` + `role="alert"`.
3. Vector `SliderRow` omits `aria-valuetext` (`HeatmapStyleControls.tsx:160-171`) so ~20 call sites announce "0.8" where the screen shows "80%" — the raster sibling and DEM/basemap scenes all set it. One-line fix, broad reach.
4. `analysisTools.truncatedNotice`, `truncatedNoticeTotal`, `ephemeralBadge.featureCountTruncated` pass reserved `count` with no plural variants, and mix locale-grouped `total` with ungrouped `count` in one sentence. [live] Related: the builder's DatasetSearchPanel card renders "1 features" (the catalog search card pluralizes correctly).
5. Both builder `<aside>` landmarks are unnamed (`BuilderRail.tsx:184`, `:230`) — "complementary" twice with no way to tell rail from panel; point `aria-labelledby` at the panel title, static `aria-label` on the rail.
6. Shortcuts sheet: Space's row-focused meaning (toggle selection) is undocumented and the existing "Space/Enter — toggle reorder" entry should be scoped to the drag handle (`StackRow.tsx:237-240` vs `KeyboardShortcutsSheet.tsx:75-93`); double-click-to-rename also absent.
7. Extend the axe e2e suite to AnalysisPanel and the layer-editor tabs — the existing coverage asserts none of the live-region behavior above.
---
Found by the 2026-07-27 builder/analysis audit. Items marked [live] were reproduced in the running v1.5.0 app.

ishiland:
[...]

PR #804, opened 2026-07-27
fix(a11y): builder quick wins — accessible names, live regions, translated drag announcements

### LINKED ISSUE #754: Viewer/builder: the analysis clip-action label overrides the accessible name of all three button states
In `frontend/src/components/builder/AnalysisPanel.tsx`, `<Label htmlFor="analysis-clip-action">Draw clip area</Label>` sits above a slot where three different buttons render by state, all reusing `id="analysis-clip-action"`: **Draw clip area**, **Cancel** (while drawing), and **Clear** (once a mask is set).

Per the accessible-name computation, the `<label for>` association overrides each button's own text, so a screen reader announces Cancel and Clear as "Draw clip area". It also means role+name queries cannot distinguish the three buttons in tests (they must query by text).

Found 2026-07-26 while writing the #726 regression tests; flagged in #729's body, now filed for the 1.5.1 polish batch alongside #725/#727/#731.

**Suggested fix:** drop the `htmlFor` (the button text is the label), or give each state's button its own id and explicit `aria-label`.

### LINKED ISSUE #782: bug(a11y): the Save button's aria-label overrides its sr-only status text
**Where:** `frontend/src/components/builder/MapTitleBar.tsx:169`, `:191` (label computed at `:88`)

The button sets `aria-label={saveButtonAriaLabel}` ("Save (⌘S)" / "Retry save") and carries an sr-only span with the status ("Saved / Saving… / Unsaved changes / Save failed") plus a comment claiming AT users hear it. `aria-label` takes precedence over element contents in the accessible-name computation, so the status is never conveyed. Same failure class as #754, different mechanism and file.

**Fix:** fold `saveStatusLabel` into the aria-label, and surface the status in a real `role="status"` region so save completion is announced without focus.
---

### LINKED ISSUE #783: bug(a11y): closing a rail panel drops keyboard focus to body
**Where:** `frontend/src/pages/MapBuilderPage.tsx:825`; `components/builder/BuilderRail.tsx:241`, `:258`

Escape and the header close button call `onPanelChange(null)`, unmounting the aside containing the focused element with no restoration — the user must Tab from the top of the document. The sibling path does it right: `handleCloseEditor` (`MapBuilderPage.tsx:1128-1149`) restores focus to `stack-row-{id}` with a deleted-row fallback.

**Fix:** mirror `handleCloseEditor` — focus the rail button for the panel that just closed; consider extracting a shared `useRestoreFocus` so the two dismissal paths can't drift again.
---

### LINKED ISSUE #784: bug(a11y): a successful analysis preview is never announced to assistive tech
**Where:** `frontend/src/components/builder/EphemeralBadge.tsx:35`; `AnalysisPanel.tsx:273-325`

The badge is the primary confirmation that a preview succeeded and it is a plain div — no `role="status"`, no live region. Only the empty-result and truncated branches raise toasts (sonner has its own live region); the ordinary success path is silent. A screen-reader user presses Preview, hears the button flip to "Running…" and back, and is never told N features rendered.

**Fix:** `role="status"` on the badge wrapper. Related: the panel's job-status `<p role="status">` is conditionally rendered (`AnalysisPanel.tsx:738-739`), so it mounts already populated and the first message ("Creating dataset…") is typically not announced — render it unconditionally with empty content instead. Add `aria-busy` to the Preview button while pending (the rail button already does this for materialize).
---

### LINKED ISSUE #785: bug(a11y): drags emit dnd-kit's untranslated default announcements containing raw layer UUIDs
**Where:** `frontend/src/pages/MapBuilderPage.tsx:1530` (`<DndContext sensors={dndSensors}>` — no `accessibility` prop)

dnd-kit's default announcements are hardcoded English built from the draggable id: "Picked up draggable item 3f2b8c1e-…." Every drag emits that into dnd-kit's live region — in English regardless of locale — while the app's own `announce()` emits the correct localized, layer-named message into a second region. Two competing announcements, one of them noise.

**Fix:** pass `accessibility={{ announcements }}` wired to the existing `a11y.drag*` keys (already present in all four locales), plus `screenReaderInstructions` documenting the grip's reorder keys.
---

ishiland:
A11y quick-wins batch from the 2026-07-27 builder/analysis audit (v1.6.0 triage).

## Per issue

**#754 — clip-action label overrides button names** (`AnalysisPanel.tsx`)
Dropped the `htmlFor="analysis-clip-action"` on the section label and the shared `id` on the three state-dependent buttons. A `<label for>` aimed at a button overrides the button's own text in the accessible-name computation, so Cancel and Clear were both announced as "Draw clip area". The buttons are self-labeling; hunks kept tightly scoped away from the sibling units/precedence PR.

**#782 — Save button aria-label swallows the status** (`MapTitleBar.tsx`)
The status is folded into the button's aria-label ("Save (⌘S), Unsaved changes" / "Retry save, Save failed") and mirrored in a persistent sr-only `role="status"` region next to the button, so save completion is announced without focus. The old sr-only span inside the button was dead code — `aria-label` takes precedence over element contents.

**#783 — closing a rail panel drops focus to body** (`BuilderRail.tsx`)
Escape and the header close button now route through a `closePanel` callback that mirrors `handleCloseEditor`: focus returns to the rail button (`data-rail-button={id}`) that owns the just-closed panel, via rAF. In sheet mode (no rail) Radix's own close-time focus restoration applies.

**#784 — successful preview never announced** (`EphemeralBadge.tsx`, `AnalysisPanel.tsx`)
Live regions only announce mutations made while the region is already in the AX tree — mounting populated is silent. The badge now mounts an empty sr-only `role="status"` twin and populates it a frame later (visible text is `aria-hidden` so browse mode doesn't read it twice). The panel's two conditional job-status `<p role="status">` blocks merged into one persistently-rendered region that mounts empty, so the first message ("Creating dataset…") and each later one arrive as announced mutations.

**#785 — dnd-kit default announcements: untranslated + raw UUIDs** (`MapBuilderPage.tsx`)
`<DndContext>` now gets an `accessibility` prop: announcements built from the existing `a11y.drag*` keys using human layer names (stack rows resolve `display_name ?? dataset_name`; the basemap row resolves its preset name), plus `screenReaderInstructions` documenting the drag handle's actual keys (new `a11y.dragInstructions` key, all four locales). Catalog drags return `undefined` — their pickup/position/drop already flow through the page's own localized `announce()` region, whose drop announcement deliberately fires only after the add-dataset mutation resolves (a guarantee dnd-kit's synchronous `onDragEnd` can't honor). No drag listeners or row spread props touched.
Key-parity note: the triage said en had one more `drag` key than es/fr/de — a full key-set diff of every namespace shows all four locales identical; the extra grep hit was the English *value* "focused drag handle" inside `a11y.shortcuts.reorderToggle`, which the other locales translate. Nothing to reconcile.

**#788 item 1 — hardcoded breadcrumb literal** (`LayerEditorPanel.tsx`)
`Basemap · {name} ›` is now `t('basemapSublayer.breadcrumb')` with a translated `breadcrumbUntitled` fallback (all four locales), so visible text and the translated aria-label agree in every locale.

**#788 item 3 — vector SliderRow missing aria-valuetext** (`HeatmapStyleControls.tsx`, `components/ui/slider.tsx`)
`SliderRow` now passes `aria-valuetext` with the formatted display value, matching `RasterSliderRow` and the DEM/basemap scenes. One deliberate extra: the ui `Slider` wrapper now forwards `aria-valuetext` to the Radix `Thumb` (exactly like the existing `aria-label` forwarding) — Radix leaves unknown props on the Root span, which has no slider role, so the valuetext set by every existing call site never actually reached AT. This makes the whole convention effective (test pins the attribute on the `role="slider"` element).

**#788 item 5 — unnamed aside landmarks** (`BuilderRail.tsx`)
Icon rail: static `aria-label` (new `rail.toolsLabel` key, all four locales). Expanded panel: `aria-labelledby` pointing at its own title span.

Part of #788 (items 1, 3, 5) — items 2, 4, 6, 7 are intentionally not addressed here.

Closes #754
Closes #782
Closes #783
Closes #784
Closes #785

## i18n
New keys added literally to en/es/fr/de: `a11y.dragInstructions`, `rail.toolsLabel`, `basemapSublayer.breadcrumb`, `basemapSublayer.breadcrumbUntitled`.

## Tests
- `AnalysisPanel.test.tsx`: clip buttons keep their own accessible names; job status renders in a persistent `role="status"` region that mounts empty.
- `MapTitleBar.test.tsx`: aria-label folds the status; `role="status"` region tracks unsaved → saving → saved; failed-state assertions updated (status text moved out of the button).
- `BuilderRail.test.tsx`: focus returns to the rail button on close-button and Escape dismissal; both asides have accessible names.
- `HeatmapStyleControls.test.tsx`: all three sliders expose formatted `aria-valuetext` on the `role="slider"` element.
- `MapBuilderPage.a11y.test.tsx`: DndContext `accessibility` wiring source pin + translated drag-handle instructions mounted in the DOM.
- `LayerEditorPanel.test.tsx`: i18n mock now interpolates params so the existing breadcrumb assertions still assert the rendered strings.

## Verification
```
cd frontend && npm ci
npm run lint          # pass
npm run typecheck     # pass
npm run test:i18n     # pass (locale parity incl. the 4 new keys)
npx vitest run src/components/builder   # 129 files, 1935 tests, pass
npx vitest run src/pages/__tests__/MapBuilderPage.header-actions.test.tsx src/pages/__tests__/PublicMapViewerPage.test.tsx src/pages/__tests__/PublicViewerPage.test.tsx src/pages/__tests__/MapViewerGate.test.tsx   # pass
```

## Skipped
- Full `npm run test:coverage` suite — CI runs it.
- E2E/axe suites (stack not running here); the e2e selectors this PR could affect were checked: `Query result · N features` (Playwright ignores aria-hidden for text queries + visibility) and the `Close panel` click in `analysis.spec.ts` are unaffected.
- No screenshots: no visual changes (sr-only/ARIA-only, plus one string now translated).

### DIFF
--- a/frontend/src/components/builder/AnalysisPanel.tsx
+++ b/frontend/src/components/builder/AnalysisPanel.tsx
@@ -937,9 +937,11 @@ export function AnalysisPanel({
 
       {operation === 'clip' && maskLayer == null && (
         <div className="space-y-1.5">
-          {/* Exactly one of the three buttons below renders at a time, so
-              they can share the id this label points at. */}
-          <Label className="text-xs" htmlFor="analysis-clip-action">
+          {/* fix(#754): no htmlFor here — a <label for> pointed at a button
+              OVERRIDES the button's own text in the accessible-name
+              computation, so Cancel and Clear were both announced as "Draw
+              clip area". The buttons below are self-labeling. */}
+          <Label className="text-xs">
             {t('analysisTools.drawMask', { defaultValue: 'Draw clip area' })}
           </Label>
           {isDrawing ? (
@@ -950,7 +952,6 @@ export function AnalysisPanel({
                 })}
               </p>
               <Button
-                id="analysis-clip-action"
                 variant="outline"
                 size="sm"
                 onClick={stopDrawing}
@@ -964,7 +965,6 @@ export function AnalysisPanel({
                 {t('analysisTools.maskSet', { defaultValue: 'Clip area set' })}
               </span>
               <Button
-                id="analysis-clip-action"
                 variant="ghost"
                 size="sm"
                 onClick={() => {
@@ -981,7 +981,6 @@ export function AnalysisPanel({
           ) : (
             <div className="space-y-1.5">
               <Button
-                id="analysis-clip-action"
                 variant="outline"
                 size="sm"
                 onClick={startDrawing}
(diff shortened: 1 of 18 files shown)


ishiland:
@codex review

PR #807, opened 2026-07-27
fix(analysis): re-add the drawn clip mask after a basemap style reload

### LINKED ISSUE #775: bug(analysis): a basemap change erases the drawn clip mask with no re-add path
**Where:** `frontend/src/components/builder/AnalysisPanel.tsx:220-235`; contrast `hooks/use-ephemeral-layers.ts:135-139`

The finished mask is kept visible via TerraDraw static mode, which owns its own MapLibre source/layers. A basemap switch calls `setStyle`, wiping every custom layer — the ephemeral-overlay hook subscribes to `style.load` for exactly this (the #394 fix), but the TerraDraw adapter (1.4.1) has no style-reload handling and AnalysisPanel registers none. The `mask` state survives, so the operation still runs correctly — but the panel says "Clip area set" over a map with no visible mask, so the user cannot verify what they are about to clip to.

**Fix:** mirror the ephemeral hook — re-add the polygon on `style.load` for the lifetime of `mask`, or render the finished mask through the ephemeral pipeline instead of TerraDraw static mode.
---

ishiland:
## What changed

A basemap switch calls `setStyle`, which wipes every imperatively added source/layer — including the TerraDraw static overlay that keeps the finished clip mask visible (`terra-draw-maplibre-gl-adapter` 1.4.x has no style-reload handling: its `register()` adds sources/layers once and nothing re-adds them). The `mask` state survived, so the operation still ran correctly, but the panel said "Clip area set" over a map with no visible mask and no re-add path.

Fix in `frontend/src/components/builder/AnalysisPanel.tsx`, mirroring the `use-ephemeral-layers` #394 pattern rather than inventing a new one:

- Extracted the kept-static TerraDraw overlay construction (previously inlined in the #793 mount-restore effect) into a shared `addStaticMaskOverlay` helper. The restore effect's latch/no-dep-array semantics are unchanged.
- New effect: while a drawn mask is set (and the map ref exists), subscribe to `style.load` and rebuild the overlay on every fire — stopping the orphaned instance first, guarded, since the adapter's `unregister()` removes its layers unguarded and the new style already dropped them. The cleanup drops the subscription when the mask clears or the panel unmounts, so a cleared mask cannot resurrect.
- Side benefit: a mask restored before the initial style finished loading (where the #793 latch-once creation could fail silently) now heals on the first `style.load`.

Mask state stays client-side — no `basemap_config` fields, no layer PATCH involvement, no API/schema/config impact. Not covered (pre-existing, separate surface): a basemap switch mid-draw, before the polygon is finished — the mask does not exist yet at that point.

Tests:

- New: `re-adds the drawn mask overlay after a basemap style reload (#775)` — mock map with a real `style.load` handler registry; asserts the overlay is rebuilt from `mask` (old instance stopped, new TerraDraw constructed, `setMode('static')`) and that clearing the mask unsubscribes so a later reload adds nothing. Fails against the previous `AnalysisPanel.tsx`, passes with the fix.
- Existing map fixtures in the two AnalysisPanel test files grew `on`/`off` stubs, since the panel now subscribes to the map whenever a mask is set.

## Verification

```
cd frontend && npm run lint
cd frontend && npm run typecheck
cd frontend && npx vitest run src/components/builder/__tests__/AnalysisPanel.test.tsx src/components/builder/__tests__/AnalysisPanel.draw-suppression.test.tsx
```

All clean; 70/70 tests pass. No new translation keys.

Closes #775

### DIFF
--- a/frontend/src/components/builder/AnalysisPanel.tsx
+++ b/frontend/src/components/builder/AnalysisPanel.tsx
@@ -347,55 +347,91 @@ export function AnalysisPanel({
   // the panel said "Clip area set" over an empty map. Recreate the
   // kept-static overlay exactly as the finish handler leaves it. Best
   // effort: if TerraDraw rejects the feature, the mask still applies (the
-  // pre-fix behavior), just without the outline.
+  // pre-fix behavior), just without the outline. Shared by the mount
+  // restore below and the fix(#775) style.load re-add.
+  const addStaticMaskOverlay = useCallback(
+    (map: MaplibreMap, maskGeometry: GeoJSON.Polygon) => {
+      // fix(#793 review): held locally so the catch can reach a PARTIALLY
+      // started instance — drawRef is only assigned on full success, so
+      // stopping drawRef there stopped nothing while the failed instance kept
+      // its map layers and handlers attached.
+      let td: TerraDraw | null = null;
+      try {
+        td = new TerraDraw({
+          adapter: new TerraDrawMapLibreGLAdapter({ map }),
+          modes: [
+            new TerraDrawPolygonMode({
+              styles: {
+                fillColor: MAP_COLORS.default.fill,
+                fillOpacity: MAP_COLORS.default.fillOpacity,
+                outlineColor: MAP_COLORS.default.stroke,
+                outlineWidth: MAP_COLORS.default.strokeWidth,
+              },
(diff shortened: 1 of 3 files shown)


ishiland:
@codex review

Issue #755, opened 2026-07-27
Viewer/builder: returning to a backgrounded tab 403-bursts vector tiles until the reactive re-mint lands


ishiland:
**Observed on the demo (1.5.0), via the in-app problem reporter**: switch away from a map tab for a few minutes, switch back, and every visible vector tile 403s before the map heals. The reporter captures pairs of suppressed `BLDR-TILE-RACE` warnings plus **unsuppressed red console `AJAXError (403)`** entries, which is what makes a recovered situation read as a breakage.

**Mechanics (not a regression — this machinery predates 1.5.0):**

- Tile tokens use `round_expiry()` with 900 s boundaries, so a signed URL lives 15–30 min (`backend/app/processing/tiles/signing.py:62`, mint at `processing/tiles/router.py:1224`).
- A backgrounded tab crosses the boundary; on `visibilitychange` back to visible, MapLibre resumes fetches with the stale `sig` → 403 per tile.
- The GUARD-03/#621 handler (`BuilderMap.tsx:663`) re-signs from cache and, since the sig is expired, kicks the throttled `recoverTileAuth()` re-mint. The map self-heals after one round trip. Recovery works; the experience and the telemetry are the problem.

**Two fixes, independently shippable:**

1. **Proactive refresh.** The client already knows `exp` (URL param) and `expires_in` (mint response) but only reacts to 403s. Add a `visibilitychange` listener: on becoming visible with any token past or within ~60 s of `exp`, re-mint and re-sign sources before MapLibre's retries hit the network. Kills the burst for the dominant tab-switch case.
2. **Stop double-logging handled 403s.** When the re-sign path recovers a tile, the raw `AJAXError` still reaches `console.error` and lands as an unsuppressed red row in the problem report next to four suppressed yellows. Either mark the correlated console entry suppressed in the reporter buffer or prevent the underlying log for the handled 401/403 vector-tile case.

Capture excerpt:

```
[warning] (maplibre) Vector tile re-signed after HTTP 403 (BLDR-TILE-RACE) (suppressed)
[error]   (console) AJAXError: (403): /api/tiles/data.prueba/4/8/5.pbf?sig=…&exp=1785148200&scope=prueba&cols=id&_v=1
```

ishiland:
[...]

PR #808, opened 2026-07-27
fix(map): stop double-logging tile-auth 403s the recovery path already handles

ishiland:
## What

Fix 2 of #755, per the v1.6.0 scoping comment (fix 2 only; the proactive `visibilitychange` re-mint stays on the issue).

Returning to a backgrounded map tab past the tile-signature boundary 403s every visible vector tile. The GUARD-03 re-sign / #621 throttled re-mint path recovers the map, but each 403 still landed as an **unsuppressed red `AJAXError` console row** in the problem report next to the suppressed recovery warnings — making a recovered situation read as a breakage.

Root cause of the duplicate: `@vis.gl/react-maplibre` subscribes the map `error` event unconditionally and falls back to `console.error(e.error)` whenever the `<Map>` component has no `onError` prop — in parallel with each surface's own `map.on('error')` recovery handler (`maplibre.ts` `_onEvent`).

- New `frontend/src/lib/map-error-log.ts`: `logUnhandledMapError` drops the log for exactly the handled first-party 401/403 vector-tile case (relative/same-origin `/tiles/` URL, or a CDN tile URL carrying our `sig=` param) and replicates the wrapper's default `console.error` for everything else.
- Wired as the `onError` prop on the three surfaces that own tile-auth recovery: `BuilderMap`, `ViewerMap`, `DatasetMap`.

A handled 403 now logs **not at all** through the wrapper path; the recovery path's suppressed `BLDR-TILE-RACE` report entries are unchanged. Third-party basemap 401/403s, raster-tile errors, 404 no-data tiles, 5xx, and status-less style errors keep the exact default single log. No API/schema/config impact; no new translation keys; `error-map.ts` untouched.

## Verification

```
cd frontend
npm run lint
npm run typecheck
npx vitest run src/lib/__tests__/map-error-log.test.ts   # 10 passed
npx vitest run src/components/builder/__tests__/BuilderMap.*.test.* \
  src/components/viewer/__tests__/ViewerMap.*.test.* \
  src/components/dataset/__tests__/DatasetMap.*.test.*    # 14 files, 106 passed
```

Part of #755 — the remainder (fix 1, proactive re-mint on `visibilitychange`) stays tracked on the issue per the scoping comment.

### DIFF
--- a/frontend/src/components/builder/BuilderMap.tsx
+++ b/frontend/src/components/builder/BuilderMap.tsx
@@ -13,6 +13,7 @@ import {
   BLANK_BASEMAP_ID,
 } from '@/lib/basemap-utils';
 import { buildClusterTileUrl, buildSignedTileUrl, isMvtSourceLayerConfigReady } from '@/lib/tile-utils';
+import { logUnhandledMapError } from '@/lib/map-error-log';
 import { MAP_COLORS } from '@/lib/map-colors';
 import { useTileTokens, useInvalidateTileTokens } from '@/hooks/use-tile-token';
 import { useTileAuthRecovery } from '@/hooks/use-tile-auth-recovery';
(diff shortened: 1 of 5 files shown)


ishiland:
@codex review

PR #801, opened 2026-07-27
fix(builder): correctness batch — phantom legend groups, JSON-apply zoom reset, empty-state copy

ishiland:
Builder correctness one-liners from the v1.6.0 triage: #769, #770, #780, plus the regression test #768 asked for.

### #769 — folder-group rows rendered phantom legend entries (on screen and in PNG export)
The synthetic `group:folder` row is built by spreading the group's first child (`folder-groups.ts` `hydrateFolderGroupLayers`), so it inherits `visible`/`show_in_legend`/`paint` — and neither legend filter had a folder-group predicate. Added the existing shared `isFolderGroupLayer()` predicate to both filter sites:
- `frontend/src/components/map-plugins/builtin/LegendPlugin.tsx` (on-screen legend)
- `frontend/src/components/builder/hooks/use-builder-save.ts` (`handleExportPNG` legend block)

New tests: LegendPlugin render tests (group row excluded; placeholder shown when only a group row qualifies) and a PNG-export test asserting the group name is never drawn. All four fail against the unfixed code.

### #770 — applying the Advanced JSON block silently reset the layer's min/max zoom
The editor shows paint/layout with `_`-prefixed builder-private keys stripped (`stripBuilderPrivate`), and Apply replaced the block wholesale — an edit-free open→Apply of the Layout block reset `_minzoom`/`_maxzoom` (the zoom sliders' only storage) to 0–22; the Paint block lost its `_`-keys the same way. The onApply wrappers in `LayerStyleEditor.tsx` now re-merge the stripped private keys (`{ ...builderPrivateKeys(original), ...applied }`). Legacy non-underscore keys removed by `stripLegacyBuilderPaint` are deliberately NOT re-merged — they are being scrubbed.

New tests: unedited Layout Apply preserves `_minzoom: 5`/`_maxzoom: 12`; Paint Apply keeps `_outline-color` while applying the edited JSON. Both fail against the unfixed code.

### #780 — first-run empty-state copy pointed at absent controls
`SUGGESTED_DATASETS` ships `[]` by design, so stock installs never render a starter list, and there is no Upload button on this surface (upload lives inside the Add Data modal, permission-gated). Reworded `unifiedStack.emptyBody` and `unifiedStack.emptyHelpBody` in **all four locales (en/es/fr/de)** — and the matching in-code `defaultValue`s in `EmptyStackState.tsx` — to name the controls that actually exist: catalog search and the header's "Add data" button (localized label in each locale). Copy-only; no logic change.

[...]

PR #811, opened 2026-07-27
fix(builder): keep groups, selection, and bulk delete consistent with the stack

### LINKED ISSUE #767: bug(builder): a folder group with no children is silently discarded on save
**Where:** `frontend/src/components/builder/folder-groups.ts:161-190` (persist), `:118-159` (hydrate)

Group identity is persisted only on children (`builder.folderGroupId` in each child's `style_config`); group rows are filtered out of the persisted payload and reconstructed only when a child carries the marker. A group whose last child is removed still renders, accepts drops, and marks the map dirty — but on save + reload it vanishes, along with any rename. Reachable via move-out (`use-folder-group-layers.ts:200-224`), drag-out (`MapBuilderPage.tsx:1085-1098`), and bulk delete of all children (`use-bulk-layer-actions.ts:316-320`).

**Fix:** persist empty groups (a map-level `group_meta` record already exists — `use-builder-layers.ts:222-224`) or prune the empty group row at the moment it empties so UI and persisted state agree.
---

### LINKED ISSUE #771: bug(builder): selection can act on rows the user cannot see, and bulk-deleting group rows is a silent no-op
**Where:** `frontend/src/pages/MapBuilderPage.tsx:633-639`; `components/builder/BulkActionBar.tsx:192`, `:215`; `hooks/use-bulk-layer-actions.ts:302-308`

Three related drifts between selection and reality:
1. `selectableRowIds` is the raw flat array, so shift-click range selection includes children of collapsed groups and rows hidden by the layer search — bulk delete/visibility/opacity then act on rows the user cannot see.
2. The delete confirmation counts the full selection while the handler silently filters out `group:folder` rows and returns `false` when nothing remains: select two group headers → "Delete 2 layers" → nothing happens, no toast, selection preserved.
3. `selectedIds` is never pruned when layers are removed outside the bulk path (`handleRemove`, `handleDeleteGroup`, `handleUngroup`), so the bar over-reports and confirms counts it will not perform.

**Fix:** derive `selectableRowIds` from the rendered row set; exclude group rows from the confirm count (or route them through `handleDeleteGroup`) and toast the zero-deletable path; prune `selectedIds` against `localLayers` in an effect.
---

### LINKED ISSUE #805: bug(builder): partial bulk-delete restores survivors at the wrong position; grouped saves emit a spurious per-child style_config PATCH
Split from #787 (items 7 and 8) — the two real data bugs in that batch, kept in v1.6.0 while the rest of #787 moves to v1.7.0.

1. **Partial bulk-delete failure restores survivors at the wrong stack position** (`use-bulk-layer-actions.ts:377-384` merges two renumbered `sort_order` schemes) — and the wrong order is then saved. Fix: rebuild from `previousLayers` minus the deleted rows.
2. **Grouped maps emit a spurious per-child `style_config` PATCH on every save** (`use-builder-save.ts:451-452` — baseline prepared without `groupMeta`), silently persisting collapse state that was deliberately non-dirtying.

Coordination note: `use-bulk-layer-actions.ts` is also in #771's blast radius (Day-3 batch) — land or rebase these together. `use-builder-save.ts` is touched by PR #801 (different hunk, legend filter).

ishiland:
Fixes three related builder stack-state drifts in one pass, since they share the same blast radius (`use-bulk-layer-actions.ts`, the selection plumbing, and the save diff).

## #767 — empty folder groups vanished on save+reload

Group identity is persisted only on children (`builder.folderGroup*` markers), so a group whose last child left had no carrier: it kept rendering, accepted drops, and marked the map dirty, then silently vanished (with any rename) on the next hydrate. The backend has no map-level `group_meta` field to persist an empty group into, so this takes the issue's other sanctioned fix: **prune the group row the moment it empties**, making the UI agree with what persistence can actually round-trip.

- New `pruneEmptyFolderGroups()` in `folder-groups.ts` (identity-stable when nothing to prune).
- Applied inside the state write at every emptying site: `handleRemove`, `handleAiRemoveLayer` (draft remove), `handleReorder` (covers the drag-out dispatch from `handleDragEnd`), `handleMoveLayerOutOfGroup`, `handleAddLayerToExistingGroup` (defensive), and `handleBulkDelete` (optimistic update, full-success baseline sync, and the partial-failure rebuild). Rollback paths restore the pre-delete snapshot, so a failed delete brings the group back with its child.

## #771 — selection could act on rows the user cannot see; group-row bulk delete was a silent no-op

1. `selectableRowIds` was the raw flat layer array, so shift-click ranges (and Shift+Arrow) swept children of collapsed groups and rows hidden by the layer search. It is now derived from the **rendered row set** via a shared `computeSelectableRowIds()` helper in `selection-utils.ts`; `UnifiedStackPanel` consumes the same helper for its row-skip decisions so the two can never drift. The panel's search query is lifted hybrid-controlled into `MapBuilderPage` (internal state remains for standalone/test usage).
2. Every delete-facing count in `BulkActionBar` (confirm label/action, deleting announcement, aria labels) now uses the **deletable count** — group rows are filtered out of the batch by the handler, so counting them promised deletions that never happened. The Delete item disables with an inline hint when nothing is deletable, and `handleBulkDelete` toasts the zero-deletable path instead of returning `false` silently.
3. A `MapBuilderPage` effect prunes `selectedIds` against the live rows, so removals outside the bulk path (`handleRemove`, `handleDeleteGroup`, `handleUngroup`, the #767 prune) no longer leave the bar over-reporting. The effect is skipped while a bulk delete is in flight so the preserved-on-failure selection survives the optimistic removal + rollback cycle.

## #805 — partial bulk-delete scrambled survivor order; grouped saves emitted a spurious per-child PATCH

1. The partial-failure path merged the already-renumbered optimistic array with failed rows still carrying their **original** `sort_order`, interleaving two numbering schemes — survivors slotted back at the wrong stack position and that order was then saved. Survivors are now rebuilt from `previousLayers` minus the actually-deleted rows.
2. `buildLayerDiff` prepared its baseline **without** `groupMeta` while the current side got it, so baseline children lacked the `folderGroupExpanded` marker and every save of a grouped map emitted a per-child `style_config` PATCH — silently persisting collapse state that is deliberately non-dirtying. Both sides now use the same `groupMeta`; collapse state still rides along only when a real change patches `style_config` or a full PUT runs.

## Config/i18n impact

Two new translation keys, literal in all four locales (en/es/fr/de): `toasts.bulkDeleteOnlyGroupRows`, `bulkActions.deleteHint`. No API, schema, or backend changes.

## Tests

New coverage, all verified to **fail against the pre-fix code** (5 targeted regression tests confirmed red with the fixes stashed) and pass with it:

- `folder-groups.test.ts`: `pruneEmptyFolderGroups` unit tests (prunes childless, keeps populated, identity-stable, ignores non-folder rows).
- `selection-utils.test.ts`: `computeSelectableRowIds` render-order/collapse/search matrix (7 cases).
- `use-builder-save.test.ts`: unchanged grouped map produces an empty diff; collapse-state-only change stays out of the diff.
- `use-builder-layers.bulk-ops.test.ts`: partial-failure survivor position is exact (`[b,c,d,e]`, not interleaved); bulk-deleting all children prunes the group row and never sends the synthetic id; group-only selection toasts instead of no-oping.
- `use-builder-layers.groups.test.ts`: move-out of the last child prunes (and keeps the group while children remain); drag-out via `handleReorder` prunes; single `handleRemove` of the last child prunes.
- `BulkActionBar.test.tsx`: mixed selection confirms the deletable count; group-only selection disables Delete with the hint; deleting state announces the deletable count.

## Verification

```
cd frontend
npm run lint
npm run typecheck
npm run test:i18n
npx vitest run src/components/builder src/pages   # 154 files, 2082 tests
```

All green on the rebased branch; the #801 layer-contiguity property test and the #794 keyboard-reorder tests are untouched and passing. No dnd-kit listener chrome was modified.

Closes #767
Closes #771
Closes #805

### DIFF
--- a/frontend/src/components/builder/BulkActionBar.tsx
+++ b/frontend/src/components/builder/BulkActionBar.tsx
@@ -80,6 +80,16 @@ export const BulkActionBar = memo(function BulkActionBar({
     [layers, selectedIds],
   );
 
+  // fix(#771): group container rows have no backend record — handleBulkDelete
+  // filters them out of the batch — so every delete-facing count must be the
+  // DELETABLE count, not the selection size. Counting group rows promised
+  // deletions that never happened ("Delete 2 layers" on two group headers →
+  // silent no-op) and the Delete item now disables when nothing is deletable.
+  const deletableCount = useMemo(
+    () => selectedLayers.filter((l) => !isFolderGroupLayer(l)).length,
+    [selectedLayers],
+  );
+
   // Derived values for display and enable/disable logic
   const avgOpacity =
     N > 0
@@ -149,7 +159,7 @@ export const BulkActionBar = memo(function BulkActionBar({
           role="toolbar" must not carry aria-live. */}
       <span className="sr-only" aria-live="polite" aria-atomic="true">
         {isDeleting
-          ? t('bulkActions.deletingLayers', { count: N })
+          ? t('bulkActions.deletingLayers', { count: deletableCount })
           : t('bulkActions.liveAnnouncement', { count: N })}
       </span>
       {isDeleting ? (
@@ -159,7 +169,7 @@ export const BulkActionBar = memo(function BulkActionBar({
         <div className="flex items-center gap-2 w-full">
           <Loader2 className="size-4 animate-spin text-muted-foreground shrink-0" aria-hidden="true" />
           <span className="text-sm text-muted-foreground flex-1">
-            {t('bulkActions.deletingLayers', { count: N })}
+            {t('bulkActions.deletingLayers', { count: deletableCount })}
           </span>
           <Button
             type="button"
@@ -168,7 +178,7 @@ export const BulkActionBar = memo(function BulkActionBar({
             className="h-8 gap-1.5 px-2 shrink-0 text-destructive cursor-not-allowed"
             disabled={true}
             aria-busy={true}
-            aria-label={t('bulkActions.deleteAriaLabel', { count: N })}
+            aria-label={t('bulkActions.deleteAriaLabel', { count: deletableCount })}
           >
             <Loader2 className="size-4 animate-spin" aria-hidden="true" />
             {t('bulkActions.delete')}
@@ -189,7 +199,9 @@ export const BulkActionBar = memo(function BulkActionBar({
             id={confirmId}
             className="flex-1 text-sm text-destructive text-center"
           >
-            {t('bulkActions.deleteConfirmLabel', { count: N })}
+            {/* fix(#771): the confirm promises only what the handler will
+                actually delete — group rows are filtered out of the batch. */}
+            {t('bulkActions.deleteConfirmLabel', { count: deletableCount })}
           </p>
           <Button
             type="button"
@@ -217,7 +229,7 @@ export const BulkActionBar = memo(function BulkActionBar({
             }}
             onPointerDown={(e) => e.stopPropagation()}
           >
-            {t('bulkActions.deleteConfirmAction', { count: N })}
+            {t('bulkActions.deleteConfirmAction', { count: deletableCount })}
           </Button>
         </div>
       ) : (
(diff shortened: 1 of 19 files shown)


ishiland:
@codex review

Issue #653, opened 2026-07-23
Admin AI surfaces gate on manage_users only — backend switches to manage_tenants in multi-tenant mode


ishiland:
Found while fixing the same class of bug in PR #652 (codex round 4).

`require_ai_status_reader` (`backend/app/modules/admin/router.py:61`) resolves to `manage_users` in single-tenant mode but `manage_tenants` in multi-tenant mode. The pre-existing admin AI surfaces gate on `can('manage_users')` alone:

- `frontend/src/components/admin/AIStatusCard.tsx` (`canManageUsers` gates the whole card and both queries)
- `frontend/src/components/admin/settings/SettingsAITab.tsx` embedding-stats/backfill block (still `canManageUsers`)

In a multi-tenant deployment a fleet operator with `manage_tenants` but not `manage_users` gets no AI status UI even though the backend would serve them, and a `manage_users`-only user fires queries the backend 403s.

Reference fix: the `canProbe` pattern from #652 — `isMultiTenant ? can('manage_tenants') : can('manage_users')` via `useEdition()`. Consider extracting a small `useAIStatusReader()` helper if more surfaces accumulate.

P3: only reachable in multi-tenant (enterprise overlay) deployments; single-tenant behavior is correct today.

https://claude.ai/code/session_01EJ3PZWCqfT3zCPQDCVuVUk

PR #816, opened 2026-07-28
fix(admin): gate the AI availability probe on the mode-aware status capability

### LINKED ISSUE #815: bug(admin): use-ai-availability gates /admin/ai-status on isAdmin, not the mode-aware reader capability
Found while fixing #653. `frontend/src/hooks/use-ai-availability.ts:55` gates its `/admin/ai-status` query on `isAdmin` from the auth store rather than capabilities. The backend endpoint requires `manage_tenants` in multi-tenant mode and `manage_users` otherwise (`require_ai_status_reader`, `backend/app/modules/admin/router.py:61`).

In a multi-tenant deployment, an admin-flagged user without `manage_tenants` fires a query the backend 403s; conversely a capability-holding non-admin never fetches availability.

Fix: swap the gate to the shared `useAIStatusReader()` hook introduced by #653 (`frontend/src/hooks/use-ai-status-reader.ts`). P3 — only reachable in multi-tenant (enterprise overlay) deployments.

ishiland:
Closes #815.

`useAIAvailability` picked between the detailed `/admin/ai-status` probe and the public `/ai/availability` signal using the `isAdmin` flag, but the backend gates the admin endpoint on the capability mirrored by `useAIStatusReader()` (added in #814 for #653). The flag both over-selected (admin-flagged users without the capability fired a probe that 403s) and under-selected (capability-holding non-admins never read the detailed status).

## Change

- Branch on `useAIStatusReader()` instead of `isAdmin` in `use-ai-availability.ts`; comment pass to match.
- Test file moved to the capability-set + edition mock pattern from the #653 tests; three new regression cases cover both directions of the old mismatch.
- CHANGELOG entry under Unreleased.

## Verification

- `npx vitest run src/hooks/__tests__/use-ai-availability.test.tsx` — 16 passed
- Consumer suites (builder/viewer/dataset/maps/import/admin) — 2495 passed
- `npm run lint`, `npm run typecheck`, `make public-surface-check` — clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,7 +69,10 @@ and releases use semantic versioning.
   users who can read AI status get the detailed signal, everyone else
   with chat access uses the public one — previously the split keyed on
   the admin flag, which could probe an endpoint the backend rejects and
-  skip users it would allow. (#815)
+  skip users it would allow. Probes now also wait for the deployment
+  mode to be known instead of guessing. The disabled-state Settings
+  shortcut shows only for users who can open admin settings, and lands
+  on the AI tab instead of General. (#815)
 - **Feature popups open from the keyboard in the shared-map viewer**,
   matching the builder.
 - **A cancelled or fenced-off analysis job cleans up its orphan output
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,11 @@ and releases use semantic versioning.
   checks**, so operators who can read AI status always see the card, and
   users without that capability no longer fire requests the backend
   rejects. (#653)
+- **The builder chat availability probe follows that same capability**:
+  users who can read AI status get the detailed signal, everyone else
+  with chat access uses the public one — previously the split keyed on
+  the admin flag, which could probe an endpoint the backend rejects and
+  skip users it would allow. (#815)
 - **Feature popups open from the keyboard in the shared-map viewer**,
   matching the builder.
 - **A cancelled or fenced-off analysis job cleans up its orphan output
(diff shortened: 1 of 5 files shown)


ishiland:
Review findings on the first commit, all addressed in the follow-up commit:

1. **Edition-loading race (should-fix)** — `canReadStatus` derives from `useEdition().isMultiTenant`, which reads as single-tenant while the edition query is still loading. With a cold cache, a `manage_users` holder in a multi-tenant deployment would probe `/admin/ai-status` during that window and 403 — the exact noise this PR exists to remove, and 403s flow through `captureQueryError` (`frontend/src/main.tsx`) into error telemetry. Both probes now hold until the edition query settles, with the window surfaced as `isLoading` so consumers keep showing the spinner. Regression test added.

2. **Disabled-state Settings CTA population (should-fix)** — `AIDisabledState` in `BuilderRail.tsx` still gated its "Go to Settings" button on the `isAdmin` flag. That lined up with the old branch split but not the new one — and neither matched the target route: `/admin/settings/:tab` is gated on `manage_settings` (`AdminCapabilityRoute`, `src/App.tsx`). The CTA now shows exactly for users the route will admit.

3. **Stale CTA link (found while fixing 2)** — the button linked `/admin/settings?tab=ai`; the bare-path redirect sends that to the General tab and drops the query. Every other surface links `/admin/settings/ai`; the CTA now does too.

Also verified during review: the two new mode-aware gate tests fail against the pre-fix hook body (both directions of the original mismatch), so they genuinely pin #815 rather than just exercising the new path.

PR #838, opened 2026-07-28
feat(settings): remove the demo-mode banner in favor of the site banner

ishiland:
## What

Removes the Demo Mode banner and its `demo_mode` setting entirely. The admin-configurable Site Banner (Settings → General) covers the same need with custom text, a color choice, and per-session dismissal, so keeping both was redundant.

Removed end to end:

- `DemoBanner` component + its mount in `AppLayout` (Site Banner stays)
- the Demo Mode toggle on Settings → General
- `demo_mode` `PersistentConfig` entry and `Settings` field
- `demo_mode` field on `GET /api/auth/config` (OpenAPI + both SDKs + frontend generated types regenerated)
- `DEMO_MODE` env var: `.env.example`, both compose files, and the env-doc drift allowlist
- `demoBanner` / `settings.general.demoMode*` i18n keys in all four locales

## Impact

- **API**: `ConfigResponse` loses the optional `demo_mode` field.
- **Config**: a set `DEMO_MODE` env var is now ignored; a stored `demo_mode` settings row is orphaned harmlessly. Deployments that relied on it (the public demo VM does) should enable the Site Banner with equivalent text instead.
- CHANGELOG has a `Removed` entry under Unreleased.

## Verification

- `uv run ruff check .` / `ruff format --check .` clean
- backend: `tests/test_auth.py tests/test_api_contract_openapi.py tests/test_settings_admin.py` — 64 passed
- frontend: `npm run typecheck`, `npm run lint`, full `npx vitest run` (344/345 files; the one failure is a pre-existing local-only StackRow rename flake that also fails on clean main and passes in CI), `npm run test:i18n`
- `make openapi-check`, `make sdks-check`, `npm run types:check` — clean worktree after regen
- TS SDK `npm run build` + `npm test` — 4 pass


ishiland:
Addressed the Codex P2 in 640f643e5: `BANNER_ENABLED` / `BANNER_TEXT` / `BANNER_COLOR` are now env-backed Settings fields (same `env_default_factory` pattern as `LANDING_FIRST`), documented in `.env.example` and both compose files, and covered by the env-doc drift gate — so `ENV_ONLY_CONFIG=true` deployments keep a banner path after `DEMO_MODE` is gone.

PR #841, opened 2026-07-28
feat(builder): cluster count toggle and legible categorical layer rows

### LINKED ISSUE #839: feat(builder): "Show counts" toggle for cluster count labels
## Problem

A map with labels disabled still shows numbers on its points at clustered zooms — the cluster count text. On the demo's "Everything That Fell From the Sky" map (`label_config: null`, cluster render mode), every cluster bubble carries a white count, and the map author reasonably read that as "labels are on even though I disabled them".

This is a control-scoping gap, not a rendering bug: the Labels tab governs per-feature labels, while the count symbol layer (`{layer}-cluster-count`, `text-field: point_count_abbreviated`) is emitted unconditionally by cluster mode with no way to turn it off. The cluster editor styles the count text (size, color) but cannot hide it.

## What other tools do

- **ArcGIS Online** models cluster labels as their own explicitly-configured control (Clustering pane → Cluster label → Enable labels), independent of the bubbles. Size encoding is automatic; the count label is opt-in.
- **CARTO Builder** likewise separates cluster aggregation (drives symbol size) from labels (independent opt-in with its own controls).
- **Kepler.gl ships count-less cluster bubbles** — size/color-only encoding, count in the tooltip. Issue keplergl/kepler.gl#1151 asking for on-bubble counts has sat unresolved for years, so size-only clustering is a legitimate shipping design, not a footgun.
- **Mapbox** has no toggle only because users hand-author the three layers; deleting the count symbol layer is the toggle. A builder that generates those layers should surface the equivalent switch.

## Recommendation

Add a **"Show counts"** toggle to the cluster style editor, directly under the radius/size controls:

- Default **on** — preserves every existing map and matches the dominant convention.
- Independent of the Labels tab — do not fold it into the feature-label toggle; ArcGIS and CARTO both model it as a separate control, and "labels" already means something else in this UI.
- Off = the count symbol layer is hidden; clusters encode magnitude by size (and the color ramp) only, with the count still available in the cluster popup.

## Implementation notes

- `style_config.builder.clusterShowCounts?: boolean` (absent = true).
- `frontend/src/components/builder/layer-adapters/cluster-adapter.ts` — gate the count layer's `visibility` in `clusterCountLayout()` / `syncClusterCountLayer()` on the new flag (the layer already exists on both add and sync paths).
- Toggle UI in `frontend/src/components/builder/LayerStyleEditor/ClusterEditor.tsx` next to `clusterTextSize`/`clusterTextColor` (grey those out while counts are hidden).
- Parity: the viewer consumes the same style config; verify the shared sync path honors it, and check whether the backend style-JSON generator (`backend/app/modules/catalog/maps/style_json.py`) emits cluster count layers for embed/share surfaces — if so it needs the same gate.
- i18n: new keys in all four locales (en/es/fr/de).

## Acceptance

- Cluster map with counts off: no numbers at any zoom in builder, viewer, and embed; cluster popup still reports the count.
- Existing maps (no flag saved) render unchanged.
- Toggling on/off live updates the map without a reload.

### LINKED ISSUE #840: feat(builder): make categorical styling legible from the layer row
## Problem

A layer styled categorically is illegible from its layer-stack row. The row icon for a multi-category layer is a 14px circle filled with a linear gradient across the category colors (`getLayerColors()` → `ShapeIcon` in `frontend/src/components/map/layer-icons.tsx`) — a 2-category amber/gray layer reads as a single muddy amber dot, and nothing on the row names the styled attribute. A user looking at the demo meteorites map could not tell what the amber vs gray points meant from the Layers panel; the answer only lives in the floating Legend panel and the style editor.

Two distinct gaps:
1. The gradient swatch misrepresents discrete categories as a continuous ramp, at a size where neither is legible.
2. The row never says *which field* drives the styling, so "what do the colors mean" has no anchor.

## What other tools do

- **Desktop GIS (QGIS, ArcGIS Pro)** makes the TOC the legend: categorized layers expand into per-class swatch rows with visibility checkboxes. Works because their trees are tall and there is no separate legend panel.
- **ArcGIS Online** — the closest analog to our layout (plain layer list + separate legend) — deliberately did NOT add inline expansion. Its answer is a Legend tab inside each layer's options pane; where the ArcGIS web stack has inline layer-list legends at all (Experience Builder "Show legend", JS SDK `LayerList` ListItemPanel) they are opt-in and collapsed by default.
- **Felt / Atlas.co** keep categories in the style editor and the (separate) legend.
- **Kepler.gl** regressed its layer-panel color bar to stop showing category breakdowns — issue keplergl/kepler.gl#531 is users reporting exactly the confusion above.

Trade-offs of full inline expansion in our drag-and-drop stack: variable row heights (a 30-category layer = a 30-row subtree), child rows must be excluded from the sortable context, expansions should collapse on drag start. High cost for a sidebar that already has a legend.

## Recommendation (two phases)

**Phase 1 — fix the row symbol (cheap, addresses the actual complaint):**
- For categorical/graduated styles, replace the gradient fill with a strip of up to 3–5 **discrete** color chips ("+" overflow implied), so discrete classes look discrete.
- Add the styled field and class count to the row subtitle: e.g. `fall · 2 categories`. Naming the attribute is what answers "what do the colors mean"; the swatch alone never will.

**Phase 2 (optional, follow ArcGIS-web convention) — opt-in disclosure:**
- A chevron on categorized rows expanding to a capped list of swatch+label rows (5–6, then "+N more" that opens the style editor).
- Collapsed by default; child rows non-draggable and not drop targets; auto-collapse all expansions on drag start.

If only one lands, ship Phase 1.

## Implementation notes

- Phase 1: `layer-icons.tsx` (`getLayerColors` already returns category colors when paint is an expression; add a `discrete` rendering mode to `ShapeIcon`/`LineIcon`), `StackRow.tsx` subtitle. Data is already on the layer (`style_config.mode`, `column`, `categories`).
- The same icon component is shared with both legend surfaces (fix(#452) parity) — verify the discrete chips look right there too or scope the change to the stack row.
- Phase 2: `StackRow.tsx` + the dnd-kit sortable context (see the existing basemap row expansion for the interaction pattern).

## Acceptance (Phase 1)

- A categorical point layer's row shows discrete chips (not a gradient) and a `column · N categories` subtitle.
- Single-color and heatmap/graduated-ramp layers are unchanged.
- No row-height change; drag/drop behavior untouched.

ishiland:
## What

Two builder UX fixes from the demo meteorites map, specced in the linked issues.

**Cluster "Show counts" toggle (#839).** Cluster mode emitted its count symbol layer unconditionally, so a map with labels disabled still showed numbers on every bubble. The cluster editor now has a "Show counts" switch, default on, stored as `style_config.builder.clusterShowCounts` (absent = on). Off hides the count layer and the count color/size controls; clusters encode magnitude by size and the color ramp, and the popup still reports the count. The gate lives in the shared cluster adapter, so the viewer and embeds follow the builder. Counts turned off also registers as destructible state for the render-mode switch warning. The style.json export emits no cluster companion layers (server-side clustering does not round-trip), so nothing changes on the export path.

**Categorical layer rows (#840, Phase 1).** A categorical layer's row icon was a 14px gradient across the category colors; a two-class amber/gray layer read as one muddy dot, and nothing on the row named the styled column. The icon now renders hard-stop color bands (capped at 4) inside the same glyph, and the row gets a muted `column · N categories` subtitle rendered inline, so row height and drag behavior are unchanged. Both legend surfaces share the icon component and pick up the banded header icon; their per-category swatch rows are untouched. Graduated ramps, heatmaps, and single-color layers render as before.

## Schema and API impact

New builder key `clusterShowCounts` / `cluster_show_counts` in both alias maps (frontend `layer-adapters/shared.ts`, backend `maps/schemas.py`); parity is pinned by `test_builder_alias_keys_drift.py`. `style_config` is an open dict, so there is no API schema change and no migration.

## Verification

- `cd frontend && npm run lint && npm run typecheck && npx vitest run` (the one failure is the pre-existing StackRow rename-input flake, which fails the same way on clean main and passes in isolation and in CI)
- `npm run test:i18n` with the new keys in all four locales, including `_one`/`_other` plurals for the category count
- `cd backend && uv run pytest tests/test_builder_alias_keys_drift.py -q` plus ruff check and format
- Live check on the local stack against "Everything That Fell From the Sky": toggling Show counts hides and restores the bubble numbers without a reload, the count controls hide while off, and the layer row shows `fall · 2 categories` next to the banded icon.

Closes #839
Closes #840


ishiland:
@codex review

ishiland:
@codex review — addressed the P2: `isDiscreteColorStyle` now excludes `render_mode === 'symbol'` (58fe6175e), with a StackRow test pinning both excluded modes.

ishiland:
@codex review — round 2 P2 addressed: `applyLayerVisibilityToMap` now gates the count layer on `clusterShowCounts` like the adapter does (adb99a613), with a test covering the hide/re-show path.

PR #844, opened 2026-07-28
fix(backup): tolerate tar live-write skew instead of failing the cycle

### LINKED ISSUE #843: Backup cycle fails on tar live-write skew (exit 1), leaving the backup container unhealthy
## Problem

A backup cycle fails whenever anything writes to the object-storage staging volume while `backup_staging()` is tarring it. GNU tar exits 1 for "file changed as we read it" — a warning, with the archive fully written — but the script treats any nonzero exit as fatal: it deletes the archive, fails the cycle, never touches `.last-success`, and (since the #712 freshness healthcheck) the container goes and stays **unhealthy**.

Observed live on the demo VM minutes after the 1.6.0 redeploy: the initial on-startup cycle produced a verified 38M dump, then spent 56s tarring 1.2G of staging while the freshly restarted api/worker were writing — `tar: .: file changed as we read it` → "ERROR: object-storage archive failed" → container unhealthy past its 10m start_period. A manually re-run cycle on the settled system passed and went healthy.

On any busy install this makes backup health a coin flip: the more active the instance, the more likely every cycle overlaps a write and the healthcheck reports a permanently failing backup — while the actually-critical DB dump succeeded and verified.

## Fix

Distinguish tar's documented exit codes in `backup_staging()` (`scripts/backup-entrypoint.sh`):

- exit 0 — success, as today.
- exit 1 ("some files differ") — keep the archive, log a warning, continue the cycle. Every quiescent file in the archive is sound; files caught mid-write re-archive next cycle. This is how mainstream backup tooling treats live filesystems.
- exit ≥ 2 (fatal) — delete the partial archive and fail the cycle, as today.

## Acceptance

- A cycle whose tar hits live-write skew completes, writes `.last-success`, and the container stays healthy; the warning is visible in the container log.
- A fatally failing tar (exit ≥ 2) still fails the cycle and leaves the healthcheck unhealthy.

ishiland:
## What

A backup cycle failed whenever anything wrote to the object-storage staging volume during the tar pass. GNU tar exits 1 for "file changed as we read it" — a warning, with the archive fully written — but `backup_staging()` treated any nonzero exit as fatal: it deleted the archive, failed the cycle, never touched `.last-success`, and the #712 freshness healthcheck reported the container unhealthy. Observed live on the demo VM minutes after the 1.6.0 redeploy: the initial on-startup cycle raced the freshly restarted api/worker during a 56-second tar of 1.2G; a re-run on the settled system passed and went healthy.

Tar's exit codes are now distinguished: 0 succeeds as before; 1 keeps the archive, logs a warning ("staging changed during archiving"), and continues the cycle — files caught mid-write re-archive next cycle; 2 and above deletes the partial archive and fails, as before.

## Tests

`backend/tests/test_backup_staging_tar_skew.py` extracts the real `backup_staging` function from the script at test time (no embedded copy to drift) and drives it with a stub tar at exits 0/1/2, asserting archive kept + cycle success on 1 and archive removed + cycle failure on 2. `bash -n` clean, ruff clean.

Closes #843

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ and releases use semantic versioning.
 
 ## [Unreleased]
 
+### Fixed
+
+- **A backup cycle no longer fails just because the instance was busy.**
+  Anything writing to object storage while the backup archived it made tar
+  report "file changed as we read it", and the cycle treated that warning
+  as fatal: it deleted the (fully written) archive and left the backup
+  container unhealthy under the new freshness probe. The warning now keeps
+  the archive and the cycle; only a fatal tar error (exit 2+) fails it.
+  (#843)
+
 ## [1.6.0] - 2026-07-28
 
 ### Added
--- a/backend/tests/test_backup_staging_tar_skew.py
+++ b/backend/tests/test_backup_staging_tar_skew.py
@@ -1,0 +1,93 @@
+"""fix(#843): backup_staging must tolerate tar's live-write-skew exit (1).
+
+GNU tar exits 1 ("some files differ", e.g. "file changed as we read it") when
+api/worker write to the staging volume mid-archive; the archive is still fully
+written and every quiescent file in it is sound. Only exit >= 2 is a fatal tar
+error. Before the fix, exit 1 deleted the archive and failed the cycle, so the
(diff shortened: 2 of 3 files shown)


ishiland:
@codex review

PR #847, opened 2026-07-28
chore(builder): drop the color-relief typing workaround after the spec bump

ishiland:
Follow-up to #744 (style-spec 19.3.3 → 26.2.1): the spec union now models the `color-relief` layer natively, so the local `ColorReliefLayerSpec` interface and the `as unknown as AddLayerObject` double cast (builder-audit #338 MAINT-02) are deleted. The layer object and its `color-relief-*` paint keys typecheck against the real spec now, and `buildElevationExpression` declares `ExpressionSpecification` with one documented cast at the dynamic-push boundary.

No behavior change — typecheck, lint, and the raster map-sync + Advanced JSON validator tests all pass.


ishiland:
@codex review

PR #849, opened 2026-07-28
fix(api): strip root_path in the slowapi lazy-route fallback

ishiland:
Follow-up to the Codex P2 on #747 (posted after that PR merged).

## What

The lazy-include fallback added in #747 matched route regexes against the raw `scope["path"]`. Route regexes never include the ASGI root_path, and starlette's own matcher strips it before comparing. In rewrite-less deployments that keep the `/api` prefix via ROOT_PATH (direct-container or custom reverse proxies; the checked nginx config rewrites the prefix away), the fallback resolved no handler and slowapi treated nested routes as exempt from the global default rate limit.

The fix mirrors starlette's `get_route_path` semantics inline (avoiding the private import): strip `root_path` only when the remainder is empty or starts with `/`.

## Verification

- New regression test `test_lazy_route_handler_resolution_strips_root_path` asserts a `/api`-prefixed scope with `root_path="/api"` resolves to the same handler as the unprefixed scope.
- `uv run pytest tests/test_middleware.py` passes (16 tests, including the existing global rate-limit test).
- ruff check and format pass.

### DIFF
--- a/backend/app/api/main.py
+++ b/backend/app/api/main.py
@@ -772,7 +772,17 @@ def _find_route_handler_with_lazy_includes(routes, scope):
         return handler
     from fastapi.routing import iter_route_contexts
 
+    # Route regexes never include the ASGI root_path, and starlette's own
+    # matching strips it (starlette._utils.get_route_path — mirrored here to
+    # avoid the private import). Without this, rewrite-less deployments that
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

PR #855, opened 2026-07-29
fix(builder): keep AI availability in loading state while permissions resolve

### LINKED ISSUE #818: bug(builder): cold-permissions mount briefly shows the no-permission AI state to permitted editors
Found during the #816 review verification pass; pre-existing (the coupling dates to Phase 1135 AI-02), not introduced there.

In `useAIAvailability` (`frontend/src/hooks/use-ai-availability.ts`), `canUse = can('use_ai_chat')` defaults to false while the permissions store is still loading, which settles `reason` to `'permission'` with `isLoading=false` (the `canUse &&` clause short-circuits the loading formula). `ProtectedRoute` gates on token presence only, so on a cold page load a genuinely permitted editor can see the "You don't have permission to use AI chat" disabled state flash in BuilderRail before permissions resolve and the panel corrects itself.

Fix shape: surface a permissions-loading signal (mirroring how #816 folded the edition query's loading into `isLoading`) so the spinner covers the window.

P4 — cosmetic flash, self-corrects.

ishiland:
## Summary

Fixes the cold-mount flash where a permitted editor briefly saw the "You don't have permission to use AI chat" disabled state in BuilderRail (#818, found during the #816 review verification pass).

`useAIAvailability` derived `canUse` from `can('use_ai_chat')`, which answers `false` while the permissions query is still loading. That settled `reason='permission'` with `isLoading=false` (the `canUse &&` clause short-circuited the loading formula), so the disabled-state UI rendered the permission-denied copy before permissions resolved.

## Change

- `frontend/src/hooks/use-ai-availability.ts`: consume `usePermissions().isLoading` (already exposed by the hook), fold it into the returned `isLoading` — the same shape #816 used for the edition query — and keep `reason=null` during that window so the UI shows the spinner. `isLoading` stays false for anonymous sessions since the disabled permissions query never enters `isFetching`.
- `frontend/src/hooks/__tests__/use-ai-availability.test.tsx`: mock now mirrors the real hook (`can()` is false and `isLoading` is true while loading); new regression test asserts a cold permissions load surfaces `isLoading=true`, `reason=null`, and fires neither status probe.

No API, schema, or config impact.

Closes #818

## Verification

- `cd frontend && npx vitest run src/hooks/__tests__/use-ai-availability.test.tsx` — 18 passed
- `cd frontend && npx vitest run src/components/builder/__tests__/BuilderRail.test.tsx src/components/viewer/__tests__/ViewerChatPanel.test.tsx src/components/dataset/__tests__/DatasetChatPanel.test.tsx src/components/builder/__tests__/MapBuilderPage.notes-ai-rail.test.tsx` — 35 passed
- `cd frontend && npm run lint` — clean
- `cd frontend && npm run typecheck` — clean

### DIFF
--- a/frontend/src/hooks/__tests__/use-ai-availability.test.tsx
+++ b/frontend/src/hooks/__tests__/use-ai-availability.test.tsx
@@ -18,10 +18,15 @@ const mocks = vi.hoisted(() => ({
   capabilities: new Set<string>(),
   isMultiTenant: false,
   editionLoading: false,
+  permissionsLoading: false,
 }));
 vi.mock('@/hooks/use-permissions', () => ({
   usePermissions: () => ({
-    can: (capability: string) => mocks.capabilities.has(capability),
+    // Mirrors the real hook: permissions are null while loading, so can()
+    // answers false for everything until the query resolves.
+    can: (capability: string) =>
+      !mocks.permissionsLoading && mocks.capabilities.has(capability),
+    isLoading: mocks.permissionsLoading,
   }),
 }));
 vi.mock('@/hooks/use-edition', () => ({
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

PR #856, opened 2026-07-29
fix(frontend): wire silent error boundaries into the problem report and guard ThemeProvider storage access

### LINKED ISSUE #834: bug(frontend): error boundaries and ThemeProvider resilience — silent boundaries, storage-blocked white screen
From the 1.6.0 pre-tag audit.

- 3 of the 6 error boundaries record nothing in the problem-report buffer, so a crash behind them produces an empty report — the exact situation the reporter exists for. Wire all boundaries through the same capture path.
- ThemeProvider reads localStorage unguarded at module scope: in storage-blocked contexts (Safari private iframes, strict cookie settings) it throws before React mounts → white screen. The embed surface is exactly where third-party iframe restrictions apply. Wrap in try/catch with an in-memory fallback (the persisted zustand stores already handle this case).

ishiland:
## Summary

Two resilience fixes from the 1.6.0 pre-tag audit (#834):

**Silent error boundaries.** Three of the six error boundaries (LazyLoadErrorBoundary, RouteErrorBoundary, PluginErrorBoundary) logged crashes but never fed the problem-report buffer, so a crash behind them produced an empty problem report. They now call `pushReportEntry` with the same `source: 'react'` shape the App/Map/Panel boundaries already use. RouteErrorBoundary is a function component, so it pushes from a `useEffect` keyed on the route error.

**ThemeProvider storage-blocked white screen.** `ThemeProvider` read `localStorage` unguarded in its state initializer and wrote it unguarded in `setTheme`. In storage-blocked contexts (Safari private-mode iframes, strict cookie settings — the embed surface) that access throws before React mounts, leaving a white screen. Reads/writes are now wrapped in try/catch with an in-memory fallback, matching how `theme-init.js` and the persisted zustand stores already tolerate blocked storage.

No API, schema, or config impact.

## Files touched

- `frontend/src/components/error/LazyLoadErrorBoundary.tsx`
- `frontend/src/components/error/RouteErrorBoundary.tsx`
- `frontend/src/components/map-plugins/PluginErrorBoundary.tsx`
- `frontend/src/components/theme-provider.tsx`
- `frontend/src/components/error/__tests__/ErrorBoundaries.test.tsx` (capture-path assertions for the three fixed boundaries)
- `frontend/src/components/__tests__/theme-provider.test.tsx` (new: storage-blocked fallback tests)

## Verification

- `cd frontend && npm run lint` — pass
- `cd frontend && npm run typecheck` — pass
- `cd frontend && npx vitest run src/components/__tests__/theme-provider.test.tsx src/components/error/__tests__/ErrorBoundaries.test.tsx` — 19 tests pass
- `cd frontend && npx vitest run src/components/map-plugins/__tests__` — 20 tests pass

Closes #834

### DIFF
--- a/frontend/src/components/__tests__/theme-provider.test.tsx
+++ b/frontend/src/components/__tests__/theme-provider.test.tsx
@@ -1,0 +1,65 @@
+// fix(#834): ThemeProvider must survive storage-blocked contexts (Safari
+// private-mode iframes, strict cookie settings) where any localStorage access
+// throws — previously this crashed before React mounted (white screen on the
+// embed surface).
+import { render, screen, fireEvent } from '@testing-library/react';
+import { ThemeProvider, useTheme } from '../theme-provider';
+
+function ThemeConsumer() {
+  const { theme, setTheme } = useTheme();
+  return (
+    <div>
+      <span data-testid="theme">{theme}</span>
+      <button onClick={() => setTheme('dark')}>go dark</button>
+    </div>
+  );
+}
+
+describe('ThemeProvider storage-blocked fallback', () => {
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it('renders with the default theme when localStorage.getItem throws', () => {
+    vi.spyOn(Storage.prototype, 'getItem').mockImplementation(() => {
+      throw new Error('SecurityError: storage blocked');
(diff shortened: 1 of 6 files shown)


ishiland:
@codex review

PR #858, opened 2026-07-29
fix(frontend): a11y/UX follow-ups batch from the 1.6.0 pre-tag audit

### LINKED ISSUE #833: bug(frontend): a11y/UX follow-ups batch — focus drops, hidden AT hints, dedupe and persistence edge cases
From the 1.6.0 pre-tag audit. Grouped small frontend correctness/a11y items:

- Focus drops to `<body>` on 3 of the 4 new InlineDeleteConfirm sites after confirm/cancel (the layer-row site restores it; the others should match).
- BulkActionBar's disabled-Delete explanation is invisible to AT: the `aria-label` overrides the subtree that contains the hint text.
- The analysis "Add to map" dedupe is not shared between the toast action and the panel button — clicking both adds the layer twice.
- Group-collapse state persists only when another edit rides along in the same save (`use-builder-save.ts:459`) — collapsing a group and saving nothing else loses the collapse on reload.
- Hiding rows via the stack filter permanently prunes them from a multi-selection instead of restoring them when the filter clears.
- `geojson-z.ts:72` sends the API key as a query parameter; the header form exists two lines up — use it (query strings land in logs).
- `use-abort-signal.ts` returns a pre-aborted signal under React StrictMode's double-invoke, silently cancelling the first real request in dev.

ishiland:
Fixes the seven grouped frontend correctness/a11y items from the 1.6.0 pre-tag audit.

## Changes

1. **InlineDeleteConfirm focus drops** — the DEM editor footer, basemap-sublayer reset, and style-editor reset confirms now hand focus back to their trigger on dismiss instead of dropping it on `<body>`, matching the layer-row site's `fix(#788)` pattern (rAF where the trigger remounts).
2. **BulkActionBar disabled-Delete hint invisible to AT** — the overflow items' `aria-label` overrides the subtree as the accessible name, so the inline disabled-reason hints never reached assistive tech. The hints (Delete plus the three sibling items with the same defect) are now exposed via `aria-describedby`.
3. **Analysis "Add to map" double-add** — the completion toast action and the panel button each kept their own single-use flag, so clicking both added the layer twice. The dedupe now lives in a shared `useAnalysisAddedStore` keyed on dataset id; a new run's dataset re-enables both, and adding the same dataset via other surfaces stays legitimate.
4. **Group collapse lost on collapse-only saves** — `buildLayerDiff` stamped the live `groupMeta` on both sides, so a collapse-only change never reached the diff and `handleToggleGroupExpand` never dirtied the save. The baseline is now prepared with its own persisted collapse state and folder-group toggles mark the map unsaved. The basemap group stays UI-only (no persisted carrier), and the #805 spurious-patch fix is preserved (unchanged grouped maps still diff empty).
5. **Stack filter permanently pruned multi-selections** — selected rows hidden by the layer search are parked and rejoin the selection when the search stops hiding them. The B7 invariant stands (the bulk bar never acts on hidden rows), and explicit selection resets (Escape/outside click, drag-start, bulk actions) drop the parked rows.
6. **API key in the query string** (`geojson-z.ts`) — the key now travels in the `X-Api-Key` header, the same form the embed-token path uses; query strings land in server/proxy logs.
7. **`use-abort-signal` pre-aborted under StrictMode** — the cleanup abort is deferred one microtask so the dev-only unmount/remount no longer cancels the first real request; real unmounts still abort.

No new translation keys; no API/schema/config impact.

Closes #833

## Verification

- `cd frontend && npm run lint` — pass
- `cd frontend && npm run typecheck` — pass
- `cd frontend && npx vitest run` — 349 files, 4023 tests, all pass (post-rebase on main)
- `cd frontend && npm run test:i18n` — pass
- Targeted suites extended: `geojson-z.test.ts`, new `use-abort-signal.test.tsx`, `AnalysisJobWatcher.test.tsx`, `AnalysisPanel.test.tsx`, `use-builder-save.test.ts`, `use-builder-layers.groups.test.ts`, `BulkActionBar.test.tsx`, `DEMEditorScene.test.tsx`, `BasemapSublayerEditorScene.test.tsx`, `LayerStyleEditor.test.tsx`

### DIFF
--- a/frontend/src/api/__tests__/geojson-z.test.ts
+++ b/frontend/src/api/__tests__/geojson-z.test.ts
@@ -64,12 +64,17 @@ describe('fetchBoundedGeoJson', () => {
     );
   });
 
-  it('uses the API key query parameter for public API-key contexts', async () => {
+  // fix(#833): the API key must travel in the X-Api-Key header, never the
+  // query string (query strings land in server/proxy logs).
+  it('uses the X-Api-Key header for public API-key contexts', async () => {
     mockFetch.mockResolvedValueOnce(jsonResponse(boundedGeoJson));
 
     await fetchBoundedGeoJson('dataset-1', { apiKey: 'key 1' });
 
-    expect(mockFetch).toHaveBeenCalledWith('/api/datasets/dataset-1/features.geojson?api_key=key%201');
+    expect(mockFetch).toHaveBeenCalledWith(
+      '/api/datasets/dataset-1/features.geojson',
+      { headers: { 'X-Api-Key': 'key 1' } },
+    );
   });
 
   it('keeps fetchGeoJsonZ as a compatibility alias', async () => {
@@ -78,7 +83,10 @@ describe('fetchBoundedGeoJson', () => {
     const response = await fetchGeoJsonZ('dataset-1', { apiKey: 'key' });
 
     expect(response).toEqual(boundedGeoJson);
-    expect(mockFetch).toHaveBeenCalledWith('/api/datasets/dataset-1/features.geojson?api_key=key');
+    expect(mockFetch).toHaveBeenCalledWith(
+      '/api/datasets/dataset-1/features.geojson',
+      { headers: { 'X-Api-Key': 'key' } },
+    );
   });
 
   it('converts bounded GeoJSON responses to plain FeatureCollections for MapLibre', () => {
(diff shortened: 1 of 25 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #866, opened 2026-07-29
test(mcp): add live-contract tier and cover server.py registration

### LINKED ISSUE #827: test(mcp): server.py has no tests and CI's path filter excludes backend changes — contract drift invisible
From the 1.6.0 pre-tag audit. The MCP server's tests are mock-only (`mcp/`), `server.py` itself is untested, and the CI job's path filter doesn't include `backend/**` — so a backend API change that breaks the MCP server's read contract runs zero MCP tests.

Fix: a small live-contract test tier (spin the API, exercise each MCP tool against it) and add `backend/openapi.json` (or `backend/**`) to the MCP job's path filter.

ishiland:
## What

Adds the missing test coverage for the MCP server and closes the CI blind spot where backend contract changes ran zero MCP tests.

**New mock-tier tests** — `mcp/tests/test_server.py` (runs in `make mcp-test`, DB-less):
- FastMCP registers exactly the five advertised tools (`search_datasets`, `get_dataset_schema`, `get_features`, `list_maps`, `get_map`).
- Each tool's argument schema (required + advertised params) is pinned.
- Every tool has a non-empty description; decorated functions stay plain callables; config is only required at invocation, never at import/list time.

**New live-contract tier** — `mcp/tests/test_live_contract.py`:
- Gated by `RUN_MCP_LIVE=1` (mirrors the `RUN_AI_EVALS` pattern); skips cleanly in normal runs.
- Logs in as admin, seeds one empty "created" dataset (plus a probe feature when the `enable_dataset_editing` flag is on — tolerates the 403 when it is off) and one map, then runs every MCP tool in `server.py` against the real API and asserts response **shape** (key presence/types), not exact data. Deletes its seeds in teardown, so it is safe against a long-lived dev instance.
- `make mcp-live-test` runs it, defaulting to the dev stack (`http://localhost:8080`, admin creds overridable via `GEOLENS_INSTANCE` / `GEOLENS_ADMIN_*`).

## CI / config impacts

Scoped to the MCP pieces of `ci.yml` only:

- **Path filter**: `backend/openapi.json` added to the `mcp` filter. Note: `sdks/python/**` was already in the filter, so most contract changes (which ship regenerated SDKs) already triggered the job — the remaining gap was schema-only changes whose regenerated SDK output is byte-identical. The issue's headline claim ("path filter doesn't include backend") was therefore already partially mitigated on main; this closes the rest cheaply without pulling every `backend/**` PR into the job.
- **mcp-test job**: after the existing mock-tier step, it now copies `.env.example` (same override recipe as the e2e jobs), boots `docker compose up -d --wait api` (db + migrate + api only — no worker/frontend/titiler), and runs `make mcp-live-test` against `http://localhost:8001` (the `API_PORT` host mapping). Adds compose logs on failure, `docker compose down` on exit, and `timeout-minutes: 20` (same budget as e2e-smoke for a smaller stack). This adds a few minutes of image build/boot to the job on pushes and on PRs matching the mcp filter.
- `mcp/uv.lock`: refreshes the two local editable package versions (1.4.7 → 1.6.0) — stale from the known `make bump` lockfile miss; surfaced when uv resolved the env for the new tests.

Not changed: `require_dataset_editing_enabled` means the CI live run exercises the empty-collection branch of `get_features` (editing is off by default); the probe-feature branch runs on instances with the flag on.

## Verification

Run locally against the dev stack:

- `make mcp-test` — 18 passed, 6 skipped (live tier gated off)
- `make mcp-live-test` — 6 passed against `http://localhost:8080`, seeds cleaned up
- `uv run --project backend ruff format --check --isolated mcp/tests/...` — clean (mcp/ has no ruff gate; checked anyway)
- `backend: uv run pytest tests/test_ci_alembic_filter_paths.py tests/test_phase_279_compose_ci.py` — 10 passed (the ci.yml-introspection tests that cover the edited filter block)
- `pre-commit run --files <changed files>` — all hooks pass; `ci.yml` YAML parses

Closes #827


ishiland:
@codex review

ishiland:
Both findings addressed in 7bdd481b7:

1. **mcp filter widened** — agreed the live tier is a behavior gate, so `backend/openapi.json` alone was too narrow. The filter now triggers on `backend/**` plus the compose build/boot inputs the job actually uses (`docker-compose*.yml`, `Dockerfile*`, `.dockerignore`, `.env.example`, `db/**`, `scripts/init-db.sh`) and on `.github/workflows/ci.yml` itself (same rationale as the backend filter's #561 entry — a workflow edit alters how this job's live tier boots).
2. **Fail, don't skip** — with `RUN_MCP_LIVE=1` set, a transport failure during login now `pytest.fail`s the fixture; verified against a dead port (`GEOLENS_INSTANCE=http://127.0.0.1:9` → 6 errors, non-zero exit). The env-var-absent case remains a clean skip via the module pytestmark.

Re-verified: mock tier 18 passed / 6 skipped; live tier 6 passed against the dev stack with seeds cleaned up; backend ci.yml-introspection tests (`test_ci_alembic_filter_paths.py`, `test_phase_279_compose_ci.py`) 10 passed; pre-commit hooks and YAML parse clean.

@codex review

ishiland:
Both round-2 findings addressed in 060a2597f:

1. **Credential resolution** — `_admin_credentials()` now resolves explicit `GEOLENS_ADMIN_USERNAME`/`GEOLENS_ADMIN_PASSWORD` env vars first, then reads both keys from the repo root `.env` (where `make dev-init` writes the generated password; `install.sh` writes both keys). Username falls back to `admin` (the installer's own default) only as a last resort; an unresolvable password fails the tier with a clear message instead of a confusing 401. The round-1 fail-on-unreachable behavior is untouched. Makefile comment updated to document the resolution order.
2. **Leak-proof seeding** — created ids go into a cleanup ledger the moment each resource is created, and the `try/finally` now starts at the first POST. Teardown deletes whatever exists (404s tolerated as plain responses; only transport errors are swallowed).

Verified with a sabotage harness against the live dev stack (map POST forced to 500 mid-seed): the fixture failed with `map seed failed`, the finally block issued `DELETE /datasets/{id}` → 204, and a follow-up search/list confirmed zero leaked datasets or maps. Also exercised all three credential paths (env vars, `.env` file incl. quoted values/comments, missing password → clear failure).

Re-run: mock tier 18 passed / 6 skipped; live tier 6 passed against the dev stack; no-creds run fails loudly (6 errors, non-zero exit); pre-commit hooks and ruff format clean on touched files.

@codex review

ishiland:
Both round-3 findings addressed in 1b7b45773:

1. **Fallback sweep for commit-then-500** — confirmed against the routers: both creation endpoints `await db.commit()` before building their response, so a non-201 can still mean a committed resource. Teardown now sweeps by the unique run marker after the ledger deletes — `/search/datasets?q={marker}` and `/maps?search={marker}` — and deletes any marker-titled match the ledger missed.
2. **Loud cleanup** — every cleanup DELETE now treats only 204 and 404 (already gone) as success; any other status or transport error is collected and `pytest.fail`s at the end of teardown with the marker and per-resource details, so pollution is visible.

Verified with two sabotage harnesses against the live dev stack:
- Commit-then-500, faithfully simulated: the map POST really created the map (201 under the hood) while the fixture saw a fabricated 500 — the ledger never recorded it, the fixture failed, and the sweep found and deleted the orphan; post-run search/list confirmed zero leftovers.
- Loud cleanup: dataset DELETE intercepted to return 409 — teardown raised `live-tier cleanup could not remove seeded resources (marker …): dataset …: HTTP 409`.

Re-run: mock tier 18 passed / 6 skipped; live tier 6 passed against the dev stack (one rerun needed after my own harness logins tripped the /auth/login rate limit — 429, transient, not a product issue); ruff format and pre-commit clean on the touched file.

@codex review

ishiland:
Both round-4 findings verified and fixed in 6db25426a:

1. **P1 confirmed empirically** — the mcp-test job on the previous push was red exactly as described: `migrate-1 | FATAL: GEOLENS_ADMIN_USERNAME must not be empty` ([job log](https://github.com/geolens-io/geolens/actions/runs/30479175983/job/90668569720)). `.env.example` ships the username blank, `docker-compose.yml` interpolates `${GEOLENS_ADMIN_USERNAME}` with no default, and the step-scoped env on the test step arrives after `compose up`. The e2e jobs dodge this by exporting the credentials to `GITHUB_ENV` **before** their compose boot (shell env beats .env in compose interpolation) — my recipe copied their .env step but replaced that export with step env. Fixed by writing BOTH `GEOLENS_ADMIN_USERNAME=admin` and the CI password into `.env` in the Create .env step, before boot; sed recipe simulated against a verbatim `.env.example` copy to confirm both keys land.
2. **P2 fixed** — `Makefile` added to the mcp path filter; the job invokes the root `make mcp-test` / `make mcp-live-test` recipes, so a Makefile-only change must run it.

Re-run: ci.yml YAML parses; backend ci.yml-introspection tests (`test_ci_alembic_filter_paths.py`, `test_phase_279_compose_ci.py`) 10 passed; pre-commit clean. The real proof for the P1 is this push's own mcp-test job — it now exercises the fixed boot path end to end.

@codex review

ishiland:
@codex review

ishiland:
@codex review

Issue #441, opened 2026-07-10
Launch readiness: v1.4.3 release blocker and demo hardening items from the 2026-07-09 audit


ishiland:
The 2026-07-09 launch and demo readiness audit passed overall but left a short list of items to close before and shortly after the v1.4.3 tag.

Release blocker:

- v1.4.2 predates the five High security fixes merged on 2026-07-09 (#429, #430), and the installer resolves the highest release tag. Cut v1.4.3 from main with a curated CHANGELOG section so fresh installs get the hardened code.

Riding the release:

- Installer: the 90 second health wait times out on hosts where the database image runs emulated (the db image is amd64-only, so Apple Silicon converges slowly). Raise the budget and treat a timeout as "still starting" guidance instead of a hard failure.
- Installer: warn when Docker has less memory than the compose limits assume, instead of letting the stack OOM silently.
- Installer: the port fallbacks used when .env omits DB_PORT/API_PORT disagree with the compose defaults (5434/8001 vs 5432/8000). Latent today because .env.example ships the ports as active values.
- /api/health should report the running version and build commit so a deployment can be verified over HTTP.

Demo VM hardening (operator side, no code change):

- Set non-zero per-user quotas (MAX_DATASETS_PER_USER, MAX_STORAGE_BYTES_PER_USER). Today any SSO sign-in gets editor with unlimited storage.
- Set ENVIRONMENT=production so the OAuth session cookie carries Secure.

Deferred (tracked, not in the release):

- Sanitized public schema for anonymous /datasets/{id}/history (Medium residual S09 from the 2026-07-07 security audit).
- Footer badge click-through and the /privacy 404 belong to the frontend and marketing site work respectively.

ishiland:
CS-04 (installer port fallbacks 5434/8001 vs compose defaults 5432/8000) is deliberately NOT in the v1.4.3 release commit. The docs contract pins the installer literals to the documented ports, which are the .env.example-shipped values a fresh install actually binds, so a correct fix needs a coordinated change to docs-contract.json, its canonical source, and the docs. Latent today: .env.example always ships the port keys, so the fallback branch never runs on a normal install.

ishiland:
[...]

PR #867, opened 2026-07-29
test(backup): prove restore.sh itself in CI and gate wait_for_healthy on declared start_period

### LINKED ISSUE #826: test(backup): CI proves a replica of restore.sh, never the script; wait_for_healthy edge cases untested
From the 1.6.0 pre-tag audit.

- `ci.yml:439-514` seeds `data.ci_probe`, replicates restore.sh's flags + re-grant block, and asserts reader SELECT via `SET ROLE geolens_reader` — a solid round-trip proof, but of a *replica*. The script itself is never invoked, so drift between `restore.sh` and the CI copy is invisible. Invoke the actual script in the round-trip.
- `upgrade.sh`'s `wait_for_healthy`: the new still-starting branch is untested (the test stub returns empty), and the branch treats *every* `(health: starting)` service as converging. A service stuck in `starting` forever (bad healthcheck) would pass the wait. Exempt/track by name or compare elapsed time against the per-service `start_period`.

ishiland:
Closes #826.

Both audit sub-items re-verified on current origin/main (post #850–#864) — neither had been addressed, and the first had concrete drift to show for itself.

## 1. CI now runs the real `scripts/restore.sh`, not a replica

The bundled-mode leg of the `backup-roundtrip` job inlined a copy of restore.sh's flags and re-grant block, so drift between the script and the CI copy was invisible. It was also already real: the replica validated the dump with `pg_restore --list`, which passes a truncated `-Fc` archive, while restore.sh deliberately uses `-f /dev/null` (#710) precisely because `--list` misses truncation.

restore.sh restores via `docker compose exec -T db`, so proving the actual script needs a compose project whose `db` service is the restore target:

- New `scripts/tests/ci-restore-db.compose.yml`: a db-only stack built from `db/Dockerfile` (the shipped image — restore.sh's fail-closed pre-restore DDL creates postgis/pg_trgm/unaccent/vector and must find them installed). No host ports, no volumes.
- The CI step brings that stack up, then runs `bash scripts/restore.sh "$DUMP_FILE"` env-parameterized (`COMPOSE_FILE`, `COMPOSE_PROJECT_NAME`, `POSTGRES_USER/PASSWORD/DB` — the CI checkout ships no `.env`, so env wins). The script's exit code is the gate; its own `-f /dev/null` validation, warning-vs-error classification, and reader re-grant all now run in CI.
- Kept: hard row-count asserts (T-1247-09) and the independent `SET ROLE geolens_reader; SELECT ... FROM data.ci_probe` assertion — now proving the *effect* of the script's re-grant rather than a mirrored copy of it.
- The seeding, backup-image boot, SigV4 upload, and the managed-mode step are untouched (managed mode documents why restore.sh's compose-exec path does not apply to an external DB).

## 2. `wait_for_healthy` verifies the tolerance claim it used to assume

`scripts/lib/common.sh` treated every `(health: starting)` service at budget end as converging, so a service stuck in `starting` (broken healthcheck) passed the wait. The pass condition is now checked per service against the full tolerance Docker actually grants: `start_period + timeout + retries × (interval + timeout)` — the extra `timeout` because moby grace-ignores a probe by its START time, so one launched just inside start_period runs up to `timeout` past the boundary before retry counting begins (Codex P2 round 3) — read in one `docker inspect` alongside `.State.StartedAt` (zero values substitute the daemon defaults — interval/timeout 30s, retries 3). The window is compared against the container's actual age (`now - StartedAt`, GNU date with a BSD fallback) — not the spent budget, which overstates the age of a restart-policy service that crashed and restarted mid-wait with a freshly reset health clock (Codex P2 round 2). A straggler whose age exceeds its whole tolerance window has outlived every verdict Docker could still be working on and fails the wait with a named diagnostic plus the rollback recipe. The tolerance math only applies while the live `.State.Health.Status` (read in the same inspect) still says `starting` — a container that flipped to `unhealthy` between the `compose ps` snapshot and the inspect fails outright, and one that flipped to `healthy` counts as converged (Codex P2 round 4).

The boundary is deliberately NOT bare start_period (Codex P2 review round): after start_period ends, failing probes only accumulate toward `retries` consecutive failures and the container honestly stays `starting` until the streak is exhausted — e.g. the prod frontend (start_period 15s, interval 30s, timeout 10s, retries 3) can legitimately report `starting` at the 90s budget after two post-grace failures, and must pass. A straggler still inside its window (e.g. the prod backup service's 10m first-dump grace) warns and passes exactly as before. An unreadable healthcheck config, unparseable `StartedAt`, or unreadable live status fails open, matching the script's other best-effort probes.

`scripts/tests/test-upgrade-order.sh` (the existing stub harness — its docker stub answered an empty, instantly-healthy `ps --format`, which is why every branch of the wait was unreachable) now drives the gate via a `DOCKER_PS_STATUS` table, per-service `DOCKER_HC_<svc>` healthcheck tuples, and a no-op `sleep` shim. New cases: in-tolerance straggler passes with the warning; mid-retry-streak service past bare start_period but inside its tolerance passes (the Codex P2 regression case); stuck-past-tolerance straggler fails with the rollback recipe; service restarted mid-wait (young age, small tolerance) passes (the round-2 regression case); container aged inside the final in-flight-probe timeout margin passes (the round-3 regression case); snapshot-vs-inspect race pins both directions (flipped unhealthy fails despite in-tolerance age, flipped healthy passes despite over-tolerance age — the round-4 regression cases); mixed case fails on the overdue service; unreadable config or garbage StartedAt fails open; `(unhealthy)` at budget end fails (the pre-existing broken branch, also previously untested). 46/46 assertions pass.

## CI / config impact

- `.github/workflows/ci.yml`: edits confined to the `backup-roundtrip` job's BUNDLED MODE step, the job header comment, and two added lines in the `backup` path filter (`scripts/tests/ci-restore-db.compose.yml`, `db/**` — the db image is now part of the tested surface).
- Behavior change to `scripts/upgrade.sh` runs (via `scripts/lib/common.sh`): an upgrade whose stack has a service stuck in `starting` past its full healthcheck tolerance now fails the health gate instead of passing. `scripts/install.sh`'s inlined copy is deliberately untouched — its timeout is already non-fatal by design (rc=2, #441 first-boot-under-QEMU contract).

## Verification

- `sh scripts/tests/test-upgrade-order.sh` — 46 passed, 0 failed (32 pre-existing + 14 new).
- `sh -n scripts/lib/common.sh scripts/upgrade.sh scripts/tests/test-upgrade-order.sh`; `bash -n scripts/restore.sh scripts/tests/test-backup-restore-roundtrip.sh`; `bash -n` on the extracted BUNDLED MODE run block. (shellcheck is not configured in this repo.)
- `docker compose -f scripts/tests/ci-restore-db.compose.yml config -q`.
- `pre-commit run --files <touched>` — all hooks pass.
- Local proof of the CI invocation shape: ran the unmodified `scripts/restore.sh` env-parameterized (`COMPOSE_FILE` + `COMPOSE_PROJECT_NAME` + scratch `POSTGRES_DB`) against the running dev stack's `db` service — seeded 113/57 rows + `data.ci_probe`, `pg_dump -Fc --no-owner --no-acl`, restored into the scratch DB via the script. Row counts matched and `SET ROLE geolens_reader` selected the probe row. Scratch DBs dropped; dev stack healthy afterwards.
- Verified the inspect template against a live daemon: `'{{.Config.Healthcheck.StartPeriod.Seconds}} {{.Config.Healthcheck.Interval.Seconds}} {{.Config.Healthcheck.Timeout.Seconds}} {{.Config.Healthcheck.Retries}}'` prints four plain integers (e.g. `30 30 5 3`); a container without a healthcheck errors to stderr with empty stdout (the fail-open path).


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
Round-2 P2 addressed in 91e485f04 — correct call on the restart-policy hole. The overdue check now compares the tolerance window against the container's actual age (`now - .State.StartedAt`, read in the same `docker inspect` as the healthcheck config) instead of the spent budget, so a service that crashed and restarted mid-wait with a freshly reset health clock is judged by its own age, not the wait's. `iso_to_epoch` parses with GNU date first and BSD `date -j -u -f` as the fallback, with a leading-year guard so GNU's `date -d ""` midnight quirk can't turn an empty read into a wrong age; unparseable `StartedAt` fails open, same as unreadable healthcheck config. Harness: the inspect stub now prepends a per-service `StartedAt`; new scenarios cover the restarted-mid-wait service (age 5s, tolerance 40s, budget 90s → passes) and garbage `StartedAt` (fails open), while the genuinely stuck case (age 120s past its 40s tolerance) still fails. 43/43 assertions pass.

ishiland:
@codex review

ishiland:
Round-3 P2 addressed in 0e4e1a003 — agreed, the moby grace rule is judged by probe start time, so one in-flight probe can carry `timeout` past the start_period boundary before the counted retry cycles begin, and the previous bound was one `timeout` short in the worst case. The allowance is now `start_period + timeout + retries x (interval + timeout)`, with the probe-start-time rule cited in the comment. Boundary tests shifted accordingly (15/30/10/3: 135s -> 145s; 10/5/5/3 stuck case: 40s -> 45s), and a new scenario pins the previously-misclassified window — a container aged 140s under 15/30/10/3 (past the old bound, inside the true one) now PASSES, while the genuinely stuck case (120s vs 45s) still fails. 44/44 assertions pass.

ishiland:
@codex review

ishiland:
@codex review

ishiland:
Round-4 P2 addressed in 089dc7e9f — agreed, the `compose ps` table is a snapshot and a probe completing before the per-service inspect could flip the container out of `starting`, letting the tolerance math pass an already-unhealthy service. The same `docker inspect` now also reads the live `.State.Health.Status` (sixth field) and the tolerance calculation applies only while it still says `starting`: a flip to `unhealthy` fails the service outright with a named diagnostic (`reported (health: starting) in the status snapshot but is (unhealthy) on inspection`), a flip to `healthy` counts as converged, and an unreadable status fails open like the rest. Harness tuples carry the live-status field and two new scenarios pin the race in both directions — flipped-unhealthy fails despite an in-tolerance age (5s vs 45s), flipped-healthy passes despite an over-tolerance age (120s vs 45s). 46/46 assertions pass.

ishiland:
@codex review

ishiland:
@codex review

Issue #845, opened 2026-07-28
Globe projection as a per-map option


ishiland:
## What

Offer globe projection as a per-map option. MapLibre GL 5.x (which we already ship — 5.24) renders `projection: {type: 'globe'}` natively; the style-spec side (root `projection` property, `vertical-perspective`, expression syntax for `projection.type`) is fully modeled in spec 20+, which we now vendor via #744.

Global datasets are where GeoLens demos best — the meteorites showcase map on a globe is an obvious win, and it is a root-style property away.

## Scope

- Projection is **root style state, not layer state**: persist it as map-level config (alongside the basemap/terrain choices), defaulting to `mercator` so every existing map is untouched.
- Builder: a toggle in map settings (mirror the 3D terrain toggle placement); viewer + `/m/:token` respect the saved value.
- Verify interplay before shipping: terrain-on-globe behavior, cluster/heatmap rendering at low zooms, the scale control, and export/thumbnail capture (static captures of a globe view).
- style.json export: emit the root `projection` property; it round-trips cleanly since it is spec-modeled.

## Non-goals

Zoom-dependent projection expressions (`vertical-perspective` blends) — start with a plain toggle.

PR #869, opened 2026-07-29
chore(frontend): delete dead exports; consolidate drifted map-component helpers

### LINKED ISSUE #835: chore(frontend): delete dead exports; consolidate drifted map-component helpers
From the 1.6.0 pre-tag audit (grep-verified, flagged by two passes).

Dead exports to delete:
- records/datasets mutation API functions with no callers
- `useGenerateMap`
- `MetadataField.tsx`, `QualityBadge.tsx`
- `getSourceId` — a live footgun, not just dead weight: it returns the wrong id for deduped layers, so any future caller inherits a bug.

Map-component drift to consolidate (BuilderMap / ViewerMap / DatasetMap):
- The style-fetch retry fix exists in one map, not three.
- Attribution handling and the `styleimagemissing` allowlist diverge.
- Three separate tile-URL classifiers (`isThirdPartyUrl` variants).
- Three `setTransformRequest` implementations — extract a shared `buildTileTransformRequest({ embedToken? })` (the viewer's missing raster Bearer, fixed in #819, was this drift biting).
- The relative-time formatter is written out 4×.

ishiland:
From the 1.6.0 pre-tag audit. Every claim was re-verified against current `origin/main` (post #848–#864) before acting — see the per-item notes below, including two audit claims that no longer hold as written.

## Dead exports deleted (all re-verified callerless today)

| Export | Location | Note |
|---|---|---|
| `updateContact`, `createDistribution`, `updateDistribution`, `deleteDistribution` (+ local `DistributionCreate` interface) | `api/records.ts` | Zero callers; the UI has no distribution editor. `listDistributions` kept (live via `useDistributions`). |
| `updatePublicationStatus` | `api/datasets.ts` | Only "caller" was the itself-unused `useUpdatePublicationStatus` hook (also deleted) plus its contract-test rows (removed). The app walks status through `setTargetStatus`. |
| `useGenerateMap` | `hooks/use-maps.ts` | Also deleted its only dependency `generateMap` in `api/maps.ts` — the app generates maps via `streamGenerateMap`. |
| `MetadataField.tsx`, `QualityBadge.tsx` | `components/dataset/`, `components/search/` | No importers. |
| `getSourceId` | `components/builder/map-sync.ts` | The audit's "live footgun" claim holds: it returned the raw per-layer id, which is wrong for SF-04-deduped vector sources. A tombstone comment points future readers at `getSourceIdForLayer`. |

All other `api/records.ts` / `api/datasets.ts` functions were checked and have live callers — nothing else deleted.

## Map-component consolidation (BuilderMap / ViewerMap / DatasetMap)

Behavior-preserving per component; intentional differences are parameterized, not flattened. Which copy won, per item:

**Tile-URL classifier → `isThirdPartyTileUrl` (`lib/tile-utils.ts`).** ViewerMap's inline classifier won verbatim — it is the only one that handles protocol-relative URLs and the configured tile-CDN origin (`resolveTileBaseUrl`). The other "variants" on current main were the crude `absUrl.includes('/raster-tiles/')`-only checks inside the Builder/Dataset transformRequests (now gone, see next item). `lib/map-error-log.ts`'s `isFirstPartyTileUrl` was deliberately left alone: it answers a different question (is this 401/403 a *handled* first-party **vector**-tile auth error — it excludes raster on purpose and detects CDN URLs by `sig=`), and folding it in would change error-suppression semantics.

**`setTransformRequest` → `buildTileTransformRequest({ embedToken?, getTileConfig? })` (`lib/tile-utils.ts`).** ViewerMap's implementation won: it is the superset (embed token + the raster Bearer restored by #819) and the only one that gates credentials on first-party origins. Each map keeps its own wiring via `onLoad` + `map.setTransformRequest()` (the `transformRequest` prop is ignored by react-maplibre v8; see `patterns.md`). The #845 frozen-`projection` prop and the token-gated `style.load` restore paths were not touched.

**Documented behavior change (deliberate tightening, flagged in internal review):** pre-PR, BuilderMap and DatasetMap attached the raster `Authorization: Bearer` to **any** URL containing `/raster-tiles/`, unconditionally. The shared builder now gates it on `!isThirdPartyTileUrl(url, tileConfig)` — the #819 viewer semantics. A deployment whose raster tile URL pointed at an origin outside the app origin / `cdn_base_url` would previously have leaked the session JWT Bearer to that third party; now it will not. Such URLs do not occur in practice — raster tile URLs derive from tileConfig and are absolutified against `window.location.origin` by raster-adapter/use-map-layers — so for every URL these surfaces actually produce this is a no-op, but it is a real semantic change and is pinned by test (`never sends the Bearer header to a third-party host, even on a /raster-tiles/ path`).

**`styleimagemissing` → `makeStyleImageMissingHandler(map, { knownImagesOnly })` (`lib/basemap-utils.ts`).** Both behaviors survive because they are both correct for their surface: the builder keeps the allowlist (`knownImagesOnly: true`) so an editor authoring a bad icon reference still sees the warning; viewer/dataset keep stub-everything (`false`) so public/preview consoles stay clean (their old comments cite `circle_11_black`, which is *not* in the allowlist — switching them to the allowlist would have regressed that).

**Style-fetch → `useRemoteBasemapStyle` (`components/map/hooks/`).** Audit claim adjusted: on current main **no map has an explicit style-fetch retry** — what exists is a duplicated ~60-line fetch+sanitize effect in BuilderMap and ViewerMap whose *failure* paths diverge deliberately (builder keeps the placeholder + notice per Phase 1051 WR-06; viewer falls back to handing MapLibre the raw URL, which is its implicit second attempt). The duplication is now one hook; the divergence rides `fallbackToRawUrlOnError` + callbacks, so a future retry fix lands in both maps at once. DatasetMap has nothing to consolidate here: it never fetches style JSON in app code (imperative `setStyle` + `transformStyle` lets MapLibre fetch).

**Attribution.** BuilderMap/DatasetMap won: they pass the basemap entry's configured attribution into `toMaplibreStyle`; ViewerMap dropped it. The viewer — the public surface where attribution matters most — now passes it too. This is the one deliberate user-visible change in the PR: raster XYZ basemaps with configured attribution now show it in the viewer's AttributionControl.

**Relative-time formatter (was 4×) → `lib/relative-time.ts`.** The `provenance-attribution` Intl implementation won as the core (broadest unit range); `ReportEntryList`'s divergences are parameterized (`maxUnit: 'hour'`, `exactSeconds: true`) and its try/catch → `''` hardening moved into the core. The two byte-identical ApiKey-section copies collapse into `formatLastUsedRelativeTime` (their floor-based i18n-key semantics kept as-is, documented as deliberate). `provenance-attribution.ts` re-exports `formatRelativeTime` so existing importers (`quality-freshness`) are untouched.

## Audit claims that did not survive re-verification

- "The style-fetch retry fix exists in one map, not three" — no explicit retry exists anywhere on current main; treated as the duplicated-fetch-effect consolidation above.
- "Three separate `isThirdPartyUrl` variants" — two structured variants exist today (ViewerMap inline, `map-error-log.ts`) plus the crude substring checks; consolidated as described, with `map-error-log.ts` intentionally left as a purpose-specific matcher.

## Impacts

- No schema/API/config changes.
- i18n: the orphaned `search:qualityBadge.*` keys (dead with the deleted QualityBadge component) are removed from all four locales; `npm run test:i18n` stays green. The orphaned `builder:mapCreate.generateFailed` key is left in place (present in all four locales, parity unaffected; belongs to a broader locale-cleanup pass).
- Two behavior changes, both deliberate and documented above: viewer attribution now shown for raster XYZ basemaps, and the raster Bearer on Builder/Dataset is now first-party-gated (JWT no longer sent to a hypothetical third-party `/raster-tiles/` URL).

## Verification

```
cd frontend && npm ci
npm run build        # pass
npm run lint         # pass
npm run typecheck    # pass
npm run test:coverage  # 351 files, 4068 tests, all pass
npm run test:i18n    # pass
```

New unit tests: `lib/__tests__/tile-utils.test.ts` (classifier + transformRequest builder, incl. the #819 raster-Bearer regression class, the #394 embed-token first-party gate, and the embed-token-vs-Bearer precedence on first-party raster URLs), `lib/__tests__/relative-time.test.ts`, `makeStyleImageMissingHandler` cases in `lib/__tests__/basemap-utils.test.ts`, and `components/map/hooks/__tests__/use-remote-basemap-style.test.ts` (placeholder → sanitized flow, viewer fallback vs builder notice). Playwright e2e not run (no stack in this environment); the existing Builder/Viewer/DatasetMap unit suites all pass unchanged.

Review follow-up commit re-ran: targeted vitest on `tile-utils.test.ts` (39 pass), `npm run test:i18n` (pass), `npm run lint` (pass), `npm run typecheck` (pass).

Closes #835


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #865, opened 2026-07-29
docs(ops): post-1.6.0 backup propagation — EGRESS rows, RUNBOOK error markers and --wait note

ishiland:
Post-1.6.0 release propagation for the backup service, per the #837 checklist. This PR covers the geolens-repo portion; the docs-site and chart-repo portions are committed locally on sibling branches (listed below) and await push after review.

## This repo

**EGRESS.md**: the egress matrix was missing three opt-in egress paths that exist in the product. Added rows for the backup offsite S3 upload (`BACKUP_S3_ENABLED` plus the shared `S3_*` vars), SMTP email notifications, and webhook notifications, all default-off, plus matching air-gap checklist lines.

**RUNBOOK.md**:
- Log markers now include the dump-verification failure (`ERROR: <filename> failed verification (pg_restore could not read it)`) and the retention boot refusal (`ERROR: BACKUP_RETENTION_DAILY='...' must be >= 1`, with the `is not a plain integer` variant). Both strings verified against `scripts/backup-entrypoint.sh`. Added a catch-all `grep 'ERROR:'` recipe and a note on the restart-loop symptom of a retention misconfiguration.
- Section 6 step 5: `docker compose up -d --wait` can exit non-zero while the backup service is still inside its 10-minute `start_period` (first dump of a large freshly restored database). Added a note stating the principle: a `health: starting` service is converging, not failed, the same rule `scripts/upgrade.sh`'s `wait_for_healthy` applies. Phrased against the principle rather than the implementation, since #826 may touch `wait_for_healthy` edge cases concurrently.

The issue's third RUNBOOK claim was partly stale: `ERROR: object-storage archive failed` was already in the log-markers table.

## Sibling-repo commits (local, not pushed)

- **getgeolens.com** `docs/backup-160-propagation` @ `4fd3786`: `backups.mdx` no longer tells operators to set the removed-for-backup `S3_ALLOW_HTTP` (the uploader follows the `S3_ENDPOINT` scheme; the knob only affects the app's own storage client), and both backup pages gain `BACKUP_MAX_AGE_MINUTES`, the freshness healthcheck, the retention >= 1 boot refusal, and the corrected `restore.sh` validation method (`pg_restore -f /dev/null`, not `--list`).
- **geolens-deployments** `docs/backup-scope-note` @ `598719e`: README "Backups" section stating the chart ships no backup workload. No chart version bump needed (README is outside the packaged `helm/geolens`).

## Chart findings (all four #837 chart items N/A)

[...]

PR #872, opened 2026-07-29
fix(tiles): convert cluster radius px to extent units and anchor the grid

### LINKED ISSUE #868: fix(tiles): server cluster grid treats cluster_radius px as MVT extent units — overlapping clusters + singles leak
Reported from the public demo meteorites map ("clusters on top of clusters").

**Symptom** (any dataset on the server-cluster path, i.e. >5000 features): cluster circles overlap heavily, and thousands of unclustered points render at low zoom despite `cluster_max_zoom`. Most clusters stay tiny, so the count ramp/labels barely engage.

**Root cause**: `_build_cluster_tile_query` in `backend/app/processing/tiles/service.py` buckets on `bounds.width * $6 / 4096.0`, treating `cluster_radius` (screen pixels — it mirrors MapLibre's `clusterRadius` on the client path; router default `Query(48)`) as MVT extent units. 44 extent units is ~5.5 CSS px at 512px tile display, roughly 8x finer than intended. Cluster centers pack tighter than the drawn circle radius (overlap), and any point alone in a ~5px cell is emitted as an unclustered single (leak).

**Evidence** (local meteorites data, one z2 tile, the query's bucketing CTE verbatim):
- radius 44 as shipped: 187 clusters + 568 unclustered singles, biggest cluster 801
- radius 352 (= 44px x 4096/512): 94 well-spaced clusters + 15 singles, biggest 2826

**Fix**
- Convert px to extent units inside the SQL (x8 for 512px tile display); keep the endpoint's px semantics and update the docstring.
- Anchor the bucket grid to absolute EPSG:3857 coordinates instead of each tile's minx/miny so grids align across tile borders (secondary seam artifact).
- The Valkey cluster-tile cache key (`_cluster_cache_table_key`, `tiles/router.py`) carries no semantic version — version it so a deploy with changed cluster SQL doesn't serve stale-geometry tiles.
- SQL-level regression test: two points N px apart cluster (or don't) as the px semantics dictate.

Demo note: demo.getgeolens.com picks this up at the next release/redeploy; the edge tile cache may need a purge then.

ishiland:
## What

Server-side cluster tiles (`/tiles/clusters/...`, the path any point dataset over 5000 features takes) built their bucket grid from `cluster_radius` as if the value were MVT extent units. It is CSS pixels: the router default `Query(48)` mirrors MapLibre's `clusterRadius` on the client-side path. At 512 px tile display one pixel is 4096/512 = 8 extent units, so the shipped grid was ~8x finer than intended. Cluster circles overlapped, and any point alone in a ~5.5 px cell leaked through as an unclustered single at low zoom.

Numbers from one z2 tile of the local meteorites dataset (the query's bucketing CTE verbatim):

| | clusters | singles | largest cluster |
|---|---|---|---|
| radius 44 as shipped | 187 | 568 | 801 |
| radius 352 (44 x 8) | 94 | 15 | 2826 |

## Changes

- `backend/app/processing/tiles/service.py`: scale `$6` (CSS px) by `_CLUSTER_PX_TO_EXTENT_UNITS` (4096/512, with the 512 px display assumption documented at the constant) inside `_build_cluster_tile_query`. The bucket grid now anchors to absolute EPSG:3857 coordinates instead of each tile's minx/miny, so adjacent tiles at one zoom share an aligned grid (bucket size at a fixed zoom is identical for every tile). The `GREATEST(..., 1.0)` degenerate-size guard stays. Docstrings now say CSS/screen px explicitly.
- `backend/app/processing/tiles/router.py`: **cache-key version bump.** `_cluster_cache_table_key` gains a SQL-semantics version component (`...:cluster:v2:r{r}:z{z}`), with a comment telling future editors to bump it when cluster SQL semantics change. Without it this deploy would keep serving stale-geometry cluster tiles from Valkey until TTL expiry. Old `:r{r}:z{z}` keys become garbage and age out on their TTL.
- `backend/tests/test_tiles.py`: SQL-level regression test. Two points ~30 CSS px apart at z0 with radius 48 land in ONE cluster (the pre-fix math split them; the test pins that arithmetic too), and a lone point comes through as an unclustered single carrying its projected attributes. It runs the shipped query's CTEs verbatim with only the final `ST_AsMVT` encode swapped for a row SELECT, since the suite carries no MVT decoder. The existing cache-key test now expects `:v2`.
- `backend/tests/test_mvt_audit_fixes.py`: query-shape tests pinning the px conversion factor and the absolute-grid anchoring.
- `backend/tests/test_layering.py`: the tiles router LOC ratchet moves 2043 -> 2046 with a written carve-out for the 3-line cache-key comment (the ratchet requires exact LOC).

No API change: the endpoint keeps px semantics and `Query(48, ge=1, le=256)`. Frontend untouched; the client-side bounded-geojson cluster path was already correct.

Cross-tile grid alignment is asserted structurally (bucketing uses absolute `ST_X`/`ST_Y`, no `bounds.minx`/`miny` subtraction). A semantic two-tile assertion would need one point to be a candidate in two tiles at once, which the per-tile bounds filter rules out, so it was skipped as disproportionate machinery.

## Review round 1 (codex P2, commit caf975491)

Anchoring the grid globally introduced a new seam bug: a bucket cell straddling a tile border was computed twice, each tile seeing only its own fragment (candidates were filtered to the tile envelope), so opposite-side points emitted as separate singles or undercounted clusters. Fixed by:

- expanding the candidate scan by one full bucket, clamped to the world envelope before the 4326 transform (past the world edge proj wraps longitude and the transformed bbox inverts, dropping candidates), so each tile sees the complete membership of every cell overlapping it;
- `ORDER BY gid` before the input cap, so neighboring tiles agree on a straddling cell's membership and centroid when the cap truncates;
- emitting each cell from exactly one tile: the one whose envelope contains the cell centroid (half-open bounds). Singles that only enter via the expanded ring drop out through the same filter, since a single's centroid is its own point.

Perf trade: the candidate scan now covers up to one bucket beyond the tile on each side (9.375% of the tile width per side at radius 48), and the deterministic ordering sorts the candidate set before the `LIMIT`. Cap semantics are unchanged.

New regression test: two points within one bucket width on opposite sides of a z2 tile border come out as one cluster (`point_count` 2) from the owning tile only; the neighbor emits nothing for that cell, and its interior point never leaks into the adjacent tile despite the expanded scan.

## Review round 2 (two codex P2s, commit 2b5cd0d9b)

**Cap saturation (service.py).** Round 1's centroid ownership was our design call, and codex is right that it breaks under cap saturation: when the candidate cap truncates different subsets in neighboring tiles (their scan envelopes differ), each tile's computed centroid for a shared cell can land on either side of the border, giving a double- or zero-emit. Preserving complete shared-cell membership under any finite cap is impossible, so instead ownership no longer depends on the data: the anchor is now the cell ORIGIN (bucket index times bucket size; the point itself past cluster max zoom), pure grid geometry that both neighbors always agree on. Worst case under cap saturation degrades to an undercounted (or missing) cluster in the owner tile, the same degradation class any capped per-tile grid has, never a cross-tile disagreement. Consequence handled: the owner-emitted rendered point can now sit up to one bucket outside the tile, so the ST_AsMVTGeom buffer scales with the radius (`GREATEST(256, ceil(px * 8) + 16)`), covering the `le=256` px extreme (2064 extent units).

Documented limit: cap-saturation undercount is inherent to a capped per-tile grid and predates this PR; asks to "fix" the undercount itself will get a documented-limits decline.

**World edge (service.py).** The strict upper bounds dropped a feature sitting exactly on the world's east/north boundary in the outermost tile, where no neighbor exists to own it (a regression vs pre-PR, and the same class exists for anchors). The east (`x = 2^z - 1`) and north (`y = 0`) edge tiles now use inclusive upper bounds; all internal borders stay half-open.

New tests: a point at lon 180 (which transforms to exactly the world XMax on PostGIS/proj, verified against the test DB) is emitted by exactly one tile, the edge tile; the round-1 seam test passes unchanged under anchor ownership (the cell origin lands on the same side its centroid did); a cap-divergence test injects a candidate cap of 2 so the neighbors see different subsets and shows the shared cell is emitted at most once, as an undercounted single from the owner tile.

## Review round 3 (two codex P2s, commit b21e48a23)

**West-edge vanish (service.py).** The grid was anchored at 0, so the origin of a cell straddling the world's west/south edge landed outside the world; the ownership lower bound rejected it in every tile and points near lon -180 vanished at clustering zooms. Fixed by construction: the grid now anchors at the world minimum (`bucket index = floor((x - world_min) / bucket)`, `anchor = world_min + index * bucket`), so any in-world point yields an in-world cell origin and the inclusive lower bounds cover the west/south edges with no special case. Re-derived the east/north inclusive upper bounds: still needed, since an anchor can land exactly on the world maximum (a point at lon 180 past max zoom, or a bucket size that divides the world width exactly, e.g. radius 64).

**Past-max-zoom ring starvation (service.py).** Past cluster max zoom nothing clusters, but the scan still used the expanded envelope; lower-gid neighbor rows could consume the entire candidate cap and then all be discarded as ring-only, leaving the tile empty - a failure mode the pre-PR tile-bounded scan never had. The expansion is now zero when `z > cluster_max_zoom` (no cross-tile cells exist there); clustering zooms keep the one-bucket expansion.

New tests: a point 1 km from lon -180 at a clustering zoom is emitted by the west edge tile (it vanished under the 0-anchored grid); a past-max-zoom scenario with three lower-gid neighbor points and an injected cap of 2 keeps the tile's own point (the expanded scan starved it); the seam, cap-divergence, and world-edge tests were re-derived for the world-min grid; query-shape assertions pin the new anchor expressions and the conditional expansion.

## Review round 4 (two codex P2s, commit 985ef1ad3)

**Visibility hole (service.py).** Anchor ownership lets a cell's centroid fall up to one bucket inside the NEIGHBOR tile; a client whose viewport covers the centroid but not the owner tile requests only the neighbor tile, where the cell is filtered out, so the cluster was invisible there. Codex is right that the MVT buffer cannot fix this (the client never requests the owner tile). Keeping anchor ownership (the cap-stable choice), the emitted geometry is now clamped into the owning tile, one MVT extent unit inside each edge. Maximum positional error from the clamp: one bucket, i.e. the cluster radius in CSS px (48 by default), deterministic, and zero for geometry already inside the tile - all past-max-zoom points and every cell whose centroid is in-tile. The round-2 radius-scaled MVT buffer is superseded and reverted to the plain 256, since emitted geometry is now always in-tile.

**Cap-before-ownership (service.py).** The feature `LIMIT` applied before the ownership predicate, so neighbor-owned cells seen through the expanded scan could consume the output budget and displace this tile's own cells. Ownership is per-row grid arithmetic (every member of a cell shares the cell's anchor), so it now filters in the aggregation CTE before `GROUP BY` and the cap. Unclustered/past-max-zoom semantics unchanged (buckets are per-point there; the anchor is the point).

New tests: a straddling cell whose raw centroid lands in the neighbor is emitted by the owner with its geometry inside the owner tile's extent (tile-space x clamped to just inside the east edge, asserted via `ST_X` on the MVT geometry) while the neighbor emits nothing; a feature-cap variant (injected cap of 1, three neighbor-owned cells plus one owned cell) keeps the tile's own cell.

## Verification (host, Postgres at localhost:5434)

- `uv run pytest tests/test_tiles.py tests/test_mvt_audit_fixes.py tests/test_dp02_read_path_binding.py` - 63 passed
- `uv run pytest tests/test_tile_gzip_offload_perf005.py tests/test_vector_tile_auth.py tests/test_dp_review_fixes.py tests/test_phase_275_demo_cluster.py` - 21 passed
- `uv run pytest tests -n 4` - full suite: 5922 passed, 81 skipped, 2 failed on the tiles-router LOC ratchet, which this change tripped; fixed via the carve-out above and re-verified with `uv run pytest tests/test_layering.py tests/test_tiles.py tests/test_mvt_audit_fixes.py` (76 passed)
- After the round-1 seam fix: `uv run pytest tests/test_tiles.py tests/test_mvt_audit_fixes.py tests/test_dp02_read_path_binding.py tests/test_layering.py` - 93 passed; adjacent tile files re-run - 21 passed
- After the round-2 anchor/edge fix: same targeted set - 95 passed; adjacent tile files - 21 passed
- After the round-3 world-min/starvation fix: same targeted set - 97 passed; adjacent tile files - 21 passed; rebased on origin/main (#869)
- After the round-4 clamp/cap-order fix: same targeted set - 99 passed; adjacent tile files - 21 passed; rebased on origin/main (#870, #865)
- `uv run ruff check .` and `uv run ruff format --check .` - clean after each round

## Notes

- Deploy note from the issue: demo.getgeolens.com picks this up at the next release/redeploy; the edge tile cache may need a purge then.
- Possible follow-up, deliberately not touched here: `expansion_zoom` is `LEAST($5+1, 22)` for every cluster, so clicking a z2 cluster targets cluster_max_zoom+1 (z15 by default) instead of the zoom where that cluster actually splits.

Closes #868

### DIFF
--- a/backend/app/processing/tiles/router.py
+++ b/backend/app/processing/tiles/router.py
@@ -1588,7 +1588,10 @@ def _ensure_clusterable_dataset(meta: _DatasetMeta) -> None:
 def _cluster_cache_table_key(
     table_name: str, *, cluster_radius: int, cluster_max_zoom: int
 ) -> str:
-    return f"{table_name}:cluster:r{cluster_radius}:z{cluster_max_zoom}"
+    # fix(#868): "v2" versions the cluster SQL semantics. Bump it whenever
+    # _build_cluster_tile_query changes the emitted tile geometry/properties,
+    # or a deploy keeps serving stale-geometry cluster tiles until TTL expiry.
+    return f"{table_name}:cluster:v2:r{cluster_radius}:z{cluster_max_zoom}"
 
 
 async def _acquire_and_serve_tile(
--- a/backend/app/processing/tiles/service.py
+++ b/backend/app/processing/tiles/service.py
@@ -66,6 +66,15 @@ _CLUSTER_INPUT_LIMIT = 100000
 # 360/(extent*2^z) degrees of longitude at zoom z.
 _MVT_EXTENT = 4096
 
+# fix(#868): cluster_radius arrives in CSS/screen pixels (MapLibre clusterRadius
+# semantics, matching the client-side cluster path). Tiles display at 512 CSS px,
+# so one pixel covers _MVT_EXTENT / 512 = 8 extent units. The old bucket math
+# divided by the extent without this factor, treating pixels as extent units and
+# producing a grid ~8x finer than requested (overlapping cluster circles plus
+# unclustered singles leaking at low zoom).
+_TILE_DISPLAY_SIZE_PX = 512
+_CLUSTER_PX_TO_EXTENT_UNITS = _MVT_EXTENT / _TILE_DISPLAY_SIZE_PX
+
 # builder-audit #338 MVT-07: simplification tolerance schedule. Below this zoom the
 # geometry is simplified; at/above it the original geometry is served untouched.
 # (Distinct from _DEFAULT_NO_ATTR_BELOW_ZOOM, which happens to share the value 10
@@ -291,7 +300,15 @@ def _build_cluster_tile_query(
 
     Parameters at execution time:
     $1=z, $2=x, $3=y, $4=source layer name, $5=cluster max zoom,
-    $6=cluster radius in tile pixels.
+    $6=cluster radius in CSS/screen pixels (MapLibre ``clusterRadius``
+    semantics; converted to MVT extent units in-query via
+    ``_CLUSTER_PX_TO_EXTENT_UNITS`` — fix(#868)).
+
+    fix(#868): the bucket grid is anchored to absolute EPSG:3857 coordinates
+    (``floor(ST_X(geom) / bucket_size)``), not to the tile's own minx/miny.
+    Bucket size at a fixed zoom is identical for every tile, so anchoring
+    absolutely makes the grids of adjacent tiles line up instead of drifting
+    by each tile's origin offset (a seam artifact at tile borders).
 
     Cluster output follows the MapLibre client-side cluster property shape:
     clustered features carry ``point_count`` and ``point_count_abbreviated``;
@@ -341,8 +358,6 @@ bounds AS (
     SELECT
         _env.geom,
         ST_Transform(_env.geom, 4326) AS geom_4326,
-        ST_XMin(_env.geom) AS minx,
-        ST_YMin(_env.geom) AS miny,
         GREATEST(ST_XMax(_env.geom) - ST_XMin(_env.geom), 1.0) AS width,
         GREATEST(ST_YMax(_env.geom) - ST_YMin(_env.geom), 1.0) AS height
     FROM _env
@@ -360,17 +375,20 @@ bucketed AS (
     SELECT
         candidates.gid,
         candidates.geom_3857,
+        -- fix(#868): bucket size converts $6 (CSS px) to extent units before
+        -- scaling by tile width, and the grid anchors to absolute 3857 coords
+        -- so adjacent tiles at the same zoom share one grid.
         CASE
             WHEN $1::integer <= $5::integer THEN floor(
-                (ST_X(candidates.geom_3857) - bounds.minx)
-                / GREATEST(bounds.width * $6::float8 / 4096.0, 1.0)
+                ST_X(candidates.geom_3857)
+                / GREATEST(bounds.width * $6::float8 * {_CLUSTER_PX_TO_EXTENT_UNITS} / {_MVT_EXTENT}.0, 1.0)
             )::integer
             ELSE candidates.gid
         END AS bucket_x,
         CASE
             WHEN $1::integer <= $5::integer THEN floor(
-                (ST_Y(candidates.geom_3857) - bounds.miny)
-                / GREATEST(bounds.height * $6::float8 / 4096.0, 1.0)
+                ST_Y(candidates.geom_3857)
+                / GREATEST(bounds.height * $6::float8 * {_CLUSTER_PX_TO_EXTENT_UNITS} / {_MVT_EXTENT}.0, 1.0)
             )::integer
             ELSE candidates.gid
         END AS bucket_y
@@ -538,7 +556,9 @@ async def get_cluster_tile(
             and popups work past cluster max zoom.
         tile_columns: Phase 269 H-23 allowlist override (None / [] / list).
         additional_columns: Runtime opt-in columns (the ``cols=`` param).
-        cluster_radius: Cluster radius in tile pixels.
+        cluster_radius: Cluster radius in CSS/screen pixels (MapLibre
+            ``clusterRadius`` semantics; converted to MVT extent units
+            inside the query — fix(#868)).
         cluster_max_zoom: Maximum zoom level at which clustering is active.
         conn: Optional already-acquired asyncpg connection (see ``get_tile``
             docstring for DP-02 details; T-1209-10).
--- a/backend/app/processing/tiles/service.py
+++ b/backend/app/processing/tiles/service.py
@@ -335,10 +335,21 @@ def _build_cluster_tile_query(
     cross-tile cells exist there, and ring candidates would only consume
     the input cap before ownership discards them, starving the tile of its
     own points. Singles that only enter via the expanded ring drop out
-    through the ownership filter, and the MVT buffer grows with the radius
-    so owner-emitted geometry up to one bucket outside the tile is not
-    clipped. ``ORDER BY gid`` keeps candidate membership deterministic per
-    envelope.
+    through the ownership filter. ``ORDER BY gid`` keeps candidate
+    membership deterministic per envelope.
+
+    fix(#868, codex round 4 on PR #872): ownership filters BEFORE the
+    feature cap (in the ``grouped`` CTE), so neighbor-owned cells can never
+    consume the output budget and displace owned cells. And because a cell
+    owned via its anchor can have its centroid up to one bucket inside the
+    NEIGHBOR tile — where a client whose viewport covers the centroid but
+    not the owner tile would never see it (no MVT buffer helps a tile the
+    client never requests) — the emitted geometry is CLAMPED into the
+    owning tile, one MVT extent unit inside each edge. Maximum positional
+    error from the clamp: one bucket, i.e. the cluster radius in CSS px
+    (48 by default), deterministic, and zero for geometry already inside
+    the tile (all past-max-zoom points; every cell whose centroid is
+    in-tile).
 
     Cluster output follows the MapLibre client-side cluster property shape:
     clustered features carry ``point_count`` and ``point_count_abbreviated``;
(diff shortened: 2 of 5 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
Good catch on the straddling cell: the absolute grid traded the old per-tile seam drift for a border cell computed twice, each side seeing only its own fragment. Fixed in caf975491. The candidate scan now covers the tile expanded by one full bucket (clamped to the world envelope before the 4326 transform, which otherwise wraps longitude past the world edge and inverts the bbox), candidates are ordered by gid so neighbors agree on a straddling cell's membership under the input cap, and each cell is emitted only by the tile whose envelope contains its centroid (half-open bounds). Singles that only enter via the expanded ring drop through the same filter. Added a two-tile SQL-level regression test: the straddling pair comes out as one cluster from the owning tile only, and the neighbor's interior point never leaks across.

ishiland:
@codex review

ishiland:
Both round-2 findings are real, and the first one is on us: round 1's centroid ownership was our design call, and it does break under cap saturation exactly as described — diverging candidate subsets move the computed centroid across the border. Complete shared-cell membership under a finite cap isn't achievable, so 2b5cd0d9b makes disagreement structurally impossible instead: ownership now anchors on the cell ORIGIN (bucket index x bucket size; the point itself past cluster max zoom), pure grid geometry both neighbors always agree on. Cap saturation now degrades to an undercounted cluster in the owner tile — the class any capped per-tile grid has — never a cross-tile double- or zero-emit. The ST_AsMVTGeom buffer scales with the radius (GREATEST(256, ceil(px*8)+16)) since the owner-emitted point can sit up to one bucket outside the tile, covering the 256 px extreme. The world-edge drop is fixed too: east (x=2^z-1) and north (y=0) edge tiles use inclusive upper ownership bounds; internal borders stay half-open. New tests: lon-180 point (transforms to exactly world XMax) emitted only by the edge tile; the round-1 seam test unchanged under anchor ownership; a cap-divergence test with an injected cap of 2 showing the shared cell emits at most once.

ishiland:
@codex review

ishiland:
Both round-3 findings fixed in b21e48a23 (branch rebased on main for #869, so SHAs moved). West edge: fixed by construction rather than another special case — the grid now anchors at the world minimum, so bucket index = floor((x - world_min)/bucket) and every in-world point yields an in-world cell origin in [world_min, world_max]; the inclusive lower bounds then cover the west/south edges with no extra clause. Re-derived the east/north inclusive upper bounds and kept them: an anchor can still land exactly on the world maximum (a lon-180 point past max zoom, or a bucket size that divides the world width exactly, e.g. radius 64 gives bucket = W/2^(z+3)). Ring starvation: the scan expansion is now zero past cluster max zoom — no cross-tile cells exist there, so ring rows could only burn the candidate cap before ownership discarded them. New tests: a point 1 km from lon -180 at a clustering zoom emits from the west edge tile (it vanished under the 0-anchored grid), and a past-max-zoom neighbor-heavy scenario with an injected cap of 2 keeps the tile's own point. Seam/cap/world-edge tests re-derived for the world-min grid; shape assertions pin the new anchor expressions and the conditional expansion.

ishiland:
@codex review

ishiland:
Both round-4 findings fixed in 985ef1ad3 (rebased on main again, SHAs moved). Visibility hole: agreed the buffer can't help a tile the client never requests, so keeping anchor ownership, the emitted point is now clamped into the owning tile, one MVT extent unit inside each edge. Max positional error is one bucket = the cluster radius in CSS px (48 default), deterministic, and zero whenever the centroid is already in-tile (which includes all past-max-zoom points). The round-2 radius-scaled buffer is superseded by the clamp and reverted to 256. Cap order: ownership is per-row grid arithmetic (all members of a cell share its anchor), so the filter moved into the aggregation CTE ahead of GROUP BY and the feature LIMIT — neighbor-owned cells can no longer consume the output budget. New tests: a straddling cell with its raw centroid in the neighbor emits from the owner with tile-space coordinates inside the extent (clamped to the east edge) while the neighbor emits nothing, and a feature-cap-of-1 scenario with three neighbor-owned cells keeps the tile's own cell. Full targeted set at 99 passed.

ishiland:
@codex review

ishiland:
@codex review

PR #876, opened 2026-07-29
chore(release): bump version to 1.6.1

ishiland:
Version bump for the v1.6.1 patch release.

- `make bump VERSION=1.6.1` across all version sites, plus the three lockfiles the bump script does not touch (backend/uv.lock, sdks/typescript/package-lock.json, frontend/package-lock.json).
- CHANGELOG: closed Unreleased as [1.6.1] - 2026-07-29. Ships the globe projection on viewer surfaces (#845), the server-side cluster grid fix (#868), API-key hardening (#864), the react-router CSRF backport, the mode-aware admin settings gate (#857), the backup tar live-write fix (#843), and the pre-tag audit fix batches.

No schema or API changes beyond the version fields. Verification: `make version-check` green locally; pre-commit hooks green.

https://claude.ai/code/session_01A8vNJqrsio7yP5vWNhauVg

PR #880, opened 2026-07-30
chore(release): stamp lockfiles in bump and gate them in version-check

### LINKED ISSUE #877: make bump misses every lockfile; stamp them in bump_version.py
`make bump VERSION=X.Y.Z` rewrites the manifests but none of the lockfiles that embed the package version, so every release has hand-stamped them — and we keep finding new ones after the fact:

- `backend/uv.lock` (known since 1.5.1)
- `sdks/typescript/package-lock.json` (known since 1.5.1)
- `frontend/package-lock.json` (missed through 1.5.1, caught at 1.6.0)
- `mcp/uv.lock` (missed at 1.6.0, caught by review on the 1.6.1 bump PR #876)
- root `package-lock.json` (sat at 1.5.1 through TWO releases, caught at 1.6.1)

Suggested fix: after rewriting manifests, `bump_version.py` (or the `bump` make target) runs the refreshers itself — `uv lock` in backend/ and mcp/, `npm install --package-lock-only --ignore-scripts` at root, frontend/, and sdks/typescript/ — and `make version-check` greps each lockfile for the expected version so a miss fails CI instead of shipping.

ishiland:
## What changed

`make bump VERSION=X.Y.Z` rewrote every version manifest but none of the five lockfiles that embed the package's own version, so every release hand-stamped them — and each one was missed at least once (the root `package-lock.json` sat at 1.5.1 through two releases; `mcp/uv.lock` was caught only by review on the 1.6.1 bump PR #876).

- `scripts/bump_version.py` now stamps all five lockfiles after the manifests.
- `scripts/check_version_coherence.py` reads each of them as a version site, so `make version-check` (a required CI job) goes red on a miss instead of the miss shipping.

Sites added to the gate — 10 → 19:

| Lockfile | Site(s) |
| --- | --- |
| `backend/uv.lock` | `[[package]] geolens-backend`.version |
| `mcp/uv.lock` | `geolens-mcp`.version **and** `geolens`.version (the editable `../sdks/python` path dep, whose version this bump also moves) |
| `package-lock.json` (root) | top-level `.version` **and** `packages."".version` |
| `frontend/package-lock.json` | same two |
| `sdks/typescript/package-lock.json` | same two |

Verified against the files rather than assumed: lockfileVersion 3 records the root package's version **twice** (top-level and under `packages.""`), and `mcp/uv.lock` carries a second repo-owned version because it resolves `sdks/python` as an editable path dependency. Both are stamped and both are gated independently.

Each site is read structurally (`tomllib` / `json`) — the package's *own* entry, never "the version string appears somewhere in the file" — so a dependency that happens to sit at the release version cannot satisfy the gate. Both scripts also refuse to touch a `uv.lock` entry that is not locally sourced (`editable`/`virtual`/`directory`), so a package that ever became a registry dependency fails loudly instead of being stamped with an unrelated version.

## Design choice: in-place stamp, not re-running the refreshers

A version bump changes no requirement, so the resolution is unchanged — an in-place stamp of the package's own entry is exactly what `uv lock` / `npm install --package-lock-only` would produce, minus the network round-trip that could drag unrelated dependency churn into a release commit. Proof it is equivalent: with the trial bump applied, `uv lock --check` passes in both `backend/` and `mcp/`; reverting only the lock's version line makes it fail (`The lockfile at uv.lock needs to be updated`), so uv does care about this field and agrees with what the script wrote.

## Gate evidence (red, then green)

Deliberate single-site sabotage, one per shape, each reverted after:

```
# mcp/uv.lock: geolens-mcp -> 1.6.0
FAIL: version drift detected. Canonical (backend/pyproject.toml) = '1.6.1'.
  - mcp/uv.lock (package 'geolens-mcp'.version): '1.6.0' != '1.6.1'

# package-lock.json: TOP-LEVEL only -> 1.5.1 (the real 1.6.1 miss)
FAIL: ...
  - package-lock.json (.version): '1.5.1' != '1.6.1'

# frontend/package-lock.json: packages."" ONLY -> 1.6.0
FAIL: ...
  - frontend/package-lock.json (packages."".version): '1.6.0' != '1.6.1'
```

Both halves of each package-lock are therefore gated independently. Green again after revert: `OK: all 19 version sites agree on 1.6.1.`

## Verification

| Command | Result |
| --- | --- |
| `make version-check` (current tree, 1.6.1) | pass — 19 sites agree; **no lockfile was lagging at HEAD**, nothing needed a catch-up stamp |
| `make bump VERSION=1.6.2` then `git diff --stat` | pass — every lockfile moves and nothing else does: `backend/uv.lock` 1 line, `mcp/uv.lock` 2, each `package-lock.json` 2; no dependency churn |
| `make version-check` at trial 1.6.2 | all 19 sites agree on 1.6.2; fails only on the expected missing `## [1.6.2]` CHANGELOG section |
| `cd backend && uv lock --check --offline` / `cd mcp && ...` | pass at the stamped 1.6.2; fails when the lock version is reverted by hand |
| `make bump VERSION=1.6.1` (round-trip back) | restores every site byte-identically; `git status --short` shows only the 5 files in this PR — no trial-bump residue |
| `python3 scripts/tests/test-changelog-gate.py` | 20 passed, 0 failed |
| `cd backend && uv run ruff check . && uv run ruff format --check .` | pass (846 files formatted; no backend Python touched) |

Docs/comments that enumerate the version sites are updated with the lockfiles: `Makefile` (both targets), the `version-coherence` job comment in `ci.yml`, `RELEASE.md`'s source-coherence bullet, and both script docstrings. No CHANGELOG entry — release tooling, not an operator-visible change (consistent with #866/#867/#872/#873).

Closes #877

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -697,7 +697,8 @@ jobs:
 
   # ---------- version coherence ----------
   # REL-04: assert every version site agrees (backend/cli/sdks×2/root+frontend
-  # package.json/openapi.json info.version/main.py fallback constant). Blocks a
+  # package.json/openapi.json info.version/main.py fallback constant/the 5
+  # lockfiles that embed the package's own version, fix(#877)). Blocks a
   # release where one site silently drifted from the rest. Pure stdlib (tomllib
   # + json + re); no DB, no Docker, runs on every push/PR.
   version-coherence:
--- a/Makefile
+++ b/Makefile
@@ -297,8 +297,10 @@ catalog-domain-discipline: ## Verify no external imports of catalog/datasets/dom
 # ----- Version management (REL-05) -----
 # `make bump VERSION=X.Y.Z` rewrites EVERY version site atomically (backend +
 # cli + sdks×2 + root/frontend package.json + openapi.json info.version + the
-# metadata fallback constant in main.py). The write side of the version
-# contract; `make version-check` is the read/verify side.
+# metadata fallback constant in main.py + the 5 lockfiles that embed the
+# package's own version — backend/mcp uv.lock and root/frontend/sdks-ts
+# package-lock.json, fix(#877)). The write side of the version contract;
+# `make version-check` is the read/verify side.
 bump: ## Rewrite all version sites to VERSION=X.Y.Z (single source of truth)
 ifndef VERSION
 	$(error VERSION is required: make bump VERSION=X.Y.Z)
@@ -307,8 +309,8 @@ endif
 
 # `make version-check` — version-coherence gate (REL-04). Reads every version
 # site (backend/cli/sdks×2/root+frontend package.json/openapi.json info.version/
-# main.py fallback constant) and exits non-zero if any disagree. Run in CI to
-# block a release where one site silently drifted from the rest.
+# main.py fallback constant/the 5 lockfiles) and exits non-zero if any disagree.
+# Run in CI to block a release where one site silently drifted from the rest.
 version-check: ## Assert all version sites agree (CI gate)
 	uv run --no-project python scripts/check_version_coherence.py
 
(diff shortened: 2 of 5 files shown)


ishiland:
@codex review

Issue #584, opened 2026-07-19
Builder popup: custom visible_fields don't show — setTiles reload dropped, null values hidden, low-zoom 'No attributes'


ishiland:
\"Popup not showing custom fields\" decomposes into three verified defects:

**(a) Live edits to popup visible_fields never reach the tiles.** Editing the popup field selection correctly rebuilds the signed tile URL with the new `cols=` and calls `VectorTileSource.setTiles()` (verified by instrumenting the source), but maplibre-gl 5.x drops the resulting reload whenever the source's TileManager is paused — which the same sync pass guarantees via its layer updates. `tile_manager.ts` `_dataHandler` returns early on a `content` event while `_paused` WITHOUT setting `_shouldReloadOnResume`, so the reload is lost, tiles keep the pre-edit column set, and the popup can never show the newly selected fields until a full page reload or an unrelated re-tile (token rotation). Verified live: `map.refreshTiles(sourceId)` (whose reload path DOES set the resume flag) immediately flushes the stale tiles. (Upstream maplibre bug; app-side workaround: schedule `refreshTiles` after `setTiles`.)

**(b) Null-valued configured fields silently vanish.** `FeaturePopup` intersects the configured allowlist with the keys present in tile properties (`visibleFields.filter(k => propMap.has(k))`). ST_AsMVT omits null-valued properties from the tile, so a configured field that is null on the clicked feature renders no row at all — the existing `--` placeholder is unreachable.

**(c) Low-zoom default popups say \"No attributes\".** At z<10 the tile server strips all attribute columns unless opted in via `cols=`; in the default \"all fields\" mode nothing is opted in, so clicking a feature on a freshly added layer at low zoom yields an empty popup with no explanation.

Fixes: (a) re-issue the reload via the public `map.refreshTiles()` after `setTiles`; (b) render every configured field, `--` when absent; (c) replace the empty-state copy with a \"zoom in to view attributes\" hint when the dataset has columns but the tile carried none.

PR #881, opened 2026-07-30
fix(tiles): re-mint tile tokens on tab return instead of after the 403 burst

### LINKED ISSUE #755: Viewer/builder: returning to a backgrounded tab 403-bursts vector tiles until the reactive re-mint lands

ishiland:
Frontend only. No schema, API, or config impact. No new `t()` keys.

## Half 1 — proactive re-mint on tab return (the outstanding half)

Tile sigs are minted on `round_expiry()` 900 s boundaries (`backend/app/processing/tiles/signing.py`), so a signed URL lives 15–30 min. A tab backgrounded for a few minutes routinely crosses a boundary; on return MapLibre resumes its fetches with the stale `sig` and every visible vector tile 403s. The reactive GUARD-03 / #621 handler does heal the map, but only after a full round trip, and the burst is what the user and the problem report see.

`useVisibleTileTokenRefresh` now sits next to `useTileAuthRecovery` in `frontend/src/hooks/use-tile-auth-recovery.ts` and is called from the three surfaces that already own tile-auth recovery — `BuilderMap`, `ViewerMap`, `DatasetMap`. On the visible edge of `visibilitychange` it inspects the tokens the surface already holds (`exp` is on every vector token) and, when any is past expiry or within 60 s of it, kicks the **same** throttled re-mint the 403 handler uses. The fresh batch then flows through each surface's existing token→`setTiles` plumbing.

Three things this deliberately does *not* do:

- **No parallel refresh path.** It calls the surface's existing `recoverTileAuth`, so the existing 30 s cooldown / 10 s settle window still bounds the mint rate across rapid tab flipping, and a 403 that still slips through now arrives *inside* the settle window — where it is reported suppressed instead of surfacing error UI.
- **Nothing on the hidden edge.** MapLibre 5 drops a `setTiles` reload while the source's `TileManager` is paused (#584) and a hidden tab has no rAF, so a re-mint issued before the tab is back would silently no-op.
- **Nothing at mount.** The builder's mount-time token race is a separate concern; a test pins that mount does not re-mint.

The listener is removed on unmount, and `getTokens` is read at event time rather than captured, so the listener registers once and still sees the current token set.

Raster tokens are skipped — they carry no `exp` (stable `tile_url`, auth rides the `Authorization` header via `setTransformRequest`).

## Half 2 — handled 403s no longer double-log

Already shipped in #808: `logUnhandledMapError` (`frontend/src/lib/map-error-log.ts`) is the `<Map onError>` prop on all three surfaces and drops the react-maplibre wrapper's `console.error` fallback for exactly the handled first-party vector-tile 401/403 case, keeping it for everything else (raster/DEM, 404 no-data, 5xx, status-less style errors). Verified wired on all three surfaces and re-ran its check (`frontend/src/lib/__tests__/map-error-log.test.ts`, 14 tests) as part of this PR's gates. No further change was needed.

## Verification

All run from `frontend/`:

| Command | Result |
| --- | --- |
| `npx vitest run src/hooks/__tests__/use-tile-auth-recovery.test.ts` | pass — 16 tests |
| `npx vitest run src/components/builder/__tests__/BuilderMap.token-visibility-refresh.test.tsx` | pass — 4 tests |
| `npx vitest run src/lib/__tests__/map-error-log.test.ts src/hooks src/components/builder src/components/viewer src/components/dataset` | pass — 191 files / 2479 tests |
| `npx vitest run` (full suite) | pass — 354 files / 4121 tests |
| `npm run lint` | pass |
| `npm run typecheck` | pass |

Mutation-checked the wiring pin: commenting out the `useVisibleTileTokenRefresh(...)` call in `BuilderMap.tsx` fails `BuilderMap.token-visibility-refresh.test.tsx` (1 failed / 3 passed), so the call site cannot be deleted silently.

## Runnable checks left behind

- `frontend/src/hooks/__tests__/use-tile-auth-recovery.test.ts` — skew window boundaries (past `exp`, +30 s, +60 s in; +61 s out), raster/empty/absent entries ignored, visible edge fires, hidden edge does not, tokens read at event time not mount, listener detached on unmount, repeated visibility flips mint once.
- `frontend/src/components/builder/__tests__/BuilderMap.token-visibility-refresh.test.tsx` — the call site itself: expired sig + `visibilitychange` re-mints once, fresh sig does not, hidden edge does not, mount does not, unmounted map does not.
- `frontend/src/lib/__tests__/map-error-log.test.ts` — the existing double-log check (unchanged).

Closes #755

### DIFF
--- a/frontend/src/components/builder/BuilderMap.tsx
+++ b/frontend/src/components/builder/BuilderMap.tsx
@@ -15,7 +15,7 @@ import { buildClusterTileUrl, buildSignedTileUrl, buildTileTransformRequest, isM
 import { useRemoteBasemapStyle } from '@/components/map/hooks/use-remote-basemap-style';
 import { logUnhandledMapError } from '@/lib/map-error-log';
 import { useTileTokens, useInvalidateTileTokens } from '@/hooks/use-tile-token';
-import { useTileAuthRecovery } from '@/hooks/use-tile-auth-recovery';
+import { useTileAuthRecovery, useVisibleTileTokenRefresh } from '@/hooks/use-tile-auth-recovery';
 import { useTileTokenError } from './hooks/use-tile-token-error';
 import { getEnvConfig } from '@/lib/env';
 import { pushReportEntry } from '@/lib/report';
@@ -567,6 +567,10 @@ export const BuilderMap = memo(function BuilderMap({
   // cached-token re-sign can't cure kicks one throttled token re-mint; the
   // token-sync effect re-signs sources when the fresh batch lands.
   const recoverTileAuth = useTileAuthRecovery(invalidateTileTokens);
+  // fix(#755): a tab backgrounded past the 900 s sig boundary comes back with
+  // stale tokens, so MapLibre's resumed fetches 403 before the handler above
+  // can heal them. Kick the same throttled re-mint on the visible edge.
+  useVisibleTileTokenRefresh(() => syncInputsRef.current.tokenMap.values(), recoverTileAuth);
 
   const handleLoad = useCallback(
     (e: MapLibreEvent) => {
--- a/frontend/src/components/builder/__tests__/BuilderMap.token-visibility-refresh.test.tsx
+++ b/frontend/src/components/builder/__tests__/BuilderMap.token-visibility-refresh.test.tsx
@@ -1,0 +1,250 @@
+// fix(#755): wiring pin for the proactive tile-token refresh. Tile sigs are
+// minted on 900 s `round_expiry()` boundaries, so a tab backgrounded for a few
+// minutes returns with an expired sig and MapLibre 403s every visible tile
+// before the reactive GUARD-03 / #621 handler heals the map. BuilderMap must
+// kick the SAME throttled re-mint on the visible edge of `visibilitychange`.
+//
+// The hook's own semantics (skew window, hidden-edge no-op, unmount detach) are
+// covered in hooks/__tests__/use-tile-auth-recovery.test.ts. This file pins the
+// CALL SITE: delete `useVisibleTileTokenRefresh(...)` from BuilderMap and these
+// fail. Uses the vi.hoisted fakeMap + @vis.gl/react-maplibre recipe from
+// BuilderMap.terrain-visibility.test.tsx.
+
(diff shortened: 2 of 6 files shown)


ishiland:
@codex review

PR #879, opened 2026-07-30
fix(admin): drop the legacy appearance redirect so the edition gate owns the tab

ishiland:
## What changed

Deleted the legacy `admin/settings/appearance` -> `/admin/settings/map` redirect from `frontend/src/App.tsx`. `AdminSettingsPage`'s own edition gate now owns that URL.

Verified before deleting that the gate really handles both directions (`AdminSettingsPage.tsx:82-89`): `visibleTabs` drops `appearance` unless `isEnterprise`, `activeTab` becomes null, and the page returns `<Navigate to="/admin/settings/general" replace />`. So community keeps a valid destination without the redirect, and enterprise renders the branding tab.

## Why

Two commits from adjacent days never met. `32c4cb60c` renamed the old basemaps "Appearance" tab to "Map" and left a redirect behind; `1f555a233` added the enterprise-only `SettingsAppearanceTab` at that same URL the next day. React Router ranks a fully static path above `admin/settings/:tab`, so the redirect fired for every edition. Enterprise admins clicking the sidebar's Appearance link landed on Map settings, and the branding tab could not be reached at all.

## Tests

- `frontend/src/App.test.tsx` (new): walks the real `appRoutes` element tree and fails on any static `admin/settings/<key>` path that collides with a key in `ALL_TAB_KEYS`. This is the only test that can catch a re-added redirect, because the gating tests in `AdminSettingsPage.test.tsx` mock `useParams` and never touch the route table. It also pre-empts the same collision for any future tab (`admin/settings/infrastructure` sits in that block today and is only harmless because `infrastructure` is not a tab key). `ALL_TAB_KEYS` is now exported so the guard cannot drift from the page.
- `e2e/admin.spec.ts`: the appearance test from #870 deliberately accepted `map|general` while the bug stood. Tightened to `/admin/settings/general` with a `General` heading assertion, and dropped the stale comment. The community invariants it already had (no `#show-badge`, no sidebar link) are unchanged.
- `AdminSettingsPage.test.tsx`: assertions were already written against the fixed behavior, so only the stale comment about the shadowing redirect needed rewriting.

## Verification

Ran from the worktree with `frontend/node_modules` symlinked from the main checkout.

| Command | Result |
| --- | --- |
| `npx vitest run src/App.test.tsx src/pages/admin/__tests__/AdminSettingsPage.test.tsx src/components/admin/__tests__/AdminSidebar.test.tsx` | pass, 22 tests |
[...]

Issue #825, opened 2026-07-28
ci: E2E and axe suites gate nothing on PRs


ishiland:
From the 1.6.0 pre-tag audit. The Playwright and axe suites run nightly only; a PR that breaks an e2e flow merges green and the breakage surfaces days later (the nightly itself was silently cancelled for a week until #814 fixed the concurrency group — the two gaps compounded).

Minimal shape:
- Run `e2e:smoke:core` on PRs touching `frontend/**` or `backend/**` (keep the full matrix nightly).
- Nightly failure should open/refresh a tracking issue so a red nightly is impossible to miss.

PR #895, opened 2026-07-30
test(e2e): unstick the two nightly-only specs

### LINKED ISSUE #894: Nightly E2E/a11y run is red
The nightly run on 2026-07-30 did not produce a green browser signal.

- `e2e-test`: failure
- `accessibility`: success
- Run: https://github.com/geolens-io/geolens/actions/runs/30531176944

`skipped` means an upstream lint/test job failed on the scheduled run, so the browser suites never started.

Close this issue once a nightly run is green again. The next red nightly comments here while it stays open, or files a fresh issue after it is closed.

ishiland:
Closes #894.

The nightly has been red on browser signal. Both failures turned out to be test-side — **no product code is wrong**, and PR #883 (which the timing pointed at) is cleared.

## `e2e/dataset-chat.spec.ts:193` — stale assertion, deterministic

Asserted the ephemeral badge read `Query result · 2 features`. PR #674 (`13fd3ec9a`, 2026-07-24) renamed that copy to `"Result"` — deliberately, once analysis previews started reusing the badge and "query result" stopped being accurate. The spec was never updated, so it has failed every nightly since. Nightlies were cancelled for most of the week before that, which is why it surfaced only now.

The feature itself works. In the same failure snapshot where the assertion failed, the CI `error-context.md` shows `status: Result · 2 features`, a `Dismiss result` button, and the coord readout already at `40.75° N · 73.95° W` — the exact value the *next* assertion wants. Both #533/#542 behaviors (overlay + `fitBounds`) had applied.

The fix is not a plain `s/Query result/Result/`: `getByText(/Result · 2 features/)` matches **two** nodes and trips strict mode, because `EphemeralBadge` renders the label both in an `sr-only` `role="status"` live region and in an `aria-hidden` visible copy. Scoping to `getByRole('status')` resolves to exactly one, since aria-hidden content is out of the accessibility tree.

## `e2e/analysis.spec.ts:58` — race, not a regression

The spec's comment says it abandons the builder *mid-job*. There is no mid-job: the materialize job completes in ~80 ms (`created_at .324 → started_at .345 → completed_at .403` in the failing run's own job-poll body; 52–136 ms locally). `useJobStatus` fetches immediately on mount, so the first poll returns `complete` while the builder is still up, `AnalysisJobWatcher` raises the toast there, and `setJob(null)` is persisted. `page.goto('/')` is a hard reload: it destroys that toast, and the rehydrated store has nothing to re-poll — zero `/api/jobs/` requests appear after the reload in the trace, and the notifications region is empty in all three attempts.

So the whole test hinged on whether the first poll landed before `page.goto('/')` started. It failed 1-in-4 locally on main and was 1 failed / 2 flaky on 07-29, 2 failed on 07-30.

Evidence it is not a #883 regression: the job **succeeded** (`{"status":"complete","dataset_id":"…","error_message":null}`), the output registered correctly (`Vector PostGIS <title> · Polygon · 1 feature · EPSG:4326`), the preview step passed, and the spec passes on current main.

Fix: navigate **client-side** instead of hard-reloading, which keeps the toast in the same document — that is the invariant the test was actually after ("completion surfaces after leaving the builder"). The action label is decided once when the toast is raised (`canAddToMap` depends on `MapBuilderPage` being mounted), so pinning `View dataset` cannot survive either ordering; the assertion now accepts either label and pins the actionable affordance instead.

Also moves the output-dataset id resolution ahead of the toast assertions, so a failure still cleans up. Previously each failed attempt leaked one dataset — visible as the catalog count climbing 3 → 4 → 5 across the three retries.

`attribute-table-ax.spec.ts:29` and `feature-editing.spec.ts:169` failed on 07-29 only, passed on retry, and were clean on 07-30. Genuine flake, no action.

## Verification

Both run against the local stack. These are spec-only changes, so a worktree run is valid here — the specs come from the branch and the app code is byte-identical to main.

    npx playwright test e2e/dataset-chat.spec.ts --project=chromium -g "ephemeral overlay" --reporter=line
    → 3 passed (20.6s)

    npx playwright test e2e/analysis.spec.ts --project=chromium -g "buffer preview" --repeat-each=5 --reporter=line
    → 7 passed (39.1s)

5/5 on the analysis spec, against ~1-in-4 failures for the old form.

## Note on coverage, worth knowing separately

`E2E Tests` runs on `schedule`/`workflow_dispatch` only (`ci.yml:1468`). PRs get `E2E Smoke (core)`, and **neither of these two specs is in any `e2e:smoke*` script**. The nightly is the only gate on them, which is why #879–#883 could not have caught the stale assertion. That is by design (#825 traded the full suite off PRs for the Actions budget), but it means a stale spec can sit red for days.

No schema, API, or config impact.

### DIFF
--- a/e2e/analysis.spec.ts
+++ b/e2e/analysis.spec.ts
@@ -78,22 +78,9 @@ test.describe('builder analysis tools', () => {
     };
     expect(jobId).toBeTruthy();
 
-    // Abandon the builder mid-job: tracking is global (AnalysisJobWatcher in
-    // RootLayout), so completion must still surface on a different page — with
-    // "View dataset" rather than "Add to map", since no builder is mounted.
-    await page.getByRole('button', { name: 'Close panel' }).click();
-    await page.goto('/');
-    // Scope to the toast: the finished dataset also lands in the catalog list
-    // behind it (which is the query invalidation doing its job).
-    const completionToast = page
-      .locator('[data-sonner-toast]')
-      .filter({ hasText: OUTPUT_TITLE });
-    await expect(completionToast).toBeVisible({ timeout: 60_000 });
-    await expect(
-      completionToast.getByRole('button', { name: 'View dataset' }),
-    ).toBeVisible();
-
-    // Resolve the created dataset id for cleanup.
+    // fix(#894): resolve the id BEFORE asserting on the toast, so an assertion
+    // failure still cleans up. Previously each failed attempt leaked one output
+    // dataset (visible as the catalog count climbing across retries).
     for (let attempt = 0; attempt < 30; attempt++) {
       const res = await fetch(`${BASE_URL}/api/jobs/${jobId}`, { headers });
       if (res.ok) {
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

ishiland:
Addressed the P2 — it was a fair hit, and the fix is not the one suggested.

**The finding is correct.** Resolving the id before the assertions meant the test waited for `complete` before navigating, so the toast was already raised on the builder and the client-side navigation merely preserved it. That is stable but vacuous: it would still pass if `AnalysisJobWatcher` stopped tracking on unmount, which is exactly the #682 coverage the spec exists for.

**The literal suggestion reintroduces the bug I was fixing.** Navigating before waiting for terminal status is the original coin flip: the job finishes in ~80 ms and `useJobStatus` fetches immediately on mount, so whether the toast lands on the builder or after it depends on how fast the UI steps run. That is what made this spec 1 failed / 2 flaky on 07-29 and 2 failed on 07-30, and ~1-in-4 locally.

**What I did instead** — hold the browser's view of the job at `running` until the builder is gone, via `page.route('**/api/jobs/*')`. "Mid-job" becomes true by construction. Three consequences:

1. Added `await expect(completionToast).toBeHidden()` while the job is **already complete server-side**. This is the assertion that gives the next one meaning: the toast can only have been raised after the builder unmounted.
2. Restored the strict `View dataset` assertion. With no builder mounted, `canAddToMap` is provably false, so the label is deterministic — no need to accept either branch.
3. Cleanup polling uses Node-side `fetch()`, which `page.route` does not touch, so it still resolves the dataset id *before* any assertion that can fail. No leak on failure.

**Verified with a negative control, not just a green run.** Patched a throwaway copy of the spec to unmask immediately after `Create dataset` and wait for the toast while still on the builder — i.e. forced the exact scenario the finding describes. It fails at the new assertion:

    Error: expect(locator).toBeHidden() failed
    > 153 |     await expect(completionToast).toBeHidden();
    1 failed

So the assertion has teeth. With the mask in place:

    npx playwright test e2e/analysis.spec.ts --project=chromium -g "buffer preview" --repeat-each=5 --reporter=line
    → 7 passed (35.9s)

5/5, against ~1-in-4 failures for the original form.

ishiland:
@codex review

PR #898, opened 2026-07-30
fix(export): filter an antimeridian bbox server-side instead of via -spat

### LINKED ISSUE #885: fix(export): a west>east bbox silently produces an empty ogr2ogr export
Found during the #697 antimeridian audit. This one is an asymmetry bug: the same query returns rows on some formats and nothing on others.

`backend/app/processing/export/ogr.py:88-99` passes the parsed bbox straight to `ogr2ogr -spat`, which interprets a `west > east` rectangle as degenerate and matches nothing. There is no error — the export succeeds and is empty.

The asymmetry is the tell:

- `parse_bbox` **explicitly documents** west>east as valid input.
- `backend/app/processing/export/parquet.py:185-190` splits it correctly.
- `backend/app/modules/catalog/features/service.py:173-177` splits it correctly.

So `?bbox=170,-20,-170,-15` returns features from `/features` and from `format=parquet`, but yields an empty GPKG / GeoJSON / SHP. A user exporting a Fiji or Aleutians AOI gets a silently empty file.

**Fix sketch.** Reuse the same split the parquet and features paths already do, and pass two `-spat` invocations or a `-clipsrc` geometry. A test asserting non-empty output for a crossing bbox on each ogr-backed format would pin it.

ishiland:
## What changed

`ogr2ogr -spat` takes one rectangle, and GDAL reads the four numbers as a corner pair and derives an envelope from it. So `-spat 170 -20 -170 -15` did not match nothing — it matched the **complement**: `minx` and `maxx` sorted to `-170..170`, i.e. everything *except* the box the caller asked for. The export exited 0, so the user just got a file.

For a crossing box (`bbox[0] > bbox[2]`), `run_ogr2ogr_export` now drops `-spat` and pushes the two-envelope split into the server-side `-where` predicate the PostgreSQL driver already sends verbatim:

```
((geom_4326 && ST_MakeEnvelope(170.0, -20.0, 180, -15.0, 4326)
  AND ST_Intersects(geom_4326, ST_MakeEnvelope(170.0, -20.0, 180, -15.0, 4326)))
 OR (geom_4326 && ST_MakeEnvelope(-180, -20.0, -170.0, -15.0, 4326)
  AND ST_Intersects(geom_4326, ST_MakeEnvelope(-180, -20.0, -170.0, -15.0, 4326))))
```

An ordinary (non-crossing) box still goes through `-spat` byte-for-byte as before. When the caller also passed `where=`, the split is ANDed with it into the single `-where` argument.

Three smaller pieces come with it:

- The split fragment is now one builder, `bbox_where_sql()` in `export/ogr.py`, and the GeoParquet writer calls it instead of keeping its own copy — the two export paths can't drift apart again. `literal=True` renders float literals for the subprocess argv (which cannot carry bind parameters); the default renders `:minx`-style binds for the `text()` caller. Every bound goes through `float()`, so nothing but a numeric can reach the SQL text.
- The router drops the bbox for a **non-spatial** dataset. `-spat` was already a silent no-op there (`ERROR 1: Cannot set spatial filter: no geometry field present in layer`, exit 0), and `csv` is the only ogr format a geometry-less dataset can reach. Without this, a `geom_4326` predicate on a table with no such column would fail the whole translation → 500. Same `has_geometry` assumption the export cap's count and the `/features` bbox path already make.
- `conftest.py`'s `requires_ogr2ogr` redirect now patches both `build_pg_conn_str` bindings. `export/ogr.py` imports the name at module scope, so it held its own reference and `run_ogr2ogr_export` in a test was reading the **dev** database. Nothing exercised that before; the new end-to-end tests do.

## Why this shape and not `-clipsrc`

`-clipsrc` with a two-part MULTIPOLYGON does select the right features (verified — same 3 rows), but it *clips* geometry to the clip region. Measured on the same seam linestring:

```
=== -where (this PR) ===
seam LineString      [[179.0, -17.5], [-179.0, -17.5]]

=== -clipsrc MULTIPOLYGON ===
seam MultiLineString [[[179.0, -17.5], [170.0, -17.5]], [[-170.0, -17.5], [-179.0, -17.5]]]
```

The feature came back as a different geometry, cut at the box edges. That would change behavior for every bbox export, not just crossing ones. `-spat`/`-where` are select-not-clip, and the export must keep handing back whole features.

Two appended `-spat` runs (one per half) were the other option and were rejected on the duplicate constraint: a geometry that straddles the seam intersects **both** halves and would be written twice. A single `OR` predicate in one pass emits it once — pinned by `test_seam_feature_is_not_duplicated`.

`-sql` was not needed: the PostgreSQL driver passes `-where` straight into the server-side `WHERE`, and `run_ogr2ogr_export` already had a `-where` code path to reuse.

## Verification

All commands run on the host against the test Postgres (localhost:5434), on this branch.

**1. The bug, reproduced with real ogr2ogr against PostGIS** (5 linestrings: two inside the Fiji-ish box, one straddling the seam, one at Greenwich, one outside the latitude band):

```
$ ogr2ogr -f GeoJSON a.geojson "PG:..." data.probe885 -spat 170 -20 -170 -15 -spat_srs EPSG:4326
features: 1 ['seam_line']            # missed both features inside the box

$ ogr2ogr -f GeoJSON c.geojson "PG:..." data.probe885 -where "<the OR split>"
features: 3 ['east_of_170', 'seam_line', 'west_of_-170']    # correct, no duplicate

$ ogr2ogr -f GeoJSON d.geojson "PG:..." data.probe885 -clipsrc "MULTIPOLYGON(...)"
features: 3 ['east_of_170', 'seam_line', 'west_of_-170']    # correct set, but clipped
```

**2. The new tests fail without the fix.** Disabling only the new branch in `run_ogr2ogr_export` (`if False and bbox and ...`):

```
$ uv run pytest tests/test_export_antimeridian.py
tests/test_export_antimeridian.py ....F.F.FFFF....FF.F..                 [100%]
E  AssertionError: assert ['greenwich', 'on_the_seam'] == ['east_of_170', 'on_the_seam', 'west_of_-170']
9 failed, 13 passed
```

That left-hand side is the whole point: pre-fix, `bbox=170,-20,-170,-15` returned a feature at **longitude 0** and neither of the two features actually inside the box. Not just empty — wrong. All four non-crossing tests passed with the fix disabled, which is the no-regression half.

**3. The new tests pass with it:**

```
$ uv run pytest tests/test_export_antimeridian.py -q
22 passed, 43 warnings in 9.03s
```

Coverage: `bbox_where_sql` unit tests; command-shape tests (crossing → `-where`, no `-spat`; non-crossing → `-spat 170.0 -20.0 180.0 -15.0 -spat_srs EPSG:4326`, no `-where`; crossing + caller `where` → one ANDed `-where`); real ogr2ogr end-to-end for **gpkg / geojson / shp / csv** on both a crossing and a non-crossing bbox; seam feature emitted exactly once; seam geometry unclipped; the crossing bbox agreeing across `/features`, GeoParquet and the ogr export; and the router's bbox drop for a non-spatial dataset.

**4. Export, parquet, layering and Rule-1 structural files:**

```
$ uv run pytest tests/test_export_antimeridian.py tests/test_export.py tests/test_export_parquet.py \
    tests/test_export_parquet_writer.py tests/test_export_hardening.py tests/test_export_access.py \
    tests/test_export_temp_cleanup.py tests/test_layering.py tests/test_tenant_ingest_export_routing.py \
    tests/test_staging_runtime.py tests/test_rule1_structural.py -q
149 passed, 48 warnings in 47.66s
```

**5. Full backend suite:**

```
$ uv run pytest -n 4 -q
5975 passed, 81 skipped, 262 warnings in 648.92s (0:10:48)
```

**6. Backend gates:**

```
$ uv run ruff check .
All checks passed!
$ uv run ruff format --check .
847 files already formatted
```

## Impact

No schema, API surface or config change — `backend/openapi.json` and the SDKs are untouched. Behavior changes only for a bbox with `minx > maxx`, plus the dropped no-op `-spat` on a non-spatial csv export.

## Known gap, left out of scope

The export cap's bounded COUNT (`_count_selected_features`) still skips the bbox clause entirely for a crossing box (`bbox[0] <= bbox[2]` guard), so a dataset over the 5M-feature ceiling counts unfiltered and 413s on a crossing bbox that would have narrowed it under the cap. That is pre-existing, deliberately conservative (the clause is `&&`-only to keep the 5M-row scan cheap), and it is a ceiling behavior rather than the wrong-rows bug this PR fixes. Worth its own issue.

Closes #885

### DIFF
--- a/backend/app/processing/export/ogr.py
+++ b/backend/app/processing/export/ogr.py
@@ -40,6 +40,40 @@ FORMAT_MAP: dict[str, dict[str, str]] = {
 }
 
 
+def bbox_where_sql(bbox: list[float], *, literal: bool = False) -> str:
+    """Build the ``geom_4326`` bbox predicate for a raw-SQL WHERE fragment.
+
+    Each envelope carries a ``&&`` index prefilter plus an exact
+    ``ST_Intersects``, mirroring the features read path
+    (``catalog/features/service.py``).
+
+    fix(#885): a west>east bbox crosses the antimeridian — ``parse_bbox``
+    documents it as valid input — and is emitted as ``[minx..180]`` OR
+    ``[-180..maxx]``. It stays a single predicate, so a feature whose own
+    geometry straddles the seam matches the OR once and is selected exactly
+    once. Shared with the GeoParquet writer (``export/parquet.py``) so the two
+    export paths cannot drift.
+
+    ``literal=True`` renders float literals instead of ``:minx``-style bind
+    parameters, for the ogr2ogr ``-where`` argument — a subprocess argv element
+    that cannot carry binds. Every bound goes through ``float()``, so the
+    rendered text is always a numeric literal (``parse_bbox`` has already
+    rejected NaN/Inf).
+    """
+    if literal:
+        minx, miny, maxx, maxy = (repr(float(v)) for v in bbox)
+    else:
+        minx, miny, maxx, maxy = ":minx", ":miny", ":maxx", ":maxy"
+
+    def envelope(west: str, east: str) -> str:
+        env = f"ST_MakeEnvelope({west}, {miny}, {east}, {maxy}, 4326)"
+        return f"(geom_4326 && {env} AND ST_Intersects(geom_4326, {env}))"
+
+    if bbox[0] > bbox[2]:
+        return f"({envelope(minx, '180')} OR {envelope('-180', maxx)})"
+    return envelope(minx, maxx)
+
+
 async def run_ogr2ogr_export(
     table_name: str,
     output_path: str,
@@ -60,7 +94,10 @@ async def run_ogr2ogr_export(
         schema: Source PostgreSQL schema. Required so exports cannot silently
             read a same-named table from the shared ``data`` schema.
         target_srs: Optional target CRS (e.g. "EPSG:3857").
-        bbox: Optional bounding box [minx, miny, maxx, maxy] in WGS84.
+        bbox: Optional bounding box [minx, miny, maxx, maxy] in WGS84. A
+            west>east box crosses the antimeridian and is filtered server-side
+            against ``geom_4326``, so callers must not pass a bbox for a layer
+            without geometry (the router drops it for non-spatial datasets).
         where: Optional SQL WHERE clause for attribute filtering.
         format_key: Format key from FORMAT_MAP for format-specific options.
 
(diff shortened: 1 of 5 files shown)


ishiland:
@codex review

Issue #886, opened 2026-07-30
fix(catalog): extent rollups manufacture a global bbox across the antimeridian


ishiland:
Found during the #697 antimeridian audit. Distinct from #697 and from the STAC-import bug: here **no individual record wraps**, but the aggregate does.

Two ordinary Fiji datasets — one at lon 179, one at lon -179 — roll up to `-180..180`. Each record is fine; the min/max fold across them spans the globe the wrong way.

Five sites do the same naive fold, so **one shared helper fixes all of them**:

- `backend/app/modules/catalog/search/router.py:571`
- `backend/app/modules/catalog/collections/service.py:312`
- `backend/app/standards/stac/router.py:635-638` and `:855-858` (STAC Collection `extent.spatial.bbox`)
- `backend/app/modules/catalog/ai/service.py:550-553` — map `center_lng` gets overwritten to the Gulf of Guinea
- `frontend/src/components/builder/BuilderMap.tsx:114-117` (auto-fit)

**Fix sketch.** A single "union of lon ranges on a circle" helper: pick the aggregate whose *gap* is largest and emit the complement, which naturally produces west>east when that is the honest answer. Then decide per consumer whether it wants west>east (STAC/OGC, spec-correct) or a split.

### Related sites found in the same audit (propagation only — they inherit a wrap rather than create one)

`processing/ingest/metadata.py:358,888`, `catalog/features/service.py:630`, `processing/vector/quicklook.py:140`, `processing/tiles/router.py:405-410` (a 360° raster extent drives `maxzoom` toward 0, so **the raster stops rendering when zoomed in**), `scripts/seed_tiles.py:115`, `frontend/src/lib/map-extent.ts:21`, `BBoxPreview.tsx:127`, `use-builder-layers.ts:438` (zoom-to-layer is a silent no-op for west>east), plus ~11 naive `to_shape(...).bounds` sites across DCAT / DCAT-US / GeoDCAT-AP / OGC.

### Also related: the NL→SQL prompt teaches the unnormalized pattern

[...]

Issue #884, opened 2026-07-30
fix(stac): remote import writes an inverted ring for RFC 7946 west>east bboxes


ishiland:
Found during the #697 antimeridian audit. This is the highest-value site in that class because it is the only place in the product that *ingests* spec-compliant antimeridian bboxes, and it has no branch for them.

`backend/app/modules/catalog/sources/stac_router.py:545-548` builds the record extent straight from `item.bbox`:

```
POLYGON((w s, e s, e n, w n, w s))
```

RFC 7946 §5.2 **mandates** `west > east` for a bbox that crosses the antimeridian. So importing a legitimate `[170, -20, -170, -15]` produces a self-intersecting ring that covers the **wrong 340°** — everything except the intended 20° strip.

Consequences are the same as #697, but arriving from remote metadata rather than from an analysis operation: the stored geometry is invalid, `extent_bbox` and the OGC collection extent serve a near-global box, and `ST_Intersects` yields feature-level false positives at arbitrary distance.

**Fix sketch.** Detect `w > e` and either split into two rings (`w..180` and `-180..e`) and store a MULTIPOLYGON, or normalize via the 0..360 shift the #697 fix uses. Note `records.spatial_extent` is `geometry(POLYGON)` (see the memory of that column constraint), so a two-ring result needs either a column change or an envelope of the shifted geometry — decide that explicitly rather than by accident.

Needs a regression test importing a west>east STAC item; every current STAC fixture is low-lon.

Issue #697, opened 2026-07-25
bug(analysis): buffer output crossing the antimeridian likely corrupts the registered extent (needs one runtime probe)


ishiland:
From the 2026-07-25 analysis scenario audit. **Unconfirmed at runtime** — filed with the probe as step 1.

**Claim.** `ST_Buffer(...::geography, d)::geometry` output is never longitude-normalized, and registration derives the extent from a plain `ST_Extent(geom_4326)` (`backend/app/processing/ingest/metadata.py:873-910`, `get_extent` padding only degenerate POINT/LINESTRING cases for the `records.spatial_extent geometry(POLYGON)` constraint). A buffer of features near lon ±180 (Fiji/NZ/Aleutians) or near a pole plausibly registers an extent that spans nearly the whole globe or carries |lon| > 180 — propagating into published STAC and OGC bboxes. The repo handles the antimeridian only for INPUT bboxes (`backend/app/core/geo.py:22-39`); there is no output-side counterpart, and every analysis test fixture is low-lat/low-lon.

**Step 1 (S):** materialize a buffer of a point at lon 179.95 and inspect the registered extent + STAC bbox.
**Step 2 if reproduced (M):** normalize via `ST_ShiftLongitude` (or split) before extent computation, or clamp in `get_extent` — decide once the failure shape is known.

Related: the edge-fixture test-matrix issue tracks adding permanent ±180/polar fixtures.

ishiland:
## Step 1 probe: reproduced

Ran the probe directly against PostGIS (18.4 / PostGIS 3.6.4 / GEOS 3.13.1) rather than through a materialize job, since the claim is about `ST_Buffer(geography)` output and `ST_Extent`, not about the job pipeline:

```sql
SELECT ST_AsText(ST_Extent(g)) FROM (
  SELECT ST_Buffer('SRID=4326;POINT(179.95 -18)'::geography, 20000)::geometry AS g
) t;
```

```
POLYGON((-179.98062896410434 -18.1805299900566,
         -179.98062896410434 -17.81946616524579,
          179.9898081297982  -17.81946616524579,
          179.9898081297982  -18.1805299900566,
         -179.98062896410434 -18.1805299900566))

xmin -179.98062896410434 .. xmax 179.9898081297982
```

A 20 km buffer whose true longitude span is ~0.38° registers an extent spanning **359.97°** — very nearly the whole globe.

## Where it actually goes wrong

[...]

PR #899, opened 2026-07-30
fix(ingest): shift 0..360 sources and report Mercator-clamp drops

### LINKED ISSUE #888: fix(ingest): clip_to_mercator_bounds silently discards data with no warning
Found while probing #697 — two distinct silent-data-loss paths in one function, neither of them a wrap.

`clip_to_mercator_bounds` (`backend/app/processing/ingest/metadata.py:1184`) intersects incoming geometry with `ST_MakeEnvelope(-180, -85.06, 180, 85.06)`.

**1. Sources in the 0..360 Pacific convention lose everything east of lon 180.** That convention is common in ocean/climate data. The intersection silently drops the out-of-range half; the user sees a successful ingest with missing features and no message.

**2. Valid polar geometry is silently discarded.** A GeoJSON point at lat -89.95 ingests as `MULTIPOINT EMPTY`. The Mercator clamp itself is intentional and defensible, but the user is never told their geometry was dropped — and the downstream failure they eventually hit (`Analysis produced no features to save`) gives no hint of the real cause. This is what made the polar half of #697 non-reproducible.

**Fix sketch.** Detect a 0..360 source (all lons ≥ 0 with values > 180) and shift rather than clip. For the polar case, emit an ingest warning that names the dropped feature count and the reason, so the failure is legible at the point where the data is lost instead of three steps later.

ishiland:
`clip_to_mercator_bounds` lost data on two separate paths and said nothing on either.

## What changed

**1. A 0..360 source is shifted, not clipped.** Sources written in the 0..360 Pacific longitude convention (ocean and climate data, mostly) had everything east of lon 180 intersected away against `ST_MakeEnvelope(-180, -85.06, 180, 85.06)`. Those coordinates are not out-of-range data, they are the same world with a different origin. `_shift_zero_to_360_longitudes` now recognises the convention and translates the out-of-range rows by -360 in X before the clip runs, so every feature survives.

**2. The clamp now reports what it removed.** The clamp itself stays: GeoLens serves Web Mercator tiles and 85.06 deg is a real limit. What changed is that the clip UPDATE counts, in the same statement that does the work, how many rows it emptied (`dropped_features`) and how many it merely trimmed (`clipped_features`). A non-zero count becomes a `mercator_clip` ingest warning on the job. A point at lat -89.95 used to become `MULTIPOINT EMPTY` silently and only surface three steps later as "Analysis produced no features to save" — which is what made the polar half of #697 non-reproducible.

**3. Rows that arrive already empty** are excluded from the clip UPDATE, so they cannot inflate the counts. The clip was a no-op for them either way.

## Guard conditions and why each is needed

All four must hold before anything moves. #883 established that a single-condition guard on this class of problem is a coin flip, so each condition is here for a specific failure:

| Condition | Rules out |
|---|---|
| Geometry column CRS is lon/lat **and** degree-based (`srtext` matches `^GEOG(CS\|CRS)` *and* `UNIT["degree`) | A projected CRS, where X is metres or feet and 300 is not out of range: a 3857 table with X of 100 and 200 satisfies **every** numeric condition below, and only this check stops it. The degree half rules out the 14 stock SRIDs that are `GEOGCS` in **grads** (4807 NTF Paris and relatives), where a full circle is 400, so -360 is not a whole turn. Added in review, see #899 codex r1. |
| Table-wide min X `>= 0` | A source already in -180..180. A single negative longitude proves the convention; a source mixing both is ambiguous, so it is left alone rather than guessed at. |
| Table-wide max X `> 180` | A dataset legitimately confined to the eastern hemisphere. Without this, every Africa/Europe/Asia dataset ending at lon 180 gets flung into -360..-180. |
| Table-wide max X `<= 360` | Coordinates that are not the 0..360 convention at all (wrong units, corrupt data). The clamp keeps handling those. |

Then, **per row**, only geometry whose own min X is `>= 180` is translated. A row already inside -180..180 keeps its exact coordinates.

### On the "0..200" near miss

Worth being explicit, because it reads two ways. A 0..200 source **is** a 0..360 source by the detection rule (all lons >= 0, at least one > 180), so it is recognised. What must not happen is a *whole-table* shift: that is the #883 failure in miniature, where an unguarded shift split a lon-0 buffer at the **prime** meridian into parts covering 49x the correct area. `test_dataset_spanning_0_to_200_shifts_only_out_of_range_rows` pins the correct behaviour: lon 0 stays 0, lon 170 stays 170, lon 200 becomes -160, and all three features survive.

The cases that are not shifted at all are covered too: ends exactly at lon 180, spans -10..170, a projected CRS with X between 180 and 360, X past 360, and a grads-based geographic CRS (with a 4326 positive control on identical coordinates).

## Warning channel

Reused, not invented. `IngestJob.user_metadata['warnings']` already carries `reserved_rename` and `dbf_truncation_collision` through the whole path, so `mercator_clip` was added as a third `kind` at every existing layer:

- producer TypedDict + `make_mercator_clip_warning()` in `backend/app/processing/ingest/warnings.py`
- `_append_mercator_clip_warning()` in `tasks_common.py`, next to the existing `_append_job_warning`
- Pydantic `MercatorClipWarning` in `backend/app/platform/jobs/schemas.py` and the parse branch in `jobs/router.py`, so `JobStatusResponse` validates it at the wire boundary
- `MercatorClipBody` in `IngestWarningsBanner.tsx`, which already renders on the import success screen and the dataset detail page
- copy in all four locales, with `_one`/`_other` plural forms matching the existing convention in these files

Emission sits at the three production clip call sites: `_finalize_ingest` (file + service ingest), `reupload_file`, and `reupload_service`. `_ingest_vector_into_staging` is not wired because nothing in `app/` calls it — it exists as the staging-pipeline unit-test boundary.

`make_mercator_clip_warning()` returns `None` for a no-loss clip, so the "warn only when the user actually lost data" decision lives in one place instead of at each call site.

## Deliberate non-goals

- **The shift is not surfaced as its own user-facing warning.** It is loss-free normalization, the same class as the SRID reprojection the pipeline already does silently. It is recorded in the structured log (`shifted_longitudes`, `rows_shifted`, `min_x`, `max_x`) for operators.
- **A feature that straddles lon 180 in a 0..360 source is not split.** Splitting it needs the antimeridian machinery tracked by #884/#886 (`render_dateline_safe` is the closest existing tool, but it gates on planar span > 180 deg, which a Pacific-only 200..300 source does not meet). Such a feature stays put, gets clipped, and is counted in the warning — legible rather than silent, which is what this issue is about.
- **The naive extent folds at `metadata.py` ~358 and ~888 are untouched.** They belong to #886. There is one interaction worth flagging, pinned in `test_pacific_crossing_source_is_preserved_with_a_naive_extent`: a Pacific-crossing 0..360 source (say 150..250) now lands half on each side of the antimeridian, which is correct per feature, but a bare `ST_Extent` over that result measures -170..170 — a 340 deg bbox for a 100 deg footprint. The old behaviour had a tidy extent only because it had deleted half the data. Preserving the data is the right trade; the extent is #886's remit, and the pinning test is where the two fixes meet.

- `backend/openapi.json`, both SDKs, and `frontend/src/types/api.generated.ts` are regenerated: `JobStatusResponse.warnings` gains a third discriminated-union member. Additive only, no removals in the generated output.
- No migration, no config, no new env var.
- `spatial_ref_sys` is read unqualified, matching the `geometry_columns` lookup already in this function. `PUBLIC` holds `SELECT` on it and `public` is in the default search_path, verified on the dev DB. The multi-tenant before-hook sees no tenant schema in that statement so it runs as the session login; the data-modifying CTE starts with `WITH ... UPDATE`, which `_statement_requires_writer` already classifies as a write, so it binds the tenant writer role.
- Cost: one extra sequential scan (`ST_Extent`) per vector ingest that has a lon/lat geometry column. The projected-CRS check runs first (single-row lookup) so projected ingests skip it. The pipeline already makes several full passes over the same freshly-written table.

## LOC ratchets

Two exact ratchets in `test_layering.py` needed a written carve-out, both documented inline at the cap:

- `processing/ingest/metadata.py` 1716 -> 1856. Roughly half the added lines are the guard write-up, kept next to the code because #883 showed how easily a guard like this gets re-simplified by the next reader. The last +23 is the review round: the angular-unit condition and the note that the envelope bounds X too.
- `processing/ingest/tasks_common.py` 1606 -> 1631, for the `StagingResult.mercator_clip` field and the `_append_mercator_clip_warning` emitter. The emitter exists partly because `reupload_file` is already at the C901 complexity ceiling (15), so an inline `if` there fails `ruff check`.

## Verification

Every command below was run from this branch's worktree against Postgres on localhost:5434 (PostGIS 3.6).

**New backend tests (23 cases, 18 DB-backed):**

```
$ uv run pytest tests/test_ingest_mercator_clip.py -p no:randomly -q --no-cov
23 passed, 43 warnings in 7.93s
```

**Full backend suite** (not just the new file):

```
$ uv run pytest -n 4 -q --tb=short --no-cov
5978 passed, 80 skipped, 267 warnings in 567.36s (0:09:27)
EXIT=0
```

**Backend lint/format:**

```
$ uv run ruff check . && uv run ruff format --check .
All checks passed!
847 files already formatted
```

**Pre-commit (all files):**

```
$ pre-commit run --all-files
SSRF Rule 2 ... Passed
Visibility Rule 1 ... Passed
AGENTS.md — no bare, unscoped finding markers in source comments ... Passed
ruff check (backend) ... Passed
ruff format --check (backend) ... Passed
check json / yaml / toml / secrets ... Passed
```

**OpenAPI + SDK drift gates:**

```
$ make openapi-check      # dump_openapi.py --check
(exit 0)

$ make sdks-check         # regenerate + git diff --exit-code -- sdks/ + npm run build && npm test
✔ 4 SDK tests pass, no diff after regeneration
```

**Frontend gates:**

```
$ npm run types:check
(exit 0, no diff in src/types/api.generated.ts)

$ npm run test:i18n
✓ All static source-referenced i18n keys exist in the en bundles.
Test Files  1 passed (1)   Tests  4 passed (4)

$ npm run check:i18n:toast-strings
No hardcoded toast string literals found for: error, success, info, warning, message.

$ npm run lint
(exit 0)

$ npm run typecheck        # tsc -b --noEmit
(exit 0)

$ npm run test:coverage
Test Files  356 passed (356)
Tests  4126 passed (4126)
Lines : 73.51% ( 13604/18506 )
EXIT=0
```

**Version gate:**

```
$ make version-check
CHANGELOG.md: '## [1.6.1]' section present (47 lines).
(exit 0)
```

No Playwright/e2e run: the dev stack bind-mounts the main checkout, so e2e from a worktree would exercise main's code rather than this branch.

Closes #888

## Review round (codex r1)

Both P2 findings were real and were reproduced against the dev DB before being fixed, in `f6fbfeb6e`:

1. **Grads-based geographic CRSs were admitted by the shift.** 14 SRIDs in a stock `spatial_ref_sys` are `GEOGCS` with `UNIT["grad"]` (the Paris-meridian family, 4807 and relatives) against 999 degree-based ones, and 4807 passed the old prefix-only check. A grads table reaching past 180 would have had those rows translated by -360 grads. Guard condition 1 now requires degrees, with a grads test plus a 4326 positive control on identical coordinates.
2. **The warning copy blamed latitude for losses the X bound caused.** The envelope bounds longitude at ±180 too, and the suite already contains a point at lon 400 / lat 0 that the clip drops. The copy in all four locales now names both axes, and `MercatorClipDetail`'s description carries the same correction so SDK and OpenAPI consumers do not re-import the latitude-only framing.


ishiland:
@codex review

ishiland:
Both P2 findings were real; each was reproduced against the dev DB before I touched anything, and both are fixed in f6fbfeb6e.

**Restrict longitude shifts to degree-based CRSs** — confirmed. Probed `spatial_ref_sys` on PostGIS 3.6: **14** SRIDs match `^GEOG(CS|CRS)` but carry `UNIT["grad"]` (4305, 4807, 4810, 4811, 4816, 4819, 4821, 4901, …, the Paris-meridian family), against 999 that are degree-based. 4807 passed the old prefix-only check, so a grads table reaching past 180 would have had those rows translated by -360 grads. Guard condition 1 now also requires the angular unit to be degrees: one pattern covers WKT1 `UNIT["degree"` and WKT2 `ANGLEUNIT["degree"` (the WKT2 spelling contains the WKT1 substring), and the `^GEOG` prefix is what still keeps 3857 out, since its srtext carries `UNIT["degree"` inside a nested `GEOGCS`.

Two tests: `test_grads_based_geographic_crs_is_not_shifted` (4807, not shifted) and `test_degree_based_geographic_crs_is_shifted` (4326 positive control, identical coordinates). The grads test drives `_shift_zero_to_360_longitudes` directly on purpose — through the public entry point it would prove nothing, because `ST_Transform` collapses the global Mercator envelope into a degenerate polygon in 4807 (every X becomes 197.396) and the clip then empties every row regardless. That collapse is pre-existing clip behaviour for exotic CRSs, unrelated to this fix.

**Describe longitude clipping accurately** — confirmed, and worse than "imprecise": the envelope bounds X at ±180 as well as Y at ±85.06, and `test_longitudes_past_360_are_not_shifted` already asserts a drop for a point at lon 400 / **lat 0**. The old copy would have told that user their data was beyond ±85.06° latitude, which is the wrong repair to go looking for. Copy in all four locales now names both axes and points at coordinate order and range. The `MercatorClipDetail` description carries the same correction so SDK and OpenAPI consumers do not re-import the latitude-only framing (openapi.json, both SDKs, and `api.generated.ts` regenerated; docstring-only, 5 added lines each).

Re-verified after the fix:

```
$ uv run pytest -n 4 -q --tb=short --no-cov
5978 passed, 80 skipped, 267 warnings in 567.36s (0:09:27)
EXIT=0

$ uv run ruff check . && uv run ruff format --check .
All checks passed!
847 files already formatted

$ npx vitest run src/components/import/__tests__/IngestWarningsBanner.test.tsx
Test Files  1 passed (1)   Tests  4 passed (4)

$ npm run test:i18n
✓ All static source-referenced i18n keys exist in the en bundles.
Test Files  1 passed (1)   Tests  4 passed (4)

$ npm run typecheck && npm run lint
(exit 0)
```

The `metadata.py` LOC ratchet moves 1833 -> 1856 with the carve-out extended for both fixes.

ishiland:
@codex review

PR #900, opened 2026-07-30
fix(analysis): buffer each component in its own planar projection

### LINKED ISSUE #891: fix(analysis): ST_Buffer(::geography) picks one planar SRID per feature — multipart features far apart in longitude lose up to half their area
Found while fixing the codex P2 on #883 (antimeridian buffer). Unrelated to the antimeridian itself — this is in the raw buffer, before and independent of any seam handling.

`ST_Buffer(geom::geography, d)::geometry` chooses a **single** planar working SRID for the whole input geometry. When one multipart feature has components spread far apart in longitude, every component gets buffered in a projection that suits at most one of them.

Measured ratio of produced area vs. the ideal circle area, two-point MULTIPOINT at increasing separation:

| component separation | area ratio vs ideal |
| --- | --- |
| 0.2° | 0.938 |
| 10° | 0.994 |
| **90°** | **0.498 — half the area** |
| 150° | 0.994 |

The 90° case is the worst because that is where the single chosen projection is maximally wrong for both components. The non-monotonic shape (0.2° also poor, 150° fine) suggests the SRID pick is driven by the geometry centroid or extent midpoint, so the error depends on where that lands relative to the members rather than on separation alone.

Practical impact: a user buffering a multi-region dataset as one feature (e.g. a MULTIPOINT of sites on several continents, or a multipart admin boundary) silently gets buffers that are substantially the wrong size. There is no error and the output is valid geometry, so nothing signals it.

This is why the #883 verification tables compare fixed-vs-raw rather than fixed-vs-ideal — comparing against an ideal circle would have folded this pre-existing error into the antimeridian numbers.

**Fix sketch.** Buffer per component in a projection chosen for that component (`ST_Dump` → per-part `::geography` buffer → re-collect), or pick an equal-area/azimuthal projection per part. #883 already established the per-component dump pattern in `render_dateline_safe()` and measured its cost (~0.4 ms, `SubPlan` only on rows that need it), so the machinery exists to extend.

ishiland:
## What changed

`ST_Buffer(geom::geography, d)::geometry` asks `_ST_BestSRID` for **one** planar working SRID and uses it for the whole input. The plan output makes the shape explicit — PostGIS rewrites the cast chain to

```
st_transform(st_buffer(st_transform(geometry(g::geography), _st_bestsrid(g::geography)), d), 4326)
```

so every component of a multipart feature is buffered in a projection chosen for the *envelope*, not for itself.

`render_geodesic_buffer()` in `backend/app/platform/analysis_sql.py` now dumps the validated source, buffers each component through its own `::geography` cast, re-collects, runs the existing #883 seam split, and dissolves with `ST_UnaryUnion`. `render_geometry_expr("buffer", ...)` is the single chokepoint both the preview (`service_analysis.build_preview_sql`) and the materialize worker (`processing/analysis/tasks.py`) call, so both paths pick it up.

Diff shape: `analysis_sql.py` +144/-12 (one new function; 19 lines of it are SQL, the rest is the docstring), `test_analysis_materialize.py` +260/-0 (6 behavioural tests), `test_analysis_preview.py` +74/-10 (one new SQL-shape test, three updated).

No schema, API, config, i18n or SDK impact — this is a rendered SQL expression behind `render_geometry_expr`, which is already the sole entry point for both the preview and the materialize worker.

## Reproduced first

The issue's table, reproduced verbatim against the dev DB (PostgreSQL 18.4, PostGIS 3.6.4, GEOS 3.13.1) on `main@cb22f5bb1` before any code changed:

```sql
WITH params(sep) AS (VALUES (0.2::float8),(10.0),(90.0),(150.0)),
src AS (
    SELECT sep, ST_SetSRID(ST_Collect(ST_MakePoint(0,45), ST_MakePoint(sep,45)),4326) AS g
    FROM params
),
buf AS (SELECT sep, ST_Buffer(ST_MakeValid(g)::geography, 10000)::geometry AS b FROM src)
SELECT sep AS separation_deg, GeometryType(b) AS gtype, ST_NumGeometries(b) AS parts,
       round(ST_Area(b::geography)::numeric,0) AS area_m2,
       round((ST_Area(b::geography) / (2*pi()*10000^2))::numeric,4) AS ratio_vs_ideal
FROM buf ORDER BY sep;

 separation_deg |    gtype     | parts |  area_m2  | ratio_vs_ideal
----------------+--------------+-------+-----------+----------------
            0.2 | POLYGON      |     1 | 589383310 |         0.9380
             10 | MULTIPOLYGON |     2 | 624289038 |         0.9936
             90 | MULTIPOLYGON |     2 | 313192925 |         0.4985
            150 | MULTIPOLYGON |     2 | 624289038 |         0.9936
```

Exact match to the reported 0.938 / 0.994 / **0.498** / 0.994. The mechanism came straight out:

```sql
SELECT sep, _ST_BestSRID(g::geography) AS whole_srid,
       _ST_BestSRID(ST_GeometryN(g,1)::geography) AS part1_srid,
       _ST_BestSRID(ST_GeometryN(g,2)::geography) AS part2_srid
FROM src ORDER BY sep;

 sep | whole_srid | part1_srid | part2_srid
-----+------------+------------+------------
 0.2 |     999031 |     999031 |     999031
  10 |     999247 |     999031 |     999032
  90 |     999000 |     999031 |     999046   <- 999000 = world Mercator
 150 |     999061 |     999031 |     999055
```

Two things fall out that the issue's table alone does not show. The 0.2° row is **not** a bug — at that separation the two circles overlap and the union is genuinely smaller than 2 π r², which is why it also stays 0.9380 after the fix. And the non-monotonic shape is not centroid-driven: it is a hard threshold, `_ST_BestSRID` dropping to world Mercator once the longitude span reaches 45°, then climbing back into a wide Lambert past ~135°.

`after` numbers below come from substituting the rendered expression — `render_geometry_expr("buffer", distance_meters=10000.0)[0]` with `geom_4326` bound to each fixture — into the same queries.

## Why this is not the antimeridian

Different failure, same expression. The seam bug (#697/#883) is about longitude normalization in the *output*. This one is about the projection chosen for the *input*, and it fires nowhere near ±180. Measured longitude-span sweep of a two-point `MULTIPOINT` at lat 45, 10 km buffer, reading the geodesic distance from each source point to every vertex of the output part containing it:

| longitude span | `_ST_BestSRID` | produced radius (m) |
| --- | --- | --- |
| under 6° | 999031 | 9 997 – 10 004 (single UTM zone) |
| 6° – 44.99° | 999247 | 9 904 – 10 097 |
| **45° and up** | **999000 (world Mercator)** | **7 079 – 7 087** |
| ~135° and up | 999061 | 9 240 – 10 824 (wide Lambert) |

The cliff is exactly at a 45.0° span (44.99° → 999247, 45.0° → 999000), and world Mercator's scale error is `1/cos(latitude)`, which is the whole story: `cos(45°)² = 0.5`.

The gate threshold sits at the *other* measured boundary — 6.0°, where PostGIS leaves the single-UTM-zone regime (5.999° → 999031, 6.0° → 999247), hence `>=` rather than `>`. Picking 6 rather than 45 is deliberate: it removes the ±1% band as well as the catastrophic tail, and it keeps a safety margin so a future PostGIS retune of `_ST_BestSRID` cannot silently reopen the half-area case.

Latitude span is harmless. Held at 0° longitude span and swept to an 80° latitude span, the SRID stayed on the UTM zone and the radius stayed within 9 990 – 10 004 m — transverse Mercator holds scale ≈ 1 along its central meridian at every latitude. So the guard tests longitude span only, the same quantity `render_dateline_safe` already tests.

## Pass ordering

Three passes, and the order is not free:

1. **Per-component buffer.** `ST_Dump` the validated source, buffer each part through its own `::geography`, dump each part's buffer (so `ST_Collect` never mixes `POLYGON` with `MULTIPOLYGON` into a `GEOMETRYCOLLECTION`), `ST_CollectionHomogenize(ST_Collect(...))`.
2. **Seam split** — the existing `render_dateline_safe`, unchanged.
3. **`ST_UnaryUnion`** to dissolve overlapping parts.

**2 before 3.** A component whose buffer wraps ±180 is self-intersecting *in the planar domain*, and unioning it in that state does not degrade gracefully — it aborts the statement:

```
ERROR:  lwgeom_unaryunion_prec: GEOS Error: TopologyException: side location
conflict at 179.90757318430857 44.915684255255684. This can occur if the input
geometry is invalid.
```

That is the same lesson #883 learned with `ST_ShiftLongitude` before `ST_MakeValid` (validate-first noded the seam into 4 slivers holding 97.9% of the area, against 2 parts holding 99.3%): the seam artefact has to be resolved before any GEOS overlay op touches the geometry.

**3 has to exist.** Buffering the whole input dissolved components whose buffers overlap; buffering per component and collecting does not. Measured on `MULTIPOINT((0 45),(0.05 45),(90 45))`, 10 km:

| dissolve | parts | `ST_IsValid` | area (m²) |
| --- | --- | --- | --- |
| none (`ST_Collect` only) | 3 | **false** | 935 908 947 (overlap counted twice) |
| `ST_MakeValid` | 3 | true | **468 078 270** (cut the overlap out of one part) |
| `ST_UnaryUnion` | 2 | true | **701 993 608** ✔ |

701 993 608 is the true answer: 390 028 685 (two overlapping circles unioned) + 311 964 923 (the distant circle). `ST_MakeValid` subtracts where a dissolve should merge, so it is not an alternative. The union is a no-op where nothing overlaps — the #883 seam fixture measures 623 944 052 m² and 3 parts with or without it — which is why it can sit unconditionally on this branch.

## Cost profile

Gated on two cheap conditions on the validated source: `ST_NumGeometries(...) > 1 AND ST_XMax(...) - ST_XMin(...) >= 6`. Single-part input takes a bare `ELSE` that renders exactly the #883 expression.

The guard is free — PostgreSQL short-circuits the `AND`, so single-part rows never reach the bbox scan. Over the same 2 000 66-vertex polygons: `ST_NumGeometries` alone 3.75 ms, the span test alone 7.34 ms, both together **3.60 ms**.

`EXPLAIN (ANALYZE, TIMING OFF)`, median of 3 warm runs, pre-fix expression vs this one, plus a third variant that is the pre-fix expression with only this function's extra `OFFSET 0` fence added, to attribute the delta:

| workload | before | fence only | after |
| --- | --- | --- | --- |
| 2 000 single-part POINT | 53.2 ms | 38.5 ms | **38.4 ms** |
| 2 000 single-part POLYGON (66 vertices) | 334.9 ms | 375.7 ms | **375.6 ms** |
| 2 000 two-part MULTIPOINT, 2° apart (under the gate) | 86.8 ms | — | **75.0 ms** |
| 2 000 two-part MULTIPOINT, 90° apart (gated) | 87.3 ms | — | **116.4 ms** |
| 200 × 100-part MULTIPOINT spanning 99° (gated) | 586.3 ms | — | **1 315.5 ms** |

So the common path's entire cost is the fence, +20 µs/row on 66-vertex polygons; on bare points it is *cheaper*, because materializing `ST_MakeValid` once beats re-deriving it. The gated path pays for the per-part buffers and the dissolve, bounded and only where the answer was wrong.

The last row is the honest worst case: 6.58 ms/row against 2.93 ms/row before, so a hypothetical source of 500 000 hundred-part features spanning 99° would blow the 300 s CTAS budget — as it already did before this change (1 465 s → 3 289 s). It is a 2.2x on a shape that was already past the ceiling, not a newly reachable failure, and `MAX_SOURCE_FEATURES["buffer"]` bounds feature count rather than part count either way.

Structural evidence — one `EXPLAIN ANALYZE` over a 2 000-row mix, 285 of them wide multipart:

```
Seq Scan on _c891_mixed (actual rows=2000.00 loops=1)
  SubPlan 6
    ->  Subquery Scan on _pb (loops=2000)
          SubPlan 3                                    <- per-component buffer
            ->  Subquery Scan on _pb_m (loops=285)
                          ->  Aggregate (loops=285)
                  SubPlan 1  -> (never executed)       <- seam split, no wrap present
          SubPlan 5                                    <- whole-input buffer
            ->  Subquery Scan on _dl (loops=1715)
                  SubPlan 4  -> (never executed)
```

285 + 1715 = 2000. Both `ST_Buffer` call sites are rendered; exactly one is reached per row. On the pure single-part table `SubPlan 3` reads `never executed` outright.

## Before / after

Same fixtures, same 10 km distance, `before` = the expression on `main@cb22f5bb1`. Ratio is against the ideal circle area (`π r²` per component), measured with:

```sql
SELECT label,
   round((ST_Area((<BEFORE>)::geography)
          / (pi()*10000^2*ST_NumGeometries(geom_4326)))::numeric,4) AS before_ratio,
   round((ST_Area((<AFTER>)::geography)
          / (pi()*10000^2*ST_NumGeometries(geom_4326)))::numeric,4) AS after_ratio,
   ST_AsEWKB(<BEFORE>) = ST_AsEWKB(<AFTER>) AS byte_identical
FROM _fixtures ORDER BY label;
```

| fixture | before | after | note |
| --- | --- | --- | --- |
| `MULTIPOINT((0 45),(0.2 45))` | 0.9380 | 0.9380 | byte-identical; the deficit is real overlap, not a bug |
| `MULTIPOINT((0 45),(10 45))` | 0.9936 | 0.9936 | area equal, radius tightened 9 904–10 097 → 9 997/10 003 |
| **`MULTIPOINT((0 45),(90 45))`** | **0.4985** | **0.9930** | the headline case |
| `MULTIPOINT((0 45),(150 45))` | 0.9936 | 0.9930 | equal-area SRID hid it; radius 9 240–10 824 → 9 997 |
| **`MULTIPOINT((0 0),(90 60))`** | **0.6216** | **0.9927** | lat-60 component 5 009 m → 10 000 m radius |
| `POINT(0 45)` | 0.9930 | 0.9930 | **byte-identical** |
| `POINT(0 89.9)` (polar) | 0.9936 | 0.9936 | **byte-identical**, still `POLYGON` |
| `MULTIPOINT((179.95 45),(0 45))` (#883) | 0.9936 | 0.9930 | 3 parts, valid, both centres covered |

The remaining 0.7% is `ST_Buffer`'s own polygonal approximation of a geodesic circle, the same figure #883 measured.

The `MULTIPOINT((0 0),(90 60))` row is why area ratios understate this: 0.6216 reads as a mild deficit, but the equatorial component was exact (Mercator scale 1 at the equator) while the lat-60 component held a **quarter** of its area. The per-component radius is the honest metric, and the new tests assert on it. Same fixtures, geodesic distance from each source component to every vertex of the output part covering it, requested radius 10 000 m:

```
                        |     BEFORE      |      AFTER
 fixture         part   |  min      max   |  min      max
------------------------+-----------------+-----------------
 sep 10 deg        1    |  9904    10097  |  9997     9997
 sep 10 deg        2    |  9971    10030  | 10003    10003
 sep 90 deg        1    |  7079     7087  |  9997     9997
 sep 90 deg        2    |  7079     7087  |  9997     9997
 sep 150 deg       1    |  9253    10814  |  9997     9997
 sep 150 deg       2    |  9240    10824  |  9997     9997
 lat 0 + lat 60    1    | 10000    10000  |  9990     9991
 lat 0 + lat 60    2    |  5009     5016  | 10000    10001
 883 seam fixture  1    |  9253    10813  |  9960     9998
 883 seam fixture  2    |  9253    10814  |  9997     9997
 single point           |  9997     9997  |  9997     9997
 polar 89.9             | 10000    10000  | 10000    10000
```

The `883 seam fixture` rows are worth calling out: at a 179.95° span the whole-input SRID was the wide Lambert, which is equal-area, so #883's area assertion at `rel=0.02` passed while both components were deformed ±8%. Per-component buffering fixes the shape too. (Component 1's 9 960 m minimum is the seam-split part, whose ring includes the ±180 cut edge.)

**Why #883's tables compared fixed-vs-raw rather than fixed-vs-ideal:** this bug was folded into every multipart number, so an ideal-circle baseline would have charged the antimeridian fix with a projection error it did not cause. With #891 fixed, fixed-vs-ideal is now the meaningful comparison, and the new tests use it (`rel=0.01` against `2 π r²`) instead of comparing against unfixed output.

## Verification

```
cd backend && uv run ruff check .          # All checks passed!
cd backend && uv run ruff format --check . # 846 files already formatted
```

```
pytest tests/test_analysis_preview.py tests/test_analysis_materialize.py
  109 passed, 43 warnings in 62.61s

pytest -k "MultipartProjection or Dateline" tests/test_analysis_materialize.py
  11 passed, 58 deselected in 13.50s     # 5 pre-existing #883 + 6 new

pytest -n 4                              # FULL backend suite
  5960 passed, 81 skipped, 264 warnings in 708.40s (0:11:48)
```

Every #883 fixture re-measured against the new expression and unchanged:

| #883 fixture | type | parts | valid | span | hits France |
| --- | --- | --- | --- | --- | --- |
| `POINT(179.95 45)` @10 km | MULTIPOLYGON | 2 | t | −180..180 | f |
| `POINT(0 45)` @10 km | POLYGON | 1 | t | 0.2534 | f |
| `POINT(0 85.06)` @100 km | POLYGON | 1 | t | 20.8158 | f |
| `POINT(0 89.9)` @100 km (polar) | POLYGON | 1 | t | 347.3388 | f |
| `MULTIPOINT((179.95 45),(0 45))` @10 km | MULTIPOLYGON | 3 | t | −180..180 | f |

Edge cases probed for GEOS exceptions and type drift — all valid, none raising: pole-encircling component plus a distant one, components at both −179 and +179, a `GEOMETRYCOLLECTION` with an `EMPTY` member, mixed-dimension collection, `GEOMETRYCOLLECTION EMPTY`, `MULTIPOINT EMPTY`, duplicate coincident points, and spans of exactly 5.999° / 6.0°.

## Not fixed here

A **single** component wider than one UTM zone is still buffered in one non-local projection — there is nothing to dump it into. A 10 km buffer of `LINESTRING(0 45, 90 45)` produces 85 485 981 084 m² against 134 146 143 319 m² for the same line segmentized to 20 km and buffered piecewise, so 63.7% of the correct area, identical before and after this change. Closing that means subdividing *within* a component, which changes the vertex structure of every large single-part output — its own decision, worth a follow-up issue.

`processing/ai/sql_generator.py` still teaches the unnormalized `ST_Buffer(geom_4326::geography, 1000)::geometry` pattern in the NL→SQL prompt; that is #886 and deliberately untouched here.

Closes #891

### DIFF
--- a/backend/app/platform/analysis_sql.py
+++ b/backend/app/platform/analysis_sql.py
@@ -85,6 +85,14 @@ MASK_SUBDIVIDE_MAX_VERTICES = 256
 DATELINE_WRAP_SPAN_DEG = 180
 
 
+# A geography buffer projects its WHOLE input into one planar SRID chosen by
+# PostGIS's ``_ST_BestSRID``, and that choice is local to every component only
+# while the input fits inside a single UTM zone — 6° of longitude. Measured on
+# PostGIS 3.6, the switch away from the UTM zone happens exactly at a 6.0° span,
+# so the guard in ``render_geodesic_buffer`` tests ``>=``, not ``>``.
+BUFFER_LOCAL_SRID_SPAN_DEG = 6
+
+
 def render_dateline_safe(geom_expr: str, *, alias: str = "_dl") -> str:
     """Split antimeridian-wrapping output of ``geom_expr`` at ±180.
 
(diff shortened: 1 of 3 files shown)


ishiland:
@codex review

Issue #903, opened 2026-07-30
fix(frontend): no guard for a west>east bbox — a dozen map consumers would misrender a seam-crossing extent


ishiland:
Surfaced by PR #901 (#892/#884). **No open issue owned this**, and it is the reason #901 could not move the map-facing extent fields to the spec form.

## Context

#901 made `records.spatial_extent` able to hold an antimeridian-crossing extent (two-ring MULTIPOLYGON) and added `extent_to_bbox()`, which returns the RFC 7946 §5.2 / STAC `west > east` pair for that shape. It then had to split every consumer into two groups: **4 spec sites** (STAC/OGC record `bbox`, the two OGC collection extents, the AI prompt) and **6 span sites** routed to a new `extent_to_span_bbox()` that returns monotonic `-180..180` instead.

The span sites are a deliberate retreat, not the destination. They exist because the frontend cannot currently handle an inverted bbox anywhere.

## The three unguarded consumers

**1. `normalizeRasterBounds` — `frontend/src/.../layer-adapters/shared.ts:60`.** No `bounds[0] > bounds[2]` check. This is the keystone: until it has one, the backend cannot serve the spec form to any map surface.

**2. `DatasetMap.tsx:569-596`** builds a planar ring with no ordering guard, so a `west > east` pair would **draw** the complementary 340° rectangle over the map. (Corrected from an earlier `:563` / `extent_bbox` citation: the destructure is at `:569`, the ring at `:571-596`, and the component's prop is `bbox`, declared at `:84` — `extent_bbox` is the API field name that feeds it, not the local.)

**3. `lib/map-extent.ts:21`** — `isLargeExtent` computes `maxx - minx`, which goes **negative** for a crossing bbox. That routes an inverted extent into `fitBounds` instead of the large-extent `flyTo` path.

Two further consequences found while verifying, both worse than a missed auto-fit:

- `map-sync.ts:1007` feeds the same value into a MapLibre vector/raster source `bounds`. An inverted pair matches **no tile**, so a seam-crossing layer renders **blank**.
- The builder's two fit-bounds paths and Zoom to Layer skip an inverted bbox, so both silently become no-ops (this is the `use-builder-layers.ts:438` site already noted in #886's related list).

## Why this is worth doing rather than leaving the span fallback

[...]

PR #925, opened 2026-07-30
fix(catalog): fold extent rollups on the circle instead of across the globe

### LINKED ISSUE #886: fix(catalog): extent rollups manufacture a global bbox across the antimeridian

ishiland:
## What changed

Two ordinary Fiji datasets — one at lon 179, one at lon −179 — rolled up to `-180..180`. No individual record wraps; the min/max fold does, and it takes the long way round. 359° of bbox for two footprints 2° apart, served as the OGC collection extent, the STAC Collection extent, the collections API `extent_bbox`, and the centre of an AI-generated map.

Five sites did the same naive fold in two different shapes. They now share one helper set in `backend/app/core/geo.py`:

| site | before | after |
|---|---|---|
| `modules/catalog/search/router.py` `_build_collection_metadata` | `ST_AsGeoJSON(ST_Envelope(ST_Collect(...)))` + GeoJSON coordinate fold | `rollup_bbox_columns()` → `rollup_bbox()` |
| `modules/catalog/collections/service.py` `batch_collection_extents` | same shape, plus `_iter_coord_pairs` | `rollup_bbox_columns()` → `rollup_span_bbox()` |
| `standards/stac/router.py` `_fetch_extents` (listing) | 4× `ST_XMin/YMin/XMax/YMax(ST_Extent(...))` | `rollup_bbox_columns()` |
| `standards/stac/router.py` `_fetch_extent` (detail) | same | `rollup_bbox_columns()` |
| `processing/ai/service.py` viewport snap | Python fold over `to_shape(ext).bounds` | `merge_bboxes()` in the extracted `_snap_viewport_to_extent()` |

New in `core/geo.py`: `rollup_bbox_columns()` (six aggregate columns), `rollup_bbox()` (spec form), `rollup_span_bbox()` (monotonic), `merge_bboxes()` (the Python twin, for folds that already hold per-record bboxes), `wrap_longitude()`.

## How it works

Aggregate in two longitude domains — as stored, and `+360`-shifted — and keep whichever range spans less. A tie keeps the unshifted domain, so nothing that does not actually cross the seam is ever re-expressed as `west > east`. The shifted domain naturally produces the RFC 7946 §5.2 crossing pair when that is the honest answer.

**The shift cannot be a bare `ST_ShiftLongitude`.** It shifts each vertex with `x < 0`, which is right for the two-ring seam form (`150..180` + `-180..-110` → a contiguous `150..250`) and wrong for a footprint crossing the prime meridian: Europe's `-10..30` becomes the vertex pair `(350, 30)`, i.e. a claimed span of `30..350`, and a rollup that then wins on span emits `30..-10` — **a bbox that excludes the Europe dataset itself.** Measured against the live PostGIS 3.6:

```
Europe (-10..30) + Fiji pair
  guarded                  domain=shifted bbox=(  178.50,   30.00) span=211.50  covers-all-inputs=True
  bare ST_ShiftLongitude   domain=shifted bbox=(   30.00,  -10.00) span=320.00  covers-all-inputs=False

UK (-10..0) + Fiji pair
  guarded                  domain=shifted bbox=(  178.50,    0.00) span=181.50  covers-all-inputs=True
  bare ST_ShiftLongitude   domain=shifted bbox=(    0.00,  -10.00) span=350.00  covers-all-inputs=False

Africa (-18..52) + Fiji pair
  guarded                  domain=shifted bbox=(  178.50,   52.00) span=233.50  covers-all-inputs=True
  bare ST_ShiftLongitude   domain=shifted bbox=(   52.00,  -18.00) span=290.00  covers-all-inputs=False
```

So footprints that reach the prime meridian are translated whole (`ST_Translate(g, 360, 0)`) instead. Translating preserves a footprint's own span, so it can never invent a range narrower than the data has; the worst it does is leave the shifted domain too wide to win, which falls back to the unshifted domain.

### Float noise cannot flip the domain

Brute-forcing the covering property over 400 random footprint sets (including prime-meridian crossers, seam-form records and global extents) turned up a case with a **single** input: an ordinary Europe-shaped extent whose emitted bbox came back with drifted edges.

`(x + 360) - 360` is not always `x`, so the two domains disagree on the same footprint in the last bits. For `-4.761789777127049..33.035082006073026`:

```
normal span   37.796871783200075
shifted span  37.79687178320006     <- smaller, so the shifted domain won
emitted west  -4.761789777127035    <- drifted, on an extent that never crosses the seam
```

Every Europe/Africa/UK extent was a candidate. The shifted domain must now beat the unshifted one by more than `_DOMAIN_MARGIN = 1e-9` degrees (~0.1 mm, the same floor `_SEAM_TOL` already uses), so **non-crossing catalogs are byte-identical to before** and a genuine narrowing clears the margin by four to eleven orders of magnitude. Pinned by `test_float_noise_cannot_flip_a_prime_meridian_extent` and `test_a_real_narrowing_still_beats_the_margin`.

The residual is documented rather than corrected: when the shifted domain legitimately wins, its edges have still been through the round-trip, so an edge can sit up to `ulp(512)/2` (~6e-14 degrees) inside the true union. Re-running the brute force:

```
checked=800 sql_not_covering=39 py_not_covering=39
worst sql shortfall = 0.000e+00 deg
worst py  shortfall = 0.000e+00 deg
```

Those 39 fail the bitwise `ST_CoveredBy` predicate while PostGIS measures the geometric gap as exactly zero. Correcting it soundly means widening every crossing edge unconditionally, which puts the same noise into the exact cases for no operational gain. (The SQL and Python folds agreeing 39/39 is also a free cross-check that the two implementations match.)

### Documented ceiling

Two cut points (−180 and 0), not the largest gap anywhere on the circle. Footprints at `-170`, `-20`, `20` and `165` have their largest gap at `-160..-20` and could be covered in 220°; both cut points fall inside data, so this returns 325°. **Always a valid covering range, sometimes broader than optimal — never inverted, never partial.** Closing that needs a real largest-gap-on-a-circle search; it is deliberately not in this change and is pinned by `test_documented_ceiling_is_covering_but_not_minimal`.

Worked examples, measured against PostGIS:

```
two Fiji datasets (179 / -179)          normal=(-179.50, 179.50) span=359.00 | shifted=(178.50, 181.50) span=  3.00 | -> shifted west= 178.50 east=-178.50
records at 0 and 170 (tie -> normal)    normal=(   0.00, 170.00) span=170.00 | shifted=(  0.00, 170.00) span=170.00 | -> normal  west=   0.00 east= 170.00
#899 pacific 150..250                   normal=(-170.00, 170.00) span=340.00 | shifted=(150.00, 250.00) span=100.00 | -> shifted west= 150.00 east=-110.00
seam-form single record 150..-110        normal=(-180.00, 180.00) span=360.00 | shifted=(150.00, 250.00) span=100.00 | -> shifted west= 150.00 east=-110.00
genuinely global single polygon          normal=(-180.00, 180.00) span=360.00 | shifted=(180.00, 540.00) span=360.00 | -> normal  west=-180.00 east= 180.00
multipolygon straddling 0 + Pacific part normal=(  -5.00, 110.00) span=115.00 | shifted=(355.00, 470.00) span=115.00 | -> normal  west=  -5.00 east= 110.00
```

## Per-consumer: spec form vs span form

`extent_to_bbox`/`extent_to_span_bbox` (#901) established the split; each rollup consumer is decided the same way.

- **STAC Collection `extent.spatial.bbox`** — both queries, through the shared `_parse_extent_row` — **spec (`west > east`)**. STAC inherits RFC 7946 §5.2, and a global bbox on a Fiji collection is exactly the "fabricated global extent" the STAC hardening batch already refused to advertise elsewhere in this router.
- **OGC "datasets" collection `extent.spatial.bbox`** (`search/router.py`) — **spec**. The per-dataset record bboxes in the *same document* already serve the spec form via `extent_to_bbox` (#901, `:798`); a span rollup would make one response self-inconsistent.
- **`CollectionResponse.extent_bbox`** (`collections/service.py`) — **span**. Its only consumer is `CollectionCard` → `BBoxPreview`, which sizes its SVG rect from `maxx - minx` and has no `bounds[0] > bounds[2]` guard yet (**#903**). This also keeps it on the same contract as the sibling per-dataset `extent_bbox` (`datasets/domain/helpers.py`, span since #901), so #903 can flip both together. Crossing rollups stay honestly global here — over-broad, never inverted.
- **AI map viewport** (`processing/ai/service.py`) — **spec, unwrapped for arithmetic**. The span form would keep the `-180..180` bbox and therefore keep snapping the centre to lon 0 (the Gulf of Guinea), so it does not fix the bug. But the existing margin logic (`spec.center_lng < min_x - dx`) assumes a monotonic width, so a crossing `east` is lifted past +180 and the LLM's longitude is brought into the same unwrapped domain before comparison; the resulting centroid is folded back with `wrap_longitude`. The block moved into `_snap_viewport_to_extent()` so it is unit-testable.

## The #899 test

`test_pacific_crossing_source_is_preserved_with_a_naive_extent` (#899, now merged) asserts a bare `ST_Extent` over the shifted *feature table* reads `-170..170` — 340° for a 100° footprint — and names #886 as the owner of that fold. That assertion is about `ST_Extent` over `data.<table>` inside `processing/ingest/metadata.py`, which produces the per-dataset `spatial_extent`; this PR fixes the **rollups over `records.spatial_extent`**, so it does not flip that assertion (and the test still passes unchanged).

The same footprint is covered here instead, in one test that asserts both readings:

```
tests/test_antimeridian_rollup.py::TestSqlRollup::test_shifted_pacific_source_measures_its_own_footprint
  naive ST_Extent  -> (-170.0, 170.0)   # 340 deg, pinned
  rollup_bbox      -> [150.0, -5.0, -110.0, 5.0]   # 100 deg, west > east
```

Making the ingest producer store the two-ring form (so the #899 assertion does flip) means restructuring `metadata.py`'s `get_extent` and the `WITH meta AS (...)` statement to build WKT through `bbox_to_extent_wkt`, which touches the reupload swap and `refresh_dataset_metadata`. That is a separate change — see "follow-ups" below.

## Tests

`backend/tests/test_antimeridian_rollup.py`, 35 tests: the arithmetic (worked examples, tie behaviour, crossing inputs, the prime-meridian guard, the documented ceiling, empty/short/NULL rows), the SQL fold against live PostGIS (Fiji pair, the #899 Pacific footprint, a stored seam-form record, a non-crossing catalog as a regression guard, span form never inverting), the per-surface decisions end-to-end (STAC detail + listing, collections API, the OGC builder), and the AI viewport snap.

Two specific things the test design pins:

- **The negative probe differs in exactly one axis.** `test_rolled_up_bbox_covers_the_data_and_not_the_complement` probes mid-Pacific (`-150..-140`) in the **same `-20..-15` latitude band** as the fixture. A probe that also differed in latitude would pass before and after the fix.
- **Area, not just intersection.** Round-tripped through `bbox_to_extent_wkt`, the rolled-up bbox is 15 deg² and the naive one 1795 deg² — the wrong-complement failure an `ST_Intersects` boolean can miss.

Verified non-vacuous by neutralizing the domain choice (`if False:` in `_narrower_domain`): 17 of the 33 fail, including all three surface tests and the AI snap.

## Verification

```
All checks passed!
851 files already formatted

$ uv run pytest tests/test_antimeridian_rollup.py -q
35 passed, 43 warnings in 6.96s

$ uv run pytest tests/test_layering.py -q
31 passed, 6 warnings in 4.26s

$ uv run pytest -n 4 -q          # full backend suite
6071 passed, 81 skipped, 261 warnings in 585.42s (0:09:45)

$ PYTHONPATH=. uv run python scripts/dump_openapi.py --check
(exit 0 — no snapshot drift; field names and types are unchanged, only values)
```

## Cost

The second aggregate domain adds one `ST_Intersects` + `ST_Translate`/`ST_ShiftLongitude` per row. Measured on 57,799 one-degree footprints against the same PostGIS:

```
rows=57799
old (one ST_Extent)                    10.3 ms
new (both domains)                     29.5 ms
```

+19 ms on a catalog ~20x larger than any real one, and the catalog-wide rollup is cached for 60 s. The per-collection rollups are grouped over far fewer rows.

## Notes

- **No API shape change.** Same fields, same types; only the values for crossing rollups differ. `openapi.json` and the SDKs are untouched.
- **LOC ratchets lowered, still exact**: `search/router.py` 1432 → 1427, `standards/stac/router.py` 1796 → 1789. Both shrank — the repeated `ST_*(ST_Extent(...))` columns collapse to one splat, and the GeoJSON coordinate fold (with the `#430 BA-20` degenerate-envelope flattening it needed) disappears, since `ST_XMin/ST_XMax` of a degenerate `ST_Extent` is just an equal-min/max bbox.
- **`_seam_split_bbox`'s one-latitude-band contract is not at risk**: none of these five consumers writes a rollup back into a geometry column; all four return float bboxes and the fifth returns a centre point.
- `test_tenant_cache_partitioning.py`'s fake extent row moved from `SimpleNamespace` to a `NamedTuple`, because the row is now sliced positionally as well as read by attribute.
- No frontend changes (#903 owns those).

## Follow-ups found while doing this (not in scope here)

1. **`processing/ingest/metadata.py` still folds the per-dataset extent naively** (`get_extent` at ~`:358` and the `WITH meta AS (...)` statement at ~`:888`). A Pacific-crossing source is stored as a single global `POLYGON`, so every reader downstream is over-broad no matter how good the rollup is. Now unblocked by #901's column widening and by the helpers here; needs `rollup_bbox_columns` + `bbox_to_extent_wkt` and a look at the reupload/refresh paths. This is what would flip the #899 assertion.
2. **`modules/catalog/features/service.py`** (`get_table_extent`, ~`:630`) is the same producer-side fold on the features path and needs the same treatment.
3. **`backend/scripts/seed_tiles.py`** reads `ST_XMin/ST_XMax(r.spatial_extent)` per record, so a seam-form record seeds tiles for the whole world. Harmless (over-seeding, not wrong tiles) but wasteful.
4. **`processing/ai/sql_generator.py:235,339` still needs its own follow-up.** #900 landed a per-component geodesic buffer in `platform/analysis_sql.py`, but the NL→SQL prompt is a separate string surface that still teaches the bare `ST_Buffer(geom_4326::geography, N)::geometry` pattern — so a generated query inherits neither #900's per-component projection (a multipart feature spanning >45° of longitude comes back `cos(φ)` too small) nor #883's seam split. Untouched here.
5. **Propagation sites confirmed already fixed, no edit needed**: every naive `to_shape(...).bounds` read in `standards/dcat*`, `standards/ogc`, `maps/_router_helpers`, `datasets/domain/*` and `processing/tiles/router.py` was redirected onto `extent_to_bbox`/`extent_to_span_bbox` by #901; the only remaining `to_shape(...).bounds` in `backend/app` is the one inside `extent_to_span_bbox` itself. `processing/vector/quicklook.py:~140` folds naively but wants a planar span for a thumbnail canvas, so it is correct as-is.

Closes #886

https://claude.ai/code/session_013weFmN2Jb2BPEpNMMJsUhE

### DIFF
--- a/backend/app/core/geo.py
+++ b/backend/app/core/geo.py
@@ -1,7 +1,9 @@
 from __future__ import annotations
 
+from collections.abc import Iterable, Sequence
+
 from geoalchemy2.shape import to_shape
-from sqlalchemy import and_, func, or_
+from sqlalchemy import and_, case, func, or_
 from sqlalchemy.sql.elements import ColumnElement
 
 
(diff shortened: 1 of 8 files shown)


ishiland:
@codex review

ishiland:
@codex review

Issue #904, opened 2026-07-30
chore(catalog): four comments still describe spatial_extent as POLYGON-only, and alembic check cannot see the new CHECK constraint


ishiland:
Two small leftovers from PR #901 (#892/#884), grouped because both are about the new `spatial_extent` representation being under-documented rather than under-implemented.

## 1. Four comments now assert something false

#901 widened `catalog.records.spatial_extent` from `geometry(Polygon,4326)` to `geometry(Geometry,4326)` plus a `chk_records_spatial_extent_type` CHECK allowing `POLYGON` or `MULTIPOLYGON`. These four comments still tell the reader it is POLYGON-only:

- `backend/app/platform/analysis_sql.py:170-176` — this is the note #892 was *filed from*, so it is the most misleading one left.
- `backend/app/processing/ingest/metadata.py:343`
- `backend/app/processing/ingest/metadata.py:886`
- `backend/app/modules/catalog/features/service.py:620`

No code path is affected: all four write `ST_Extent`-derived POLYGONs, which the CHECK still accepts. They were left alone in #901 because those files were owned by concurrent PRs (#888, #891) at the time.

Worth fixing precisely because the column constraint has been load-bearing folklore in this repo — a wrong comment here is how the next person concludes a two-ring extent cannot be stored.

## 2. The CHECK constraint is invisible to `alembic check`

Alembic does not compare CHECK constraints, so `make alembic-check` will not notice if the model's `CheckConstraint` and migration `0030`'s constraint drift apart. #901 covers the current state with a DB test that asserts the constraint rejects a POINT and accepts both POLYGON and MULTIPOLYGON, but there is no general gate.

That is a small hole in a gate the project otherwise relies on for schema-adjacent changes. Options, cheapest first: accept it and note the limitation next to the constraint; or add a test that reflects every `CheckConstraint` declared on the ORM models and asserts a same-named constraint exists in the migrated DB. The latter generalizes to any future CHECK, which is the only reason it might be worth more than a comment.

## Related nit found at the same time

A zero-height bbox (`south == north`) yields an invalid zero-area ring. Pre-existing on the non-crossing path and unchanged by #901, but the crossing path now matches that behavior rather than fixing it, so both are wrong together. Filed here only so it is written down; it needs its own decision about whether to pad (as `get_extent` already does for degenerate POINT/LINESTRING cases via `ST_Expand`) or reject.

## Operator note worth surfacing somewhere more visible

[...]

Issue #893, opened 2026-07-30
chore: small leftovers from the #871/#874/#877/#755/#697 batch


ishiland:
Minor items surfaced while shipping PRs #879-#883, plus a second batch from #884-#894 (PRs #895, #897-#901). Each is too small for its own issue but real. Grouped so they are not lost; pick off individually.

The two comments below are the original filings and remain as history. This body is the live list — verified against `main` on 2026-07-30.

---

## Closed out

**1. Document the worktree e2e recipe in AGENTS.md.** *Shipped in #926 — AGENTS.md now carries the "Working from a git worktree" section, including the spec-only exception and the `.env.test` note.*

**2. `expansion_zoom` at the ceiling is a no-op click.** *Shipped in #926 — `frontend/src/components/map/cluster-interactions.ts:143-153` guards on `target > currentZoom`, covering both the ceiling case and the overzoomed-parent case that was easing users backwards.*

**3. Published Docker images carry no version metadata.** *Struck — false premise. `publish.yml` passes `docker/metadata-action` labels, whose defaults include `version`; the live registry confirms it on all four images. The real label bug in that area is the opposite one, filed as #932.*

**6. `check_version_coherence.py` docstring is inaccurate.** *Shipped in #926 — the docstring now states outright that `docs-contract.json` is verified by `check_docs_contract.py` instead, and warns against reading the list as the bump's mirror image.*

**8. Cosmetic: duplicate route comments in `App.tsx`.** *Shipped in #926 — one "Redirects from old routes" comment remains, at `frontend/src/App.tsx:135`.*

**10. LOC-ratchet coverage is inconsistent.** *Lifted to #958. It outgrew this list: `analysis_sql.py` is still absent from `_MODULE_LOC_CAPS` and has grown 504 → 522 lines since this item was written, which makes it a live gate gap rather than a nit. `metadata.py` at 1856 lines travels with it as a decomposition candidate.*

---

## Still open

**4. Frontend coverage thresholds are stale.** The comment in `frontend/vite.config.ts:242-253` records 2026-05-07 actuals and warns that a +1 buffer would have failed. Both halves of that warning are now wrong: thresholds are `41/39/37/42` against measured actuals of `71.51/66.77/65.91/73.62`, so there is roughly 30 points of headroom, not zero. The staleness is the whole problem — someone planning coverage work reads this and believes any uncovered line will trip a global threshold. Correct the comment to the current actuals, or ratchet the thresholds up to match and rewrite it once.

[...]

Issue #892, opened 2026-07-30
schema(catalog): records.spatial_extent geometry(POLYGON) cannot express a west>east antimeridian bbox


ishiland:
Recorded as a deliberate non-fix in `render_dateline_safe()` (`backend/app/platform/analysis_sql.py`) while shipping #883. This is the root representation gap behind the antimeridian issue class, and it blocks doing #884 and #886 properly.

**The gap.** `records.spatial_extent` is `geometry(Polygon, 4326)`. RFC 7946 §5.2 and the STAC spec both express an antimeridian-crossing extent as a bbox with **west > east**. A single POLYGON ring in a planar 4326 column cannot represent that — the honest extent of a dataset straddling the seam is either two rings or a west>east pair, and neither fits.

So even with #883 merged, a dataset that genuinely straddles the seam still registers a `-180..180` extent. #883 fixed the corrupt *geometry* (which was the serious half — it produced feature-level false positives 15 000 km away); the *recorded extent* remains globe-spanning.

**This predates the analysis tools** and equally affects directly ingested Fiji-area data — it is not an analysis bug. It also constrains the fixes proposed in:

- #884 (STAC remote import of west>east bboxes) — the natural fix wants to store two rings or a west>east box, and today it can store neither.
- #886 (extent rollups across the seam) — a correct rollup of two Fiji datasets *is* west>east.

**Options, none of them free:**

1. Widen the column to `geometry(MultiPolygon, 4326)` (or plain `geometry`) and store the split rings. Migration touches every record; every reader that assumes a single ring needs an audit. `get_extent` currently pads degenerate POINT/LINESTRING cases specifically to satisfy the POLYGON constraint, so that logic changes too.
2. Keep the geometry column for spatial indexing and add explicit `bbox_west`/`bbox_east` (or a `crosses_antimeridian` flag) as the serving truth for STAC/OGC/DCAT output. Less invasive, but introduces two sources of truth that can drift.
3. Store the shifted (0..360) geometry with a flag. Keeps one column, but every consumer must know to unshift.

Worth deciding once, deliberately, because five or six downstream fixes all depend on which representation wins. Suggest resolving this before starting #884 or #886.

PR #926, opened 2026-07-30
chore: gate mcp/uv.lock, guard cluster click zoom, and correct four spatial_extent comments

### LINKED ISSUE #889: ci(deps): uv lock --check runs only for backend/ — mcp/ has no lock-sync gate
Found while fixing #877. `.github/workflows/dep-audit.yml:34` runs `uv lock --check` with `working-directory: backend` only. `mcp/uv.lock` has no full-lock sync gate at all.

That asymmetry explains the history in #877: `backend/uv.lock` lagging was "known since 1.5.1" because something eventually surfaced it, while `mcp/uv.lock` slipped two whole releases unnoticed.

#877 closed the narrow hole — `make version-check` now asserts the *version fields* in both locks (and `mcp/uv.lock` carries two repo-owned versions: `geolens-mcp` plus the editable `geolens` path dep for `sdks/python`). But a dependency-tree drift in `mcp/uv.lock` is still ungated.

**Fix:** add a `uv lock --check` step with `working-directory: mcp` alongside the backend one. Cheap and symmetrical.

ishiland:
A mixed chore batch: one real gate, one real behavior fix, and four comment corrections. Each item is its own commit so any one can be reverted alone.

## 1. `mcp/uv.lock` had no lock-sync gate (`Closes #889`)

`.github/workflows/dep-audit.yml` ran `uv lock --check` under the job's `working-directory: backend` default, so `mcp/` — the only other project with its own `uv.lock` — was never checked. `make version-check` (#877) asserts the version *fields* in both locks, but a dependency-tree drift in `mcp/uv.lock` stayed ungated, which is how it slipped two releases unnoticed. Added a symmetrical step with `working-directory: mcp`, verified green on the current lock.

## 2. Worktree recipes in AGENTS.md (#893 item 1)

New `### Working from a git worktree` next to the existing single-test recipes. Three facts, all of which cost real time: the dev stack bind-mounts the **main** checkout, so `localhost:8080` serves `main` whatever branch a worktree is on and Playwright run from a worktree validates code nobody wrote; the host-Vite recipe that does work; and the spec-only exception (editing `e2e/*.spec.ts` with no app-code change *is* validly testable in place, which is how #895 was verified). Plus a note that the documented host pytest recipe cannot be `source`d verbatim from a worktree, and the copy-in workaround.

Every claim was checked against the tree rather than transcribed: `API_PROXY_TARGET` in `frontend/vite.config.ts:13`, `E2E_BASE_URL` in `playwright.config.ts:14`, `--reload --reload-dir /app/app` at `docker-compose.yml:336`, the two bind mounts at `:289`/`:555`, `API_PORT=8001` in `.env.example`, and `.env.test` matching `.gitignore`'s `.env.*`.

## 3. Cluster click could ease you backwards (#893 item 2)

`activateClusterFeature` handed `expansion_zoom` straight to `easeTo` with no comparison against the zoom the map is already at.

**The reported case** is the ceiling: at z22 with `cluster_max_zoom=22`, the server's now-honest value (#882) clamps to the zoom the cluster is drawn at, so the click can only pan and the cluster never breaks apart. **Nothing can zoom out of that one** — MapLibre's default `maxZoom` is 22 (the app never overrides it) and the server clusters all the way there, so no zoom exists that splits the cluster. I chose **recentre without zooming**: it is the whole honest response available, and inventing a nudge the map would immediately clamp away would be a lie about what happened. The value of the guard here is that the case is now a deliberate branch with the reasoning attached, instead of an accident of arithmetic.

**The case the guard actually fixes** turned up while checking the first one, and is worse. With the default `cluster_max_zoom=14`, a z14 tile's clusters carry `expansion_zoom: 15` (the `generate_series($1+1, $5)` split search is empty at the max, so it falls through to `cluster_max_zoom + 1`). Zoom to 20 and MapLibre keeps drawing the **overzoomed z14 parent** until the z20 tile lands — and those cluster features stay queryable, so a click reads `expansion_zoom: 15` against a current zoom of 20. The old code eased the user **backwards** to 15, out of the view they were zooming into. `target > currentZoom ? target : currentZoom` makes a cluster click unable to lose ground.

To be exact about the evidence: that is traced through the code and the constants above, and pinned by a unit test that fails without the guard — I did not reproduce it in a browser. The window is also not necessarily brief; if the deep tile 403s (the tile-token race), the overzoomed parent stays on screen.

Also names the repeated `22` as `CLUSTER_ZOOM_CEILING` and routes the three clamps through one helper, so the ceiling case is greppable.

Four cases added to `cluster-interactions.test.ts`. The backwards case is a genuine regression test — with the guard reverted it fails (evidence below); the ceiling case passes either way by design, and is there to pin the decision.

## 4. Stale `spatial_extent` comments (#904, the comment half)

#901 widened `catalog.records.spatial_extent` to `geometry(Geometry,4326)` plus `chk_records_spatial_extent_type`, which admits POLYGON **or MULTIPOLYGON**. Four comments still told the reader it is POLYGON-only — including the one in `platform/analysis_sql.py` that #892 was filed from, which claimed the column "cannot express the west > east bbox". It can now; what remains is the write side.

- `backend/app/platform/analysis_sql.py` — the `_dateline_safe_buffer` note
- `backend/app/processing/ingest/metadata.py` — `get_extent`'s docstring and the SQL comment in the fast-path CTE
- `backend/app/modules/catalog/features/service.py` — `_refresh_count_and_extent`

No behavior change: all four write `ST_Extent`-derived POLYGONs, which the CHECK still accepts. Each comment's actual point — a degenerate POINT/LINESTRING extent is still rejected and must be padded — is kept; only the claim about the column's type is corrected. **Both `metadata.py` edits are line-neutral**, so its exact LOC ratchet (1856) does not move and this stays a comment-only PR.

## 5. Two nits (#893 items 6 and 8)

`check_version_coherence.py` claimed it and `bump_version.py` "enumerate the same set of sites". They differ by exactly one: the bump writes `docs-contract.json`'s `.version`, which `check_docs_contract.py` asserts against `backend/pyproject.toml` instead. No coverage gap, but the docstring reads as a guarantee the file does not make.

`App.tsx` had two consecutive `{/* Redirects from old routes */}` headers with routes between them belonging to the same block. Merged into the one header that also carries the `fix(#871)` note. No route moved relative to another, so matching is unchanged.

## Dropped: #893 item 3 (image version metadata) — the premise is false

The published images already carry `org.opencontainers.image.version`. `publish.yml`'s "Build and push" step passes `labels: ${{ steps.meta.outputs.labels }}`, and `docker/metadata-action`'s default label set includes `version`, `revision`, `created` and `url`. `build-push-action`'s `--label` flags override Dockerfile `LABEL`s, so adding one to the Dockerfile would be shadowed on the publish path and empty everywhere else — strictly worse than absent. Verified against the real registry, all four images:

```
$ docker buildx imagetools inspect ghcr.io/geolens-io/geolens-api:1.6.1 --format '{{ json .Image }}'
### geolens-api
  org.opencontainers.image.title = 'geolens'
  org.opencontainers.image.version = '1.6.1'
  org.opencontainers.image.description = 'PostGIS-native GIS data catalog API'
### geolens-worker
  org.opencontainers.image.title = 'geolens'
  org.opencontainers.image.version = '1.6.1'
  org.opencontainers.image.description = 'PostGIS-native GIS data catalog worker'
### geolens-backup
  org.opencontainers.image.title = 'geolens'
  org.opencontainers.image.version = '1.6.1'
  org.opencontainers.image.description = 'Automated pg_dump backup service with S3 offload'
### geolens-frontend
  org.opencontainers.image.title = 'geolens'
  org.opencontainers.image.version = '1.6.1'
  org.opencontainers.image.description = 'PostGIS-native GIS data catalog frontend'
```

An image pulled by digest is therefore identifiable, and no new version site is introduced — which also means nothing needed registering in `bump_version.py` / `check_version_coherence.py`.

That same output turned up a real defect worth its own issue, left out of this PR: **`org.opencontainers.image.title` is `geolens` on all four images**. The Dockerfile sets a distinct per-image title (`geolens-api`, `geolens-worker`, `geolens-backup`, `geolens-frontend`), but `metadata-action` defaults `title` to the repository name and its label wins, so the Dockerfile's four titles are silently discarded. `description` survives only because `publish.yml` sets it explicitly from `matrix.description`; `title` is not in that list. One line to fix (`org.opencontainers.image.title=${{ matrix.image }}` in the `meta` step) but it changes published metadata, so it wants its own issue.

## Issue status

`Closes #889`.

**#893 and #904 stay open** — both are grouped issues and only part of each is addressed here.

- #893: items **1, 2, 6, 8** done. Item **3** dropped as a false premise (above). Items **4, 5, 7** (coverage thresholds, low-zoom cluster tile latency, `version-check` observability) untouched — they need decisions, not edits. Items **9–15** from the follow-up comment untouched.
- #904: the **four comments** are done. The `alembic check` cannot see CHECK constraints half, the zero-height-bbox nit, and the `downgrade()` operator note for RUNBOOK.md remain.

## Verification

Backend suite, full, from this worktree — run twice, once before rebasing onto #925 and once after (the count moves because #925 added tests):

```
$ uv run pytest -n 4 -q      # base a6570a69b
6036 passed, 81 skipped, 264 warnings in 608.92s (0:10:08)

$ uv run pytest -n 4 -q      # base 0176bfce2 (#925)
6069 passed, 81 skipped, 264 warnings in 601.34s (0:10:01)
```

```
$ cd backend && uv run ruff check .
All checks passed!
$ uv run ruff format --check .
851 files already formatted
```

```
$ python3 scripts/check_version_coherence.py
OK: all 19 version sites agree on 1.6.1.
...
  CHANGELOG.md: '## [1.6.1]' section present (47 lines).
```

The new gate, on this PR's own CI run (job `Python audit (uv)`, 31s, pass):

```
Verify lockfile in sync — MCP server  Run uv lock --check
Verify lockfile in sync — MCP server  Using CPython 3.14.4
Verify lockfile in sync — MCP server  Resolved 40 packages in 1ms
```

A gate nobody has seen fail is not a gate, so: add one unlocked dependency to `mcp/pyproject.toml` and it does fail, then passes again once reverted.

```
$ cd mcp && uv lock --check          # with "tabulate>=0.9" added to dependencies
Resolved 41 packages in 327ms
The lockfile at `uv.lock` needs to be updated, but `--check` was provided.
EXIT=1

$ cd mcp && uv lock --check          # restored
Resolved 40 packages in 4ms
EXIT=0
```

```
$ cd frontend && npm run lint
> eslint .
$ npm run typecheck
> tsc -b --noEmit
$ npx vitest run src/components/map/__tests__/cluster-interactions.test.ts
 Test Files  1 passed (1)
      Tests  9 passed (9)
$ npm run test:coverage
 Test Files  358 passed (358)
      Tests  4157 passed (4157)
Statements   : 71.51% ( 15116/21137 )
Branches     : 66.77% ( 12306/18428 )
Functions    : 65.91% ( 3583/5436 )
Lines        : 73.62% ( 13616/18493 )
```

Sidebar on #893 item 4, which I did not touch: the thresholds are `41/39/37/42`, so the headroom is ~30 points, not the "~0 by design" the `vite.config.ts` comment describes. That comment records 2026-05-07 actuals of `41.51/39.42/37.99/42.69`; the suite has since gained roughly 30 points on every dimension. The item's premise (stale) holds; its stated consequence (a global threshold trips on any uncovered line) has not been true for a while.

The new backwards-ease case proves it is a regression test, not decoration — with `const zoom = target > currentZoom ? target : currentZoom` reverted to `const zoom = target`:

```
× fix(#893): never eases backwards when a stale parent tile reports a shallower expansion zoom 2ms
AssertionError: expected "vi.fn()" to be called with arguments: [ ObjectContaining {"zoom": 20} ]
      Tests  1 failed | 8 passed (9)
```

No API surface moved, so `make openapi-check` / `make sdks-check` are not implicated. No Playwright run: this branch is not spec-only, so per the very recipe item 2 documents, e2e from a worktree would have tested `main`.

### DIFF
--- a/.github/workflows/dep-audit.yml
+++ b/.github/workflows/dep-audit.yml
@@ -32,6 +32,14 @@ jobs:
           cache-dependency-glob: "backend/uv.lock"
       - name: Verify lockfile in sync
         run: uv lock --check
+      # fix(#889): mcp/ is the other project with its own uv.lock, and it had no
+      # full-lock gate at all — which is why it silently lagged two releases
+      # while the backend lock's drift eventually surfaced. `make version-check`
+      # (#877) asserts only the version FIELDS in both locks; this catches
+      # dependency-tree drift.
+      - name: Verify lockfile in sync — MCP server
+        working-directory: mcp
+        run: uv lock --check
       - name: Export locked prod requirements
         run: uv export --locked --no-dev --format requirements-txt -o /tmp/reqs.txt
       - name: pip-audit (OSV + PyPI advisory DB)
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -25,6 +25,14 @@ The React/Vite frontend is in `frontend/src/`: `components/`, `pages/`, `hooks/`
 - Frontend: `cd frontend && npx vitest run src/path/foo.test.ts`.
 - E2E: `npx playwright test e2e/foo.spec.ts --project=chromium` (stack must be running).
 
+### Working from a git worktree
+
+The dev stack bind-mounts the MAIN checkout (`./frontend` → `/app`, and `backend/app` → `/app/app` with `--reload`), so `localhost:8080` always serves `main` no matter which branch your worktree is on. Running `npx playwright test` from a worktree therefore validates code you did not write. Treat the result as meaningless: it has produced a false FAILURE, and the symmetric case is worse, because a worktree change that breaks e2e passes when the stack never had it. To exercise worktree app code, run Vite on the host at `:5174` with `API_PROXY_TARGET=http://localhost:8001` and point `E2E_BASE_URL` at it.
+
+One exception: a spec-only change (editing `e2e/*.spec.ts` with no app-code change) is validly testable from a worktree, because Playwright reads the specs from your worktree while the app code stays byte-identical to `main`.
+
+The host backend recipe above is also unrunnable verbatim from a worktree — the sandbox refuses `source` on a path outside the worktree and denies reading `.env*`. Copy `.env.test` into the worktree instead (it is gitignored; delete it when you are done) and run pytest from a wrapper script.
+
 ## Architecture
 
 Services (`docker-compose.yml`): Nginx (prod proxy; Vite proxy in dev) fronts the FastAPI `api` (catalog, search, OGC/STAC, vector tiles) and Titiler (COG raster tiles). A `worker` runs GDAL/ogr2ogr ingestion, dispatched via the Procrastinate job queue that lives *inside* PostgreSQL (no separate broker). PostgreSQL 18 (PostGIS + pgvector + pg_trgm) is the single source of truth; object storage is MinIO/S3; Valkey is the tile/query cache.
--- a/backend/app/modules/catalog/features/service.py
+++ b/backend/app/modules/catalog/features/service.py
@@ -617,12 +617,14 @@ async def _refresh_count_and_extent(
     Returns (feature_count, extent_wkt) in a single query instead of the
     5 queries that extract_metadata() runs.
     """
-    # fix(#430 BA-18): records.spatial_extent is a POLYGON column, but ST_Extent of a
-    # single point / axis-collinear points casts to POINT / LINESTRING, which the
-    # column rejects (previously the caller silently skipped storing it, leaving a
-    # stale/NULL extent). ST_Expand always returns the bounding-box POLYGON, so we
-    # pad ONLY the degenerate (non-polygon) cases into a valid sub-mm-padded
-    # polygon; genuine polygon extents are returned byte-identical (no epsilon).
+    # fix(#430 BA-18): records.spatial_extent admits only POLYGON or MULTIPOLYGON
+    # (fix(#892) widened the typmod to geometry(Geometry, 4326) and moved the type
+    # guard into chk_records_spatial_extent_type), but ST_Extent of a single point /
+    # axis-collinear points casts to POINT / LINESTRING, which is still rejected
+    # (previously the caller silently skipped storing it, leaving a stale/NULL
+    # extent). ST_Expand always returns the bounding-box POLYGON, so we pad ONLY the
+    # degenerate (non-polygon) cases into a valid sub-mm-padded polygon; genuine
+    # polygon extents are returned byte-identical (no epsilon).
     result = await session.execute(
         text(
             f"SELECT COUNT(*), "
--- a/backend/app/platform/analysis_sql.py
+++ b/backend/app/platform/analysis_sql.py
@@ -176,12 +176,14 @@ def render_dateline_safe(geom_expr: str, *, alias: str = "_dl") -> str:
     over 2 000 rows keeps ``ST_Buffer`` at one evaluation per row.
 
     NOT fixed here, and deliberately: a dataset that genuinely straddles the
-    seam still registers a -180..180 ``spatial_extent``, because the
-    ``geometry(Polygon, 4326)`` column cannot express the west > east bbox that
-    RFC 7946 § 5.2 and the STAC spec use for antimeridian-crossing extents.
-    That representation gap predates the analysis tools and equally affects
-    directly ingested Fiji-area data; it needs a schema change, not a wider
-    expression.
+    seam still registers a -180..180 ``spatial_extent``. The column can now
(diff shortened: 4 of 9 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #928, opened 2026-07-30
fix(catalog): stop float noise flipping a rollup into the shifted domain

ishiland:
Follow-up to #925 (#886). That PR merged while this fix was still being verified, so it lands separately.

## The bug

Brute-forcing #925's covering property over 400 random footprint sets — prime-meridian crossers, seam-form records, global extents — turned up a failure with a **single** input: an ordinary Europe-shaped extent whose emitted bbox came back with drifted edges.

`(x + 360) - 360` is not always `x`, so `_narrower_domain` compares two spans that describe the *same* footprint and disagree in the last bits:

```
extent        -4.761789777127049 .. 33.035082006073026
normal span   37.796871783200075
shifted span  37.79687178320006     <- smaller, so the shifted domain won
emitted west  -4.761789777127035    <- drifted, on an extent that never crosses the seam
```

Nothing here crosses the antimeridian and nothing needs the shifted domain, but the bbox changes anyway. Every Europe / Africa / UK extent was a candidate, which makes it a quiet output change for ordinary catalogs rather than a crossing-only edge case.

## The fix

Require the shifted domain to beat the unshifted one by more than `_DOMAIN_MARGIN = 1e-9` degrees before preferring it. That is ~0.1 mm, the same floor `_SEAM_TOL` already uses in this module: four to eleven orders of magnitude above the noise (~6e-14) and far below any real narrowing. Non-crossing catalogs are byte-identical to pre-#925 again.

## It is not a Python-side quirk — the SQL path has it too

`ST_Translate(g, 360, 0)` rounds exactly like Python's `x + 360`, so PostGIS hands back the same skewed pair. Running the *production* `rollup_bbox_columns` expressions against a real geometry column holding that extent:

```
PostGIS returned:
  unshifted xmin/xmax : -4.761789777127049 33.035082006073026
  shifted   xmin/xmax : 355.23821022287297 393.035082006073
  unshifted span      : 37.796871783200075
  shifted   span      : 37.79687178320006
  bare '<' would pick shifted? True
  with margin picks shifted?   False

what the bare comparison WOULD have emitted:
  west=-4.761789777127035 east=33.035082006073026
what rollup_bbox emits now:
  west=-4.761789777127049 east=33.035082006073026
  byte-identical to the stored extent: True
```

Both paths funnel through the one `_narrower_domain` comparison, so the single margin covers both — this is not a Python-only fix. `test_postgis_shows_the_same_last_bit_asymmetry` pins the SQL half in CI: it inserts the extent, asserts PostGIS's own aggregates disagree in the direction that matters (`sxmax - sxmin < xmax - xmin`), and then asserts the emitted bbox is the stored extent byte-for-byte. A hand-built row cannot show that the aggregates themselves disagree, which is why the DB test earns its place next to the unit test.

## What is documented rather than corrected

When the shifted domain legitimately wins, its edges have still been through the round-trip, so an edge can sit up to `ulp(512)/2` (~6e-14 degrees) inside the true union. Re-running the brute force after the fix:

```
checked=800 sql_not_covering=39 py_not_covering=39
worst sql shortfall = 0.000e+00 deg
worst py  shortfall = 0.000e+00 deg
```

Those 39 fail the bitwise `ST_CoveredBy` predicate while PostGIS measures the geometric gap as exactly zero with `ST_Distance`. The only sound correction is widening every crossing edge unconditionally, which would push the same noise into the exact cases (`178.5` becoming `178.49999999999999`) for no operational gain — the stored geometries are float64 too, so their own edges are fuzzy at the same scale, and no consumer does an exact containment test against a rollup. The docstring now says so, and the SQL and Python folds agreeing 39/39 is a free cross-check that the two implementations still match.

## Tests

Three additions to `backend/tests/test_antimeridian_rollup.py` (33 → 36):

- `test_float_noise_cannot_flip_a_prime_meridian_extent` — asserts the noise exists (`(east + 360) - (west + 360) < east - west`) and that the emitted bbox is byte-identical to the input rather than the round-tripped approximation.
- `test_a_real_narrowing_still_beats_the_margin` — a 0.1-degree genuine improvement still wins, and pins the residual at `< 6e-14`.
- `test_postgis_shows_the_same_last_bit_asymmetry` — the DB half: PostGIS's own aggregates disagree the same way, and the extent survives untouched.

## Verification

```
All checks passed!
851 files already formatted

$ uv run pytest tests/test_antimeridian_rollup.py -q
36 passed, 43 warnings in 6.90s

$ uv run pytest -n 4 -q          # full backend suite, rebased onto 8e63c095f
6072 passed, 81 skipped, 261 warnings in 561.34s (0:09:21)
```

No API shape change, no migration, no LOC-cap movement (this touches `core/geo.py` and a test file only).

Refs #886

### DIFF
--- a/backend/app/core/geo.py
+++ b/backend/app/core/geo.py
@@ -151,6 +151,17 @@ def bbox_to_extent_wkt(west: float, south: float, east: float, north: float) -> 
 # tear apart (see _shifted_longitude_geom).
 _PRIME_MERIDIAN_WKT = "LINESTRING(0 -90,0 90)"
 
+# fix(#886): how much narrower the shifted domain must be before it is preferred.
+# Adding and subtracting 360 is not bit-exact, so the two domains disagree by up
+# to ~3e-14 degrees on the SAME footprint: `(-4.761789777127049 + 360) - 360`
+# comes back as `-4.761789777127035`, and the shifted span of an ordinary
+# prime-meridian-crossing extent (Europe, Africa, the UK) measured
+# 37.79687178320006 against a normal 37.796871783200075. Without a margin that
+# noise wins the comparison and rewrites a non-crossing bbox with drifted edges.
+# 1e-9 degrees is ~0.1 mm -- far above the noise, far below any real gain, and
+# the same floor _SEAM_TOL already uses.
+_DOMAIN_MARGIN = 1e-9
+
 
 def wrap_longitude(lng: float) -> float:
     """Fold a longitude from the shifted domain back into ``[-180, 180]``.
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

ishiland:
@codex review

PR #962, opened 2026-07-30
ops(db): set temp_file_limit so a spill fails the query, not the cluster

### LINKED ISSUE #959: ops(db): set temp_file_limit so an analysis spill fails the query instead of the cluster
Split out of #696 (analysis resource-governance knobs), where it was the only item that prevents an outage rather than tuning a soft limit, and it was blocked behind three unrelated design questions (a Settings-schema change, a `work_mem` ceiling policy, and new concurrency machinery). This one is a single line plus a RUNBOOK note.

## Problem

`temp_file_limit` is unset in `db/postgresql.conf`, which means unlimited. Verified on `main` at 28e205ef2: `grep temp_file_limit db/postgresql.conf` returns nothing.

`work_mem` is 8MB (`db/postgresql.conf:93`), so anything larger than that spills to disk. A grouped dissolve over millions of rows spills the entire geometry column to `pgsql_tmp`, and `pgsql_tmp` lives inside the data directory, which in our compose layout is the same volume as `pgdata` (`docker-compose.yml:172`, `pgdata:/var/lib/postgresql`, a single mount since the PG18 change in #704).

There is no separate temp tablespace, so a large enough spill fills the volume holding the database. Postgres does not degrade gracefully when its data volume fills; it stops. The 300s `MATERIALIZE_TIMEOUT` does not help, because the disk fills before the statement times out.

With a limit set, the same query fails with `temporary file size exceeds temp_file_limit` and everything else keeps running. That is the entire point of this change: it converts a cluster-wide outage into one failed query with a clear message.

## The change

One line in the Memory block of `db/postgresql.conf` (currently `:85-95`), next to `work_mem`:

```
temp_file_limit = 4GB       # per-session ceiling on spill to pgsql_tmp
```

Rationale for 4GB rather than a smaller number: the limit is **per session**, not global, and it exists to stop a runaway, not to tune throughput. It has to sit above any legitimate analysis spill or it becomes a source of spurious failures. A dissolve or buffer that genuinely needs more than 4GB of temporary files is already past the point where `MAX_SOURCE_FEATURES` (`backend/app/platform/analysis_sql.py:48`) should have rejected it at enqueue, so hitting this ceiling is a signal worth surfacing rather than accommodating.

Worth adding in the same change, since it makes the limit diagnosable instead of mysterious:

```
log_temp_files = 64MB       # log any spill above this, to see them coming
```

Sizing note for anyone revisiting the number: `max_connections = 80` (`db/postgresql.conf:11`), so the theoretical worst case is far above any real disk. The per-user materialize cap (`router_analysis.py:279-292`) means one analysis CTAS per user at a time, which is what actually bounds concurrent spills in practice.

## Verification

The limit is not exercised by any normal test, so verify it deliberately:

1. `docker compose up -d --wait db`, then `docker compose exec db psql -U geolens -c "SHOW temp_file_limit;"`, expecting `4GB`, confirming the mounted conf took effect rather than being silently ignored.
2. Force a spill past the ceiling in a scratch session and confirm the error is `temporary file size exceeds temp_file_limit (SQLSTATE 53400)` and that the server is still accepting connections afterwards. A large `SELECT ... ORDER BY` over `generate_series` is enough; it does not need real data.
3. Confirm `log_temp_files` produces a log line for a spill that stays under the ceiling, so operators see pressure before they see failures.

## Acceptance criteria

- [ ] `temp_file_limit` set in `db/postgresql.conf` with a comment explaining that it converts an outage into a query error.
- [ ] `log_temp_files` set, or a note in this issue recording why it was left out.
- [ ] `SHOW temp_file_limit` confirms the value is live in the running container, not just present in the file.
- [ ] A forced overrun returns SQLSTATE 53400 and leaves the cluster serving.
- [ ] RUNBOOK.md documents the error signature under monitoring (§4), so an operator meeting `temporary file size exceeds temp_file_limit` in the logs knows it is a deliberate guard and knows the two responses: filter the dataset, or raise the ceiling if the disk supports it.
- [ ] Existing test suite unaffected; this should be invisible to every current test.

## Not in scope

The other three items from #696 (scoped `work_mem` for materialize, promoting the timeout constants to Settings, a global preview concurrency bound) stay there. This issue deliberately does not touch application code.

ishiland:
## What changed

- `db/postgresql.conf`: `temp_file_limit = 4GB` and `log_temp_files = 64MB` in the Memory block, with a comment explaining the guard.
- `RUNBOOK.md` §4 (Monitoring): new subsection documenting the SQLSTATE 53400 error signature (quoted verbatim from a live PG18 container) and the two operator responses — filter the dataset, or raise the ceiling and `pg_reload_conf()`.

## Why

`pgsql_tmp` lives on the same volume as `pgdata` (single mount since the PG18 change, #704) and `temp_file_limit` was unset (unlimited). A grouped dissolve spilling past `work_mem` (8MB) could fill the data volume and stop the whole cluster. With the ceiling, the same query fails with a clear error and everything else keeps running. 4GB sits above any legitimate analysis spill; the limit is per session.

No application code touched. Invisible to every existing test.

## Verification

Ran the worktree conf in a scratch `geolens-db` container (`postgres -c config_file=...`, same as compose):

- `SHOW temp_file_limit` → `4GB`; `SHOW log_temp_files` → `64MB` (conf took effect, not silently ignored).
- Forced overrun (`SET temp_file_limit='32MB'` + large `ORDER BY` over `generate_series`) → `ERROR: temporary file size exceeds "temp_file_limit" (32768kB)`; `SELECT 1` on a fresh connection succeeded immediately after — cluster still serving.
- Under-ceiling spill produced `LOG: temporary file: path "base/pgsql_tmp/..."` lines in `$PGDATA/log/` via `log_temp_files`.

Post-merge on a live deployment: `docker compose exec db psql -U geolens -c "SHOW temp_file_limit;"` — the setting is reloadable, so `SELECT pg_reload_conf()` suffices without a restart.

Closes #959

### DIFF
--- a/RUNBOOK.md
+++ b/RUNBOOK.md
@@ -402,9 +402,14 @@ only the offending session failed. Two responses:
   area of interest, fewer features) before rerunning.
 - **Raise the ceiling**: if the disk has the headroom and the workload is
   legitimate, increase `temp_file_limit` in `db/postgresql.conf` and reload
-  (`docker compose exec db psql -U geolens -c "SELECT pg_reload_conf();"`).
-  The limit is per session, so size it against a single worst-case query, not
-  the sum of connections.
+  (`docker compose exec db psql -U "$POSTGRES_USER" -c "SELECT pg_reload_conf();"`).
+  The limit is enforced **per session (backend process)**, so several
+  sessions spilling at once can each use the full ceiling — size it against
+  available disk headroom divided by the number of concurrent spill-heavy
+  sessions you expect, not against one query in isolation. In practice the
+  per-user materialize cap keeps analysis at one CTAS per user at a time, so
+  the concurrency to budget for is the number of simultaneously active
+  analysis users plus any ad-hoc sessions.
 
 `log_temp_files = 64MB` logs every spill above 64MB, so temp-file pressure is
 visible in the database logs before it reaches the ceiling — a rising number of
--- a/RUNBOOK.md
+++ b/RUNBOOK.md
@@ -432,6 +432,15 @@ only the offending session failed. Two responses:
   leaves your version in place, so re-check it against the release after any
   upgrade that mentions a Postgres setting.
 
+  Two cases where the upgrade will not do it for you, both needing the
+  recreate command above by hand. The upgrader excludes itself from the
+  release-file sync so it is never swapped under itself mid-run, so an install
+  still running an upgrader from before this behaviour shipped skips the
+  config sync entirely; it picks it up on the following upgrade. And a
+  customised file is left alone by design. Either way, `SHOW
+  temp_file_limit;` after an upgrade tells you whether the running cluster has
+  the release's value.
+
   The limit is enforced **per process**, not per query: every session
   spilling at once gets the full ceiling, and a parallel query spills from
   each worker process separately (`max_parallel_workers_per_gather = 1` in
--- a/RUNBOOK.md
+++ b/RUNBOOK.md
@@ -427,8 +427,10 @@ only the offending session failed. Two responses:
   Verify either way with `SHOW temp_file_limit;`.
 
   `scripts/upgrade.sh` syncs `db/postgresql.conf` from the target release and
-  recreates the db container when it changed, but only if you have not edited
-  the file. A customised config is never overwritten — the upgrade warns and
+  recreates the db container when it changed, but only while the file still
+  matches the release you are upgrading *from*. The comparison is on file
+  contents, not on `git status`, so tuning you committed or staged counts as
+  tuning: a customised config is never overwritten. The upgrade warns and
   leaves your version in place, so re-check it against the release after any
   upgrade that mentions a Postgres setting.
 
--- a/RUNBOOK.md
+++ b/RUNBOOK.md
@@ -402,13 +402,24 @@ only the offending session failed. Two responses:
   area of interest, fewer features) before rerunning.
 - **Raise the ceiling**: if the disk has the headroom and the workload is
   legitimate, increase `temp_file_limit` in `db/postgresql.conf` and reload
-  (`docker compose exec db psql -U "$POSTGRES_USER" -c "SELECT pg_reload_conf();"`).
-  The limit is enforced **per session (backend process)**, so several
-  sessions spilling at once can each use the full ceiling — size it against
-  available disk headroom divided by the number of concurrent spill-heavy
-  sessions you expect, not against one query in isolation. In practice the
+  from inside the container, where Compose has already set the credentials
+  (single quotes on purpose — the variables must expand in the container
+  shell, not on the host, where `.env` values are not exported):
+
+  ```bash
+  docker compose exec db sh -c \
+    'psql -U "$POSTGRES_USER" -d "$POSTGRES_DB" -c "SELECT pg_reload_conf();"'
+  ```
+
+  The limit is enforced **per process**, not per query: every session
+  spilling at once gets the full ceiling, and a parallel query spills from
+  each worker process separately (`max_parallel_workers_per_gather = 1` in
+  the bundled config, so budget up to 2 processes per query). Size the
+  ceiling against available disk headroom divided by the peak number of
+  spilling processes — concurrent spill-heavy sessions times (1 + parallel
+  workers per gather) — not against one query in isolation. In practice the
   per-user materialize cap keeps analysis at one CTAS per user at a time, so
-  the concurrency to budget for is the number of simultaneously active
+  the session count to budget for is the number of simultaneously active
   analysis users plus any ad-hoc sessions.
 
 `log_temp_files = 64MB` logs every spill above 64MB, so temp-file pressure is
--- a/RUNBOOK.md
+++ b/RUNBOOK.md
@@ -379,6 +379,37 @@ targets just need network reach to `api:8000` and `worker:8001`.
 > the `geolens_db_pool_*` gauges are not emitted and `GeoLensDbPoolSaturated`
 > never fires — pool health is the provider's responsibility there.
 
+### Database temp-file ceiling (`temp_file_limit`)
+
+The bundled Postgres config sets `temp_file_limit = 4GB` (`db/postgresql.conf`).
+Temporary spill files (`pgsql_tmp`) live on the same volume as the database
+itself, so an unbounded spill from a single runaway query — typically a large
+analysis dissolve or buffer — could fill the data volume and stop the whole
+cluster. The ceiling is a deliberate guard that converts that outage into one
+failed query.
+
+If you see this in the database logs or as a query error:
+
+```
+ERROR:  temporary file size exceeds "temp_file_limit" (4194304kB)
+```
+
+(SQLSTATE `53400`) — the guard worked as designed. The cluster keeps serving;
+only the offending session failed. Two responses:
+
+- **Filter the dataset**: the query genuinely needed more than 4GB of temp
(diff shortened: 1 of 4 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
Rebased onto current `main` (the two `Merge branch 'main'` commits are gone; the three content commits are unchanged from the tree codex last passed clean on at `210bd8f76e`). Re-running review against the new head so the clean pass is anchored to the exact sha being merged.

@codex review

ishiland:
Fixed the finding on `932abfed66` — you were right that the guard never reached an existing install.

`db/postgresql.conf` is now synced from the target tag in `scripts/upgrade.sh`, with one deliberate constraint: it is only overwritten when it still matches the installed checkout. That file is one operators are told to tune (this PR's own RUNBOOK section says to raise `temp_file_limit` in it), so silently replacing a customised config during an upgrade would be a worse regression than the one being fixed. A locally edited file now produces a warning with merge instructions and is left alone.

The activation step is a **recreate**, not a restart or a reload, and this matters more than it looks: `git checkout` writes a new inode, and a single-file bind mount keeps resolving the one the container started with, so `pg_reload_conf()` would re-read the old contents and appear to do nothing. It runs after `compose pull` and before the migrate step, while the old app containers are the ones about to be replaced anyway, and waits on the healthcheck (`compose up -d --force-recreate --no-deps --wait db`).

That same inode trap was in the RUNBOOK recipe this PR adds — "edit the file and reload" silently fails for any editor that saves by rename, which is most of them. The raise-the-ceiling step now says recreate, keeps the reload as the in-place-edit case, and points at `SHOW temp_file_limit;` to verify.

Verified: `sh -n scripts/upgrade.sh` clean; `scripts/tests/test-upgrade-order.sh` 46 passed, 0 failed; `test-upgrading-doc-consistency.sh` 15 passed; `check_env_doc_drift.py` and `check_docs_contract.py` pass. Rebased onto current main.

@codex review

ishiland:
**:300 (P2) — stale HEAD comparison.** Correct and now fixed. A path-restricted checkout writes the index and worktree but leaves HEAD at the install tag, so from the second upgrade onward the previous upgrader's own file would read as a customization and the config would freeze permanently. The two tests are now separated and both index-relative:

- *needs changing*: worktree vs the target tag's blob (`git show $TAG:db/postgresql.conf` into a temp file, `cmp -s`)
- *operator edited*: worktree vs the index (`git diff --quiet -- db/postgresql.conf`)

Neither depends on HEAD, so repeated upgrades keep working, and a config already equal to the target is a no-op rather than a needless container bounce.

Also extended `test-upgrade-order.sh`: the compose stub records a `--no-deps db` recreate as `db_recreate` rather than `app_up` — otherwise the existing "migrate precedes app up" and "no app_up after a failed migrate" assertions would read a config bounce as bringing the app back — and a new case pins the bounce between the pull and migrate. 47 passed, 0 failed.

**:299 (P1) — the installed upgrader runs, not the new one.** True, and I am not fixing it here, because it is not specific to this change. `scripts/upgrade.sh` deliberately excludes itself, `restore.sh`, and `install.sh` from the target-tag checkout so the script is never swapped under itself mid-run (that exclusion and its rationale are at lines 267-270 and predate this PR). Every improvement ever made to the upgrader therefore takes effect one release later; that is the accepted cost of the self-exclusion design, not a defect introduced by adding a file to the sync list.

The two alternatives both look worse to me:

- Bootstrapping a newer upgrader mid-run reintroduces exactly the swap-under-itself hazard the exclusion exists to prevent, and it is a change to the upgrade contract that does not belong in a PR about a Postgres GUC.
- Delivering `temp_file_limit` through the compose `command:` (a path the old flow does refresh) would reach existing installs one release sooner, but a command-line `-c` overrides the config file, which would silently defeat the operator tuning this PR's own RUNBOOK section tells people to do in `db/postgresql.conf`. Trading permanent untunability for one release of latency is the wrong trade.

What I did instead is make the gap visible and actionable: the RUNBOOK now names both cases where the upgrade will not apply the config for you (an older upgrader, or a customized file), gives the one-command manual fix, and points at `SHOW temp_file_limit;` to check what the running cluster actually has.

Verified: `sh -n scripts/upgrade.sh` clean; `test-upgrade-order.sh` 47 passed; `test-upgrading-doc-consistency.sh` 15 passed; `check_env_doc_drift.py` passes. Rebased onto current main.

@codex review

ishiland:
Fair, and fixed. `git diff` (staged or not) is the wrong instrument here — an operator who version-controls their tuning is precisely the one most likely to have it, and to git they look pristine.

The customization test is now content-based and git-status-blind: **unmodified** means "the worktree file matches the blob at the release you are upgrading *from*" (`git show v$CURRENT_VERSION:db/postgresql.conf`, `cmp -s`). That is true whether the operator's edit is unstaged, staged, or committed, and it also keeps the stale-HEAD property from the last round, since `CURRENT_VERSION` comes from `.env` and the upgrader maintains it. The current tag is fetched best-effort, and if it cannot be resolved the code falls through to the keep-your-version branch — the failure mode is never clobbering.

The test harness now actually models the file rather than stubbing around it: the git stub answers `show <tag>:db/postgresql.conf` per tag and its `checkout` writes the target content into the worktree, and `seed_prod_env` seeds the installed version with an override argument. Two new cases:

- `seed_prod_env 'temp_file_limit = 64GB # tuned by the operator'` → the file still contains 64GB after the upgrade, no `db_recreate`, and the "keeping your version" warning is printed.
- default seed → synced to the release's 4GB and the bounce fires.

I checked these are real guards rather than decoration: restoring the `git diff --quiet` form makes exactly those three customization assertions fail (`48 passed, 3 failed`), with the clobbered file reported as `temp_file_limit = 4GB`. With the fix in place, `test-upgrade-order.sh` is 51 passed, 0 failed.

RUNBOOK updated to describe the actual rule — matches the release you are upgrading from, compared on contents rather than `git status`.

@codex review

ishiland:
Good catch, and it pushed me to a better design than the one you suggested. Deciding this from git state cannot be made correct — after a failed attempt the version in `.env` may or may not have been bumped, so an installed-blob comparison gets the retry right in one ordering and wrong in the other.

The container is the authority on what is live, so I ask it: compare the worktree file against `/etc/postgresql/custom.conf` as the running `db` container is serving it. Any mismatch means the release's config is on disk but not in effect, whatever git thinks. It is gated on the on-disk file already being the release's (`DB_CONF_AT_TARGET`), so an operator's own uncommitted tuning is still never applied as a side effect of upgrading, and it degrades to "no bounce" when there is no `db` service (external/managed Postgres) or the file cannot be read.

This subsumes the whole class: interrupted-then-retried, synced-this-run, and already-applied all resolve from one comparison.

Two new cases in `test-upgrade-order.sh` (the stub now answers `compose exec -T db cat`, with `DOCKER_DB_RUNNING_CONF` modelling a container still on the pre-upgrade inode):

- disk at `4GB`, container still serving `0` → recreates (your scenario)
- disk at `4GB`, container already serving `4GB` → no bounce

Both are real guards: disabling the live comparison fails them (`49 passed, 4 failed`, including the normal sync case). With it, `test-upgrade-order.sh` is 53 passed, 0 failed. `sh -n` clean on both files; `test-upgrading-doc-consistency.sh` 15 passed; `check_env_doc_drift.py` passes.

Not rebased onto the newest main in this push — there is a merge-contention hold in effect and I am not competing for the window; the branch is a fast-forward from `f247c0430`.

@codex review

PR #963, opened 2026-07-30
fix(tiles): classify degree CRSs from stored WKT, not epsg == 4326

### LINKED ISSUE #939: fix(tiles): maxzoom derivation tests epsg == 4326, so every other geographic CRS reads degrees as metres (15 zoom levels too high)
Found during the post-merge smoke check of #924, in live data. Not a regression from that PR — pre-existing.

## The defect

`_native_resolution_meters` (`backend/app/processing/tiles/router.py:406`) decides whether a raster's stored resolution is in degrees or metres with an exact equality test:

```python
if asset.epsg == 4326:
    values.extend(_degrees_resolution_to_meters(res_x, res_y, bounds))
else:
    # GeoLens raster ingest normally stores COGs in meter-based CRSs
    values.extend(v for v in (res_x, res_y) if v is not None)
```

`4326` is not the only degree-based CRS. Every other geographic CRS falls into the `else` branch and has its **degrees read as metres**, which collapses the derived native resolution by ~5 orders of magnitude and pins `maxzoom` to the `_MAX_RASTER_MAXZOOM` cap.

## Measured, on the live dev stack

The affected dataset is real, not hypothetical: **ETOPO 2022 Global Relief (60 arc-second)** is stored as `epsg=9518`, `res_x=res_y=0.016666…`, 21600x10800. Its tile token returns `maxzoom: 22`.

Running the real function with one input varied — only the EPSG changes:

```
ETOPO fixture: 60 arc-second (1/60 deg) global, 21600x10800
  epsg=4326  native_res_m=1855.33      maxzoom=7     <- correct
  epsg=9518  native_res_m=0.02         maxzoom=22    <- shipped behaviour
  epsg=4269  native_res_m=0.02         maxzoom=22
  epsg=4979  native_res_m=0.02         maxzoom=22
  epsg=4258  native_res_m=0.02         maxzoom=22
```

**15 zoom levels too high.** The client is told to keep requesting tiles to z22 for a raster that holds no detail past z7, so every level beyond that is upsampled mush plus wasted TiTiler work and cache pressure.

Note this is the **inverse** of the symptom #887 fixed. There a seam-crossing raster measured 360° wide and *lost* five levels of maxzoom; here a non-4326 geographic raster *gains* fifteen. Same root category, opposite direction.

## Affected CRSs

Any geographic CRS other than literally 4326, including several that are common rather than exotic:

- **4269** — NAD83 geographic, ubiquitous in US federal and state data
- **4258** (ETRS89), **4283** (GDA94)
- **4979** (WGS 84 3D), **9518** / **7912** / **9057** — WGS 84 and ITRF realizations, which is what modern GDAL increasingly writes

Local dev data has one today (`epsg=9518`), because the ETOPO import produced it without anyone doing anything unusual.

## Fix

Test the CRS's own units rather than its code. The principle from #924 holds — never an EPSG allow-list, never a unit *name* alone (names vary by PROJ build). But the helper that principle produced is the wrong one to reach for **here**, so this section is worth reading before starting.

### Not `_is_degree_based`

`_is_degree_based` (`backend/app/processing/raster/vrt.py:245-261`) is the most rigorous test in the codebase: it reads `crs.units_factor` and compares radians-per-unit against `math.pi / 180`, which is exactly what excludes grads (0.01571, a full 10% away).

It takes a **pyproj/rasterio CRS object**. `_native_resolution_meters` has no CRS object — it has `asset.epsg`, an integer. Adopting it means constructing `CRS.from_epsg(asset.epsg)` per call on a tile-token path, which is a PROJ database lookup this request did not previously need.

### The zero-cost route

`RasterAsset.crs_wkt` is a column on the **same model row** already in hand (`backend/app/processing/raster/models.py:75`, declared alongside `epsg` / `res_x` / `res_y`). No extra query, no extra dependency.

And `wkt_is_geographic(crs_wkt)` already exists at `backend/app/core/geo.py:416`. It was added by fix(#569) for *this exact symptom class* — its docstring reads "the frontend rendered geographic-CRS pixel resolutions as meters ('60 arc-second' ETOPO showed '2 cm')" — and it explicitly handles **EPSG:9518**, naming the COMPOUNDCRS nesting that this issue's live repro lands on. It checks `PROJCRS`/`PROJCS` first, because WKT1 nests a `GEOGCS` inside every `PROJCS`.

### The caveat that has to be handled

`wkt_is_geographic` matches keywords only and performs **no unit check**, so a `GEOGCS` carrying `UNIT["grad"]` returns `True`. That is precisely the blind spot codex flagged during #924, and its docstring's "(degree units)" over-claims.

So pair it with a degree-unit test on the same WKT. The pattern already exists in this repo, in SQL: `backend/app/processing/ingest/metadata.py:1195` and `:1206` define `_GEOGRAPHIC_SRTEXT_RE` (`^GEOG(CS|CRS)`) and `_DEGREE_UNIT_SRTEXT_RE` (`UNIT\["degree`), used together at `:1244-1253`. Mirroring that pairing in Python over `crs_wkt` gives geographic **and** degrees without a PROJ lookup.

Whichever way this goes, fix `wkt_is_geographic`'s docstring or its behavior — leaving a function that claims degree units while admitting grads is how the next site inherits the same bug.

### Test

A test asserting the derived **maxzoom** for a degree-based non-4326 fixture, differing from the 4326 fixture in exactly the EPSG — every wrong answer here is a plausible integer, so asserting "not 22" is not enough. Use the ETOPO shape from the measurements above: 21600x10800 at 1/60°, expecting `maxzoom: 7`. Cover 4269 at minimum, plus a grads-CRS case pinning whatever the chosen behavior is.

## Acceptance criteria

- The ETOPO dataset already in local dev (`epsg=9518`) derives `maxzoom: 7`, not 22.
- 4269, 4258, 4979, and 7912 derive the same maxzoom as the identical raster tagged 4326.
- Projected meter-based CRSs (3857 and friends) are unchanged.
- A grads CRS is handled deliberately — whether it is treated as degrees, rejected, or falls through to the bounds-derived branch, the choice is pinned by a test rather than left to whichever helper was picked.
- No new per-request PROJ database lookup on the tile path.

## Umbrella: four ways this codebase decides "is this CRS in degrees?"

Worth listing, because this issue exists as the fourth one, and the next author should pick deliberately instead of inventing a fifth:

| mechanism | location | tests units? | needs |
| --- | --- | --- | --- |
| `asset.epsg == 4326` | `processing/tiles/router.py:404` | no — this is the bug | an integer |
| `_is_degree_based` | `processing/raster/vrt.py:245` | yes, `units_factor` — most rigorous | a pyproj CRS object |
| `wkt_is_geographic` | `core/geo.py:416` | **no** — keyword only, admits grads | a WKT string |
| srtext regex pair | `processing/ingest/metadata.py:1195`, `:1206`, used `:1244-1253` | yes — geographic AND degree | a SQL round-trip |

They disagree on grads CRSs, and three of the four are reachable from data a single `RasterAsset` row already carries. Consolidating is out of scope here, but a comment at each site pointing at this table would stop the count growing. #906 is the other live instance.

ishiland:
## What changed

`_native_resolution_meters` (`backend/app/processing/tiles/router.py`) decided degrees-vs-metres with `asset.epsg == 4326`, so every other geographic CRS had its degree resolutions read as metres, collapsing native resolution by ~5 orders of magnitude and pinning maxzoom at `_MAX_RASTER_MAXZOOM`. The live ETOPO 2022 dataset (`epsg=9518`, 60 arc-second, 21600x10800) derived `maxzoom: 22` instead of 7.

The decision now reads `RasterAsset.crs_wkt`, a column on the same row already in hand — no per-request PROJ lookup on the tile path:

- `wkt_is_geographic` (existing, `core/geo.py`) classifies geographic vs projected.
- New `wkt_has_degree_unit` (`core/geo.py`) checks the angular unit; it mirrors `_DEGREE_UNIT_SRTEXT_RE` from `processing/ingest/metadata.py` and matches WKT1 `UNIT["degree"` and WKT2 `ANGLEUNIT["degree"` with one pattern.
- Grads GEOGCS (Paris-meridian family, e.g. 4807): geographic but resolutions are neither metres nor degrees, so the stored resolution is skipped and the bounds/width estimate takes over. Pinned by test.
- No stored WKT: falls back to the historical `epsg == 4326` test, so meter-CRS assets without WKT behave exactly as before.
- `wkt_is_geographic`'s docstring no longer claims "(degree units)" while admitting grads, and points to the #939 catalogue of unit tests so a fifth variant does not appear.

## Acceptance criteria from the issue

- ETOPO shape at `epsg=9518` (compound WKT) derives maxzoom 7, not 22 — tested.
- 4269, 4258, 4979, 7912 derive the same maxzoom (7) as the identical raster tagged 4326 — parametrized test asserting the exact value, since every wrong answer is a plausible integer.
- Projected meter CRSs (3857, with and without WKT) unchanged — tested.
- Grads CRS handled deliberately (bounds-estimate fallback) — pinned by test.
- No new PROJ database lookup on the tile path — string checks on the stored WKT only.

No schema or API changes.

## Verification

```
uv run pytest tests/test_wkt_is_geographic.py tests/test_raster_tiles.py tests/test_raster_antimeridian_887.py -v
```

159 raster-tile tests + the wkt suite pass locally.

Closes #939


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
Addressed, but by removing the mechanism rather than narrowing the window again.

Rounds 1-9 here were the same defect in different clothes: a regex reading WKT structure. Each fix narrowed the scan (quoted names, GEODCRS, the WKT2 `CS[` section, the BOUNDCRS `SOURCECRS` window, the target's PRIMEM) and each was followed by another element that legitimately carries an angular unit outside that window — PRIMEM, now MERIDIAN, with BEARING and the conversion PARAMETERs still unclaimed. That sequence doesn't terminate: WKT is a nested grammar, and a flat scan cannot decide which subtree a unit belongs to.

So `wkt_has_degree_unit` now asks PROJ (via rasterio, already a backend dependency) instead of scanning. It resolves the tree, so a BoundCRS reports its SOURCE's units and a geographic 3D CRS reports its axis unit rather than its MERIDIAN's — the whole class at once. The MERIDIAN case in this finding is covered by `test_axis_meridian_unit_does_not_masquerade_as_the_axis_unit`, and `test_real_epsg_wkt_round_trips_through_both_helpers` pins real EPSG WKT for 4326/4269/4258/4979/9518/4807/3857/32633.

The keyword sniff survives only as the fallback for WKT that PROJ rejects, because this helper classifies whatever GDAL wrote at ingest rather than guaranteed-valid CRS text. On that path `wkt_has_degree_unit` returns None (honestly unknown), which the caller's `is not False` test already reads as the historical degrees assumption. Net: ~55 lines of scope-window regex deleted, and the unit comparison now shares a constant and a method with `_is_degree_based` in `raster/vrt.py`.

@codex review

PR #964, opened 2026-07-30
fix(builder): refresh the session JWT on tab return for raster-only maps

### LINKED ISSUE #907: fix(builder): raster/DEM tiles have no proactive JWT refresh, so a raster-only map 401s its whole tile surface once on tab return
Follow-up from #890 (PR #897), which unified the token-refresh policy but left this gap.

Raster and DEM tiles authenticate with the session JWT via the `Authorization` header (`buildTileTransformRequest`, `lib/tile-utils.ts`), not with a signed tile token. Two things follow:

1. `hasExpiringVectorToken` ignores raster tokens entirely — there is no `exp` to inspect, because there is no tile token.
2. `buildTileTransformRequest` attaches the JWT with **no expiry check**, and MapLibre fetches tiles directly rather than through `apiFetch`, so it bypasses `apiFetch`'s proactive refresh (`api/client.ts:142`, which calls `tryRefresh()` within 30 s of access-token expiry).

So a **raster/DEM-only** map returning from a long hidden period still 401s its entire tile surface once before healing. #897 made that failure legible (it now reports once, unsuppressed, instead of double-logging) but did not prevent it.

## Why #897 made this sharper rather than fixing it

Codex flagged during #897 that the raster re-mint looked useless and should be dropped. It was wrong to drop, and the mechanism is worth recording because it is unintuitive: the mint goes `getTileTokensBatch` → `apiFetch` → `authenticatedRawFetch`, which calls `tryRefresh()` and writes the fresh token into the same store the raster `Bearer` header reads. **The "useless" re-mint was the only thing renewing a raster tile's credential.** #897 therefore narrowed the *claim* that raster was handled, while keeping the re-mint on all three surfaces.

That leaves the current behavior: raster credentials are renewed only as a side effect of a vector token mint. A map with no vector layers has nothing to trigger it.

## Suggested fix

Have the #881 visible-edge hook (`useVisibleTileTokenRefresh`) also nudge a refresh when `useAuthStore.getState().expiresAt` is near, independent of whether there are vector tokens to mint. That reuses the existing throttle (30 s cooldown / 10 s settle) and the existing visible-edge trigger, so it should not need new machinery — and it must stay on the visible edge, because MapLibre drops `setTiles` on a paused map (#584).

Worth confirming the failure first on a raster-only map: hide the tab past the access-token lifetime, return, and check whether every DEM/raster tile request 401s once.

## The fix is smaller than the wording above suggests

Verified against `main` at `28e205ef2`. The phrase "no vector tokens to mint" reads as though a raster-only map has no token query to invalidate at all. It does.

`RasterTileToken` is a real token kind (`api/tiles.ts:13-23`, `kind: 'raster'`), returned by the same batch mint as vector tokens, so a raster-only map still has a live `useTileTokens` query. Invalidating it re-runs `getTileTokensBatch` → `apiFetch` → the proactive refresh. **`recover()` therefore already does the right thing on a raster-only map** — the credential gets renewed as a side effect exactly as described above.

The only thing standing in the way is the gate in `useVisibleTileTokenRefresh`:

```ts
if (!hasExpiringVectorToken(getTokensRef.current())) return;
```

and `hasExpiringVectorToken` skips every non-vector token (`if (token?.kind !== 'vector') continue;`). Both are in `hooks/use-tile-auth-recovery.ts`.

So this is a change to the gate condition, not new refresh machinery. Widening it to also fire when `expiresAt` is near is sufficient; calling `tryRefresh()` directly is available as an alternative (it is exported from `api/client.ts:58`) but should not be necessary.

ishiland:
## What changed

Raster and DEM tiles authenticate with the session JWT via `buildTileTransformRequest`'s `Authorization` header, not with a signed tile sig. `hasExpiringVectorToken` only inspects `kind === 'vector'` tokens, so on a map with no vector layers the visible-edge refresh never fired and every DEM/raster tile 401'd once after a long hidden period.

This is a gate-condition change, not new machinery:

- `hasExpiringSession(expiresAt, nowMs)` in `hooks/use-tile-auth-recovery.ts` — the session-JWT half of the same 60 s skew window. `null` (anonymous / embed-token surfaces) is not expiring.
- `useVisibleTileTokenRefresh` fires when either the vector sig **or** the session JWT is inside that window.

The re-mint path is untouched. `RasterTileToken` is a real token kind returned by the same batch mint, so a raster-only map still has a live `useTileTokens` query; invalidating it re-runs `getTileTokensBatch` → `apiFetch` → `tryRefresh()`, which writes the fresh token into the store the raster `Bearer` header reads. The existing 30 s cooldown / 10 s settle still bound the mint rate, and it stays on the **visible** edge because MapLibre drops a `setTiles` reload while the tab is paused (#584).

## Security impact

None beyond the intent: no credential is broadened, and no refresh happens while signed out (`expiresAt === null` is a no-op).

## Verification

    cd frontend && npx vitest run src/hooks/__tests__/use-tile-auth-recovery.test.ts   # 29 passed
    cd frontend && npx vitest run src/components/viewer/__tests__/use-viewer-tokens.test.ts src/components/builder/__tests__/BuilderMap.token-visibility-refresh.test.tsx   # 12 passed

Three new hook tests cover the raster-only map: expiring session re-mints, fresh session does not, and the hidden edge still no-ops.

Closes #907

### DIFF
--- a/frontend/src/components/builder/BuilderMap.tsx
+++ b/frontend/src/components/builder/BuilderMap.tsx
@@ -13,7 +13,7 @@ import {
 } from '@/lib/basemap-utils';
 import { buildClusterTileUrl, buildSignedTileUrl, buildTileTransformRequest, isMvtSourceLayerConfigReady, refreshRasterTileSources } from '@/lib/tile-utils';
 import { useRemoteBasemapStyle } from '@/components/map/hooks/use-remote-basemap-style';
-import { isRasterTileAuthError, logUnhandledMapError } from '@/lib/map-error-log';
+import { isRasterTileAuthError, isRefreshableRasterAuthError, logUnhandledMapError } from '@/lib/map-error-log';
 import { useTileTokens, useInvalidateTileTokens } from '@/hooks/use-tile-token';
 import { isSessionRenewalPending, useTileAuthRecovery, useVisibleTileTokenRefresh } from '@/hooks/use-tile-auth-recovery';
 import { useTileTokenError } from './hooks/use-tile-token-error';
@@ -656,7 +656,10 @@ export const BuilderMap = memo(function BuilderMap({
         // fix(#907): a raster auth failure while a session renewal is running is
         // the visibility race, not a dead session — the post-rotation reload is
         // about to cure it, so it reports like any other transient.
-        const rasterAuthDuringRenewal = rasterAuthError && isSessionRenewalPending();
+        // Only a 401 — a raster 403 is an upstream denial a fresh JWT cannot
+        // cure, so it must still surface.
+        const rasterAuthDuringRenewal =
+          isRefreshableRasterAuthError(e) && isSessionRenewalPending();
         const isClientError = Boolean(status && status >= 400 && status < 500);
         const suppressedClientError = isClientError && (!rasterAuthError || rasterAuthDuringRenewal);
         pushReportEntry({
--- a/frontend/src/components/builder/BuilderMap.tsx
+++ b/frontend/src/components/builder/BuilderMap.tsx
@@ -11,7 +11,7 @@ import {
   toMaplibreStyle,
   BLANK_BASEMAP_ID,
 } from '@/lib/basemap-utils';
-import { buildClusterTileUrl, buildSignedTileUrl, buildTileTransformRequest, isMvtSourceLayerConfigReady } from '@/lib/tile-utils';
+import { buildClusterTileUrl, buildSignedTileUrl, buildTileTransformRequest, isMvtSourceLayerConfigReady, refreshRasterTileSources } from '@/lib/tile-utils';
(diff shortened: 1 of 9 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #967, opened 2026-07-30
chore(report): count distinct failures in the reporter badge

### LINKED ISSUE #908: chore(frontend): the report buffer records one map failure as two rows, so the badge count is doubled
Found during #890 (PR #897). This is the reason a codex finding on that PR was declined rather than fixed — the honest fix is here, not in the raster path.

Every unrecovered map error produces **two** rows in the report buffer:

1. a `maplibre` row pushed by the surface handler, which carries `sourceId`
2. a `console` row from the intercepted `console.error`

`ReportProblemHost`'s badge counts rows, so "2 errors" can mean **1 failure**. This has always been true — a tile 5xx does it on `main` today — and it is not specific to auth errors.

## Which failures actually double the badge

Re-verified against `main` at `28e205ef2`. The two-row write is universal, but the **badge** double-count is not, and the difference matters when reproducing.

`ReportProblemHost.tsx:28` counts only rows with `severity === 'error'`. In `BuilderMap.tsx`, `suppressedClientError = isClientError && !rasterAuthError`, so a 4xx that is not a raster auth error is pushed as `'warning'` **and** suppressed — it never reaches the badge at all. The `console` row is always `'error'` (`lib/report/capture.ts:44-52`).

So the badge doubles for **5xx tile failures and raster 401/403 specifically**. Reproducing with a 404 no-data tile shows a badge of 1 and looks like the bug is already gone.

## Why it was not collapsed in #897

Codex asked #897 to stop recording raster auth failures twice. Collapsing either half locally loses something real:

- drop the `maplibre` row → lose `sourceId`, which is what makes a report actionable
- drop the `console` row → the viewer and dataset preview end up with **no trace at all**, because they push no report row of their own

So #897 kept both and pinned the shape with a test (`records a raster 403 the same way an unrecovered 5xx is recorded`) asserting identical row shapes and hard counts against an unrecovered 5xx — i.e. it pinned raster auth as *consistent with existing behavior* rather than making it a special case.

## The fix

**Primary ask: stop counting rows in the badge and count distinct failures instead.** This is the smaller change with most of the benefit, it needs no correlation key, and it makes the badge mean what a user assumes it means. `ReportProblemHost.tsx:28` is the only place that has to change.

**Stretch: a cross-source dedupe in `pushReportEntry`** — when a `console` row describes the same failure as a `maplibre` row already buffered, merge rather than append. That would fix tile 5xx, raster auth, and style errors in one place, and would also tidy the technical-details view rather than just the count.

The dedupe needs a correlation key, and that is the open question. The `maplibre` row has `sourceId`, the console row has only message text, and `pushReportEntry`'s existing collapse compares `entries[0]` with a matching `source`, so cross-source merging is new behaviour rather than a tweak. Whether those can be matched reliably probably determines whether the stretch is worth doing at all — which is why the count change is the primary ask and not a fallback.

## Related nit from the same PR

`refetchIntervalInBackground` is left implicit at **both** sites in `hooks/use-tile-token.ts` — `:37` in `useTileToken` and `:79` in `useTileTokens` — while only the viewer's hook states it explicitly (`components/viewer/hooks/use-viewer-tokens.ts:84`). The unified hidden-tab policy now rests on that TanStack default. One line at each of the two sites would make the policy structural rather than inherited, which matters because this default is exactly what made the #755 403 burst guaranteed rather than probabilistic.

ishiland:
## What changed

`ReportProblemHost` counted **rows** with `severity === 'error'`, and every unrecovered map error writes **two** of them:

1. a `maplibre` row from the surface handler, carrying `sourceId` (what makes a report actionable)
2. a `console` row derived from `logUnhandledMapError`'s `console.error` (the only trace the viewer and dataset preview leave at all)

Neither half can be dropped, so this takes the issue's primary ask: stop counting rows, count distinct failures. `ReportProblemHost.tsx:28` is the only call site that changed; the technical-details view still shows both rows.

## How a failure is identified

- **`maplibre` rows key on the source id AND the normalized message.** Neither is unique alone: two cluster layers over one dataset get their own MapLibre source and fail with byte-identical URLs, while two custom basemaps are both handed the fixed id `basemap` by `toMaplibreStyle()`.
- **`console` rows correlate by message** with a `maplibre` row when one exists, so the pair counts once. A console row with nothing behind it — viewer, dataset preview, which push no `maplibre` row — counts on its own.
- **Every other source keys on its own identity and detail**, so two `PanelErrorBoundary` instances throwing the same common message from different panels stay two failures.

## What the message normalizer does, and deliberately does not

It collapses only what is **unambiguously per-tile**: the tile address in MapLibre's supported forms (`{z}/{x}/{y}`, `{quadkey}` at any zoom, `{bbox-epsg-3857}`, and the `{ratio}` → `@2x` suffix) plus the rotating `sig`/`exp`/`scope`/`_v` params. `cols`, `cluster_radius` and `cluster_max_zoom` are kept — they distinguish real sources.

It does **not** touch host or path segments, and that is a decision rather than an omission. A resolved `{prefix}` shard is indistinguishable from a meaningful short label in a single URL: `a.tiles.example.com` looks exactly like `ca.tiles.example.com`, and `/a3/` like `/de/`. Earlier revisions of this PR tried to normalize both and each attempt merged two genuinely different basemaps into one badge entry. Between the two errors this keeps the **over**-count, because hiding a broken source is the worse failure for a tool whose whole job is surfacing problems. The reasoning lives in the code, not only here.

One quirk worth knowing for anyone touching this: `redact()` runs at capture, and it reads a retina tile path ending `/1539@2x.png` as an email address — so what reaches the key is `/12/1205/[redacted-email]`, with the coordinate already gone. The z/x/y rule accepts that placeholder as the `y` segment for exactly that reason.

**Not done: the stretch cross-source dedupe in `pushReportEntry`.** It needs a correlation key reliable enough to merge rows permanently; the count change gets the user-visible benefit without betting on one.

### Related nit from the same issue

`refetchIntervalInBackground: false` is now explicit at both sites in `hooks/use-tile-token.ts`, matching `use-viewer-tokens.ts:84`. Behavior is unchanged (it was the TanStack default) — but that default is what made the #755 403 burst guaranteed rather than probabilistic, so the policy should be stated, not inherited.

## Verification

    cd frontend && npx vitest run src/lib/report src/components/report
    cd frontend && npm run lint && npm run typecheck && npm run test:coverage   # 359 files, 4176 tests

Tests cover: the maplibre+console pair counting once (buffer level and through the rendered badge); every failing tile of one source counting once; two cluster sources over one dataset staying apart; two custom basemaps sharing the `basemap` id staying apart; two panels with the same message staying apart; quadkey (including low zoom), bbox and `@2x` normalization; and the host/path-label cases asserting the trade-off above explicitly.

Closes #908

### DIFF
--- a/frontend/src/components/report/ReportProblemHost.tsx
+++ b/frontend/src/components/report/ReportProblemHost.tsx
@@ -1,7 +1,7 @@
 import { useTranslation } from 'react-i18next';
 import { LifeBuoy } from 'lucide-react';
 import { useAuthStore } from '@/stores/auth-store';
-import { useReportDialog, useReportEntries } from '@/lib/report';
+import { countDistinctFailures, useReportDialog, useReportEntries } from '@/lib/report';
 import { cn } from '@/lib/utils';
 import { Tooltip, TooltipContent, TooltipTrigger } from '@/components/ui/tooltip';
 import { ReportProblemWizard } from './ReportProblemWizard';
@@ -25,7 +25,9 @@ export function ReportProblemHost() {
 
   if (!token) return null;
 
-  const errorCount = entries.reduce((count, entry) => count + (entry.severity === 'error' ? 1 : 0), 0);
+  // fix(#908): distinct failures, not rows — one unrecovered map error writes
+  // both a `maplibre` row and a `console` row, and the badge used to read "2".
+  const errorCount = countDistinctFailures(entries);
   const hasErrors = errorCount > 0;
   const badge = errorCount > 9 ? '9+' : String(errorCount);
 
--- a/frontend/src/components/report/__tests__/ReportProblemHost.test.tsx
+++ b/frontend/src/components/report/__tests__/ReportProblemHost.test.tsx
@@ -34,6 +34,29 @@ describe('ReportProblemHost', () => {
     expect(screen.getByText('1')).toBeInTheDocument();
   });
 
+  // fix(#908): the badge counts distinct failures. An unrecovered tile error
+  // writes both a `maplibre` row (which carries the sourceId) and a `console`
+  // row from logUnhandledMapError, and it used to read "2" for one failure.
+  it('counts one failure for the maplibre + console pair a 5xx tile writes', async () => {
+    signIn();
+    render(<ReportProblemHost />);
+    pushReportEntry({
+      severity: 'error',
+      source: 'maplibre',
+      message: 'AJAXError: Internal Server Error (500)',
+      detail: 'source: dataset-abc',
+    });
+    pushReportEntry({
+      severity: 'error',
+      source: 'console',
+      message: 'AJAXError: Internal Server Error (500)',
+    });
+
+    const button = await screen.findByRole('button', { name: /report a problem/i });
+    expect(button).toBeInTheDocument();
+    expect(screen.getByText('1')).toBeInTheDocument();
+  });
+
   it('opens the wizard from the floating button', async () => {
     signIn();
     render(<ReportProblemHost />);
--- a/frontend/src/hooks/use-tile-token.ts
+++ b/frontend/src/hooks/use-tile-token.ts
@@ -41,6 +41,12 @@ export function useTileToken(datasetId: string | undefined) {
       return Math.max(d.expires_in * 800, 30_000);
     },
     staleTime: 60_000,
+    // fix(#908): explicit because it is the whole hidden-tab policy — a hidden
+    // tab must not refresh, since MapLibre drops the resulting setTiles reload
+    // while the source's TileManager is paused (fix(#584)). It matched the
+    // TanStack default before, which is exactly what made the #755 403 burst
+    // guaranteed rather than probabilistic.
+    refetchIntervalInBackground: false,
   });
 }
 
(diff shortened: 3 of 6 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

Issue #936, opened 2026-07-30
sec(ci): Rule 2's pre-commit hook only greps httpx.AsyncClient, so in-process GDAL/rasterio fetches are unguarded


ishiland:
Surfaced by a codex P1 on PR #924 (#887). The finding there is being fixed in that PR; this issue is the enforcement gap it exposed.

> **Blocked on #937.** Option 1 below would have a grep assert that `gdal_safe_env` is referenced, but #937 may delete that helper's headline clamp (`GDAL_HTTP_FOLLOWLOCATION`, which turns out not to be a real GDAL option). Building a gate around a key that is on its way out is wasted work. Land #937's keep-vs-remove decision first, then gate whatever survives.

## The gap

AGENTS.md Rule 2 is "SSRF redirect-revalidation", and its stated enforcement is a pre-commit hook. That hook matches files constructing `httpx.AsyncClient(` while `follow_redirects=True` appears in the file (`.pre-commit-config.yaml:12-55`). Rule 2's text also covers the GDAL side — "the same applies to ogr2ogr / GDAL subprocess calls that fetch user-supplied URLs: set `GDAL_HTTP_FOLLOWLOCATION=NO` in the subprocess env" — but **nothing enforces that half**.

So a new **in-process** GDAL access is invisible to the gate:

```python
with rasterio.open(path) as ds:   # path may be an HTTP(S) URL for a remote STAC asset
    ...
```

No `httpx`, no subprocess, no `follow_redirects` — the hook cannot see it, and neither can a reviewer skimming for the documented pattern.

**The gap is wider than in-process access.** A new GDAL *subprocess* that fetches a user-supplied URL without going through `gdal_safe_env()` also trips nothing: the hook's `files:` pattern only reaches `backend/app/**.py`, and inside those files it greps exclusively for `httpx.AsyncClient(`. `backend/app/modules/catalog/sources/preview.py:116` is the shape — it runs `ogrinfo` against a user-supplied service URL and hand-rolls its env inline instead of calling `gdal_safe_env()`, so it sits outside both the hook and the helper. Rule 2's GDAL sentence is unenforced in both directions, not only for the in-process case.

## Why it matters concretely

[...]

PR #968, opened 2026-07-30
fix(sec): remove the no-op GDAL_HTTP_FOLLOWLOCATION redirect clamp

### LINKED ISSUE #937: sec: GDAL_HTTP_FOLLOWLOCATION=NO does not work — the documented SSRF redirect clamp is a no-op
Surfaced while fixing a codex P1 on PR #924 (#887), then verified independently against the **production worker image**.

## The finding

`GDAL_HTTP_FOLLOWLOCATION` is not a GDAL configuration option. Setting it to `NO` does not stop GDAL following an HTTP redirect.

**Reproduced on GDAL 3.10.3 (the worker image itself), in-container, with a local redirect server:**

    default (no config)      -> requests=['/start.tif', '/TARGET.tif', '/']  FOLLOWED_REDIRECT=True
    FOLLOWLOCATION=NO        -> requests=['/start.tif', '/TARGET.tif', '/']  FOLLOWED_REDIRECT=True

A `302` from `/start.tif` to `/TARGET.tif` is followed identically with and without the config. Independently observed on GDAL 3.12.1, where `gdalinfo --config GDAL_HTTP_FOLLOWLOCATION NO --debug ON` also answers `Warning 1: Unknown configuration option 'GDAL_HTTP_FOLLOWLOCATION'`. (On 3.10.3 that warning is not emitted for *any* unknown option — a bogus control option is equally silent — so absence of a warning there is not evidence either way. The behavioural test above is what settles it.)

**Confirmed against GDAL's own registry**, which settles step 1 below without needing a second experiment. `cpl_known_config_options.h` is the header that drives the "Unknown configuration option" warning, and it contains no `GDAL_HTTP_FOLLOWLOCATION`. The only redirect-named entry in it is `CPL_VSIL_CURL_USE_S3_REDIRECT`. The documented option list adds `CPL_VSIL_CURL_AUTHORIZATION_HEADER_ALLOWED_IF_REDIRECT` (GDAL >= 3.10, default `IF_SAME_HOST`), which governs whether the Authorization header survives a redirect but does not stop the redirect. So GDAL exposes no host-level redirect control at all, and the mitigation has to be structural.

Sources: https://gdal.org/en/stable/user/configoptions.html and https://gdal.org/en/stable/doxygen/cpl__known__config__options_8h_source.html

## Where the repo relies on it

Thirteen files, twenty-six lines. Line numbers below are against `28e205ef2`; the original filing predated PR #924, so `AGENTS.md` has moved from `:104` to `:112` and `cog.py` from `:229` to `:322`.

**The definition and its docstrings**

- **`backend/app/processing/raster/vrt.py:20`** — `_VRT_SAFE_ENV`, the central definition. Every raster subprocess gets the key from here through `gdal_safe_env()`, so this is the site to change; `cog.py` is only a consumer.
- **`backend/app/processing/raster/vrt.py:38-40`** — the strongest statement of the falsehood anywhere in the repo: "`GDAL_HTTP_FOLLOWLOCATION="NO"` — pinned with the SEC-S04 SSRF redirect-bypass defense; libcurl will not follow 3xx hops out of the explicitly-validated source URL."
- **`backend/app/processing/raster/vrt.py:47`** — the `ValueError`-on-collision contract, which treats the key as a security clamp callers may not override.
- **`backend/app/processing/raster/vrt.py:81-88`** — already correct. PR #924 added `gdal_safe_open_env()` with an accurate `fix(#887)` note saying the option does not exist and "provides no redirect protection to anybody". **The same file now asserts both things fifty lines apart.** Anyone opening `vrt.py` will find the correction already written and may conclude the file is done.

**The subprocess call sites**

- **`backend/app/processing/ingest/ogr.py`** — `:862` sets it; the comments at `:842` and `:849` state it is "the only way to disable" redirect following and that the path is "bounded by `GDAL_HTTP_FOLLOWLOCATION=NO` (no redirect rebinding)". `:705-709` explains why `run_ogr2ogr` deliberately does *not* set it.
- **`backend/app/modules/catalog/sources/preview.py:116`** — **a fourth subprocess site that the original filing missed.** `run_service_preview` runs `ogrinfo` against a **user-supplied service URL** with `env = {**os.environ, "GDAL_HTTP_FOLLOWLOCATION": "NO"}` as its only claimed redirect defense. This is the most directly user-controlled instance in the repo.
- **`backend/app/processing/raster/cog.py:322`** — consumer comment naming the clamp.

**Runtime configuration**

- **`docker-compose.yml:507`** and **`docker-compose.prod.yml:418`** — set on the TiTiler service, commented "TiTiler must not follow redirects".

**Tests that pin the no-op in place**

Seven assertions across five files. This is the remediation blast radius, and it is why this is not a comment-only change:

- **`backend/tests/test_titiler_redirect_env.py`** — the whole 22-line file exists only to assert `env["GDAL_HTTP_FOLLOWLOCATION"] == "NO"` in both compose files (`:17`, `:22`).
- **`backend/tests/test_ogr_subprocess_env.py:123`**
- **`backend/tests/test_preview_token_sec021.py:111`**
- **`backend/tests/test_vrt_hardening.py:261`**
- **`backend/tests/test_cog_subprocess_env.py:266`** — asserts `gdal_safe_env(extras={"GDAL_HTTP_FOLLOWLOCATION": "YES"})` raises, i.e. it enshrines the key as a clamp that cannot be overridden.

**Documentation**

- **`AGENTS.md:112`** — Rule 2's GDAL half: "set `GDAL_HTTP_FOLLOWLOCATION=NO` in the subprocess env", citing `backend/app/processing/ingest/ogr.py` as the reference implementation.
- **`CHANGELOG.md:947`** claims it as a shipped mitigation.

So the control is asserted in code comments, the contributor guide, the runtime compose files, the release notes, and the test suite — and it does nothing.

## What this does and does not mean

It does **not** automatically mean an exploitable SSRF. The API-layer defences are separate and unaffected: `validate_url_for_ssrf`, `make_safe_client()`'s per-hop `_revalidate_redirect` hook, and the `CPL_VSIL_CURL_ALLOWED_EXTENSIONS` allow-list (which is a real GDAL option and does constrain what a redirect target can be). The accurate statement is that **one layer people believe exists does not**, and any reasoning that concluded "safe because GDAL will not follow the redirect" needs redoing.

The exposure to assess: a URL that passes `validate_url_for_ssrf` at import time but later redirects to a private address, reached by a GDAL/ogr2ogr fetch that assumed the clamp.

**The concrete reachable instance is TiTiler serving a remote STAC asset.** `tiles/router.py:850` calls `resolve_open_path(meta.asset_uri)`, and `titiler_url.py:159-161` passes `http(s)` URLs straight through as a documented STAC pass-through. TiTiler then opens that caller-supplied URL in-process. Its only real constraint is `CPL_VSIL_CURL_ALLOWED_EXTENSIONS: ".tif,.tiff,.cog,.vrt"`; the redirect defense named two lines below it in the same compose block does nothing. `preview.py:116` is the second instance, on a service URL supplied directly by the user.

**SEC-008 needs its threat model redone, not its comment edited.** `ogr.py:849` accepts a known DNS-rebinding TOCTOU on the libcurl path and argues it "is bounded by `GDAL_HTTP_FOLLOWLOCATION=NO` (no redirect rebinding) and must be mitigated operationally". Half of that bound does not exist, so the residual risk SEC-008 signed off on is larger than recorded. Whoever picks this up should re-derive the SEC-008 conclusion rather than reword the sentence.

## Suggested next steps

1. **Establish what actually constrains redirects in GDAL** — **answered above: nothing does.** `CPL_VSIL_CURL_ALLOWED_EXTENSIONS` limits the target's extension but not its host, and no host-level redirect control exists in the option registry. GDAL fetches of caller-supplied URLs cannot be made redirect-safe from the inside, so the mitigation is structural.
2. **Prefer not fetching caller-controlled hosts with GDAL at all.** PR #924 takes this route for its new probe — it refuses to open `http://`, `https://`, `/vsicurl/`, `/vsicurl_streaming/` sources and treats them conservatively, on the grounds that `/vsis3/`, `/vsiaz/` and local paths are managed storage (bucket from settings, key already validated by `resolve_storage_key`). That distinction is reusable, and `preview.py` plus the TiTiler pass-through are where it would apply next.
3. **Fix the comments, AGENTS.md, and the compose files** so nobody re-derives the false guarantee. A no-op env var that reads as a security control is worse than no env var. If it is kept for defence-in-depth against a future GDAL that implements it, say exactly that.
4. **A behavioural test** — the redirect-server probe above is about 25 lines and would catch a future GDAL either implementing the option or changing the default.

## The keep-vs-remove decision has to be made first

The two options have different test consequences, so decide before touching anything:

- **Remove the key.** Seven assertions across the five test files above fail, and `test_titiler_redirect_env.py` loses its entire reason to exist. Cleanest end state, largest diff.
- **Keep it, documented as inert.** Every test stays green, but each of the five files needs its rationale rewritten to say the assertion pins a placeholder rather than a defense, or the suite goes on teaching the falsehood to the next reader.

Record the choice in the PR. Either way `CHANGELOG.md:947` should get a corrections entry in the current release rather than an edit to the shipped one, since the original note went out as a security claim.

## AGENTS.md is the process-sensitive part

`AGENTS.md` is auto-loaded into every session, so a stale Rule 2 keeps teaching the false guarantee to everyone who touches this area. The full sync set, all of which must land together:

- `AGENTS.md` Rule 2 (the GDAL sentence at `:112`, and the "Enforcement" paragraph that cites it)
- the `.pre-commit-config.yaml` hook description that points at Rule 2
- `vrt.py:20`, `:38-40`, `:47`, and the already-correct `:81-88` that currently contradicts them
- `ogr.py:705-709`, `:842`, `:849`, `:862`, including the SEC-008 re-derivation
- `preview.py:116`
- `docker-compose.yml:507` and `docker-compose.prod.yml:418`
- the five test files listed above
- `CHANGELOG.md:947`, as a corrections entry

## Acceptance

- No file in the repo asserts that this variable provides redirect protection. That includes code comments, docstrings, compose comments, AGENTS.md, test docstrings, and the changelog.
- The keep-vs-remove decision is recorded in the PR body with its test consequence.
- `vrt.py` no longer says two contradictory things about the same variable.
- SEC-008's residual-risk statement reflects that the redirect bound does not exist.
- `grep -rn GDAL_HTTP_FOLLOWLOCATION` returns only sites consistent with the recorded decision.

Related: #936 (Rule 2's pre-commit hook only greps `httpx.AsyncClient`, so in-process GDAL fetches are unguarded either way), #887/PR #924.

**#936 is blocked on this issue.** Its cheapest option is a grep asserting that `gdal_safe_env` is present, which would gate on a key this issue may delete. Land this one first.

ishiland:
## What

`GDAL_HTTP_FOLLOWLOCATION` is not a GDAL configuration option and never blocked a redirect (absent from `cpl_known_config_options.h`; behaviourally verified on GDAL 3.10.3 and 3.12.1 in #937). This PR removes it everywhere and makes every comment, docstring, test, and guide describe the defenses that actually exist.

**Decision: remove, not keep-as-inert** — made by the maintainer decision comment on #937 ("remove the key everywhere, do not keep it as documentation"), cited here per that comment; not re-litigated. Test consequence of that choice: the seven assertions pinning the no-op are rewritten to assert the key stays **absent**, and `test_titiler_redirect_env.py` is repurposed to pin the real `CPL_VSIL_CURL_ALLOWED_EXTENSIONS` allow-list plus the key's absence in both compose files.

## Changes

- `backend/app/processing/raster/vrt.py` — key removed from `_VRT_SAFE_ENV`; `gdal_safe_env` / `gdal_safe_open_env` docstrings no longer contradict each other; both now state no GDAL redirect control exists and point at the structural defense (managed-storage-only fetches).
- `backend/app/processing/ingest/ogr.py` — env key removed from the service-ingest subprocess; SEC-S04 comment rewritten; **SEC-008 re-derived**: both the DNS-rebinding TOCTOU and post-validation redirects are open on the libcurl path and must be bounded operationally (worker egress rules), not "bounded by" a variable that does nothing.
- `backend/app/modules/catalog/sources/preview.py` — key removed from the `ogrinfo` env; comment names the real defense (`validate_url_for_ssrf` at submission).
- `backend/app/processing/raster/cog.py` — consumer comment no longer lists the key as a clamp.
- `docker-compose.yml` / `docker-compose.prod.yml` — key removed from the titiler service; do-not-re-add comments left in place.
- `AGENTS.md` Rule 2 — GDAL half rewritten: GDAL cannot be made redirect-safe from the inside; prefer never handing caller-controlled URLs to GDAL, validate at the API layer, use `gdal_safe_env()` for the real clamps.
- Five test files rewritten coherently (absence assertions; the `gdal_safe_env` collision test now seeds a real clamp key).
- `CHANGELOG.md` — corrections entry under Unreleased for the 1.4.3 security claim; the shipped 1.4.3 note is left as history.

## Security impact

No behaviour change to any fetch path — the variable never did anything. The change is that the repo stops claiming a defense that does not exist, and SEC-008's recorded residual risk now matches reality. Related follow-up: #936 (structural gate for GDAL/rasterio call sites) lands on top of this.

## Verification

```
  uv run pytest tests/test_titiler_redirect_env.py tests/test_ogr_subprocess_env.py \
    tests/test_preview_token_sec021.py tests/test_vrt_hardening.py \
    tests/test_cog_subprocess_env.py -q        # 43 passed
grep -rn GDAL_HTTP_FOLLOWLOCATION . --exclude-dir=node_modules --exclude-dir=.git
# every remaining hit states the variable is inert / must not be re-added
uv run ruff check . && uv run ruff format --check .   # clean on touched files
```

Closes #937


ishiland:
@codex review

PR #969, opened 2026-07-30
fix(export): count a crossing bbox in the 5M cap guard instead of skipping it

### LINKED ISSUE #905: fix(export): the 5M feature-count guard skips a crossing bbox, so a narrow Pacific AOI 413s
Found while fixing #885 (PR #898), left out to keep that PR scoped.

`_count_selected_features` (`backend/app/processing/export/router.py:101`) guards its bbox clause on `bbox[0] <= bbox[2]`. So for a `west > east` bbox it counts the dataset **unfiltered**.

Consequence: a dataset over the 5M-feature export ceiling returns 413 for a crossing AOI that would have narrowed the count well under the cap. The user is told their export is too large when the AOI they asked for is small.

This is the same asymmetry class as #885, one layer up — and the same tell: `parquet.py` is exempt from this guard and counts the split correctly, so the two paths disagree about how many features a Fiji bbox selects.

## Why it was deliberately left

The clause is `&&`-only (bounding-box operator, no `ST_Intersects` recheck) specifically to keep a 5M-row scan cheap. Fixing it means either a third split variant of the predicate or putting `ST_Intersects` on that scan. #898 introduced a shared `bbox_where_sql(bbox, *, literal=False)` builder in `processing/export/ogr.py` that already emits the correct split predicate, so the pieces exist — the open question is the cost of using it here, not how to write it.

Worth measuring before choosing: a `&&`-only OR of two envelopes should still be index-assisted on both halves, in which case this is nearly free and the concern is unfounded.

## Related, and probably the better fix

There are now three antimeridian-split implementations in the backend:

- `core/geo.py::make_bbox_filter` — SQLAlchemy Core, the read path
- `modules/catalog/features/service.py:173-177` — raw text
- `processing/export/ogr.py::bbox_where_sql` — the shared builder #898 added

`features/service.py` cannot reuse the export one: `catalog/` must not import `app.processing.*` at module scope and `processing/` must not import `app.modules.catalog.*` (CATPORT-02/04, PROCESS-02/04, enforced by `backend/tests/test_layering.py`). The single correct home is `core/geo.py`, beside `make_bbox_filter` — which was off-limits during #898 because #892/#884 owned that file. **PR #901 has since landed**, so `core/geo.py` is free now and already hosts the seam-aware `extent_to_bbox` / `extent_to_span_bbox` / `bbox_to_extent_wkt` trio. Consolidating there and having all four call sites (including this count guard) use one predicate builder would close the class rather than patch the instance.

## What number decides it

"Worth measuring before choosing" needs a decision rule, so here is one.

The guard's own contract is the bar: its docstring (`processing/export/router.py:73-79`) states it "does strictly less work than the export it gates", because the inner `LIMIT` stops the scan at cap+1 rows. Any replacement predicate has to keep that true.

Run `EXPLAIN (ANALYZE, BUFFERS)` on the count query against a dataset at or above `_MAX_EXPORT_FEATURES`, with a crossing bbox, for three variants: today's unfiltered count, the `&&`-only OR of two envelopes, and `bbox_where_sql`'s full `&&` + `ST_Intersects` OR.

**Adopt `bbox_where_sql` if** both halves of the OR show a GiST index scan on `geom_4326` (no sequential scan), and the crossing-bbox count runs no slower than today's unfiltered count. Both are likely: today's path scans to cap+1 rows *unfiltered*, so an index-assisted split that returns far fewer rows should win outright rather than merely break even. If that is what the numbers say, the cost concern that deferred this was unfounded and the consolidation is free.

**Fall back to an `&&`-only third variant if** the `ST_Intersects` recheck measurably costs on the 5M-row case. That keeps the guard conservative in the correct direction — `&&` is a superset of exact intersection, so it can only over-count, which errs toward 413 rather than toward letting an oversized export through.

**Do not** leave the current behavior in place on the grounds that it is "conservative". It is conservative about the wrong thing: it protects the cap by ignoring the filter entirely, which is what produces the false 413.

## Acceptance criteria

- A crossing bbox that selects under the cap returns a normal export, not a 413.
- The count for a crossing bbox agrees with what `export/parquet.py` selects for the same bbox — the two paths stop disagreeing.
- A crossing bbox that genuinely selects over the cap still 413s.
- Whichever variant is chosen, the measured plan is recorded in the PR so the next person does not re-derive it.

ishiland:
## What changed

`_count_selected_features` (`backend/app/processing/export/router.py`) guarded its bbox clause on `bbox[0] <= bbox[2]`, so a west>east (antimeridian-crossing) bbox counted the dataset **unfiltered**. An oversized dataset then returned 413 for a narrow Pacific AOI that selected far fewer rows than the cap, and the count path disagreed with `export/parquet.py`, which already splits the bbox.

The guard now uses `bbox_where_sql` from `processing/export/ogr.py` — the shared split predicate #898 introduced, already used by parquet and the ogr2ogr `-where` path — so the count uses the exact predicate the export executes and the two paths cannot drift.

## Measured plan (the issue's decision rule)

`EXPLAIN (ANALYZE, BUFFERS)` on a synthetic 5M-row point table with a gist index on `geom_4326` (uniform global points), Fiji crossing bbox `[177,-20,-178,-15]`, `LIMIT 5000001`:

| variant | plan | execution |
| --- | --- | --- |
| V1 today: unfiltered (crossing bbox skipped) | Seq Scan to cap+1 rows | **509 ms** (41 728 buffers) |
| V2 `&&`-only OR | BitmapOr of 2 Bitmap Index Scans on `geom_4326` gist | 7.9 ms |
| V3 `bbox_where_sql` (`&&` + `ST_Intersects` OR) | BitmapOr of 2 Bitmap Index Scans on `geom_4326` gist | **3.5 ms** (2 046 buffers) |

Both halves of the OR are index scans, no seq scan; the adopt condition is met outright — the `ST_Intersects` recheck costs nothing measurable at this selectivity, so the `&&`-only fallback variant was not needed. The docstring's "strictly less work than the export it gates" contract holds with ~145x margin.

## Acceptance criteria

- Crossing bbox selecting under the cap exports normally (was a false 413) — tested (`test_crossing_bbox_under_cap_passes`, real Pacific table, cap lowered to 2).
- Crossing bbox genuinely over the cap still 413s — tested.
- Count now agrees with `parquet.py` structurally: both call the same `bbox_where_sql`.
- Measured plan recorded above.

Not done here: the issue's "probably the better fix" consolidation of all three split implementations into `core/geo.py` touches `modules/catalog/features/service.py`, outside this PR's scope; the class-level consolidation remains open in the issue thread.

No schema or API changes.

## Verification

```
uv run pytest tests/test_export.py tests/test_export_antimeridian.py -v
```

55 tests pass locally.

Closes #905


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

Issue #927, opened 2026-07-30
Dataset visibility can't be changed after import — add a control to the Access tab


ishiland:
Part of the #940 visibility-ladder epic. **Sequencing: this ships private/public only. Do not add the `internal` option until #930's backend branches have merged** — until then the value writes a state that makes the dataset invisible in every list, including its owner's.

## Problem

A user asked in [discussion #89](https://github.com/geolens-io/geolens/discussions/89#discussioncomment-17840175) how to change a dataset from private to public. There is currently no way to do this in the UI. Visibility can only be chosen at import time (`frontend/src/components/import/RegisterForm.tsx`, `frontend/src/components/import/ImportMetadataForm.tsx`). After that, the dataset page shows it read-only.

Maps have the Share button for this; datasets have nothing equivalent.

## Current behavior

- The Access tab renders visibility as a read-only badge with help text: `frontend/src/components/dataset/tabs/AccessTab.tsx` (the "Visibility" card near the bottom of the file).
- The Metadata tab has no visibility field.
- The backend already supports changing it: `PATCH /api/datasets/{dataset_id}` accepts `visibility` in `DatasetMetaUpdate` (`backend/app/modules/catalog/datasets/api/router.py`, `update_dataset_metadata`). Access is owner-or-admin via `check_dataset_write_access`. Verified working against a dev stack: flipping `private` → `public` immediately makes the dataset readable anonymously, and flipping back returns it to 404 for anonymous requests.

So this is frontend-only work. No backend or schema changes.

One backend behaviour to know about before wiring the error path: moving a dataset **away from public** while a public map uses it is rejected with a 422 by `_apply_visibility_change` (`backend/app/modules/catalog/datasets/domain/service_metadata.py:156-172`). This control is the first UI that can trigger it. Surfacing that message is #931's job, not this issue's, but the mutation here must not swallow the error.

## Proposed change

In the Visibility card in `AccessTab.tsx`, render a select instead of the static badge when the current user can edit the dataset. Keep the badge for everyone else.

Implementation notes:

[...]

PR #970, opened 2026-07-30
fix(catalog): exempt the creator from the restricted-visibility grant check

### LINKED ISSUE #929: Restricted dataset visibility is a dead end: grants have no write path and owners lock themselves out
Part of the #940 visibility-ladder epic. **This is the independent prerequisite and can land first and alone** — the creator exemption below does not depend on #930 or #927.

Follow-up from #927, which will add a visibility control to the dataset page. Investigating what the "restricted" option actually does surfaced two problems. The first is decided and deferred (see the non-goal section at the bottom); the second is a live bug and is this issue's core.

## The bug: the owner of a restricted dataset loses access to it

`filter_visible` allows: public to everyone, private to its creator, restricted to grant holders. There is no creator clause on the restricted branch, and `can_access_dataset` has the same shape (`backend/app/platform/extensions/defaults_extensions.py`). So a non-admin owner who sets their own dataset to restricted immediately loses read access to it, unless their role happens to hold a grant that can only exist via manual SQL.

Precisely, in `defaults_extensions.py`:

- `filter_visible` builds its `conditions` list with public (`:78`) and private+owner (`:79-82`), then appends restricted+grant (`:90-100`). Nothing matches a restricted record for its own creator. The `created_by == user.id` at `:104` is part of `status_filter`, which is **AND**ed with `or_(*conditions)` at `:106`, so it cannot rescue the owner.
- `can_access_dataset` (`:136-145`) returns solely on the grant lookup for restricted records.

Both the list path and the detail path deny the owner.

This is reachable today: the import forms (`frontend/src/components/import/RegisterForm.tsx`, `frontend/src/components/import/ImportMetadataForm.tsx`) offer restricted at upload time. #927 would have made it reachable post-import as well.

## The fix

**Add a creator exemption to the restricted branches.** In `defaults_extensions.py`, extend the restricted condition in `filter_visible` to also match `record_cls.created_by == user.id`, and in `can_access_dataset` return `True` for the creator before the grant lookup. This makes restricted mean "owner, admins, and grant holders", which is almost certainly what anyone selecting it expects, and it removes the lockout regardless of what happens with grant management.

It stays worth doing even though restricted is leaving the pickers, because SQL-managed deployments can still hold restricted datasets and their owners are locked out today.

**Add pinning tests for the authenticated owner-of-restricted case in both the detail and list paths.** The existing restricted tests cover anonymous (`backend/tests/test_datasets.py:387`) and the authenticated non-grantee with an admin-bypass leg (`:400`). Neither covers the owner.

**Remove `restricted` from the two import pickers** so the lockout is no longer reachable at upload time. Both files need the type and the option removed:

- `RegisterForm.tsx:17` (the `DatasetVisibilityChoice` union) and `:301` (the `SelectItem`)
- `ImportMetadataForm.tsx:35` (the `VISIBILITY_OPTIONS` entry)

This half may land in #927's PR instead if that is more convenient; what matters is that it lands with or before the new select. #927's select never offers restricted.

**Legacy values.** A dataset already `restricted` must still DISPLAY as restricted wherever visibility renders — the badge stays — and #927's select shows the current value as a non-reselectable disabled option with private/public as the available moves. A one-way exit, no silent coercion.

**Release notes** should mention both behaviour changes: the creator exemption (a restricted dataset's owner regains read access) and the picker removal.

## Non-goal: grant management

**Decision, 2026-07-30.** Restricted comes out of the pickers and grant management is deferred until there is demonstrated demand. It may end up an enterprise-tier feature, which is a separate roadmap call. Recorded here with the evidence so the reasoning survives.

Restricted visibility means "users whose role holds a `DatasetGrant` for this dataset" (`defaults_extensions.py`, `filter_visible` and `can_access_dataset`). Nothing in the product can create a grant:

- No API endpoint references `DatasetGrant` for writes. Every reference is a read for visibility filtering: `processing/tiles/router.py:789-795`, `processing/ai/service.py:317` and `:546`, `platform/sandbox/validator.py:424`, `platform/jobs/router.py:965-976`, `maps/service_layers.py:59-62`, `maps/service_shared.py:306`, plus the permission extension itself.
- No service function, no seed, no migration inserts one. The only `dataset_grants` writes in the repo are the table creation in `alembic/versions/0001_baseline.py:1384` and a test fixture in `backend/tests/test_permission_overlay.py`.
- The frontend has no grant UI and the generated SDK has no grant endpoints, because there is nothing to generate from.

The only way to make restricted visibility do anything is a hand-written SQL `INSERT INTO catalog.dataset_grants`.

If this is ever revisited, the shape is known: grants key a role (`dataset_id` + `role_id`, `backend/app/modules/catalog/datasets/domain/models.py:694`), not a user, so the UI would be a role multi-select rather than a user search, and the endpoints would be `GET/PUT/DELETE /api/datasets/{id}/grants` gated by `check_dataset_write_access` per Rule 1 in AGENTS.md. That work is out of scope here.

The need restricted was pretending to serve — team-visible datasets without grants machinery — is met by `internal` instead. See #930.

ishiland:
## What changed

`backend/app/platform/extensions/defaults_extensions.py`:

- `filter_visible`: the restricted condition now matches `record_cls.created_by == user.id` OR the existing grant subquery. The owner clause could not be rescued by the trailing `status_filter` because that is ANDed with the visibility conditions.
- `can_access_dataset`: returns `True` for the creator before the grant lookup.

Restricted now means "owner, admins, and grant holders" on both the detail and list paths. Security impact: this widens access only to the dataset's own creator — no change for anonymous users, non-owner non-grantees, or any other visibility level.

`backend/tests/test_datasets.py`: new `test_owner_of_restricted_dataset_keeps_access` pins the authenticated owner-of-restricted case on GET `/datasets/{id}`, GET `/datasets/`, and GET `/search/datasets/`, each with a non-grantee viewer control. Existing restricted tests (anonymous, non-grantee + admin bypass) are unchanged and still pass.

## Scope notes

- Per the issue, grant management is a non-goal.
- The import-picker removal halves (`RegisterForm.tsx` / `ImportMetadataForm.tsx`) are frontend and belong with #927's visibility-select work per the issue body ("may land in #927's PR instead"); this PR is the independent backend prerequisite and lands alone.

## Verification

- `tests/test_datasets.py`: 32 passed.
- Reverting the extension change alone makes the new test fail (owner gets 404 on detail), so the test pins the bug.
- `tests/test_permission_overlay.py`: 5 skipped (overlay importorskip; community checkout), unchanged.

Closes #929

### DIFF
--- a/backend/app/modules/catalog/maps/service_layers.py
+++ b/backend/app/modules/catalog/maps/service_layers.py
@@ -6,7 +6,7 @@ from sqlalchemy import delete, select
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from app.core.identity import Identity
-from app.modules.auth.models import UserRole
+from app.modules.catalog.authorization import apply_visibility_filter
 from app.modules.catalog.datasets.domain.models import Dataset, DatasetGrant, Record
 from app.modules.catalog.maps.models import MapLayer
 from app.modules.catalog.maps.schemas import MapLayerInput, split_legacy_builder_paint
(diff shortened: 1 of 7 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
Cross-domain note: this PR now touches processing/tiles/router.py (agent B's area) with coordinator approval — same authorization predicate as the other two gates, restricted-branch only.

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #972, opened 2026-07-30
fix(analysis): hold the per-user materialize slot on a heartbeat lease

### LINKED ISSUE #691: Analysis materialize: hold the per-user job slot on a heartbeat lease, not job status alone

ishiland:
## What changed

`backend/app/modules/catalog/datasets/api/router_analysis.py` (the only app-code file, per the issue's step 3):

- New module constant `MATERIALIZE_LEASE_SECONDS = 90` — 3x the 30s `HEARTBEAT_INTERVAL_SECONDS`, tolerating two missed renewals; recorded on the issue. Kept a constant on purpose; Settings promotion is #696.
- Cap predicate changed from `status.in_(('pending','running'))` to `status = 'pending' OR (status = 'running' AND coalesce(heartbeat_at, started_at) >= cutoff)` — the shape from `_leasing_ingest_job_ids` (worker.py:305), at the short lease rather than the 60-minute timeout.
- The pending branch is kept status-only, with a comment spelling out why: a pending job has never been claimed, so both timestamps are NULL and any cutoff comparison silently drops it, letting a user stack unbounded queued CTASes.
- The 19-line no-lease rationale comment is rewritten to describe the lease.

`frontend/src/components/analysis/AnalysisJobWatcher.tsx`: comment-only update recording why the client deliberately still applies no staleness rule (it has no heartbeat visibility; a released lease means the next create simply succeeds server-side).

## Tests (`backend/tests/test_analysis_materialize.py`)

- `test_old_pending_job_still_blocks` — unchanged, still passes (the regression test for the silent failure mode).
- `test_long_running_job_with_fresh_heartbeat_still_blocks` — the old `test_long_running_job_still_blocks` updated for the lease: 45-minute job, heartbeat now → 429.
- `test_stale_heartbeat_releases_the_slot` — running, heartbeat 5 minutes stale → next materialize admitted (200, the endpoint's success status).
- `test_null_heartbeat_with_fresh_started_at_still_blocks` — the coalesce leg → 429.

## Verification

- `tests/test_analysis_materialize.py`: 71 passed.

Closes #691


ishiland:
@codex review

ishiland:
@codex review

PR #973, opened 2026-07-30
fix(builder): register fill-pattern images in the mixed-geometry adapter

### LINKED ISSUE #919: fix(builder): mixed-geometry layers never register the fill-pattern images, so a patterned polygon renders empty
Found in the 2026-07-30 style-tab audit (pass 2). Re-verified on main @ 28e205ef2.

Scope note: this issue now covers the missing image registration only. The second finding from the original report, that a data-driven ramp on a mixed layer colors only the polygon family, moved to #952 because it needs a product decision first.

## What happens

On a mixed-geometry layer (`GEOMETRY` or `GEOMETRYCOLLECTION`), a `fill-pattern` in paint makes the polygon family render empty instead of patterned. MapLibre has no image registered under that id, so it draws nothing.

`frontend/src/components/builder/layer-adapters/mixed-adapter.ts` imports `FILL_OWNED_PAINT_PROPERTIES` (`:16`), which includes `fill-pattern` (`fill-adapter.ts:20-28`), and syncs the whole set onto its fill sublayer in `syncFillLayer` (`:211-216`). It is the only fill-rendering adapter that never calls `ensureFillPatternImages`. The two existing call sites are both in the standalone fill adapter (`fill-adapter.ts:76` and `:173`).

The key really does reach MapLibre: `mixedFillPaint` runs `filterPaintForLayerType(input.paint, 'fill')` (`mixed-adapter.ts:94-97`), and that filter keeps `fill-pattern`, since it starts with `fill-`, is not a `fill-extrusion-` property, and is not a custom underscore prop (`frontend/src/components/builder/layer-adapters/shared.ts:171-192`).

The pattern picker is hidden for these layers, because `isPolygon` is a `.includes('POLYGON')` test that both generic sentinels fail (`LayerStyleEditor.tsx:203`). So the key only arrives via Advanced JSON, an AI style action, or a pasted style. It is reachable, just not from the picker.

## Fix

Call `ensureFillPatternImages(map)` at the top of the mixed adapter's `addLayers` and `syncPaint`, the same way `fill-adapter.ts` does at `:76` and `:173`. Two lines plus a test.

## Acceptance

- Advanced JSON `"fill-pattern": "geolens-fill-hatch"` on a mixed layer: polygons render the pattern instead of vanishing.
- The line and point sublayers are unaffected.
- `cd frontend && npm run lint && npm run typecheck && npx vitest run src/components/builder/__tests__/layer-adapters.test.ts`

`layer-adapters.test.ts:1424` and `:1439` are the equivalent registration pins for the standalone fill adapter and are the tests to mirror.

ishiland:
## What changed

On a mixed-geometry layer (`GEOMETRY` / `GEOMETRYCOLLECTION`), a `fill-pattern` in paint made the polygon family render empty. `mixed-adapter.ts` syncs the whole `FILL_OWNED_PAINT_PROPERTIES` set onto its fill sublayer, `fill-pattern` included, and `filterPaintForLayerType(paint, 'fill')` keeps the key — but it was the only fill-rendering adapter that never called `ensureFillPatternImages`, so MapLibre had no image registered under the id and drew nothing.

One call at the top of `addLayers`. `syncPaint` self-heals through `addLayers`, so that single call covers both entry points; the test pins both.

The pattern picker stays hidden for these layers (`isPolygon` is a `.includes('POLYGON')` test both generic sentinels fail), so the key only arrives via Advanced JSON, an AI style action, or a pasted style. Reachable, just not from the picker.

## Verification

    cd frontend && npm run lint && npm run typecheck && npx vitest run src/components/builder/layer-adapters/__tests__/mixed-adapter.test.ts src/components/builder/__tests__/layer-adapters.test.ts

124 tests pass. Three new pins mirror the standalone fill adapter's registration tests: addLayers registers, syncPaint registers, and the `hasImage` idempotency gate still skips.

Closes #919

### DIFF
--- a/frontend/src/components/builder/layer-adapters/__tests__/mixed-adapter.test.ts
+++ b/frontend/src/components/builder/layer-adapters/__tests__/mixed-adapter.test.ts
@@ -6,6 +6,7 @@ import {
   mixedLinesLayerId,
   mixedPointsLayerId,
 } from '../mixed-adapter';
+import { FILL_PATTERN_IDS } from '../fill-pattern-images';
 import type { AdapterLayerInput } from '../types';
 
 /**
@@ -28,6 +29,8 @@ function createMockMap(opts: { layerExists?: boolean } = {}) {
     setPaintProperty: vi.fn(),
     getLayoutProperty: vi.fn().mockReturnValue(undefined),
     getPaintProperty: vi.fn().mockReturnValue(undefined),
+    hasImage: vi.fn().mockReturnValue(false),
+    addImage: vi.fn(),
   };
 }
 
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #975, opened 2026-07-30
fix(ci): stamp per-image OCI titles on published images

### LINKED ISSUE #932: fix(ci): all four published images report org.opencontainers.image.title=geolens, discarding the per-image titles
Found while working #893 item 3 (PR #926), left out to keep that PR to comment-and-gate changes.

`Dockerfile` sets a distinct title per stage — `geolens-api` (`:163`), `geolens-worker` (`:178`), `geolens-backup` (`:211`), `geolens-frontend` (`:287`). None of them survive publication.

`publish.yml` passes `labels: ${{ steps.meta.outputs.labels }}` (`:155`) from `docker/metadata-action`, whose defaults include `org.opencontainers.image.title` derived from the **repository name**. A `--label` on the build command overrides a `LABEL` in the Dockerfile, so every published image reports `title=geolens` and the four Dockerfile titles are silently discarded.

`description` escapes this only because `publish.yml` explicitly sets it from `matrix.description` (`:92-93`). `title` has no such override, so it falls through to the default.

**Verified against the live registry:** all four images carry `title=geolens`.

**Fix** is one line in the metadata step's `labels:` input: `org.opencontainers.image.title=${{ matrix.image }}`. The matrix key is confirmed — `publish.yml:38/41/44/47` hold `geolens-api`, `geolens-worker`, `geolens-frontend`, `geolens-backup`, so that expression is correct as written. Small, but it changes published image metadata, which is why it wants its own PR rather than riding a comment-fix batch.

**`title` is not the only label being overridden.** `metadata-action`'s defaults also emit `org.opencontainers.image.licenses` and `org.opencontainers.image.source`, so the Dockerfile's `LABEL` lines for those (`:165-166`, `:180-181`, `:213-214`, `:289-290`) are equally dead on the publish path. They happen to land on correct values — the repo license is Apache-2.0 and the source URL matches — so there is no bug to fix there. Worth knowing anyway: a fix that restores only `title` leaves three quarters of each Dockerfile `LABEL` block as decoration on published images, and the next person to read those lines will assume they do something. Decide whether to keep them for the non-publish build paths (where they *are* the only source of the metadata) or to drop them, and say which in the PR.

## Related, worth doing at the same time

`dep-audit.yml:51` has a comment above the CLI/SDK license gates claiming "Both ship their own uv.lock; `--project` resolves each locked env." Neither ships one, and the reason is deliberate rather than an oversight: `cli/.gitignore:2` and `sdks/python/.gitignore:12` both list `uv.lock`, so those lockfiles are gitignored by design and `git log --all` confirms neither has ever been committed.

That makes the comment wrong in a way that matters. The backend gate (`:34`) and the MCP gate (`:42`) both run `uv lock --check` against a committed lockfile, so they audit a pinned dependency tree. The CLI and Python SDK gates (`:54`, `:58`) resolve fresh at CI time, so they audit whatever resolves that day. Both still block strong copyleft, which is what the gate is for, but they are not locked and the comment says they are. The corrected comment should say the two projects deliberately do not commit a lockfile and that their license gates therefore run against an unlocked resolution. It now sits directly below the `mcp/uv.lock` gate added in #926, where it reads as contradicting that step.

## Note: #893 item 3 was a false premise — do not act on it

That item said published images carry no `org.opencontainers.image.version`. They do: `metadata-action`'s defaults include `version`, `revision`, `created`, and `url`, and the registry confirms all four images tagged `1.6.1` carry `version=1.6.1`. Adding a Dockerfile `LABEL org.opencontainers.image.version` would be *worse* than absent — overridden by `--label` on the publish path, and empty on every other build path — and it would create a new version site to register in `bump_version.py` and `check_version_coherence.py` for no benefit.

#893 item 3 is being struck from #893 directly as part of this remediation round, so this issue no longer needs to carry the instruction. The paragraph above stays as the record of why.

ishiland:
## What

One-line fix from #932 (verified correct in the issue body): add `org.opencontainers.image.title=${{ matrix.image }}` to the metadata-action `labels:` input in `publish.yml`. Without it, metadata-action's default derives `title` from the repository name, `--label` overrides Dockerfile `LABEL`, and all four published images report `title=geolens` — the per-stage Dockerfile titles never survive publication. `description` already escapes the default the same way; `title` now does too.

**Dockerfile LABEL decision (as the issue asks the PR to state):** the per-stage `LABEL` blocks are KEPT. On non-publish build paths (local `docker build`, compose builds) they are the only source of title/licenses/source metadata; a comment above the `labels:` block in `publish.yml` records that they are overridden on the publish path so nobody assumes the labels there are decorative or vice versa.

Also the related dep-audit correction from the issue: the comment above the CLI/SDK license gates claimed both "ship their own uv.lock". Neither does — `cli/.gitignore` and `sdks/python/.gitignore` both ignore `uv.lock` by design, and neither lockfile has ever been committed — so those two gates audit an unlocked, resolve-at-CI-time tree. The comment now says exactly that, which stops it reading as a contradiction of the `mcp/uv.lock` gate directly above it.

Per the issue, #893 item 3 (adding a version LABEL) is a false premise and is deliberately NOT acted on: registry images already carry `version` from metadata-action defaults.

## Impact

Published-image metadata only; no runtime behavior change. Takes effect on the next tag publish.

## Verification

- `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/publish.yml')); yaml.safe_load(open('.github/workflows/dep-audit.yml'))"` — both parse.
- Dry-run reasoning: the metadata step's `labels:` list is passed verbatim to `build-push-action` at `publish.yml:155`; explicit entries win over metadata-action defaults (documented precedence, and the mechanism `description` already relies on in this same block). `matrix.image` holds exactly `geolens-api`/`geolens-worker`/`geolens-frontend`/`geolens-backup` (`publish.yml:38-47`), so each image's title matches its Dockerfile stage title. The next `v*.*.*` tag publish will show `title=geolens-api` etc. on the registry, replacing `title=geolens`.

Closes #932

### DIFF
--- a/.github/workflows/dep-audit.yml
+++ b/.github/workflows/dep-audit.yml
@@ -48,7 +48,11 @@ jobs:
         run: uv run --with pip-licenses pip-licenses --fail-on="GPL-3.0-only;GPL-3.0;GPL-2.0;AGPL-3.0;SSPL-1.0;BSL-1.1"
 
       # Extend the BLOCKING license gate to the CLI + Python SDK (same allowlist).
-      # Both ship their own uv.lock; `--project` resolves each locked env.
+      # fix(#932): unlike the backend and MCP gates above, these two run against
+      # an UNLOCKED resolution — cli/ and sdks/python/ deliberately gitignore
+      # uv.lock (cli/.gitignore, sdks/python/.gitignore) and have never
+      # committed one, so each run audits whatever resolves that day. That is
+      # acceptable for a copyleft block, but it is not a pinned-tree audit.
       - name: License gate — CLI (block strong/network copyleft)
         working-directory: cli
         run: uv run --with pip-licenses pip-licenses --fail-on="GPL-3.0-only;GPL-3.0;GPL-2.0;AGPL-3.0;SSPL-1.0;BSL-1.1"
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -89,7 +89,14 @@ jobs:
             # unreleased candidate (same spirit as #325 gating release.yml's
             # isPrerelease). Fresh `:latest` pulls must track the last GA.
             type=raw,value=latest,enable=${{ github.event_name == 'push' && github.ref_type == 'tag' && !contains(github.ref_name, '-') }}
+          # fix(#932): --label overrides Dockerfile LABEL, and metadata-action's
+          # defaults derive title from the REPO name — without the explicit
+          # title below, all four published images reported title=geolens and
+          # the per-stage Dockerfile titles were discarded. The Dockerfile
+          # LABEL blocks stay: they are the only metadata source on non-publish
+          # build paths (local docker build, compose).
           labels: |
+            org.opencontainers.image.title=${{ matrix.image }}
             org.opencontainers.image.description=${{ matrix.description }}
             org.opencontainers.image.licenses=Apache-2.0
             org.opencontainers.image.source=${{ github.server_url }}/${{ github.repository }}


ishiland:
@codex review

Issue #931, opened 2026-07-30
fix(catalog): surface the public-map visibility block, and close the internal-map gap


ishiland:
Part of the #940 visibility-ladder epic. #927 has shipped (#976), so the Access-tab control this hangs off now exists.

> **Rewritten 2026-07-30.** The original body claimed `PATCH /api/datasets/{id}` applies a visibility change with no map-impact check, and proposed a frontend confirm dialog as the guard. That premise was wrong, a hard block already exists, and the acceptance criteria derived from it were not implementable. The corrected scope is below; the real gap turns out to be narrower and in a different place.

## What already exists

Moving a dataset away from `public` while a **public** map uses it is already rejected. `_apply_visibility_change` (`backend/app/modules/catalog/datasets/domain/service_metadata.py:156-172`), reached from `:293`:

```python
if new_visibility != "public" and record.visibility == "public":
    public_maps = await find_public_maps_using_dataset(session, dataset_id)
    if public_maps:
        raise ValueError(
            f"Cannot restrict visibility: dataset is used in public maps: {', '.join(public_maps)}"
        )
```

That is a 422 naming the offending maps, and it covers every move away from public: private, restricted, and internal alike. Two tests in `backend/tests/test_datasets.py` pin it, `test_restrict_dataset_blocked_when_used_in_public_map` at `:241` (422, map name in `detail`) and `test_restrict_dataset_allowed_when_no_public_maps` at `:282` (200).

So there is no silent breakage for public maps, and no confirm-then-apply flow is possible for them. Confirming cannot apply a change the server refuses.

## The actual gaps

The two gaps below are not equally ready. Gap 1 is implementable today and blocked on nothing. Gap 2 needs #930 to land first and needs a decision recorded here. They are written up together because gap 2's backend option changes a string gap 1 has to match, which is covered under gap 2.

### 1. The block is invisible in the UI

`find_public_maps_using_dataset` returns map names precisely so a human can act on them, and those names currently do not reach the user.

#976 did the first half. The Access-tab select routes its 422 through `formatMutationError` into a toast (`frontend/src/components/dataset/tabs/AccessTab.tsx:249-255`), with an inline comment deferring the localized message to this issue. So the error is no longer swallowed, and the select no longer snaps back silently. What the user sees is the raw backend prose or a generic fallback, depending on how `error-map.ts` resolves it.

[...]

Issue #930, opened 2026-07-30
Dataset visibility 'internal' is accepted but unimplemented: hidden from lists, readable by any signed-in user by ID


ishiland:
Part of the #940 visibility-ladder epic. Steps 1 and 2 of that epic have shipped (#929 in #970, #927 in #976), so this is the next issue in the sequence.

Found while investigating #927. Dataset visibility `"internal"` is accepted by two write paths but has no implementation in the permission layer, so datasets in that state behave inconsistently.

## How the state is reached

- `PATCH /api/datasets/{id}` accepts it: the `Visibility` Literal is `"private" | "restricted" | "internal" | "public"` (`backend/app/modules/catalog/datasets/domain/schemas.py:116`).
- The CLI manifest path writes it: publication intent `internal` maps to `("internal", "internal")` for (visibility, record_status) in `backend/app/processing/ingest/manifest_sources.py:46`, so `geolens apply` with that intent creates the state directly.

**These two paths do not produce the same bug**, and the difference matters when reproducing. See below.

## What actually happens

The permission layer only knows public, private, and restricted (`DatasetVisibility` enum in `backend/app/modules/catalog/authorization.py:26-31`; branch logic in `backend/app/platform/extensions/defaults_extensions.py`):

- `filter_visible` (list/search/tiles paths, `:63-112`): the OR-conditions built at `:77-107` cover public, private+owner, and restricted+grant. `visibility == "internal"` matches none of them, so the dataset is invisible to every non-admin, **including its owner**.
- `can_access_dataset` (single-dataset paths, `:114-158`): the private and restricted branches don't match, so the function falls through to `return True` at `:158`. On a published record, any signed-in user can read the dataset by direct ID.

So a `PATCH`-created internal dataset (visibility `internal`, record_status `published`) is hidden from lists but readable by direct link for all authenticated users. Neither half matches the other, and neither is documented.

**The manifest-created case behaves differently.** `geolens apply` writes record_status `internal`, not `published`. The status gate at `can_access_dataset:136` checks `record.record_status != "published" and record.created_by != user.id` and returns False before ever reaching the fall-through, so those datasets are **not** readable by any signed-in user. They are invisible in lists even to their creator, and readable by ID only by the owner and admins. Anyone verifying the "any signed-in user can read it" claim against `geolens apply` data will fail to reproduce it and may wrongly conclude this issue is stale. Reproduce the fall-through with a `PATCH` to a published dataset.

[...]

PR #976, opened 2026-07-30
feat(dataset): change dataset visibility from the Access tab

### LINKED ISSUE #927: Dataset visibility can't be changed after import — add a control to the Access tab

ishiland:
Frontend-only. The backend already accepts `visibility` in `PATCH /api/datasets/{id}` (`DatasetMetaUpdate`, owner-or-admin via `check_dataset_write_access`) — no schema or API change here.

## What changed

**`AccessTab.tsx` — the Visibility card.** Renders a `Select` when the viewer can edit the dataset, and the existing read-only badge for everyone else. `canEdit` is threaded `DatasetPage → DetailPanel → AccessTab` rather than re-derived: `DetailPanel` already holds the exact `isEditor() && useCanMutate(dataset)` value it passes to `SourcesTab`, so a non-owner editor correctly gets the badge.

Options are **private and public only**:

- **No `restricted`.** Per #929 it leaves the pickers entirely — a non-admin owner who selected it lost access to their own dataset, and grants have no write path.
- **No `internal` yet.** #930 adds it with the import pickers.
- **A stored `restricted` dataset still displays as Restricted**, rendered as a *disabled* current option with private and public as the available moves. A one-way exit, never a silent coercion. The guard is `!SELECTABLE_VISIBILITIES.includes(...)` rather than `=== 'restricted'`, so a SQL-set `internal` gets the same treatment until #930 lands.

**Both import pickers lose `restricted` — type and option.** `RegisterForm.tsx` (`DatasetVisibilityChoice` union + the `SelectItem`) and `ImportMetadataForm.tsx` (`VISIBILITY_OPTIONS`). Deleting only the option would have left a union member that still typechecks against a value the picker can no longer produce.

**The 422 is not swallowed.** Moving a dataset away from public while a public map uses it is rejected by `_apply_visibility_change`; this control is the first UI that can trigger it. `onError` raises a toast through `formatMutationError` so the select does not just snap back in silence. Turning that server prose into its own localized message is #931's job, not this PR's — an unmapped detail still collapses to the generic 422 string today.

**i18n.** `metadataEdit.visibilityHelp` is rewritten in en/es/fr/de: the old text described `restricted`, which this PR removes from every picker. Two new keys (`visibilityUpdated`, `visibilityFailed`) added to all four locales.

## Reused, not rebuilt

`visibilityColors`, `VisibilityIcon`, and `getVisibilityLabel` already handle every visibility value including `restricted` and `internal`, so legacy datasets keep their badge, icon and colour for free. The mutation is the existing `useUpdateDataset`, which already invalidates the dataset query — the select and the "Access:" header badge both refresh without a reload.

## Verification

    cd frontend && npx vitest run src/components/dataset/__tests__/AccessTab.test.tsx   # 8 passed
    cd frontend && npx vitest run src/components/import src/components/dataset src/pages # 406 passed

New tests: owner/admin sees the select and a change calls the update with the new value; viewer/non-owner sees the badge; `restricted` is not offered as a move; a stored `restricted` renders as the disabled current value with private and public selectable; a rejected change raises an error toast; an unchanged value fires no mutation.

No screenshot: the Playwright MCP browser profile is held by another parallel session right now, and Chrome MCP renders no map tiles. The control is a standard `ui/select` in the existing card layout, pinned by the tests above.

Closes #927


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #978, opened 2026-07-30
ci(e2e): name the failing specs in the nightly failure report

### LINKED ISSUE #896: ci(e2e): a stale spec can sit red for days — nightly-only specs have no PR-time gate and no staleness signal
Surfaced while diagnosing #894. Not a bug in any product code — a gap in how test rot is detected.

> **Still open after #895.** `4c4dcf6eb` unstuck the two specs but touched only `e2e/analysis.spec.ts` and `e2e/dataset-chat.spec.ts`. It changed no workflow and no script, so every structural condition below is unchanged at `28e205ef2`.

## What happened

`e2e/dataset-chat.spec.ts:193` asserted the ephemeral badge read `Query result · 2 features`. PR #674 (`13fd3ec9a`, 2026-07-24) deliberately renamed that copy to `"Result"` — the badge stopped being query-only once analysis previews reused it. The spec was never updated, so it failed **every** nightly from 07-24 onward and nobody noticed for six days.

Two things combined to hide it:

1. **`E2E Tests` runs on `schedule` / `workflow_dispatch` only.** `.github/workflows/ci.yml:1468` gates the job on `github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'`. PRs get `E2E Smoke (core)` instead (`:1555`), which is `npm run e2e:smoke:core` = admin / auth / search / dataset-detail / collections / permissions / console-hygiene. **Neither `analysis.spec.ts` nor `dataset-chat.spec.ts` is in any `e2e:smoke*` script.** So #674 — the PR that renamed the string — could not have caught its own fallout, and neither could #879–#883.
2. **Nightlies were `cancelled` for most of the preceding week** (07-25 through 07-28 all show `cancelled`), so the first two runs that actually completed were 07-29 and 07-30, six days after the break landed.

The PR-time trade is deliberate (#825 traded the full suite off PRs for the Actions budget) and this issue is not asking to undo it. The problem is that a nightly-only spec has no signal between "someone changes a string" and "someone reads a red nightly six days later".

## How much of the suite this covers

Two specs is how it surfaced, but the exposed surface is most of the suite. Of the **30 specs in `e2e/`**:

- **7 are gated at PR time**, via `e2e:smoke:core`: `admin`, `auth`, `search`, `dataset-detail`, `collections`, `permissions`, `console-hygiene`.
- **2 more are gated on push to main only**, via the `accessibility` job (`ci.yml:1743` runs on `push || schedule || workflow_dispatch`): `accessibility`, `keyboard-nav`. Those catch rot after a merge, not before.
- **21 have no CI gate at all before the nightly**: `analysis`, `attribute-table-ax`, `builder`, `builder-dem-editor`, `builder-hardening`, `builder-styling`, `builder-unified-stack`, `builder-v1-5`, `dataset-chat`, `download-cog-token`, `export-runtime`, `feature-editing`, `non-spatial`, `plugin-lifecycle`, `post-impl-validation`, `record-detail-ux-audit`, `reupload-multi-layer-gpkg`, `sec-audit`, `showcase-smoke`, `showcase-smoke-anonymous`, `upload`.

Two entries in that last group are worth naming on their own. `sec-audit.spec.ts` is a security spec with no gate before the nightly. And six of the 21 are the builder group (`builder`, `builder-dem-editor`, `builder-hardening`, `builder-styling`, `builder-unified-stack`, `builder-v1-5`), which is the most actively edited surface in the repo right now — the whole #910–#923 batch lands there.

The smoke scripts do not close this, because most of them are never run by CI. `e2e:smoke:core` is the only `e2e:smoke:*` script any job invokes. `e2e:smoke:builder`, `e2e:smoke:fixtures`, `e2e:smoke:reupload`, and `e2e:smoke:audit` exist purely as local recipes, which is what `ci.yml:1459` means when it says reviewers run the remaining groups before approving. (`e2e:smoke:builder-hardening` is the exception in reverse: its specs do run nightly, through a direct config invocation at `:1519` rather than the npm script.) So the scripts read like a tiered gate and are not one.

## Why the existing nightly-report automation did not help

The `Nightly E2E Failure Report` job did its job — #894 was filed. But the report is coarse: it said `e2e-test: failure` / `accessibility: success` and guessed that `skipped` meant an upstream job failed, which was wrong (E2E genuinely failed). It names jobs, not specs. So the issue gave no hint that one failure was a deterministic six-day-old string mismatch and the other was a coin-flip race — that took downloading traces to establish.

## Worth considering, roughly in order of cost

1. **Name the failing specs in the nightly report.** The Playwright JSON reporter already knows them. A report that said `dataset-chat.spec.ts:193 failed 3/3` vs `analysis.spec.ts:58 failed 3/3` would have made the deterministic-vs-flaky split obvious without touching an artifact. Cheapest change with the most diagnostic value.
2. **Distinguish "failed every night" from "failed once".** A spec red on N consecutive nightlies is a different animal from a flake, and the report already has the history to say so. This is the signal that was actually missing — six consecutive identical failures should read louder than one.
3. **Report a nightly that never completes.** Note the original framing here was wrong: the cause of the four cancelled runs is already fixed. `ci.yml:16-20` puts `github.event_name` in the concurrency group precisely so a dependabot push stops cancelling the scheduled run, and the comment there records that this is what rotted the suite during the v1.6.0 audit. What is still missing is the *signal*: `nightly-e2e-report` only fires on `github.event_name == 'schedule'` with a completed `e2e-test` result (`ci.yml:1911-1914`), so a scheduled run that is cancelled or never starts produces no output at all. Four silent days is the failure mode to close, and the fix is detect-and-report a non-completing scheduled run, not another concurrency change.
4. **Consider whether copy renames need a cheap cross-check.** The specific rot here was an i18n string an e2e spec asserted on. A grep-style gate is probably the wrong tool (too brittle, too many false hits), but it is worth noting that `frontend/src/i18n/locales/en/*.json` values are asserted verbatim in e2e specs with no link between them. `ViewerChatPanel.test.tsx:64` carries a comment documenting the #674 rename — the unit layer knew, the e2e layer did not.

## Adjacent nit found in the same diagnosis

`e2e/analysis.spec.ts` resolved its output dataset id *after* the assertion it was failing on, so every failed attempt leaked one dataset (the catalog count climbed 3 → 4 → 5 across three retries). Fixed in the #894 PR, but the pattern is worth checking for elsewhere: any spec that assigns a cleanup handle after its assertions leaks on failure.

ishiland:
## What

Implements the cheapest option from #896 (option 1: spec-level detail in the nightly report), and states the design trade below.

- `playwright.config.ts`: add the `json` reporter (`playwright-summary.json`, gitignored) to the CI reporter list alongside `github` + `html`.
- `ci.yml` `e2e-test`: a `Summarize failing specs` step (`if: !cancelled()`, so it runs on the failure path that matters) jq-extracts `file:line`, title, and attempt count for every `ok: false` spec into a job **output**. Job outputs are readable from `needs` even when the job failed, so no artifact plumbing or new (unpinned) download action is introduced.
- `ci.yml` `nightly-e2e-report`: the tracking-issue body now lists the failing specs. `(attempts: 3)` on a 2-retry suite reads as deterministic rot; `(attempts: 1..2)`-then-pass never appears (only `ok: false` specs are listed), so what shows up is real reds with their retry burn — the dataset-chat six-day break would have read `dataset-chat.spec.ts:193 ... (attempts: 3)` on night one.

## The trade (as #896 asks the PR to state)

This buys diagnosis, not prevention. Deliberately NOT done here:

- **No PR-time gate for the 21 nightly-only specs** — #825's Actions-budget trade (full suite off PRs) stands; this PR does not relitigate it.
- **No consecutive-failure history** (issue option 2) — needs cross-run state; the per-night spec list in a single threaded issue gives a human the same signal by reading two consecutive comments.
- **No missing-nightly watchdog** (issue option 3) — a cancelled/never-started scheduled run still produces silence; that needs a separate scheduled job and stays open on #896's thread if wanted.
- **builder-hardening suite is not itemized** — it runs `--reporter=line` under its own config; the report body says so explicitly, so a red `e2e-test` with an empty spec list points there.

## Verification

- `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/ci.yml'))"` — parses.
- jq expression exercised against a sample of the Playwright JSON shape (nested suites, mixed ok true/false): emits exactly `- \`analysis.spec.ts:58\` runs analysis (attempts: 3)` and nothing for passing specs.
- Dry-run reasoning: the summarize step only reads a file the json reporter writes at repo root; when Playwright never ran (e.g. `npm ci` failed) the file is absent and the output is empty, and the report body skips the section on an empty `FAILED_SPECS`. The heredoc output delimiter (`PW_SPECS_EOF`) cannot collide with jq's output format.

Closes #896


ishiland:
@codex review

Issue #944, opened 2026-07-30
fix(catalog): a zero-height bbox builds an invalid zero-area extent ring


ishiland:
Noted in passing in #904 and split out so it gets its own decision. Pre-existing, not a regression from #901.

## The defect

A bbox whose `south == north` produces a ring of four collinear points, zero area, and not a valid polygon:

    POLYGON((w s, e s, e s, w s, w s))

`bbox_to_extent_wkt` (`backend/app/core/geo.py:154-196`) already guards the *width* case deliberately: the crossing branch at `:187-191` drops halves with `x0 < x1` false, and falls back to the full `-180..180` band at `:193` when neither half has real width, on the stated principle that "an extent must never silently narrow to nothing". The *height* case has no equivalent guard on either branch, and `_ring` (`:150-151`) emits whatever it is handed.

The non-crossing branch at `:168-169` is unguarded in both dimensions: it returns `POLYGON({_ring(west, south, east, north)})` directly, so a zero-width non-crossing bbox has the same defect as a zero-height one.

So the two paths are now consistently wrong together, which is why #901 left it: the crossing path was made to match existing non-crossing behavior rather than to fix it.

Reachable whenever a source's features are collinear in latitude, a single point or a run of points on one parallel. The producer paths that pad this today do so with `ST_Expand`, and only for the degenerate geometry *type* (`ST_Extent` casting to POINT/LINESTRING), not for an areal-typed bbox that happens to have zero height. See `get_extent` (`backend/app/processing/ingest/metadata.py:365`, the `ST_Expand` at `:389`) and `_refresh_count_and_extent` (`backend/app/modules/catalog/features/service.py:613`, the `ST_Expand` at `:638`) for the existing padding pattern.

## Decision: pad, following the precedent already in this file

**Decision, 2026-07-31.** Pad. This is not an open question any more; it was settled in practice by #934, which hit the identical problem one caller upstream and solved it inside `seam_extent_wkt_for_table` at `backend/app/core/geo.py:417-423`:

```python
# Mirror the callers' degenerate padding (their non-crossing path runs
# ST_Expand 1e-9 on POINT/LINESTRING extents): a crossing extent can
# never be zero-width, but a single-parallel source is zero-height and
# would produce invalid zero-height rings. Related: #944.
if north - south < 1e-12:
    south -= 1e-9
    north += 1e-9
```

[...]

PR #980, opened 2026-07-30
fix(catalog): fold per-dataset extents seam-aware at every producer site

### LINKED ISSUE #934: fix(catalog): per-dataset extent producers still fold naively, so a Pacific source stores one global POLYGON
Found while shipping #886 (PRs #925/#928). That work fixed the **rollup** (aggregate-across-records) side; these are the **producer** (single-record) sites, which still take a naive min/max and were deliberately out of scope.

## The sites

1. **`backend/app/processing/ingest/metadata.py`** — `get_extent`, and the consolidated `WITH meta AS (...)` extract statement. A Pacific-crossing source stores **one global POLYGON**.
2. **`backend/app/modules/catalog/features/service.py`** — `_refresh_count_and_extent` (`:612`, the `ST_Extent` fold at `:632-636`), plus the second `ST_Extent` producer around `:692-700`. The same producer-side fold.
3. **`backend/scripts/seed_tiles.py`** — reads raw `ST_XMin`/`ST_XMax` per record, so a seam-form record seeds world-wide tiles. Wasteful rather than wrong, but it will look like a bug when someone profiles it.

## Why this now matters more than it did

#888 (PR #899) made this **reachable through ordinary ingest**. `clip_to_mercator_bounds` no longer discards the out-of-range half of a 0..360 Pacific source — it shifts it — so a source spanning 150..250 correctly lands on both sides of the antimeridian. A naive `ST_Extent` over that reads **−170..170 — a 340° bbox for a 100° footprint** — published to the datasets API and the OGC collection extent. Those are the numbers the pinning test actually asserts (`backend/tests/test_ingest_mercator_clip.py:207-215`, seeding 150/170/190/250, which shift to 150/170/−170/−110); an earlier revision of this issue quoted −160..150 / 310°, which matches no endpoint of a 150..250 footprint — 250 wraps to −110, not −160. Before #899 the extent looked tidy only because half the data had been deleted.

#899 pinned that as `test_pacific_crossing_source_is_preserved_with_a_naive_extent`. **That test's comment on `main` says #886 owns the fix — which is now stale**, since #886 is closed and covered only the rollup side. Worth correcting the comment to point here as part of this work.

## The tools already exist

`backend/app/core/geo.py` now has everything needed (from #901 and #925/#928):

- `bbox_to_extent_wkt(west, south, east, north)` — splits a west>east bbox into two rings at ±180, never lets an extent narrow to nothing
- `extent_to_bbox()` / `extent_to_span_bbox()` — the spec vs monotonic readers
- `rollup_bbox()` / `rollup_span_bbox()` (`core/geo.py:321` and `:333`) / `merge_bboxes()` / `_narrower_domain` — the two-domain fold with its `_DOMAIN_MARGIN`. `rollup_span_bbox` is the monotonic sibling and matters here: picking between it and `rollup_bbox` is the same spec-vs-span decision each producer site has to make (see below).

Two traps that cost real time on #925, worth inheriting rather than rediscovering:

- **Shift whole footprints, not vertices.** A bare `ST_ShiftLongitude` moves each vertex, so a prime-meridian-spanning footprint (Europe `-10..30` → `(350, 30)`) produces an extent that excludes its own data. Verified with PostGIS: `covers-all-inputs=False`.
- **Comparing spans needs a margin.** `±360` round-tripping is not bit-exact, so a bare `<` lets the shifted domain win on last-bit noise for footprints that never cross the seam. `_DOMAIN_MARGIN = 1e-9` exists for this.

Also check the reupload/refresh paths, not just first ingest — an extent recomputed on refresh goes through the same producer code.

Related: #886 (rollup half, closed), #888 (made it reachable), #892/#884 (the column and helpers), #903 (the frontend guard that still blocks serving the spec form to map surfaces).

## Which form each site should emit

This is the decision most likely to be got wrong, so it is worth stating flatly: **two of the three sites should not be choosing a bbox form at all.**

Sites 1 and 2 write `catalog.records.spatial_extent`, which is a geometry column, not a bbox. Since migration `0030` it accepts the two-ring MULTIPOLYGON. So the fix there is to derive the honest extent geometry — the producer-side helper is `bbox_to_extent_wkt(west, south, east, north)` (`core/geo.py:146`), which splits a `west > east` bbox into two rings at ±180 and never lets an extent narrow to nothing.

The spec-vs-span choice then stays where it already lives: at the **read** boundary, in `extent_to_bbox` versus `extent_to_span_bbox`, whose call sites #901 already split deliberately. A producer that picks a bbox form has moved that decision to the wrong layer and flattens the extent for every consumer at once — which is the present bug, just relocated.

| site | emits | helper |
| --- | --- | --- |
| `ingest/metadata.py` — `get_extent` (`:339`) and the `WITH meta` extract (`:878-900`) | extent **geometry**, two-ring when crossing | `bbox_to_extent_wkt` |
| `catalog/features/service.py` — `_refresh_count_and_extent` (`:612`) | extent **geometry**, two-ring when crossing | `bbox_to_extent_wkt` |
| `scripts/seed_tiles.py` (`:115,117`) | a tile range, so a **bbox** — and here the choice is real | see below |

`seed_tiles.py` is the one genuine bbox consumer. Reading raw `ST_XMin`/`ST_XMax` off a two-ring extent gives `-180..180` and seeds the world. The correct answer is to read the extent through `extent_to_bbox` and seed each lobe as its own longitude range when `west > east`. `extent_to_span_bbox` is the acceptable fallback if per-lobe seeding is more surgery than this script deserves — it is honest and safe, it just keeps the waste this issue is complaining about, so choosing it should be a written decision rather than an accident.

Note the degenerate-extent padding these paths already do (`ST_Expand` for `ST_Extent` results that cast to POINT/LINESTRING) has to survive the rewrite. Related: #944 covers the zero-height ring case that both branches get wrong today.

## Acceptance criteria

- A Pacific-crossing source ingested end to end stores a two-ring `spatial_extent`, and its datasets-API `extent_bbox` reports the real footprint rather than a near-global one.
- `backend/tests/test_ingest_mercator_clip.py:178` (`test_pacific_crossing_source_is_preserved_with_a_naive_extent`) **starts failing** and is updated — the test says so itself, and it is the meeting point of #899 and this fix. Its comment still names #886 as the owner; correct that to this issue as part of the work.
- Non-crossing extents are byte-identical before and after. A Europe footprint spanning the prime meridian must not move.
- The reupload/refresh paths are covered, not just first ingest — an extent recomputed on refresh goes through the same producer code.
- Whichever form `seed_tiles.py` adopts is stated in the PR body.

ishiland:
## What changed

Since #888 shifts a 0..360 Pacific source onto both sides of the antimeridian instead of clipping it, ordinary ingest produces tables whose naive `ST_Extent` reads -170..170 — a 340-degree bbox for a 100-degree footprint — stored verbatim in `records.spatial_extent` and published through the datasets API and OGC collection extents. The rollup side was fixed by #886; these are the producer (single-record) sites.

Per the issue's form table:

| site | now emits |
| --- | --- |
| `ingest/metadata.py` — `get_extent` + `extract_metadata` CTE | extent geometry, two-ring MULTIPOLYGON when crossing, via `bbox_to_extent_wkt` |
| `catalog/features/service.py` — `_refresh_count_and_extent` (feature-write refresh path) | same |
| `scripts/seed_tiles.py` | **stated decision: per-lobe seeding.** Reads the extent in spec form via `extent_to_bbox` (`ST_AsBinary` + `WKBElement`) and seeds `west..180` and `-180..east` as separate ranges (`spec_bbox_to_tiles`). Fiji strip at z3 seeds the two edge columns, not the 0..7 world sweep. |

Shared machinery: new `seam_extent_wkt_for_table` in `core/geo.py`, built on the existing `rollup_bbox_columns` / `rollup_bbox` two-domain fold — so both #925 traps are inherited, not rediscovered (whole footprints shifted, never vertices; `_DOMAIN_MARGIN` guards the span comparison).

Design points:

- The spec-vs-span choice stays at the read boundary (`extent_to_bbox` / `extent_to_span_bbox`), where #901 already split it — producers store the honest geometry.
- Each producer gates the second aggregate scan on naive width > 180 (the only case the shifted domain can mathematically win), so ordinary datasets keep one scan and **byte-identical output**, including the degenerate POINT/LINESTRING `ST_Expand` padding.
- A crossing zero-height (single-parallel) extent is padded 1e-9 in latitude before `bbox_to_extent_wkt`, mirroring the naive path's `ST_Expand` and avoiding the invalid-ring class #944 tracks.
- `ingest/metadata.py` grew 52 lines; its exact LOC ratchet in `test_layering.py` is bumped with the required rationale comment.

## Acceptance criteria

- Pacific-crossing table stores a two-ring `spatial_extent` whose `extent_to_bbox` reports the real footprint — tested for all three producer entry points (`tests/test_extent_producers_934.py`).
- `test_pacific_crossing_source_is_preserved_with_a_naive_extent` updated (now `..._with_an_honest_extent`): it asserts `get_extent` over the #899-shifted table reports 150..-110, and its stale comment naming #886 as owner is corrected to #934.
- Non-crossing extents byte-identical (prime-meridian Europe, wide-but-not-crossing, degenerate point — all asserted against the pre-#934 SQL verbatim).
- Refresh path covered: `_refresh_count_and_extent` is the reupload/feature-write recompute and is tested directly; `get_extent` is the reupload-swap producer.
- `seed_tiles.py` form stated above; lobe seeding pinned in `tests/test_seed_tiles.py`.

**No migration** — `spatial_extent` accepts the two-ring form since migration 0030 (#892).

## Editing a test that #990 added

`TestBufferAtDateline::test_buffer_across_dateline_is_split_not_wrapped` (`tests/test_analysis_materialize.py`) arrived on main with #990, and its extent assertion fails here. I changed that assertion. Why it is the assertion and not this fix that was wrong:

- **It was written as a pin, and it named this PR as the thing that would break it.** Its comment read: *"fix(#892) widened the column so a two-ring MULTIPOLYGON … is now storable, but teaching the analysis extent rollup to emit one is the separate follow-up; pinned, not fixed, here."* This is that follow-up. Analysis registers through `register_existing_table`, so it shares the `ingest/metadata.py` producer changed above.
- **The behaviour that test actually guards is untouched.** #697 is about the buffer RESULT geometry being split rather than wrapped: `MULTIPOLYGON`, 2 parts, `ST_IsValid`, every vertex inside the WGS84 domain, and no false `ST_Intersects` against a bbox over central France. All five of those assertions pass unchanged — only the separate `catalog.records.spatial_extent` assertion moved.
- **The new shape is the improvement this issue exists to make.** The pinned extent was a single `POLYGON` spanning the full -180..180 band — 360° to describe a footprint 0.25° across — which is precisely the defect #934 removes. It is now two rings, 179.823..180 and -180..-179.923.

The replacement asserts structure rather than just the gtype, per review: one lobe per hemisphere meeting at the seam (order-independent — some ring ends on +180, some ring starts on -180), neither lobe wider than 1°, and a read-back through `extent_to_bbox` giving the RFC 7946 §5.2 crossing (`west > east`) pair with a ~0.25° span and an unchanged latitude band.

I swept every backend test referencing `spatial_extent` (23 files, 588 tests) to confirm this was the only cross-PR interaction; nothing else asserts extent shape in a way this fix changes.

Merge-train note: this should merge before #903's frontend guard work (session C), per the wave plan.

## Verification

```
uv run pytest tests/test_extent_producers_934.py tests/test_seed_tiles.py tests/test_ingest_mercator_clip.py tests/test_antimeridian_extent.py tests/test_antimeridian_rollup.py tests/test_features_crud.py tests/test_staging_pipeline.py tests/test_reupload_service.py tests/test_layering.py -v

# cross-PR interaction sweep: every backend test referencing spatial_extent
uv run pytest tests/test_processing_port.py tests/test_stac_import.py tests/test_ogc_record_enrichment.py \
  tests/test_validation.py tests/test_antimeridian_extent.py tests/test_ogc_collection_metadata.py \
  tests/test_search.py tests/test_geodcat_ap.py tests/test_analysis_materialize.py \
  tests/test_ai_send_sample_values.py tests/test_reupload_swap_lock_retry.py tests/test_permission_overlay.py \
  tests/test_stac_serializer.py tests/test_dcat.py tests/test_extent_producers_934.py \
  tests/test_search_cache.py tests/test_collections.py tests/test_metadata_roundtrip.py \
  tests/test_antimeridian_rollup.py tests/test_raster_antimeridian_887.py tests/test_ingest_mercator_clip.py \
  tests/test_seed_tiles.py tests/test_layering.py
```

254 tests pass locally across the first set; the `spatial_extent` sweep is 588 passed, 5 skipped.

Closes #934


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
Pushed d2c99541b. The failure was a genuine cross-PR interaction with #990, not a race — details and reasoning are now in the PR body under "Editing a test that #990 added".

Short version: `TestBufferAtDateline`'s extent assertion was written as a pin and its own comment named this PR as the follow-up that would change it ("teaching the analysis extent rollup to emit one is the separate follow-up; pinned, not fixed, here"). Analysis registers via `register_existing_table`, so it shares the ingest producer this PR changes. The #697 guarantees that test exists for — buffer result split not wrapped, valid, in-domain, no false France intersect — are asserted separately and all still pass untouched; only the `spatial_extent` assertion moved, from a single 360°-wide ring to the two-ring seam form, which is the defect #934 exists to remove.

New assertions check structure rather than gtype: one lobe per hemisphere meeting at the seam (order-independent), neither lobe wider than 1°, and an `extent_to_bbox` read-back giving the RFC 7946 §5.2 crossing pair at ~0.25° span.

Swept all 23 backend test files referencing `spatial_extent` (588 passed, 5 skipped) — this was the only interaction.

@codex review

ishiland:
Re-triggering at the final head 6b6cedf5b. Since the last request I resolved a conflict with main by hand (#963 and #906 both landed and both touch this diff):

- `backend/app/core/geo.py` — import-block conflict only; #963 added `functools.lru_cache`, this PR added `typing.TYPE_CHECKING`. Kept both.
- `backend/tests/test_layering.py` — #906 and #934 each bumped the same `ingest/metadata.py` LOC ratchet. Resolved to the actual merged count, 2002, keeping both rationale comments.

Re-verified the merged result locally: 201 tests across layering / wkt classification / extent producers / antimeridian extent / mercator clip / jobs router / seed tiles, plus 277 across analysis materialize / raster antimeridian / antimeridian rollup / raster tiles. All pass, and `ruff check` is clean.

@codex review

ishiland:
Merged current main (with #1003's revert) at 027890271. No conflicts — this is a pure main merge, no change to this PR's own diff since the codex-clean pass.

The inherited #1001 failure is gone: `tests/test_ai_sql_prompt_935.py` now passes alongside this branch's own suites (161 tests across ai_sql_prompt / extent_producers_934 / analysis_materialize / antimeridian_extent / layering).

@codex review

PR #981, opened 2026-07-30
fix(frontend): guard the map surfaces against a west>east bbox

### LINKED ISSUE #903: fix(frontend): no guard for a west>east bbox — a dozen map consumers would misrender a seam-crossing extent

ishiland:
**Merge order: after #934.** This is an enabler, not a live user-facing bug — every frontend-facing producer emits the span form today. Until the per-dataset extent producers fold at the seam, this guard has little real data to act on.

## The shared helper

`frontend/src/lib/bbox.ts` mirrors the backend split, so each call site declares which form it wants:

| helper | for |
| --- | --- |
| `crossesAntimeridian(b)` | the `west > east` test, in one place |
| `splitBbox(b)` | the two seam halves; a one-element array otherwise — consumers that can draw more than one rectangle |
| `toSpanBbox(b)` | the monotonic `-180..180` widening, mirroring `extent_to_span_bbox()` |
| `bboxLonSpan(b)` / `bboxCenterLon(b)` | span the short way round; center on the data, not its antipode |
| `toFitBounds(b)` | MapLibre corners, letting east run past 180 |

## The per-consumer split

**`normalizeRasterBounds` (`layer-adapters/shared.ts`) → span form.** The keystone. Every result feeds a MapLibre source `bounds`, which accepts one increasing box and nothing else; an inverted pair matches no tile, so a seam-crossing layer renders **blank**. Two boxes are not expressible on one source, so the choice is over-broad-but-visible or correct-and-invisible. Fixing the helper fixes all four inheriting sites (`map-sync.ts:108/800/1007`, `hillshade-adapter.ts:128`).

**`lib/map-extent.ts` → span + center helpers.** `isLargeExtent` computed a negative `maxx - minx`, so a Fiji extent read as "small". `computeLargeExtentView` was the worse of the two: the negative span was swallowed by `Math.max(lonSpan, 1)` into a world-wide zoom, and `(minx + maxx) / 2` put the center on the **antipode** — a confident fly-to on the far side of the planet. Both DatasetMap branches (`:503-511`, `:859-877`) inherit the fix.

**`DatasetMap.tsx` → split for the drawing, fit-bounds form for the fitting.** The extent ring becomes a MultiPolygon of the seam halves (a planar ring from a crossing pair drew the complementary 340° rectangle). `handleZoomToExtent` and `initialViewState.bounds` go through `toFitBounds`.

**`BuilderMap.tsx` → unwrap.** `getVisibleLayerBounds` dropped any crossing bbox outright, which is what made both fit paths and Zoom to Layer silent no-ops. It now unwraps east past 180 before merging, so a single crossing layer fits the few degrees it occupies. Merging several layers still min/maxes in unwrapped space — no worse than the planar merge it replaces, and a genuine union across the seam is the rollup problem #886/#934 solve on the backend.

**`BBoxPreview.tsx` → split.** Not in the issue's inventory but reached by three of the sites it lists (`CollectionCard:17`, `CollectionDetailPage:74`, `SearchResultCard`). It already detected a crossing extent for its viewBox, but still drew one rect from a negative width — `Math.max(width, 2)` collapsed it into a 2°-wide sliver pinned at the east edge. Now it draws both halves.

`DatasetPage.tsx:346` and `SourceQualityTab.tsx:171` are pass-throughs: the former feeds `DatasetMap`, the latter `SpatialExtentCard`, which prints the four numbers verbatim (correct as-is for a spec-form pair).

### Deliberately still rejecting

`use-builder-layers.ts:435-440` and `ChatPanel.tsx:515` keep refusing an inverted bbox. Both are no-ops rather than misrenders, which stays the correct conservative behavior, and both sit in another parallel session's domain — turning them into real handling is a clean follow-up now that the helper exists.

## Verification

    cd frontend && npx vitest run src/lib/__tests__/bbox.test.ts src/lib/__tests__/map-extent.test.ts   # 25 passed
    cd frontend && npx vitest run src/components/builder/layer-adapters/__tests__/shared.test.ts src/components/layout/__tests__/BBoxPreview.test.tsx
    cd frontend && npm run lint && npm run typecheck && npm run test:coverage   # 360 files, 4179 tests passed

Tests cover the Fiji bbox (`[178.5, -20, -178.5, -15]`) through every helper, plus the raster-bounds and extent-span sites the acceptance criteria name, and the two-rect preview. Each helper suite also asserts the non-crossing case is untouched, since that is the great majority of data.

No live-map QA: the dev stack serves `main`, and worktree Playwright results would be meaningless here.

Closes #903


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
Merge-order hold: this must land **after** #980 (issue #934, seam-aware extent producers). Per #903's body, the producers need to be writing honest two-ring extents before this guard ships, otherwise the span-to-spec flip order breaks.

Coordinator will merge this once #980 is on main. CI/codex work can proceed normally in the meantime.

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #982, opened 2026-07-30
ux(builder): validate per-value symbol icons, drop empty entries, translate anchors

### LINKED ISSUE #920: ux(builder): symbol editor batch — unvalidated per-value icons, empty-icon junk, silent 6-category cap, untranslated anchors
Found in the 2026-07-30 style-tab audit (pass 2). Four small confirmed items in the symbol-mode editor (`frontend/src/components/builder/LayerStyleEditor/SymbolEditor.tsx`), batched because they're all in one file. Pointers re-verified on main @ 28e205ef2.

1. **Per-value icons are unvalidated free text.** The primary icon uses `IconPicker` (`:29-34`) but each category row is a raw `<Input>` (`:114-119`). A typo becomes `geolens:markr` in the match expression (`symbol-adapter.ts:51-52`, `:92`) and that category silently draws nothing.

   Validate the typed value against the icon list and show an inline warning on the row when it doesn't resolve. That is the scoped fix here. Swapping in `IconPicker` per row is the tempting answer but it is a real layout question, not a drop-in: `IconPicker` (`frontend/src/components/builder/IconPicker.tsx:19-95`) renders a five-column swatch grid plus a file-upload control, so stacking one per category row rebuilds the whole section. If that's wanted, raise it as its own issue with a design call. (Query cost is not the concern — every instance shares one `useMapIcons()` cache entry.)

2. **Clearing a per-value icon leaves junk and a blank field.** `updateCategory` stores `{value, icon: ''}` (`:21-24`). The adapter correctly skips empty icons (`symbol-adapter.ts:80`) so the map falls back to the primary icon, but the display expression `find(...)?.icon ?? symbolConfig.iconImage ?? 'marker'` (`:110`) never reaches the fallback because `''` is not nullish, so the input shows blank while the map shows the fallback, and empty entries accumulate in saved `style_config`. Drop empty-icon entries from `categories` instead of storing them.

3. **The category list shows at most 6 values and says nothing about it.** `sampleValues` is `.slice(0, 6)` (`:17`) with no indication more exist, so a 20-category column shows 6 rows and nothing tells the user the rest are unmapped.

   The obvious fix — "showing 6 of N" — is not available, and it is worth knowing why before starting. `dataset_sample_values` is not the column's distinct values. The backend caps it at **10 distinct values per column** (`LIMIT 10` at `backend/app/processing/ingest/metadata.py:609`, documented at `:545-546`) and draws them from only the first 10,000 rows. So the largest `N` the frontend can ever compute is 10, and it is a sample, not a cardinality. Rendering "6 of 10" for a 20-category column just swaps one wrong impression for another.

   Scope for this issue: raise the slice to the full array the backend sends (up to 10 rows) and label the section so it reads as a sample rather than the complete set — something like "sampled values" with an i18n key in all four locales. Do not synthesise a total.

   A real "6 of 20" needs a distinct-count from the server, which means a new endpoint or a new column on the dataset metadata. That is out of scope here; open it separately if the truncation turns out to matter to users.

4. **Anchor dropdown shows raw tokens in every locale.** `:63-66` renders `['center','top','bottom-left',...]` as-is. LineEditor translates its equivalent cap/join options (`LineEditor.tsx:146`, `:162`); mirror that with `style.symbol.anchor.*` keys in all four locales.

## Acceptance

- A per-value icon that doesn't resolve is flagged inline instead of silently drawing nothing; a cleared row falls back visibly and saves no empty entry; the category list reads as a sample and shows every value the backend returned; anchor options are translated.
- `cd frontend && npm run lint && npm run typecheck && npm run test:i18n && npx vitest run src/components/builder/__tests__/LayerStyleEditor.test.tsx`

ishiland:
## What changed

All four items from the issue, all in `LayerStyleEditor/SymbolEditor.tsx`.

**1. Per-value icons are validated.** Each category row was raw free text, so a typo became `geolens:markr` in the match expression and that category silently drew nothing. The typed value is now checked against the same `useMapIcons()` list the primary `IconPicker` uses — one shared cache entry, no extra query — and an unresolved value gets an inline warning plus `aria-invalid` on the input. The check is skipped while the query is in flight so an unloaded list can't flag every row.

Swapping `IconPicker` in per row was deliberately not done: it renders a five-column swatch grid plus an upload control, so stacking one per row rebuilds the section. That's a design call for its own issue.

**2. Clearing a row drops the entry.** `updateCategory` stored `{value, icon: ''}`. The adapter skipped empty icons so the map fell back correctly, but the display expression never reached its `??` fallback (`''` is not nullish), so the field showed blank while the map showed the fallback — and empty entries accumulated in saved `style_config`.

**3. The list shows every sampled value.** The `.slice(0, 6)` is gone. `dataset_sample_values` is capped at **10 distinct values per column** server-side (`LIMIT 10`, drawn from the first 10k rows), so it is a sample and not a cardinality — the section is now labelled as such rather than claiming "6 of N". No total is synthesised; a real "6 of 20" needs a distinct-count from the server and is out of scope.

**4. Anchor options are translated.** `style.symbol.anchorOption.*` in all four locales, mirroring how `LineEditor` translates its cap/join options.

New keys (`anchorOption.*`, `sampledValues`, `unknownIcon`) land in en/es/fr/de.

## Verification

    cd frontend && npm run lint && npm run typecheck && npm run test:i18n && npx vitest run

359 files / 4161 tests pass; i18n parity green. New `SymbolEditor.test.tsx` covers all four items.

Closes #920

### DIFF
--- a/frontend/src/components/builder/LayerStyleEditor/SymbolEditor.tsx
+++ b/frontend/src/components/builder/LayerStyleEditor/SymbolEditor.tsx
@@ -3,7 +3,14 @@ import { Select, SelectContent, SelectItem, SelectTrigger, SelectValue } from '@
 import { SliderRow } from '../HeatmapStyleControls';
 import { IconPicker } from '../IconPicker';
 import type { BaseStyleEditorProps } from './types';
+import { useMapIcons } from '@/hooks/use-maps';
 import type { SymbolStyleConfig } from '@/types/api';
+
+// fix(#920): the anchor dropdown rendered these raw tokens in every locale.
+const ICON_ANCHORS = [
+  'center', 'top', 'bottom', 'left', 'right',
+  'top-left', 'top-right', 'bottom-left', 'bottom-right',
+] as const;
 
 export function SymbolEditor({
   layer,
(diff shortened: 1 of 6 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #983, opened 2026-07-30
test(builder): cover the BuilderMap popup pipeline and analysis registration

### LINKED ISSUE #943: test(builder): close the frontend analysis/builder Vitest gaps
Split out of #789, which mixed backend pytest, frontend Vitest, and Playwright work into one issue and so never got picked up whole. #789 keeps the backend items; the Playwright gap is #945. All line references below were re-verified against `main` at `28e205ef2` — the originals had drifted.

From the coverage track of the 2026-07-27 audit — the seams, not the middle.

1. **Panel job lifecycle + Add to map (P1).** Partly closed since the original report: #793 added `components/analysis/__tests__/AnalysisJobWatcher.test.tsx`, which now exercises `useJobStatus` directly (`:8`, `:17`, `:41`). What remains:
   - The real `MapBuilderPage` registration is still untested at both ends. It lives at `frontend/src/pages/MapBuilderPage.tsx:919` (`analysisAddToMap.current = layers.chatLayerActions?.onAddDataset ?? null`), with the consumer ends at `:1896` and `:2189`. The watcher tests hand-seed the module ref, so nothing covers the wiring itself. Note the original issue cited `components/builder/MapBuilderPage.tsx:797-804`; the file is under `pages/`, not `components/builder/`.
   - Extracting an `AnalysisJobStatus` presentational component still has not happened and still makes the status-line coverage cheap. `components/builder/__tests__/AnalysisPanel.test.tsx` is now 1845 lines and spends its first ~110 on eight `vi.mock` blocks before any assertion.

2. **#729's consumer half (P1).** `BuilderMap.tsx` sets the guards at `:536-558` (`measureActiveRef`, `drawActiveRef`) and reads them at `:981`, `:1016`, and `:1046`. The suppression suite covers only the store-flag producer; `measureActiveRef` is equally unpinned. One `BuilderMap.draw-guard` test: set the flag, fire a click over a feature, assert no popup; assert a pre-open popup dismisses. (Original cited `:585-593` and `:941`/`:991`/`:1021`.)

3. **BuilderMap popup pipeline.** The click/mousemove/keydown hit-test path (`BuilderMap.tsx:895-1021` — clusters, `popup_config.enabled === false`, measure/draw guards) has no unit test in any of the **seven** `BuilderMap.*` test files. The original said five; the count has grown while the coverage has not.

## Running these

`cd frontend && npx vitest run src/components/builder/__tests__/<file>`

ishiland:
Tests only — no source change.

## 1. BuilderMap popup pipeline

`BuilderMap.popup-pipeline.test.tsx`. The click/mousemove hit-test path (`BuilderMap.tsx:895-1021`) had no unit test in any of the **seven** existing `BuilderMap.*` files. Covers: a hit on a popup-enabled layer opens the popup with the right layer; `popup_config.enabled === false` opens none; a hit on an unknown (basemap) layer opens none; a later empty click closes an open popup; the pointer cursor is set over an interactive feature and cleared off one.

Built on the `vi.hoisted` fakeMap + `@vis.gl/react-maplibre` recipe from `BuilderMap.terrain-visibility.test.tsx`, with `queryRenderedFeatures` added and `FeaturePopup` stubbed so what the pipeline produced is readable from the DOM.

## 2. #729's consumer half

Same file. The existing suppression suite covered only the store-flag **producer**; `measureActiveRef` and `drawActiveRef` — the refs the handlers actually read at `:981`, `:1016`, `:1046` — were unpinned. Covers: drawing suppresses the popup on click, dismisses one already open, and leaves the cursor alone; popups return when drawing ends; the measurement plugin suppresses the same way.

## 3. MapBuilderPage analysis registration

Added to `MapBuilderPage.header-actions.test.tsx`, reusing its mock set rather than duplicating ~200 lines of it. `AnalysisJobWatcher.test.tsx` hand-seeds the module ref, so the real registration at `MapBuilderPage.tsx:919` was covered at neither end. Covers: the builder registers its own `chatLayerActions.onAddDataset` (asserted by identity, not just non-null) with the open map id; unmount clears both fields so the watcher falls back to "View dataset"; an unsaved builder registers a null map id — which is what the watcher's `mapId === job.mapId` gate turns on.

**Not done: extracting an `AnalysisJobStatus` presentational component.** The issue offers it as a way to make status-line coverage cheap, not as a requirement, and it is a refactor of a 1845-line test file's subject rather than a coverage gap. Worth its own issue if the status line grows.

## Verification

    cd frontend && npx vitest run src/components/builder/__tests__/BuilderMap.popup-pipeline.test.tsx   # 10 passed
    cd frontend && npx vitest run src/pages/__tests__/MapBuilderPage.header-actions.test.tsx            # 8 passed
    cd frontend && npm run lint && npm run typecheck && npm run test:coverage   # 359 files, 4170 tests passed

Closes #943

### DIFF
--- a/frontend/src/components/builder/__tests__/BuilderMap.popup-pipeline.test.tsx
+++ b/frontend/src/components/builder/__tests__/BuilderMap.popup-pipeline.test.tsx
@@ -1,0 +1,341 @@
+// fix(#943): the BuilderMap click/mousemove hit-test path had no unit test in
+// any of the seven BuilderMap.* files, and #729's consumer half was equally
+// unpinned — the suppression suite covered only the store-flag producer, never
+// the two refs the handlers actually read (`measureActiveRef`, `drawActiveRef`).
+//
+// Uses the vi.hoisted fakeMap + @vis.gl/react-maplibre recipe from
+// BuilderMap.terrain-visibility.test.tsx, with queryRenderedFeatures added so a
+// click can resolve a hit, and FeaturePopup stubbed so what the pipeline
+// produced is readable from the DOM.
+
+import type { ReactNode } from 'react';
+import { act, render, screen } from '@/test/test-utils';
+import type { MapLayerResponse } from '@/types/api';
+import type { VectorTileToken } from '@/api/tiles';
+import { BuilderMap } from '../BuilderMap';
+import { useMapDrawStore } from '@/stores/map-draw-store';
+import { usePluginStore } from '@/stores/map-plugin-store';
+
+vi.mock('@/hooks/use-settings', () => ({
+  useBasemaps: () => ({
+    data: [
+      {
+        id: 'openfreemap-positron',
+        label: 'Light',
+        url: 'https://tiles.example.com/styles/basic',
+        enabled: true,
+      },
+    ],
+  }),
+  useMapDefaults: () => ({ data: { center_lng: 0, center_lat: 0, zoom: 2 } }),
+  useTileConfig: () => ({ data: { cdn_base_url: null, mvt_source_layer_prefix: 'data' } }),
+  useEnabledPlugins: () => ({ data: ['measurement'], isLoading: false }),
+}));
+
+const vectorToken: VectorTileToken = {
+  kind: 'vector',
+  sig: 'sig',
+  exp: 9_999_999_999,
+  scope: 'test',
+  expires_in: 900,
+} as VectorTileToken;
+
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

Issue #726, opened 2026-07-26
bug(builder): a clip-drawing vertex click also opens the clicked feature popup


ishiland:
Found while filming the 1.5 analysis demo. Drawing a clip mask over a dense layer leaves a trail of feature popups behind it, and the last one sits on top of the clip result.

## What happens

In `AnalysisPanel`, **Draw clip area** starts a `TerraDraw` polygon mode on the live map. Each click places a vertex. It also falls through to `BuilderMap`'s own click handler, which resolves the feature under the cursor and calls `setPopupInfo`, so the clicked feature's popup opens.

Three consecutive vertex clicks over Manhattan Building Heights, every one of them:

```
vertex at (500,300): popup opened -> MANHATTAN BUILDINGS | Built 1927 | Construction Year | 1,927 | Height Roof | 420.8 ...
vertex at (700,300): popup opened -> MANHATTAN BUILDINGS | Americas Tower Offices | Construction Year | 1,988 ...
vertex at (700,470): popup opened -> MANHATTAN BUILDINGS | Americas Tower Offices | Construction Year | 1,988 ...
clip area registered: true
popups still open after finishing: 1
```

The vertex still lands, so the click does double duty rather than being stolen. `Clip area set` appears and the mask is correct. The damage is cosmetic but persistent: a popup is left open over the result, and Escape closes them one at a time.

## Cause

`BuilderMap`'s click handler has no awareness of an active draw mode. `AnalysisPanel` owns its `TerraDraw` instance directly (the `BboxMapPicker` precedent) and nothing tells the map to stop resolving feature clicks while drawing.

The measurement plugin will want the same suppression if it ever grows a draw mode, so a shared "a draw mode owns the pointer" signal is probably better than a one-off check.

## Suggested fix

Suppress `setPopupInfo` while any draw mode is active. Either a small piece of builder state that `AnalysisPanel` sets in `startDrawing` and clears in `stopDrawing`, or a check on the TerraDraw instance's mode from the map's click handler. Closing any open popup when drawing starts would be a reasonable extra.

## Repro

1. Open a map with a dense polygon layer whose popups are enabled (Manhattan Building Heights on the seeded showcase works).
2. Analysis panel, pick that layer, Operation: Clip, click **Draw clip area**.
3. Click a few vertices over the buildings. A popup opens on each one.
4. Double-click to finish. The last popup is still open, over the area you are about to preview into.

## Workaround

[...]

PR #985, opened 2026-07-30
test(e2e): cover dissolve materialize and drawn-mask clip in the analysis spec

### LINKED ISSUE #945: test(e2e): cover dissolve materialize and drawn-mask clip in the analysis spec
Split out of #789 (item 5), which mixed backend pytest, frontend Vitest, and Playwright work into one issue. #789 keeps the backend items; the Vitest gaps are #943.

`e2e/analysis.spec.ts` contains exactly one test — `:58`, "buffer preview, materialize, and page-level completion toast". Buffer is the only operation covered.

Two gaps worth closing, both chosen because e2e is the only place they can be caught:

1. **Dissolve materialize.** Materializing is the *only* way a user validates dissolve — it has no preview by design — and its pipeline is never wired together anywhere else: the `col:` sentinel, the collision guard, and `enable_hashagg=off` are each unit-tested in isolation but never end to end.

2. **Drawn-mask clip.** The pointer path that regressed twice, in #726 and again in #729. The draw-guard unit test proposed in #943 covers the guard flag; it does not cover a real drawn geometry reaching the clip operation.

## Gotchas

The stack must be running, and the dev stack bind-mounts the **main** checkout. Running `npx playwright test` from a git worktree validates `main`'s app code, not your branch — a spec-only change is the one case that is validly testable that way, and this issue is spec-only, so a worktree is fine here. If any app code changes alongside it, run Vite on the host at `:5174` with `API_PROXY_TARGET=http://localhost:8001` and point `E2E_BASE_URL` at it.

`npx playwright test e2e/analysis.spec.ts --project=chromium`

ishiland:
Spec-only, no app code. `e2e/analysis.spec.ts` had exactly one test, and buffer was the only operation covered.

## Dissolve materialize

Materializing is the only way a user validates dissolve — it has no preview by design (#779) — and the pipeline behind it is never wired together anywhere else: the `col:` sentinel, the collision guard, and `enable_hashagg=off` are each unit-tested in isolation. The test picks Dissolve, asserts the hint that names the only way to run it and the *absence* of the Preview button, groups by the fixture's own column (exercising the `col:` sentinel rather than the no-grouping path), then waits for the job Node-side, asserts the materialize request body, and waits for the completion toast.

## Drawn-mask clip

The pointer path that regressed in #726 and again in #729. The draw-guard unit test in #943 covers the guard flag; nothing covered a drawn geometry actually reaching the operation. The test draws a box with real mouse clicks plus a double-click to close, asserts "Clip area set", then asserts the **preview request body carries a Polygon `mask`** — the half the unit-level guard cannot see.

**It also asserts the preview actually returns features, and that assertion is the valuable one.** Writing this caught a real trap: the builder does not sit over the seeded data by default, so a box drawn at the canvas center landed in the Atlantic and the clip matched nothing. The test now zooms to the layer *with the Analysis panel already open*, so the fit runs against the final layout — and the assertion then only passes if the drawn screen coordinates unproject to the lng/lat the feature is actually at. A version that only checked the request body would have passed with the mask in the wrong hemisphere.

Job-id resolution for cleanup happens before every assertion that can fail, so a Playwright retry cannot leak output datasets into the shared catalog.

## Verification

    npx playwright test e2e/analysis.spec.ts --project=chromium
    # 5 passed — setup, the existing buffer test, both new tests, teardown

Run against a live stack. Note that `analysis.spec.ts` is in none of the `e2e:smoke:*` subsets and the full E2E job is skipped on frontend-only PRs, so CI green here does not execute these two tests — the run above is the evidence.

Closes #945

### DIFF
--- a/e2e/analysis.spec.ts
+++ b/e2e/analysis.spec.ts
@@ -250,11 +250,21 @@ test.describe('builder analysis tools', () => {
     await page.getByLabel('Operation').click();
     await page.getByRole('option', { name: 'Clip', exact: true }).click();
 
+    // Put the data under the cursor, with the panel already open so the fit
+    // runs against the final layout. Without this the map sits at its default
+    // view and a box drawn at the center lands in the Atlantic, so the clip
+    // matches nothing — which is what makes the preview assertion below
+    // meaningful: it passes only if the drawn screen coordinates unproject to
+    // the lng/lat the feature is actually at.
+    // Scoped by layer name: the basemap row's kebab shares the same prefix.
+    await page.getByRole('button', { name: `Layer options for ${seed.title}` }).click();
+    await page.getByTestId('kebab-zoom-to-layer').click();
+    await page.waitForTimeout(1500); // the fly-to animation
+
     await page.getByRole('button', { name: 'Draw clip area' }).click();
     await expect(page.getByText('Draw on the map — double-click to finish')).toBeVisible();
 
-    // Draw a box around the map center, where the single seeded feature sits
-    // (the builder fits to the layer's extent on load).
+    // Draw a box around the map center, which is now the seeded feature.
     const box = await canvas.boundingBox();
     if (!box) throw new Error('map canvas has no bounding box');
     const cx = box.x + box.width / 2;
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #986, opened 2026-07-30
fix(ingest): skip the Mercator clip when the safe envelope degenerates in the source CRS

### LINKED ISSUE #906: fix(ingest): ST_Transform of the global Mercator envelope silently degenerates for narrow-validity CRSs
Found while fixing #888 (PR #899). Pre-existing, unrelated to that fix, and currently silent — which is the worrying part.

`clip_to_mercator_bounds` (`backend/app/processing/ingest/metadata.py`) intersects incoming geometry with `ST_MakeEnvelope(-180, -85.06, 180, 85.06, 4326)`, transformed into the source CRS.

For a CRS with narrow area-of-validity, that transform collapses. Measured in **EPSG:4807** (NTF Paris): every X of the global envelope becomes `197.396`, so the envelope degenerates and the intersection would **empty the entire table**.

The user would see a successful ingest with zero features and no explanation. That is the same failure shape #888 fixed for the 0..360 and polar cases, but by a different mechanism, so #888's warning does not cover it.

## Why #899 didn't fix it

It is orthogonal to the 0..360 convention work, and it surfaced as a test-design problem rather than a product symptom: the grads-CRS regression test in #899 has to drive the private helper directly, precisely *because* `ST_Transform` collapses the envelope in 4807 and the clip would empty every row, hiding the behavior under test. That comment is in the test file.

## Scope worth checking

EPSG:4807 is one of 14 SRIDs in a stock `spatial_ref_sys` that are `GEOGCS` with `UNIT["grad"]`, but narrow validity is not limited to grads CRSs — any projected CRS with a small usable footprint (national grids, local engineering systems) is a candidate. The right first step is probably to enumerate which SRIDs in `spatial_ref_sys` produce a degenerate result for this specific envelope transform, rather than reasoning about it CRS-family by CRS-family.

## Suggested direction

Detect the degenerate transform rather than trying to enumerate CRSs: after transforming the envelope into the source CRS, check it still has non-zero area (and ideally that it still contains the data's own extent). If it does not, either skip the clip for that ingest or clip in 4326 after reprojection instead of in the source CRS. Either way, say something — a silent whole-table drop is the worst available outcome, and #888 already added a `mercator_clip` warning kind to `IngestJob.user_metadata['warnings']` that this could reuse.

## Ordering constraint a fix must not break

`clip_to_mercator_bounds` calls `_shift_zero_to_360_longitudes` at `backend/app/processing/ingest/metadata.py:1338` — **before** the envelope is transformed at `:1343`. That ordering is load-bearing: the shift is what puts a 0..360 Pacific source onto both sides of the antimeridian (#888/#899), and the clip then runs against already-shifted data.

So a degeneracy guard belongs between the shift and the `UPDATE`, and must not be implemented by moving the clip earlier or by short-circuiting out of the function before the shift has run. Skipping the clip for a narrow-validity CRS must still leave the shift applied.

## Precedent for the CRS probe, 90 lines up in the same file

Do not invent a new way to interrogate the CRS. `_shift_zero_to_360_longitudes` already does it, at `metadata.py:1244-1253`: a query against `spatial_ref_sys.srtext` matching `_GEOGRAPHIC_SRTEXT_RE` (`^GEOG(CS|CRS)`, defined at `:1195`) and `_DEGREE_UNIT_SRTEXT_RE` (`UNIT\["degree`, at `:1206`). That pair tests geographic **and** degree-unit, which is what excludes the grads family (EPSG:4807 and its 13 siblings) that this issue's own repro lands on.

That said, the suggested direction above — measure the transformed envelope's area — is still the better primary test, because narrow validity is not a property of the CRS family and no keyword probe will enumerate it. Use the geometry check to decide, and reach for the srtext probe only if a CRS classification is needed alongside it.

See also the umbrella note on #939: there are now four separate mechanisms in this codebase for deciding whether a CRS is in degrees, and this file holds one of them.

## Acceptance criteria

- Ingesting a source in a narrow-validity CRS (EPSG:4807 is the measured one) preserves its features instead of emptying the table.
- When the clip is skipped or relocated for such a CRS, the job records a warning the user can see. The `mercator_clip` warning kind already exists — `backend/app/processing/ingest/warnings.py:45` and `make_mercator_clip_warning` at `:104`, appended via `_append_mercator_clip_warning` (`tasks_common.py:291-300`) — so this needs a new reason, not a new mechanism. A silent whole-table drop is the outcome to eliminate; a silent whole-table *skip* is not much better.
- Geometry genuinely outside the Mercator safe latitude band still gets clipped for ordinary CRSs, with `dropped_features` / `clipped_features` counts unchanged for existing fixtures.
- 0..360 shifting still happens for a narrow-validity source, per the ordering constraint above.
- The enumeration suggested above (which SRIDs in a stock `spatial_ref_sys` degenerate under this envelope transform) is worth doing first and recording in the PR, but the shipped guard should be the geometric check, not the resulting list.

ishiland:
## What changed

`clip_to_mercator_bounds` intersects incoming geometry with the global Mercator safe envelope transformed into the source CRS. For a narrow-validity CRS that transform collapses (EPSG:4807 turns every X into 197.396), and the intersection then silently empties the whole table behind a successful ingest.

## Enumeration (the issue's first step, recorded here)

Probed every SRID in a stock `spatial_ref_sys` (8500 rows) with `ST_Transform(ST_MakeEnvelope(-180,-85.06,180,85.06,4326), srid)`:

- **4415 SRIDs** produce a zero-area or collapsed envelope — EPSG:4807, **every UTM zone** (32633 measured), France's 2154, the Netherlands' 28992, and most national grids.
- **108 SRIDs** error under the transform outright.
- Partial collapses with positive area also exist: EPSG:2263 (NY Long Island ftUS) leaves a **19-square-foot bowtie** across a 3e16 sq ft bbox; EPSG:5070 a 0.04 m² one; EPSG:27700 collapses to a 0.005 m² *square* (area/bbox ratio 1, so no single check covers all shapes).

So this is a class, not a corner — and per the issue, the shipped guard is the geometric check, not the list.

## The guard

New `_mercator_envelope_degenerates` probe, run **between the 0..360 shift and the clip UPDATE** — the #888/#899 ordering constraint is preserved and pinned by a test (a skipped clip still leaves the shift applied). Degenerate means any of:

1. NULL/empty/zero-area envelope or collapsed X/Y range (the 4415);
2. a sliver: area < 1e-6 of its own bbox area (2263, 5070);
3. for **projected** CRSs only (gated via the existing `_GEOGRAPHIC_SRTEXT_RE` probe — see the four-mechanism table on #939): an envelope narrower or shorter than the table's own extent (27700, polar stereo 3031). Geographic CRSs are excluded from this clause deliberately, so genuinely polar geographic data keeps being clipped with unchanged counts (4979 fixture asserts the exact pre-guard counts);
4. the transform raising — probed inside a SAVEPOINT so the phase-2 transaction stays usable; the clip's own UPDATE would have raised identically.

## The warning (new reason, existing mechanism)

- `MercatorClipCounts` gains `clip_skipped: NotRequired[bool]`; `make_mercator_clip_warning` now emits the existing `mercator_clip` kind for a skip even at 0/0 counts.
- Pydantic `MercatorClipDetail.clip_skipped: bool = False` — stored pre-#906 warnings still validate (tested).
- Frontend `IngestWarningsBanner` shows a dedicated `skippedDescription` for the skip instead of claiming geometry was clipped away; key added to **all four locales** (en/es/fr/de).

## API/schema impact

`MercatorClipDetail` gains one optional field. Regenerated in the same commit: `backend/openapi.json`, `frontend/src/types/api.generated.ts`, `sdks/python/geolens/models/mercator_clip_detail.py`, `sdks/typescript/src/client/types.gen.ts`. Hand-typed `frontend/src/types/api.ts` mirror updated additively. No migration.

## Acceptance criteria

- EPSG:4807 ingest preserves features (previously emptied the table) — tested, plus UTM 32633 and the 2263 partial collapse (whose test fails against the old code with `dropped_features: 2`, i.e. the data destruction is real).
- The skip records a user-visible warning via the existing kind — tested end-to-end producer → Pydantic contract → banner rendering.
- Ordinary CRSs keep clipping: 3857 counts shape byte-identical (no `clip_skipped` key), 4979 polar clip counts unchanged.
- 0..360 shift still runs before a skipped clip — pinned with a forced-probe test (no real degree-geographic CRS degenerates).

## Verification

```
uv run pytest tests/test_ingest_mercator_clip.py tests/test_staging_pipeline.py tests/test_ingest_progress.py -v
cd ../frontend && npm run lint && npm run typecheck && npm run test:i18n
npx vitest run src/components/import/__tests__/IngestWarningsBanner.test.tsx
```

Closes #906


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #987, opened 2026-07-30
chore(builder): style-editor leftovers from the 2026-07-30 audit

### LINKED ISSUE #922: chore(builder): style-editor small leftovers from the 2026-07-30 audit (zoom-editor draft clobber, dead props, orphaned i18n keys)
Small confirmed leftovers from the 2026-07-30 style-tab audit (pass 2), batched like #787/#893. All in `frontend/src/components/builder/`. Pointers re-verified on main @ 28e205ef2.

1. **ZoomExpressionEditor clobbers in-progress edits on re-hydration.** The sync effect (`ZoomExpressionEditor.tsx:95-98`) resets local `draft` whenever the `parsed` object identity changes, and `parsed` is `useMemo(() => parseZoomExpression(value), [value])`. Invalid drafts are never emitted (`:105-110`), so when that reset fires, an in-progress edit (an emptied stop field, a lone `-`) is silently reverted.

   It does **not** fire when another control in the panel changes paint, so don't try to reproduce it that way. Every hop between `layer.paint` and the `value` prop is a shallow spread that preserves the underlying array reference: `handlePaintProp` does `{...paint, [key]: value}` (`LayerStyleEditor.tsx:273`), `handlePaintChange` does `{...l, paint: newPaint}` (`hooks/use-layer-map-sync.ts:247`), `buildBuilderControlPaint` does `{...paint, ...overrides}` (`LayerStyleEditor/builder-paint-map.ts:75`), and `getEditableNumericPaintValue` returns the array itself (`LayerStyleEditor/utils.ts:99-103`). There is no `structuredClone` or JSON round-trip anywhere in the builder hooks. So `value` keeps its identity, the memo doesn't recompute, and the effect stays put.

   It fires where the array reference genuinely is replaced. Repro: put an opacity or width control into Varies-by-zoom, empty one stop's value field so the draft is invalid, then trigger a re-hydration — save the map, or switch render mode, or paste a copied style onto the layer. The half-typed stop reverts.

   Fix as originally scoped: guard the reset on an actual value change (deep-compare or serialized compare) rather than object identity. Same family as the applyLayerUpdate stale-ref clobber.

2. **FillEditor's `onFillPatternChange` fallback encodes forbidden semantics, and tests pin it.** `FillEditor.tsx:79` falls back to `(id) => onPaintProp('fill-pattern', id)`, which sets the key to `undefined` instead of deleting it, exactly what the EDIT-05 comment forbids (`LayerStyleEditor.tsx:345-348`). Unreachable in production (the orchestrator always supplies the handler, `LayerStyleEditor.tsx:463`), but `LayerStyleEditor/__tests__/FillEditor.test.tsx:419,439` assert the fallback as if it were the contract — `:439` asserts `onPaintProp('fill-pattern', undefined)` outright. Make the prop required and delete the fallback plus the tests that pin it.

3. **Pattern picker "None" swatch has no visible affordance.** It is a childless button (`FillPatternPicker.tsx:71-81`) carrying `text-2xs font-medium` classes it never uses; sighted users see an empty square with nothing saying "solid". Give it a visible slash or solid glyph. (Its `aria-label` and `title` are already set, so this is a sighted-user gap only.)

   The other half of this item — the `currentColor` CSS previews at `:26-54` matching neither the grey the map draws nor the layer colour — has moved to #914, which owns the tint decision.

4. **Dead fields on the editor mega-prop.** No editor consumes `renderMode`, `clusterAvailable`, or `onStyleConfigChange` from `BaseStyleEditorProps` (`LayerStyleEditor/types.ts:20,32,35`; passed at `LayerStyleEditor.tsx:447,453,456`). `clusterAvailable` also costs a `getRenderAsOptions(layer)` call on every layer/render-mode change (`LayerStyleEditor.tsx:209-212`) to feed a prop nobody reads. Remove all three and the memo.

5. **Eight orphaned i18n keys, present in all four locales.** Zero non-test references: `style.sections.render`, `style.sections.renderDescription`, `style.renderAs`, `style.renderPoints`, `style.renderSymbol`, `style.renderHeatmap`, `style.renderCluster`, `style.raster.renderMode`. The live Render-as section uses `layerEditor.section.renderAs` and the `renderAs.*` namespace. Delete from all four locales in the same commit so the parity gate stays green.

## Acceptance

- Each item lands as its own commit; `cd frontend && npm run lint && npm run typecheck && npm run test:i18n && npm run test:coverage` green after each.

ishiland:
## What changed

Five confirmed leftovers, one commit each.

**1. ZoomExpressionEditor no longer clobbers an in-progress edit.** The sync effect re-seeded local `draft` whenever the memoized `parsed` changed identity, and invalid drafts are never emitted — so a re-hydration that replaces the paint array with an *equal* one (save the map, switch render mode, paste a copied style) silently reverted a half-typed stop. The memo is now keyed on the serialized value, so an unchanged expression keeps its identity and the effect stays put. Same family as the `applyLayerUpdate` stale-ref clobber. The regression pin was confirmed red before the fix.

**2. `onFillPatternChange` is required; the fallback is gone.** The fallback `(id) => onPaintProp('fill-pattern', id)` set the key to `undefined` instead of deleting it — exactly what the EDIT-05 comment forbids. Unreachable in production, but two `FillEditor` tests asserted it as if it were the contract (one asserted `onPaintProp('fill-pattern', undefined)` outright). Both now pin the real handler and additionally assert `onPaintProp` is *not* called for `fill-pattern`.

**3. The pattern picker's "None" swatch has a visible affordance.** It was a childless button carrying `text-2xs font-medium` classes it never used, so sighted users saw an empty square. It now draws a solid block, which is what "no pattern" renders. (Its `aria-label`/`title` were already correct.)

**4. Three unread fields off the editor mega-prop.** No editor consumes `renderMode`, `clusterAvailable`, or `onStyleConfigChange` from `BaseStyleEditorProps`. `clusterAvailable` also cost a `getRenderAsOptions(layer)` call on every layer/render-mode change to feed a prop nobody read; that memo and its import go too.

**5. Eight orphaned i18n keys deleted** from all four locales in one commit, so the parity gate stays green: `style.sections.render`, `style.sections.renderDescription`, `style.renderAs`, `style.renderPoints`, `style.renderSymbol`, `style.renderHeatmap`, `style.renderCluster`, `style.raster.renderMode`. The live Render-as section uses `layerEditor.section.renderAs` and the `renderAs.*` namespace.

## Verification

    cd frontend && npm run lint && npm run typecheck && npm run test:i18n && npx vitest run

One unrelated flake: `StackRow.test.tsx > "a space can be typed into the rename input…"` fails intermittently under the full parallel run and passes in isolation. Confirmed it fails identically on a clean `origin/main` worktree with this diff stashed.

Closes #922


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #988, opened 2026-07-30
fix(builder): clear the unsaved flag when a style Revert restores the saved state

### LINKED ISSUE #913: fix(builder): pending-style Revert leaves the map flagged "Unsaved changes"
Reproduced locally on main @ cb22f5bb1. Pointers re-verified and corrected on main @ 28e205ef2.

## What happens

1. Open a layer's Style tab and change anything (a pattern, an opacity nudge). The "Pending style preview" banner appears.
2. Click the banner's Revert. The layer visibly returns to its saved state.
3. The header still shows the highlighted Save button with "Unsaved changes", and closing the tab now triggers the beforeunload "unsaved changes" prompt even though nothing differs from the server.

Saving at this point is harmless (the save diff finds nothing meaningful), but the user is told they have unsaved work when they don't, and gets a scary prompt on close for nothing.

## Why

`handleRevertToSaved` (`frontend/src/components/builder/LayerStyleEditor.tsx:413-424`) restores the baseline through the ordinary mutation handlers (`onStyleConfigChange`, `onLayoutChange`, `onOpacityChange`), and those handlers mark the map dirty. So the revert itself re-marks the map dirty and nothing ever clears the flag.

The dirty-marking is **one** call site, not many: every layer mutation funnels through `applyLayerUpdate` in `frontend/src/components/builder/hooks/use-layer-map-sync.ts:195-228`, which calls `setHasUnsavedChanges(true)` unconditionally at `:216`. That single function backs the paint, opacity, layout, style-config, label, popup, and visibility handlers. (An earlier revision of this issue pointed at `use-builder-layers.ts:73-86` — that is the `hasUnsavedChanges` useState declaration plus the basemap-config dirty wrapper, and is not involved.)

Because there is only one choke point, the fix is smaller than "add equality checks to every mutation path" suggests.

## Fix sketch

After a banner Revert, recompute instead of assuming: if every local layer now matches the saved baseline, clear the flag. The saved baseline is `savedLayerBaselineRef` (`frontend/src/components/builder/hooks/use-builder-layers.ts:98`). A reasonable junior-sized route is to thread an `onRevertToSaved` callback down from the page, which owns both the baseline and `setHasUnsavedChanges`.

**Do not reuse `hasUnsavedStyleChanges` as the whole-map cleanliness oracle without extending it first.** That helper (`frontend/src/components/builder/LayerStyleEditor/utils.ts:17-28`) compares only `paint`, `layout`, and `style_config`. It ignores `opacity`, `visible`, `sort_order`, `display_name`, `filter`, `label_config`, and `popup_config`. Used as-is it reports a false "Saved" whenever the only outstanding dirt is an opacity change, a reorder, a rename, a visibility toggle, or a layer add/remove — which would break the third acceptance criterion below (rename the map, edit a layer, revert the layer). Either extend the helper to cover the remaining fields, or compare the full layer objects against the baseline.

Careful: only clear when the whole map is clean. Reverting one layer while another layer has pending edits must stay "Unsaved changes". Non-layer dirt (map name, description, basemap) must also keep the flag.

## Acceptance

- Edit one layer, Revert: header returns to "Saved", no beforeunload prompt.
- Edit two layers, Revert one: still "Unsaved changes".
- Rename the map, edit a layer, Revert the layer: still "Unsaved changes".
- Change only a layer's master opacity on layer A, Revert layer B: still "Unsaved changes".
- `cd frontend && npm run lint && npm run typecheck && npx vitest run src/components/builder/hooks/__tests__ src/components/builder/__tests__/LayerStyleEditor.test.tsx`

ishiland:
## What changed

Editing a layer's style, then clicking the "Pending style preview" banner's Revert, left the header on "Unsaved changes" and armed the beforeunload prompt even though nothing differed from the server.

`handleRevertToSaved` restores the baseline through the ordinary mutation handlers (`onStyleConfigChange`, `onLayoutChange`, `onOpacityChange`), and every layer mutation funnels through `applyLayerUpdate`, which calls `setHasUnsavedChanges(true)` unconditionally. So the revert re-marked the map dirty on its way back to clean, and nothing ever cleared the flag.

Rather than special-casing the revert path, the flag is now **recomputed**. A banner Revert calls `onRevertToSaved`, threaded from the page (which owns both `savedLayerBaselineRef` and `setHasUnsavedChanges`) through the editor handlers. The recheck clears the flag only when:

- every layer matches the saved baseline, compared as **whole layer objects** and in order, and
- no map-level field diverges from server state: name, description, legend title, basemap, basemap labels, `basemap_config`, `terrain_config`, and
- no page-owned dirt is outstanding.

Per the issue's warning, `hasUnsavedStyleChanges` is **not** reused as the cleanliness oracle — it compares only `paint`/`layout`/`style_config` and would report a false "Saved" whenever the outstanding dirt is an opacity change, reorder, rename, visibility toggle, or layer add/remove.

Page-owned dirt the hook cannot re-derive from server state (plugin toggles, and anything else going through `markDirty`) sets a flag that blocks the recheck outright, so a plugin-only edit can never be cleared by reverting a layer. The exported `setHasUnsavedChanges` clears that flag whenever it is set to `false`, so save and map-switch reset it.

The recheck runs in an effect rather than inline, so it observes the reverted layers after commit; the request nonce batches into the same render, so one revert costs one pass.

Terrain hydration moved into a shared `savedTerrainConfig(mapData)` helper so the load path and the recheck cannot drift.

## Verification

    cd frontend && npm run lint && npm run typecheck && npx vitest run

359 files / 4162 tests pass. New `use-builder-layers.clean-recheck.test.ts` covers all four acceptance cases (revert one layer → clean; two edited, one reverted → dirty; renamed map + reverted layer → dirty; opacity on layer A + revert layer B → dirty) plus the markDirty block. `MapBuilderPage.settings-wiring.test.tsx` F2 updated: the plugin toggle now asserts on `markDirty` rather than the raw setter.

Closes #913

### DIFF
--- a/frontend/src/components/builder/LayerEditorPanel.tsx
+++ b/frontend/src/components/builder/LayerEditorPanel.tsx
@@ -32,6 +32,10 @@ export interface LayerEditorHandlers {
   // boundaries (e.g. line→arrow leaving behind line-cap / line-join).
   onRenderModeChange?: (layerId: string, mode: RenderAsId) => void;
   onRemove: (layerId: string) => void;
+  /** fix(#913): called after a banner Revert so the page can re-derive whether
+   *  the map is still dirty (the revert itself routes through the ordinary
+   *  mutation handlers, which all mark it dirty). */
+  onRevertToSaved?: () => void;
 }
 
 interface LayerEditorPanelProps {
@@ -555,6 +559,7 @@ export const LayerEditorPanel = memo(function LayerEditorPanel({
                       onOpacityChange={handlers.onOpacityChange}
                       onStyleConfigChange={handlers.onStyleConfigChange}
                       onLayoutChange={handlers.onLayoutChange}
+                      onRevertToSaved={handlers.onRevertToSaved}
                     />
                   )}
                 </div>
--- a/frontend/src/components/builder/LayerStyleEditor.tsx
+++ b/frontend/src/components/builder/LayerStyleEditor.tsx
@@ -41,6 +41,13 @@ interface LayerStyleEditorProps {
   onOpacityChange?: (layerId: string, opacity: number) => void;
   onStyleConfigChange: (layerId: string, config: StyleConfig | null, paint: Record<string, unknown>, opts?: { replace?: boolean }) => void;
   onLayoutChange: (layerId: string, layout: Record<string, unknown>) => void;
+  /**
+   * fix(#913): banner Revert restores the baseline through the ordinary mutation
+   * handlers, each of which marks the map dirty — so the revert re-dirtied the
+   * map and the header kept claiming unsaved work. The owner of the baseline
+   * re-derives cleanliness here; it only clears when the WHOLE map is clean.
+   */
+  onRevertToSaved?: () => void;
 }
 
 // Re-export hasUnsavedStyleChanges so existing callers (tests etc.) can still
@@ -129,6 +136,7 @@ export const LayerStyleEditor = memo(function LayerStyleEditor({
   onOpacityChange,
   onStyleConfigChange,
   onLayoutChange,
+  onRevertToSaved,
 }: LayerStyleEditorProps) {
   const { t } = useTranslation('builder');
   const geomType = getLayerType(layer.dataset_geometry_type);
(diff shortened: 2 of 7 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #989, opened 2026-07-30
test(backend): late-bind fixture-redirected helpers and guard the import shape

### LINKED ISSUE #909: test(backend): module-scope import bindings escape monkeypatch redirects — a test silently hit the dev database
Found and fixed narrowly during #885 (PR #898). Filing the class, because the instance was latent for as long as the fixture existed and the failure mode is silent.

## What happened

`requires_ogr2ogr` in `backend/tests/conftest.py` redirected `app.processing.ingest.ogr.build_pg_conn_str` so ogr2ogr subprocesses in tests point at the test database. But `app.processing.export.ogr` does `from ... import build_pg_conn_str` at import scope (`backend/app/processing/export/ogr.py:11`), which binds its **own** name. Patching the origin module does not touch that binding.

So `run_ogr2ogr_export` in a test read the **dev** database. Nothing exercised that path before #898 added the first test that does, which is the only reason it surfaced now. #898 fixed it by patching both bindings (`backend/tests/conftest.py:1900-1901`).

## Why this is a class, not a one-off

This repo already has a documented instance of the same shape: a top-level `async_session` import binds the dev-DB session before test config can rebind it, and the fix there was to late-bind. Same root cause — `from X import name` at module scope snapshots the value, so any fixture that redirects `X.name` misses every module that imported it directly.

The failure is silent in both directions: a test can read or **write** the wrong database and still pass.

## The live `async_session` bindings

The `async_session` precedent is not just history. `backend/tests/conftest.py:1618-1621` redirects the session factory by assigning to the origin module:

```python
original_engine = db_module.engine
original_session = db_module.async_session
db_module.engine = test_engine
db_module.async_session = test_session_factory
```

That is exactly the origin-module patch a module-scope `from app.core.db import async_session` snapshots past. Three modules still hold their own binding:

- `backend/app/api/main.py:30` — `from app.core.db import async_session, engine`
- `backend/app/core/dependencies.py:6` — `from app.core.db import async_session`
- `backend/app/processing/ai/token_usage.py:14` — `from app.core.db import Base, async_session`

Two more re-export it rather than use it, and are worth checking at the same time because a future consumer importing from them inherits the same snapshot: `backend/app/processing/ingest/tasks.py:29` (an explicit `# noqa: F401` re-export) and `backend/app/processing/ingest/tasks_vrt.py:31`.

Every other consumer in the tree already late-binds inside a function (`from app.core.db import async_session` at call scope), which is why this has not bitten yet. The three above are live instances of the same trap, not hypotheticals.

## Suggested guard

A structural test that, for each name a fixture redirects, asserts every module binding of that name is patched. Concretely: walk `sys.modules` for attributes that are the same object as the un-patched target and fail if any survive after the fixture applies. That is a small amount of reflection and it generalizes to whatever the next redirected helper is, rather than requiring someone to remember this trap.

Cheaper alternative if that proves fiddly: forbid module-scope `from ... import build_pg_conn_str` (and the handful of similar env-sensitive helpers) via a lint rule or a `test_layering.py` check, forcing call sites to `import module` and reference `module.helper` so there is only ever one binding to patch. That trades a small style constraint for making the whole class impossible.

**There is already a working template for the second option.** `backend/tests/test_layering.py:1659`, `test_no_module_level_provider_sdk_imports_in_processing()`, enforces precisely this shape of rule: no module-scope import of a named set of symbols inside a given package. Copying that test and swapping its symbol list is a much smaller job than deriving the guard from scratch, and it lands in the file where the repo already keeps its structural import rules.

Seed list for whichever guard is chosen: `build_pg_conn_str` and `async_session`. Both are env-sensitive, both are redirected by a fixture, and both have at least one module-scope binding today.

Either way the value is the same — the current situation is that a fixture *looks* like it isolates the database and does not.

## Acceptance

1. **Fix the three live bindings first.** Convert `api/main.py:30`, `core/dependencies.py:6`, and `processing/ai/token_usage.py:14` to late-bind inside the functions that use them, matching what the rest of the tree already does. This is the narrow, immediately verifiable part.
2. **Then add the guard, and prefer the structural one.** The `test_layering.py:1659` pattern extended to the seed list is the recommended path: it makes the class impossible rather than detecting it after the fact, it has a precedent in-repo, and it fails loudly at test time instead of silently at database time. The `sys.modules` reflection walk stays the fallback if a symbol turns out to need a module-scope binding for a legitimate reason.
3. A test proves the guard catches a regression: reintroducing a module-scope `from app.core.db import async_session` in `backend/app/` fails the suite.

ishiland:
## What changed

**The three live bindings from the issue, plus two more the new guard found:**

- `backend/app/api/main.py` — module-scope `from app.core.db import async_session, engine` removed; each seed/sweep function and `lifespan` late-binds.
- `backend/app/core/dependencies.py` — `get_db` late-binds.
- `backend/app/processing/ai/token_usage.py` — `record_token_usage` late-binds (`Base` stays module-scope; it is not redirected).
- `backend/app/processing/ingest/tasks_vrt.py` — the issue listed this as a re-export; it actually USES the binding in `ingest_vrt`/`regenerate_vrt` (6 call sites). Both now late-bind.
- `backend/app/processing/embeddings/backfill.py` — found by the guard: `__main__` block held a module-scope binding.
- `backend/app/processing/ingest/tasks.py` — the `# noqa: F401` re-export removed (no consumers; comment explains why it must not return).
- `backend/app/processing/export/ogr.py` — `build_pg_conn_str` moved to call scope; the conftest double-patch from #898 collapses back to a single origin-module patch.

**The guard (`backend/tests/test_layering.py`):** `test_no_module_scope_imports_of_fixture_redirected_helpers`, AST-based rather than the git-grep template so untracked files and multi-name import lists are covered. Seed list per the issue: `async_session` (origins `app.core.db`, `app.core.db.session`) and `build_pg_conn_str` (`app.processing.ingest.ogr`). `app/core/db/__init__.py` is the one allowlisted binding — the package attribute is exactly what conftest patches. `test_fixture_redirect_guard_catches_seeded_violation` proves a seeded `from app.core.db import async_session` is caught (issue acceptance §3).

**Tests that patched the removed snapshot bindings** (the anti-pattern itself) now patch the origin or drop the redundant re-point: `test_seed_admin_concurrency`, `test_seed_bootstrap_identity`, `test_tenant_job_recovery`, `test_regenerate_vrt_integration`, `test_vrt_source_management_174`, `test_ai_token_budget_daily`. `api/main.py`'s LOC ratchet updated (1282 → 1292).

## Verification

- Full backend suite at CI scale: `pytest -n 4` → **6194 passed, 81 skipped, 0 failed** (the six initially failing tests were all patchers of removed bindings; fixed as above and re-run green at scale — this bug class is xdist-sensitive).
- Guard negative-control: seeded violation fails with file:line surfaced.

Closes #909

### DIFF
--- a/backend/app/api/main.py
+++ b/backend/app/api/main.py
@@ -27,7 +27,10 @@ from app.observability.metrics import init_metrics
 from app.platform.cache.provider import init_tile_cache
 
 # settings already imported above for the tempdir override — do NOT reimport
-from app.core.db import async_session, engine
+# fix(#909): async_session/engine are late-bound inside each function that
+# uses them. A module-scope `from app.core.db import ...` snapshots the
+# dev-DB objects before the test fixture rebinds app.core.db, so lifespan
+# seed functions in tests would silently hit the dev database.
 from app.core.db.tenant_session import tenant_job_context
 from app.core.logging_config import setup_logging
 from app.core.tenancy import is_multi_tenant
@@ -81,6 +84,8 @@ async def seed_roles() -> None:
     unique constraint and crashing the loser's startup *before* the admin-seed
     lock is ever reached.
     """
+    from app.core.db import async_session  # fix(#909): late-bind for tests
+
     async with async_session() as session:
         await session.execute(
             text("SELECT pg_advisory_xact_lock(:k)"), {"k": _SEED_LOCK_KEY}
@@ -135,6 +140,8 @@ async def seed_initial_admin() -> None:
     xact-scoped advisory lock makes exactly one worker seed; the rest see
     count>0 and no-op. The lock releases when the session's transaction ends.
     """
+    from app.core.db import async_session  # fix(#909): late-bind for tests
+
     async with async_session() as session:
         await session.execute(
             text("SELECT pg_advisory_xact_lock(:k)"), {"k": _SEED_LOCK_KEY}
@@ -190,6 +197,7 @@ async def sweep_stale_jobs_once(
     ``ingest_jobs`` reads and writes. Recovery is best-effort per tenant: one
     broken tenant must not prevent the remaining tenants from being swept.
     """
+    from app.core.db import async_session  # fix(#909): late-bind for tests
     from app.platform.jobs.router import fail_stale_jobs
 
     if not is_multi_tenant():
@@ -255,6 +263,8 @@ async def sweep_stale_jobs_once(
 
 @asynccontextmanager
 async def lifespan(app: FastAPI):
+    from app.core.db import engine  # fix(#909): late-bind for tests
+
     for attempt in range(1, 4):
         try:
             async with engine.connect() as conn:
--- a/backend/app/core/db/rls.py
+++ b/backend/app/core/db/rls.py
@@ -318,7 +318,10 @@ async def apply_tenancy_rls_from_engine(*, verify_runtime_role: bool = True) -> 
         may pass ``verify_runtime_role=False`` while preparing the schema; API
         and worker bootstrap always use the default verification.
     """
-    from app.core.db.session import engine
+    # fix(#909): via the app.core.db façade, not app.core.db.session — the test
+    # fixture reassigns the façade attribute, so importing from the origin
+    # module resolves the un-patched dev-database engine even when late-bound.
+    from app.core.db import engine
 
     async with engine.connect() as conn:
         # Use AUTOCOMMIT so each ALTER TABLE is its own implicit transaction
--- a/backend/app/core/db/tenant_schema.py
+++ b/backend/app/core/db/tenant_schema.py
@@ -230,7 +230,8 @@ async def apply_tenant_data_schema_from_engine(tenant_id: str) -> None:
     The SECURITY DEFINER function runs in one ordinary transaction.  Tenant
     creation paths must instead pass their existing session directly.
     """
-    from app.core.db.session import engine
+    # fix(#909): façade import — see the note in rls.py.
+    from app.core.db import engine
 
     async with engine.begin() as conn:
         await apply_tenant_data_schema(conn, tenant_id)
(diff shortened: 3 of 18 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
Fixed the open finding on `f1e2f0bf2c` (test_layering.py:1721). You were right, and I confirmed it empirically rather than by reading:

```
facade.engine = sentinel
exec('from app.core.db import engine')          -> sentinel      (patched)
exec('from app.core.db.session import engine')  -> original engine (NOT patched)
```

conftest.py:1620-1621 reassigns `app.core.db.engine` / `.async_session` only, so a `app.core.db.session` import escapes the redirect at **every** scope — late-binding cannot save it, because the attribute it defers to was never patched. Listing that module under a module-scope-only check was therefore just wrong.

Changes:

- Split the guard into two failure modes. `_FIXTURE_REDIRECTED_SYMBOLS` (module-scope only, fix by late-binding) keeps `app.core.db` and `app.processing.ingest.ogr`. New `_UNPATCHED_ORIGIN_SYMBOLS` bans `from app.core.db.session import engine/async_session` at **any** scope; only `app/core/db/__init__.py` is exempt, since its binding is the patch target. Failures now report which mode they are and therefore which fix applies.
- Routed the four remaining offenders through the façade: `app/core/db/rls.py:321` (your cited `apply_tenancy_rls_from_engine`) and `app/core/db/tenant_schema.py:233,241,277`. Those were the only `app.core.db.session` imports under `backend/app/` outside the façade.
- `test_fixture_redirect_guard_catches_unpatched_origin_at_any_scope` asserts both the module-scope and the late-bound form are reported and the façade form is not. Offenders are sorted, since `ast.walk` is breadth-first.

Verified: `tests/test_layering.py` plus the tenancy suites that drive these helpers (`test_dp01_tenant_schema_helpers`, `test_dp05_shard_routing_pgbouncer`, `test_worker_bootstrap_parity`, `test_dp_single_tenant_byte_identical`, `test_event_ingest`, `test_notification_registration`, `test_worker_overlay_resolution`, `test_tenant_job_recovery`) — 136 passed, 6 skipped. `ruff check .` and `ruff format --check` clean. Rebased onto current main.

@codex review

ishiland:
Rebased onto `f05a11117` for up-to-date protection. Content is byte-identical to `1ea0cb8932`, the sha you passed clean at: `git diff 1ea0cb8932 6a5f5eb84 -- backend/` is empty. Re-running so the pass is anchored to the sha that actually merges.

@codex review

PR #990, opened 2026-07-30
fix(analysis): slice wide buffer inputs into sub-zone bands before projecting

### LINKED ISSUE #902: fix(analysis): a single component wider than one UTM zone is still buffered in a bad projection
Follow-up from #891 / PR #900. That PR fixed the **multi**-component case by buffering each part in its own planar projection. A single component that is itself wider than one projection's usable span is unaffected and still wrong.

## Measured

    LINESTRING(0 45, 90 45) buffered 10 000 m
      produced:          85 485 981 084 m²
      piecewise truth:  134 146 143 319 m²
      ratio:            63.7%

Identical before and after #900 — per-component dumping cannot touch it, because there is only one component.

## Why

`ST_Buffer(geom::geography, d)` calls `_ST_BestSRID` on the whole geometry, and that has a **hard threshold at a 45.0° longitude span**: 44.99° picks SRID 999247, 45.0° falls back to **999000 = world Mercator**, whose error is exactly `1/cos(latitude)`. A 90°-wide single LINESTRING lands in world Mercator, so the buffer radius is wrong everywhere except the equator. Past roughly 135° PostGIS climbs into a wide Lambert (999061) which is equal-area, so *area* recovers while *shape* stays deformed by about ±8% — which is why measurements across separations look non-monotonic.

Radius produced for a requested 10 000 m, by span:

| longitude span | SRID | radius |
| --- | --- | --- |
| < 6° | 999031 | 9 997–10 004 |
| 6–44.99° | 999247 | 9 904–10 097 |
| **>= 45°** | **999000 (world Mercator)** | **7 079–7 087** |
| ~135°+ | 999061 | 9 240–10 824 |

Latitude span is harmless out to 80°, so any guard should test longitude only. #900 ships `BUFFER_LOCAL_SRID_SPAN_DEG = 6` for the per-component gate.

## Why it was left out of #900

Fixing it means subdividing *within* a component — splitting the geometry, buffering each piece in its own projection, and dissolving. That changes vertex structure for **every** large single-part buffer, not just seam-crossing ones, so it is a visible output change for existing datasets rather than a bug fix with a narrow blast radius. #900 documents it as explicitly not fixed, in the `render_geodesic_buffer()` docstring.

## Notes for whoever takes it

- A dissolve is mandatory and the ordering is forced. In #900, union-*before*-split aborted the statement outright (`TopologyException: side location conflict at 179.90757318430857 44.915684255255684`). With no dissolve at all, per-component collect regressed to **invalid** geometry (935 908 947 m² against a true 701 993 608 m²). `ST_MakeValid` is not a substitute — it subtracts where a dissolve should merge (468 078 270 m²). The order that worked: buffer per part → seam split (#883) → `ST_UnaryUnion`.
- Measure per-component radius, not pooled area. A pooled ratio hides the failure: in the lat-0 + lat-60 pair the lat-60 member held a **quarter** of its area while its equatorial sibling was exact, and the pooled number read as a mild 0.62 deficit.
- The cost budget from #900: the span guard is free (the `AND` short-circuits), and the measurable cost was the `OFFSET 0` materialization fence — polygons 334.9 → 375.6 ms (+20 µs/row), points actually *faster* 53.2 → 38.4 ms.

## Acceptance criteria

**Subdivision width.** Reuse `BUFFER_LOCAL_SRID_SPAN_DEG` (6, `analysis_sql.py:93`) as the slice width rather than introducing a second constant. The measured table above is the argument: under 6° the radius holds 9 997–10 004 m (±0.04%), while the 6–44.99° band already costs ±1%. Slicing at the same threshold the gate tests keeps one number to reason about, and a slice is by construction in the band where `_ST_BestSRID` picks a local UTM zone.

**Correctness bar.** Per-component produced radius within ±1% of the requested distance, for every component, at every latitude out to ±80°. That is the accuracy the 6–44.99° band already delivers, so it is a bar the codebase has met before rather than a new invention.

**Measure per component, never pooled.** Repeating the warning above because it is the specific way this gets mis-verified: assert the radius from each source vertex to the buffer part containing it. A pooled area ratio hid a component holding a quarter of its area behind an exact equatorial sibling and read as a mild 0.62 deficit.

**Regression fixture.** `LINESTRING(0 45, 90 45)` buffered 10 000 m, asserted against the piecewise truth (134 146 143 319 m²), currently 63.7%. Add a second fixture past 135° so the wide-Lambert band (999061) is covered too — area recovers there while shape stays deformed by about ±8%, so an area-only assertion passes while the output is still wrong.

## Blast radius, and what has to be accepted before starting

This is the reason #900 stopped short, so it should be decided up front rather than discovered in review.

Fixing this changes vertex structure for **every** single-part buffer whose longitude span reaches the threshold — not only seam-crossing ones. The `ST_NumGeometries(...) > 1` half of the current gate (`analysis_sql.py:339`) is exactly what keeps single-part output byte-identical today; the fix necessarily gives that up. Output stays geometrically correct and `ST_IsValid`, but a caller comparing WKT or vertex counts will see a difference.

Tests that exercise the buffer path and should be assumed to move: `backend/tests/test_analysis_preview.py` and `backend/tests/test_analysis_materialize.py`. Any assertion there that pins exact WKT, vertex count, or part count for a wide single-part buffer needs rewriting to assert the geometric property instead. Assertions on genuinely narrow buffers (under the threshold) must **not** move — if they do, the gate is wrong.

Cost budget to stay inside, from #900: the span guard itself is free because the `AND` short-circuits, and the measurable cost there was the `OFFSET 0` materialization fence — polygons 334.9 → 375.6 ms (+20 µs/row), points 53.2 → 38.4 ms. A dense single-part source can reach `MAX_SOURCE_FEATURES` (`analysis_sql.py:48`), so the subdivision must not fire below the threshold.

ishiland:
## What changed

`ST_Buffer(...::geography, d)` projects its whole input into one planar SRID via `_ST_BestSRID`, which falls back to world Mercator at a 45° longitude span. A **single** component wider than one UTM zone was therefore still buffered in a bad projection after #900's per-component pass — there is only one component, so there is nothing to dump into. Measured: a 10 km buffer of `LINESTRING(0 45, 90 45)` produced 85 485 981 084 m² against the 134 146 143 319 m² piecewise truth (63.7%).

The wide branch of `render_geodesic_buffer` now slices the input into longitude bands and buffers each band piece in its own local projection (buffer distributes over union, so slicing + dissolve equals buffering the whole):

- **Great-circle densification first** (`ST_Segmentize` on geography, new `BUFFER_SLICE_SEGMENTIZE_M = 20_000`): geography buffers geodesic edges, so cutting the bare planar chord buffered a different line — measured 141.7e9 m² vs the 134.1e9 truth on the 90° fixture. 20 km keeps chord-vs-arc deviation under 8 m.
- **Band width reuses `BUFFER_LOCAL_SRID_SPAN_DEG`** per the acceptance criteria, cut a hair under (`6 - 0.001`): the SRID switch away from the local UTM zone happens AT exactly a 6.0° span (the same reason the gate tests `>=`) — measured, exact-width slices produced the 9 905 m radius signature of the 999247 fallback where 5.999° slices hold 9 997 m+.
- Pass order is unchanged and still load-bearing: buffer per piece → seam split (#883) → `ST_UnaryUnion` dissolve.

## Blast-radius decision (stated up front, per the issue)

The `ST_NumGeometries(...) > 1` half of the #891 gate is **dropped**. Every buffer whose longitude span reaches the 6° threshold — single-part included — now changes vertex structure; output stays geometrically correct and `ST_IsValid`. Narrow inputs (span < 6°) keep the byte-identical whole-input path, pinned by `test_narrow_single_part_is_still_byte_identical`, the existing narrow-multipart byte-identity test, and the single-point byte-identity sweep (all unchanged and passing).

## Measured against the acceptance criteria

| fixture | area | per-source-vertex radius |
| --- | --- | --- |
| `LINESTRING(0 45, 90 45)` (world-Mercator band) | 134 146 160 613 m² vs truth 134 146 143 319 (+0.00001%) | 9 997 m |
| `LINESTRING(-80 45, 80 45)` (160°, wide-Lambert band) | — | 9 984 – 10 003 m |
| `LINESTRING(0 60, 150 60)` (high latitude) | — | 10 000.4 m |

All within the ±1% bar (best band previously delivered ±0.04–1%). The tests additionally bound every boundary vertex's distance to the source line to [-2%, +1%]: the floor is looser than the bar because `ST_UnaryUnion` noding exposes vertices on the interior of circle-approximation chords (0.48% sagitta at quad_segs 8 plus the neighbor slice's ±0.25%), an artifact even an exact whole-input buffer has as chord dip. Both unfixed failure modes sit far outside it (Mercator −29%; Lambert ±8%, which an area-only assertion cannot see — hence the radius metric on the >135° fixture).

## Tests

- New `TestWideSingleComponentBuffer` (materialize): the two issue fixtures + a lat-60 case + narrow byte-identity.
- `test_buffer_sql_projects_each_component_locally` → `test_buffer_sql_slices_wide_inputs_into_local_projections` (SQL shape).
- All existing #883/#891/#900 fixtures pass unchanged, including the seam-component, dissolve, and byte-identity assertions.

Cost: the span gate short-circuits as before; slicing runs only on the gated branch (span ≥ 6°), swapping the per-component dump for a `generate_series` + `ST_Intersection` in the same cost class.

No schema or API changes.

## Verification

```
uv run pytest tests/test_analysis_materialize.py tests/test_analysis_preview.py tests/test_ai_chat_run_analysis.py -v
```

141 tests pass locally.

Closes #902


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #991, opened 2026-07-30
test(db): pin ORM CheckConstraints to same-named DB constraints

### LINKED ISSUE #942: test(db): alembic check cannot see CHECK constraints, so ORM and migration can drift
Split out of #904, whose other half (four stale comments) shipped in 8e63c095f. Filed on its own because it is a gate hole rather than a documentation fix, and #904's title made it easy to close by accident.

## The gap

Alembic does not compare CHECK constraints, so `make alembic-check` (`Makefile:63`, `alembic check` in the api container) will not notice if the ORM's `CheckConstraint` and a migration's constraint drift apart.

The concrete pair today:

- ORM: `backend/app/modules/catalog/datasets/domain/models.py:66-70` — `chk_records_spatial_extent_type`, allowing `POLYGON` or `MULTIPOLYGON`.
- Migration: `backend/alembic/versions/0030_records_spatial_extent_type.py:49`.

Change either one alone and every gate stays green. #901 pinned the *current* state with DB tests (`backend/tests/test_antimeridian_extent.py:185-260` — rejects a POINT, accepts both areal types, asserts the typmod is generic `Geometry`), but nothing generalizes.

This matters more than one constraint's worth: `AGENTS.md` tells contributors to run `make alembic-check` for schema-adjacent changes, so the gate is carrying more trust than it earns. There are 33 `CheckConstraint(` declarations across `backend/app` today, none of them covered.

## Options, cheapest first

1. Accept the limitation and note it next to the constraint, so the next reader does not assume `alembic check` has them covered.
2. Add a test that reflects every `CheckConstraint` declared on the ORM metadata and asserts a same-named constraint exists in the migrated DB (`pg_constraint` join on `contype = 'c'`).

Option 2 generalizes to any future CHECK, which is the only reason it might be worth more than a comment. Name-existence is the cheap 80%: comparing the *expression* text is not reliable, because PostgreSQL normalizes and re-prints the SQL it stores, so a byte comparison against the ORM string will produce false failures.

## Acceptance criteria

- Deleting `chk_records_spatial_extent_type` from either the ORM model or migration `0030` — but not both — fails CI.
- The check enumerates constraints from ORM metadata rather than a hand-maintained list, so a new `CheckConstraint` is covered the day it lands with no second edit.
- Any constraint deliberately not mirrored (if such a case exists) is exempted by an explicit, commented allow-list rather than by the check silently not looking.
- The test states in a comment that it compares names and not expression text, and why.

Refs #904, #892, #901.

ishiland:
## What changed

New `backend/tests/test_check_constraint_parity.py` (option 2 from the issue):

- `test_every_orm_check_constraint_exists_in_migrated_db` — enumerates every named `CheckConstraint` from `Base.metadata` (no hand-maintained list; a new constraint is covered with no second edit) and asserts a same-named `contype='c'` constraint exists in the migrated test DB (`pg_constraint` join).
- Compares **names, not expression text**, with the reason stated in a comment per the acceptance criteria: PostgreSQL normalizes and re-prints stored SQL, so byte comparison against the ORM string produces false failures.
- Deliberate mismatches go through `_ORM_ONLY_ALLOWLIST` (explicit + commented; empty today), with a staleness assert so removed constraints cannot leave dead allowlist entries.
- Unnamed ORM CheckConstraints fail with instructions to name them — a nameless constraint cannot be parity-checked (none exist today; all declared constraints are named and mirrored).
- `test_parity_gate_catches_a_dropped_constraint` — negative control per the acceptance criteria: drops `chk_records_spatial_extent_type` via uncommitted DDL, confirms the gate flags it, rolls back (test DB untouched), and confirms the constraint is back.

No app code, migrations, or existing tests touched.

## Verification

- `pytest tests/test_check_constraint_parity.py tests/test_antimeridian_extent.py` → 31 passed.
- Negative control exercised as the permanent second test.

Closes #942

### DIFF
--- a/backend/tests/test_check_constraint_parity.py
+++ b/backend/tests/test_check_constraint_parity.py
@@ -15,7 +15,15 @@ This gate closes the hole by (table, name) pair, not by expression:
   cannot masquerade as parity (codex review on the #942 PR).
 - Reverse: a CHECK in the database on an ORM-managed table with no ORM
   declaration is flagged too — that is exactly what an ORM-side deletion
-  leaves behind — unless listed in the explicit DB-only allowlist.
+  leaves behind — unless listed in the explicit DB-only allowlist. The
+  managed-table set comes from ``Base.metadata.tables``, not from the
+  surviving constraint pairs, so deleting a one-check table's only
+  ``CheckConstraint`` cannot remove the table from the comparison.
+
+Both allowlists are themselves checked: an entry has to stay a *live*
+one-sided mismatch. Once the constraint it excuses exists on both sides or on
+neither, the exemption is dead and fails the gate, so a stale entry can never
+pre-authorize the drift it was never meant to cover.
 
 Expression text is deliberately NOT compared — PostgreSQL normalizes and
 re-prints the SQL it stores (casts, quoting, operator spelling), so a byte
@@ -45,6 +53,21 @@ _ORM_ONLY_ALLOWLIST: frozenset[tuple[str, str]] = frozenset()
 # migration with no ORM declaration) on tables the ORM manages. Same rule:
 # explicit pair + comment, or the reverse leg fails.
 _DB_ONLY_ALLOWLIST: frozenset[tuple[str, str]] = frozenset()
+
+
+def _orm_managed_tables() -> set[str]:
+    """Return every table fullname the ORM metadata declares.
+
+    Derived from ``Base.metadata`` rather than from the constraint pairs
+    (codex review on the #942 PR): a table whose *only* ``CheckConstraint``
+    was deleted would otherwise vanish from the managed set, and the reverse
+    leg would filter out the very leftover it exists to catch. Several tables
+    are one-check today (``catalog.record_translations``,
+    ``catalog.ingest_jobs``, ``catalog.map_layers``).
+    """
+    from app.core.db import Base
+
+    return {table.fullname for table in Base.metadata.tables.values()}
 
 
 def _orm_check_constraints() -> dict[tuple[str, str], str]:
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
Both open findings on `281d3cecc4` are fixed, with negative controls.

- **:112 (P2) — managed tables from ORM metadata.** `orm_tables` now comes from `Base.metadata.tables`, not from the surviving constraint pairs. You were right about the concrete blast radius: `catalog.record_translations`, `catalog.ingest_jobs`, and `catalog.map_layers` are all one-check today, so deleting their sole `CheckConstraint` dropped the table out of the comparison entirely. `test_reverse_leg_flags_a_tables_only_orm_constraint_being_deleted` asserts the leftover DB CHECK is now reported, and asserts the pairs-derived set is what used to hide it.
- **:130 (P2) — dead exemptions.** Each allowlist is now validated against its own live one-sided difference (`_ORM_ONLY_ALLOWLIST` against `orm_pairs - db_pairs`, `_DB_ONLY_ALLOWLIST` against `db_pairs - orm_pairs`) instead of the three-way intersection, so an entry whose constraint has vanished from both sides is reported rather than lying in wait. `test_allowlist_entry_is_flagged_once_the_constraint_disappears` covers all three states: gone from both (dead), one-sided (live), present on both (dead).

To make both legs testable without a database, the set comparison moved into a pure `_parity_failures()` helper; the DB-backed test and the original negative control are otherwise unchanged.

Verified: `uv run pytest tests/test_check_constraint_parity.py` — 4 passed. `ruff check` and `ruff format --check` clean. Rebased onto current main.

@codex review

ishiland:
Rebased onto `8e557699d` (up-to-date protection); no content change from the tree you passed clean at `67fc12e00f`. Re-running so the pass is anchored to the exact merge sha.

@codex review

PR #992, opened 2026-07-30
test(db): consolidate the seven _run_alembic helpers with seam-extent normalization

### LINKED ISSUE #933: test(backend): every per-file _run_alembic helper needs seam-extent normalization before downgrade
Deferred from PR #924 (#887), where the immediate instance was fixed with a `clean_tables` fixture. This issue is the class.

## The trap

Migration `0030` (from #892/#901) **deliberately refuses to coerce** on downgrade when any `catalog.records` row holds a MULTIPOLYGON `spatial_extent`. That refusal is correct: the only POLYGON containing a two-ring seam extent is `-180..180`, which silently re-registers the globe-spanning extent the migration exists to eliminate.

The consequence is that **any test which commits a seam-crossing extent breaks every migration test that downgrades past `0030`** — and the failure surfaces in a file with nothing to do with the change that caused it.

Observed on #924:

    tests/test_record_translations_migration.py::test_upgrade_normalizes_legacy_values_and_validates_constraint
      alembic downgrade 0013_backfill_geoparquet_distributions -> returncode 1
      Running downgrade 0030_records_spatial_extent_type -> 0029_api_key_hardening
      RuntimeError: 1 catalog.records row(s) hold a MULTIPOLYGON spatial_extent

Two things make this expensive to diagnose:

- **It does not reproduce from a single-file run.** The mechanism is intra-worker ordering (each xdist worker owns its own database, so there is no cross-worker race) — you need the leaking test and a migration test in the same worker's run.
- **The PR-checks summary truncates the traceback** before the `RuntimeError`. `gh run view <id> --log` and grepping the raw log is what surfaces the actual cause. The truncated form looks like a generic `assert 1 == 0` on an alembic subprocess.

## Why it wasn't fixed in #924

The robust fix is to normalize before downgrading, using the remediation SQL `0030`'s own error message already names — in a test database that loss is fine:

```sql
UPDATE catalog.records SET spatial_extent = ST_Envelope(spatial_extent)
WHERE GeometryType(spatial_extent) = 'MULTIPOLYGON';
```

But there is no single place to put it: **seven separate per-file `_run_alembic` helpers** each shell out to alembic independently (the count was originally given as five; see the table below). Doing this properly means extracting a shared migration-test helper, which is a refactor rather than a fix and did not belong in a raster PR.

## Suggested shape

Extract one shared `_run_alembic` (or a `downgrade_to()` wrapper) used by all seven files, and have it run the normalization before any downgrade that crosses `0030`. Then the guard holds for whatever the next seam-storing test is, instead of relying on every author remembering to clean up.

Worth checking while in there whether any *other* migration has a refuse-to-coerce downgrade with the same property — this is the first one, but the pattern is a good one and will likely recur, and each instance adds another way for an unrelated test to fail.

Related: #892/#884 (PR #901, which added `0030`), #887 (PR #924, where it surfaced).

## The actual inventory

Corrected against `28e205ef2`. Seven files define their own `_run_alembic`, not five, and six of them downgrade:

| file | helper | downgrade target |
| --- | --- | --- |
| `backend/tests/test_saml_column_drift_gate.py` | `:40` | none — never downgrades |
| `backend/tests/test_email_verification_migration.py` | `:40` | `-1` |
| `backend/tests/test_embedding_vector_migration.py` | `:51` | `0011_allow_generic_geometry_type` |
| `backend/tests/test_tenant_rls_migration.py` | `:61` | `-1` |
| `backend/tests/test_record_translations_migration.py` | `:40` | `0013_backfill_geoparquet_distributions` |
| `backend/tests/test_oauth_github_migration.py` | `:45` | below `0010` |
| `backend/tests/test_dormant_tenancy_migration_roundtrip.py` | `:62` | `-1` |

**Current head is `0030_records_spatial_extent_type`.** That is the detail that makes the exposure bigger than it looks: the three files doing a plain `downgrade -1` cross `0030` on their **first step**, so they are just as exposed as the two aiming explicitly at `0011` / `0013`. Only the SAML file is out of scope today, and it becomes exposed the moment someone adds a downgrade to it.

Note the `-1` files are also the ones that go stale silently — `test_email_verification_migration.py:141-146` already carries a comment about its "downgrade is a NO-OP" premise expiring as new heads landed. A shared helper is the place to absorb that drift once.

## Acceptance criteria

- One shared helper, used by all seven files; no per-file `_run_alembic` survives.
- Any downgrade crossing `0030` runs the `ST_Envelope` normalization first, without each test author having to know that.
- A test that commits a seam-crossing extent no longer breaks an unrelated migration test in the same xdist worker. Reproducing the original failure needs the leaking test and a migration test in one worker's run — a single-file run will not show it.
- The helper carries a comment explaining why the normalization exists and that the data loss is acceptable **only** in a test database, so nobody lifts it into application code.
- While in there, check whether any other migration has a refuse-to-coerce downgrade with the same property. `0030` is the first, and each additional one is another way for an unrelated test to fail.

## Relationship to #909

Same theme, different mechanism — worth reading together, not merging.

#909 is about module-scope `from X import name` bindings escaping a fixture's monkeypatch. This one is about duplicated subprocess helpers. Neither causes the other. What they share is the shape of the fix: both want one structural guard in the test harness instead of a convention every future author has to remember, and both fail silently today. Whoever picks up the second one should read the first one's resolution, because a shared answer to "how does this repo enforce test-harness invariants" would serve both.

ishiland:
## What changed

**New `backend/tests/alembic_helpers.py`** — the one shared `run_alembic()` used by all seven former copies. Same env contract throughout: per-worker test DB via `POSTGRES_DB`, `PYTHONPATH`, `-c alembic.ini`, `extra_env` passthrough, and the `EMBEDDING_DIMS`/`ENV_ONLY_CONFIG` strip (previously only in the embedding file, now uniform for determinism — every caller sees exactly what its test injects).

**Seam-extent normalization before every downgrade**: migration 0030 refuses to downgrade while any `catalog.records` row holds a MULTIPOLYGON `spatial_extent`, so a test that committed a seam extent broke whichever unrelated migration test later downgraded past 0030 in the same xdist worker (the #924 failure — not reproducible from a single-file run). The runner executes the `ST_Envelope` remediation from 0030's own error message before any `downgrade`, unconditionally: head is above 0030, so even a plain `downgrade -1` crosses the guard on its first step, and the UPDATE is idempotent and free when no seam extents exist. Guarded for schema states where `catalog.records`/`spatial_extent` do not exist. The helper's docstring says why the data loss is acceptable **only** in a test database.

**Refuse-to-coerce inventory** (acceptance: 'check whether any other migration has the same property'): two more exist — `0010` (refuses while GitHub OAuth providers exist) and `0005` (refuses while tenant-scoped data exists). Documented in the helper with why they are not auto-cleaned: deleting provider rows or tenant data is semantic test state, and the only tests that create it manage their own cleanup.

**The seven files** (`test_saml_column_drift_gate`, `test_email_verification_migration`, `test_embedding_vector_migration`, `test_tenant_rls_migration`, `test_record_translations_migration`, `test_oauth_github_migration`, `test_dormant_tenancy_migration_roundtrip`) now import the shared runner; no per-file `_run_alembic` survives.

**New `backend/tests/test_alembic_helpers.py`** — commits a Fiji-style two-ring extent, then downgrades to 0029 (crossing 0030) via the runner and asserts success, restoring head in a finally.

## Verification

- All seven migration suites + the new test: 55 passed.

Closes #933


ishiland:
@codex review

PR #993, opened 2026-07-30
docs: document migration 0030's refusing downgrade and the page-0 search augmentation

### LINKED ISSUE #946: docs(runbook): document migration 0030's refuse-rather-than-coerce downgrade
Split out of #904. An operator-facing behavior currently documented only in the migration source.

## What is undocumented

Migration `0030_records_spatial_extent_type` **refuses rather than coerces** on downgrade. If any `catalog.records` row holds a MULTIPOLYGON `spatial_extent`, `downgrade()` raises and leaves the column alone (`backend/alembic/versions/0030_records_spatial_extent_type.py:57-79`).

The refusal is correct and should stay: the only POLYGON that contains a two-ring seam extent is `-180..180`, which silently re-registers the globe-spanning extent the migration exists to eliminate. An operator is told the data cannot round-trip instead of having it quietly widened.

The error message is already good — it prints the inspect SQL and the explicit `ST_Envelope` UPDATE needed to proceed deliberately. The gap is that an operator only discovers this mid-rollback, from a `RuntimeError`, with the stack of a failed `alembic downgrade` in front of them.

## Suggested shape

`RUNBOOK.md` has no rollback section today; the closest neighbour is §6 (Major PostgreSQL version upgrade). Either a short subsection there or a new "Schema rollback" section, covering:

- Downgrading past `0030` fails by design when seam-crossing extents exist, and that failure is data protection, not a broken migration.
- The inspect query and the `ST_Envelope` remediation, quoted from the migration's own message so the two cannot drift.
- What is actually lost by running it: each affected record's extent widens to `-180..180`, so Pacific datasets go back to reporting a global footprint.
- Update the RUNBOOK table of contents (`RUNBOOK.md:14-21`) if a new numbered section is added.

Worth checking whether any other migration has a refuse-to-coerce downgrade while writing this — `0030` is the first, and #933 makes the same point from the test-harness side.

## Acceptance criteria

- An operator hitting the `RuntimeError` can find the remediation by searching RUNBOOK for the constraint or migration name.
- The documented SQL matches the migration's error text verbatim.
- The data-loss consequence is stated, not just the mechanics.

Refs #904, #892, #901, #933.

### LINKED ISSUE #949: docs(api): note the page-0 collection augmentation on the dataset search endpoint
Extracted from #569 (item 14), now closed. Small documentation gap, no code change expected.

`GET /api/search/datasets/?q=...&limit=1` can return `numberReturned: 2`, because page 0 is augmented with up to five matching collections alongside the datasets. This is deliberate (`fix(#315)`, `backend/app/modules/catalog/search/router.py:275-291`) and the code comment explains it, but nothing user-facing does.

The README advertises this endpoint, so an API consumer who passes `limit` and counts the response sees the limit apparently ignored. That reads as a bug, and it is the kind of thing that produces a report rather than a question.

One sentence in the API guide, ideally alongside the README's mention of the endpoint, would prevent it: page 0 includes matching collections in addition to the requested number of datasets, so `numberReturned` can exceed `limit` on the first page only.

Docs-site content lives in the getgeolens.com repo, so the fix may end up split across both.

ishiland:
Folds the two small doc issues into one PR per the wave plan.

## #946 — RUNBOOK schema-rollback section

New §7 (plus TOC entry): downgrading past `0030_records_spatial_extent_type` fails **by design** when antimeridian-crossing extents exist — the failure is data protection, not a broken migration (the only POLYGON containing a two-ring seam extent is `-180..180`, which re-registers the globe-spanning extent the migration eliminates). The inspect query and `ST_Envelope` remediation are quoted **verbatim from the migration's error message** so the two cannot drift, and the data-loss consequence is stated: affected extents widen to the full longitude span, so Pacific datasets report a global footprint in catalog/OGC/STAC/search until re-upgraded and recomputed. Searchable by both the migration name and the constraint name per the acceptance criteria. The other two refuse-to-coerce downgrades found while writing this (0010 with GitHub OAuth providers, 0005 with tenant data) get a one-line mention — same check #933 makes from the test-harness side.

## #949 — README note on the search endpoint

One short paragraph after the search example: page 0 of `GET /api/search/datasets/` is augmented with up to five matching collections, so `numberReturned` can exceed `limit` on the first page only (`fix(#315)`, search/router.py). Previously documented only in a code comment; to an API consumer counting results it reads as a limit bug. The docs-site half (getgeolens.com repo) is out of this repo's scope, noted for whoever syncs docs.

Closes #946
Closes #949


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #994, opened 2026-07-30
fix(ai): render the NL-to-SQL prompt's metric buffer from analysis_sql

### LINKED ISSUE #935: fix(ai): the NL→SQL prompt still teaches bare ST_Buffer(::geography), inheriting neither the seam split nor the per-component projection
Split out from #886, where it was listed as a related site but was never in scope. Both fixes it should inherit have now landed, so this is the last place in the product still teaching the broken pattern.

`backend/app/processing/ai/sql_generator.py:235,339` instructs the model to emit:

```sql
ST_Buffer(geom::geography, N)::geometry
```

That is a **separate string surface** from `backend/app/platform/analysis_sql.py`, so it inherited neither of the two fixes applied to the real analysis path:

- **#697 / PR #883** — antimeridian split. A buffer crossing ±180 produces a geometry that is corrupt rather than merely wrong: it `ST_Intersects` a bbox over central France ~15 000 km away.
- **#891 / PR #900** — per-component planar projection. `_ST_BestSRID` picks one working SRID for the whole input, and it has a hard threshold at a **45° longitude span**, falling back to world Mercator (SRID 999000) whose error is `1/cos(latitude)`. Measured: a component at lat 60 kept a **quarter** of its area while its equatorial sibling was exact.

## Why it is lower severity than it sounds, but still real

The output is **display-only** — never persisted. `chat_analysis.py:185` forwards only the parameters back into the now-fixed parameterized path, so a materialized dataset is correct.

But it does reach the user's screen. `chat_geojson.py:320-334` folds a naive min/max bbox over the result; the frontend guard at `ChatPanel.tsx:513` **passes**, because geography output stays within ±180; and `use-ephemeral-layers.ts:124` then `fitBounds` to the whole world with a band drawn across the map. So a user asking the assistant to buffer Pacific or multi-region data sees a visibly wrong overlay and a zoomed-out map, with nothing signalling an error.

## Suggested fix

Have the prompt emit whatever `render_geometry_expr` / `render_geodesic_buffer` produce, rather than restating the SQL as prose in a second place. The duplication is the actual defect — a prompt string that hand-writes SQL will keep drifting from the module that generates it, and this is the second time it has been left behind.

If reusing the generator is impractical for prompt-engineering reasons, then at minimum add a regression test asserting the prompt text and `analysis_sql.py` agree on the buffer shape, so the next divergence fails CI instead of surfacing as a wrong-looking map.

Note there is also an evals suite (`make ai-evals`, live-provider, costs tokens) that exercises NL→SQL — worth checking whether a case there should cover a seam-crossing or multi-region buffer.

Related: #883 (#697), #900 (#891), #886.

## The full site list — there are four, not two

Verified on `28e205ef2`. A grep for the two lines named above leaves two behind:

| line | what it teaches |
| --- | --- |
| `sql_generator.py:194` | `ST_Buffer(geom, radius) -> geometry` in the PostGIS function reference — **no geography cast at all**, so read literally it teaches the degrees form the "Geography Casts" section three pages later exists to correct |
| `sql_generator.py:235` | `For buffer in meters: ST_Buffer(a.geom_4326::geography, 1000)` — the cast cheat-sheet |
| `sql_generator.py:305-309` | the "Buffer + intersect" worked example, a multi-line `ST_Buffer((SELECT geom_4326 ...)::geography, 50000)::geometry` — third instance of the same broken pattern |
| `sql_generator.py:339` | the CTE worked example, `ST_Buffer(geom_4326::geography, 10000)::geometry AS buf` |

`:194` is a different defect from the other three and needs its own call: it is the only one that teaches degrees rather than a wrongly-projected metric buffer. Either give it the geography form or mark it explicitly as the degrees-unit primitive, so the two adjacent sections stop contradicting each other.

The count is itself the argument for the suggested fix. Four hand-written copies of an expression that `render_geodesic_buffer` already generates is exactly how this drifted twice; a fifth will appear the next time someone adds an example.

## Acceptance criteria

- No `ST_Buffer` in the prompt teaches a form that `render_geodesic_buffer` would not produce — all four sites, not the two originally named.
- A regression test asserts the prompt text and `analysis_sql.py` agree on the buffer shape, so the next divergence fails CI rather than surfacing as a wrong-looking map. This is the load-bearing deliverable: without it, fixing the four strings buys one release.
- If the prompt cannot reuse the generator for prompt-engineering reasons, the test still holds, and the reason is written down next to it.
- Check whether `make ai-evals` should carry a seam-crossing or multi-region buffer case. Note it is live-provider and costs tokens, so this is a judgement call to make explicitly rather than by default.

ishiland:
## What changed

`sql_generator.py` taught `ST_Buffer(geom::geography, N)::geometry` at four sites (the function reference at :194 additionally taught the unitless/degrees form), a separate string surface that inherited neither the antimeridian split (#883) nor the per-projection pass (#891/#900/#902). Display-only, but a Pacific/multi-region chat buffer drew a band across the map and fit-bounds'd the overlay to the whole world.

Per the issue's suggested fix, the prompt now emits what `render_geodesic_buffer` produces — rendered at import time from the same generator, which removes the drift class by construction:

- New **Metric Buffers** section embeds the generated expression as a `<GEOM>`/`<METERS>` template with copy-verbatim instructions, plus guidance to prefer `ST_DWithin` when no buffer geometry is needed.
- The **Buffer + intersect** example embeds a fully rendered expression (Denver, 50 km).
- The **CTE example** switches from a buffer to `ST_DWithin` — the better primitive for the question it poses, and one fewer 2 KB embed.
- The **function reference** line now states `ST_Buffer`'s planar/degrees semantics (the :194 defect), and a **Common Mistakes** line marks the bare geography-cast form WRONG.

## The load-bearing deliverable

`backend/tests/test_ai_sql_prompt_935.py` (always-run, no provider):

- both embeds must equal freshly rendered `render_geodesic_buffer` output — an `analysis_sql.py` shape change propagates automatically (import-time render), and unplugging that wiring fails here;
- every other `ST_Buffer` mention in the prompt must be negative teaching (`-- WRONG:` line, the degrees-declaring reference line, or the warning prose) — a fifth hand-written copy fails CI.

## Eval (explicit judgement call, per the issue)

Added `test_seam_crossing_buffer_uses_the_geodesic_expression` to `tests/evals/` with a Fiji-straddling + spread station fixture: the generated SQL must carry the geodesic expression's machinery (`ST_UnaryUnion` + span-gated CASE) and no returned buffer may span ≥ 180° of longitude (the exact corrupt-seam symptom). It is live-provider and costs tokens, so per the wave instructions it is **written and verified to parse/collect** (9 tests collected, skipped without `RUN_AI_EVALS=1`) — **the live run is left to the maintainer**:

```
cd backend && set -a && source ../.env.test && set +a && RUN_AI_EVALS=1 uv run pytest tests/evals/ -v
```

Prompt-size note: the two embeds add ~4 KB to `SQL_SYSTEM_PROMPT`, which is the cacheable static block (per its own docstring), so per-request cost is amortized by provider prompt caching.

No schema or API changes.

## Verification

```
uv run pytest tests/test_ai_sql_prompt_935.py tests/test_sql_engine.py tests/test_chat_narrative.py tests/test_ai_chat_dataset.py tests/test_ai_chat_run_analysis.py -v
uv run pytest tests/evals/ --collect-only -q
```

Closes #935


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #999, opened 2026-07-30
fix(builder): tighten the clean-state recheck against four false verdicts

ishiland:
## What changed

Follow-up to the review on #988 (issue #913), which landed with four open findings. All four are in the map-cleanliness recompute the banner Revert triggers.

**1. Folder expansion was not compared** (the only one that loses data). Expansion state is persisted — `prepareLayersForPersistence` reads `groupMeta` — and `handleToggleGroupExpand` marks the map dirty. But `computeMapIsClean` compared layers and map fields only, so expanding a saved folder, then editing and reverting a child's style, reported clean; the flag cleared, `useBuilderSave` stamped the unsaved expansion into its baseline, and the change was dropped. `groupMeta` is now compared against the hydrated saved shape via a shared `savedGroupMeta(mapData)` helper, mirroring the load path so the two cannot drift.

**2. Opaque dirt was over-applied.** `markDirty` set `opaqueDirtyRef`, which blocks the recheck outright — but `MapTitleBar` (name/description) and the basemap swap/remove handlers call it, and those fields *are* compared. Once any of them fired, the unsaved indicator could never clear for the rest of the session, even with the map matching the server. The opaque flag is now its own marker, `markOpaqueDirty`, used only where the hook genuinely cannot compare: the dock's notes (page state) and plugin toggles (separate store). `markDirty` is a plain dirty mark again.

**3. Normalization did not match the save payload.** `handleSave` persists `localName || undefined`, `localDescription.trim() || null`, and a trimmed-or-null legend title. Comparing the raw local strings meant a description saved with surrounding whitespace (or a whitespace-only one) reported the map permanently dirty. The comparison now normalizes the same way the payload does.

**4. A failed recheck never retried.** The effect fired only on the nonce, so a revert racing the post-save refetch compared against stale `mapData`, failed once, and left the indicator stuck. A request now stays pending until it actually finds the map clean, and re-runs when the saved baselines it compares against change.

## Verification

    cd frontend && npm run lint && npm run typecheck && npx vitest run

363 files / 4212 tests. Four new pins in `use-builder-layers.clean-recheck.test.ts`, each confirmed red against the pre-fix hook: re-derivable edit put back clears; whitespace-only description clears; pending folder expansion stays dirty; opaque dirt stays dirty. `MapBuilderPage.settings-wiring.test.tsx` F2 now asserts the plugin toggle uses `markOpaqueDirty`.

One unrelated flake: `StackRow.test.tsx > "a space can be typed into the rename input…"` fails intermittently under the full parallel run and passes in isolation; it reproduces on a clean `origin/main` with this diff stashed.

Refs #913

### DIFF
--- a/frontend/src/components/builder/__tests__/MapBuilderPage.settings-wiring.test.tsx
+++ b/frontend/src/components/builder/__tests__/MapBuilderPage.settings-wiring.test.tsx
@@ -14,10 +14,11 @@ import { usePluginStore } from '@/stores/map-plugin-store';
 
 let mockMapData: Record<string, unknown>;
 
-const { mockSetHasUnsavedChanges, mockSetBasemapConfig, mockMarkDirty } = vi.hoisted(() => ({
+const { mockSetHasUnsavedChanges, mockSetBasemapConfig, mockMarkDirty, mockMarkOpaqueDirty } = vi.hoisted(() => ({
   mockSetHasUnsavedChanges: vi.fn(),
   mockSetBasemapConfig: vi.fn(),
   mockMarkDirty: vi.fn(),
+  mockMarkOpaqueDirty: vi.fn(),
 }));
 
 vi.mock('react-i18next', () => ({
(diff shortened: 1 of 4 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

Issue #1000, opened 2026-07-30
ci: enable a merge queue on main (and fix ci.yml's merge_group conditionals first)


ishiland:
## Why

`main` had "require branches to be up to date before merging" (`strict: true`) enabled. Combined with a CI cycle of 10-15 minutes and several PRs merging per hour, that produced a livelock during the 2026-07-30 remediation wave: a PR would call `update-branch`, start a fresh ~13-minute CI cycle, and flip back to `BEHIND` before its checks finished, so auto-merge never fired. One PR (#980) burned five full CI cycles and reached a single pending check twice without ever landing. At peak, 13 PRs sat armed with auto-merge and stalled.

Worth recording, because it is not obvious: **auto-merge does not update a stale branch in this repo.** PRs sat at `BEHIND` with `auto=true` untouched for 8+ minutes. `--auto` closes the update-then-merge race but cannot drain a backlog on its own; something still has to call `update-branch`, which is exactly the contention point.

`strict` was turned off on 2026-07-30 to drain the wave. That is a stopgap: it restores throughput by giving up the guarantee that each merge was tested against the exact current `main`. A merge queue restores that guarantee *and* raises throughput, because it can test a batch of PRs together instead of one per CI cycle.

## The trap that has to be fixed first

Do not enable the merge queue before doing this work. All three required contexts (`Detect Changes`, `Pre-commit`, `CI OK`) come from `.github/workflows/ci.yml`, so adding `merge_group:` to its `on:` block is a one-line change. The job conditionals underneath make a naive enable actively dangerous:

- **12 of the workflow's 25 jobs** are gated on `needs.changes.outputs.X == 'true' || github.event_name == 'push'`. In a queue the event is `merge_group`, matching neither `push` nor `pull_request`, so all twelve skip. They are `backend-lint` (:227), `installer-test` (:319), `backup-roundtrip` (:378), `openapi-snapshot` (:637), `sdks-check` (:673), `cli-test` (:773), `mcp-test` (:897), `backend-test` (:966), `alembic-clean-db` (:1144), `frontend-lint` (:1176), `frontend-test` (:1237), and `security-scan` (:1318).
[...]

PR #1002, opened 2026-07-31
fix(sandbox): recalibrate the amplification guards to the banded buffer renderer

### LINKED ISSUE #1001: fix(sandbox): the canonical geodesic buffer is refused on the fail-closed allowlist (17 names), so every buffer question is dead
**Size L. This is design work, not a fix to be picked up casually.** The obvious approaches have been tried and rejected with reasons; read "Why the guards came out" before proposing one.

Rewritten 2026-07-31 against `main` at `9a788ba92`. The previous body described the fix as resolved in a linked PR. That PR was #1002 and it never merged. #1003 merged instead, and it went the other way.

## Live impact

**Every buffer question in the NL to SQL surface is refused today.** Not silently wrong, refused, which was the deliberate choice, but it is a capability that is off.

Reproduced against `9a788ba92`:

```python
import sqlglot
from app.platform.analysis_sql import render_geodesic_buffer

sql = "SELECT " + render_geodesic_buffer("ST_MakeValid(geom_4326)", 1000.0) + " AS g FROM data.t"
validate_sql(sql)
```

```
sandbox.unlisted_postgis_function  function=st_unaryunion
SandboxError: Query uses a disallowed spatial function
```

`st_unaryunion` is simply the first name the walk reaches. The full gap is seventeen.

## How it got here

#994 added the prompt's live-rendered buffer test together with the SEC-025 amplification guards from the #935 line. Those guards were written against the buffer expression as `render_geodesic_buffer` emitted it *at the time*. #990 then rewrote that renderer to slice wide buffer inputs into longitude bands before projecting.

Neither PR is wrong on its own, and both were green. They were never green *together*: #990 merged at 22:06 UTC, #994 at 23:39 UTC with head `5c387e937f`, which does not contain #990. Branch protection had `strict` (require branches up to date) turned off at the time to drain a merge livelock, so nothing forced #994 to re-test against the newer `main`. That prevention question is #1000; this issue is the repair.

## Current state, all three parts verified

**The renderer still emits the banded buffer.** `render_geodesic_buffer` (`backend/app/platform/analysis_sql.py:241`) bands wide inputs, segmentizes per component, and dissolves with `ST_UnaryUnion`. Nothing about #990 or #902 was reverted, and it should not be: the banding is what keeps a wide buffer from landing in world Mercator and coming back cos(latitude) too small.

**The prompt still mandates it verbatim.** `backend/app/processing/ai/sql_generator.py:201` builds `_GEODESIC_BUFFER_TEMPLATE` by calling `render_geodesic_buffer("<GEOM>", 99999)` at import time, and `:204` builds a worked Denver example the same way. So the model is taught to emit exactly the expression the sandbox refuses.

**The validator refuses it on the name allowlist.** `_check_function_allowlist` (`backend/app/platform/sandbox/validator.py:310`) walks `find_all(exp.Func)` and requires each name in `_ALLOWED_FUNCTIONS` (:113) or, for `st_`-prefixed names, `_ALLOWED_POSTGIS_FUNCTIONS` (:263). Walking the rendered expression against those two sets gives seventeen names present in neither:

```
exists                     st_dumpsegments      st_segmentize
exploding_generate_series  st_intersection      st_shiftlongitude
st_collectionextract       st_isempty           st_unaryunion
st_collectionhomogenize    st_makeenvelope      st_wrapx
st_dimension               st_makevalid         st_xmax
st_dump                    st_xmin
```

Two of those are not PostGIS functions at all. `exists` is sqlglot's canonical name for `exp.Exists`, and `exploding_generate_series` is what it calls `generate_series` in FROM position. The rest are the banding, seam-splitting, and dissolve machinery.

`exists` is being handled separately in **#1017**: it is a two-line skip beside the existing `exp.Connector` branch, it is independent of everything below, and it breaks every non-buffer `EXISTS` question on its own. Landing #1017 does not admit the buffer, and this issue does not need to wait for it.

## Why the guards came out, and why not to put them back the same way

#1003 (`bf1696e5c`) removed #994's five allowlist entries **and** all three cost guards. What is left of `_validate_function_cost` (`validator.py:286-298`) is the `st_collect`/`st_union` arity check and the `st_buffer` segments check. `_select_subtree_scans_a_table` and the `ST_Segmentize` numeric floor are gone entirely.

The reasoning, from that commit, is the part worth keeping:

> Three attempts to recalibrate them in place (#1002) each drew a real P1, and each fix opened the next hole. The last settles it: the buffer segmentizes an alias, `_pb_d0.c0`, several derived levels from its input, so proving a target's units and admitting the canonical buffer are contradictory under any target-inspection scheme. That is data-flow analysis, not a predicate to tighten.

Read that literally, because it is the constraint on any proposed fix. A guard that inspects a function's *argument* to decide whether the call is safe cannot work here. The renderer hands `ST_Segmentize` an alias produced several `SELECT` levels above the call site, so the guard sees a column reference and has no way to know its units or its cardinality without following the alias back through the derived tables. Tightening the predicate does not fix that; it only moves which shape gets wrongly admitted or wrongly refused. Anyone whose proposal starts "just also check that the argument is..." is on attempt four.

The removal was also the *stricter* choice, not a capitulation. #994 had added `st_dump`, `st_dumpsegments`, `st_segmentize`, `generate_series` and `st_unaryunion` to the allowlist and then bounded them with guards that had known holes. Taking both out puts the amplification-prone functions back outside the fail-closed allowlist entirely.

And #994's prompt change stayed on purpose. A full revert would have restored the hand-written buffer teaching, and `st_buffer` is allowlisted independently of #994, so the model would emit the bare `ST_Buffer(geom::geography, N)::geometry` form and it would **execute**, returning the world-spanning polygon across ±180° that #883, #900 and #990 fixed. A refused buffer question beats a silently wrong overlay and bbox. Any fix here has to keep that property: nothing should land that admits the naive form.

## What a solution has to do

Readmit the seventeen names without readmitting the amplification classes they were kept out for, given that argument inspection at the call site cannot distinguish them. Some directions, none of them free:

- **Recognize the generated expression as a unit** rather than validating its parts independently, for example by matching the rendered template structurally, since the renderer and the prompt are the same generator and the shape is known ahead of time. The cost is a coupling between the sandbox and the renderer that has to be enforced somewhere, which is the coupling this whole incident was about.
- **Do the data-flow analysis** the commit message names: resolve aliases back through the derived tables so a guard can reason about what actually reaches `ST_Segmentize` and the set-returning functions. Correct in principle, and a genuine piece of work.
- **Move the trust boundary** so that a query the model produced from the trusted template is validated differently from one assembled from arbitrary model output, if that distinction can be made soundly.

Pick deliberately. The one thing that is settled is that a fourth round of predicate tightening is not the answer.

## Acceptance

- `test_rendered_buffer_expression_is_refused_until_1001_lands` (`backend/tests/test_ai_sql_prompt_935.py:73`) flips back to asserting **admission**. Its docstring already says this is the trigger, so this test is the definition of done. Rename it back when it flips.
- **Every readmitted function ships with a paired negative control.** This is not optional and it is what #1003 removed. Recover the five from `bf1696e5c`'s diff of `backend/tests/test_sec025_sandbox_allowlist.py` rather than writing them fresh, because each one encodes an abuse a codex review round actually found:
  - `test_rejects_table_scanning_unary_st_collect_however_aliased` (a table scan hidden one derived level down, or borrowing the canonical alias names)
  - `test_rejects_tiny_or_dynamic_segmentize_length` (`1e-100`, and a column reference as the max-edge argument)
  - `test_allows_canonical_scale_segmentize` (the positive control that pairs with it)
  - `test_rejects_table_scanning_st_dump` (three shapes: `CROSS JOIN LATERAL ST_Dump`, `(ST_Dump(col)).geom` over a table, and `LATERAL ST_DumpSegments`)
  - `test_rejects_unknown_postgis_function`, whose stand-in unknown was reverted from `ST_GeneratePoints` to `ST_MakeEnvelope`; if `st_makeenvelope` is readmitted, this test needs `ST_GeneratePoints` back as its unknown
- The naive `ST_Buffer(geom::geography, N)::geometry` form is still refused or still absent from what the prompt teaches, so the ±180° regression cannot come back through this door.
- `cd backend && uv run pytest tests/test_ai_sql_prompt_935.py tests/test_sec025_sandbox_allowlist.py -v` passes.

## The part worth keeping from the original filing

Guards that encode the shape of a generated template are coupled to that generator, and nothing in the repo expresses that coupling. `render_geodesic_buffer` can be edited by anyone working on analysis geometry with no signal that a sandbox guard depends on its output shape. #994's live-rendering test is the right instinct and is what caught this; it just needs to run on the combination, which is what `strict` or the merge queue (#1000) guarantees.

Whatever lands here should leave a comment at `render_geodesic_buffer` pointing at the sandbox, so the coupling is visible from the side that changes more often.

ishiland:
Restores `main`, which is red on `test_ai_sql_prompt_935.py::test_rendered_buffer_expression_passes_the_sandbox_validator`.

Closes #1001.

## What broke

#990 rewrote `render_geodesic_buffer` to slice wide buffer inputs into sub-zone bands. #994 added the SEC-025 amplification guards, which were written against the render shape that existed before that rewrite, together with the live-rendering test that catches the mismatch. Both PRs were green and neither is wrong on its own. They were never green together: #990 merged at 22:06Z, and #994 merged at 23:39Z with a head that does not contain it. Branch protection had `strict` off at the time to drain a merge livelock, so nothing forced #994 to re-test against the newer `main`.

The practical effect is worse than a red test. All three guards reject the canonical buffer expression that the NL→SQL prompt instructs the model to copy verbatim, so no buffer question could execute at all.

## Changes

**`_select_subtree_scans_a_table` no longer counts a function as a relation.** sqlglot parses a set-returning function in FROM position as an `exp.Table` whose `this` is the function, so `FROM ST_DumpSegments(...)` and `FROM generate_series(...)` both read as table scans. #990's banding put them in the aggregate's own scope, so the unary `ST_Collect` exception stopped applying and the buffer rejected itself. The real relation was never in that scope, only further up the ancestor chain. Only a named relation counts now; anything unrecognized still counts, so the check stays fail-closed.

**`exp.Exists` is skipped as a callable.** #990's banding predicate uses `EXISTS (SELECT 1 FROM ST_DumpSegments(...) ...)`, and sqlglot models `exp.Exists` as a `Func` subclass, so the allowlist saw an unlisted function named `exists`. This is the same trap #538 hit with `AND`/`OR`, and it takes the same fix. `find_all` still walks the subquery, so its contents are validated on their own merits.

**The `ST_Segmentize` floor follows the unit.** It was a flat `>= 1000`, reasoning that "1000 in any unit the sandbox serves cannot amplify." #990 added a degree-unit call for areal input, `ST_Segmentize(geom, 0.1)`, which is about 11 km but far below a metres floor. The geography cast now picks the unit: metres keep the 1000 floor, bare geometry gets 0.01°, roughly 1.1 km. A full 360° span segmentizes to about 36k vertices at that floor, and it is 10x finer than the canonical 0.1.

**`st_dimension` added to the PostGIS allowlist**, which #990's areal/linear branch needs. It reads one geometry's topological dimension and returns a small int.

## Security impact

Three of these loosen a fail-closed guard, so each ships with a paired negative control rather than only the passing case:

- A genuine relation in the same scope as a FROM-position function still rejects, so the narrowed table check cannot be used to hide a table scan.
- A blocked function and an unbounded aggregate hidden inside an `EXISTS` subquery both still reject, proving the walk reaches inside it.
- `1e-9` degrees still rejects, and the metres floor is unchanged, so the bare-geometry path does not become a way to smuggle a tiny length past the guard.

No schema, API, or config impact.

## Verification

```
uv run pytest tests/test_sec025_sandbox_allowlist.py tests/test_ai_sql_prompt_935.py tests/test_sandbox.py -q   # 162 passed
uv run pytest -q -n 4                                                                                          # full suite
```

## Follow-up

Guards that encode the shape of a generated template are coupled to that generator, and nothing in the repo expresses that coupling. Someone editing `render_geodesic_buffer` gets no signal that sandbox guards depend on its output. #1001 notes a comment at the renderer pointing back at the guards as a cheap improvement, and #1000 (merge queue) is the structural fix for the class of failure that let this land.

### DIFF
--- a/backend/app/platform/sandbox/validator.py
+++ b/backend/app/platform/sandbox/validator.py
@@ -276,6 +276,10 @@ _ALLOWED_POSTGIS_FUNCTIONS: frozenset[str] = frozenset(
         "st_centroid",
         "st_collect",
         "st_contains",
+        # fix(#1001): #990's banding branches on ST_Dimension to segmentize
+        # lines and areas differently. It reads one geometry's topological
+        # dimension and returns a small int — no rows, no amplification.
+        "st_dimension",
         "st_distance",
         "st_dwithin",
         "st_intersects",
@@ -380,11 +384,22 @@ def _select_subtree_scans_a_table(func: exp.Func) -> bool:
     already reads. A collect whose enclosing SELECT (including nested derived
     tables) scans a real table aggregates unbounded rows and stays rejected,
     however it is aliased.
+
+    fix(#1001): a set-returning function in FROM position parses as an
+    ``exp.Table`` whose ``this`` is the function, so counting every
+    ``exp.Table`` read ``FROM ST_DumpSegments(...)`` and ``FROM
+    generate_series(...)`` as table scans. #990's sub-zone banding put both in
+    the aggregate's own scope, which made the canonical buffer reject itself.
+    A function call is not a relation, and its own amplification is policed by
+    the generate_series shape check and the ST_Dump guard below. Only a named
+    relation counts here; anything unrecognized still does, to fail closed.
     """
     sel = func.find_ancestor(exp.Select)
     if sel is None:
         return True  # fail closed: no recognizable scope
-    return any(True for _ in sel.find_all(exp.Table))
+    return any(
+        True for tbl in sel.find_all(exp.Table) if not isinstance(tbl.this, exp.Func)
+    )
 
 
 def _validate_function_cost(func: exp.Func, fn_name: str, sql: str) -> None:
@@ -400,14 +415,28 @@ def _validate_function_cost(func: exp.Func, fn_name: str, sql: str) -> None:
     # fix(#935 codex r2): ST_Segmentize's max-edge argument controls vertex
     # amplification — a tiny or dynamic length manufactures an enormous
     # vertex count before any outer LIMIT applies. Require a numeric literal
-    # of at least 1000 (the canonical buffer uses 20 000 metres; 1000 in any
-    # unit the sandbox serves cannot amplify).
+    # no finer than the floor for its unit.
+    #
+    # fix(#1001): that floor was a flat 1000 on the assumption that the sandbox
+    # only ever segmentized in metres. #990's banding added a degree-unit call
+    # for areal input (ST_Segmentize(geom, 0.1), ~11 km) alongside the existing
+    # geography one (ST_Segmentize(geom::geography, 20000)), and 0.1 tripped a
+    # metres floor. The cast picks the unit: geography is metres, bare geometry
+    # is SRID units, i.e. degrees for the 4326 the sandbox serves. 0.01° keeps
+    # the two overloads bounded to comparable vertex budgets over a global
+    # extent — 40 075 km / 1000 m is ~40k vertices, 360° / 0.01° is 36k — so
+    # neither unit is the cheap way past the guard.
     if fn_name == "st_segmentize":
         length = func.expressions[1] if len(func.expressions) > 1 else None
+        target = func.expressions[0] if func.expressions else None
+        in_metres = (
+            isinstance(target, exp.Cast) and "geography" in target.to.sql().lower()
+        )
+        floor = 1000.0 if in_metres else 0.01
         segmentize_ok = (
             isinstance(length, exp.Literal)
             and length.is_number
-            and float(length.this) >= 1000
+            and float(length.this) >= floor
         )
         if not segmentize_ok:
             logger.info("sandbox.unbounded_segmentize", sql=sql)
@@ -482,6 +511,14 @@ def _check_function_allowlist(stmt: exp.Expression, sql: str) -> None:
         # their operands regardless, so a disallowed function inside either
         # side of an AND/OR is still caught below.
         if isinstance(func, exp.Connector):
+            continue
+        # fix(#1001): sqlglot models EXISTS as a Func subclass too, so #990's
+        # banding predicate (EXISTS (SELECT 1 FROM ST_DumpSegments(...)))
+        # parsed as an unlisted function named "exists". EXISTS is a
+        # short-circuiting predicate that yields one boolean, not a callable,
+        # and find_all still walks its subquery — every function and table in
+        # there is checked on its own merits below.
+        if isinstance(func, exp.Exists):
             continue
         if isinstance(func, exp.Anonymous):
             fn_name = func.name.lower() if hasattr(func, "name") else ""
--- a/backend/app/platform/sandbox/validator.py
+++ b/backend/app/platform/sandbox/validator.py
@@ -402,8 +402,46 @@ def _select_subtree_scans_a_table(func: exp.Func) -> bool:
     )
 
 
-def _validate_function_cost(func: exp.Func, fn_name: str, sql: str) -> None:
-    """Reject function arguments that can amplify a one-row query into a DoS."""
+def _statement_reprojects_off_4326(stmt: exp.Expression) -> bool:
+    """True when the statement reprojects to anything but a literal 4326.
+
+    fix(#1001 codex r1): the segmentize floor infers its unit from the target,
+    but ST_Transform is allowlisted, so a bare geometry is not necessarily in
+    degrees — ST_Transform(geom_4326, 3857) makes 0.01 a centimetre, and a
+    world-scale feature at that edge length is billions of vertices. Alias
+    indirection makes the target's own CRS unknowable in general (the banded
+    renderer segmentizes ``_pb_d0.c0``, several derived levels from its input),
+    so this asks the whole statement instead: any reprojection away from 4326,
+    or to an SRID we cannot read as a literal, forces the metres floor on every
+    segmentize in the query. The canonical buffer reprojects via ``::geography``
+    casts and contains no ST_Transform, so it is unaffected.
+    """
+    for fn in stmt.find_all(exp.Func):
+        if isinstance(fn, exp.Anonymous):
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
Closing in favour of #1003, which reverts #994 outright.

Three rounds here, three real P1s, each fix opening the next hole. The third finding settles it: the canonical buffer's own segmentize target is an alias (`_pb_d0.c0`), several derived levels from its input, so "prove the target's units" and "admit the canonical buffer" are contradictory requirements under any inspection scheme. The same alias indirection defeats deciding whether a correlated source traces to a real relation, since the buffer's input legitimately does. That is data-flow analysis, not a predicate to tighten, and it does not belong in a hotfix for a red `main`.

Both P1s in the last round were correct and neither is addressed by this branch — nothing here should be cherry-picked as-is.

Reverting is also the stricter endpoint, which is what decided it. #994 added `st_dump`, `st_dumpsegments`, `st_segmentize`, `generate_series` and `st_unaryunion` to the allowlist and then bounded them with these guards; removing both puts them back outside the fail-closed allowlist entirely.

The two changes worth keeping are recorded in #1001 for the redo: the `exp.Exists` skip (the #538 `exp.Connector` class, one node type over) and `st_dimension` in the PostGIS allowlist. The redesign direction is whole-template matching with the input as a single hole — the canonical shapes are input-independent, which is what makes it tractable.

PR #1003, opened 2026-07-31
revert(sandbox): drop #994's allowlist entries and cost guards, keep its prompt

ishiland:
Restores `main`, red since #994 landed, which blocks #980, #981 and #965 — all three codex-clean and waiting on nothing else.

Refs #1001, which carries the redo.

## Why a revert and not a fix

#994's cost guards were written against the buffer expression as `render_geodesic_buffer` emitted it *before* #990 rebuilt that renderer to band wide inputs. The two merged 93 minutes apart and never ran together, because `strict` was off. All three guards reject the canonical buffer the prompt instructs the model to copy verbatim.

I tried recalibrating them in place (#1002). Three rounds, three real P1s, each fix opening the next:

1. Exempting function-backed FROM sources let a correlated lateral dump scan an outer table: `SELECT x.geom FROM data.cities c CROSS JOIN LATERAL (SELECT d.geom FROM ST_DumpSegments(c.geom_4326) d) x`.
2. Inferring degrees from "not a geography cast" admitted projected geometry, where `ST_Segmentize(ST_Transform(geom_4326, 3857), 0.01)` is one centimetre.
3. Gating that on `ST_Transform` still admitted a geography cast behind a derived-column alias: `WITH g AS (SELECT geom_4326::geography AS x FROM data.cities) SELECT ST_Segmentize(x, 0.01) FROM g`.

The third closes the door. The canonical buffer's own segmentize target is an alias, `_pb_d0.c0`, several derived levels from its input, so "prove the target's units" and "admit the canonical buffer" are contradictory under any target-inspection scheme. The same indirection defeats deciding whether a correlated source traces to a real relation, because the buffer's input legitimately does. That is data-flow analysis, and it does not belong in a hotfix for a red `main`.

## Why this is the stricter endpoint

#994 **added** `st_dump`, `st_dumpsegments`, `st_segmentize`, `generate_series` and `st_unaryunion` to the PostGIS allowlist to support the buffer, then added the cost guards to bound them. Removing both puts those amplification-prone functions back outside the fail-closed allowlist entirely. The sandbox here is stricter than `main` is today, not looser.

## Why #994's prompt change stays

This is the part I got wrong first time round, and review caught it.

A full revert restores the hand-written buffer teaching, and `st_buffer` is allowlisted independently of #994 — `test_allows_st_buffer` confirms `ST_Buffer(geom_4326::geography, 1000)::geometry` passes. So a full revert would instruct the model to emit the bare form **and it would execute**, returning the world-spanning or zone-distorted polygon that #883, #900 and #990 fixed. The "fails closed" claim I made for the full revert was simply false.

Keeping the prompt means the model emits the *safe* banded expression and the sandbox refuses it outright. A refused buffer question beats a silently wrong chat overlay and bbox.

So the user-visible effect is: buffer questions in chat now return an error instead of a wrong answer, until #1001 lands. Nothing else in NL→SQL changes.

## What is untouched

`render_geodesic_buffer` and the analysis-side buffer path that calls it directly are not affected. This is limited to the sandbox allowlist, the cost guards, and their tests.

`test_rendered_buffer_expression_*` is inverted to assert the refusal, with a docstring saying it must go back to asserting admission when #1001 lands, so the temporary state cannot quietly become permanent.

## The redo

#1001. The tractable design is matching the whole rendered template with the input as a single hole, rather than characterising the expression one function at a time. The canonical shapes are input-independent, measured across three different inputs, which is what makes it work:

```
st_collect:      3 shapes, input-independent
st_dumpsegments: 1 shape,  input-independent
st_segmentize:   2 shapes, input-independent
st_dump:         7 shapes, one embeds the buffer distance
```

`_canonical_band_series_shapes` already does this for `generate_series`, so the precedent is in the file. Two changes from #1002 are worth carrying over and are recorded on the issue: skipping `exp.Exists` in the callable walk (the #538 `exp.Connector` class, one node type over) and `st_dimension` in the allowlist.

## Verification

```
uv run pytest tests/test_sec025_sandbox_allowlist.py tests/test_sandbox.py tests/test_ai_sql_prompt_935.py -q   # 154 passed
uv run pytest -q -n 4                                                                                          # full suite
```

### DIFF
--- a/backend/app/platform/sandbox/validator.py
+++ b/backend/app/platform/sandbox/validator.py
@@ -171,14 +171,6 @@ _ALLOWED_FUNCTIONS: frozenset[str] = frozenset(
         "greatest",
         "least",
         "width_bucket",
-        # fix(#935): GENERATE_SERIES → sqlglot sql_name
-        # "exploding_generate_series". Named here so it clears the fail-closed
-        # allowlist, but _validate_function_cost only admits the EXACT series
-        # call the canonical geodesic buffer expression emits (bounded by a
-        # 4326 longitude span, at most ~61 terms); every other
-        # generate_series — literal, dynamic, or composed — stays rejected.
-        "exploding_generate_series",
-        "generate_series",  # Anonymous-node spelling on some sqlglot paths
         "pi",
         "degrees",
         "radians",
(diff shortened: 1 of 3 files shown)


ishiland:
@codex review

ishiland:
@codex review

Issue #923, opened 2026-07-30
fix(builder): style copy/paste merges paint and reintroduces the fill-color + fill-pattern collision


ishiland:
Close-out finding from the 2026-07-30 style-tab audit. Same collision class as #918, different entry point. Pointers re-verified on main @ 9a788ba92.

## What happens

Copy/paste style merges paint instead of enforcing the fill-color/fill-pattern mutual exclusion:

- Copy a solid-color polygon style, paste onto a patterned layer: the target's `fill-pattern` survives the spread, so the merged paint carries both keys, the pattern keeps winning on the map, and the paste looks like a no-op.
- Copy a patterned style, paste onto a solid layer: the target's `fill-color` survives alongside the copied `fill-pattern`. The map shows the pattern, but saved paint carries both keys.

Either direction violates the EDIT-05 invariant that saved paint never carries both (`frontend/src/components/builder/LayerStyleEditor.tsx:347-350`).

## Why

`applyCopiedStyleToLayer` (`frontend/src/lib/builder/layer-style-clipboard.ts:95-108`) builds `{ ...target.paint, ...deepClone(copied.paint) }` at `:99-102`. The merge is deliberate (target-only keys like tile params must survive), but for the fill family it needs the same exclusion the picker handler enforces.

## Fix

After the merge, apply the exclusion for polygon-class pastes:

- if `copied.paint` has `fill-pattern`, delete `fill-color` from the merged paint;
- else if `copied.paint` has `fill-color`, delete `fill-pattern` from the merged paint.

Pure function, so this is a couple of lines plus two unit cases in the clipboard helper's test file.

Please write the exclusion as a small **exported** helper in `layer-style-clipboard.ts` (something like `resolveFillCollision(mergedPaint, copiedPaint)`) rather than inline logic, so the rule has one home.

Good news on blast radius: `applyCopiedStyleToLayer` has three consumers, and fixing the pure function fixes all of them with no extra code. Each just gets its own acceptance check below.

| Consumer | Location |
| --- | --- |
| Single-layer paste | `frontend/src/components/builder/hooks/use-layer-style-clipboard.ts:64` |
| Bulk apply-style across a multi-selection | `frontend/src/components/builder/hooks/use-bulk-layer-actions.ts:177` |
| Bulk apply-style, per-target branch | `frontend/src/components/builder/hooks/use-bulk-layer-actions.ts:187` |

## Note: a sibling fix is landing in parallel

[...]

Issue #917, opened 2026-07-30
fix(maps): exported style.json passes builtin fill-pattern ids the served sprite does not contain


ishiland:
Found in the 2026-07-30 style-tab audit (pass 2). Backend counterpart of the builder fill-pattern work (#910/#911/#914). Pointers re-verified on main @ 9a788ba92.

## What happens

Save a map whose polygon layer uses a builtin fill pattern (`paint['fill-pattern'] = 'geolens-fill-hatch'`). The served `style.json` document passes that value through, and its `sprite` points at the map-icons sprite. The sprite index is built only from the `map_icons` table, so `geolens-fill-*` is not in it. Any external MapLibre client that consumes the exported style gets a missing-image warning and renders no fill at all for that layer.

In-app surfaces don't hit this only because the viewer and builder go through the client-side adapter registry, which generates and registers the pattern images in the browser (`frontend/src/components/builder/layer-adapters/fill-pattern-images.ts`). The exported document has no such luck.

## Where

- `backend/app/modules/catalog/maps/style_json.py:1053`: the fill allowlist includes `fill-pattern` and the value passes through verbatim, unlike `icon-image` which is rewritten via `_sprite_icon_id` (`:535`).
- Sprite declared on the document at `style_json.py:1405`.
- Sprite index built from `map_icons` only: `backend/app/modules/catalog/maps/sprites.py:590-603`.
- `grep -rn "geolens-fill" backend/` has zero hits.

## Fix options

1. Bake the five builtin patterns into the served sprite: port the five 16x16 generators from `fill-pattern-images.ts` (they're simple pixel loops, fixed `rgb(80,80,80)`, see `TILE = 16` at `:17`) into the sprite builder as static entries alongside the `map_icons` rows. Then the exported document is self-contained.
2. Or strip `fill-pattern` during style.json serialization and emit the fallback `fill-color` instead, documenting that patterns are an in-app feature.

Option 1 keeps exported maps faithful and is the better end state; option 2 is the two-line stopgap. Whichever lands, add a test asserting a patterned layer's exported style either resolves its sprite id or carries no `fill-pattern`.

Option 2 is implementable right now and independently of everything else in the fill-pattern cluster. It is backend-only, it touches one serializer, and it does not read or write anything the builder issues are changing. If the goal is to stop external clients rendering an empty polygon, take it today rather than sequencing behind the frontend work.

## Sequencing: option 1 depends on #914

[...]

Issue #914, opened 2026-07-30
ux(builder): fill patterns ignore the fill color and are nearly invisible at default opacity


ishiland:
Reproduced locally on main @ cb22f5bb1 (World Countries polygon layer, Positron basemap). Pointers re-verified on main @ 9a788ba92.

Scope note: this issue covers the tinting half only. The legend and layer-list swatch mismatch moved to #951, which shipped in #979.

## What happens

Selecting any fill pattern paints thin dark-gray lines (fixed `rgb(80,80,80)`, 1px in a 16px tile) at the layer's fill opacity, which defaults to 0.3 for fills. On a light basemap the result reads as "my fill disappeared"; on a dark basemap it's close to invisible. The picker swatches preview the patterns in the UI foreground color, which suggests the pattern will follow the chosen fill color. It never does.

## Why

- Pattern tiles are generated with a hardcoded pixel color (`makeHatch` and its four siblings in `frontend/src/components/builder/layer-adapters/fill-pattern-images.ts`, all `setPixel(..., 80, 80, 80, 255)` at `:34`, `:49`, `:63`, `:77`, `:90`). `makeFillPatternImage(id)` at `:106` takes no color argument.
- The default fill opacity is 0.3 (`MAP_COLORS.default.fillOpacity`, `frontend/src/lib/map-colors.ts:34`), so a 1-in-4-pixel gray line lands at 30% alpha.
- MapLibre ignores `fill-color` when `fill-pattern` is set, and the builder deletes `fill-color` anyway (EDIT-05, `LayerStyleEditor.tsx:351-366`).
- The CSS previews are drawn with `currentColor` (`patternPreviewStyle` in `frontend/src/lib/fill-pattern-preview.ts`), which is what creates the expectation that the pattern follows the fill color.

## Proposed fix (enhancement)

Tint patterns with the layer's fill color:

1. Parameterize the generators: `makeFillPatternImage(id, rgb)` instead of the hardcoded gray.
2. Register tinted variants lazily under namespaced ids, for example `geolens-fill-hatch#1d4ed8`, in `ensureFillPatternImages` (keep the plain ids registered as the gray fallback).
3. In the fill adapter (`fill-adapter.ts` `addLayers` at `:73` and `syncPaint` at `:170`), when `fill-pattern` is one of the builtin ids and a tint color is available, rewrite the paint value to the tinted id before `setPaintProperty`.
4. Tint source: the `fillColorSaved` stash from #910, falling back to gray when it is absent.

[...]

PR #1022, opened 2026-07-31
fix(builder): keep fill-color and fill-pattern mutually exclusive in both directions

### LINKED ISSUE #910: fix(builder): fill-pattern picker destroys data-driven and solid fill colors; gate it and stash the color
Reported on the demo (private map `038206f8` on demo.getgeolens.com, v1.6.1) and reproduced locally on main @ cb22f5bb1 with the "Manhattan - A Century of Skyline" showcase map. Pointers re-verified on main @ 9a788ba92.

Consolidates #911, which covered the solid-color half of the same handler. Both halves are one gating problem in one component and one handler, so they are fixed together here.

## What happens

Three variants of the same defect, all triggered from the Fill Pattern swatches.

**1. Data-driven layer, the destructive one.** The Buildings layer is categorical (fill color by `era`, 7 classes) and renders as 3D extrusion. Clicking any swatch in "Fill Pattern" instantly repaints every building in the uniform default blue. No pattern appears. Clicking the "None" swatch does not bring the era colors back; the layer stays default blue. Meanwhile the editor still says "Styled by: era", the Data-driven section still shows the 7 class colors, and the legend still lists the era classes. The only recovery is the pending-style banner's Revert. The same breakage applies to any data-driven polygon layer in plain Fill mode: the classification expression is destroyed and "None" restores the default color, not the expression.

**2. Solid-color round-trip (was #911).** Pick a custom fill color, then click a pattern swatch: the color is discarded. Click the "None" swatch: the fill comes back as the default blue, not the color you had.

**3. Dead color picker while a pattern is active (was #911).** The Color row still renders with a stale swatch and stays fully interactive. It does not merely do nothing. It snaps to the default blue, because `fill-color` was deleted and `getPaintValue` falls back to `FILL_DEFAULTS`, and clicking a new color writes `fill-color` back into paint next to `fill-pattern`. That re-breaks the EDIT-05 both-keys-never invariant with zero visible effect on the map. Hiding it is correctness, not polish.

## Why

`handleFillPatternChange` (`frontend/src/components/builder/LayerStyleEditor.tsx:351-366`, EDIT-05 comment at `:347-350`) enforces fill-color/fill-pattern mutual exclusion by deleting `paint['fill-color']`, and restores `FILL_DEFAULTS['fill-color']` on None. Nothing distinguishes a string color from an expression, and nothing stashes the previous value. For a data-driven layer that key holds the generated `match`/`step` expression, so the delete destroys the classification while `style_config` still claims it exists.

Two gating gaps expose this:

1. `FillEditor` hides the color picker for data-driven layers (`frontend/src/components/builder/LayerStyleEditor/FillEditor.tsx:59-69`) but still renders `FillPatternPicker` for any polygon (`FillEditor.tsx:76-82`), and never hides the color picker while a pattern is active.
2. The picker also renders in 3D-extrusion render mode. The extrusion adapter deliberately does not author `fill-extrusion-pattern` (SPEC-11 comment, `frontend/src/components/builder/layer-adapters/fill-adapter.ts:30-33`) and reads `rawPaint['fill-color']` for the extrusion color (`fill-adapter.ts:148` and `:211`), so a pattern click resets the whole extrusion to the default blue.

There is an existing convention for the stash half: the Fill visibility toggle stashes `fillOpacitySaved` in the builder style config and restores it on re-enable (`LayerStyleEditor.tsx:278-289`). Builder-config fields persist through save and reload via `style_config.builder`.

### Correction to the original report

An earlier version of this issue said patterns "can never show" in 3D-extrusion mode. That is wrong, and the fix should not be argued that way in review. `fillAdapter.addLayers` keeps the flat `fill` layer alongside the extrusion companion (`fill-adapter.ts:105-113` and `:144-163`), and `DEFAULT_EXTRUSION_MIN_ZOOM = 14` (`frontend/src/components/builder/layer-adapters/builder-defaults.ts:85`), so below zoom 14 the patterned flat fill is the only thing drawn. The real reason to hide the picker in extrusion mode is the color reset described above, which hits non-data-driven extrusion layers too.

### 1. Gating truth table

`isDataDriven` and `currentHeightCol` are already in `FillEditor`'s props (`FillEditor.tsx:33` and `:43`), so all four rows are decidable where the controls render.

| Layer state | Pattern picker | Color picker |
| --- | --- | --- |
| Plain solid polygon, Fill or Fill+Stroke | shown | shown |
| Data-driven (`style_config.column` set) | hidden, with hint | already hidden ("Styled by") |
| 3D extrusion (`currentHeightCol` set) | hidden | shown |
| `paint['fill-pattern']` set | shown (so it can be cleared) | hidden, with hint |

Use the same muted hint styling as the existing "Styled by" swap at `FillEditor.tsx:59-62`.

### 2. Stash the fill color across the pattern round-trip

In `handleFillPatternChange`, stash the current string `fill-color` as `fillColorSaved` in the builder config before deleting it, mirroring the `fillOpacitySaved` pattern. In the None branch, restore `fillColorSaved ?? FILL_DEFAULTS['fill-color']` and clear the stash. Only stash a string; if `fill-color` holds an expression there is nothing to save (and per the truth table the picker is not reachable in that state anyway).

### 3. New builder key: three tables, not one

`fillColorSaved` is a new builder key, and builder keys are stored snake_case with a three-table contract. Miss the backend entry and the key persists as camelCase among snake_case siblings. Miss the frontend alias and the reload path never produces `fillColorSaved`, so the save-and-reload acceptance case below fails silently with nothing in the console to explain it. All three edits are one line each:

| Table | File | Entry to add |
| --- | --- | --- |
| TS type | `frontend/src/types/api.ts:893-971` (`BuilderStyleConfig`, next to `fillOpacitySaved` at `:896`) | `fillColorSaved?: string;` |
| Load-side alias | `frontend/src/components/builder/layer-adapters/shared.ts:533-560` (`BUILDER_STYLE_KEY_ALIASES`, next to `fill_opacity_saved` at `:536`) | `fill_color_saved: 'fillColorSaved',` |
| Storage canonicalizer | `backend/app/modules/catalog/maps/schemas.py:79-106` (`_BUILDER_CAMEL_TO_SNAKE_KEYS`, next to `fillOpacitySaved` at `:82`) | `"fillColorSaved": "fill_color_saved",` |

The backend table is the authoritative one (see its comment at `schemas.py:71-78`); `style_json.py` derives the export direction from its inverse (`BUILDER_SNAKE_TO_CAMEL_KEYS`, `schemas.py:113-115`), so adding it there covers both directions.

### Three, not eight: why the other five sites do not apply

`grep -rn "fillOpacitySaved\|fill_opacity_saved\|_fill-opacity-saved" backend/app frontend/src` returns eight registration sites, and copying all eight is the obvious wrong move here. Five of them exist only to migrate the legacy underscore paint key `_fill-opacity-saved`, which predates the `style_config.builder` block and still turns up in maps saved by older clients:

- `frontend/src/components/builder/layer-adapters/shared.ts:11` (`CUSTOM_PAINT_PROPS`), which strips the underscore key so it never reaches MapLibre.
- `frontend/src/components/builder/LayerStyleEditor/builder-paint-map.ts:39` (`BUILDER_PAINT_FIELDS`), the builderKey/paintKey pairing.
- `frontend/src/lib/normalize-style-config.ts:92` (`PAINT_FALLBACK_FIELDS`), which reads the underscore key when the builder block has no value.
- `backend/app/modules/catalog/maps/schemas.py:35` (`LEGACY_BUILDER_PAINT_KEYS`), the underscore-to-snake rewrite on storage.
- `backend/app/processing/ai/schemas.py:22`, which allowlists the underscore key on the AI paint path.

`fillColorSaved` is new and has never had an underscore paint-key form, so there is no legacy value to migrate and nothing to strip. Do not add it to any of those five. Three tables is the whole contract for a builder-only key.

### 4. Translations

Two new keys under `style` in `frontend/src/i18n/locales/{en,es,fr,de}/builder.json`, next to `fillPattern` (`en/builder.json:249`). A `defaultValue` alone fails the `npm run test:i18n` locale-parity gate, so all four locales need real values. Suggested strings, worth a second look from anyone who reads the language:

```
en  fillPatternUnavailableDataDriven: "Patterns are unavailable while styling by an attribute"
    fillColorUnavailablePattern:      "Color is unavailable while a pattern is applied"
es  "Los patrones no están disponibles al estilizar por un atributo"
    "El color no está disponible mientras se aplica un patrón"
fr  "Les motifs sont indisponibles lors du stylisme par attribut"
    "La couleur est indisponible lorsqu'un motif est appliqué"
de  "Muster sind beim Stilisieren nach einem Attribut nicht verfügbar"
    "Farbe ist nicht verfügbar, solange ein Muster angewendet wird"
```

## Must land with #918

Hiding the pattern picker on data-driven layers removes the only control that can clear an existing `fill-pattern`. A layer that is patterned first and then classified would keep the pattern in paint, keep it winning on the map, and offer no way out. That is worse than today's bug, which the None swatch at least escapes.

So either land this together with #918 (which deletes `fill-pattern` when an expression takes ownership of the color), or, if this ships first, clear a stale `fill-pattern` from paint at the moment the picker is hidden. Do not ship the gate alone.

The dependency runs both ways: #918 step 2 clears the `fillColorSaved` stash this issue introduces, so whichever lands second picks up the other half.

## Acceptance

- Data-driven polygon layer: no pattern swatches shown, hint shown, classification cannot be destroyed from this section.
- Polygon layer in 3D-extrusion mode: no pattern swatches shown.
- Plain solid-color polygon layer in Fill or Fill+Stroke mode: picker unchanged.
- Solid color A, then Hatch, then None: the layer is color A again.
- Solid color A, then Hatch, then save, reload, then None: still color A. This is the case that fails if either alias table in step 3 is missed.
- With a pattern active there is no interactive color picker for the fill.
- A layer that already carries `fill-pattern` can still reach a state where the pattern is gone, whether by this issue's gate or by #918.
- `cd frontend && npm run lint && npm run typecheck && npx vitest run src/components/builder/LayerStyleEditor/__tests__/FillEditor.test.tsx src/components/builder/__tests__/LayerStyleEditor.test.tsx`

`FillEditor.test.tsx:448-463` already has the two tests to copy for the new hidden-state cases (`does NOT render Fill Pattern section when isPolygon=false` and the `fillEnabled=false` sibling). The EDIT-05 handler tests live at `LayerStyleEditor.test.tsx:1443-1504`.

Kept as a good first issue after the merge: it is still one component, one handler and one test file, and the three table edits above are quoted verbatim so there is no hunting involved. Budget a full day rather than an hour, though. Between the four-row gate, the stash round-trip, three registration tables, eight locale strings and the save-and-reload acceptance case, it is a wide first issue even though no single step is hard.

### LINKED ISSUE #918: fix(builder): applying a data-driven ramp leaves an active fill-pattern in paint, which overrides the ramp on the map
Found in the 2026-07-30 style-tab audit (pass 2). Complements #910: that issue hides the pattern picker on already-data-driven layers, this one covers the reverse order. Pointers re-verified on main @ 9a788ba92.

## What happens

Apply a fill pattern to a plain polygon layer, then apply a data-driven color ramp (categorical or graduated). The paint now carries BOTH `fill-color` (the expression) and `fill-pattern`. MapLibre gives the pattern precedence, so the map keeps drawing gray hatching, while the appearance section swaps to "Styled by: {column}" and the legend renders the ramp classes. Everything in the UI says classification worked; the map says otherwise. This also violates the EDIT-05 invariant that saved paint never carries both keys (`frontend/src/components/builder/LayerStyleEditor.tsx:347-350`).

## Why

The data-driven apply paths spread the previous paint and only set the color property:

- `frontend/src/components/builder/DataDrivenStyleEditor.tsx:468` (categorical) and `:532` (graduated) build `stripStaleLineGradient({ ...layer.paint, [colorProp]: expression }, colorProp)` and never remove `fill-pattern`.
- The clear paths at `:599-618` (`handleClear`), `:620-632` (`handleColumnChange`) and `:634-661` (`handleModeChange`) reset `fill-color` to the default but likewise leave `fill-pattern` in place, so clearing a ramp on a previously patterned layer leaves the pattern winning over a freshly defaulted color.

## Fix: the chokepoint, not the two call sites

~~One-line deletes at `DataDrivenStyleEditor.tsx:468` and `:532`.~~ Fixing it there works for the manual editor and misses everything else. Do it at the shared funnel instead.

`handleStyleConfigChange` (`frontend/src/components/builder/hooks/use-layer-map-sync.ts:353-423`) is the single handler every style write lands in, and it already implements the exact analogous exclusion for a different pair of incompatible keys: a data-driven solid color is incompatible with `line-gradient`, so the handler drops the stale paint key (`:367-370`), drops the matching `builder.lineGradient` intent stub (`:392-398`), and imperatively clears the property on the live map before the adapter repaint (`:405-419`). Its own comment at `:360-361` notes that it serves both the manual `DataDrivenStyleEditor` path and the AI `set_style_config` action.

`fill-color` versus `fill-pattern` is the same shape. Add it alongside, inline, using the `isDataDrivenColor` flag the handler already computes at `:362-365`. Write the rule directly in this handler rather than importing one: the line-gradient block a few lines above is the template, and matching it keeps the whole exclusion story readable in one place.

1. When `isDataDrivenColor` is true and the resolved color property is `fill-color`, delete `fill-pattern` from `effectivePaint`, in the same `if` shape as the `line-gradient` drop at `:367-370`.
2. Clear the `fillColorSaved` stash from the builder config in the same branch, mirroring how `builder.lineGradient` is dropped at `:392-398`. Once an expression owns `fill-color`, a saved solid color from before the ramp is stale, and restoring it later would resurrect a color the user set several edits ago. (That stash is introduced by #910; if #910 has not landed yet, leave a note and add the clear in whichever PR lands second.)
3. Imperatively clear `fill-pattern` on the live map in the map-side callback, next to the existing `setPaintProperty(mapLayerId, 'line-gradient', undefined)` at `:412`, so the repaint does not wait on a full adapter re-add.
4. Fix the clear paths in `DataDrivenStyleEditor` too (`:599-618`, `:620-632`, `:634-661`). Those write a non-data-driven config, so the chokepoint branch above does not fire for them and the pattern would survive a ramp clear.

### Relationship to #923

#923 fixes the same collision at the clipboard entry point and exports a `resolveFillCollision` helper from `frontend/src/lib/builder/layer-style-clipboard.ts`. This issue does not wait for it and does not import it. Land the inline rule here; once #923's helper exists, swapping this handler over to it is a mechanical follow-up that either PR can carry. The two are independent in both directions.

### The fourth writer this covers

The exclusion is currently enforced in exactly one of four places that write fill paint:

| Writer | Location | Enforces today? |
| --- | --- | --- |
| Pattern picker | `LayerStyleEditor.tsx:351-366` | yes, destructively (#910) |
| Data-driven apply | `DataDrivenStyleEditor.tsx:468,532` | no, this issue |
| Clipboard paste and bulk apply | `lib/builder/layer-style-clipboard.ts:99-102` | no, #923 |
| AI `set_style` action | `backend/app/processing/ai/schemas.py:26` | no |

The AI path allows `fill-pattern` through its paint allowlist at `backend/app/processing/ai/schemas.py:26` with no exclusion logic anywhere on either side of the wire, so an AI style action can produce both keys today. It lands in `handleStyleConfigChange` like everything else, which is why the chokepoint framing closes it at no extra cost. The two-line version at the call sites would leave it open.

## Acceptance

- Pattern then ramp: the map renders the ramp, paint carries no `fill-pattern`, and the pattern picker (visible again per #910 only when not data-driven) shows "None" active after clearing the ramp.
- Pattern then ramp then clear the ramp: paint carries neither `fill-pattern` nor a stale `fillColorSaved`, and the layer is a plain default solid.
- Ramp on a never-patterned layer: unchanged.
- An AI `set_style` action that sets a data-driven `fill-color` on a patterned layer ends with `fill-pattern` gone, same as the manual path.
- `cd frontend && npm run lint && npm run typecheck && npx vitest run src/components/builder/__tests__/DataDrivenStyleEditor.test.tsx src/components/builder/hooks/__tests__/use-layer-map-sync.test.ts`

`use-layer-map-sync.test.ts` already covers the line-gradient exclusion, so the tests for this one can be written directly against that precedent.

Kept as a good first issue: the chokepoint has a working template a few lines above the insertion point, and the change is additive to a branch that already exists.

ishiland:
Pairs #910 and #918. They are two halves of one paint-model problem and must land together: hiding the pattern picker on data-driven layers (#910) removes the only control that can clear an existing `fill-pattern`, and #918 is what deletes the pattern when an expression takes ownership of the color. #918 also clears the `fillColorSaved` stash that #910 introduces.

## What was broken

`fill-color` and `fill-pattern` are mutually exclusive (EDIT-05), and the rule was enforced in exactly one of the four places that write fill paint — destructively.

- **Pattern applied to a data-driven layer** (#910): `handleFillPatternChange` deleted `paint['fill-color']` without distinguishing a string from an expression, so a categorical/graduated classification was destroyed. The editor kept saying "Styled by: era", the legend kept listing the classes, and "None" restored the default blue rather than the expression. In 3D-extrusion mode the same delete reset the extrusion, which reads `rawPaint['fill-color']`.
- **Solid color round-trip** (#910): color A → Hatch → None came back as the default blue, and the color picker stayed live behind an active pattern, writing `fill-color` back in next to `fill-pattern` with no visible effect on the map.
- **Ramp applied over a pattern** (#918): paint carried both keys. MapLibre gives the pattern precedence, so the map drew hatching while the appearance section, the legend and saved paint all reported the ramp had applied.

## What changed

**#918 — the chokepoint, not the call sites.** The rule goes in `handleStyleConfigChange` (`use-layer-map-sync.ts`), the funnel every style write lands in, right alongside the `line-gradient` exclusion it already implements for the identical reason. One insertion covers the manual editor, the clipboard/bulk-apply path and the AI `set_style` action (whose paint allowlist admits `fill-pattern` with no exclusion logic on either side of the wire). The branch drops the pattern from paint, drops the now-stale `fillColorSaved` stash, and clears `fill-pattern` imperatively on the live map so the ramp shows without waiting on a full adapter re-add.

The three clear paths in `DataDrivenStyleEditor` (`handleClear`, `handleColumnChange`, `handleModeChange`) write a non-data-driven config, so that branch does not fire for them — they drop the pattern themselves, which also rescues layers already saved with both keys by today's bug.

**#910 — gate it instead of letting it destroy state.**

| Layer state | Pattern picker | Color picker |
| --- | --- | --- |
| Plain solid polygon | shown | shown |
| Data-driven | hidden, with hint | already hidden ("Styled by") |
| 3D extrusion | hidden | shown |
| `fill-pattern` set | shown (so it can be cleared) | hidden, with hint |

The solid color round-trips through a new `fillColorSaved` stash in `style_config.builder`, mirroring the `fillOpacitySaved` convention next to it. Only a string is stashed; an expression means the layer is data-driven, where the picker is now unreachable and #918 owns the pattern instead.

## Schema / API impact

No migration, no OpenAPI or SDK change. `fillColorSaved` is a builder-only style-config key, so it registers in three tables — the TS type (`types/api.ts`), the load-side snake_case alias (`layer-adapters/shared.ts`), and the authoritative backend canonicalizer (`maps/schemas.py`, whose derived inverse drives style export). `backend/tests/test_builder_alias_keys_drift.py` pins the frontend/backend parity of those maps in CI. It has never had a legacy underscore paint-key form, so the five `_fill-opacity-saved` migration sites deliberately do not get an entry.

Untouched on purpose: `frontend/src/lib/builder/layer-style-clipboard.ts` (#923, external contributor). Once that PR lands, swapping this handler over to its `resolveFillCollision` helper is a mechanical follow-up.

## Verification

```
# 366 files, 4292 tests pass

npx vitest run src/components/builder/LayerStyleEditor/__tests__/FillEditor.test.tsx \
  src/components/builder/__tests__/LayerStyleEditor.test.tsx \
  src/components/builder/hooks/__tests__/use-layer-map-sync.test.ts \
  src/components/builder/__tests__/DataDrivenStyleEditor.test.tsx

cd backend && uv run pytest tests/test_builder_alias_keys_drift.py tests/test_maps_style_json.py -q
# 69 passed
```

New tests: the four gate rows in `FillEditor.test.tsx`; the stash round-trip including the save-and-reload case that fails if either alias table is missed; the `#918` exclusion in `use-layer-map-sync.test.ts` (drop + stash clear + imperative map clear, with negative controls for a non-data-driven write and a size-target change); and the ramp-clear case in `DataDrivenStyleEditor.test.tsx`.

Live QA screenshots to follow on this PR.

Closes #910
Closes #918


ishiland:
@codex review

ishiland:
Both P2s fixed and verified live.

**1. Pattern-to-pattern dropped the stash.** Right: Hatch then Dots found no `fill-color` left to stash, and the `undefined` cleared the existing `fillColorSaved` through `compactBuilder`, so the eventual None came back default blue. The stash now seeds from its current value and is cleared only on the way to None.

**2. Extrusion stranded an active pattern.** Also right, and it was the worse of the two — with both controls hidden there was no way to remove the pattern from that panel at all, which is exactly the trap the issue warns about for shipping the gate alone. Two changes:

- The picker stays reachable in extrusion mode *when a pattern is set*. It is still hidden when there is none, so a pattern still cannot be applied to an extrusion layer; only cleared.
- `fill-adapter` colors the extrusion companion from `builder.fillColorSaved` when `fill-color` is absent. The companion authors no pattern of its own (SPEC-11) and was silently reverting to the default blue. A data-driven `fill-color` expression still wins over the stash, so the categorical extrusion case is unchanged — pinned by a second test.

## Live QA

Worktree Vite on `:5174` against the dev stack, throwaway map over "New York Median Household Income by County" (62 MULTIPOLYGON), state read back from `GET /api/maps/{id}` after each save.

```
[gate]        fill Color row hidden while patterned, hint shown
[save]        paint {"fill-opacity":0.3,"fill-pattern":"geolens-fill-hatch"}   (no fill-color)
              builder {...,"fillColorSaved":"#ff0000"}
[reload+None] fill colour #ff0000                        <- the alias-table case
[918]         pattern picker hidden, data-driven hint shown
[918/save]    paint keys ["fill-color","fill-opacity"]   fill-pattern gone
              fill-color is an expression: true
              builder {"outline_color":"#1d4ed8","outline_width":1}   fillColorSaved cleared
[clear]       paint {"fill-color":"#3b6fd4","fill-opacity":0.3}, picker back
[extrusion]   pattern picker hidden, fill colour still editable
console errors: 0
```

The two review cases, on a fresh map:

```
[P2-1] after Hatch->Dots  paint {"fill-opacity":0.3,"fill-pattern":"geolens-fill-dots"}
       stash survived     {...,"fillColorSaved":"#ff0000"}
       colour after None  #ff0000
[P2-2] pattern picker reachable in extrusion: yes, None reachable: yes
       colour after clearing from extrusion: #ff0000
       picker hidden again once no pattern: yes
console errors: 0
```

Visually confirmed: the ramp-over-pattern case draws the ramp with no hatching anywhere on the map (before this change the map kept drawing gray hatch while the legend listed the classes), and Hatch → None returns the red fill rather than the default blue.

One incidental finding worth recording, since it is direct evidence the third registration table matters: the dev stack runs `main`'s backend, so `fillColorSaved` persisted camelCase next to its `outline_color` / `outline_width` siblings. With the `schemas.py` entry in this PR it canonicalizes to `fill_color_saved`. The frontend round-trips either way, which is why the reload case passes above — the casing is what drifts, and `test_builder_alias_keys_drift.py` is what pins it.

ishiland:
@codex review

ishiland:
@codex review

ishiland:
Right on both counts, and the test that was supposed to cover it passed vacuously.

`withBuilderConfig` returns `null` when `fillColorSaved` was the builder's only field, and `use-layer-map-sync.ts:389-391` reads a null config as "keep the existing builder" — so the None click emitted a clear that the funnel immediately undid. The assertion I wrote for that case (`config?.builder?.fillColorSaved` is undefined) was true because `config` itself was null, so it never touched the behaviour.

Fixed by passing `replace` from this one call site (the config it builds is already complete, so verbatim restore is both the fix and the accurate signal). `updateBuilderConfig`'s other callers keep their three-arg shape. The test now pins the null **and** the flag, and `use-layer-map-sync.test.ts` pins the funnel in both directions: replace wipes an emptied builder, and a plain null config still preserves it, which is the revert contract #461 added.

Worth recording: the same collapse-to-null hazard sits on the fill and stroke visibility toggles, whose re-enable path writes `{ fillDisabled: undefined, fillOpacitySaved: undefined }`. On a layer whose `style_config` holds nothing but that builder, the funnel would restore `fillDisabled: true`. It predates this PR and I have left it alone rather than widen the diff; it wants its own issue.

Gates re-run at `ae84f8224`: lint, typecheck, i18n, 4298 frontend tests, and the backend alias-parity suite.

ishiland:
@codex review

ishiland:
Both are real, and both are paths that reach the pattern state without going through the picker.

**Expression-valued `fill-pattern`.** Right — `AdvancedJsonEditor` accepts spec-valid paint and `fill-pattern` takes expressions, so a string-only predicate read that layer as unpatterned: the colour picker stayed live and editing it wrote both keys back. The gate now counts any present value. The picker's `value` still requires a string, so an expression shows None active and remains clearable rather than pretending a swatch is selected.

**Reset leaving the stash.** Also right, and I had drawn the boundary in the wrong place — I noted the collapse-to-null hazard for the fill/stroke toggles and did not look at Reset, which passes a plain `null` config through the same branch. Reset now names the config it wants and replaces it: same content the null branch preserved, minus the stash. Two existing Reset tests move from a three-arg assertion to a four-arg one, so the `replace` signal is pinned where Reset is already covered, plus a new case for a layer whose builder has other fields (they survive; the stash does not).

Gates at `99de06bf0`: lint, typecheck, i18n, 4313 frontend tests, backend alias-parity suite.

For the record on the remaining sibling: the fill and stroke visibility toggles still write `{ fillDisabled: undefined, fillOpacitySaved: undefined }` through the non-replace path, so on a layer whose `style_config` holds nothing but that builder, the funnel restores the disabled state. It predates this PR and is not a fill-pattern bug; I would rather file it than keep widening this diff.

ishiland:
@codex review

ishiland:
That is a real regression and it is mine — thank you for catching it.

Reset must destroy the classification and the render mode. Building the config with `withBuilderConfig(layer.style_config, …)` copied all of it, and `replace` then persisted `mode` / `column` / `categories` / `render_mode` verbatim, so a classified layer came back with default paint and a live "Styled by". The funnel's null branch had been dropping those all along, and my change quietly stopped it.

Fixed by building the reset config **builder-only** — `compactBuilder({ ...builderConfig, fillColorSaved: undefined })` — which reproduces exactly what the null branch preserved, minus the stash, and nothing else.

Fail-before is proved rather than asserted. Reverting just that expression and running the new test:

```
× drops mode, column and render_mode on Reset
  AssertionError: expected 'categorical' to be undefined
```

and it passes on the shipped version. The test covers mode, column, categories and render_mode being gone while `builder.outlineWidth` survives and `builder.fillColorSaved` does not.

Gates at `3287d0c43`: lint, typecheck, i18n, 4306 frontend tests.

ishiland:
@codex review

ishiland:
Agreed, and the deletion alone was the wrong shape. Switching Mode is an exploratory act — nothing about it should repaint the layer a colour it never had, and losing the red into the stash while showing default blue is the worst of both.

Both of your options had a catch, so the fix takes the second one properly:

- *Delay the removal until a column takes ownership* would reopen the case #918 exists for. A layer saved with both keys (which today's bug produces) clears its ramp, gets a defaulted `fill-color`, and the surviving pattern keeps winning — "reads as cleared, still renders the pattern".
- *Restore and clear the saved colour* holds for both. The three clear paths now go through one helper: drop the pattern, put the **stashed colour** in as the solid fill, and drop the stash in the same write. Red hatching → Mode switch → red solid, and the pattern picker is right there to re-apply it.

Clearing the stash from these paths needs `replace` (they write a null config, which the funnel reads as "keep the existing builder"), so `DataDrivenStyleEditor`'s `onStyleConfigChange` prop gains the same optional `opts` the funnel already accepts.

New tests: the mode-switch case asserts `fill-color` comes back as `#ff0000` with the pattern gone, `replace` set, `builder.fillColorSaved` gone and `builder.outlineWidth` kept; a second covers a patterned layer with no stash falling back to the default colour and emitting a null config. One existing assertion gains the fourth argument.

Gates at `139d9fccc`: lint, typecheck, i18n, 4308 frontend tests.

ishiland:
@codex review

ishiland:
Confirmed at both call sites — `handleAdvancedPaintChange` falls through to `onPaintChange` for every non-symbol render mode, and the chat action's paint branch calls it directly. So the pattern can arrive on a classified layer without ever passing the funnel, and my gate then hid the only control able to remove it. Same stranding shape as the extrusion case, different entrance.

Took your first option, because an active pattern deserves its clearing control no matter how it got there: the picker now shows whenever `fill-pattern` is set, and applying a **new** pattern stays gated to plain solid polygons.

That exposed a second edge which had to go with it. Making the picker reachable on a classified layer means `handleFillPatternChange` can now run there — and its unconditional `delete paint['fill-color']` is the #910 defect itself. The delete is now conditional on the value being a string: only a solid colour is stashable and only a solid colour is removed. On a classified layer the ramp survives and None simply drops the pattern, which is the one useful action there.

I did not take the second option (enforcing the exclusion in the general paint funnel). It would mean silently deleting a key a power user typed into Advanced JSON, and it is a behavioural change to a shared write path that deserves its own issue and its own live QA rather than a sixth-round edit here.

New tests: the picker renders for a data-driven layer that already carries a pattern, and applying a pattern to a classified layer leaves the expression untouched with nothing stashed.

Gates at `ee81e5aef`: lint, typecheck, i18n, 4310 frontend tests.

ishiland:
@codex review

ishiland:
Confirmed by tracing it: `handleFillPatternChange` keeps the data-driven config, so `isDataDrivenColor` is true in the funnel, `fill-color` (the expression) and `fill-pattern` are both present, and the exclusion strips the pattern the user just clicked. A dead control — my previous round made the swatches visible without making them work.

The picker takes a `clearOnly` mode for data-driven and extrusion layers: None only, with a label that says so. Applying a pattern there is impossible by EDIT-05 and the funnel is right to refuse it; the honest UI is to not offer the swatch.

Knock-on I want to flag rather than bury: the no-delete-on-an-expression guard from last round is now unreachable through the picker on a classified layer, so its test would have been passing vacuously. It drives the path that IS still reachable instead — Advanced JSON can write an expression into `fill-color` with no `style_config` mode, and the full picker is live there. Same guard, real route.

One new locale key in all four (`style.fillPatternClearOnly`).

Gates at `1feff74ee`: lint, typecheck, i18n parity, 4310 frontend tests.

ishiland:
@codex review

ishiland:
Real, and my own test at `FillEditor.test.tsx:564` was asserting the bug — it pinned `aria-pressed="true"` on None for an expression-valued pattern, which is exactly backwards from what the map draws.

Fixed with a display-only sentinel rather than a second state prop: `EXPRESSION_PATTERN` is exported from `FillPatternPicker` and passed as `value` when `fill-pattern` is present but not a string. Nothing in the swatch grid equals it, so None reads `aria-pressed="false"` and no swatch reads pressed — the honest rendering of "a pattern is active and none of these five is it". It is never a pattern id: `onChange` still only ever emits `undefined` or a real `FILL_PATTERN_IDS` entry, so it cannot reach paint.

The test now asserts both buttons unpressed; it fails on the pre-fix component (`aria-pressed="true"` vs expected `"false"`).

Separately, that same round-7 commit shipped a `tsc` break I caught running the gates here: `queryByRole('button', { name: 'Hatch', exact: true })` — `exact` is not a `ByRoleOptions` key. Testing Library's string `name` match is already exact (unlike Playwright's, which is why I reached for it), so dropping the option keeps the assertion's meaning.

Gates at `65ba6c988`: lint, typecheck, i18n parity, 366 files / 4310 tests.

ishiland:
@codex review

ishiland:
Confirmed and fixed — this one crossed a boundary I had not swept. `fill-adapter.ts` learned to read `fillColorSaved` for the extrusion companion, and `_fill_companion_layers` is the export twin of exactly that code path; I fixed the live renderer and left the exported style.json still resolving `paint.get("fill-color", DEFAULT_FILL_COLOR)` against paint that EDIT-05 guarantees no longer has the key.

```python
fill_color = paint.get("fill-color")
if fill_color is None:
    fill_color = builder.get("fillColorSaved")
if fill_color is None:
    fill_color = DEFAULT_FILL_COLOR
```

`is None` rather than `or`, deliberately: `fill-extrusion-color` accepts a data-driven expression, and truthiness testing a list would be fine today but the explicit check documents that an expression must pass through. That mirrors the `??` chain on the frontend side.

Two tests, both in `test_maps_style_json.py`. The stash one fails before the fix with `assert '#3b82f6' == '#94a3b8'` — the default-blue export you described. The second pins the expression case in the same function, since that is the branch the `is None` check exists to protect.

`style_json.py` grew 9 lines, so its `test_layering.py` size cap moves 1432 → 1441.

Gates at `720c97bc8`: ruff check + format clean, `tests/test_maps_style_json.py` and `tests/test_layering.py` 104 passed; frontend lint, typecheck, i18n parity, 366 files / 4310 tests.

ishiland:
@codex review

ishiland:
Rebased onto main (`368114331c`) to clear the conflict. One collision, in `backend/tests/test_layering.py`: #1052 took the `maps/schemas.py` LOC cap 1312 → 1316 (+4 for `thumbnail_updated_at`) while this branch took it 1312 → 1313 (+1 for the `fillColorSaved` alias entry). Both comments kept, cap resolved to **1317**.

That is not arithmetic on the deltas — the ratchet is exact-match (`test_module_loc_caps_have_no_headroom`), so I checked the file: `schemas.py` is 1317 lines at the merge, and `test_layering.py` passes. No other file conflicted; my `style_json.py` bump (1432 → 1441) was untouched by main.

The diff is otherwise unchanged from the head you last reviewed (`720c97bc8`), so this is a merge commit plus one integer.

Gates at `0e8074187`: ruff check + format, `test_layering.py` / `test_maps_style_json.py` / `test_builder_alias_keys_drift.py` 105 passed; frontend typecheck, i18n parity, 366 files / 4331 tests.

Not arming auto-merge — merges are being serialized centrally.

ishiland:
@codex review

ishiland:
Real, and it is the gap I left when I scoped the #918 exclusion to `isDataDrivenColor`. Traced it: `handlePasteStyle` calls `applyCopiedStyleToLayer`, which spreads the copied paint over the target's, so the target's `fill-color` survives next to the copied `fill-pattern`; the config it forwards is builder-only, `isDataDrivenColor` is false, and the drop never runs.

Fixed at the chokepoint rather than in `applyCopiedStyleToLayer` — that file is #923's.

```ts
const patternOwnsFill =
  !isDataDrivenColor && 'fill-color' in effectivePaint && 'fill-pattern' in effectivePaint;
```

The pattern wins, because MapLibre draws it either way, so the stray colour comes out of the paint object and off the live map (the imperative `setPaintProperty(…, 'fill-color', undefined)` mirrors the sibling pattern clear — without it a later None click flashes the target's old colour before the repaint).

On the stash, your point about which colour None restores decided the precedence: **an incoming `fillColorSaved` wins over the value I just dropped.** On a paste that stash is the SOURCE layer's colour, which is what the user copied. The dropped colour is only used when nothing was stashed, so the fallback cannot overwrite the copied intent. Strings only — the stash feeds the extrusion companion and #914's tint resolver, both of which need a solid colour, so an expression is dropped without being stashed.

Three tests, all failing on the previous head:

- the paste shape (pattern + target colour + copied stash) drops the colour, keeps the pattern, and keeps `fillColorSaved: '#ff0000'` — not the target's `#0000ff`,
- with no incoming stash, the dropped colour is stashed and unrelated builder fields survive,
- an expression-valued `fill-color` is dropped but never stashed.

Gates at `67513a3ca`: lint, typecheck, 366 files / 4334 tests.

ishiland:
@codex review

ishiland:
Real, and a third bypass of the same rule. Verified `handleBulkApplyStyle` at `use-bulk-layer-actions.ts:176-188`: it calls `applyCopiedStyleToLayer` directly for the atomic `setLocalLayers` and again in the map-sync loop, so neither the #918 data-driven branch nor the paste branch ran. Both directions you name reproduce.

Routing bulk through the funnel per target was not an option — that handler's contract is one atomic write ("never field-by-field per layer"), and N funnel calls would be N state writes. So I did what you asked and shared the normalization instead: `resolveFillExclusions`, `stashExcludedFillColor` and `clearExcludedPaintOnMap` are now exported from `use-layer-map-sync.ts` (where the bulk hook already sources `applyLayerVisibilityToMap`/`applyLayerOpacityToMap`), and the bulk handler applies all three before both its state and map writes.

`builder.lineGradient` deliberately stayed inline in the funnel: dropping the gradient paint without also dropping the intent stub lets map-sync's `lineGradientNeededFor()` put it straight back.

The extraction is behavior-preserving for the funnel — the map-clear flattened out of its nested `if (isDataDrivenColor)` because `dropsFillPattern` already implied it, so the same flags reach the same clears. The 350 existing builder-hook tests pass unchanged, which is the evidence for that claim.

Tests, both failing before the wiring (`'fill-pattern' in paint` / `'fill-color' in paint` still true):
- bulk ramp-over-pattern drops the pattern and the now-stale `fillColorSaved`
- bulk pattern-over-colour drops the colour and keeps the SOURCE stash, not the target's displaced blue
- four direct `clearExcludedPaintOnMap` cases (keys per flag, the wrong-geometry throw, layer absent) — the bulk hook test harness reports `isStyleLoaded() === false`, so its map write is unreachable from there and the helper is covered directly instead

Gates at `14768e461`: typecheck, lint, 366 files / 4340 tests.

ishiland:
@codex review

ishiland:
Both real, both fixed — though the second one's consequence is narrower than stated, and it changed how I framed the fix.

**Expression fill-color.** Right, and worse than the reproduction you gave. It doesn't need a pattern click at all: an Advanced-JSON layer already carrying both keys re-sends that paint through `handleStyleConfigChange` on *any* later builder edit, so changing the outline width silently destroyed the classification. And you're right that it contradicted `LayerStyleEditor.test.tsx:1625` (`never deletes an expression-valued fill-color when a pattern is applied`) — that test asserts what the editor *emits*, so it passed while the funnel deleted the value one layer downstream. My own test was pinning the bug: it asserted the expression is dropped-but-not-stashed.

Narrowed the branch to a string-valued `fill-color`, which is the rule the PR already claimed ("only a solid colour is stashable, and only a solid colour is safe to remove") and makes `strandedFillColor` a `string | undefined` by construction. Both keys then persist for that case with the pattern winning on the map — the pre-existing Advanced-JSON quirk, and far cheaper than unrecoverable data loss. I did not gate the picker instead: the picker is deliberately live whenever a pattern is active (#919/#951), so gating it would strand an existing pattern with no way to clear it.

**Non-string stash.** The fix is right, the predicted failure is not: an invalid document can't reach consumers, because `_strip_wrong_typed_values` already pops any non-string `*-color` before the style ships. I checked rather than assumed — with `fillColorSaved = 42` the emitted companion has `['fill-extrusion-base', 'fill-extrusion-height', 'fill-extrusion-opacity', 'fill-extrusion-vertical-gradient']` and no colour key at all. So the real damage is that being *stripped* is worse than being resolved: MapLibre falls back to the spec default black instead of the brand blue this fallback exists to supply, and every export logs a wrong-typed-value warning for a layer nobody mistyped by hand. Resolved at the source with `isinstance(str)`, and the comment says that instead of the invalid-document claim.

Tests, each failing before its fix:
- `keeps an expression-valued fill-color rather than dropping what it cannot stash` — was `expected undefined to deeply equal [ 'match', … ]`
- `test_extrusion_companion_ignores_a_non_string_fill_color_stash` over 42 / dict / list / bool / None — was `KeyError: 'fill-extrusion-color'`, which is how I found the stripping in the first place

`style_json.py` LOC cap 1441 → 1451 (ruff reflowed the new comment).

Gates at `33f5a7122`: typecheck, lint, 366 files / 4340 frontend tests; `test_maps_style_json.py` + `test_layering.py` 105 passed; ruff check and format clean.

ishiland:
@codex review

ishiland:
Both real, and the first one is a bug I introduced in round 12's fix.

**Solid-over-pattern.** Confirmed: the pattern-wins tie-break is only correct when the pattern is what was *copied*. A copied solid style merges its `fill-color` beside the target's surviving `fill-pattern`, `isDataDrivenColor` is false, so the resolver deleted the colour the user had just copied and left the old pattern drawing — the paste appeared to do nothing.

Took your first remedy rather than the second: the funnel genuinely cannot answer this, because provenance is gone by the time `applyCopiedStyleToLayer` has merged. But both clipboard call sites still hold the source, so `applySourceFillPrecedence(mergedPaint, sourcePaint)` runs there — in `handlePasteStyle` and in the bulk helper — and drops the target's leftover key before the resolver sees anything. The resolver's tie-break then applies only to the Advanced-JSON case where both keys genuinely pre-exist, which is the one place it's right. `layer-style-clipboard.ts` itself is untouched (#923).

One deliberate omission, commented in place: the target's now-inert `fillColorSaved` is left alone. It is read only while a pattern is active, and picking a pattern overwrites it with the current colour first, so a stale value cannot resurface — and clearing it would hit the collapse-to-null hazard for no gain.

**Non-string stash in `fill-adapter.ts`.** Right, and the consequence is worse here than in the backend twin I fixed last round: the backend has `_strip_wrong_typed_values` to pop it, but MapLibre rejects the colour outright and `addLayers`' catch swallows the failure, so the whole 3D companion goes missing. Both call sites now share `resolveExtrusionFillColor` — string-or-array from paint (an expression is valid and must pass through), string only from the stash. It also removes two `as string | undefined` assertions that were being made on values nobody had checked; tsc rejected the honest signature until I typed it as `FillExtrusionColor`, which is the proof those casts were load-bearing fiction.

Tests, each failing before its fix:
- copied solid displaces the target pattern, via bulk AND via paste — both were `expected true to be false` on `'fill-pattern' in paint`
- `fill-extrusion` falls back to a real colour for a stash of `42` / `{r:1}` / `['#fff']` / `true` — was `expected 'number'/'object'/'boolean' to be 'string'`, i.e. junk reaching MapLibre

Gates at `12e1f2afd`: typecheck, lint, 366 files / 4346 tests. Worth noting the new adapter test passed under vitest while failing `tsc` (the `it.each` junk union is not comparable to the stash's declared type) — caught locally, fixed by widening before the cast.

ishiland:
@codex review

ishiland:
Both real, and read together they say the rule was in the wrong place. So I stopped patching paths and moved it.

**The rule, decided once:** whichever of `fill-color`/`fill-pattern` the write *introduced* is what the user asked for, so the other goes. Provenance comes from diffing the committed paint against the layer's previous paint — no caller declares its intent, so a path nobody enumerated cannot escape it. When nothing was introduced (an Advanced-JSON layer that already carried the pair) it falls back to pattern-wins, displacing a solid colour to the stash and leaving an expression alone.

**Where:** `applyLayerUpdate`, the single boundary every handler in the hook commits through. Two opt-outs — a reference-equal paint means the write never touched paint at all (visibility, opacity, layout, popup), and `restore` skips normalization outright.

**On your first finding, one correction that changed the fix.** `replace` is not a verbatim marker, so skipping normalization when it is set would have been wrong: of the 8 style callers passing `replace`, only `handleRevertToSaved` is a restore. The other 7 are forward edits — Reset, the pattern picker, the three data-driven clears — and they *need* normalizing. (The two other `replace: true` hits in the tree are react-router `setSearchParams`.) So the restore gets its own explicit flag and `replace` keeps meaning what it meant.

**Your second finding is fixed as a consequence rather than as a patch.** `onPaintChange` commits through the same boundary, so the AI `set_style` colour-over-pattern write now normalizes with no change at that call site. `handlePaintChange`'s map sync also had to stop feeding the adapter its raw `newPaint`, which would have repainted the key the commit just dropped.

**This subsumes rounds 12-14 and deletes code.** `applySourceFillPrecedence` is gone — the paint diff already knows what the clipboard introduced — so both clipboard sites are back to a plain merge, and the bulk helper just passes the target's paint as the baseline. Net for the round: 231 insertions, 74 deletions across 7 files, with the funnel shorter than when the round started.

Five tests at the boundary, verified against two separate mutations. Disabling boundary normalization fails three (AI adds colour → pattern displaced; AI adds pattern → colour displaced and stashed; a non-restore `replace` still normalizes) while the two skip-cases correctly still pass. Removing just the `verbatim` opt-out fails the restore test alone: `expected { Object (fill-pattern) } to deeply equal { 'fill-color': '#0000ff', … }` — the baseline coming back stripped, which is the dirty-forever bug you described.

One existing test changed and it is the honest signal: the revert assertion now expects `{ replace: true, restore: true }`.

Gates at `77aaa308d`: typecheck, lint, 366 files / 4352 tests.

ishiland:
@codex review

ishiland:
Real, and fallout from moving the rule to the boundary — thanks for catching the seam I created.

Your diagnosis is exact: a colour can now win by simply being *introduced*, but the stash was still cleared only for a data-driven colour. So a paint-only solid write dropped the pattern and left the old stash, the next pattern write found the slot occupied (incoming-stash-wins declines to overwrite), and None restored a colour from two edits earlier — extrusion companion included.

I took the invariant one step further than "clear it whenever `dropsFillPattern`", because that is still a list of reasons and a list of reasons is what just failed. **The stash means something only while a pattern owns the fill**, since that is precisely what None brings back. So it is now keyed off the RESOLVED paint: no `fill-pattern` in the committed paint, no stash. That subsumes the data-driven case it replaces, covers your solid-colour case, and cannot be outgrown by some future third reason a colour might win.

Three tests, on the sequence rather than the single write, since the single write was never the broken part:
- a solid colour taking the fill clears the stale stash
- so a later pattern write stashes `#222222` (the current colour) rather than `#111111` — this was `expected '#111111' to be '#222222'`, your symptom exactly
- the stash SURVIVES a write that leaves the pattern owning the fill (a `fill-opacity` change), which is the case the fix must not overreach into — it passes both before and after

Gates at `fee8decc0`: typecheck, lint, 366 files / 4355 tests.

Separately, worth recording because it only appeared in composition: pre-flighting #914 (tinted fill patterns) on top of this branch made my own round-14 junk-stash tests fail with `expected undefined to be defined`. `fillPatternTint` read the same unvalidated stash and handed junk to `ensureTintedFillPatternImage`, whose throw is swallowed by `fillAdapter.addLayers`' catch — so the whole layer silently failed to build, not just the tint. Fixed in #914 with the same string check plus direct unit tests. Neither branch's own suite could see it.

ishiland:
@codex review

ishiland:
Real. Presence is not intent, and my diff only checked presence.

On a layer that already carried both keys, editing `fill-color` from `#111111` to `#222222` left both "added" checks false, so it fell to pattern-wins and deleted the colour the user had just set. Reachable from Advanced JSON, the AI `set_style` action, and a bulk apply of a solid style onto such a layer.

Took the value-comparison option. An absent key compares unequal to any value, so "touched" **subsumes** "added" instead of sitting beside it as a second condition — the narrower check was the entire bug, and adding a parallel condition would have left the same shape in place to fail again. That is also why I did not take the propagate-the-intent option: a flag each caller passes is only as good as the callers that remember, which is what moving this rule to the commit boundary was meant to stop relying on.

Touching both keys at once stays genuinely ambiguous — nothing can read intent out of it — so it keeps the documented pattern-wins fallback, which is what MapLibre draws regardless. Now pinned by a test so the fallback reads as a decision rather than an accident.

Three tests, and only the first one moves:
- editing the colour keeps `#222222` and drops the pattern — was `expected undefined to be '#222222'` under presence-only
- editing the pattern keeps the pattern and stashes `#111111` — passes either way
- touching both takes the fallback — passes either way

The two unchanged tests are the point of including them: they show the fix narrowed nothing that already worked.

Gates at `090fad995`: typecheck, lint, 366 files / 4358 tests.

ishiland:
@codex review

ishiland:
Real, and it marks the limit of how I'd framed the rule. I keyed the stash off a *collision*, but the displacement does not need one: a wholesale Advanced JSON replacement (or `replace_paint`) hands over a pattern-only object that has already dropped `fill-color`, so nothing collided, nothing was recorded, and None fell back to default blue with the extrusion companion following it.

What actually matters is the **transition to a pattern-owned fill**. When the resolved paint is pattern-only and the incoming write no longer carries a colour, the displaced value now comes from the PREVIOUS paint. Strings only, as everywhere that touches this stash.

I also removed `patternOwnsFill` from the stash-write condition while I was in there. `strandedFillColor` is set only where a pattern took the fill, so requiring both was redundant at best — and at worst it is precisely the shape of this bug: a second condition that had to be kept in step with the first and wasn't. One condition now.

Three tests, only the first moving:
- a replacement that dropped the colour still stashes `#ff0000` — was `expected undefined to be '#ff0000'`
- a previous EXPRESSION colour is not stashed (no solid form to restore)
- an existing stash survives a pattern→pattern replacement with no colour to displace

---

**Rebased onto `639b3d3c2e`, and one of the PRs in that batch collides with this one semantically.** #1054 (#917's export strip) deletes a builtin `fill-pattern` and falls back to `DEFAULT_FILL_COLOR` "when paint has no colour" — but EDIT-05 means a builder-patterned polygon *never* has a colour in paint, because it lives in `builder.fillColorSaved`. So every patterned polygon was exporting brand blue and discarding the user's choice.

Measured rather than assumed: a POLYGON with `fill-pattern: hatch` and a stash of `#ff0000` emitted `{'fill-opacity': 0.5, 'fill-color': '#3b82f6'}`.

`_strip_builtin_fill_pattern` can't read the stash — it runs from `_validate_emitted_style`, which walks emitted layers with no access to the source layer — so I seed the colour earlier, in the fill branch where the builder is already in scope for the `strokeDisabled` check. #1054's function is untouched; the strip simply finds a colour present and keeps it. Gated to a string stash, only when paint has no colour of its own, and only for the five builtin ids: a non-builtin id names a real `map_icons` entry that #917 deliberately preserves, and seeding a colour beside a surviving pattern would alter a working exported style. All three cases pinned.

Three LOC-cap collisions resolved along the way; `style_json.py` is 1554 with both #917's and this branch's carve-outs kept.

Gates at `5656b9a7a`: typecheck, lint, 367 files / 4413 frontend tests; `test_layering.py` + `test_maps_style_json.py` 120 passed; ruff clean.

ishiland:
@codex review

ishiland:
Real, and it is the map-disagrees-with-the-legend failure this PR exists to remove, so it gets fixed rather than deferred.

Traced the sequence: Advanced JSON (or `replace_paint`) swaps a categorical `fill-color` expression for a pattern-only paint object and leaves `style_config` behind, so the classification is already orphaned before None is clicked. None then installed a solid colour beside it, `DataDrivenStyleEditor` saw a persisted config matching its own state and skipped regenerating the expression, and the layer drew solid while the editor and legend both still reported attribute styling.

Took the clear rather than the rebuild. None means "plain solid fill", and this component cannot regenerate a ramp expression — that is `DataDrivenStyleEditor`'s job, and reaching into it from the fill picker would put the same rule in two places. The classification being dropped is not live styling: its expression was destroyed by the replacement, one step earlier.

Scoped to a config that targets COLOUR. A graduated `target: 'radius'` or `'width'` describes something the fill picker never touched, so it survives; only the builder block is kept otherwise, which is the shape `DataDrivenStyleEditor`'s own clear paths already produce.

Two tests, and the second is the guard on the first:
- the orphaned categorical config is gone after None, and paint carries a real colour — was `expected 'categorical' to be undefined`
- a `target: 'width'` classification is untouched — passes either way, which is what shows the clear is scoped and not a blanket wipe

Gates at `b1bf45e5d`: typecheck, lint, 367 files / 4415 tests.

---

On process, since this is round 19: rounds 12 through 19 each found a real defect, several of them mine, so the overrun earned its keep. But the findings have moved from "this rule misses a call path" to increasingly deep multi-step Advanced-JSON sequences, and that is the point where a PR is better served by landing than by another round. Unless the next pass surfaces something that loses user data or lies to the user about their styling, I intend to file it as a follow-up issue against #910 rather than keep extending this branch.

ishiland:
@codex review

ishiland:
Two findings. Fixing the second, declining the first, and the reason matters.

## Retain classification when None exposes its expression — fixed

This one is a regression I introduced last round, and it is the same lie pointing the other way: a layer can legitimately hold an expression `fill-color` AND a `fill-pattern` — the picker produces that state deliberately, because deleting an expression is unrecoverable — and None removes only the pattern there. The expression keeps drawing while my clear stripped the metadata the editor and legend read.

Gated on whether None actually **installed** a solid colour, which is precisely what makes a classification orphaned. Paint that already carried a colour keeps its config. Test: expression + pattern + categorical config → None → expression retained, pattern gone, `mode` still `categorical`. It was `expected undefined to be 'categorical'` against the previous commit.

## Drop a target expression when a copied pattern wins — declining

Checked this against the branch before answering, and the two cases are indistinguishable where you have anchored it. `resolveFillExclusions` sees "pattern introduced, colour untouched" for both the paste you describe **and** for `handleFillPatternChange` applying a pattern to a layer whose colour is an expression — which this PR does on purpose, and which `use-layer-map-sync.test.ts` pins with previous paint `{fill-color: RAMP}` and incoming `{fill-color: RAMP, fill-pattern: hatch}`. Implementing the remedy deletes the expression in both, so it would revert an earlier decision in this same PR and break the test guarding it.

On the consequence: the state you describe is the trade-off that decision already accepted, not a new defect. Both keys persist, the pattern wins on the map, and the expression survives because nothing can bring it back — `fillColorSaved` holds strings only. After None the map draws the ramp and the legend reports the ramp, so they agree; the disagreement exists only while the pattern is active, which is the known EDIT-05 residue on the Advanced-JSON path.

That leaves a real product question I do not think a serializer should answer unilaterally: should pasting a patterned style over a classified layer discard that classification, the way pasting a solid style does? If yes it belongs with the clipboard's own semantics and wants its own issue, because it is about what "apply this style" means, not about which paint key wins. Happy to file it against #910 — say the word if you would rather it land here instead, and I will take the expression deletion with the test updated deliberately rather than incidentally.

Gates at `2e6391cb2`: typecheck, lint, 367 files / 4416 tests.

ishiland:
@codex review

ishiland:
Round 20's second finding was mine to own, and fixing it where it was reported turned out to be the wrong move twice over. `2a087801a` moves it.

**The None click was where it was visible, not where it broke.** The invariant is that `style_config` and `paint` must never disagree about whether a colour classification exists. Advanced JSON and the AI `replace_paint` action are what break it; the fill picker is just the first control downstream to trip over the result. Both of my gates at that site were wrong in one direction or the other — round 19's stripped a live classification, round 20's fix kept an orphaned one whenever None found a colour already in paint.

So the rule now lives at the commit boundary every in-session write already passes through, keyed off the **resolved paint** — is there actually a colour expression in the key this geometry uses — rather than off which caller wrote it or why. The None-site guard is deleted rather than kept alongside it.

**The skip guard in `DataDrivenStyleEditor` is the second half, and separately wrong.** `styleConfigAlreadyMatches` compares the saved config against the effect's own state, which silently assumes paint agrees with config — so it cannot detect the one case where that assumption fails. It matters beyond the boundary because a layer arrives orphaned from **storage** (written straight through the API/SDK, or saved before this rule existed), where no in-session write ever happened for the boundary to see. Both colour effects now check first, and regenerating is also the repair.

**Two states that look orphaned and are not**, which is where this could have gone badly:

- A config with no classes. An all-null column legitimately yields `categories: []`, and `buildCategoricalExpression` correctly emits a bare colour rather than a below-arity `match` (#527). Treating that as orphaned makes the effect regenerate on every render — the #461 loop.
- `render_mode` heatmap/symbol, which park a classification while the renderer paints something else. That is also what leaves `use-render-mode-layers.ts` — the third writer of layer paint — with nothing to reconcile.

The strip drops exactly the fields the predicate reads to detect the claim, so detection and removal cannot drift apart. `render_mode`, `symbol` and the builder block survive: folding an orphaned classification into a builder-only config would silently un-cluster a point layer. `column` and `ramp` survive too, so reopening the editor re-asserts the classification instead of quietly forgetting the user's column.

**Fixture debt this surfaced.** Nine tests across four suites paired a classification with a flat colour — a layer no save can produce, since the config and its expression are written together and persist together. They now carry the expression, which is what makes them test the guard rather than the fixture. One of them cited `scripts/seed-showcase.py`; I checked the script, and the real seeded layer does carry a `case`/`match` expression, so the fixture was the incomplete part, not the seed.

Verification: `npm run typecheck`, `npm run lint`, 367 files / 4433 tests green. Each guard in the new predicate is mutation-checked — inverting `targetsColor`, `claimsClasses`, the render-mode exemption, the geometry key, and the strip itself each fail at least one test.

Same stopping rule as before: unless the next pass finds something that loses user data or misreports styling, I'll file it as a follow-up against #910 rather than keep extending this branch.

ishiland:
@codex review

ishiland:
Real, reachable, and it loses styling the user set — fixed in `3916b60f2`.

I verified the path before fixing rather than taking the reproduction on faith: `AdvancedJsonEditor.handleApply` does `JSON.parse(text)` on the whole block and hands the result straight to `onPaintChange`, so **every array-valued key comes back as a fresh object** even when its text never changed. `previousPaint[key] !== effectivePaint[key]` then read an untouched `fill-color` expression as a deliberate colour change, the pattern lost the tiebreak, and it was deleted from both the saved paint and the live map.

Worth noting why this only bites now: an expression `fill-color` sitting alongside a `fill-pattern` is a state this PR *deliberately* preserves (an expression has no stashable form, so deleting one is unrecoverable — round 13). The fix that created that state is what made this reference comparison reachable.

`deepEqual` from `LayerStyleEditor/utils.ts` is the fix, not a new helper: it is the same comparison the builder's dirty check already runs over `paint`, so "same JSON, new object" now means *unchanged* in both places instead of only one. That inconsistency was the actual bug.

**I did not take the second half of the suggestion.** Propagating which JSON keys the editor changed would put intent back in the caller's hands one caller at a time, and moving this rule to the commit boundary was specifically to stop that — the writers are an open set, and a flag is only as complete as the callers who remember to pass it. Structural comparison keeps the decision keyed off state, where an unenumerated path can't escape it.

Two tests, both against the real shape (`JSON.parse(JSON.stringify(paint))`, then edit one key):

- an unrelated `fill-opacity` edit keeps the pattern, the expression and the new opacity. Fail-before with `!==` restored: `expected undefined to be 'geolens-fill-hatch'`.
- a genuinely different reparsed expression still drops the pattern — otherwise the fix would just disable provenance for expressions rather than repair it.

`npm run typecheck`, `npm run lint`, 367 files / 4435 tests green.

ishiland:
@codex review

ishiland:
Real, and it's a regression the previous commit introduced. Fixed in `6fb856e91` — though not with the remedy you led with, because I measured that one and it doesn't work.

**Reproduced first.** Rendering the editor on a settled classification and then re-rendering with what the boundary actually commits after a paint replacement (`mode` gone, `column`/`ramp` kept, flat colour in paint):

```
CALLS AFTER RETIREMENT: 1
  regenerated mode: categorical
  fill-color is expression: true
```

So the user's solid colour is overwritten by the ramp on the very next render. Your diagnosis is exactly right, including *why*: local `mode`/`column` drive it.

**But stripping `column` and `ramp` doesn't fix it.** Same harness, config stripped all the way to `null`:

```
CALLS AFTER RETIREMENT: 1
```

`column` and `mode` are `useState` seeded at mount and never re-read from the config, so removing config fields is invisible to an editor that is already open. It would only help the closed-editor case.

**The actual shape.** The classification lives in **three** places — the paint expression, `style_config`, and this component's local state — and the third is the copy nobody reconciled. The previous commit taught the first two to agree and left the third to re-assert itself. That is your second suggestion, arrived at from the other end: local state has to follow the layer.

Clearing `column` is enough — all three effects bail on it, and it is what the editor's own clear paths (`handleClear`, `handleColumnChange('')`, `handleModeChange`) already do.

Two details that decide whether this works:

- **Adjust-state-during-render, not an effect.** React re-runs the component before running any effect, so effect 1 never gets a pass with a stale column. An effect declared *above* effect 1 would still lose the race, because a state update doesn't apply mid-pass.
- **Keyed off the claim being RETIRED, not absent.** A first-time column choice moves the other way — local state legitimately ahead of a config the effects are about to write — so a rule keyed on absence would clear the column the user just picked before it took effect.

It also fixes something nobody filed: **Reset was being partially undone the same way.** It writes a builder-only config, which retires the claim, and the editor re-asserted the classification straight back if the data-driven section was open.

Kept `column`/`ramp` in the retired config deliberately. They are the user's selections, not the claim, and with local state now following the layer they no longer cause a re-assertion — they just mean re-enabling a classification starts from where the user left off.

Fail-before on the new test: `expected "vi.fn()" not to be called at all, but it was called 1 times`. `npm run typecheck`, `npm run lint`, 367 files / 4436 tests green.

ishiland:
@codex review

ishiland:
Real, and it's the same lie one step further in — an array was being taken as proof the classification still held. Fixed in `1642efb0b`.

A classification **names a column**, so an expression that never reads that column cannot be the one the config describes. That is the whole test.

**I deliberately did not go as far as "inspect the expression's shape."** Working out *which* `get` in a hand-authored expression is the classification has no right answer — `['case', ['has', 'zone'], ['match', ['get', 'era'], …], '#ccc']` reads two columns and is perfectly valid — and a wrong guess deletes hand-authored category colours, which #461 exists to protect. The conservative direction gives up nothing real, because every builder-generated expression reads the config's column by construction; it only declines to catch a *replacement* expression that happens to still mention the old column. I'd rather under-fire there than delete a live classification.

**One thing worth flagging, because I got it wrong first.** `expressionReadsColumn` already existed, privately, in `basemap-utils.ts`. My first attempt exported it from `maplibre-filter-utils.ts` beside its sibling `isGetExpression` — and three hook suites mock that module wholesale, so the helper came back `undefined`. That reads as "this expression does not mention the column" and **retires live classifications**. Two of my own boundary tests failed immediately, which is how I found it. It now lives in a new `maplibre-expressions.ts` that nothing stubs, and the reason is written down there so it doesn't get moved back.

Fixtures now pair each expression with its own config's column through `colorExprFor(column)`, a helper that takes the column, so a fixture can't quietly claim one column while painting another. Two of them were doing exactly that once this check existed.

Tests: the `era`→`status` replacement is orphaned; a `case`-wrapped expression is found at depth; a multi-column expression that still reads `era` is accepted; a config naming no column falls back to the array check. Fail-before on the first: it returned `false`. The over-fire mutation (accept any array) breaks two.

`npm run typecheck`, `npm run lint`, 367 files / 4439 tests green.

Unrelated to the review: main has moved four times in the last half hour and #1022 is now DIRTY on `backend/tests/test_layering.py` — #1044 relocated the two LOC-cap entries this branch edits. The resolution is mechanical (main's structure, my line counts: 1554 and 1317, neither source file touched by the batch) and I'm holding it until main settles rather than rebasing into a landing batch.

ishiland:
@codex review

ishiland:
You're right, and this is the half of your round-22 remedy I declined. Fixed in `0c5ffaf65`.

I said then that stripping `column` didn't fix anything, because local state drives the regeneration — and I proved that for an editor that is **already open**. What I didn't check was the editor being **closed and reopened**, where stripping `column` is the only thing that helps. Reproduced at the component:

```
CALLS ON FRESH MOUNT: 1
  mode: categorical | expression: true
```

The mechanism is exactly as you describe: a fresh `DataDrivenStyleEditor` seeds local `column` from the config and defaults local `mode` to `'categorical'`, so a config carrying a column and no mode reads to it as a live categorical classification. And `claimedLastRender` initialises to the already-false state, so the transition guard has nothing to observe — the claim was gone before the component first rendered.

So the two fixes cover different cases and both are needed:

| case | what covers it |
| --- | --- |
| editor already open when the claim is retired | the transition guard (`6fb856e91`) |
| editor closed, opened afterwards | dropping `column`/`ramp` (this commit) |

Which also means my stated rationale — "keeping the column means reopening the editor re-asserts the classification instead of forgetting it" — was describing the bug as if it were a feature. Two tests asserted that decision explicitly, with comments explaining it. I flipped them rather than deleting them, since the reasoning they recorded is the thing that was wrong.

`ramp` goes too. It is meaningless without a mode, and leaving it behind is the stray-field hazard #392 already documents.

Fail-before on the narrow key list: `drops the claim and keeps everything that is not the claim` and `drops the claim when a paint replacement leaves no expression behind` both fail. There is also a new component-level test pinning the consequence where the damage would show up — no classification on mount from a column with no mode behind it — so this can't regress silently if the strip list is ever narrowed again.

`npm run typecheck`, `npm run lint`, 367 files / 4440 tests green.

ishiland:
@codex review

ishiland:
Rebased onto `bddfb8d1d6` (`c55692a3e`) after main went quiet — the previous head `0c5ffaf65` had a clean pass, but the branch was DIRTY, and a DIRTY branch gets **no CI at all**, so the clean pass was the only signal it had.

One conflicting file, `backend/tests/test_layering.py`: #1044 moved `maps/style_json.py` out of the second cap dict into `_OPEN_CORE_SIZE_CAPS`, which is exactly where this branch's commits edited it. Resolution keeps #1044's structure and its written inclusion rule, and re-applies this branch's delta at the new location — 1517 → 1554, **measured against the file** rather than carried across from the old entry, since `test_module_loc_caps_have_no_headroom` is exact-match. `maps/schemas.py` needed nothing; the 1317 this branch set merged cleanly and still matches.

Re-verified after the rebase, not before: `test_layering.py` 36 passed, `test_maps_style_json.py` 86 passed, frontend typecheck + lint clean, 367 files / **4449** tests (up from 4440 — main's batch brought nine frontend tests with it).

Not armed.

ishiland:
@codex review

ishiland:
Both real, both fixed in `d9a7088eb`, and they're the same mistake in two languages: asking whether a fill key is **present** when the question is whether it's **active**.

**Frontend.** Confirmed by fail-before — with the presence-only collision test restored, `expected undefined to be '#ff0000'`. The solid colour is deleted and stashed on an unrelated `fill-opacity` edit, and the layer drops to MapLibre's spec default.

Your pointer to `FillEditor`'s `paint['fill-pattern'] != null` is the part that made this worth fixing rather than debating: the codebase already had a convention for "active pattern" and the exclusions disagreed with it. Same shape as the `deepEqual` finding two rounds ago — an internal inconsistency, not a missing feature. It now applies at **all three** places the exclusions ask, including the stale-stash check, where a null pattern makes the stash exactly as stale as an absent key does. That third site has its own fail-before (the collision mutation alone doesn't reach it).

**Backend.** I measured the export before fixing, because the stated consequence didn't match what I expected `_strip_wrong_typed_values` to do:

```
emitted paint: {'fill-color': None, 'fill-opacity': 0.5}
```

You're right and my expectation was wrong — the null is **not** popped, so it ships and MapLibre resolves it to its own default rather than the saved `#ff0000`. The seed now accepts absent-or-null, and the test records the measured before-state so the next reader doesn't have to re-derive it.

Checked the neighbours for the same blind spot while I was in there: `colorClassificationIsOrphaned` already handles it correctly (a null colour isn't an array, so a classification claiming one is genuinely orphaned), and `patternOwnsFill`'s `typeof fillColor === 'string'` was never fooled by null.

Ratchet 1554 → 1559 for the seed comment; `test_layering.py` 36 passed, `test_maps_style_json.py` 87 passed, ruff check + format clean, frontend typecheck + lint, 367 files / 4451 tests.

ishiland:
@codex review

ishiland:
Correct analysis, and I'm declining it here — filed as #1087 instead. This is the first finding since round 20 that doesn't clear the bar I set, and the honest fix is in a file this branch is barred from touching.

**Where it actually breaks.** The distinguishing fact is which fill keys the *source style* owned, and `applyCopiedStyleToLayer` erases that before any rule downstream can see it: `{...target.paint, ...copied.paint}` cannot express "the source's fill is solid, so drop the pattern," because a spread has no way to remove a key the source doesn't mention. The equal-valued case you found is where that erasure becomes visible; the unequal case only works by accident, because the new colour looks like an introduction.

So the fix is in `layer-style-clipboard.ts` — a pasted style should **replace** the fill representation, not union with it. That file belongs to #923 and I'm under instruction not to touch it.

**Why I won't compensate for it on my side.** `handleBulkApplyStyle` still holds the `CopiedStyle` and could recover the ownership; `handlePasteStyle` merges first and hands the flattened paint to `handleStyleConfigChange`, which cannot. Fixing the half that has the source would make paste and bulk-apply disagree with each other — and it would put caller-declared intent back into a rule that was moved to the commit boundary precisely to keep it out. That trade was the whole point of rounds 15–20.

**Why it stays below the bar.** Nothing is lost: the displaced colour is in `fillColorSaved` and the pattern is the user's own. Nothing misreports either — I checked `FillEditor`, which reads `paint['fill-pattern'] != null`, so the editor, the legend swatch and the map all agree the layer is patterned. And it needs two uncommon conditions together: a target already holding both keys (only reachable from Advanced JSON, the API/SDK or an import) *and* a pasted colour exactly equal to the dormant one.

#1087 has the reproduction, the root cause, the reason it can't be fixed from here, and a suggested test including the mirror case so the replacement lands symmetric rather than special-cased.

Head is unchanged at `d9a7088eb`, which your previous pass already reviewed clean apart from the two null findings I fixed there.

PR #1023, opened 2026-07-31
chore(scripts): show the version-site list even when the CHANGELOG check fails

### LINKED ISSUE #1019: chore(scripts): version-check hides the per-site list when the CHANGELOG check fails
Decomposed from #893 (item 7). Verified against `main` at `9a788ba92` on 2026-07-31.

## The problem

`scripts/check_version_coherence.py` checks two things: that every version site agrees on the canonical version, and that `CHANGELOG.md` carries a populated `## [<version>]` section. When the CHANGELOG half fails, the script returns before printing anything about the first half, so a failing `make version-check` never reveals whether the lockfiles and manifests stamped correctly.

The order in the source:

- `:228` and `:234` — the "no `## [<version>]` section" failure path, returning 1.
- `:237` and `:242` — the "section is empty" failure path, returning 1.
- `:244` — `print(f"OK: all {len(sites)} version sites agree on {canonical}.")`
- `:246` — the per-site loop that prints each site and its value.

Both CHANGELOG returns sit above the per-site print, so the information you most want during a trial bump is exactly the information you do not get.

## Fix

Print the per-site list before the CHANGELOG checks run, or collect both failures and report them together rather than returning on the first. Either shape works; the requirement is that a failing run shows the version-site state.

Keep the exit code semantics unchanged: 0 for coherent, 1 for drift, as the module docstring at `:48` states.

## Why it is worth doing at all

Only noticeable when trial-bumping, which is exactly when someone is least able to tell a lockfile that did not stamp from a CHANGELOG section they forgot to write. `make version-check` is the CI gate for `make bump`, and `AGENTS.md` says never to edit a version string by hand, so this script's output is the whole feedback loop for that workflow.

## Verification

Trial-bump locally with a deliberately missing CHANGELOG section and confirm the per-site list still prints:

```
make bump VERSION=9.9.9   # then remove the CHANGELOG section it wrote
make version-check
```

Revert the bump afterwards.

ishiland:
## What changed

`scripts/check_version_coherence.py` checks two independent things: that every version site agrees with the canonical `backend/pyproject.toml` version, and that `CHANGELOG.md` carries a populated `## [<version>]` section. Both CHANGELOG failure paths returned above the per-site print, so a failing `make version-check` told you the CHANGELOG was wrong and said nothing about whether the 19 version sites had stamped.

That is backwards for the one workflow the script exists to serve. During a trial bump the per-site list is exactly what you need to tell a lockfile that did not stamp from a CHANGELOG section you have not written yet.

- The per-site list now prints before any check can return, each site tagged `[ok]` or `[DRIFT]`.
- Drift and CHANGELOG verdicts are collected and both reported, instead of returning on the first.
- The `OK: all N version sites agree on X.` summary moved to the end, where it now means "both halves passed" rather than heading a list.

No change to exit-code semantics: 0 coherent, 1 drift, as the module docstring states.

None. Stdlib-only script, no runtime code, no CI contract change — `.github/workflows/ci.yml:729` and `make version-check` only read the exit code.

## Verification

Clean tree:

```
python3 scripts/check_version_coherence.py   # exit 0, 19 sites + CHANGELOG line, then OK
python3 scripts/tests/test-changelog-gate.py # 20 passed, 0 failed
```

Failure paths, exercised by temporarily mangling the tree and restoring with `git checkout --`:

- CHANGELOG header renamed away: exit 1, all 19 sites still listed.
- Root `package.json` set to `9.9.9` **and** the CHANGELOG header renamed away: exit 1, both `FAIL:` blocks on stderr, `package.json` tagged `[DRIFT]` in the listing.

Closes #1019

### DIFF
--- a/scripts/check_version_coherence.py
+++ b/scripts/check_version_coherence.py
@@ -207,16 +207,30 @@ def main() -> int:
     # Canonical = backend distribution version (what the app derives at runtime).
     canonical = sites[f"{_rel(BACKEND_PYPROJECT)} ([project].version)"]
 
-    offenders = {site: v for site, v in sites.items() if v != canonical}
+    # fix(#1019): print every site BEFORE any check can return. Both CHANGELOG
+    # failure paths used to sit above this listing, so a `make version-check`
+    # that failed on the CHANGELOG never said whether the lockfiles and
+    # manifests had stamped — which is the whole reason you run it mid-bump.
+    # For the same reason both halves are now collected and reported together
+    # instead of returning on the first.
+    print(
+        f"Version sites ({len(sites)}), "
+        f"canonical (backend/pyproject.toml) = {canonical}:"
+    )
+    for site, v in sites.items():
+        print(f"  [{'ok' if v == canonical else 'DRIFT'}] {site}: {v}")
 
+    failed = False
+
+    offenders = {site: v for site, v in sites.items() if v != canonical}
     if offenders:
+        failed = True
         sys.stderr.write(
             f"FAIL: version drift detected. Canonical (backend/pyproject.toml) = {canonical!r}.\n"
             f"Run `make bump VERSION={canonical}` to resync, or correct the offending site:\n"
         )
         for site, v in offenders.items():
             sys.stderr.write(f"  - {site}: {v!r} != {canonical!r}\n")
-        return 1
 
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review

PR #1024, opened 2026-07-31
fix(sandbox): admit EXISTS subqueries in the SQL validator

### LINKED ISSUE #1017: fix(sandbox): every EXISTS subquery is rejected — exp.Exists resolves to an unlisted function name
Split out of #1001, which is a large design problem. This one is not: it is two lines, it is independent of the buffer work, and it is live on `main` today.

## The defect

Every `EXISTS` subquery is rejected by the NL to SQL sandbox validator, whatever it contains.

Reproduced against `main` at `9a788ba92`:

```python

validate_sql(
    "SELECT name FROM data.t AS a "
    "WHERE EXISTS (SELECT 1 FROM data.u AS b WHERE b.id = a.id)"
)
```

```
sandbox.unlisted_function  function=exists
SandboxError: Query uses a disallowed function
```

## Why

`_check_function_allowlist` (`backend/app/platform/sandbox/validator.py:310`) walks `stmt.find_all(exp.Func)` and requires every node's canonical name to appear in an allowlist. sqlglot models `exp.Exists` as an `exp.Func` subclass (its MRO is `Exists -> Expression -> Func -> SubqueryPredicate -> Predicate -> Condition`, and `sql_names()` returns `['EXISTS']`), so the walk yields it with the name `exists`. That name is in neither `_ALLOWED_FUNCTIONS` (:113) nor `_ALLOWED_POSTGIS_FUNCTIONS` (:263), and the check fails closed, so the query is refused.

`EXISTS` is a subquery predicate, not a callable. There is nothing to allowlist about it, and `find_all` walks into its operand regardless, so any genuinely disallowed function inside the subquery is still caught on its own node.

This is the same shape as #538, where `AND` and `OR` hit the identical trap: sqlglot reaches them through `exp.Connector`, which sits on a `Func` path, so a compound `WHERE` clause was rejected as an unlisted function.

## Fix

Skip `exp.Exists` the way `exp.Connector` is already skipped, beside the existing branch at `validator.py:332`:

```python
if isinstance(func, exp.Connector):
    continue
```

Add the `exp.Exists` case there with a comment in the same style, referencing this issue and #538, and one test asserting a plain `EXISTS` subquery validates. A negative control comes free: assert that a disallowed function placed inside the `EXISTS` subquery is still rejected, which proves the skip did not open a hole.

## Scope note

Landing this does **not** admit the canonical geodesic buffer. `exists` is one of seventeen names that expression needs, and the other sixteen are the subject of #1001. The two can land in either order, and this one should not wait: it fixes every non-buffer `EXISTS` question in the NL to SQL surface on its own.

## Acceptance

- `validate_sql` admits a plain `EXISTS` subquery.
- A disallowed function inside an `EXISTS` subquery is still rejected.
- `cd backend && uv run pytest tests/test_sec025_sandbox_allowlist.py -v` passes.

ishiland:
## What changed

`_check_function_allowlist` walks `stmt.find_all(exp.Func)` and requires every
node's canonical name to be allowlisted. sqlglot models `exp.Exists` as an
`exp.Func` subclass, so the walk yielded it as function `exists`, a name that is
in neither `_ALLOWED_FUNCTIONS` nor `_ALLOWED_POSTGIS_FUNCTIONS`. Every `EXISTS`
subquery was therefore refused on `main`:

```
sandbox.unlisted_function  function=exists
SandboxError: Query uses a disallowed function
```

`exp.Exists` is now skipped beside the `exp.Connector` skip that #538 added for
the identical trap with `AND`/`OR`. `EXISTS` is a subquery predicate, not a
callable, and `find_all` still walks its operand, so a disallowed function
inside the subquery is caught on its own node.

## Security impact

No widening. The skip suppresses the check for one predicate node; every node
inside the subquery is still validated. Three negative controls cover that: a
blocked function (`pg_sleep`), an unlisted function (`version()`), and an
unlisted spatial function (`ST_MakeEnvelope`) placed inside the `EXISTS`
subquery are all still rejected, and a recursive CTE inside `EXISTS` still trips
the recursive-CTE guard.

Those controls assert the validator's logged `function` field, not just that
something raised. Without that, they would pass on the pre-fix `exists`
refusal and prove nothing. Verified by stashing the validator change: six of the
seven new tests fail without it.

## Scope

This does not admit the canonical geodesic buffer; `exists` is one of the
seventeen names that expression needs, and the other sixteen are #1001.

## Verification

```
cd backend && uv run pytest tests/test_sec025_sandbox_allowlist.py tests/test_sandbox.py tests/test_sandbox_tenant_schema.py -q
# 162 passed
```

Closes #1017

### DIFF
--- a/backend/app/platform/sandbox/validator.py
+++ b/backend/app/platform/sandbox/validator.py
@@ -331,6 +331,13 @@ def _check_function_allowlist(stmt: exp.Expression, sql: str) -> None:
         # side of an AND/OR is still caught below.
         if isinstance(func, exp.Connector):
             continue
+        # fix(#1017): sqlglot reaches EXISTS through the same Func path (#538),
+        # so every EXISTS subquery was rejected as an unlisted "function".
+        # EXISTS is a subquery predicate, not a callable — and find_all walks
+        # its operand regardless, so a disallowed function inside the subquery
+        # is still caught below.
+        if isinstance(func, exp.Exists):
+            continue
         if isinstance(func, exp.Anonymous):
             fn_name = func.name.lower() if hasattr(func, "name") else ""
         else:
--- a/backend/tests/test_sec025_sandbox_allowlist.py
+++ b/backend/tests/test_sec025_sandbox_allowlist.py
@@ -15,6 +15,7 @@ RED → GREEN history:
 from __future__ import annotations
 
 import pytest
+from structlog.testing import capture_logs
 
 from app.platform.sandbox.schemas import SandboxError
 from app.platform.sandbox.validator import validate_sql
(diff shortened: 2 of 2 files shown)


ishiland:
@codex review

Issue #787, opened 2026-07-27
bug(builder): frontend polish batch from the 2026-07-27 audit (P3)


ishiland:
Small confirmed items, batched. Original numbering is kept so the triage comment below still reads; struck items stay listed rather than deleted.

Kept as one issue on purpose. Six live items across six files, each a handful of lines, and every one of them is decidable from source now. Splitting them into six issues costs more in triage than the work itself. Pointers re-verified on main @ 9a788ba92.

1. The "Result · N features" chip renders bottom-left **under** the open Legend plugin panel (distinct from #731, which was chip vs toolbar). This no longer needs a live repro; it is decidable from source. `frontend/src/components/builder/EphemeralBadge.tsx:48` carries `z-[8]`, and every `PluginHost.tsx` slot is `z-10` (`:16-19`), so a plugin panel wins wherever the two overlap. The bottom-left slot is positioned to clear the badge (`PluginHost.tsx:13,18`), but a panel taller than that offset overlaps it and the z-index decides. Fix: raise the badge above the plugin slots, one line at `EphemeralBadge.tsx:48`, then confirm the raised badge does not cover anything else in that corner.

2. ~~The panel's post-run "Add to map" never names which dataset it adds and stays active after use~~. **Shipped:** the affordance now dedupes on dataset id via `useAnalysisAddedStore` (`stores/analysis-job-store.ts`), settled globally in `useAddLayer` (`hooks/use-maps.ts`); `AnalysisPanel` disables the button once the id is pending or added and renders `analysisTools.addToMapNamed` ("Add “{{name}}” to map").

3. Closing the panel mid-preview suppresses result callbacks and never aborts the request. Thread an `AbortSignal` through `previewAnalysis` (`frontend/src/api/analysis.ts:9`); `materializeMutation` (`AnalysisPanel.tsx:784`) already works around the unmount-suppression, `previewMutation` (`:692`) does not. Smaller than it looks: `apiFetch` already composes a caller-supplied signal with its own timeout via `AbortSignal.any` (`frontend/src/api/client.ts:219-226`), so the work is passing the signal down, not building cancellation.

4. Moved to #941. The history entry is accurate about an immediate server POST, so it is a copy fix on the server-side summary rather than a polish fix here.

[...]

PR #1025, opened 2026-07-31
fix(builder): abort a closed panel's preview and hold the map in state

ishiland:
Two items from the #787 audit batch, carved out ahead of the analysis-operation wave so they land before `AnalysisPanel.tsx` starts collecting new operations.

**Item 3 — a closed panel's preview kept running.** TanStack suppresses observer callbacks after unmount, so the result was dropped, but the request itself ran to completion and held its sandbox query open. `previewAnalysis` now takes an optional `AbortSignal`; `apiFetch` already composes a caller signal with its own request timeout (`AbortSignal.any`), so this is threading rather than new cancellation machinery. The panel aborts on unmount, and also when a newer preview supersedes an in-flight one — the sequence guard already dropped that response, this stops the work as well.

**Item 10 — the clip Draw button read a ref during render.** `disabled={!mapInstanceRef?.current}` was evaluated in the render pass, and a ref assignment re-renders nothing, so the button could stay dead until an unrelated state change happened to commit. The instance now lives in state: seeded from the ref at mount (the common case, where the map loaded before the panel opened) and resynced from a no-dep-array effect, the same retry-every-commit idiom the mask-overlay effects above it use.

Refs #787 — closes items 3 and 10 only. Items 1, 5, 6 and 9 live in other files and stay open.

No API, schema, or security impact. No new translation keys.

## Verification

```
cd frontend
npm run lint          # clean
npm run typecheck     # clean
npm run test:coverage # 366 files, 4286 tests, all passing
```

Two new tests in `AnalysisPanel.test.tsx`: one asserts the signal handed to `previewAnalysis` is aborted by the unmount, the other that a map arriving after mount enables the Draw button. The ten existing `previewAnalysis` call assertions gained `expect.any(AbortSignal)` for the new third argument.

### DIFF
--- a/frontend/src/api/analysis.ts
+++ b/frontend/src/api/analysis.ts
@@ -9,10 +9,14 @@ import type {
 export async function previewAnalysis(
   datasetId: string,
   body: AnalysisPreviewRequest,
+  // fix(#787 item 3): callers cancel a preview they no longer want. apiFetch
+  // composes this with its own request timeout, so whichever fires first wins.
+  signal?: AbortSignal,
 ): Promise<AnalysisPreviewResponse> {
   return apiFetch<AnalysisPreviewResponse>(`/datasets/${datasetId}/analysis/preview/`, {
     method: 'POST',
     body: JSON.stringify(body),
+    signal,
   });
 }
 
--- a/frontend/src/components/builder/AnalysisPanel.tsx
+++ b/frontend/src/components/builder/AnalysisPanel.tsx
@@ -422,6 +422,22 @@ export function AnalysisPanel({
     },
     [],
   );
+  // fix(#787 item 10): the clip Draw button gated on `mapInstanceRef.current`
+  // read during render. Ref assignments don't re-render, so the button could
+  // stay dead until an unrelated state change happened to re-render the panel.
+  // Mirror the instance into state from an effect — the legal place to read a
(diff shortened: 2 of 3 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
On the server-side half of the last finding: correct, and the comments now say so. Aborting the fetch ends the client's wait, which is what re-enables Preview; it does not reach `analysis_preview_endpoint`, which never checks `Request.is_disconnected()`, so the sandbox statement runs to its own timeout and holds the advisory lock until then.

Making the server cancel on disconnect is a backend change to the preview path (racing the query against the disconnect and cancelling the DB task), which is out of scope for a P3 frontend polish item — #787 item 3 is explicitly scoped to threading the signal, and lock contention on that path is #696's subject. Filing it separately if it is wanted.

ishiland:
@codex review

ishiland:
Confirmed and fixed. I wrote a test expecting the observer to be unsubscribed by then, and it failed: the toast fired. `onError` now returns early on `AbortError`, which covers all three abort sites, and the test is in the PR. Timeouts still report, because `safeFetch` normalizes `TimeoutError` into an `ApiError` and only a caller-initiated abort keeps that name.

ishiland:
@codex review

Issue #565, opened 2026-07-17
MCP server: add read-only `query` tool via a hardened sandbox SQL endpoint


ishiland:
## Context

The read-only MCP server (#550) ships a v1 with 5 tools — `search_datasets`, `get_dataset_schema`, `get_features`, `list_maps`, `get_map` — each a thin `GET` over an existing anonymous-capable REST endpoint. The 6th proposed tool, **`query`** (SQL through the sandbox), was deliberately **deferred**: unlike the others it is *not* a thin wrapper, because the read-only SQL sandbox (`validate_and_execute`, `backend/app/platform/sandbox/`) has **no REST endpoint** — its only caller is the AI-chat pipeline (`backend/app/processing/ai/chat_actions.py`), which takes natural language, generates SQL via an LLM, and streams SSE. There is no path today that accepts raw SQL and returns `{columns, rows}`.

Adding `query` therefore requires a new backend endpoint **plus** security hardening. This issue tracks doing it safely.

## What the risk actually is now

The original write-up leaned on the S03 finding from `sec-audit-20260710.md` and quoted two payloads. Both are now rejected at validation, verified against `9a788ba92`:

```
SELECT ST_GeneratePoints(ST_MakeEnvelope(0,0,1,1,4326), 1000000000)
  -> sandbox.unlisted_postgis_function: "Query uses a disallowed spatial function"
SELECT array_agg(generate_series(1,1000000000))
  -> sandbox.unlisted_function: "Query uses a disallowed function"
```

Four of S03's five prescribed fixes have shipped since the audit: the `st_` prefix shortcut was replaced with an explicit `_ALLOWED_POSTGIS_FUNCTIONS` allowlist, the set-generating and unbounded-aggregate functions were dropped from `_ALLOWED_FUNCTIONS`, a per-user advisory concurrency gate was added (`executor.py:157-161`), and the AI SQL statement timeout came down to `DEFAULT_TIMEOUT_MS = 10_000` (`executor.py:26`). Both allowlists are fail-closed: an unlisted name raises rather than falling through (`validator.py:342-353`). The one S03 fix not shipped is "require at least one authorized `data.*` table" — `check_table_access` (`validator.py:429`) iterates the referenced tables and does not require the set to be non-empty.

So the cost story is much better than the audit's snapshot, and the remaining exposure is narrower and more specific:

[...]

PR #1026, opened 2026-07-31
fix(sandbox): reject nested DML and thread timeout_ms through the wrapper

### LINKED ISSUE #1011: fix(sandbox): reject DML nested in CTEs and thread timeout_ms through validate_and_execute
Two mechanical gaps in the read-only SQL sandbox, split out of #565 so they can land ahead of the endpoint design work. Both are self-contained and both have a reproducible payload.

## 1. `validate_sql` accepts a writable CTE

`validate_sql` (`backend/app/platform/sandbox/validator.py:358`) isinstance-checks only the **top-level** node against `exp.Select/Union/Intersect/Except` (`:382`), then checks `exp.Into` (`:387`), recursive CTEs, and the function allowlist. Nothing walks the tree for DML nodes, so a DML statement nested inside a CTE passes validation.

Confirmed live against `9a788ba92`:

```python
>>> from app.platform.sandbox.validator import validate_sql
>>> validate_sql("WITH x AS (INSERT INTO data.foo (a) VALUES (1) RETURNING a) SELECT a FROM x")
ValidatedQuery(tables={('data', 'foo'), ('', 'x')}, cte_names={'x'})   # accepted
```

The only thing that stops this statement from executing is `SET TRANSACTION READ ONLY` in `execute_safe`. That control works, which is why this is defense-in-depth rather than an open hole. It is also the *only* control here: any future path that runs validated SQL outside that transaction, or any change to how the transaction is opened, turns a one-line regression into arbitrary writes. A validator whose docstring says it "Rejects: INSERT, UPDATE, DELETE, DROP, CREATE" should reject them.

**Fix:** reject `exp.Insert`, `exp.Update`, `exp.Delete`, `exp.Merge`, and `exp.Command` found anywhere in the tree, not just at the root. `stmt.find(...)` already walks, so this mirrors the existing `exp.Into` check.

**Tests:** use the payload above as a fixture, plus one per node type (`UPDATE ... RETURNING`, `DELETE ... RETURNING`, `MERGE`, and a `Command` node such as a nested `COPY`), and at least one nested two levels deep inside a CTE within a CTE. Assert the error is the sandbox's own `invalid_query` refusal, not a database error. Keep a passing case that proves an ordinary read-only CTE still validates.

## 2. `timeout_ms` is not threaded through `validate_and_execute`

`execute_safe` already accepts it:

```python
# backend/app/platform/sandbox/executor.py:116
timeout_ms: int = DEFAULT_TIMEOUT_MS,   # DEFAULT_TIMEOUT_MS = 10_000 at :26
```

and applies it at `executor.py:211-213` (`SET LOCAL statement_timeout`). But the public wrapper `validate_and_execute` (`backend/app/platform/sandbox/__init__.py:31-38`) exposes only `row_limit` and `restrict_tables`, and its call into `execute_safe` (`:86-92`) passes neither a timeout nor a way to supply one. Every caller silently gets the 10 s default, and no caller can ask for less.

**Fix:** add `timeout_ms: int = DEFAULT_TIMEOUT_MS` to `validate_and_execute`'s keyword-only parameters, document it in the docstring alongside `row_limit`, and pass it through to `execute_safe`. Behaviour for existing callers must be byte-identical, since the default is the value they get today.

**Tests:** assert an explicit `timeout_ms` reaches the `SET LOCAL statement_timeout` statement, and that omitting it produces the current default. Assert an existing caller path is unchanged.

## Why these are worth doing now

#565 needs both before it can add a raw SQL endpoint: the endpoint wants a tighter budget than 10 s, and it widens who can reach the validator. Landing them separately keeps #565's design questions (cost bounds for unbounded self-joins, mandatory dataset scope, the reader-role binding) from blocking two changes that are clear-cut.

## Gates

`cd backend && uv run ruff check . && uv run ruff format --check .`, then `make test`. No schema change, no OpenAPI change, no SDK regeneration.

Related: #565, `docs-internal/audits/sec-audit-20260710.md` (S03).

ishiland:
Two mechanical gaps in the read-only SQL sandbox, both reproduced against
`9a788ba92`.

## 1. `validate_sql` accepted a writable CTE

`validate_sql` isinstance-checks only the **root** node against
`exp.Select/Union/Intersect/Except`. Nothing walked the tree, so a write
statement nested inside a CTE — or a scalar subquery — validated:

```python
>>> validate_sql("WITH x AS (INSERT INTO data.foo (a) VALUES (1) RETURNING a) SELECT a FROM x")
ValidatedQuery(tables={('data', 'foo'), ('', 'x')}, cte_names={'x'})   # accepted
```

The only thing stopping that statement from executing is `SET TRANSACTION READ
ONLY` in `execute_safe`. That control works, so this is defense-in-depth rather
than an open hole — but it is the *only* control, and any future path that runs
validated SQL outside that transaction turns a one-line regression into
arbitrary writes.

`_reject_nested_mutation` now walks the tree, mirroring the existing
`exp.Into` check.

**Two deliberate departures from the issue's list, both flagged rather than
assumed:**

- The set includes `exp.Create`, `exp.Drop` and `exp.Copy` alongside
  `Insert/Update/Delete/Merge/Command`. `WITH x AS (CREATE TABLE data.evil AS
  SELECT 1) SELECT 1`, `WITH x AS (CREATE VIEW ...) SELECT 1` and `WITH x AS
  (DROP TABLE data.foo) SELECT 1` all validate on `main` today — same defect,
  same nesting path — and the validator's own docstring already says it
  "Rejects: INSERT, UPDATE, DELETE, DROP, CREATE".
- The list is explicit rather than keyed on `exp.DDL` / `exp.DML`. Those base
  classes do not partition the way the names suggest in sqlglot 30.14:
  `Insert` is both, `Drop` and `Alter` are neither. An explicit tuple is what
  stays fail-closed across sqlglot upgrades.

`ALTER ... ` and `TRUNCATE` inside a CTE fail to parse, so they never reach the
walk; `exp.Alter` is listed anyway for the same fail-closed reason.

## 2. `timeout_ms` was not threaded through `validate_and_execute`

`execute_safe` has always accepted `timeout_ms` and applied it as `SET LOCAL
statement_timeout`, but the public wrapper exposed only `row_limit` and
`restrict_tables`, so every caller silently got the 10 s default and none could
ask for less. Added as a keyword-only parameter defaulting to
`DEFAULT_TIMEOUT_MS` and passed through.

No production caller passes it, so existing behaviour is byte-identical. #565
needs it: a synchronous raw-SQL endpoint wants a tighter budget than 10 s.

## Security impact

Tightening only. No new function, table or statement form is admitted; the
read-only CTEs the SQL generator emits are covered by passing cases, including
a CTE nested inside a CTE.

## Verification

```
cd backend && uv run pytest tests/test_sandbox.py tests/test_sandbox_tenant_schema.py \
  tests/test_sec025_sandbox_allowlist.py tests/test_sql_engine.py tests/test_ai_chat.py \
  tests/test_ai_chat_dataset.py tests/test_ai_chat_actions_phase1135.py \
  tests/test_chat_narrative.py -q -n 4
# 256 passed
```

Closes #1011

### DIFF
--- a/backend/app/platform/sandbox/__init__.py
+++ b/backend/app/platform/sandbox/__init__.py
@@ -15,7 +15,7 @@ import structlog
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from app.core.identity import Identity
-from app.platform.sandbox.executor import execute_safe
+from app.platform.sandbox.executor import DEFAULT_TIMEOUT_MS, execute_safe
 from app.platform.sandbox.schemas import SandboxError, SandboxResult
 from app.platform.sandbox.validator import (
     build_table_allowlist,
@@ -34,6 +34,7 @@ async def validate_and_execute(
     user: Identity | None,
     *,
     row_limit: int = 1000,
+    timeout_ms: int = DEFAULT_TIMEOUT_MS,
     restrict_tables: frozenset[str] | None = None,
 ) -> SandboxResult:
     """Validate and safely execute a SQL query.
@@ -49,6 +50,10 @@ async def validate_and_execute(
         db: Async database session.
         user: Current user (None for anonymous).
         row_limit: Maximum rows to return (default 1000).
+        timeout_ms: Statement timeout in milliseconds, applied as SET LOCAL
+            statement_timeout on the execution connection (default 10000).
+            Callers that need a tighter budget than the shared default — a
+            synchronous request path, say — pass a smaller value.
         restrict_tables: Optional surface-level scope: when set, the effective
             allowlist is the INTERSECTION of the user's RBAC allowlist with
             this set — it can only narrow access, never widen it. Used by
@@ -88,6 +93,7 @@ async def validate_and_execute(
             db,
             validated.sql,
             row_limit=row_limit,
+            timeout_ms=timeout_ms,
             concurrency_key=concurrency_key,
         )
 
--- a/backend/app/platform/sandbox/validator.py
+++ b/backend/app/platform/sandbox/validator.py
@@ -777,6 +777,37 @@ def _canonical_buffer_exempt_ids(stmt: exp.Expression) -> set[int]:
     return exempt
 
 
+# Statement types that write. `validate_sql` isinstance-checks only the ROOT
+# node, so before #1011 any of these nested inside a CTE or a scalar subquery
+# passed validation — `WITH x AS (INSERT ... RETURNING a) SELECT a FROM x` was
+# accepted, and only `SET TRANSACTION READ ONLY` in `execute_safe` stopped it
+# from running. Listed explicitly rather than by sqlglot base class: `exp.DDL`
+# and `exp.DML` do not partition the way the names suggest (Insert is both,
+# Drop and Alter are neither), so an explicit tuple is what stays fail-closed
+# across sqlglot upgrades.
+_MUTATING_STATEMENTS: tuple[type[exp.Expression], ...] = (
+    exp.Insert,
+    exp.Update,
+    exp.Delete,
+    exp.Merge,
+    exp.Copy,
+    exp.Create,
+    exp.Drop,
+    exp.Alter,
+    exp.Command,  # sqlglot's catch-all for statements it does not model
+)
+
+
+def _reject_nested_mutation(stmt: exp.Expression, sql: str) -> None:
+    """Reject write statements found anywhere in the tree, not just at the root."""
+    node = stmt.find(*_MUTATING_STATEMENTS)
+    if node is not None:
+        logger.info(
+            "sandbox.nested_mutation", sql=sql, statement_type=type(node).__name__
+        )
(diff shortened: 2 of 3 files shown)


ishiland:
@codex review

ishiland:
Rebased onto `main` (now at `c081fc6b1`). One conflict, in `validator.py`, and it was positional rather than semantic: main added the #1001 canonical-geodesic-buffer exemption block at the same point this branch adds `_MUTATING_STATEMENTS` and `_reject_nested_mutation`. Both are kept, and both stay wired into `validate_sql`.

The post-rebase diff is byte-identical to the pre-rebase one: 3 files, 189 insertions, 1 deletion. Nothing was dropped or absorbed.

Verified whole-tree rather than on the touched file, since a rebase can produce defects present in neither branch: `ruff check`, `ruff format --check`, `test_layering.py`. The sandbox suite runs in CI.

ishiland:
@codex review

PR #1027, opened 2026-07-31
ops(backup): capture pg_dumpall --globals-only in the backup cycle

### LINKED ISSUE #995: ops(backup): capture pg_dumpall --globals-only in the backup cycle
Follow-up from #950's PR review (codex on PR #965).

A database-only pg_dump can never restore cluster roles, and the RUNBOOK's migration-based reconstruction fallback cannot serve every multi-tenant dump: 0020/0021 downgrades rebuild GLOBAL uniqueness (datasets.table_name, collection names, OAuth subjects) that the current schema scopes per tenant, so a valid dump with cross-tenant name reuse makes `alembic downgrade 0016` refuse on data the schema permits.

RUNBOOK now documents `pg_dumpall --globals-only` as the preferred fresh-cluster path (`RUNBOOK.md:166-178`) and tells operators to add the artifact to their backup cycle by hand. The durable fix is for `scripts/backup-entrypoint.sh` to write a globals dump next to each database dump (tiny, cheap, prunes with the same retention), so the preferred path always has its input.

## Three things the implementation has to get right

**Write it under `umask 077`.** `pg_dumpall --globals-only` emits role password verifiers. Under the default 022 umask the file lands world-readable inside the backup volume and then, with offsite upload enabled, in the bucket. RUNBOOK already models the correct form at `:173-178`:

```
(umask 077; docker compose exec -T db \
  pg_dumpall --globals-only -U "$POSTGRES_USER" > globals.sql)
```

Shipping this without the umask turns a backup improvement into a credential-exposure regression, so it belongs in the same commit and not as a follow-up.

**A `pg_dumpall` failure fails the cycle.** The existing paths already disagree with each other deliberately, so pick the right precedent. A failed S3 upload sets `cycle_failed=1` (`scripts/backup-entrypoint.sh:157-160`) and the cycle returns non-zero after retention has run, which means `.last-success` is never touched and the freshness healthcheck goes unhealthy. A `tar` exit of 1 during the staging archive is downgraded to a warning (`:214-228`) because a partially-fuzzy staging archive is still worth keeping. Globals belong with S3, not with tar: an unrestorable-roles backup should not be reported as a good cycle. Follow the `cycle_failed=1` shape so retention pruning still runs before the non-zero return.

**Copy the staging-tarball block's structure.** `backup_staging()` (`scripts/backup-entrypoint.sh:190-237`) is already exactly this pattern: a second artifact named with the **same timestamp** as the dump so restore can pair them, written into `DAILY_DIR`, copied into `WEEKLY_DIR` on Sundays, its path printed on stdout while all logging goes to stderr, and pruned by a matching `prune_old_*` helper wired into the same retention counts (`:166-169`). A globals artifact should be the same function shape with `pg_dumpall` in place of `tar`, and needs its own prune helper so the retention counts cover it the way they cover the staging archives.

## Acceptance

- Each backup cycle writes a globals artifact alongside the dump, with the dump's timestamp, pruned by the same daily/weekly retention.
- The file is created under `umask 077`; a test or an assertion confirms the mode, since this is the part that silently regresses.
- A `pg_dumpall` failure fails the cycle (non-zero, `.last-success` untouched) rather than logging and continuing.
- `restore.sh` (or RUNBOOK) points at the artifact for fresh-cluster restores.
- The RUNBOOK multi-tenant section drops the "consider adding one by hand" hedge.

## Test surfaces

`scripts/tests/test-backup-restore-roundtrip.sh` is the fast leg and the place to assert that a cycle produced the globals artifact with the right mode and the right timestamp pairing. It is driven in CI by the `backup-roundtrip` job (`.github/workflows/ci.yml:375`, invoked at `:477`), which also runs on every push, so a regression here fails visibly rather than waiting for someone to try a restore.

ishiland:
A database-only `pg_dump` can never restore cluster roles, so the fresh-cluster path in RUNBOOK § 2 had no input unless an operator remembered to capture a globals dump by hand. The RUNBOOK told them to add one to their backup cycle themselves; now the cycle does it.

## What changed

`scripts/backup-entrypoint.sh` gains `backup_globals()`, shaped like the existing `backup_staging()`:

- writes `globals-<timestamp>.sql` into `DAILY_DIR` with the **dump's** timestamp, so `restore.sh` can pair them by filename
- copies to `WEEKLY_DIR` on Sundays, prints its path on stdout with all logging on stderr
- uploads offsite alongside the dump it belongs to (a globals artifact that only exists on the disk you just lost is no use)
- prunes via `prune_old_globals()` on the same daily/weekly retention counts

Two properties the issue called out, and why they are what they are:

**`umask 077`.** `pg_dumpall --globals-only` emits role password verifiers. Under the default 022 umask the file lands world-readable in the backup volume and, with `BACKUP_S3_ENABLED=true`, in the bucket. The subshell scopes the umask to the redirect that creates the file — the form RUNBOOK already documented.

**A `pg_dumpall` failure fails the cycle.** It follows the S3 path (`cycle_failed=1`, so retention still prunes, the cycle returns non-zero, and `.last-success` is never touched) rather than `tar`'s downgrade-to-warning. `tar` exit 1 is downgraded because a partially-fuzzy staging archive is still worth keeping; there is no equivalent partial success here, since a truncated globals file replays as a partial set of roles.

## restore.sh

The paired globals dump is now reported **before** the destructive `--clean`, not after — replaying globals is a pre-restore step. It lists any role the globals dump defines that the target cluster lacks, so the message is actionable instead of unconditional noise (on a same-cluster restore it prints one line confirming every role is present). Non-fatal by design: it is a warning, not a gate.

Its dump-timestamp parse moved above the restore (the staging block reuses it) and is now anchored at the end of the filename. The old leftmost `grep -oE '[0-9]{8}_[0-9]{6}' | head -n1` matched any 8-digit_6-digit run inside `$POSTGRES_DB` and sent every sibling lookup down the unpaired-fallback path. This was a latent bug that the new globals pairing made load-bearing — the test caught it immediately, because the throwaway database names contain exactly that shape.

## Test

`BACKUP_DIR` gained a `${BACKUP_DIR:-/backups}` default, the same shape `STAGING_DIR` already had, so `scripts/tests/test-backup-restore-roundtrip.sh` can drive a real `--run-backup` cycle into a temp directory rather than mirroring one. Nothing in either compose file sets it, so every deployment still gets `/backups`. Running the real script matters for this specific property: a mirrored copy of the `pg_dumpall` call would assert the umask of the copy.

The new leg (step 5/5) asserts:

- a `globals-<ts>.sql` exists carrying the **dump's** timestamp, is non-empty, and contains `CREATE ROLE`
- its mode is `-rw-------`, checked with `ls -l` rather than `stat` (whose flags differ between BSD and GNU), and under a deliberately permissive ambient `umask 022` so a dropped `umask 077` cannot be masked by the caller's environment
- with `pg_dumpall` shadowed by a failing stub on `PATH`, the cycle returns non-zero, logs why, leaves `.last-success` at its previous value, and leaves no empty artifact behind

## Security impact

The new artifact contains role password verifiers. It is created mode 0600, the weekly copy re-asserts 0600, and it inherits the dump's offsite handling. RUNBOOK now says so where the artifact is introduced.

## Verification

```
bash scripts/tests/test-backup-restore-roundtrip.sh    # PASS, incl. new [5/5]
python3 scripts/check_env_doc_drift.py                 # OK
python3 scripts/check_docs_contract.py                 # OK
bash -n on all three changed scripts
```

Negative check, by hand: deleting `umask 077;` from `backup_globals` makes the new leg fail with `globals dump is -rw-r--r--, expected -rw-------`.

All four changed files are in the `backup` path filter, so the `backup-roundtrip` CI job runs on this PR and exercises `restore.sh`'s new pre-restore block against a real cycle's output.

Closes #995

### DIFF
--- a/RUNBOOK.md
+++ b/RUNBOOK.md
@@ -78,12 +78,20 @@ Artifacts land at:
 
 Default retention: 7 daily, 4 weekly (set `BACKUP_RETENTION_DAILY` /
 `BACKUP_RETENTION_WEEKLY` in `.env` to override). The count applies to the
-`.dump` files; the paired `staging-*.tar.gz` and `globals-*.sql` artifacts are
-pruned when the dump they belong with ages out. Retention therefore evicts whole
-backup sets, and a companion artifact never outlives its dump — which matters
-because a cycle can produce a dump without a companion (an empty staging volume,
-or a failed `pg_dumpall`), and counting each kind separately would then prune
-complete sets while leaving orphans behind.
+`.dump` files, ordered by the timestamp in the filename; the paired
+`staging-*.tar.gz` and `globals-*.sql` artifacts are pruned when the dump they
+belong with ages out. Retention therefore evicts whole backup sets, and a
+companion artifact never outlives its dump — which matters because a cycle can
+produce a dump without a companion (an empty staging volume, or a failed
+`pg_dumpall`), and counting each kind separately would prune complete sets while
+leaving orphans behind.
+
+One exception: the **newest complete set** — a dump that still has its globals
+dump — is held back on top of the window, so a directory can hold `keep` + 1
+dumps. A run of `pg_dumpall` failures still produces valid database dumps each
+cycle, and without the exemption those would walk the last restorable-onto-a-
+fresh-cluster set out of retention while the healthcheck was already reporting
+the problem. The cost is one extra dump plus a few KB of SQL.
 
 ### Offsite (S3) upload
 
--- a/RUNBOOK.md
+++ b/RUNBOOK.md
@@ -64,7 +64,7 @@ The staging archive is omitted silently when the `upload_staging` volume is abse
 or empty (fresh install with no uploaded datasets).
 
 The globals dump is the piece a database-only `pg_dump` can never give back, and
-it is what makes the fresh-cluster restore in [§ 2](#multi-tenant-role-reconstruction-after-a-fresh-cluster-restore)
+it is what makes the fresh-cluster role reconstruction in [§ 2](#2-restore--bundled-postgres-mode)
 work. It contains **role password verifiers**, so it is written mode `0600` and
 must keep the same protection as the dump itself wherever it is copied. A
 `pg_dumpall` failure fails the whole cycle: a backup set that cannot restore its
@@ -456,8 +456,8 @@ directory as the dump — here `./restore`). For the staging archive it prints t
 `docker run` line above with the real paths filled in; copy the printed command
 from the restore output rather than hand-editing it. For the globals dump it
 reports, **before** anything destructive runs, which roles the target cluster is
-missing — replay it first if that list is non-empty (see
-[§ multi-tenant role reconstruction](#multi-tenant-role-reconstruction-after-a-fresh-cluster-restore)).
+missing — replay it first if that list is non-empty, following the role
+reconstruction procedure earlier in this section.
 
 ### Finding the dump to restore
 
--- a/RUNBOOK.md
+++ b/RUNBOOK.md
@@ -842,14 +842,20 @@ Determine what was lost:
 # List all daily dumps in the backup_data volume, newest first
 docker compose exec backup sh -c 'ls -lt /backups/daily/*.dump'
 
-# Copy the chosen dump (and its paired staging archive) out of the volume.
+# Copy the chosen dump and BOTH its paired artifacts out of the volume — the
+# globals dump included, or a fresh-cluster restore has no roles to replay and
+# restore.sh cannot report which ones are missing.
 # Replace <project> with your Compose project name (see `docker volume ls`).
-mkdir -p ./restore
+# umask 077: globals-*.sql holds role password verifiers.
+mkdir -p ./restore && chmod 700 ./restore
 docker run --rm \
   -v <project>_backup_data:/backups:ro \
   -v "$(pwd)/restore":/out \
-  alpine sh -c 'cp /backups/daily/geolens_<YYYYmmdd_HHMMSS>.dump /out/ && \
-                cp /backups/daily/staging-<YYYYmmdd_HHMMSS>.tar.gz /out/ 2>/dev/null; ls -lh /out'
+  alpine sh -c 'umask 077; \
+                cp /backups/daily/geolens_<YYYYmmdd_HHMMSS>.dump /out/ && \
+                cp /backups/daily/staging-<YYYYmmdd_HHMMSS>.tar.gz /out/ 2>/dev/null; \
+                cp /backups/daily/globals-<YYYYmmdd_HHMMSS>.sql /out/ 2>/dev/null; \
+                ls -lh /out'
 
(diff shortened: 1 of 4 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
Dropped from integration batch #1086 after it went red. The failure is real but **your PR's own CI structurally could not have caught it**, so filing the mechanism rather than just the bug: #1088.

```
2 failed, 6659 passed, 83 skipped
```

`backend/tests/test_backup_staging_tar_skew.py` (on main since #844) extracts `backup_staging` from `scripts/backup-entrypoint.sh` and runs that function **alone** under `set -euo pipefail`. Your `CYCLE_IS_WEEKLY` is set by the cycle function and read inside `backup_staging`, which is correct in production, where bash globals reach callees. In the extracted context the caller never runs, so `-u` makes the read fatal.

**Why it was green here:** `Backend Tests -> SKIPPED` on this PR. It touches only `scripts/` and `RUNBOOK.md`, which does not trip the `backend` path filter — and the filter's premise is false for this one file, because a backend test reads it.

Smallest fix is `${CYCLE_IS_WEEKLY:-0}` at the read sites inside `backup_staging`, defaulting to the non-weekly path. A function its own test runs in isolation should not depend on a global its caller sets. #1088 also proposes adding the script to the `backend` filter, which is what prevents the next instance.

Reproduce without waiting for a batch:

```
cd backend && uv run pytest tests/test_backup_staging_tar_skew.py -q
```

Re-batchable as soon as that passes.

ishiland:
Pushed `a1c033382` to this branch: a one-line default for `CYCLE_IS_WEEKLY` at the read inside `backup_staging()`.

Context, since this PR looked ready and was not. Its Backend Tests job was **skipped**, not passed. The `backend` paths-filter in `ci.yml` had no `scripts/` entries, so changing `scripts/backup-entrypoint.sh` triggered Backup Restore Round-trip (which passed) but never pulled in the backend suite. `test_backup_staging_tar_skew.py` extracts `backup_staging()` on its own and runs it under `set -euo pipefail`, where the caller's assignment does not exist:

```
bash: line 42: CYCLE_IS_WEEKLY: unbound variable
```

Reproduced at `cdb9e9cd8` (2 failed, 1 passed), then 3 passed with the default applied. `bash -n` clean.

Only the read at line 267 is inside the function. The three in the caller sit after the assignment and are untouched. The assignment stays in the caller on purpose, so a cycle running across midnight into Sunday cannot switch modes partway through.

The filter gap is #1088, fixed in #1090. Once that lands, update-branch this PR and Backend Tests will actually run and report on the change.

Issue #958, opened 2026-07-30
test(backend): LOC-ratchet coverage is inconsistent — analysis_sql.py is ungated and grew 504 to 522 unnoticed


ishiland:
Lifted out of #893 (item 10), where it was filed as a nit. It has since demonstrated its own thesis, so it gets its own issue.

## The gap

`_MODULE_LOC_CAPS` in `backend/tests/test_layering.py:1025-1111` ratchets nine modules with exact-equal caps, and a separate glob gate scans `**/router.py`. `backend/app/platform/analysis_sql.py` is in neither: it is not a router, so the glob misses it, and nobody added it to the dict.

When this was written it was 504 lines. As of 2026-07-31 it is **651**. That is 147 lines through an ungated file, which is what the ratchet exists to prevent, and a reasonable demonstration that "it isn't listed" is not the same as "it doesn't need listing".

## The inconsistency worth deciding on

The ratchet's exact-equal design means any change to a listed file requires raising or lowering its cap in the same PR. `test_module_loc_caps_have_no_headroom` fails on shrinkage too:

```python
verb = "shrank below" if actual < cap else "exceeds"
```

That makes it a "document your growth" gate rather than a ceiling, a defensible design, and the carve-out comments in the dict are genuinely useful history. But it should be applied consistently. Right now `ingest/metadata.py` at 2002 lines cannot gain a line without a reviewed comment, while a 651-line one can grow freely.

Two things to settle:

1. **Coverage.** Either add `analysis_sql.py` (and audit for other non-router modules in the same position), or define the inclusion rule explicitly so its absence is a decision rather than an oversight. The table below is the data that decision should be made against.
2. **`metadata.py` is a decomposition candidate at 2002 lines** regardless of the ratchet. The cap comments on it are already four carve-outs deep (#888, #899 codex r1, #906, #934). Each fix now pays a cap-raise tax, which is friction pointed at the wrong thing.

## Audit: what "ungated" actually covers, measured 2026-07-31

Three separate mechanisms in `test_layering.py` cap module size, with different semantics, and a module is gated if any one of them names it:

- `_MODULE_LOC_CAPS` (:1025-1111), enforced by `test_module_loc_caps_have_no_headroom` (:1115), exact-equal, nine entries.
- The `**/router.py` glob in `test_router_orchestrator_modules_stay_within_loc_cap` (:1201), default ceiling 1500, overridable from `_MODULE_LOC_CAPS`.
[...]

PR #1028, opened 2026-07-31
docs: correct five comments that described something the code does not do

### LINKED ISSUE #1018: chore(docs): five comments and docstrings that describe something the code does not do
Decomposed from #893, which grouped these under "a comment, docstring, or helper that describes something the code does not do". They touch five files, they have no behaviour change between them apart from one line reorder in a spec, and they are all cheap. One PR closes the set. Each reference re-verified against `main` at `9a788ba92` on 2026-07-31.

## 1. `test_export_invalid_bbox_bounds` states the reason backwards (#893 item 9a)

`backend/tests/test_export.py:302` reads:

```python
"""Request with bbox=10,10,5,5 (minx > maxx) returns 400."""
```

`parse_bbox` (`backend/app/modules/catalog/features/service.py:96`) explicitly allows that. Its docstring says *"Allows antimeridian-crossing bboxes where minx > maxx (e.g. 170,-45,-170,-30)"*, and the only bounds check is at `:124`:

```python
# Only validate latitude (lon wraps at antimeridian)
if values[1] >= values[3]:
    raise ValueError("bbox miny must be less than maxy")
```

So the 400 comes from `miny (10) >= maxy (5)`, not from the longitude pair. The assertion is correct; the stated reason is exactly backwards, and it is backwards in front of the antimeridian code, which is the code most likely to read it. Fix the docstring to name the latitude check.

## 2. `_ingest_vector_into_staging` has no production caller (#893 item 11)

Defined at `backend/app/processing/ingest/tasks_common.py:810` and re-exported through `tasks.py:20`. Every call site in the repo is a test: `backend/tests/test_staging_pipeline.py` and `backend/tests/test_staging_pipeline_integration.py`. Production ingest goes through `_finalize_ingest`.

Decide and record the decision, either is fine:

- Delete it and fold the two test files onto whatever path production actually uses, or
- Keep it and mark it test-only in its docstring, so the next reader does not assume it is live and keep it in step with a path it does not share.

What should not survive this issue is a public-looking helper with no production caller and nothing saying so.

## 3. `AnalysisJobWatcher`'s comment overstates the guarantee (#893 item 14)

`frontend/src/App.tsx:59-62` says a job that outlives the builder still reports through a "tab reloaded" path. That holds only for a job still *running* at reload time. A job that completed roughly 400 ms before a hard reload is unrecoverable, because `setJob(null)` is already persisted.

Low severity, since no real user hard-reloads 400 ms after clicking Create. The comment simply reads as a stronger guarantee than the code gives. Narrow it to the running-at-reload case.

## 4. A spec assigns its cleanup handle after the assertion (#893 item 12)

`e2e/collections.spec.ts:192` sets `createdDatasetId = dataset.datasetId` after the `Add Datasets` heading assertion at `:187-189`. If that assertion fails, the dataset is created and never cleaned up, so each failed retry leaks one row. `e2e/analysis.spec.ts` had the same shape and was fixed in #895; the catalog count there was visibly climbing 3, 4, 5 across three retries.

**The sweep is done: `collections.spec.ts:192` is the only remaining site.** Move the assignment above the assertion. This is the one line in this issue that changes behaviour.

## 5. The frontend coverage comment records 2026-05 actuals (#893 item 4)

`frontend/vite.config.ts:242-253` records `2026-05-07 actuals: statements 41.51 / branches 39.42 / functions 37.99 / lines 42.69` and warns that a +1 buffer would have failed because actuals sit under a point above their integer floors.

Both halves of that are now wrong. Measured on `main` at `9a788ba92` with `npm run test:coverage`:

| Dimension | Threshold | Actual |
|---|---|---|
| statements | 41 | 72.08 |
| branches | 39 | 67.35 |
| functions | 37 | 66.40 |
| lines | 42 | 74.19 |

That is roughly 30 points of headroom on every dimension, not zero. The staleness is the whole problem: someone planning coverage work reads this and believes any uncovered line will trip a global threshold.

Correct the comment to the current actuals and **drop the zero-headroom warning**, which no longer describes anything. Whether to also ratchet the thresholds up to match is a separate call and should not be smuggled into this PR; if you do ratchet, the comment gets rewritten once rather than twice.

## Verification

```
cd backend && uv run pytest tests/test_export.py -v
npx playwright test e2e/collections.spec.ts --project=chromium
```

ishiland:
Five stale or wrong comments, one PR, as #1018 asks. Three are comment-only; two change behaviour (`collections.spec.ts` cleanup, and a `_MODULE_LOC_CAPS` bump the docstring growth requires).

## 1. `backend/tests/test_export.py` — the reason was backwards

The docstring said the 400 on `bbox=10,10,5,5` came from `minx > maxx`. `parse_bbox` deliberately **allows** that — it is how an antimeridian-crossing bbox is spelled (`170,-45,-170,-30`), and its only bounds check is `values[1] >= values[3]`. The assertion was right, the stated reason exactly inverted, and it sat in front of the antimeridian code most likely to be read by someone acting on it. Now names the latitude check.

## 2. `_ingest_vector_into_staging` — kept, marked test-only

`backend/app/processing/ingest/tasks_common.py:810`, re-exported through `tasks.py:20`. Every call site in the repo is a test (`test_staging_pipeline.py`, `test_staging_pipeline_integration.py`).

**Decision: keep it, mark it.** Deleting means rewriting ~900 lines of tests onto `_finalize_ingest`, a far wider blast radius than the problem justifies.

The docstring now maps what it mirrors, and review corrected that map:

- Its **pre-staging** steps (`run_ogr2ogr` → `rename_reserved_columns` → DBF-truncation check → `user_wants_geom` override) are a parallel copy of `tasks_vector.py:459-560` and `tasks_reupload.py:261-340`.
- Its **staging** steps are not forked at all — they are the real `_run_staging_pipeline` (`:650`). But production reaches that function *only* on the re-upload path (`tasks_reupload.py:337`); new vector ingest reruns the same eight steps inline in `_finalize_ingest` (`:1069`, block at `:1114-1177`). **Three sites, not two.**

The same correction fixes `_run_staging_pipeline`'s own docstring, which claimed `_ingest_vector_into_staging` was its "new ingests" caller — the same class of defect this issue collects, one function up.

## 3. `frontend/src/App.tsx` — narrowed to the guarantee the code gives

The `AnalysisJobWatcher` host comment listed "tab reloaded" alongside panel-closed and navigated-away. That holds only while the job is still running, which is what the persisted store carries. After a terminal status the watcher has already persisted `setJob(null)`. Comment now says so; no code change, since it takes a reload inside a few hundred ms.

## 4. `e2e/collections.spec.ts` — the leak is real, one frame further in than the issue says

The issue pins this at `:192`, "the assignment sits after the `Add Datasets` heading assertion". On `9a788ba92` that assertion is at `:186-189` and `createCollectionDataset()` at `:191` — creation happens *after* the assertion, so that path cannot leak.

The leak it describes is real, one frame deeper: `createCollectionDataset` commits the ingest and only then awaits `waitForDatasetJob`. A failed import, a 45-attempt timeout, or a bad response mid-poll all throw with `createdDatasetId` still null, and `afterAll` returns early having deleted nothing — one leaked dataset per retry, the shape #895 fixed in `analysis.spec.ts` (its catalog count climbed 3 → 4 → 5 across three retries).

Fixed after two review rounds, and the second round is the interesting one: **there is no earlier moment to claim the id.** `_finalize_ingest` sets `dataset_id` in the *same atomic update* that flips the job to `complete`, so "claim it as soon as a poll reports one" — my first attempt — was a no-op. What survives:

- `createdDatasetTitle` registered before the upload can create anything (`afterAll` needs both halves of the handle).
- `createdJobId` registered at commit — the only handle that exists in the gap.
- Teardown polls that job for the id, **bounded** at 30 × 2s. A single lookup would almost always still see `running`, because the reason we are in that branch is that the job outlived its 45-second window; and the bound matters as much as the wait, since teardown must not hang on a wedged worker. Stops early on a failed job, which produces no dataset to delete.

## 5. `frontend/vite.config.ts` — coverage actuals were 12 weeks stale

Recorded `2026-05-07 actuals: statements 41.51 / branches 39.42 / functions 37.99 / lines 42.69` plus a warning that even a +1 threshold buffer would fail for lack of headroom.

Measured with `npm run test:coverage`, 2026-07-31:

| Dimension | Threshold | Actual |
|---|---|---|
| statements | 41 | 72.05 |
| branches | 39 | 67.31 |
| functions | 37 | 66.29 |
| lines | 42 | 74.16 |

~30 points of headroom, not zero, so the warning described nothing and is dropped. **Thresholds are unchanged** — ratcheting is the separate call the issue asks not to smuggle in.

Review caught that correcting only the numbers left the block self-contradictory: the surviving prose said every threshold is `floor(actual_pct)` from the most recent run, while the recorded actuals now sit 30 points above them. The invariant is reworded to what is actually true — "thresholds are a floor from a recorded measurement" — with the ratchet recipe kept.

## The cap bump

Item 2's docstring costs 23 lines in a file ratcheted exact, so `_MODULE_LOC_CAPS["...tasks_common.py"]` goes 1631 → 1654 with the carve-out comment the ratchet asks for. Caught by the full backend suite, not by a focused run — which is #958's other half.

None.

## Verification

```
cd backend && uv run pytest tests/test_layering.py tests/test_export.py tests/test_staging_pipeline.py \
  tests/test_staging_pipeline_integration.py tests/test_event_ingest.py -q     # 107 passed
cd frontend && npm run lint && npm run typecheck && npm run test:coverage      # 366 files / 4284 tests
npx playwright test e2e/collections.spec.ts --project=chromium                 # 9 passed (re-run after each spec change)
```

The Playwright run is valid from this branch: the spec is the only behavioural change that the running stack could see, and Playwright reads specs from the working tree while the app code the stack serves is byte-identical to `main` (the other edits are comments and a test-file cap).

Closes #1018

### DIFF
--- a/backend/app/processing/ingest/tasks_common.py
+++ b/backend/app/processing/ingest/tasks_common.py
@@ -824,8 +824,19 @@ async def _ingest_vector_into_staging(
 ) -> StagingResult:
     """Load a vector source into staging and return extracted staging metadata.
 
-    This helper preserves the staging-pipeline unit-test boundary while the
-    production ingest path continues to inline the broader job lifecycle.
+    TEST-ONLY (#1018). Nothing in ``app/`` calls this. Every caller is a test:
+    ``tests/test_staging_pipeline.py`` and
+    ``tests/test_staging_pipeline_integration.py``. It exists to give those
+    tests a callable seam over the staging half of ingest, which production
+    inlines inside its own job lifecycle rather than factoring out.
+
+    So it is a PARALLEL COPY, not the live path, and it can drift. The two
+    production sequences it mirrors are ``tasks_vector.py`` (run_ogr2ogr ->
+    rename_reserved_columns -> DBF-truncation check -> geometry override ->
+    _run_staging_pipeline, :459-560) and ``tasks_reupload.py`` (:261-340).
+    A change to either has to visit here, or these tests start asserting
+    against a shape production no longer has.
+
     It intentionally performs no commits.
     """
     from app.processing.ingest.metadata import rename_reserved_columns
--- a/backend/app/processing/ingest/tasks_common.py
+++ b/backend/app/processing/ingest/tasks_common.py
@@ -836,31 +836,35 @@ async def _ingest_vector_into_staging(
     tests a callable seam over vector ingest's pre-staging half, which
     production runs inline inside its own job lifecycle.
 
-    What it mirrors, and what it shares (fix(#1018 review)):
+    What it mirrors, and what it shares (fix(#1018 review)). Deliberately no
+    line numbers: this docstring's own growth moved them twice while it was
+    being written, so it names FUNCTIONS, which do not drift.
 
-    - The pre-staging steps here — run_ogr2ogr, rename_reserved_columns, the
-      DBF-truncation check, the ``user_wants_geom`` override — are a PARALLEL
-      COPY of ``tasks_vector.py`` (:459-560) and ``tasks_reupload.py``
-      (:261-340). Those can drift from this, and only the tests would notice.
-    - The staging steps are the real ``_run_staging_pipeline`` (:650), so this
-      helper does not fork them — but almost nothing else calls it either.
-      Where the sequence is written out, and how much of it:
+    - Pre-staging. ``run_ogr2ogr``, ``rename_reserved_columns``, the
+      DBF-truncation check, then ``_detect_and_override_geometry`` under
+      ``user_wants_geom``. ``tasks_vector.ingest_file`` runs the same four —
+      it is the only production path with the override, so this helper's
+      ``user_wants_geom`` branch has exactly one counterpart.
+      ``tasks_reupload.reupload_file`` runs the first three and passes its
+      detected geometry type straight to ``run_ogr2ogr`` instead
+      (fix(#1018 review): an earlier draft claimed all four).
+    - Staging. This helper calls the real ``_run_staging_pipeline``, so it does
+      not fork those steps — but little else calls it either. The sequence
+      exists in three independent places:
 
-        1. ``_run_staging_pipeline`` (:650) — the full eight steps
+        1. ``_run_staging_pipeline`` — the full eight steps
            (ensure_geom_column, clip_to_mercator_bounds, add_4326_column,
            grant_reader_access, extract_metadata, detect_3d_metadata,
            promote_z_to_elev, get_sample_values). Reached in production only by
-           ``reupload_file`` (``tasks_reupload.py:337``), and by this helper.
-        2. ``_finalize_ingest`` (:1069, block at :1114-1177) — the same eight,
-           inlined. New vector ingest, via ``tasks_vector.py:572``.
-        3. ``reupload_service`` (``tasks_reupload.py:463``, block at :670-690)
-           — a SHORTER copy: ensure_geom_column, clip, add_4326_column,
-           grant_reader_access, extract_metadata, get_sample_values. No 3D
-           detection and no elevation promotion, so do not "fix" it by
-           symmetry with the other two; find out why first (fix(#1018 review)).
+           ``tasks_reupload.reupload_file``, and by this helper.
+        2. ``_finalize_ingest`` in this module — the same eight, inlined. New
+           vector ingest, via ``tasks_vector.ingest_file``.
+        3. ``tasks_reupload.reupload_service`` — a SHORTER copy: no 3D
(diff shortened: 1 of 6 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

Issue #940, opened 2026-07-30
epic(catalog): dataset visibility ladder — private/internal/public


ishiland:
Tracking issue for the dataset visibility work split across #929, #930, #927, and #931. Those four issues touch the same three frontend files, the same type union, and the same help-text string, so they are sequenced here rather than picked up independently.

## The ladder

Dataset visibility becomes **private / internal / public**, matching what maps already do.

- `private`: the owner and admins.
- `internal`: any signed-in user, on a published record. Implemented to match map semantics (`MapVisibility`, `backend/app/modules/catalog/maps/schemas.py:253-258`).
- `public`: everyone, including signed-out visitors.

`restricted` is **not** part of the ladder. It stays a valid stored value for deployments that manage `catalog.dataset_grants` in SQL, it still renders wherever visibility is displayed, but it is removed from every picker. See #929 for the decision and its grounds.

## Order of work

The sequence is load-bearing. Taking these out of order ships a state where a user's dataset disappears from their own list.

| Step | Issue | Scope | Status |
|---|---|---|---|
| 1 | #929 | Creator exemption in the permission layer, plus removing `restricted` from the two import pickers. | **Done**, shipped in #970 |
| 2 | #927 | The Access-tab visibility control, shipping private and public only. | **Done**, shipped in #976 |
| 3 | #930 | Implement `internal` in the permission layer, widen the frontend `DatasetVisibility` union, promote the option on the Access tab. | Next |
| 4 | #931 | Map-impact surfacing (frontend). | After #930 |

Step 2 shipped deliberately without `internal`. `SELECTABLE_VISIBILITIES` at `frontend/src/components/dataset/tabs/AccessTab.tsx:51` is `['private', 'public']`, and until #930's permission branches merge, offering `internal` would write a value that makes the dataset invisible in every list, including its owner's. Promoting the option is #930's job, not a leftover from #927.

#931 needs #927's control to hang off, which now exists, and needs #930 for the internal rung. #931 was rewritten after the guard it originally asked for turned out to already exist.

## Consolidated i18n surface

Stated once here so it is not re-derived in four places. Every key below lands in all four locales (en/es/fr/de) or `npm run test:i18n` fails the build.

| Key | Change | Owning issue |
|---|---|---|
[...]

PR #1029, opened 2026-07-31
fix(catalog): implement internal dataset visibility across the permission layer

### LINKED ISSUE #930: Dataset visibility 'internal' is accepted but unimplemented: hidden from lists, readable by any signed-in user by ID

ishiland:
Step 3 of the #940 visibility ladder. Closes #930.

## What was broken

`internal` was accepted by the dataset `PATCH` Literal and written by the CLI
manifest path (`geolens apply` intent `internal`), but the permission layer only
knew public / private / restricted:

- `filter_visible` built no condition matching it, so an internal dataset was
  invisible to every non-admin **including its owner**.
- `can_access_dataset` fell through to `return True`, so a `(internal, published)`
  dataset was readable by direct ID by any signed-in user.

Hidden from lists, readable by link, and nothing pinned either half.

## What changed

Implemented with map parity — **internal = any signed-in user, on a published
record** — per the decision recorded on the issue.

Backend
- `authorization.py`: `DatasetVisibility.INTERNAL`.
- `defaults_extensions.py`: an `internal` condition in `filter_visible`, and an
  explicit `internal` branch in `can_access_dataset` placed ahead of the
  fall-through so the outcome is policy rather than an unhandled value.

Frontend
- `types/api.ts`: widened the hand-maintained `DatasetVisibility` union.
- `AccessTab.tsx`: `internal` joins `SELECTABLE_VISIBILITIES` between private
  and public. A stored `internal` value stops rendering through the legacy
  branch. The import pickers (`ImportMetadataForm`, `RegisterForm`) deliberately
  stay at private/public — `RegisterForm` declares its own local union and is
  untouched by the widening.
- `metadataEdit.visibilityHelp` gains the internal rung in all four locales,
  wording adapted from the maps share dialog ("anyone signed in to GeoLens").

Cleanup (the open #940 task, riding along as planned)
- Deleted the dead `register.visibilityHelp` and both `visibilityRestricted`
  keys from `import.json` in all four locales. The pickers stopped offering
  `restricted` in #929 and no component references those keys; the only rendered
  visibility help is `metadataEdit.visibilityHelp` on the Access tab.

No new `t()` key was needed. The issue expected `metadataEdit.visibilityInternal`,
but the select labels resolve through `getVisibilityLabel` →
`common:enums.visibility.internal`, which already exists in all four locales. Its
`metadataEdit.visibility{Public,Private,Restricted}` siblings in `dataset.json`
are themselves unreferenced, so adding a fifth dead key would have contradicted
the deletion above.

## One correction to the issue's step 2

The body specified the `filter_visible` condition as
`and_(visibility == "internal", record_status == "published")`, with an inline
comment asserting that without the inner check "an owner's draft internal
dataset leaks to every signed-in user".

That is not what happens, and I verified it rather than reasoning about it.
`status_filter` is ANDed over every condition and is
`record_status == "published" OR created_by == <the caller>` — the creator
clause tests the **caller**, not the record's owner. So another user's draft is
already hidden from the team. What the inner check actually does is hide an
owner's own draft from the owner: it reintroduces exactly the list/detail split
this issue closes, and repeats the #929 creator-exemption bug.

Measured, with the issue-literal condition in place:

```
test_internal_draft_still_visible_to_its_own_owner FAILED
  assert '2a17b9b0-…' in ['890c2229-…', 'c2db8497-…']
```

`GET /datasets/{id}` returned 200 for that same dataset. Private and public
drafts stay visible to their owner today, so internal must match.

Shipped instead as a bare `record_cls.visibility == "internal"`, the same shape
as the `public` condition directly above it and for the same reason, with a
comment recording why the stricter-looking variant is wrong so it is not
"fixed" later. Both behaviours are pinned by test: a non-owner's draft stays
hidden, an owner's own draft stays visible.

## Security impact

Widens read access for `visibility='internal'` datasets, which is the point of
the issue. Anonymous access is unchanged — internal is never visible signed-out,
and `_is_publicly_cacheable` in the tile router still requires public+published,
so internal tiles never enter the auth-less cache. Unpublished internal records
remain owner-and-admin-only via the existing `record_status` gate.

## Verification

```
  tests/test_datasets.py tests/test_permission_extension.py \
  tests/test_permissions.py tests/test_permission_overlay.py -q
  # 64 passed, 5 skipped

cd frontend && npm run test:coverage   # 366 files, 4286 tests passed
```

New backend coverage is `TestInternalVisibility` in `backend/tests/test_datasets.py`:
`(internal, published)` visible to a signed-in non-owner on both list and detail,
visible to a non-admin owner, `(internal, internal)` (the `geolens apply` state)
hidden from a non-owner on both, `(internal, draft)` hidden from a non-owner but
visible to its owner, and internal hidden from anonymous.


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

Issue #1015, opened 2026-07-31
ops(analysis): per-tenant cap on active materialize jobs


ishiland:
Split out of #696 (item 5, which arrived there from #695 recommendation 3). It was sequenced behind #691. **#691 has shipped, so this is unblocked.**

## The gap

The materialize admission cap is per user only. `create_analysis` counts active analysis jobs filtered on `IngestJob.created_by == user.id` (`backend/app/modules/catalog/datasets/api/router_analysis.py:295-315`) and returns 429 when one exists. A tenant with N users can therefore hold N concurrent CTASes, each an unbounded-ish statement, regardless of what the host can absorb.

That matters more now than it did when the item was filed, because the sibling split on scoped `work_mem` wants to raise the per-statement memory ceiling. A raised ceiling times an unbounded concurrent count is the same outage in a new place.

## The predicate to reuse

#691 replaced status-only counting with a heartbeat lease, and the new cap must share that rule rather than inventing a second one. The live predicate (`router_analysis.py:292-315`) is:

```python
lease_cutoff = datetime.now(timezone.utc) - timedelta(seconds=MATERIALIZE_LEASE_SECONDS)
...
IngestJob.user_metadata.has_key("analysis"),
or_(
    IngestJob.status == "pending",
    and_(
        IngestJob.status == "running",
        func.coalesce(IngestJob.heartbeat_at, IngestJob.started_at) >= lease_cutoff,
    ),
),
```

`MATERIALIZE_LEASE_SECONDS` is `ANALYSIS_MATERIALIZE_LEASE_SECONDS`, imported under an alias from `backend/app/platform/jobs/heartbeat.py` (`router_analysis.py:43`) and also used by the stale-job sweeper (`backend/app/platform/jobs/router.py:740`).

Three properties of that predicate are load-bearing and the tenant version must keep all three, because the comments at `:274-291` record why each exists:

1. **A running job whose lease has gone stale releases the slot.** A hard-killed worker (SIGKILL, OOM) would otherwise hold the slot until the 60-minute `JOB_TIMEOUT_SECONDS` backstop fails the row.
2. **The pending branch must stay status-only.** A pending job has never been claimed, so `heartbeat_at` and `started_at` are both NULL, and any cutoff comparison silently drops it from the count. That would let users stack unbounded queued CTASes while every test still passes.
[...]

PR #1030, opened 2026-07-31
test(analysis): close backend coverage gaps at the analysis seams

### LINKED ISSUE #789: test(analysis): close the backend analysis coverage gaps
From the coverage track of the 2026-07-27 audit — the seams, not the middle (the materialize suite is genuinely strong).

**Re-scoped 2026-07-30.** The original list mixed backend pytest, frontend Vitest, and Playwright work into one issue spanning three test runners, which is why it never got picked up whole. This issue now keeps the **backend pytest** items only:

- Frontend Vitest gaps → #943
- Playwright gap (dissolve + drawn-mask clip) → #945

Line references re-verified 2026-07-31 against `main` at `9a788ba92`; several had drifted.

## Items

1. **Centroid materialize semantics (P2).** `centroid` is the cheap lifecycle op throughout `tests/test_analysis_materialize.py` — the titles are `"First"`, `"Should be blocked"`, `"Not allowed"` and so on, all job-slot and permission scaffolding. Not one of those occurrences asserts its output geometry or carry-columns on the write path. Preview does assert them, which makes this easy to misread as covered. Mirror `test_buffer_materialize_creates_dataset` (`:733`). (Occurrence counts are deliberately not stated here; they have drifted twice already. Grep the file.)

2. **Sandbox error → status mapping (P2).** `_SANDBOX_STATUS` (`app/modules/catalog/datasets/api/router_analysis.py:77-81`, consumed at `:172`) is untested. Only `tests/test_sandbox.py:516` asserts the exception *category*; nothing asserts the HTTP mapping. `query_busy` → 429 is the routine one (shared advisory lock with AI chat).

   **How to patch it.** Make the service raise, then assert the status the route produces. The target is `router_analysis.run_analysis_preview` — the name the module imported, not the definition in `service_analysis` — patched exactly the way the existing suite patches the queue hop:

   ```python
   with patch.object(
       router_analysis, "run_analysis_preview", AsyncMock(side_effect=SandboxError(...))
   ):
   ```

   The precedent is `patch.object(router_analysis, "defer_async_with_tenant", AsyncMock())` at `tests/test_analysis_materialize.py:81-83`. Cover all four outcomes: `query_busy` → 429 with `user_message` in the detail, `query_timeout` → 422, `query_data_error` → 422, and an unrecognised category → 500 via the `.get` default.

3. **Job-slot fixture footgun.** `test_materialize_returns_job` (`tests/test_analysis_materialize.py:72`) hand-releases the per-user slot at `:116` with a "(shared DB)" comment. It is not alone — there are three such sites, `:116`, `:155`, and `:569`, each doing the same three lines: fetch the `IngestJob` by the id from the response, set `status = "failed"`, commit. Under `pytest -n 4` a forgetful new test 429s unrelated victims (distinct from #713).

   **Fixture shape.** An `autouse` teardown fixture in the analysis test modules that, after each test, fails every non-terminal analysis job belonging to the test user — a single `UPDATE` over `IngestJob` filtered to that user with a status not in the terminal set, then commit. Autouse rather than opt-in, because the failure mode is exactly a new test forgetting to opt in, and the fixture is a no-op when there is nothing to release. Land it, then delete the three hand-release blocks; the fixture retires all of them and the tests keep their assertions.

4. **One-liner:** assert `job.current_step == "queued"` at enqueue (`router_analysis.py:347`) — the #703 below-default priority makes the stamp the only honest "waiting" signal. Still gated on #695, which is open.

## Sequencing

#699's edge-case fixture matrix makes items 1 and 2 materially cheaper — it builds the POINT/LINE and mixed/NULL/empty geometry sources that a real output-geometry assertion wants. Worth landing that first rather than hand-rolling fixtures here.

---
Found by the 2026-07-27 builder/analysis audit.

ishiland:
Backend pytest items from #789 (the issue was re-scoped on 2026-07-30 to backend only; the Vitest and Playwright halves are #943 and #945).

## Item 1 — centroid materialize semantics

`centroid` is the cheap lifecycle op throughout `test_analysis_materialize.py`: the titles are "First", "Should be blocked", "Not allowed", all job-slot and permission scaffolding. None of them asserts what the write path produced. Preview *does* assert the output geometry, which is what makes this easy to misread as covered.

`test_centroid_materialize_creates_dataset` mirrors `test_buffer_materialize_creates_dataset` and pins the output points at (0.5, 0.5) and (10.5, 10.5) — the same values `test_centroid_preview` pins on the read path, so preview and the saved dataset are held to agreeing — plus `ST_Point` geometry, the catalog `geometry_type`, the 1:1 feature count, and the carried `name` column.

## Item 2 — sandbox error to HTTP status

`_SANDBOX_STATUS` (`router_analysis.py:77-81`, consumed at `:172`) had no test. `test_sandbox.py:516` asserts the exception *category*, which is a different guarantee. Parametrized over all four outcomes: `query_busy` → 429, `query_timeout` → 422, `query_data_error` → 422, and an unrecognised category → 500 through the `.get` default. It patches `router_analysis.run_analysis_preview` — the name the module imported, not the definition in `service_analysis` — the way the suite already patches the queue hop at `test_analysis_materialize.py:81-83`.

The detail assertion holds on every branch including the 500: `SandboxError.user_message` is already the sanitized text, and the shared `HTTPException` handler passes `detail` through.

## Item 3 — job-slot teardown fixture

The materialize endpoint admits one non-terminal analysis job per user, and the tests share a database, so a test that enqueues a job and leaves it pending 429s whatever runs next — under `-n 4`, an unrelated victim in another file. Three tests hand-released the slot "for later tests (shared DB)". An autouse teardown now fails every analysis job left behind, so a new test cannot forget; the three annotated hand-releases are gone and those tests keep their assertions.

Scope note on what was **not** removed: the remaining `status = "failed"` lines in this file are a different thing. Two are load-bearing mid-test (`test_second_materialize_while_active_is_rejected` proves a finished job releases the slot; `test_source_size_gates_dissolve_and_buffer` enqueues twice), and the rest release rows the test constructed directly rather than the endpoint slot. The three that existed purely for *later* tests are the ones the fixture replaces.

`TestUserErrorMessage` overrides the fixture with a sync no-op: those are pure-function tests, and handing a sync test an async fixture errors during setup.

## Not included

Item 4 (assert `current_step == "queued"` at enqueue) stays gated on #695 per the issue body, so this does not close #789.

## Verification

```
uv run pytest tests/test_analysis_materialize.py tests/test_analysis_preview.py \
  tests/test_ai_chat_run_analysis.py tests/test_jobs_retry_analysis_copy.py -q
# 155 passed
uv run ruff check tests/ && uv run ruff format --check tests/
```

Tests only — no schema, API or security impact.

Refs #789


ishiland:
@codex review

ishiland:
Conflict resolved and the base is current.

One conflict, in `test_analysis_materialize.py`, additive on both sides: main's tenant-cap helpers (`_create_other_user_id`, `_fill_tenant_analysis_slots`, `_release_jobs`, from #1015) and this branch's autouse job-slot teardown landed at the same seam below `_create_job`. Both kept.

They compose rather than collide. `_fill_tenant_analysis_slots` counts what is already active instead of assuming zero, precisely because the shared worker database can carry a sibling's leftovers — the autouse teardown means there are none, so those tests get more deterministic, not less. `_release_jobs` is redundant for the same reason and left alone; it costs nothing and is not this branch's code to prune.

Head is `1919d7930`, `origin/main` (`f0113cdfa`) is an ancestor, and GitHub now reports `mergeable=MERGEABLE`. Local: 127 passed across `test_analysis_materialize.py` and `test_analysis_preview.py`, including main's new tenant-cap tests alongside the fixture; ruff clean.

ishiland:
@codex review

ishiland:
Batch 4's red is this PR, and it is fixed at `a13d4804a2`.

**What broke.** Twenty-five errors in `test_analysis_materialize.py`, of which one was real and twenty-four were cascade — a failed fixture corrupts pytest's finalizer state, so every later test in the file errors with a bare `assert not self._finalizers` and the log reads like a massacre.

The real one:

```
ERROR at setup of TestMaterializeWorker.test_timeout_message_names_the_budget_that_fired
'test_timeout_message_names_the_budget_that_fired' requested an async fixture 'client',
with no plugin or hook that handled it.
```

This PR's autouse teardown was ASYNC and depended on `test_db_session` → `client`. #1013, also in the batch, adds two plain `def` tests to `TestMaterializeWorker`. pytest refuses to hand an async fixture to a sync test, so they error at setup. Neither PR is wrong alone and both were green on their own branches.

**The fix, and why it is not another override.** I had already met this exact trap here with `TestUserErrorMessage` and answered it with a per-class sync override. That patches one occurrence and loses to the next, which is precisely what happened — the next occurrence just came from another session. So the fixture no longer cares what shape the tests are: it is synchronous and uses its own short-lived connection rather than the request's async session. That also makes teardown ordering irrelevant, which is the thing the first version got wrong in the other direction. The class override is deleted with it, and a test that never took a client skips the connection entirely.

**Verified against the combination that failed, not just this branch.** Merged this fix into `integration/batch-4` locally, confirmed both of #1013's sync tests and the new fixture are present, and ran it: **85 passed**, where the batch produced 25 errors. Also 171 passed across the analysis and chat suites under `-n 4`, matching how CI runs it.

ishiland:
@codex review

ishiland:
Rebased onto current main — head `58624f28d`. Thanks for the two corrections; both changed what I did.

**On #1057 vs #1013.** You are right, and it makes this much cleaner: the sync tests are on main's HEAD, not in a batch. So the failing combination now exists in this PR alone, no batch branch needed. Confirmed after the merge — both `test_timeout_message_names_the_budget_that_fired` and `test_analysis_timeouts_track_their_settings` are present alongside the new fixture, so this PR's own CI is the red-to-green.

**On blast radius.** My "25 errors" was a single-file local run, which is exactly the measurement that cannot see the cascade — with only that file collected, the poisoned FixtureDef has ~80 tests to reach instead of ~950. The real number is what you measured, 934 to 951 errors spanning the whole worker stream. That also means my verification was too narrow, so I widened it: running the full backend suite locally under CI's exact configuration, `-n 4 --dist loadgroup -m 'not perf'`.

**On the mechanism.** Ran the standalone repro: poisoner first gives 3 errors, and `finish()` early-returns on `cached_result is None` without clearing `_finalizers`, so every later `execute()` of that fixture trips the assert for the rest of the worker's session. There is no recovery once it happens — which is why this had to prevent the setup error rather than clean up after it, and why a second class-level shadow would only have moved the trap.

ishiland:
@codex review

ishiland:
Full-suite verification done, under CI's exact configuration.

```
uv run pytest -n 4 --dist loadgroup -m 'not perf'
6541 passed, 83 skipped, 0 errors, 0 failures   (634s)
```

On head `58624f28d`, which is this PR merged with current main — so #1057's two sync tests and the new fixture are both in the tree. Batch 4's run of the same suite was 5675 passed / 951 errors.

That is the measurement that was missing before. A single-file run cannot see this bug: with only `test_analysis_materialize.py` collected, the poisoned FixtureDef has about 80 tests left to reach, so it reports ~25 errors and looks contained. Under `--dist loadgroup` it has the rest of the worker's stream — around 950 across `test_stac_visibility`, `test_tiles`, `test_validation`, `test_vrt_*`, `test_services_endpoints`, `test_tile_signing`, `test_stale_pending_reaper` and others. Same fault, two orders of magnitude apart, and the smaller number is the one that invites a narrow fix.

PR #1031, opened 2026-07-31
test(builder): drop the real-timer delay from the StackRow rename test

### LINKED ISSUE #997: test(builder): StackRow rename test is flaky under the full Vitest run
`src/components/builder/__tests__/StackRow.test.tsx:878` — *"a space can be typed into the rename input and Enter commits without re-firing the row action"* — fails intermittently during `npm run test:coverage`, and passes reliably in isolation.

Observed on `2026-07-30` across six full-suite runs on two unrelated branches (report-buffer and tile-auth work; neither touches `StackRow` or anything it imports): **1 failure in 6 runs**, and 3/3 passes with

    npx vitest run src/components/builder/__tests__/StackRow.test.tsx

## What was and wasn't checked

- The failing runs were on feature branches, but their diffs do not touch `StackRow`, its imports, or its test file — so the cause is not those changes.
- The file passes in isolation on a clean `origin/main` checkout (57/57).
- ~~**A full-suite run on clean `main` was not attempted**~~ — one has now run. `npm run test:coverage` on `main` at `9a788ba92` (2026-07-31) completed clean. That is a single data point against an event seen once in six runs, so it neither reproduces nor rules out the flake; several more runs are needed before "does not reproduce on main" is a supportable claim.

The practical cost either way: an intermittent red Frontend Tests job on PRs that did not cause it, which sends the author chasing a change they did not make.

## Likely cause

The test is one of the few in the file that drives `userEvent` asynchronously:

```ts
const user = userEvent.setup();
...
await user.type(input, 'a b');
expect(input).toHaveValue('a b');
await user.keyboard('{Enter}');
```

`userEvent.setup()` inserts a real-timer delay between keystrokes. Under the parallel full-suite run the worker competes for CPU, so the assertion can observe a partially-typed value or the `{Enter}` can land before React has flushed the last keystroke. The sibling rename tests use synchronous `fireEvent` and have not been seen to flake.

`userEvent.setup({ delay: null })`, or `fireEvent.change` + `fireEvent.keyDown` matching the neighbouring tests, would both remove the timing dependence — but the space-handling behaviour under test is exactly what makes `user.type` attractive here, so the fix deserves a moment's thought rather than a blind swap.

**Scope of the pattern in this file, re-counted 2026-07-31.** Exactly three of the file's 57 tests call `userEvent.setup()`: `:823`, `:842`, and `:879` (the failing test). The first two are useful comparison cases rather than fixes to make. Both belong to the `row keydown target guard (v1.6.0 audit A7)` block, both drive a single `user.keyboard(' ')` or `user.keyboard('{Enter}')` against a focused button, and neither has been seen to flake. So the distinguishing feature of `:879` is not `userEvent.setup()` on its own but the multi-keystroke `user.type` plus the intermediate value assertion, which narrows what a fix has to change and gives two in-file controls to check a fix against.

## Not attempted here

Found while running gates for #907/#908; `StackRow` belongs to the builder-editor domain, so this is filed rather than fixed in an unrelated PR.

ishiland:
## What changed

One line plus its rationale, in `src/components/builder/__tests__/StackRow.test.tsx`:

```ts
const user = userEvent.setup({ delay: null });
```

on *"a space can be typed into the rename input and Enter commits without re-firing the row action"*.

## Why this test and not the other two

Three of the file's 57 tests call `userEvent.setup()` — `:823`, `:842`, and this one. The first two fire a single `user.keyboard(...)` against a focused button and have never been seen to flake. This one runs `await user.clear(input)`, `await user.type(input, 'a b')`, an intermediate `expect(input).toHaveValue('a b')`, then `await user.keyboard('{Enter}')`.

user-event's default `delay: 0` still awaits a real `setTimeout(0)` between keystrokes, so this test yields to the macrotask queue four times where the siblings yield once. Under `npm run test:coverage` the worker competes for CPU and those yields stretch, which matches the observed signature: 1 failure in 6 full-suite runs on branches that touch neither `StackRow` nor anything it imports, against 3/3 passes in isolation.

`delay: null` removes the yields. The two siblings deliberately keep the default and stay as in-file controls.

## Why not fireEvent

The neighbouring rename tests use `fireEvent`, and swapping would be the smaller diff, but the block's header comment says what it is for: user-event 14 implements real keyboard activation *gated on `defaultPrevented`*, and these tests prove the descendants work again after the A7 row-keydown fix. `fireEvent.keyDown` does not model that gate, so the swap would keep the test green while removing what it tests. Dropping the delay takes out the timing dependence and leaves the dispatch semantics intact.

## Verification

`npx vitest run src/components/builder/__tests__/StackRow.test.tsx` — 57 passed, 3 consecutive runs.

`npm run test:coverage` — 366 files / 4284 tests passed, twice (once before this change, once after).

Per-test duration from `--reporter=verbose`, same machine, back to back:

| test | default delay | `delay: null` |
|---|---|---|
| Space on the focused eye toggle | 11ms | 9ms |
| Enter on the focused eye toggle | 7ms | 7ms |
| **the rename test** | **25ms** | **15ms** |

The two controls are unchanged inside noise; the fixed test drops ~10ms, which is the four real-timer yields going away.

Being straight about the limit of this evidence: green runs cannot prove a 1-in-6 flake is gone, and I have not reproduced the failure. The argument is mechanical — the test no longer depends on real-timer scheduling, and the two tests that still do are the ones that never flaked.

None. Test-only.

Closes #997

### DIFF
--- a/frontend/src/components/builder/__tests__/StackRow.test.tsx
+++ b/frontend/src/components/builder/__tests__/StackRow.test.tsx
@@ -876,7 +876,22 @@ describe('row keydown target guard (v1.6.0 audit A7)', () => {
   });
 
   it('a space can be typed into the rename input and Enter commits without re-firing the row action', async () => {
-    const user = userEvent.setup();
+    // fix(#997): delay: null. user-event's default inserts a real-timer
+    // setTimeout(0) between keystrokes, and this is the only test in the file
+    // that types more than one — three keystrokes plus a clear, each yielding
+    // to the macrotask queue. Under the parallel full-suite run the worker
+    // competes for CPU and those yields stretch, which is what made this test
+    // (and only this test) fail roughly 1 run in 6 while passing 3/3 in
+    // isolation. The two siblings above keep the default: they fire a single
+    // key against a focused button, have never been seen to flake, and are
+    // worth keeping as controls.
+    //
+    // Not swapped to fireEvent, which is how the neighbouring rename tests are
+    // written. This block exists to prove real keyboard activation works again
+    // after the A7 fix, and that activation is gated on defaultPrevented —
+    // something only user-event models. Dropping the delay removes the timing
+    // dependence without giving up the dispatch semantics under test.
+    const user = userEvent.setup({ delay: null });
     const onRename = vi.fn();
     const onSelectLayer = vi.fn();
     const layer = makeLayer({ id: 'sp-layer', dataset_name: 'Old' });


ishiland:
@codex review

PR #1032, opened 2026-07-31
fix(catalog): pad a degenerate bbox instead of building a zero-area extent ring

### LINKED ISSUE #944: fix(catalog): a zero-height bbox builds an invalid zero-area extent ring

ishiland:
Closes #944. Split out of #904; pre-existing, not a regression from #901.

## The defect

`bbox_to_extent_wkt` builds `POLYGON((w s, e s, e s, w s, w s))` when
`south == north` — four collinear vertices, zero area, not a valid polygon.

The *width* case was already guarded, but only on the crossing branch (drop
zero-width halves, fall back to the full band). The non-crossing branch at
`geo.py:168-169` returned its ring directly and was unguarded in **both**
dimensions.

Of the three callers, `seam_extent_wkt_for_table` pre-pads and a raster
footprint always has real extent, so the live exposure is
`catalog/sources/stac_router.py:554`, which passes `item.bbox` through
unvalidated. A STAC item describing a single point, or a scanline of points on
one parallel, stored an invalid ring directly.

## What changed

Pad both axes, on both branches, adopting the convention #934 established one
caller upstream rather than inventing a second one: trigger on a span below
`1e-12`, widen by `1e-9` in each direction. `1e-9` is also the magnitude the
`ST_Expand` calls in `processing/ingest/metadata.py` and
`catalog/features/service.py` use on degenerate POINT/LINESTRING extents, so a
padded extent produced here is indistinguishable in size from one produced by
ingest.

One addition to #934's version: the pad is clamped to the coordinate domain. An
axis sitting exactly on its edge — a point on the antimeridian, or at a pole —
has no room outward, and padding it unclamped would emit longitude
`180.000000001`. The sliver grows inward instead.

Rejecting was the alternative and is closed: `chk_records_spatial_extent_type`
admits POLYGON, so an invalid zero-area ring stores happily today, and anything
that starts rejecting it turns a working ingest into a failed one.

Second edit, per the issue: `seam_extent_wkt_for_table` drops its own copy of
the pad. Leaving both is not a double-pad (the seam pad widens the span past the
`1e-12` trigger, so the helper's never fires) — it is worse, because the
caller's copy stays the one that runs and the helper's becomes dead code on that
path, with two copies of one convention to keep in sync. That is what the
"Related: #944" note in #934's comment was flagging as temporary.

## Verification

```
  tests/test_antimeridian_extent.py tests/test_antimeridian_rollup.py \
  tests/test_extent_producers_934.py tests/test_export_antimeridian.py \
  tests/test_raster_antimeridian_887.py tests/test_stor_vsi_seam_lint.py -q
  # 41 + 193 passed
```

`TestBboxToExtentWkt` gains two parametrised tests. The first covers zero height
non-crossing, zero height crossing (both halves stay valid), zero width
non-crossing, a single point degenerate on both axes at once, and the four
domain-edge cases where the pad clamps. All assert `is_valid`, the geometry
type, the padded bounds, and `area > 0`.

The second pins that padding never moves a genuine bbox, including a `1e-9`-span
bbox — the tightest the `ST_Expand` producers emit, which must read as genuine
rather than be re-padded.

### DIFF
--- a/backend/app/core/geo.py
+++ b/backend/app/core/geo.py
@@ -151,6 +151,30 @@ def _ring(x0: float, south: float, x1: float, north: float) -> str:
     return f"({x0} {south},{x1} {south},{x1} {north},{x0} {north},{x0} {south})"
 
 
+# fix(#944): a span this narrow is degenerate, and the sliver it is widened to.
+# Both come from :func:`seam_extent_wkt_for_table`, which hit the same problem
+# one caller upstream first; 1e-9 is also the magnitude the ``ST_Expand`` calls
+# in ``processing/ingest/metadata.py`` and ``catalog/features/service.py`` use
+# on POINT/LINESTRING extents, so a padded extent produced here is
+# indistinguishable in size from one produced by ingest.
+_DEGENERATE_SPAN = 1e-12
+_DEGENERATE_PAD = 1e-9
+
+
+def _pad_degenerate(low: float, high: float, limit: float) -> tuple[float, float]:
+    """Widen a zero-span axis to a sliver, without leaving ``±limit``.
+
+    Padding outward is what the ``ST_Expand`` callers do, but an axis sitting
+    exactly on its domain edge (a point on the antimeridian, or at a pole) has
+    no room on one side, so the pad is clamped and the sliver grows inward
+    instead. Non-degenerate spans are returned untouched, so a genuine bbox
+    never moves by an epsilon.
+    """
+    if high - low >= _DEGENERATE_SPAN:
+        return low, high
+    return max(-limit, low - _DEGENERATE_PAD), min(limit, high + _DEGENERATE_PAD)
+
+
 def bbox_to_extent_wkt(west: float, south: float, east: float, north: float) -> str:
     """Build extent WKT for an RFC 7946 §5.2 bbox, splitting it at ±180.
 
@@ -164,8 +188,19 @@ def bbox_to_extent_wkt(west: float, south: float, east: float, north: float) -> 
     same latitude band -- which ``catalog.records.spatial_extent`` accepts as of
     migration ``0030_records_spatial_extent_type`` and which
     :func:`extent_to_bbox` reads back as the original ``west > east`` pair.
+
+    fix(#944): a degenerate axis is padded to a sliver first. A bbox with
+    ``south == north`` (a single point, or a run of points on one parallel)
+    otherwise yields four collinear vertices -- zero area, and not a valid
+    polygon -- and the same applies to a zero-width non-crossing bbox. Only the
+    width case was guarded before, and only on the crossing branch. Padding
+    rather than rejecting follows the producers, which already ``ST_Expand``
+    degenerate POINT/LINESTRING extents by the same 1e-9; rejecting would turn
+    a working ingest into a failed one for data the schema stores happily.
     """
+    south, north = _pad_degenerate(south, north, 90.0)
     if west <= east:
+        west, east = _pad_degenerate(west, east, 180.0)
         return f"POLYGON({_ring(west, south, east, north)})"
 
     # A crossing bbox whose west sits at +180 (or east at -180) has a zero-width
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

Issue #821, opened 2026-07-28
sec(auth): API-key hardening — no expiry, no staleness check, query-param lane writes the credential into logs


ishiland:
From the 1.6.0 pre-tag audit (latent, P3). Three related gaps in the API-key lane:

- No `expires_at` column: keys live forever unless manually revoked (`auth/models.py:176-202`).
- No `token_version` staleness check: a role demotion doesn't invalidate previously minted keys the way it does JWTs.
- The query-param fallback (`_resolve_api_key()`, `auth/dependencies.py:62-68`) puts the credential in request URLs, so it lands in access logs and any upstream proxy logs.

Suggested shape: add optional `expires_at` (default forever for backward compat, encourage rotation in docs), check `token_version` at resolution, and log-redact or deprecate the query-param lane (it exists for tile URLs; those now have signed URLs and embed tokens, so an audit of remaining legitimate users is the first step).

PR #1033, opened 2026-07-31
fix(migrations): guard 0029's downgrade against dropping key expiry and epoch state

### LINKED ISSUE #1016: fix(migrations): 0029's downgrade silently resurrects expired and epoch-revoked API keys
Split out of #998, which tracks the wider question of making 0019's tenant-ownership adoption reachable without a destructive downgrade. That question is a maintainer decision with no chosen direction. This piece is not: it is a scoped guard on one migration, and it is worth doing whichever direction #998 eventually takes.

## The gap

`0029_api_key_hardening` downgrades by dropping three columns with no guard at all (`backend/alembic/versions/0029_api_key_hardening.py:67-70`):

```python
def downgrade() -> None:
    op.drop_column("api_keys", "key_epoch", schema="catalog")
    op.drop_column("api_keys", "expires_at", schema="catalog")
    op.drop_column("users", "key_epoch", schema="catalog")
```

Both dropped columns carry security state, and losing them fails in the unsafe direction:

- `api_keys.expires_at` is the expiry. Once the column is gone, key resolution has no expiry to check, so a time-limited key that has already passed its expiry comes back as a permanent key.
- `users.key_epoch` and `api_keys.key_epoch` are the revocation primitive (#821): a security event bumps the owner's epoch, and a key whose snapshot no longer matches its owner stops resolving. Drop both columns and that comparison disappears, so keys revoked by an epoch bump come back live.

Neither produces an error. The downgrade succeeds, the schema is valid, and the deployment is quietly less secure than the operator believes.

## Why it matters even though downgrades are rare

The one documented procedure that reaches this migration is the multi-tenant fresh-cluster DR path, `alembic downgrade 0016` (RUNBOOK § disaster recovery). That is a procedure someone runs on their worst day, usually for the first time.

Worse, 0029's drops are not gated by the later refusals. Walking back from head the order is 0030, then 0029, then 0027, 0022, 0021, 0020. Several of those migrations do index work inside Alembic's `autocommit_block()`, which commits the DDL preceding it, so a refusal at 0021 does not roll back what 0029 already dropped (RUNBOOK.md:338-347 spells this out). An operator who runs the documented recipe and is stopped by a refusal several steps later has still lost their key expiry and revocation state, in a database RUNBOOK then tells them is unsalvageable.

## What to do

Mirror the refusal pattern 0030 already uses (`backend/alembic/versions/0030_records_spatial_extent_type.py:57-79`): count the affected rows first, and if any exist, raise a `RuntimeError` whose message names the count, gives a query to inspect the rows, and gives the explicit SQL that makes the loss acceptable, so an operator who genuinely wants it opts in deliberately rather than discovering it afterwards.

The guard predicate: refuse when either

- any `catalog.api_keys` row has a non-null `expires_at`, or
- any `catalog.users` row has `key_epoch > 1`.

The second condition is the right coarse test because `api_keys.key_epoch` is a mint-time snapshot of the owner's value; if no owner has ever been bumped past the default of 1, no revocation state exists to lose. For the error message, the precise set of keys that would come back live is the mismatch:

```sql
SELECT ak.id, ak.user_id, ak.expires_at, ak.key_epoch, u.key_epoch AS owner_epoch
FROM catalog.api_keys ak
JOIN catalog.users u ON u.id = ak.user_id
WHERE ak.expires_at IS NOT NULL OR ak.key_epoch <> u.key_epoch;
```

Naming those rows is more useful than a bare count, because the remediation an operator most likely wants is to delete the affected keys rather than to accept their resurrection.

## Test impact, which is the part that gets missed

`backend/tests/alembic_helpers.py` is the single place that knows about refuse-to-coerce downgrades, and its docstring says so directly (`:22-23`: "If a future migration gains a refuse-to-coerce downgrade, teach this module about it rather than adding cleanup to individual test files"). It already handles three: 0030 is normalized automatically before every downgrade, while `0010_oauth_github_provider_type` and `0005_dormant_tenancy` are documented but deliberately not auto-cleaned because their state is semantic test state.

A new guard on 0029 has to be added there, and the choice between the two treatments is a real one. Auto-cleaning it means test code that nulls out expiry and resets epochs, which is the same class of thing #933 warns against lifting into application code. Leaving it documented means any test that creates an expiring or epoch-bumped key owns its own cleanup, and a missed cleanup surfaces as a failure in an unrelated file sharing the same xdist worker, which is exactly the failure mode #933 existed to end. Decide it in the PR and write down why.

## Acceptance

- [ ] 0029's downgrade refuses when expiry or epoch state exists, with a message carrying the count, the inspection query, and the opt-out SQL.
- [ ] `backend/tests/alembic_helpers.py` knows about the new guard, with a comment saying which treatment was chosen and why.
- [ ] A test asserts the refusal fires on a database holding an expiring key, and that a clean database still downgrades.
- [ ] Backend `ruff check` / `ruff format --check` clean, and `make alembic-check` passes.

ishiland:
`0029_api_key_hardening` downgraded by dropping three columns with no guard. Both dropped columns carry security state, and losing them fails in the unsafe direction:

- with no `api_keys.expires_at`, key resolution has no expiry to check, so a time-limited key that has already passed its expiry comes back as a permanent key
- with no `key_epoch` pair, the revocation comparison (#821) disappears, so keys revoked by an epoch bump come back live

Neither raises. The downgrade succeeds, the schema is valid, and the deployment is quietly less secure than the operator believes — on the one documented procedure that reaches this migration (`alembic downgrade 0016`, the multi-tenant fresh-cluster DR path), run on someone's worst day.

## The guard

Mirrors 0030's refusal: count first, then raise a `RuntimeError` naming the counts, the query that lists the exact keys that would come back live, and the SQL that revokes them. It refuses when any `catalog.api_keys` row has a non-null `expires_at`, or any `catalog.users` row has `key_epoch > 1`. The second is the right coarse test because `api_keys.key_epoch` is a mint-time snapshot of the owner's value: if no owner has ever been bumped past the default of 1, no revocation state exists to lose.

The remediation offered is the one an operator most likely wants — delete the affected keys rather than accept their resurrection — plus the `UPDATE catalog.users SET key_epoch = 1` needed to clear the second condition, since deleting keys alone leaves the guard still refusing.

## Treatment in `alembic_helpers.py`: auto-clean, and why

The issue asked for this to be decided in the PR. **Auto-cleaned, alongside 0030 — not documented-only alongside 0010/0005.**

Those two are semantic test state, created by the single test that cares about them. This is not. `test_api_key_auth.py` commits both an expiring key (`test_api_key_with_future_expiry_accepted_and_surfaced`) and a `key_epoch` bump (`test_stale_key_epoch_api_key_rejected`) as ordinary auth coverage, cleans up neither, and has no relationship to migrations. Meanwhile more than twenty downgrade call sites across six files cross 0029 on their way to 0004/0005/0009.

Leaving it documented would reproduce #933 exactly, so I measured it rather than arguing it. With the normalization disabled:

```
$ uv run pytest tests/test_api_key_auth.py tests/test_dormant_tenancy_migration_roundtrip.py
AssertionError: setup downgrade to 0005 failed: RuntimeError: 2 API key(s) carry an
expiry and 4 user(s) have had their key_epoch bumped; 6 key(s) would come back live
or permanent if these columns were dropped.
```

A file with nothing to do with API keys, failing on which xdist worker the auth suite landed on. That is the failure mode #933 existed to end.

The normalization clears the state rather than deleting rows (a test database's keys are fixture noise, and there is no need to destroy rows a sibling test may still read), and each column is guarded independently so it is a no-op at any schema revision.

`run_alembic` gained `normalize=False` for the tests that assert a guard actually fires. Nothing else should pass it, and the docstring says so.

## Tests

Three new tests in `test_alembic_helpers.py`:

- the guard fires on a database holding an expiring key and a bumped epoch, and the message carries the counts, the inspection query, and the opt-out SQL
- the same state is absorbed by the runner (the #933 shape)
- a clean database still downgrades — run with `normalize=False` so nothing clears state on the way through, which is what makes the refusal meaningful

The two `normalize=False` tests call `normalize_seam_extents()` explicitly first: 0030 is crossed on the way to 0029, so a seam extent left by a sibling test in the same worker would fire the wrong guard and the assertions would test nothing.

## Schema impact

Downgrade-path only. No column, index, or constraint changes; `alembic check` reports no new upgrade operations. The upgrade path is untouched.

## Verification

```
cd backend && uv run pytest tests/test_alembic_helpers.py -v                       # 4 passed
cd backend && uv run pytest tests/test_api_key_auth.py \
    tests/test_dormant_tenancy_migration_roundtrip.py                              # 34 passed
cd backend && PYTHONPATH=. uv run alembic check                                    # No new upgrade operations detected
cd backend && uv run ruff check . && uv run ruff format --check .                   # clean
```

Closes #1016

### DIFF
--- a/RUNBOOK.md
+++ b/RUNBOOK.md
@@ -233,7 +233,7 @@ upgrade is what fixes that, and the only way to reach it is to downgrade below
 > | Migration | What its `downgrade()` does |
 > |---|---|
 > | 0030 | **Refuses** while any `catalog.records` row holds a MULTIPOLYGON `spatial_extent` — i.e. any antimeridian-crossing footprint, which the current schema exists to store |
-> | 0029 | **Discards** `api_keys.expires_at` and both `key_epoch` columns. The re-upgrade recreates them as NULL and 1, turning time-limited keys into permanent ones and un-revoking keys that a `key_epoch` bump had invalidated |
+> | 0029 | **Refuses** while any API key carries an expiry or any user's `key_epoch` has been bumped. Dropping `api_keys.expires_at` and both `key_epoch` columns would turn time-limited keys into permanent ones and un-revoke keys a `key_epoch` bump had invalidated; the error message lists the affected keys and the SQL that revokes them |
 > | 0027 | **Refuses** while any dataset uses a `parquet`, `json`, `xlsx`, or `xls` source format — all currently supported |
 > | 0022 | **Discards** `tenant_id` on `catalog.audit_logs` and `catalog.ingest_jobs`. The re-upgrade re-derives it from each row's live parent, so rows whose parent is gone lose tenant attribution permanently |
 > | 0021, 0020 | **Refuse** when two tenants share a collection name, OAuth subject, or `datasets.table_name` — all legal under the current per-tenant scoping |
@@ -339,9 +339,15 @@ SQL
 from the dump.** A failed downgrade does not roll back the ones that already
 succeeded. Several of these migrations do index work inside Alembic's
 `autocommit_block()` (0020, 0021, 0022 among them), which commits the DDL
-preceding it, so a refusal at 0021 leaves you past 0029's and 0022's discards
-with no transaction to undo them. Drop and recreate the target database, then
+preceding it, so a refusal at 0021 leaves you past 0022's discards with no
+transaction to undo them. Drop and recreate the target database, then
 re-run 2a and 2b against the untouched dump before attempting anything else.
+
+0029 is the one place where the order works in your favour: walking back from
+head it is reached second, before any of the discarding migrations, so its
+refusal stops the run while the database is still whole. Deal with its
+remediation there rather than discovering the loss after a refusal further
+down (fix(#1016) — it used to drop those columns silently).
 The dump file itself is never modified, so nothing is lost by restarting — but
 continuing in a half-downgraded database is how a recovery turns into a second
 incident.
@@ -1024,8 +1030,9 @@ been accepted.
 
 Rolling the schema back (`alembic downgrade`) is rare — the supported
 recovery paths are restore-from-backup (§2/§3) and the upgrade rollback
-quick-reference in UPGRADING.md. When you do downgrade, one migration can
-refuse by design.
+quick-reference in UPGRADING.md. When you do downgrade, several migrations
+refuse by design rather than discard state the re-upgrade cannot rebuild. The
+two you are most likely to meet are documented in full below.
 
 ### Downgrading past `0030_records_spatial_extent_type` fails when antimeridian-crossing extents exist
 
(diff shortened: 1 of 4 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1034, opened 2026-07-31
fix(builder): reword the immediate layer-add history summary

### LINKED ISSUE #941: fix(builder): reword the layer-add history summary so it does not read as a claim about the saved map
Split out of #787 (item 4). Original report:

> [live] History logs "Added … layer" for a not-yet-saved change — discard the map edit and the entry describes something that never persisted.

Reproduced in the running app. Pointers verified on main @ 9a788ba92.

## Why the entry is not simply wrong

The history entry is accurate about what the server was asked to do. Adding a layer is an immediate server POST: `handleAddDataset` (`frontend/src/components/builder/hooks/use-builder-layers.ts:603`) calls `addLayerMutation.mutate` straight away at `:616`, while style, filter, label, popup, and reorder edits are held locally and flushed on save. So the builder has two classes of edit with different persistence timing, and history faithfully records one class and not the other.

Discarding the map edit does not roll the POST back. The layer row really was created; what "never persisted" is the surrounding unsaved state the user was assembling around it. The entry is true and reads false, which is a copy problem rather than a write-path problem.

## Decision: accept the behavior, fix the copy

Two alternatives were considered and are not being taken:

1. **Defer layer creation to save.** Would make the two edit classes uniform, but the add path relies on the server-assigned layer id to render the row (`onSuccess` merges `createdLayer` into `localLayers`). Deferring means minting a client-side provisional id and reconciling it on save, which touches the save diff, the baseline sync, and group parenting. Too much blast radius for a wording problem.
2. **Filter history display.** Suppress or mark entries whose map edit was later discarded. The builder does not record "this edit was discarded" anywhere, so the signal would have to be invented and threaded through.

What ships instead: the summary text stops making a claim about the user's saved map.

## The summaries are server-side, and only one of the two needs changing

The strings are pre-rendered on the server and stored, not composed in the UI. `record_map_history_event` writes them into `map_edit_history_events` (`backend/app/modules/catalog/maps/models.py:183-184`) and `HistoryPanel.tsx:76` renders `event.summary` verbatim.

There are two `Added … layer` writers and they are not equivalent:

- `backend/app/modules/catalog/maps/router.py:1223`, inside `add_layer_endpoint` (POST `/{map_id}/layers`, `:1163`). **This is the one to change.** It fires on the immediate add, before and independently of any save, so its entry can outlive a discarded edit.
- `backend/app/modules/catalog/maps/router.py:707`, inside `patch_map_layers_endpoint` (PATCH `/{map_id}/layers`, `:619`). This is the save-diff flush: it only records rows that appear in `body.added` and were absent before (`:697-698`). Its "Added … layer" entry does describe a saved map and is already accurate. **Leave it alone**, and do not let a find-and-replace catch it.

## Fix

Reword the summary at `router.py:1223` so it describes the row that was created rather than implying the map was saved. Suggested:

```python
summary=f"Created {target_name} layer (saved immediately)"
```

The exact string is the reviewer's call; the requirement is that a reader who then discards their edit does not come back to an entry that contradicts what they see. Two constraints on whatever wording lands:

- Do not change `action="layer.add"` at `:1222`. That is a machine key, filtered on by the admin audit viewer (`frontend/src/components/admin/AuditLogViewer.tsx:61`), and it stays stable.
- Do not change `:707`, per above.

That is the whole change: one f-string, plus a test pinning that the POST path and the PATCH path produce different summaries so the distinction cannot be collapsed by a later refactor.

## Explicitly out of scope: the i18n debt

These summaries are stored as rendered English in `map_edit_history_events` and shown verbatim to every user regardless of locale. A German user's history panel is in English today, and rewording the string does not change that. Fixing it properly means storing a message key plus structured `details` and rendering client-side, which is a schema-and-backfill change across every `summary=` site in the maps router, not just this one. That is a separate issue and should not be attempted here.

## Acceptance

- Add a layer, then discard the map edit: the history entry describes a layer row being created, not the map being changed or saved.
- Add a layer through the save path (a layer present in the PATCH `added` diff): the entry is unchanged.
- `action` values are unchanged for both paths.
- `make test` (backend).

ishiland:
Closes #941. Split out of #787 (item 4). The decision is "accept the behaviour,
fix the copy" — recorded on the issue, not reopened here.

## Why the entry was not simply wrong

`handleAddDataset` calls `addLayerMutation.mutate` immediately, while style,
filter, label, popup and reorder edits are held locally and flushed on save. So
the builder has two classes of edit with different persistence timing, and
history faithfully records one and not the other. Discarding the map edit does
not roll the POST back — the layer row really was created. What "never
persisted" is the surrounding unsaved state.

## What changed

One f-string, at the `add_layer_endpoint` writer:

```
-summary=f"Added {target_name} layer",
+summary=f"Created {target_name} layer (saved immediately)",
```

Deliberately untouched:

- The PATCH save-diff writer in `patch_map_layers_endpoint`. It only records
  rows present in `body.added` and absent before, so its "Added ..." entry does
  describe a saved map and is already accurate.
- `action="layer.add"` on both paths — a machine key
  `AuditLogViewer.tsx` filters on.

## Out of scope

These summaries are stored as rendered English and shown verbatim regardless of
locale. Rewording does not change that; fixing it means storing a message key
plus structured details and rendering client-side, across every `summary=` site
in the maps router. Separate issue.

## Verification

```
cd backend && set -a && source ../.env.test && set +a && uv run pytest tests/test_maps.py -q
  # 161 passed
```

New: `TestMapHistory::test_the_two_layer_add_paths_keep_distinct_summaries`
drives both writers on one map and pins each summary exactly, so a later
refactor cannot collapse the distinction. Nothing else in the repo asserted the
old string.


ishiland:
@codex review

ishiland:
@codex review

ishiland:
Heads up: I resolved this branch's merge conflict from the wave-coordination side, so its head moved to `10fdaf1df` without its own session doing anything. **No app code changed** — the whole resolution is one integer.

The collision was in `backend/tests/test_layering.py`. #1052 took the `maps/router.py` LOC cap 1385 → 1417 (+32 for `_record_image_capture`) while this branch took it 1385 → 1393 (+8 for the reworded add-layer summary). Both comments kept, cap resolved to **1425**.

The ratchet is exact-match (`test_module_loc_caps_have_no_headroom`), so I verified against the file rather than trusting the two deltas: `router.py` is 1425 lines at the merge, and `test_layering.py` passes (34 tests). `test_maps.py` auto-merged with no conflict.

I did not run the DB-backed `test_maps.py` locally — this worktree has no credentials for the test database — so CI is the first place that runs it. Nothing else about the branch is touched, and auto-merge is left off.

ishiland:
@codex review

PR #1035, opened 2026-07-31
test(sec): close the two Rule 2 structural-gate resolution edges

### LINKED ISSUE #996: test(sec): remaining resolution edges in the Rule 2 structural gate
Follow-up to #936 / #974. The Rule 2 structural gate merged with two known resolution edges, both raised by review on the final pass. The gate was merged deliberately with these open: it fails conservative on the security side, and a test-file guard that catches the common regressions is worth having on main while these are refined. Neither is a security hole in current code — the tree has no instance of either shape.

## 1. Argv detection misclassifies ordinary data (fix this one first)

Any list or tuple whose first element resembles a GDAL utility name is treated as a command vector, even when it is plain data:

```python
SUPPORTED_TOOLS = ["gdalinfo", "ogrinfo"]   # flagged today
```

A contributor adding a constant like that, or a function returning UI choices, gets a confusing security failure and the only ways out are a misleading allowlist entry or a pointless call to the environment helper. **This is a false positive that blocks correct code**, which is worse for a CI gate than a theoretical hole, so it should be fixed before anyone trips it.

Fix direction: confirm the literal actually reaches a subprocess invocation (it appears as the args argument of a `subprocess.*` call, directly or via a local binding), rather than classifying any GDAL-looking sequence as an argv.

Review reference: https://github.com/geolens-io/geolens/pull/974#discussion_r3686073827

## 2. Comprehension targets are not modelled as their own scope

`_iter_immediate` treats a comprehension target as a binding in the surrounding function rather than in the comprehension's implicit scope. So a function containing a comprehension whose target is named `rasterio` resolves a genuine module-imported `rasterio.open(path)` elsewhere in that same function as `_OTHER`, and the collector reports zero opens even when the call is unwrapped.

This one fails in the unsafe direction (an unwrapped open goes unreported), though it needs a contributor to name a comprehension target `rasterio`, which nothing in the tree does.

Fix direction: treat list/set/dict comprehensions and generator expressions as lexical scopes so their targets cannot leak into the enclosing function's binding table. The scope machinery added in #974 (`_scope_info` / `_resolve_credit`) already walks innermost-first, so this is an additional scope kind rather than new analysis.

Review reference: https://github.com/geolens-io/geolens/pull/974#discussion_r3686073825

## Acceptance

- A GDAL-named constant that never reaches a subprocess does not trip the gate; an argv that does reach one still trips it when unclamped.
- A comprehension target named `rasterio` no longer blinds detection of an unwrapped `rasterio.open` in the same function.
- Both shapes get synthetic-module regressions alongside the 29 already in `backend/tests/test_rule2_structural.py`.
- `cd backend && uv run pytest tests/test_rule2_structural.py -v` and a re-verification against the real tree showing no new false positives, the way #974 verified its prefix and family rules.

ishiland:
Follow-up to #936 / #974, which merged the Rule 2 structural gate with two edges known and open. Both are closed here, and five rounds of review turned the first one into a materially more careful piece of analysis than it started as.

## 1. Argv detection misclassified ordinary data

Any list or tuple whose head resembled a GDAL utility name was a command vector, so this failed a **security** gate:

```python
SUPPORTED_TOOLS = ["gdalinfo", "ogrinfo"]
```

and the only exits were a misleading allowlist entry or a pointless call to the env helper. A false positive that blocks correct code is worse in a CI gate than a theoretical hole, which is why the issue asked for this one first.

A literal is now a command vector when its value **escapes** — passed to a call, returned, or yielded. Escapes are followed:

- out of **transparent wrappers**: container literals, ternaries, `+` concatenation;
- through the **name it is bound to** by `=`, annotated `=`, walrus, or a `for` / `async for` / comprehension target — and through **alias chains** (`cmd = [...]`, `alias = cmd`, `run(alias)`) to a fixed point;
- with **positional** matching on unpacking, so `ignored, choices = (None, [...])` does not read `choices` as escaping through `ignored`, and chained unpacking keeps every target.

And stopped where the value is consumed rather than moved:

- a **single-index subscript** yields an element, not the vector, so `return cmd[0]` leaves the list inert — while a slice still yields a sequence that could be spawned, and `commands["inspect"]` still yields the argv, because the analysis tracks per name whether it holds the vector or a container of vectors;
- a **nested scope that rebinds the name** breaks the link, so `def inner(cmd): consume(cmd)` is not a use of an outer `cmd`. Both kinds of local binding count — a canonical import lands in `canonical`, not `bound`.

Second rule: a multi-element literal whose **every element names a GDAL utility** is a name list, not a command — but only while it stays out of a call. `subprocess.run(["gdalinfo", "ogrinfo"])` is a real invocation, since a dataset or output path may legitimately be named after a utility, and shape alone cannot separate the two. Iterating such a list (`for tool in [...]`) binds strings, not the list, so it stays data.

**"Escapes" is deliberately wider than the issue's suggested "appears as the args argument of a `subprocess.*` call."** Taken literally that would have blinded the gate to most of its own sites: `processing/raster/vrt.py` owns the only `subprocess.run` and callers reach it as `run_gdal(cmd, env=..., tool=...)`, while `ingest/ogr.py` splats into `asyncio.create_subprocess_exec(*cmd)`. Returning counts too — five of the pre-existing synthetic regressions build an argv and return it.

## 2. Comprehension targets were not their own scope

`_iter_immediate` recorded a comprehension target as a binding in the surrounding function, so a function containing `[rasterio for rasterio in tools]` resolved a genuine `rasterio.open(path)` elsewhere in that same function to `_OTHER`, and the collector reported zero opens with the call sitting there unwrapped. A miss in the unsafe direction.

Comprehensions and generator expressions are lexical scopes now (`_LEXICAL_SCOPES`), kept separate from `_SCOPE_NODES` because the two answer different questions: `_SCOPE_NODES` means "a callable whose body runs LATER" (the boundary rules in `_inside_safe_open_env` / `_scope_uses_safe_env`, and the only nodes with `.args`), `_LEXICAL_SCOPES` means "a scope that owns its names".

Three carve-outs, all matching Python:

- the **outermost iterable** is evaluated before the comprehension's scope exists, so `[x for rasterio in rasterio.open(p)]` resolves the open outside. Same shape as #974 round 11's signature rule, so it reuses `_in_signature` — with a `grandchild` argument, since the path runs through the `ast.comprehension` node;
- a **walrus binds in the containing scope** (PEP 572), and its VALUE propagates, so `[(ropen := rasterio.open) for _ in x]` then `ropen(path)` is still a detected open. That also closes the plain-walrus case, which predates this PR;
- comprehensions stay **transparent to CLI call-credit** (`stop_at_comprehensions=False`). Only def and lambda defer.

## Where this stops, and why

Eight review rounds. Every one surfaced a real Python shape; **none of them exists in `backend/app/`** — the tree has come back byte-identical at every round (11 argv sites, 6 rasterio, zero violations). That is the signal that the rounds were exploring the space of possible evasions rather than fixing live exposure, and it is where this stops.

The module docstring now states the boundary and what it knowingly leaves on each side:

- **Loud (over-reports):** a rebinding under an `if` still reports the dead literal; a vector passed into a helper and spawned there is credited to the helper's scope, not the caller's.
- **Silent (under-reports):** a vector reached through a computed key (`registry[name]`) has no path and is not tracked; neither is one that leaves via a mutation the analysis cannot name (`box.append(cmd)`).

**This gate is a detector, not a runtime control.** Rule 2's actual defenses are the safe-env helpers and the reviewed-justification requirement; the structural test exists to catch a violation entering the tree, and its failure mode is a missed detection that review and the pre-commit hook still cover. The file's own docstring has said "reviewer territory, documented, not promised" since #974 — so documenting the boundary restores the stated contract rather than retreating from it. A detector with documented limits is a legitimate artifact; a detector with silent limits is the dangerous one.

The rule for the next shape someone finds: if it exists in `backend/app/`, fix it. If it does not, add it to the limits list and use an allowlist entry with a written justification — the same answer #974 reached for the wrapper and credit questions. Chasing an undecidable problem with a CI guard has no natural stopping point, so the stopping point is chosen here rather than discovered.

## Regressions

22 new synthetic modules alongside the 29 already there, each paired with its control — inert vs. returned, tool-name list vs. the same list in a call, single-index vs. slice, nested shadow vs. genuine closure use, unpacked position that escapes vs. the sibling that does not.

## Re-verification against the real tree

The #974 way — the detector must see the same sites it saw before, at every round:

| | before | after |
|---|---|---|
| rasterio open/Env sites | 6 | 6 |
| violations | 0 | 0 |

`MIN_GDAL_CLI_ARGV_SITES` (6) and `MIN_RASTERIO_OPEN_SITES` (5) still clear.

## Security impact

Gate-only; no runtime code. The comprehension edge tightens in the safe direction (unwrapped opens that were invisible are now reported); the argv edge relaxes only for values that provably do not move, and every relaxation has a control test asserting the adjacent shape still trips.

## Verification

```
cd backend && uv run pytest tests/test_rule2_structural.py tests/test_layering.py -q
```

Closes #996

### DIFF
--- a/backend/tests/test_rule2_structural.py
+++ b/backend/tests/test_rule2_structural.py
@@ -1438,31 +1438,6 @@ def _argv_escape_kind(node: ast.List | ast.Tuple, rel: str) -> int:
     return best
 
 
-def _use_reaches_the_binding(used: ast.Name, scope: ast.AST, rel: str) -> bool:
-    """True when ``used`` is the scope's binding rather than a nested shadow.
-
-    fix(#996 review): the escape search walks the whole scope, so a nested
-    ``def inner(cmd): consume(cmd)`` counted as a use of an outer ``cmd`` and
-    turned an inert constant into a security failure. Any lexical scope between
-    the use and the binding that binds the same name breaks the link -- the
-    same rule ``_classify_name`` applies to rasterio names.
-
-    ``bound`` is not the whole binding table: a canonical import
-    (``from rasterio import open as cmd``) is recorded in ``canonical``
-    instead, and reading only ``bound`` missed that shadow.
-    """
-    current: ast.AST | None = getattr(used, "_rule2_parent", None)
-    while current is not None and current is not scope:
-        if isinstance(current, _LEXICAL_SCOPES):
-            info = _scope_info(current, rel)
-            if used.id in info.bound or any(
-                used.id in names for names in info.canonical.values()
-            ):
-                return False
-        current = getattr(current, "_rule2_parent", None)
-    return True
-
-
 def _collect_gdal_cli_violations(
     modules: list[tuple[str, ast.Module]],
     allowlist: dict[tuple[str, str, str], tuple[int, str]],
--- a/backend/tests/test_rule2_structural.py
+++ b/backend/tests/test_rule2_structural.py
@@ -73,6 +73,23 @@ Known limits (accepted trade-offs, same posture as the Rule-1 guard):
   territory.
 - Argv built dynamically (``cmd = [tool_var, ...]``) is not matched. Every
   GDAL CLI call in the codebase starts from a string-literal argv head.
+- A GDAL-headed literal counts as an argv only when its value ESCAPES —
+  handed to a call, returned, or yielded — or through the local it is bound
+  to (fix(#996)). Inert data (``SUPPORTED_TOOLS = ["gdalinfo", "ogrinfo"]``,
+  a constant only subscripted or compared) is not a command vector, and a
+  literal whose every element names a GDAL utility is a name list rather
+  than a command. Deliberately "escapes", not "reaches ``subprocess.*``":
+  most argvs here are built in one function and spawned in another
+  (``run_gdal(cmd, env=...)``), so a literal ``subprocess.*`` requirement
+  would blind the gate to the sites it exists for. Two or more elements are
+  required for the name-list rule, since ``("gdalinfo",)`` is
+  indistinguishable from a bare invocation.
+- Comprehensions and generator expressions are binding scopes (fix(#996)),
+  so their targets cannot shadow a name used elsewhere in the enclosing
+  function. The outermost iterable is excluded from that scope, matching
+  Python: it is evaluated before the comprehension's scope exists. They
+  remain transparent to CLI call-credit, where the question is whether the
+  scope's immediate body runs the helper and only def/lambda defer.
 - Wrapping is judged lexically, with two execution-order rules (codex round
   6): credit stops at def/lambda boundaries (a callable defined inside a
   wrapped block runs after the context exits), and within one ``with``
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1036, opened 2026-07-31
fix(sandbox): admit the canonical geodesic buffer by whole-template match

### LINKED ISSUE #1001: fix(sandbox): the canonical geodesic buffer is refused on the fail-closed allowlist (17 names), so every buffer question is dead

ishiland:
Readmits the canonical geodesic buffer to the NL→SQL sandbox. This is the one
live production defect in the backlog: reproduced against `9a788ba92`, every
buffer question in the surface is refused today.

## The shape of the fix

The prompt mandates `render_geodesic_buffer`'s output verbatim
(`sql_generator.py` renders it at import time), and that expression uses
sixteen function names neither allowlist carries, so `_check_function_allowlist`
rejected it on the first one the walk reached (`st_unaryunion`).

Allowlisting those names globally is what SEC-025 exists to prevent —
`st_dump`, `st_dumpsegments`, `st_segmentize` and `generate_series` are the
row/vertex amplification classes. #994 admitted them and bounded them with
per-call cost guards; #1002 spent three review rounds recalibrating those
guards and each fix opened the next hole. The last one settles it: the buffer
segmentizes an alias, `_pb_d0.c0`, several derived levels from its input, so
proving a call's argument is safe and admitting the canonical buffer are
contradictory under any argument-inspection scheme.

**So nothing is admitted per call.** A subtree is exempted from the name
allowlist and the cost guards only when it is exactly what the renderer emits
around its own input:

1. **extract loosely** — the input expression from the renderer's `AS g OFFSET 0`
   fence, the distance from its `ST_Buffer` call;
2. **verify exactly** — re-render the template around that same input and
   require the two ASTs to be equal.

Extraction is allowed to be wrong. Verification cannot be, because a mismatch
grants no exemption and the buffer stays refused. Numeric literals compare by
value, so `10000`, `10000.0` and `1e4` are the same distance while every
scaffold constant still has to agree exactly.

Three properties follow structurally rather than by argument inspection:

- **The scaffold's constants are not caller-chosen.** The segmentize lengths,
  band width and `ST_WrapX` bounds come from the reference render, so the
  `1e-100` vertex-amplification payload #1002 chased with a numeric floor
  simply is not the rendered template.
- **The alias-rebinding attack cannot match.** The template includes the
  derived-table scaffold that binds `_pb`, `_pb_d0`, `_pb_p`, so a real table
  bound to those names is not the rendered expression.
- **Amplification is bounded by one geometry, not by row count.** The banding
  series is bounded by a 4326 longitude span and the dumps run over a single
  fenced row geometry, because that is what the scaffold does — which is the
  property #994 tried and failed to prove one call at a time.

The input expression is **not** exempt. It is the model's, so its functions,
its cost guards and its tables are validated exactly as before; the buffer's
scaffold contributes no table of its own, which the admission tests assert.

## Security review notes

- `_BLOCKED_FUNCTIONS` is still checked **inside** a match. Nothing in the
  rendered scaffold can be a blocked function, and keeping the check
  unconditional means a renderer change could never smuggle one in. A test
  asserts the rendered scaffold holds none.
- The naive `ST_Buffer(geom::geography, N)::geometry` form is unchanged:
  `st_buffer` was always allowlisted, and the prompt still teaches it only as a
  negative (`test_no_bare_geography_buffer_taught_positively`). Nothing here
  reopens the ±180° regression #883/#900/#990 fixed.
- The distance literal is caller-chosen, as it already was for the
  independently-allowlisted `st_buffer`. It never reaches `ST_Segmentize`.
- Match attempts are capped at 8 per statement. Each re-renders and re-parses
  ~4 KB of SQL, measured at ~16 ms, so a statement stuffed with buffer-shaped
  scaffolds cannot turn the validator into the DoS. Past the cap nothing more
  is exempted, so the query is refused rather than slow — asserted by a test.

## Tests

`test_rendered_buffer_expression_is_refused_until_1001_lands` flips back to
asserting admission and is renamed `test_rendered_buffer_expression_validates`,
per the acceptance criteria.

**Every readmitted name ships a paired negative control.** Rather than a
hard-coded list, the controls are parametrized over the names derived from the
rendered buffer itself, so a renderer change moves the set and the controls
follow it. That yields exactly the sixteen names in the issue (the
seventeenth, `exists`, is #1017) and asserts each is still refused on its own.

Four of the five controls #1003 removed come back from `bf1696e5c`'s diff:
`test_rejects_table_scanning_unary_st_collect_however_aliased`,
`test_rejects_tiny_or_dynamic_segmentize_length`,
`test_rejects_table_scanning_st_dump`, and `ST_GeneratePoints` restored to
`test_rejects_unknown_postgis_function` — **alongside** `ST_MakeEnvelope`
rather than replacing it, since `st_makeenvelope` is not readmitted globally
here and both are valid unknowns.

The fifth, `test_allows_canonical_scale_segmentize`, is **not** restored, and
that is a deliberate consequence of the design rather than an oversight: it was
the positive control for a global `st_segmentize` admission with a numeric
floor, and this fix makes no such admission. `ST_Segmentize(geom, 20000)` on
its own stays refused, which is asserted as a third case inside
`test_rejects_tiny_or_dynamic_segmentize_length`. The positive control it
paired with is now the canonical-buffer admission itself.

New adversarial coverage: a disallowed, a blocked and an unbounded-generator
function in the buffer input; a canonical buffer not exempting a sibling
expression; a tampered segmentize length; a tampered band series; a scaffold
borrowing the renderer's aliases; a truncated lookalike; a buffer rendered
under a non-default alias; and the attempt cap in both directions.

## The coupling

`render_geodesic_buffer` now carries a comment pointing at the sandbox. Both
sides render through that one function, so a shape change re-admits its own new
shape and nothing else — but what the match cannot follow is a **new function
name** added to the renderer, since the exemption would then cover it. The
comment says exactly that, on the side that changes more often.

## Verification

```
cd backend && uv run pytest tests/test_ai_sql_prompt_935.py tests/test_sec025_sandbox_allowlist.py -v
cd backend && uv run pytest tests/test_sandbox.py tests/test_sandbox_tenant_schema.py \
  tests/test_sec025_sandbox_allowlist.py tests/test_ai_sql_prompt_935.py tests/test_sql_engine.py \
  tests/test_ai_chat.py tests/test_ai_chat_dataset.py tests/test_chat_narrative.py \
  tests/test_analysis_preview.py tests/test_analysis_materialize.py \
  tests/test_ai_chat_run_analysis.py tests/test_layering.py -q -n 4
# 461 passed
```

Closes #1001

### DIFF
--- a/backend/app/platform/analysis_sql.py
+++ b/backend/app/platform/analysis_sql.py
@@ -238,6 +238,17 @@ def render_dateline_safe(geom_expr: str, *, alias: str = "_dl") -> str:
     )
 
 
+# fix(#1001): this function has a SECOND consumer that most edits here will
+# not have in mind. The NL->SQL prompt embeds its output verbatim
+# (``processing/ai/sql_generator.py`` renders it at import time), and the SQL
+# sandbox admits the sixteen amplification-prone functions below ONLY inside a
+# subtree that is exactly what this renderer emits — ``_matches_canonical_buffer``
+# in ``platform/sandbox/validator.py`` re-renders the template and compares.
+# The match follows a shape change automatically, because both sides call this
+# function; what it cannot follow is a NEW function name that is itself unsafe
+# outside the template, since the exemption would then cover it. Weigh that
+# before adding one, and keep the default ``alias`` — a buffer rendered under
+# any other alias fails the match and the sandbox refuses it.
 def render_geodesic_buffer(
     geom_expr: str, distance: float, *, alias: str = "_pb"
 ) -> str:
--- a/backend/app/platform/sandbox/validator.py
+++ b/backend/app/platform/sandbox/validator.py
@@ -283,6 +283,18 @@ _ALLOWED_POSTGIS_FUNCTIONS: frozenset[str] = frozenset(
 )
 
 
+def _func_name(func: exp.Func) -> str:
+    """Canonical lowercase name of a function node.
+
+    Anonymous nodes carry the raw SQL identifier; named Func subclasses carry
+    sqlglot's own name, which may differ from the SQL keyword (see the
+    _ALLOWED_FUNCTIONS docs).
+    """
+    if isinstance(func, exp.Anonymous):
+        return func.name.lower() if hasattr(func, "name") else ""
+    return func.sql_name().lower() if hasattr(func, "sql_name") else ""
+
+
 def _validate_function_cost(func: exp.Func, fn_name: str, sql: str) -> None:
     """Reject function arguments that can amplify a one-row query into a DoS."""
     if fn_name in {"st_collect", "st_union"} and len(func.expressions) < 2:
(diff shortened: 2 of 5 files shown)


ishiland:
@codex review

ishiland:
Fixed the P1. The finding is right and the fix is structural rather than another argument check.

Both payloads reproduce on the previous head. `ST_Buffer(ST_SetSRID(ST_MakePoint(0,0),4326), 1000000000)` is a PLANAR buffer, so the radius is degrees — a two-billion-degree span, hundreds of millions of bands out of the scaffold's `generate_series`, billions of vertices out of its `ST_Segmentize`. And `ST_Transform(geom_4326, 3857)` reaches the same place with no large literal at all, by handing the scaffold metres.

Adding a cost bound on the input's functions is the units problem #1002 died on — the validator cannot tell a 1000-metre geography buffer from a 1000-degree planar one, and that is exactly the inference that drew a P1 there. So the exemption now requires the input to be a **stored** geometry: a bare column reference, or a scalar subquery projecting one. Both shapes the prompt teaches qualify; everything else is refused.

That also refuses a buffer of a buffer, which is bounded in fact (a geography buffer returns 4326, so its span is at most 360) but needs recursion to prove. Failing closed there is the right direction.

New tests cover the reported payload, the reprojected variant, the nested buffer, and a wrapped column, plus a positive control for the aliased-subquery input shape.

@codex review

ishiland:
Follow-up in the same vein: the prompt now states the bounded-input rule too.

The sandbox admits the rendered buffer only around a stored geometry, and nothing told the model that — so a reprojected or constructed `<GEOM>` would be generated and then silently refused. That is the same prompt-vs-validator disagreement this whole chain is about, one level down, so it is worth closing here rather than leaving as a surprise. Pinned by a test.

@codex review

ishiland:
Fixed the P1. The finding is right — a bare column reference proves nothing when an alias can launder the manufactured geometry back in, and the `ST_Transform` sits outside the exempt subtree so it clears the allowlist on its own. Reproduced in the CTE, derived-table and scalar-subquery forms.

The column's lineage is now followed through CTE and derived-table projections until it reaches a base table, and everything undecidable is refused: an unknown qualifier, a name bound in more than one place, a projection that is an expression rather than a column, an unqualified column with more than one table in scope, or a lineage deeper than four hops.

Two details that took a round to get right. A schema-less table node is a *reference* to a CTE rather than a base table, so it is skipped when counting bindings — otherwise `FROM x` and the CTE defining `x` made every CTE name look ambiguous and refused every legitimate pass-through. And the single-base-table test for an unqualified column has to look at the scope's own from/join clause rather than walking the subtree, because the buffer scaffold is itself a stack of aliased derived tables.

A two-hop pass-through through sibling CTEs is refused. That is a documented limit rather than a goal — sibling CTEs are not in each other's scope for this resolver — and stopping is a refusal, so the failure direction is right. The prompt teaches zero hops or one.

Tests cover all three laundering shapes plus a two-hop variant, the three lineages that should resolve, and the four undecidable cases.

@codex review

ishiland:
Both P1s fixed.

**Alias shadowing.** Right, and the mistake was resolving a qualifier against CTE *definitions* instead of against the from/join reference that actually carries the binding. `FROM bad AS x` binds `x` to `bad`; the resolver read the CTE literally named `x` and admitted the transform. Bindings now come only from from/join references, and a schema-less reference resolves through the CTE's own name — which also drops the double-count that had made the plain `FROM x` case look ambiguous against its own definition, so the aliased pass-through `FROM good AS x` now works too.

**Terminal column.** Also right, and it is the sharper of the two because it needs no attacker cleverness at all. Ingest keeps a dataset's original `geom` in whatever CRS it arrived in and adds `geom_4326` alongside — `processing/ingest/service.py:574` probes for both — so a Web Mercator `s.geom` has a 40-million-unit span and drives the scaffold exactly like a reprojection does. The terminal column must now be the managed `geom_4326`.

The prompt says so too, naming the original-geom case specifically, since that is the one a model would not guess.

Tests: the shadowing payload plus its aliased-pass-through control, and the projected base column in three shapes (qualified, unqualified, through a CTE).

@codex review

ishiland:
Rebased onto main now that #1024 (#1017) has landed, and codex round 4 is fixed on top.

**Conflict resolution.** #1017 and this PR both edit `_check_function_allowlist`: #1017 added the `exp.Exists` skip and kept the inline name extraction, this PR replaced that extraction with the `_func_name` helper. Resolved by keeping both — the Exists skip stays, and it now feeds `_func_name(func)`.

**Re-run rather than just resolved.** Confirmed both changes are actually present in the same tree (`exp.Exists` skip at validator.py:820, `_canonical_buffer_exempt_ids` at :805), then ran the sandbox and analysis sets against that head — **455 passed** across `test_sandbox`, `test_sandbox_tenant_schema`, `test_sec025_sandbox_allowlist`, `test_ai_sql_prompt_935`, `test_sql_engine`, `test_ai_chat`, `test_ai_chat_dataset`, `test_analysis_preview`, `test_analysis_materialize`, `test_layering`. Two individually-green branches touching the same allowlist is the shape that took main down before, so the combination is exercised rather than assumed.

**Codex round 4, both P1s fixed.**

A positional column-alias list renames a physical column into the managed name: `FROM data.cities AS s(gid, name, geom_4326)` binds `geom_4326` to whatever the third column is, which may be a projected `geom`. That needs the table's real column order to decide, which the validator does not have, so it fails closed.

And the binding lookup was scanning the whole statement, so an unrelated nested `(SELECT count(*) FROM data.cities s)` made an outer `s` look ambiguous and refused a buffer that was fine. Resolution is now lexical — a SELECT's own from/join bindings, walking outward for correlated references and for WITH clauses on enclosing SELECTs.

Walking WITH clauses outward also made sibling CTEs visible to each other, so a two-hop pass-through now resolves where it used to be refused, while the same two hops with a reprojection at the far end stay refused and a lineage past the hop cap still fails closed.

One documented consequence, in the code and in the prompt: a bare top-level `geom_4326` is now refused, because the scaffold interposes its own `(SELECT ... OFFSET 0) AS _pb` scope and deciding whether the name belongs to `_pb` or the outer table needs the column list. The prompt teaches the qualified form and now says so; the same bare name inside its own subquery still resolves.

@codex review

Issue #1010, opened 2026-07-31
test(analysis): verify materialized outputs carry tenant attribution end to end


ishiland:
Split out of #699, whose multi-tenant bullet was the one item that could not be finished in the same sitting as the rest of that issue.

## Correct the check first

#699 prescribed `grep tenant_id backend/app/processing/analysis/tasks.py` and claimed it was empty. It is not empty, and it does not test what it appears to test. The two hits are at `tasks.py:359-360`, inside `_output_table_adopted` (`:343-360`), which is a tenant-scoped **read** probe against `catalog.datasets`. A grep-based guard over that file passes whether or not the stamp works.

## How the stamp actually works

`Dataset.tenant_id` and `Record.tenant_id` are stamped by a **database trigger**, not by application code.

Migration `0018_tenant_insert_stamping` installs `catalog.stamp_current_tenant_on_insert()` as a `BEFORE INSERT` trigger on `catalog.records` and `catalog.datasets` (among six tables). It reads `current_setting('app.current_tenant', true)`; when that GUC is unset it returns the row unchanged (the single-tenant path), and when it is set it stamps `NEW.tenant_id` and rejects a mismatched caller-supplied value.

So the application deliberately never passes `tenant_id`. `create_dataset` constructs `Record(...)` at `backend/app/modules/catalog/datasets/domain/service_create.py:221-228` and `Dataset(...)` at `:233-248` with no tenant field, and the columns themselves are dormant nullable (`domain/models.py:226-229` for Record, `:444-447` for Dataset).

**The invariant to verify is therefore: at the moment the analysis worker inserts, its connection has `app.current_tenant` set.** The insert path is `materialize_analysis` (`backend/app/processing/analysis/tasks.py:836`, decorated `@tenant_task`) into `register_existing_table` at `:767`, which reaches `port.create_dataset` at `backend/app/processing/ingest/service.py:627-635` and then `service_create.create_dataset`.

## Why this needs a test rather than a reading

The mechanism is correct as far as static reading goes, and it survives on a detail that is easy to break by accident.

[...]

PR #1037, opened 2026-07-31
test(analysis): cover the remaining analysis edge-case fixtures

### LINKED ISSUE #699: test(analysis): remaining edge-case fixtures — by_field adversarial cases, empty/raster sources, job-store tests
From the 2026-07-25 analysis audit, **re-scoped 2026-07-30**, citations refreshed 2026-07-31 against `main` at `9a788ba92`.

The original body listed a geometry-type matrix that was empty at the time. Much of it has since been filled by the antimeridian batch (#884-#894, #897-#901, #924) without this issue being updated. The already-covered bullets are struck below so nobody re-derives work that exists.

Test counts have been stated in this body twice and been wrong both times, because the suites move faster than the issue does. They are gone; locate coverage by test name, not by count.

## Covered since this issue was filed, and do not redo

Verified present, with locations, so a future reader can confirm rather than trust. All line references are into `backend/tests/`.

- ~~anything near ±180 / the poles~~: `test_analysis_materialize.py:2459-3349`, three classes deep. `TestBufferAtDateline` (`:2459`) covers a buffer at lon 179.95 (`:2464`), `ST_ShiftLongitude` behaviour (`:2564`), the highest reachable latitude 85.06 (`:2603`), a pole-encircling buffer at 89.9 (`:2643`), and one feature with components at both seams (`:2677`). `TestBufferMultipartProjection` (`:2782`) and `TestWideSingleComponentBuffer` (`:3015`) carry the per-component projection and wide-span cases. This is now the most thoroughly covered corner of the suite.
- ~~any operation on a LINESTRING source~~: `test_analysis_materialize.py:1818-1977`. `test_clip_by_layer_does_not_fragment_a_line_at_mask_seams` (`:1818`, source `geometry_type="LINESTRING"` at `:1860`) asserts the output types at `:1898`; `test_clip_by_layer_keeps_touching_multiline_parts_separate` (`:1900`, `geometry_type="MULTILINESTRING"` at `:1941`) asserts at `:1976`.
- ~~NULL geometries in the source~~: `test_analysis_materialize.py:850-889` (`test_null_geometry_rows_are_dropped_from_output`), `:1259-1304` (2,025 NULL decoy rows, `:1271`), and `:1978-2066` (`test_clip_by_layer_bbox_only_rows_do_not_count_toward_size_cap`, whose docstring at `:1982-1987` states the NULL-vs-size-cap ordering).
- ~~mixed-geometry / GEOMETRYCOLLECTION sources~~: `test_analysis_materialize.py:2344` (`test_mixed_geometry_dissolve_stays_typed`), using `geometry_type="GEOMETRY"` at `:2377` and asserting the union does not register as a GEOMETRYCOLLECTION.
- ~~buffer on a POINT source~~: `_create_point_dataset_at` (imported at `test_analysis_materialize.py:39`) drives the antimeridian block and the mixed-geometry setup at `:2437`.
- ~~dissolve `by_field` over HTTP~~: **partially**: unknown column at `test_analysis_materialize.py:455`, `source_count` conflict at `:520-548`, non-groupable JSON column at `:474`. The adversarial cases below remain.
- ~~clip of a POINT source against a polygon mask~~: **partially**: `test_analysis_preview.py:876-899` (`test_degenerate_mask_row_stays_polygonal`) covers the zero-survivor case, a single point against a collapsed mask. The expecting-survivors variant remains.
- ~~clip of a LINE source expecting survivors~~: covered by the two clip-by-layer line tests cited above; both keep real geometry through the clip rather than testing an empty result.

## Genuinely open

**Backend, dissolve `by_field` over HTTP.** Validation exists and looks correct (`_validate_dissolve_by_field`, regex + `column_info` membership + a non-groupable type check, `backend/app/modules/catalog/datasets/api/router_analysis.py:179-204`) but the adversarial cases are untested end to end: an injection payload in the column name, NULL group values, and a reserved-word column.

**Backend, source-shape edges.**
- An empty (0-row) source dataset. `test_empty_result_fails_job` (`test_analysis_materialize.py:1555`) covers an empty *result*, which is a different path.
- An actual raster/DEM record rejected by preview/materialize/mask. The defense-in-depth path is exercised only via a synthetic `geometry_type=None` dataset (`test_analysis_preview.py:460-478`, `test_non_vector_dataset_rejected`); no fixture uses a real raster row. No frontend test supplies `is_dem: true`.
- Clip of a **POINT** source against a real polygon mask **expecting survivors**. The line equivalent is done; the point one is not.

**Backend, HTTP-level parameter bounds.** Negative, zero, and NaN `distance_meters` are covered at the SQL-builder and chat layers but not over HTTP. Likewise the title length bound, which is `AnalysisMaterializeRequest.title = Field(min_length=1, max_length=500)` at `backend/app/modules/catalog/datasets/domain/schemas.py:944`. There is a second 500 downstream on the worker's registration payload, `RegisterRequest.title` at `backend/app/processing/ingest/schemas.py:296-297`; the request bound is the one an HTTP test should be pinning, and the pair only agree by convention.

**Frontend.**
- `analysis-job-store` has **no test file at all** — `frontend/src/stores/__tests__/` holds `analysis-form-store.test.ts`, the auth-store pair, and the drawing/map-plugin/search stores, and nothing for the job store. Persistence, the auth-identity clear at `frontend/src/stores/analysis-job-store.ts:42-46`, and the watcher across a reload are all unpinned.
- The rendered truncation copy. `frontend/src/components/builder/__tests__/AnalysisPanel.test.tsx` sets `truncated: true` (`:405`, `:417`) but asserts nothing about the rendered `analysisTools.truncatedNotice*` strings.

**AI serialization.** A test asserting the four provider call sites route through `model_safe_tool_result`; only the pure function is covered (`test_ai_chat_run_analysis.py:448-465`). `test_extensions.py:239-302` (`test_pre_pr_import_surface_is_intact`) asserts the extension *binding name*, which is not the same guarantee. The four call sites are `backend/app/processing/ai/streaming.py:355`, `streaming.py:667`, `backend/app/platform/extensions/defaults_ai_anthropic.py:179`, and `backend/app/platform/extensions/defaults_ai_openai.py:216`.

## Split out

The multi-tenant item (verify the cloud overlay stamps `Dataset.tenant_id` at the `port.create_dataset` seam, then add a "materialized analysis output is tileable end-to-end" multi-tenant test) has been moved to its own issue. It is cross-repo, and it was the reason this issue could not be finished in one sitting. Everything remaining here is single-repo and landable.

## Acceptance criteria

- [ ] Every bullet under "Genuinely open" is either covered by a new test or explicitly deferred with a reason recorded here.
- [ ] The `by_field` injection case asserts the payload is rejected at enqueue with a 422, not merely that the query fails later.
- [ ] The raster fixture uses a real raster/DEM record, not `geometry_type=None`.
- [ ] `analysis-job-store` gains a test file covering persistence, the identity-change clear, and rehydration after reload.
- [ ] The `model_safe_tool_result` test fails if any one of the four call sites is changed to serialize the raw result.
- [ ] Coverage stays above the 80% `fail_under` in `backend/pyproject.toml`.

## Overlapping work

The `analysis-job-store` test file is also created by the cross-tab job-sync issue split out of #790. Whichever lands second extends the existing file rather than creating it.

## Relationship to #789

Near-disjoint in subject and worth keeping separate: this issue is the **data/geometry fixture matrix** (does the SQL behave against unusual geometry inputs), while #789 is the **seam and wiring matrix** (panel job lifecycle, `BuilderMap` draw guards, sandbox error to status mapping, e2e dissolve and drawn-mask clip). The only adjacency is the HTTP-level parameter bounds above versus #789's `_SANDBOX_STATUS` item.

Sequencing: do this one first. The fixtures here, particularly the raster record, the empty source, and the `by_field` adversarial cases, are the inputs #789's backend items need, so building them first makes that issue cheaper rather than duplicating it.

ishiland:
Everything under "Genuinely open" in #699. The geometry-type matrix the issue opened with was filled by the antimeridian batch (#884-#894, #897-#901, #924); what stayed open was the material around it.

## Adversarial dissolve `by_field`

- **Injection payload.** Planted in `column_info` so the membership half of `_validate_dissolve_by_field` passes and the identifier regex is what has to refuse it. Asserts 422 **at enqueue**, plus `defer` never awaited and no `IngestJob` row — a later failure would already have cost the queue wait and the caller's one analysis slot.
- **Reserved-word column.** `order` is a legal PostgreSQL identifier that merely has to be quoted, so it must NOT be refused as a precaution. Runs the whole path — enqueue validator, then the worker — and reads the grouped output back through the quoted column.
- **NULL group values.** A partially populated column yields a NULL-keyed output row rather than dropping those features, which is indistinguishable from data loss when read off the job status.

## Source shapes

- **A real raster record.** New `_create_raster_dataset` fixture builds what `tasks_raster._register_raster` actually writes: `Record(record_type='raster_dataset')`, a `raster_`-prefixed `table_name` backed by no physical table, `source_format='geotiff'`, and a `RasterAsset` carrying `is_dem`. Refused as a preview source, a materialize source, and a clip mask. The old synthetic `geometry_type=None` dataset is a shape no ingest path produces.
- **A 0-row source.** Previews as zero features with a null bbox, and fails the materialize job for both query shapes. Dissolve is the one worth having: its no-`GROUP BY` aggregate still returns one row, with a NULL geometry, so what stops it is the in-CTAS filter plus the post-CTAS DELETE. Distinct from `test_empty_result_fails_job`, which covers an empty *result*.
- **Clip keeps points.** The line equivalents assert survivors; the point case only ever asserted the zero-survivor degenerate-mask path, which is the dimension where `ST_CollectionExtract` takes its narrowest branch.

## HTTP-level parameter bounds

Negative, zero and non-finite `distance_meters`, and the `title` bound with its accepted boundary at exactly 500. The non-finite case pins its message: `allow_inf_nan` defaults True, so NaN and the infinities parse as floats and fail `le` (not `gt`, which the finite cases fail) — if that ever changed, the builder's own `isfinite` check would silently stop being the second line.

## Tool-result serialization

The existing `model_safe_tool_result` tests pass whether or not the provider call sites use it, and `test_extensions.py` asserts the extension *binding name*, which is a different claim. A structural test now walks the three modules that build tool-result messages and requires every dict carrying a `tool_use_id`/`tool_call_id` to wrap its `content` payload in `model_safe_tool_result` before `json.dumps` — with an exact site count, so a new provider path has to opt in rather than ride the rule.

Verified by mutation: removing the wrapper at `defaults_ai_openai.py:216` fails the test naming that line.

## Frontend

- `src/stores/__tests__/analysis-job-store.test.ts` is new: persistence, rehydration after a reload, the version-mismatch discard, and the identity-change clear including the token rotation it has to survive.
- The truncation toast now pins which notice renders and what it interpolates — the `truncatedNoticeTotal` and count-only branches were interchangeable before. `t` became a spy with an unchanged return contract, so the existing raw-`{{...}}` template assertions in that file still hold.

`AnalysisPanel.tsx` itself is untouched; only its test file changed.

## Not in scope

The multi-tenant bullet moved to #1010 before this was picked up.

## Verification

```
uv run pytest tests/test_analysis_materialize.py tests/test_analysis_preview.py \
  tests/test_ai_chat_run_analysis.py -q        # 167 passed
uv run ruff check tests/ && uv run ruff format --check tests/

cd frontend
npx vitest run src/stores/__tests__/analysis-job-store.test.ts \
  src/components/builder/__tests__/AnalysisPanel.test.tsx
npm run lint && npm run typecheck
```

Tests and fixtures only — no schema, API or security impact. No new `t()` keys, so locale parity is unaffected.

Closes #699


ishiland:
@codex review

ishiland:
@codex review

ishiland:
Merge-order note for whoever picks this up if I am not around.

This conflicts with #1030 (also mine) in exactly one place: the import block at the top of `backend/tests/test_analysis_preview.py`. Both PRs add imports there and nothing else overlaps — `git merge-tree` reports a single conflict hunk between the two branches.

#1030 should land first. The resolution is purely additive; the merged block is:

```python
import math
import uuid
from unittest.mock import AsyncMock, patch

import pytest
from httpx import AsyncClient
from sqlalchemy import text
from sqlalchemy.ext.asyncio import AsyncSession

from app.modules.catalog.datasets.api import router_analysis
from app.modules.catalog.datasets.domain.models import Dataset, Record
from app.modules.catalog.datasets.domain.schemas import AnalysisPreviewRequest
from app.modules.catalog.datasets.domain.service import build_preview_sql
from app.platform.sandbox.schemas import SandboxError

from tests.factories import create_dataset, get_user_id
```

One follow-up once #1030 is in: `test_title_at_the_cap_is_accepted` here hand-releases the per-user job slot at the end. #1030 adds an autouse teardown that does exactly that for every test in the module, so those three lines become redundant and should go.

ishiland:
@codex review

ishiland:
Fixed, and the base is current now.

The set of module names collapsed `streaming.py`'s two messages into one entry, so losing either one left the anchor satisfied by the other — exactly the silent-pass the anchor exists to prevent. Counted per module now (`streaming.py: 2`, each adapter: 1). Verified by renaming one of the two streaming keys: the new check fails with `(1, 2)`, the set-based one passed that mutation.

Also merged current main in, so the next CI result is against a base that exists. The previous red on this PR was a 15:36 run whose merge-base predates both #1063 and the thirteen-PR batch, failing on `test_ingest_progress.py` with `unexpected connection_lost()` — a file this diff has no mechanism to touch.

Head is `d5338df7e`. Local: 204 passed across the analysis, chat and layering suites; ruff clean.

ishiland:
@codex review

ishiland:
Real, and it would have broken the frontend gate. Fixed at `a0f3b0965`.

Merging main brought in the ownership stamp #1043 added to `setJob`, so two assertions comparing the persisted job against the three-field fixture saw a fourth key. `toMatchObject` rather than `toEqual` in both: these tests are about the job round-tripping, not about who owns it — the stamp has its own tests in `analysis-job-store.cross-tab.test.ts` — and pinning the exact shape makes every future field addition fail three unrelated tests.

The third `toEqual(job)` in that file stays. It is the rehydration case, which seeds storage with the unstamped fixture directly, so the value never passes through `setJob` and deep equality is the honest assertion there.

Mine to have caught before you did: I ran the backend suites after the merge and not the frontend one, while the merge carried a frontend type-shape change from a PR I had written myself. Full frontend gate now — lint, typecheck, 368 files and 4368 tests passing.

ishiland:
@codex review

ishiland:
Correct — the test was passing for the wrong reason. Fixed at `b9440a7bc`.

`previewAnalysis` is a shared mock that nothing clears, so waiting for it to have been called at all resolved on an earlier test's call and the negative assertion ran before this request's `onSuccess`. It now waits on a per-test `onPreviewResult` spy, which fires in the same synchronous success handler that raises the notice — so once it has been called, a toast would already have been raised.

Verified rather than assumed: forcing the notice on unconditionally makes the test fail, where the old form passed that same mutation.

Full frontend gate: lint, typecheck, 368 files, 4368 tests.

ishiland:
@codex review

ishiland:
Rebased onto current main — head `a97c8a989`.

One conflict, in `test_ai_chat_run_analysis.py`'s import block, additive on both sides: this branch adds `Counter` and `Path` for the per-module serialization anchor, #1071 (now on main) adds `patch` for its mask-size tests. All three kept, nothing else collides.

#1071 added clip cases to the same file but no new tool-result serialization sites, so the structural test's per-module counts are unaffected — checked by running it rather than reasoning about it.

Verified: 249 passed across the analysis, chat, layering, extensions and processing-port suites under `-n 4 --dist loadgroup`; 96 passed on the frontend side (`AnalysisPanel.test.tsx`, `analysis-job-store.test.ts`).

**One thing still pending, so this is not the last rebase.** #1030 is not on main yet and conflicts with this branch in `test_analysis_preview.py`'s import block — a separate, purely additive hunk I recorded earlier. When #1030 lands I will rebase again and, in the same pass, drop the now-redundant hand-release of the job slot in `test_title_at_the_cap_is_accepted`: #1030's autouse teardown makes it dead code. It is still needed today, because that fixture is not on this branch or on main.

ishiland:
@codex review

ishiland:
Second rebase done — head `8997ff51a`, and it now includes #1083.

**The conflict** was the one I measured when these branches first diverged: `test_analysis_preview.py`'s import block, additive both sides. #1030 adds `router_analysis`, this branch adds `Dataset, Record`. Both kept.

**Dropped the hand-release** of the per-user job slot at the end of `test_title_at_the_cap_is_accepted`. #1030's autouse teardown now fails every analysis job a test leaves behind, so those three lines are dead code — and leaving them would read as intentional to the next person adding a test here.

**Ran F's gate rather than assuming it passes.** `test_fixture_scope_structural.py` — 2 passed. This branch adds no autouse fixtures, so I expected it to be clean, but that is exactly the kind of expectation that has been wrong twice today.

Verification: 229 passed across the analysis, chat, layering and fixture-scope suites under `-n 4 --dist loadgroup`. Frontend full gate: lint, typecheck, 368 files, 4381 tests. `origin/main` is an ancestor, and the diff against it is only this PR's five files.

codex re-triggered.

ishiland:
@codex review

ishiland:
Real, and it is the same class as the previous two on this test — each round has removed a way the guard could miss rather than a way it could over-report. Fixed at `a3d186137`.

A dict literal is only the shape these happen to take today. `dict(tool_call_id=..., content=...)`, or a provider's own message class built with the same keywords, is the identical payload and was invisible to a literal-only walk. So a fourth adapter written in either style would have serialized an untrimmed result while all three known-module counts stayed satisfied — the exact silent pass this exists to prevent.

It now matches on the **key names** rather than the construction style, because the names are what the provider APIs fix and they read the same whether they arrive as strings or keywords. `ast.Dict` and `ast.Call` both go to one matcher.

Verified both new styles rather than reasoning about them:

- a probe module using `dict(role=..., tool_call_id=..., content=json.dumps(result))` → reported at its file and line
- a probe using a `ToolMessage(tool_use_id=..., content=...)` class → reported at its file and line

Neither was caught before the change. Both probes removed.

229 passed across the analysis, chat, layering and fixture-scope suites under `-n 4 --dist loadgroup`.

On the round economics, since it is worth stating: this one was free by the definition we have been using — widening detection can only make the gate report more, and a false alarm here is visible and cheap while a silent miss is the failure that matters. If a later round asks it to report *less*, that is the point where it should become its own PR rather than round five of this one.

ishiland:
@codex review

Issue #1013, opened 2026-07-31
ops(analysis): promote the hardcoded analysis timeouts to settings


ishiland:
Split out of #696 (item 3), which was a cluster of four unrelated resource-governance decisions. This one stands alone.

## The gap

Two analysis timeouts are module constants with no way for an operator to change them:

- `MATERIALIZE_TIMEOUT = "300s"` (`backend/app/processing/analysis/tasks.py:58`), applied via `SET LOCAL statement_timeout` before the materialize CTAS (`:697`). It carries an ordinary comment at `:56-57` that already says the intent: "Hardcoded ceiling; promote to persistent-config if operators hit it."
- `DEFAULT_TIMEOUT_MS = 10_000` (`backend/app/platform/sandbox/executor.py:26`), the preview sandbox budget.

300 seconds covers roughly 150k to 600k buffered polygon rows. An ordinary one-million-parcel buffer fails, and the operator's only recourse is to edit Python and rebuild the image.

There is a third in the same file that should be decided in the same pass rather than left as the odd one out: `REGISTRATION_TIMEOUT = "600s"` (`tasks.py:66`, applied at `:763`). It exists because the commit after the CTAS ends the transaction and its `SET LOCAL` with it, so registration needs its own budget (#692). Promoting two of the three and leaving one hardcoded is the kind of inconsistency that costs someone an afternoon later.

## Scope: four surfaces per value, not one

This is the reason it was worth splitting out. A promoted knob is not a one-line change:

1. **The `Settings` field** in `backend/app/core/config.py`.
2. **Validation.** Both values have meaningful bounds and both are dangerous when wrong. `statement_timeout = '0'` disables the timeout entirely rather than erroring, so a zero must be rejected at boot, not silently accepted into an unbounded statement. Negative and unparseable values likewise. Follow the existing validator style in `config.py` (for example `validate_worker_queues` at `:484-492`), which raises from a `field_validator` so a bad value fails startup.
3. **`.env.example`**, with a comment saying what the budget covers and what raising it costs.
4. **The docs site entry**, since `.env.example` is not where operators look first.

Anything that skips surfaces 2 through 4 has moved the constant without making it usable.

## What raising the ceiling does and does not buy

[...]

PR #1038, opened 2026-07-31
feat(ingest): cap total rows and cells on the parquet ingest path

### LINKED ISSUE #948: feat(ingest): cap total rows and cells on the parquet ingest path
Extracted from #569 (item 4), now closed. Deferred by design when the rest of that batch shipped in #575.

## The gap

`load_parquet_to_postgis` (`backend/app/processing/ingest/parquet.py:326-429`) batches inserts through `pf.iter_batches(batch_size=_INSERT_BATCH_ROWS, ...)` (`:412`) but never checks a total. There is no row cap, no cell cap (rows times columns), and no wall-clock budget. A file that passes the upload size check can still expand into an arbitrarily large table, and the only backstop is the 500 MB upload cap, which parquet's compression makes a weak proxy for row count.

**Correction to the original text.** This is not a parity gap with ogr2ogr, because ogr2ogr has no row cap either. What that path has is wall-clock timeouts: `OGRINFO_TIMEOUT_SECONDS = 300`, `OGR2OGR_FILE_TIMEOUT_SECONDS = 3600`, `OGR2OGR_SERVICE_TIMEOUT_SECONDS = 1800` (`backend/app/processing/ingest/ogr.py:144-146`). Both paths additionally share the archive-bomb ceilings in `backend/app/processing/ingest/validation.py:53-56` (`MAX_COMPRESSION_RATIO`, `MAX_DECOMPRESSED_BYTES`, `MAX_ARCHIVE_ENTRIES`, `MAX_CENTRAL_DIRECTORY_BYTES`).

The real asymmetry is narrower and worse: a runaway ogr2ogr is a subprocess that gets killed on a timer, while the parquet path runs in-process in the worker with no backstop of any kind. Nothing stops it until the table is written or the container dies.

## Why it stayed open

The original finding called it an extreme edge case, and it is: reaching it takes a densely compressed file near the upload ceiling. It was explicitly deferred to a future ingest-budget pass rather than bolted onto the #575 batch, and that pass has not happened.

## Suggested shape

**Fail fast from the file metadata, not from a running counter.** Parquet carries the row count in its footer, and this module already reads it: `pf.metadata.num_rows` is used at `parquet.py:267` to populate `feature_count`. Checking it before the insert loop rejects an oversized file in milliseconds, before a single row is written, and is far easier to test than a counter that has to be driven through several batches to trip. Counting during iteration would only be needed if the footer could not be trusted, and nothing here suggests it cannot.

Check both a row ceiling and a cell ceiling. Cells (`pf.metadata.num_rows` times `len(plan)`, the columns actually being inserted) bound the work better than rows alone, because a 400-column file is a very different load from a 4-column one at the same row count. Raise `IngestionError` with a message naming the limit and the observed value, the same way the geometry-only guard at `parquet.py:397-405` fails.

**Put the caps in module constants, not `Settings`, for this change.** They belong beside `_INSERT_BATCH_ROWS` (`parquet.py:30`). A `Settings` field is four surfaces (field, validation, `.env.example`, docs site) for a limit no operator has yet asked to tune, and #696 item 3 is the issue that establishes the pattern for promoting ingest and analysis constants to settings. If these caps turn out to need tuning in the field, promote them there and follow that pattern rather than inventing a second one here.

### Starting numbers to review

Both are proposals, not decisions. Pick or adjust them in review; the point is that the PR should not invent them silently.

- Rows: **5,000,000**. Well above any real upload seen so far, low enough that hitting it is a genuine signal.
- Cells: **100,000,000** (rows times inserted columns). At the row cap this permits 20 columns; a wider table trips the cell cap sooner, which is the intent.

## How to verify

Build the fixtures with a small pyarrow script rather than committing large files, and monkeypatch the constants down to a handful of rows for the test rather than generating a five-million-row file. Being module constants makes that a one-line `monkeypatch.setattr`, which is a second reason to keep them there for now.

The test file already exists: `backend/tests/test_ingest_parquet.py`, and `test_geometry_only_without_geometry_raises` (`:60`) is the sibling to copy, since it exercises the same "construct a parquet file, run it through the loader, assert `IngestionError`" shape.

- A fixture over the cap fails with a message naming the limit, and no table is created.
- A fixture just under the cap still succeeds.
- A narrow file over the cell cap but under the row cap, or the reverse, so each ceiling is exercised on its own rather than both tripping together.

ishiland:
`load_parquet_to_postgis` batched inserts through `pf.iter_batches` but never checked a total: no row cap, no cell cap, no wall-clock budget. A file that passes the 500 MB upload check can still expand into an arbitrarily large table, and parquet's compression makes upload size a weak proxy for row count.

The asymmetry with ogr2ogr is narrower than "no row cap" — ogr2ogr has no row cap either. What that path has is wall-clock timeouts (`OGRINFO_TIMEOUT_SECONDS`, `OGR2OGR_FILE_TIMEOUT_SECONDS`, `OGR2OGR_SERVICE_TIMEOUT_SECONDS` in `ogr.py:144-146`) on a **subprocess that gets killed on a timer**. This loader runs in-process in the worker with no backstop of any kind. Nothing stops it until the table is written or the container dies.

## What changed

Two ceilings, read from the parquet footer before the DDL runs:

- `_MAX_TOTAL_ROWS = 5_000_000`
- `_MAX_TOTAL_CELLS = 100_000_000` (rows x inserted columns)

`pf.metadata.num_rows` is already read at `:267` to populate `feature_count`, so checking it costs nothing and rejects an oversized file in milliseconds. A running counter would have to be driven through several batches to trip and would leave a partial table behind; the check sits before `CREATE TABLE`, so nothing is created.

At the row cap the cell cap permits 20 columns. A wider table trips the cell cap sooner, which is the intent — a 400-column file is a very different load from a 4-column one at the same row count.

**Numbers.** The issue offered these as proposals to settle in review rather than invent silently. I kept both: 5M rows is well above any real upload seen so far and low enough that hitting it is a genuine signal, and 100M cells falls out of it at a reasonable column count. Cells count `len(plan)`, the columns actually being inserted (skipped geometry/covering columns are excluded, as is geometry itself under `include_geometry=False`).

**Module constants, not `Settings`.** They live beside `_INSERT_BATCH_ROWS`. A `Settings` field is four surfaces (field, validation, `.env.example`, docs site) for a limit no operator has yet asked to tune, and #1013 is the issue that establishes the promotion pattern if these ever need it. Being module constants is also what makes the tests a one-line `monkeypatch.setattr` instead of a five-million-row fixture.

## Tests

`TestParquetIngestBudget` in `backend/tests/test_ingest_parquet.py`, following `test_geometry_only_without_geometry_raises`:

- **row cap alone** — 6 rows x 1 column is 6 cells, well under the cell cap, so only the row ceiling can fire
- **cell cap alone** — 4 rows is under a row cap of 100, but 4 x 5 = 20 cells is over a cell cap of 10
- **just under both** — the ceilings are exclusive; a file exactly at both still ingests and its rows land

The two rejection tests also assert `to_regclass` finds no table afterwards, which is the property that checking the footer up front buys over a running counter.

## Verification

```
cd backend && uv run pytest tests/test_ingest_parquet.py -v          # 28 passed
cd backend && uv run pytest tests/test_layering.py -q                # 34 passed
cd backend && uv run ruff check . && uv run ruff format --check .    # clean
```

Closes #948

### DIFF
--- a/backend/app/processing/ingest/ogr.py
+++ b/backend/app/processing/ingest/ogr.py
@@ -119,6 +119,18 @@ class IngestionError(Exception):
     """Raised when an ingestion subprocess fails."""
 
 
+class IngestBudgetExceededError(IngestionError):
+    """Raised when a source exceeds an ingest resource ceiling (fix(#948)).
+
+    A subclass so the preview route can surface THIS message verbatim without
+    widening the generic ``IngestionError`` handler, which also carries GDAL
+    subprocess output. The text is server-authored and names the limit, the
+    observed value, and what to do about it — telling a user only that their
+    file "may be malformed or unsupported" when it is merely too large leaves
+    them with nothing to act on. Raised from the parquet path today.
+    """
+
+
 def validate_layer_name_argv(layer_name: str) -> None:
     """Reject option-like layer names before they reach a GDAL argv.
 
--- a/backend/app/processing/ingest/parquet.py
+++ b/backend/app/processing/ingest/parquet.py
@@ -66,17 +66,17 @@ def _check_ingest_budget(num_rows: int, num_columns: int) -> None:
     fix(#948). Raises ``IngestionError`` naming the limit and the observed
     value, the same way the geometry-only guard fails.
     """
-    from app.processing.ingest.ogr import IngestionError
+    from app.processing.ingest.ogr import IngestBudgetExceededError
 
     if num_rows > _MAX_TOTAL_ROWS:
-        raise IngestionError(
+        raise IngestBudgetExceededError(
             f"Parquet file has {num_rows:,} rows, above the "
             f"{_MAX_TOTAL_ROWS:,}-row ingest limit. Split the file or load a "
             "subset."
         )
     cells = num_rows * num_columns
     if cells > _MAX_TOTAL_CELLS:
-        raise IngestionError(
+        raise IngestBudgetExceededError(
             f"Parquet file has {cells:,} cells ({num_rows:,} rows x "
             f"{num_columns:,} columns), above the {_MAX_TOTAL_CELLS:,}-cell "
             "ingest limit. Split the file or drop columns you do not need."
--- a/backend/app/processing/ingest/parquet.py
+++ b/backend/app/processing/ingest/parquet.py
@@ -29,6 +29,33 @@ if TYPE_CHECKING:
 
 _INSERT_BATCH_ROWS = 5000
 
+# fix(#948): ingest budget for this path.
+#
+# The 500 MB upload cap is a weak proxy for how much work a parquet file
+# represents, because parquet compresses well — a densely compressed file just
+# under the ceiling can still expand into an arbitrarily large table. That
+# matters more here than on the ogr2ogr path: a runaway ogr2ogr is a subprocess
+# killed on a timer (OGR2OGR_FILE_TIMEOUT_SECONDS and friends in ogr.py), while
+# this loader runs in-process in the worker with no backstop of any kind.
+#
+# Both ceilings are checked from the file footer before a single row is
+# written, not from a running counter — parquet carries the row count in its
+# metadata, so an oversized file is rejected in milliseconds and no table is
+# created.
+#
+# Deliberately module constants rather than Settings fields: a Settings knob is
+# four surfaces (field, validation, .env.example, docs site) for a limit no
+# operator has asked to tune. #1013 establishes the pattern for promoting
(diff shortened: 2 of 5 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

Issue #1007, opened 2026-07-31
test(ai): tool-choice eval harness so a chat routing regression fails a run


ishiland:
## Gap

The repo has AI evals, but none of them can catch a tool-routing regression. `backend/tests/evals/test_sql_generation_evals.py` calls `generate_sql()` directly and executes the result through the sandbox. It asserts on the SQL text and on the executed values. It never runs the tool-calling loop, so there is no assertion anywhere about which tool the model chose for a given user message.

That is the layer where routing bugs live. #549 is one: "show me X" lands on `set_filter` when it should land on `query_data`, and the fix is a prompt or tool-description edit whose whole risk is that a later edit quietly reverses it. Nothing in CI or in `make ai-evals` would notice.

## What to build

A tool-choice eval harness alongside the existing SQL evals: drive the live chat loop with a user message and a tool set, then assert on the tool name the model emits.

- Drive the real path, not a reconstruction. The tool set comes from `select_chat_tools(can_edit, has_map)` (`backend/app/processing/ai/tools.py:397`), and the system prompt from the same builder the chat service uses, so an edit to either is what the eval sees.
- Assert on the emitted tool name (and, where it matters, that no editing tool was emitted at all). Do not assert on the response prose.
- Cover both directions of each ambiguous pair, since a prompt edit that fixes one direction commonly breaks the other. The #549 pair is the first case: "Show the ADA accessible stations served by the A line" must emit `query_data`, "Filter to ADA accessible stations on the A line" must emit `set_filter`.
- Cover the neighbouring `run_analysis` vs `query_data` split too, which was settled in prose and has the same unprotected shape.
- Cover the surface variations: the same message under `CHAT_TOOLS_READONLY` and under `has_map=False` must not emit a tool that surface does not offer.

## Constraints

Same cost and gating shape as the existing evals. This is live-provider, it spends real tokens, and it belongs behind the `RUN_AI_EVALS=1` skip and the `make ai-evals` target rather than in the default test run. Follow the flake policy already written into the SQL evals' module docstring: temperature 0.0, and assertions that leave the model freedom everywhere it legitimately has it. A tool-name assertion is a stricter thing to pin than a SQL predicate, so expect to spend time deciding which cases are stable enough to assert and which are only worth reporting.

## Why this is separate from #549

[...]

PR #1039, opened 2026-07-31
fix(ai): give verb classification one owner in builder chat

### LINKED ISSUE #549: Builder chat: 'show me X' phrasing routes unpredictably between map edits and data queries

ishiland:
## What changed

Three sites, one rule.

**`chat_service.py` — the system prompt now owns verb classification and says
so.** The QUESTION/CHANGE split that already lived there gains the ambiguous
verbs spelled out, the reported phrasing pair as worked examples, and bare
"show" settled explicitly as a QUESTION verb with only the narrowing forms
("show only", "hide everything else", "just the ...") counting as CHANGE.
Geometry operations outrank both lists, so "show the centre point of each
parcel" still reaches `run_analysis`.

**`tools.py`, `run_analysis` — stops classifying phrasing.** It dropped
"show the centre point of each parcel" and the "Use query_data instead for
questions ANSWERED from the data" cross-reference, and now describes only what
the tool does and what its output looks like.

**`tools.py`, `set_filter` — gains a pointer, not a second copy.** It carried
no routing guidance at all, which is how "show me the ..." landed on a
persistent layer filter. It now says it is the map-edit path and defers to the
prompt's verb classes.

## Why not just add the guidance

The issue's original premise was wrong and the body says so: the QUESTION/CHANGE
prose already existed at `chat_service.py:268-273`. It did not catch this
phrasing because "show" is in neither of its buckets, while `run_analysis`'s
description taught "show" as a transform verb. Adding a third copy of the rule
would have left three sites able to contradict each other.

## The structural test

`test_no_tool_description_routes_across_the_split` fails any tool description
that names `query_data` or `run_analysis` — the two tools on the
question/transform side of the split. That is deliberately narrower than "no
description may name any sibling", because two existing cross-references are
not verb classification and stay:

- `set_style` → `set_data_driven_style` is a **capability** difference; both
  are CHANGE tools.
- `add_layer` → `search_datasets` is a **sequencing** step; you need a
  `dataset_id` first.

Neither says anything about which user phrasing wins. The exemption is by name
with the reason in the test, so a future edit has to argue with it.

## Scope and blast radius

Only the builder surface with edit rights can reach `set_filter`.
`select_chat_tools` hands `CHAT_TOOLS_READONLY` (`query_data` + `run_analysis`)
to any caller that cannot edit, and dataset-scoped chat additionally passes
`has_map=False`, which drops `run_analysis` too. So the ambiguity is not
reachable from viewer chat or dataset chat, and the fix cannot regress them:
`set_filter` is not in the read-only set, so those surfaces see only the
prompt's verb classes, and every class they can act on routes to `query_data`.

## Verification

```
cd backend && uv run pytest tests/test_ai_chat_verb_routing_549.py -v
cd backend && uv run pytest tests/test_ai_chat_verb_routing_549.py tests/test_chat_narrative.py \
  tests/test_sql_engine.py tests/test_ai_chat.py tests/test_ai_chat_actions_phase1135.py \
  tests/test_ai_chat_dataset.py tests/test_ai_chat_run_analysis.py tests/test_layering.py -q -n 4
# 169 passed
```

**Live provider verification is NOT included and has to be run by a human.**
The issue's recipe needs the builder against a live provider, and there is no
tool-selection harness yet (`test_sql_generation_evals.py` calls
`generate_sql()` directly and never runs the tool-calling loop; building one is
#1007). This branch was developed in a worktree, where the running stack serves
`main`'s code, so any result from it would be meaningless. The recipe, for
whoever runs it:

1. Open a map in the builder as a user with edit rights, on a layer whose
   columns support the phrasing.
2. In fresh conversations, send "Show the ADA accessible stations served by
   the A line" and "Filter to ADA accessible stations on the A line".
3. Confirm the emitted tool is `query_data` for the first and `set_filter` for
   the second.
4. Repeat "show the centre point of each parcel" and confirm it still reaches
   `run_analysis`.
5. Repeat the "show" phrasing against viewer chat and dataset chat to confirm
   neither surface changed behaviour.

Closes #549

### DIFF
--- a/backend/app/processing/ai/chat_service.py
+++ b/backend/app/processing/ai/chat_service.py
@@ -274,10 +274,13 @@ You are a map editing assistant. The user has a map with these layers:
     -- "find datasets about flood zones", "is there a layer for parcels",
     "search for census data". Use search_datasets. query_data can only reach
     layers already on the map, so a catalog lookup sent there fails.
-  - The object is a LAYER, or its labels -- "show the layer", "show @Parks",
-    "hide the labels", "show only @Parks" when @Parks is a LAYER among
-    several. Use toggle_visibility. set_filter cannot hide a sibling layer;
-    it only filters features inside one.
+  - The object is a LAYER -- "show the layer", "show @Parks", "hide @Roads",
+    "show only @Parks" when @Parks is a LAYER among several. Use
+    toggle_visibility. set_filter cannot hide a sibling layer; it only
+    filters features inside one.
+  - The object is a layer's LABELS -- "hide the labels", "turn the labels
+    off", "label them by name". Use set_label, with column null to turn
+    labels off. toggle_visibility would hide the layer's features too.
   - The object is the DATA in a layer, and the request asks for something
     ANSWERED from it (counts, statistics, spatial relationships, distances,
     finding features). QUESTION verbs -- show, find, list, which, what,
@@ -290,7 +293,8 @@ You are a map editing assistant. The user has a map with these layers:
   - "show" is a QUESTION verb by default: "show me the ADA accessible
     stations" asks for a list, not a map change. Three exceptions, all
     CHANGE, and all of them have the MAP as the object rather than the data:
-    - a LAYER or its labels, per the rule above -- use toggle_visibility.
+    - a LAYER, per the rule above -- use toggle_visibility; its LABELS --
+      use set_label.
     - NARROWING to a feature PREDICATE -- "show only the accessible ones",
       "just the ones on the A line" -- use set_filter.
     - BROADENING an existing filter -- "show all", "show everything again",
--- a/backend/app/processing/ai/chat_service.py
+++ b/backend/app/processing/ai/chat_service.py
@@ -276,9 +276,18 @@ You are a map editing assistant. The user has a map with these layers:
     make -- ask to CHANGE the map (colors, filters, labels, visibility,
     add/remove layers). Use the map editing tools.
     "Filter to ADA accessible stations on the A line" is a CHANGE.
-  - "show" on its own is a QUESTION verb, not a filter request. Only the
-    explicit narrowing forms -- "show only", "hide everything else", "just the
-    ..." -- are CHANGE.
+  - "show" is a QUESTION verb by default: "show me the ADA accessible
+    stations" asks for a list, not a map change. Three exceptions, all
+    CHANGE, and each is recognizable because the object is the MAP rather
+    than the data:
+    - a LAYER or its labels is the object -- "show the layer", "show
+      @Parks", "hide the labels" -- use toggle_visibility.
+    - NARROWING -- "show only", "hide everything else", "just the ..." --
+      use set_filter.
+    - BROADENING an existing filter -- "show all", "show everything again",
+      "show all the features" -- use set_filter with a null expression to
+      CLEAR the filter. Answering that one as a question leaves the
+      persistent filter in place, which is not what was asked.
   - A request that names a geometry OPERATION is a TRANSFORM whichever verb
     introduces it, and outranks both lists above.
 - query_data takes a natural language question -- the server generates and
--- a/backend/app/processing/ai/chat_service.py
+++ b/backend/app/processing/ai/chat_service.py
@@ -265,10 +265,22 @@ You are a map editing assistant. The user has a map with these layers:
   Example filters: ["==", ["get", "status"], "active"], ["all", [">", ["get", "population"], 50000], ["==", ["get", "state"], "CA"]]
 - For compound requests that include both a question and a map change, use both query_data and editing tools in a single response.
 - To add a new layer, first use search_datasets to find the dataset, then use add_layer with the dataset_id.
-- When the user asks a QUESTION about their data (counts, statistics, spatial
-  relationships, distances, finding features), use the query_data tool.
-- When the user asks to CHANGE the map (colors, filters, labels, visibility,
-  add/remove layers), use the map editing tools.
+- Tool selection is decided HERE, by the verb the user used. The tool
+  descriptions say what each tool does; they do not decide which phrasing
+  wins.
+  - QUESTION verbs -- show, find, list, which, what, where, how many, how much
+    -- ask for something ANSWERED from the data (counts, statistics, spatial
+    relationships, distances, finding features). Use the query_data tool.
+    "Show me the ADA accessible stations served by the A line" is a QUESTION.
+  - CHANGE verbs -- filter, style, color, label, hide, show only, change, set,
+    make -- ask to CHANGE the map (colors, filters, labels, visibility,
+    add/remove layers). Use the map editing tools.
+    "Filter to ADA accessible stations on the A line" is a CHANGE.
+  - "show" on its own is a QUESTION verb, not a filter request. Only the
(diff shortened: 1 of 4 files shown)


ishiland:
@codex review

ishiland:
All three fixed.

**P2, `show the layer`.** Right, and the builder contract already uses that exact phrase for `toggle_visibility`. Treating "show" as a QUESTION categorically would have queried a hidden layer instead of making it visible. Visibility is now a named exception.

**P2, `show all features again`.** Also right. With a filter applied, that asks to CLEAR it, and answering it as a question returns a temporary result while leaving the persistent filter in place. Broadening is now a named exception pointing at `set_filter` with a null expression.

Both sit alongside the narrowing forms, and all three share one tell: the object is the map, not the data. That is what the prompt now says, rather than three unrelated escapes.

**P2, self-classification.** The sharpest of the three. `query_data`'s description did classify requests for itself — "use this when the user asks a question about their data ... do NOT use this for map styling or layer management" — and the name-based check passed it because it named no sibling. It is now behavioural, and a second guard rejects the prose shapes that classify without naming anything.

That guard flagged `set_style`'s "Do NOT use this for data-driven coloring" too. That pointer is a capability difference rather than verb classification, so it stays, reworded to say the same thing without reading as request-classification prose.

On the wider claim: you are right that no assertion over free text can prove a description says nothing about routing. The markers are heuristics covering the shape this incident actually took, and the test says so in its own comment rather than claiming more than it delivers.

@codex review

ishiland:
Both fixed, and they turned out to be the same mistake, so the rule is restructured rather than patched twice.

**Catalog searches.** Right — "find datasets about flood zones" is a QUESTION verb whose object is the catalog, and `query_data` can only reach layers already on the map. Routing it there turned a valid catalog lookup into a failed SQL question while `search_datasets` stayed the advertised catalog tool, so the rule contradicted itself.

**Layer-only narrowing.** Also right. "Show only @Parks" among several layers means hide the others, and `set_filter` cannot do that — it only filters features inside one layer. Pointing both narrowing cases at `set_filter` produced a no-op filter and left the conflicting instruction you describe.

The common cause is classifying on the verb alone. The prompt now reads the OBJECT first — catalog, layer, the data in a layer, the map's appearance — and only then the verb inside that family. The three "show" exceptions fall out of the object rule instead of sitting beside it as three unrelated escapes, and layer-level narrowing goes to `toggle_visibility` while feature-predicate narrowing goes to `set_filter`.

@codex review

ishiland:
Fixed. Right — `toggle_visibility`'s schema only shows or hides the whole layer, so "hide the labels" routed there would have taken the features with them. Labels are turned off with `set_label(column: null)`.

The mistake was grouping a layer and its labels as one object when they are two, with two tools. Split into two object rules.

@codex review

PR #1040, opened 2026-07-31
fix(catalog): serve the dataset extent bbox in the RFC 7946 spec form

### LINKED ISSUE #1004: fix(frontend): span-bbox flattening makes isLargeExtent fire before the #903 seam guard, so a seam-crossing dataset previews as globe-spanning
## What happens

Ingest a dataset whose features straddle the antimeridian, then open its detail page. The preview map zooms to the whole world and draws the extent as a dashed band spanning ±180, for data that occupies a 1.5° x 1.6° box near Fiji.

## Reproduce

Upload this, then open `/datasets/{id}`:

```json
{"type":"FeatureCollection","features":[
 {"type":"Feature","properties":{"name":"Seam West"},"geometry":{"type":"Point","coordinates":[179.5,-16.5]}},
 {"type":"Feature","properties":{"name":"Seam East"},"geometry":{"type":"Point","coordinates":[-179.5,-16.6]}},
 {"type":"Feature","properties":{"name":"Suva"},"geometry":{"type":"Point","coordinates":[178.44,-18.14]}}
]}
```

## Why this happens

Each piece is correct on its own.

Storage is right. `records.spatial_extent` holds the #901 two-ring split:

```
MULTIPOLYGON(((178.44 -18.14, 180 -18.14, 180 -16.5, 178.44 -16.5, 178.44 -18.14)),
             ((-180 -18.14, -179.5 -18.14, -179.5 -16.5, -180 -16.5, -180 -18.14)))
```

OGC is right. `/api/collections/{id}` emits the RFC 7946 §5.2 pair, west > east:

```json
{"bbox": [[178.44, -18.14, -179.5, -16.5]]}
```

`/api/datasets/` is right by its own contract. `extent_bbox` comes from `extent_to_span_bbox` (`backend/app/core/geo.py:100`), which returns plain shapely `.bounds` and so reports `[-180.0, -18.14, 180.0, -16.5]`. #892's docstring is explicit that this is deliberate: "over-broad, but never inverted", for consumers that cannot accept an inverted pair (span arithmetic, planar WKT serialization, tile-source bounds).

The gap is where they meet. `DatasetMap.tsx` tests `isLargeExtent(bbox)` before it can reach the #903 seam branch:

```ts
if (isLargeExtent(bbox!)) {
  const { center, zoom } = computeLargeExtentView(bbox!);
  map.flyTo({ center, zoom, duration: motionDuration(1000) });
} else {
  // fix(#903): an antimeridian-crossing extent is expressed by letting east
  // run past 180 ...
  map.fitBounds(toFitBounds(bbox!), { padding: 60, duration: motionDuration(1000) });
}
```

`isLargeExtent` (`frontend/src/lib/map-extent.ts:30`) fires at a longitude span above 90°. The flattened bbox spans 360°, so the seam branch is unreachable on this path. #903's guard is dead code for the datasets it was written for.

## Three affected sites, not one

The reproduction above exercises the initial camera, not the button the quoted snippet belongs to. All three consumers in `DatasetMap.tsx` are defeated by the same flattened bbox and all three need the fix:

- `initialViewState`, around `DatasetMap.tsx:894` — the second `isLargeExtent` gate, and the one the reproduction hits.
- `handleZoomToExtent`, `DatasetMap.tsx:507` — the quoted snippet, reachable only via the Zoom to Extent button.
- `bboxGeojson`, `DatasetMap.tsx:581-600` — the dashed extent band, drawn via `splitBbox`. `splitBbox` returns a one-element array for anything that does not cross, and the flattened bbox does not cross, so the band paints as one global rectangle.

## Why this cannot be fixed on the frontend

An earlier reading of this suggested teaching `isLargeExtent` to detect the seam case before it measures span. That is not implementable. `extent_to_span_bbox` flattens the Fiji extent to `[-180.0, -18.14, 180.0, -16.5]`, which is bit-identical to what a genuinely global dataset at the same latitude band would produce. `crossesAntimeridian` (`frontend/src/lib/bbox.ts:25-27`) can only key on `west > east`, and the flattened pair is monotonic. The seam information is destroyed server-side before the payload is serialized, so no client-side test can recover it. The producer has to stop destroying it.

## The fix: flip the producer

Change `backend/app/modules/catalog/datasets/domain/helpers.py:184` from `extent_to_span_bbox(record.spatial_extent)` to `extent_to_bbox(record.spatial_extent)`, so `/api/datasets/` serves the RFC 7946 spec form, and update the `fix(#892 codex P2)` comment above it that explains the old choice.

This is the sequencing the frontend helpers were written for. The header of `frontend/src/lib/bbox.ts:11-15` says so:

> They do not change what the API sends today — they are what has to exist before the backend can flip a consumer back to the spec form (see the issue's sequencing: #934, then this, then one consumer at a time).

#934 is closed (via #980), the helpers exist, and this is the first consumer.

Supporting evidence that the spec form is safe here: `search/router.py:793` and `standards/ogc/router.py:316` already emit the spec form, which is why `BBoxPreview` renders a seam-crossing dataset correctly on the search results page but not on the dataset detail page.

## What NOT to flip

Do not change `extent_to_span_bbox` itself, and specifically **do not flip `backend/app/modules/catalog/maps/_router_helpers.py:117`** (`dataset_extent_bbox`). That field feeds `frontend/src/components/builder/map-sync.ts:243`, which passes it to the vector tile source's `bounds`; a west > east pair bounds the source to no tiles and the layer renders blank. `backend/tests/test_antimeridian_extent.py:391-422` pins the monotonic form there with the reasoning written out. One database column, two consumers with opposite requirements.

Span consumers that stay on the span form: `processing/tiles/router.py:1213` (raster tile-source bounds) and the three DCAT serializers.

## Tests

`backend/tests/test_stac_import.py:518-523` currently pins `extent_bbox == [-180.0, -20.0, 180.0, -15.0]` with a comment citing the DatasetMap unguarded-ring reasoning. That reasoning no longer holds now that #903 added the guards, so the assertion has to be updated to the spec form as part of the flip, not worked around.

Add a spec-form pin in `backend/tests/test_antimeridian_extent.py` beside the existing map-layer pin, so the two opposite contracts sit next to each other and a future refactor that unifies them fails loudly. Cover the three `DatasetMap` sites in `DatasetMap.test.tsx` against the Fiji fixture.

## Related

Sibling bug on the collections path: `collections/service.py:321` uses `rollup_span_bbox`, so `CollectionCard` and `CollectionDetailPage` have the same flattened-seam behaviour. Tracked separately.

Found during a post-rebuild smoke check of `9a788ba92`.

ishiland:
Closes #1004.

## The gap

Every piece was correct on its own. Storage holds the #901 two-ring split; OGC
emits the spec pair; `/api/datasets/` emitted the span form, which its own
`#892` docstring justifies as "over-broad, but never inverted".

The gap is where they meet. `extent_to_span_bbox` flattens a Fiji extent to
`[-180, -18.14, 180, -16.5]`, which is **bit-identical** to what a genuinely
global dataset at the same latitude band produces. `DatasetMap` tests
`isLargeExtent(bbox)` before it can reach its #903 seam branch, and the flattened
pair spans 360°, so the seam branch was unreachable and a 1.5° x 1.6° box near
Fiji previewed as a dashed band across the whole world.

This cannot be fixed client-side: `crossesAntimeridian` can only key on
`west > east`, and the flattened pair is monotonic. The seam information is
destroyed before serialization, so the producer has to stop destroying it.

Three sites were affected, not just the one the issue's snippet quotes:
`initialViewState` (what the reproduction hits), `handleZoomToExtent`, and
`bboxGeojson` (drawn via `splitBbox`, which returns one ring for anything that
does not cross).

## What changed

`dataset_to_response` uses `extent_to_bbox`. `frontend/src/lib/bbox.ts`'s header
names this as the sequencing it was written for: #934 first (closed via #980),
then one consumer at a time. This is the first consumer.

Also regenerates `backend/openapi.json`, `frontend/src/types/api.generated.ts`
and both SDKs, for the field description.

## What is deliberately NOT flipped

`dataset_extent_bbox` in `maps/_router_helpers.py` stays on the span form. It
feeds `map-sync.ts:243` → the vector tile source's `bounds`, where a `west > east`
pair bounds the source to no tiles and the layer renders blank. One database
column, two consumers with opposite requirements.

`extent_to_span_bbox` itself is unchanged, as are the raster tile-source bounds
and the three DCAT serializers.

## Consumer audit before the flip

`DatasetResponse.extent_bbox` reaches `DatasetMap` (fixed by this change) and
`SpatialExtentCard`, which prints the four numbers and does no span arithmetic.
`e2e/export-runtime.spec.ts` reads it through a `toBBox` helper that already
returns `null` on `minX > maxX` and falls back to computing the bbox from
feature positions. `dataset-validation-navigation.ts` only matches the field
name. Nothing else consumes it.

## Verification

```
  tests/test_antimeridian_extent.py tests/test_antimeridian_rollup.py \
  tests/test_collections.py tests/test_fe_sdk_type_drift.py \
  tests/test_maps_style_json.py tests/test_processing_port.py \
  tests/test_stac_import.py tests/test_vrt_catalog_175.py tests/test_datasets.py -q
  # 250 passed
make openapi-check
cd frontend && npx vitest run src/components/dataset src/lib/__tests__  # 924 passed
```

Backend: `test_dataset_response_bbox_uses_the_spec_form` is added directly
beside the existing `test_map_layer_response_bbox_stays_monotonic`, so the two
opposite contracts on one column sit together and a refactor that unifies them
fails loudly. `test_stac_import.py`'s pin is updated to the spec form rather
than worked around — its comment cited the unguarded-ring reasoning that #903
retired.

Frontend: `DatasetMap.test.tsx` gains a block covering all three sites against
the Fiji fixture, each paired with a genuinely global bbox that must still take
the large-extent branch — same code, two bboxes, opposite outcomes.

### DIFF
--- a/backend/app/modules/catalog/datasets/domain/helpers.py
+++ b/backend/app/modules/catalog/datasets/domain/helpers.py
@@ -20,7 +20,7 @@ from app.modules.catalog.sources.provenance import (
     derive_last_edited,
     resolve_actor,
 )
-from app.core.geo import extent_to_span_bbox, wkt_is_geographic
+from app.core.geo import extent_to_bbox, wkt_is_geographic
 
 
 async def _load_actor_identities(
@@ -174,14 +174,18 @@ def dataset_to_response(
         n_dims=dataset.n_dims,
         z_min=dataset.z_min,
         z_max=dataset.z_max,
-        # fix(#892 codex P2): the SPAN, not the RFC 7946 spec bbox. This field is
-        # documented as "[minx, miny, maxx, maxy]" and its consumers are map
-        # viewers, not a standards serializer: DatasetMap draws an unguarded
-        # planar ring from it and calls fitBounds, so a west > east pair would
-        # paint the complementary 340° and hand MapLibre an inverted camera.
-        # The spec form is served where the spec demands it (STAC/OGC record
-        # bbox via extract_bbox, OGC collection extents), not here.
-        extent_bbox=extent_to_span_bbox(record.spatial_extent),
+        # fix(#1004): the RFC 7946 §5.2 spec bbox, west > east on a crossing.
+        # This was the span form under #892, when DatasetMap drew an unguarded
+        # planar ring and called fitBounds. #903 added the seam guards, and the
+        # span form then defeated them: extent_to_span_bbox flattens a Fiji
+        # extent to [-180, s, 180, n], which is bit-identical to a genuinely
+        # global dataset, so isLargeExtent fired first and no client-side test
+        # could tell the two apart. The seam information has to survive the
+        # wire. Do NOT flip the sibling `dataset_extent_bbox` in
+        # maps/_router_helpers.py: it bounds a vector tile source, where an
+        # inverted pair yields no tiles at all. One column, two contracts —
+        # both pinned in tests/test_antimeridian_extent.py.
+        extent_bbox=extent_to_bbox(record.spatial_extent),
         column_info=dataset.column_info,
         quality_detail=dataset.quality_detail,
         license=record.license,
--- a/backend/app/modules/catalog/datasets/domain/schemas.py
+++ b/backend/app/modules/catalog/datasets/domain/schemas.py
@@ -251,7 +251,11 @@ class DatasetResponse(BaseModel):
     )
     feature_count: int | None
     extent_bbox: list[float] | None = Field(
-        default=None, description="Bounding box [minx, miny, maxx, maxy]"
+        default=None,
+        description=(
+            "Bounding box [west, south, east, north] per RFC 7946 §5.2. "
+            "west > east on an antimeridian-crossing extent."
+        ),
     )
     column_info: list[ColumnInfo] | None = Field(
         default=None, description="Column names, types, and stats"
--- a/backend/openapi.json
+++ b/backend/openapi.json
@@ -5081,7 +5081,7 @@
                 "type": "null"
               }
             ],
-            "description": "Bounding box [minx, miny, maxx, maxy]",
+            "description": "Bounding box [west, south, east, north] per RFC 7946 \u00a75.2. west > east on an antimeridian-crossing extent.",
             "title": "Extent Bbox"
           },
           "feature_count": {
--- a/backend/tests/test_antimeridian_extent.py
+++ b/backend/tests/test_antimeridian_extent.py
@@ -12,6 +12,7 @@ take an inverted pair.
 from __future__ import annotations
 
 import uuid
+from datetime import UTC, datetime
 
 import pytest
 from geoalchemy2 import WKBElement, WKTElement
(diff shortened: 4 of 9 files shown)


ishiland:
@codex review

PR #1041, opened 2026-07-31
fix(builder): four polish items from the 2026-07-27 audit

ishiland:
Closes items 1, 5, 6 and 9 of #787. Items 3 and 10 are out of scope here (both `AnalysisPanel`) and stay open on the issue.

## Item 1 — the result chip drew under the Legend panel

`EphemeralBadge` sat at `z-[8]`, under every `z-10` `PluginHost` slot. The bottom-left slot is offset to clear the badge, but a panel taller than that offset overlaps it and the z-index decides. `z-20`, matching the other map overlays (`MapTitlePill`, the BuilderMap warning box).

## Item 5 — dataset delete fired three 404 refetches

`useDeleteDataset` invalidated the deleted dataset's detail query, and `related/` + `maps/` sit under the `['datasets']` prefix its list invalidation matches, so all three refetched a resource that had just been deleted. Now: cancel those three, and keep the list invalidation off the deleted id via a predicate.

Cancel rather than remove, which the live check settled: the caller navigates away only after `mutateAsync` resolves, so the detail observer is still mounted when this runs, and removing an *active* query makes the observer rebuild it and fetch. Removing left one 404 where cancelling leaves none. Measured on a throwaway ingested dataset, deleted through the real dialog:

```
cancel + remove + predicate  ->  1 404  ["/api/datasets/{id}"]
cancel + predicate           ->  0 404s, 0 console errors     <- shipped
main                         ->  3 404s (detail, related/, maps/)
```

The stale cache entry ages out through `gcTime` instead.

## Item 6 — auto-fit jumped the viewport with no way back

The auto-fit effect captures the camera (center, zoom, bearing, pitch) before `fitBounds` and offers a one-shot "Return to previous view" through the existing toast-with-action pattern.

- Not on the first fit: there is no view the user chose yet, so the offer would be page-load noise.
- A second auto-fit dismisses the older offer rather than stacking it — its camera is no longer the one the user was looking at.
- The toast is dismissed once used.
- It runs longer than the default 4s. An auto-fit usually rides along with another toast ("Layer removed"), sonner collapses the stack, and the offer was expiring while it was still hidden behind the one in front. Found in QA, not in review.

Live: removing a layer whose extent drove the fit moved the camera to `42.80° N · 75.82° W · z 6.6`; the offer survived the sibling toast; clicking it returned the camera to `20.00° N · 0.00° E · z 2.0` and dismissed itself. No toast on page load.

## Item 9 — the history panel told people to close and reopen it

Refetching is what closing and reopening did, so the error state now carries the same "Try again" button `DatasetSearchPanel` uses, and the description says what to check instead of prescribing a workaround.

## Verification

```
# 366 files, 4288 tests pass
```

New tests: the badge's stacking context; the delete hook's cancel/predicate behaviour, including that it never removes an active query and that the predicate excludes only the deleted id; the history panel's retry button and that it calls `refetch`. Eight new locale entries across en/es/fr/de (two `map.*`, two `history.*`), plus a reworded `history.errorDescription`.

Item 6 has no unit test — the auto-fit path needs a live maplibre camera, and the QA run above is the check. Screenshots on request; the camera readouts above are from that run.

### DIFF
--- a/frontend/src/components/builder/BuilderMap.tsx
+++ b/frontend/src/components/builder/BuilderMap.tsx
@@ -1311,6 +1311,9 @@ export const BuilderMap = memo(function BuilderMap({
   if (prevVisibleBoundsKeyRef.current === undefined) {
     prevVisibleBoundsKeyRef.current = visibleLayerBoundsKey(getVisibleLayerBounds(layers));
   }
+  // fix(#787 item 6): id of the live "return to previous view" toast, so a second
+  // auto-fit replaces the offer instead of stacking a stale one.
+  const returnToViewToastRef = useRef<string | number | undefined>(undefined);
 
   // Auto-fit to visible layers (skip on initial load if saved view exists)
   useEffect(() => {
@@ -1325,7 +1328,8 @@ export const BuilderMap = memo(function BuilderMap({
     prevVisibleBoundsKeyRef.current = visibleBoundsKey;
 
     // First auto-fit trigger: skip if we have a saved view to restore
-    if (!initialFitDoneRef.current) {
+    const isFirstFit = !initialFitDoneRef.current;
+    if (isFirstFit) {
       initialFitDoneRef.current = true;
       if (hasSavedView) return; // MapGL already positioned via initialViewState
     }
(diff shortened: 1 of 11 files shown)


ishiland:
@codex review

ishiland:
Both right, and the first one is a fair hit on a trade-off I waved through in the PR body.

**Deleted dataset cached under Back.** Cancelling left a *successful* detail entry with a 60s `staleTime`, so Back rendered the deleted dataset with no request. `refetchType: 'none'` gets both properties: nothing refetches while the observer is still mounted (no 404 at delete time), and the entry is stale, so a revisit issues a real request. Verified live on a throwaway ingested dataset:

```
[delete]  404s at delete time: 0
[Back]    re-requested the dataset: true  (detail, related/, maps/ -> 404)
[Back]    still renders the deleted title: false
[Back]    shows a not-found surface: true
```

**Toast outliving the builder.** Dismissed in the existing unmount cleanup (which is `[]`-deps, so unmount-only). Verified by taking the same auto-fit path and navigating away instead of clicking the action:

```
[unmount] offer alive on the builder: 1
[unmount] offer on the destination page: 0
console errors: 0
```

Gates re-run at `b6236bdc5`: lint, typecheck, i18n, 4288 frontend tests.

ishiland:
@codex review

PR #1043, opened 2026-07-31
feat(analysis): sync the analysis job across tabs and dedup its completion

### LINKED ISSUE #1008: feat(analysis): cross-tab sync and completion dedup for the analysis job store
Split out of #790 (item 3), which was carrying four unrelated pieces of work. Verified 2026-07-31 against `main` at `9a788ba92`.

Two open tabs on the builder each poll the same materialize job, each raise a completion toast, and each invalidate the query cache. A second tab also does not know the first tab has a job in flight, so its Create button stays enabled and the user earns a 429 rendered as generic rate-limiting.

**This is the second time it has been filed.** It was item 6 of #698 ("Multi-tab duplicate completion toasts: the persisted job store has no cross-tab listener and the watcher mounts per tab"), which closed without it shipping. Recording that here so the history is not lost a third time.

## What is actually there today

`frontend/src/stores/analysis-job-store.ts` persists a single `TrackedAnalysisJob` (`jobId`, `title`, `mapId`) under the key `geolens-analysis-job` (`:29-33`). Its only subscription is `useAuthStore.subscribe` at `:42-46`, which clears the job when the user identity changes. There is no `storage` listener and no `BroadcastChannel`.

`frontend/src/components/analysis/AnalysisJobWatcher.tsx` has no cross-tab dedup either. Its `toastId` at `:64` is a per-job id, which collapses StrictMode's double invoke **within one tab** and does nothing across tabs, since each tab has its own toaster.

## Two separate mechanisms, and conflating them is the trap

**Mechanism 1, the mirror.** A `storage` event listener that rehydrates the persisted store when another tab writes it, so tab B learns tab A started a job and disables Create. `frontend/src/stores/__tests__/auth-store.cross-tab.test.ts` is the working pattern for this half: same `persist` middleware, same `storage`-event shape, same test setup.

**Mechanism 2, the claim.** The mirror does not dedup anything. If three tabs mirror the same job and all three poll it to completion, all three fire a toast and all three invalidate. Deduping needs a written claim that exactly one tab wins.

Use last-writer-wins on a `completedAt` field added to the persisted store: on observing a terminal status, a tab writes `completedAt` for that `jobId` into the persisted record and then re-reads it; if the value it reads back is not its own, another tab already claimed the completion and this tab stays silent. `localStorage` writes serialize per origin, so the read-back is the arbiter. The auth-store test is **not** a precedent for this part — it exercises a mirror, not a claim — so the dedup wants its own test that drives two store instances against one shared storage mock.

## Acceptance criteria

- [ ] A second tab reflects the first tab's in-flight job without a reload, and its Create button is disabled.
- [ ] Exactly one completion toast and exactly one cache invalidation across N open tabs, driven by the claim rather than by timing.
- [ ] The claim is durable: a tab that reloads after another tab claimed the completion does not re-toast.
- [ ] The identity-change clear at `analysis-job-store.ts:42-46` still works, and clearing in one tab clears the others.
- [ ] Covered by tests in the style of `auth-store.cross-tab.test.ts` for the mirror, plus a distinct test for the claim.
- [ ] Frontend `npm run lint` / `npm run typecheck` clean.

## Overlapping work

#699 also creates `frontend/src/stores/__tests__/analysis-job-store.test.ts` (its frontend bullet: persistence, the identity-change clear, rehydration after reload). Whichever of the two lands second extends the existing file rather than creating it, and neither should assume the file is absent.

ishiland:
Two builders open on the same map each poll the same materialize job, each raise a completion toast, and each invalidate the catalog cache. The second tab also does not know the first has a run in flight, so its Create button stays enabled and the user earns a 429 rendered as generic rate limiting.

This is the second time the item has been filed — it was item 6 of #698 and closed without shipping.

## Mechanism 1, the mirror

A `storage` listener on `geolens-analysis-job` that rehydrates when another tab writes, the same shape `auth-store.ts:113-129` already ships. `analysisJobRunning` in the Analysis panel (`AnalysisPanel.tsx:505`) is a live store subscription, so tab B's Create button disables without a reload, and a clear propagates the same way. **No change to `AnalysisPanel.tsx` was needed** — the subscription was already there.

## Mechanism 2, the claim

The mirror dedups nothing: three tabs mirroring one job all still poll it to completion, and each has its own toaster (`toastId` is per job, which only collapses StrictMode's double invoke *within* a tab).

`completedAt` is a written claim on the persisted record. A tab checks for an existing claim on that `jobId`, writes its own if there is none, then reads back; the read-back is the arbiter, since localStorage writes serialize per origin.

Both halves are load-bearing. The check alone loses a genuine race (two tabs both see no claim). The read-back alone lets a tab that writes and reads before the other tab writes come out true as well, so both would report. Together, exactly one.

The record carries the writing tab's id, not a bare timestamp: two tabs can observe the same terminal poll in the same millisecond, and a timestamp then cannot say whose claim survived.

Being written rather than timed is also what makes it durable — a tab that reloads after another already reported reads the existing claim and stays silent.

## Schema note

The persisted `version` is deliberately **not** bumped for the new field. Zustand's default shallow merge leaves `completedAt` at its initial `null` when an older payload omits it, and a bump would discard a job that was in flight across the deploy.

## Tests

`src/stores/__tests__/analysis-job-store.cross-tab.test.ts` is new — the mirror half in the style of `auth-store.cross-tab.test.ts`, and the claim half as its own thing, because a mirror test is not a dedup test. The two-tab case drives a genuinely second store instance (`vi.resetModules()` + re-import, so it gets its own tab id) against the shared `localStorage`, which is the only way to get a distinct claimant out of a module singleton. The same trick doubles as the reload-durability case.

Deliberately a sibling file rather than an extension of `analysis-job-store.test.ts`: #699 (PR #1037) creates that file, this creates no file it owns, and the split matches the `auth-store.test.ts` / `auth-store.cross-tab.test.ts` precedent the issue points at. The two PRs do not touch a shared file.

`AnalysisJobWatcher.test.tsx` gains the claim behaviour end to end: silent on a completion and on a failure another tab claimed (with the local cleanup still running so Create re-enables), and exactly one toast plus exactly two invalidations when it wins. Verified by mutation — disabling the claim gate fails both silence tests.

## Verification

```
cd frontend
npx vitest run src/stores/__tests__/analysis-job-store.cross-tab.test.ts \
  src/components/analysis/__tests__/AnalysisJobWatcher.test.tsx
npm run lint && npm run typecheck && npm run test:coverage   # 367 files, 4297 tests
```

No new `t()` keys, so locale parity is unaffected. No backend, schema or API impact.

Closes #1008

### DIFF
--- a/frontend/src/components/analysis/AnalysisJobWatcher.tsx
+++ b/frontend/src/components/analysis/AnalysisJobWatcher.tsx
@@ -1,4 +1,4 @@
-import { useEffect } from 'react';
+import { useEffect, useRef } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useNavigate } from 'react-router';
 import { useQueryClient } from '@tanstack/react-query';
@@ -7,7 +7,12 @@ import { queryKeys } from '@/lib/query-keys';
 import { ApiError } from '@/api/client';
 import { useJobStatus } from '@/components/import/hooks/use-ingest';
 import { useAnalysisFormStore } from '@/stores/analysis-form-store';
-import { analysisAddToMap, useAnalysisAddedStore, useAnalysisJobStore } from '@/stores/analysis-job-store';
+import {
+  analysisAddToMap,
+  useAnalysisAddedStore,
+  useAnalysisJobStore,
+  type TrackedAnalysisJob,
+} from '@/stores/analysis-job-store';
 
 /**
  * Global notifier for a materialize-analysis job (renders nothing).
@@ -20,8 +25,14 @@ import { analysisAddToMap, useAnalysisAddedStore, useAnalysisJobStore } from '@/
 export function AnalysisJobWatcher() {
   const { t } = useTranslation('builder');
   const job = useAnalysisJobStore((s) => s.job);
+  const completedAt = useAnalysisJobStore((s) => s.completedAt);
   const claimCompletion = useAnalysisJobStore((s) => s.claimCompletion);
   const clearJobIfCurrent = useAnalysisJobStore((s) => s.clearJobIfCurrent);
+  // Runs whose terminal status this tab has already acted on, so neither the
+  // effect below nor the propagated-clear cleanup repeats itself.
+  const handledRef = useRef<Set<string>>(new Set());
+  // The job as of the previous render, so a disappearance is detectable.
+  const lastSeenRef = useRef<TrackedAnalysisJob | null>(null);
   const queryClient = useQueryClient();
   const navigate = useNavigate();
   // Shares its query key with the Analysis panel, so both watching costs one poll.
@@ -60,6 +71,15 @@ export function AnalysisJobWatcher() {
     // staleness rule of its own; it mirrors the server's verdict.
     if (status !== 'complete' && status !== 'failed') return;
 
+    // Once per job per tab. The clear used to be synchronous, so a re-run
+    // simply fell out at `if (!job)`; now the claim is awaited, and any render
+    // in between re-enters — including the one the claim's own store write
+    // causes, and the one a fresh `t` identity causes on every render. This
+    // also subsumes StrictMode's double invoke, which is why the effect needs
+    // no cleanup-based guard.
+    if (handledRef.current.has(job.jobId)) return;
+    handledRef.current.add(job.jobId);
+
     // fix(#1008 codex P2): the cache invalidation is deliberately NOT behind
     // the claim below. Every document owns its own QueryClient and the app
     // disables refetch-on-window-focus, so gating this would leave every tab
@@ -75,12 +95,7 @@ export function AnalysisJobWatcher() {
     // arbiter rather than a timing accident, so it also holds for a tab that
     // reloads after another already reported.
     const tracked = job;
-    let superseded = false;
-    void claimCompletion(tracked.jobId).then((claimed) => {
-      // The effect re-ran while the claim was in flight (StrictMode, or a new
-      // poll response). That run owns the outcome; acting here would double
-      // whatever it already did.
-      if (superseded) return;
+    void claimCompletion(tracked.jobId, status).then((claimed) => {
       if (claimed) {
         // Stable id so a re-run (StrictMode's double invoke) replaces the toast
         // instead of stacking a second one. That collapses duplicates WITHIN a
@@ -202,9 +217,6 @@ export function AnalysisJobWatcher() {
       }
       void clearJobIfCurrent(tracked.jobId);
     });
-    return () => {
-      superseded = true;
-    };
   }, [
     job,
     status,
@@ -216,6 +228,28 @@ export function AnalysisJobWatcher() {
     clearJobIfCurrent,
     t,
   ]);
+
+  // fix(#1008 codex P2): the clear propagates, so a tab that was throttled
+  // through the terminal poll simply watches its tracked job vanish. Its
+  // effect above then returns at `if (!job)` and it never runs the cleanup
+  // that is document-local: its own QueryClient (focus refetching is off, so
+  // nothing else refreshes it) and its own remembered form title. The claim
+  // carries the status precisely so a tab that did not observe the run's end
+  // still knows what kind of end it was.
+  useEffect(() => {
+    const previous = lastSeenRef.current;
+    lastSeenRef.current = job;
+    if (job || !previous) return;
+    // This tab reported the run itself, so it already did all of this.
+    if (handledRef.current.has(previous.jobId)) return;
+    // A clear with no claim behind it is a logout or an identity change, not
+    // a completion — nothing was created and nothing needs refreshing.
+    if (completedAt?.jobId !== previous.jobId) return;
+    if (completedAt.status !== 'complete') return;
+    queryClient.invalidateQueries({ queryKey: queryKeys.datasets.all });
+    queryClient.invalidateQueries({ queryKey: queryKeys.search.all });
+    useAnalysisFormStore.getState().clearTitleForMap(previous.mapId);
+  }, [job, completedAt, queryClient]);
 
   return null;
 }
--- a/frontend/src/components/analysis/AnalysisJobWatcher.tsx
+++ b/frontend/src/components/analysis/AnalysisJobWatcher.tsx
@@ -44,7 +44,7 @@ export function AnalysisJobWatcher() {
       // (#793 review): restoring it would re-enable Create with the finished
       // run's name.
       useAnalysisFormStore.getState().clearTitleForMap(job.mapId);
-      clearJobIfCurrent(job.jobId);
+      void clearJobIfCurrent(job.jobId);
       return;
     }
     // Only a terminal status stops tracking — deliberately no client-side
@@ -200,7 +200,7 @@ export function AnalysisJobWatcher() {
       if (status === 'complete') {
         useAnalysisFormStore.getState().clearTitleForMap(tracked.mapId);
       }
-      clearJobIfCurrent(tracked.jobId);
+      void clearJobIfCurrent(tracked.jobId);
     });
     return () => {
       superseded = true;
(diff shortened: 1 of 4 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
Thanks — all three land as changes.

**P1, the clear erasing a newer job.** Correct, and it is a regression the mirror introduced: the unconditional `setJob(null)` used to be this tab's own business. `clearJobIfCurrent` now reads the persisted job and stands down when a different run owns the slot, on both the terminal and the swept-job paths.

**P2, claim atomicity.** Also correct — the interleaving you name defeats it, and the code comment claiming otherwise was wrong. localStorage serializes individual operations, not the sequence. The check/write/read now runs inside a Web Lock (`navigator.locks.request`), which does give mutual exclusion across same-origin documents, with the old best-effort path kept as the fallback where Web Locks is unavailable. `claimCompletion` is async as a result and the watcher effect awaits it behind a supersede guard.

On testing it: the race cannot be staged in a single JS realm, since `attempt()` runs to completion before the other caller starts. The new test pins the mechanism instead — the claim must go through the lock — and says why the race itself is not reproducible here.

**P2, invalidation in every tab.** Agreed, and this one is a deliberate departure from the issue's acceptance criterion ("exactly one cache invalidation across N open tabs"), which is wrong on the merits for the reason you give: per-document QueryClient plus focus-refetch disabled. Invalidation moved out of the claim, so every observing tab refreshes; only the reporting is claimed. The watcher test that asserted otherwise now asserts the losing tab still invalidates.

Pushed as 64b9cb8.

ishiland:
Correction to the sha in my previous comment: the branch was rebased onto main as part of that push, so those changes are at `9fcba9a`, not `64b9cb8`.

ishiland:
@codex review

ishiland:
Right again, and the fix uncovered a second-order problem worth recording.

Every write to this store persists the whole snapshot, so `set({ completedAt })` republished this tab's `job` as a side effect. Each write now refreshes what it is *not* changing from storage first, falling back to memory only when storage has nothing to offer, so a first write or an unreadable read cannot null a field out. The throttled-tab sequence now ends with storage still naming `job2`, and the guarded clear stands down as intended. Pinned by a test that stages exactly that ordering, plus one that a `setJob` no longer drops a claim another tab wrote.

The second-order problem: `readPersisted` parses fresh objects out of JSON on every call, so merging them in unconditionally handed React a new `job` identity on every write and re-ran every effect subscribed to the store. That surfaced immediately as the watcher invalidating its caches twice for one completion. The merge now keeps the in-memory object whenever it describes the same run.

Pushed as 32122e1.

ishiland:
@codex review

ishiland:
All three taken, and the shape changed as a result: every mutation on this record now runs under one Web Lock, not just the claim.

**P1, the guarded clear.** Correct — it read the persisted job and then wrote through a *second* read, so a run started between the two got overwritten with null, which is the case the guard exists for. `setJob`, `claimCompletion` and `clearJobIfCurrent` all take the same lock now, and each reads once and writes from that same snapshot, so a decision and the write it justifies cannot straddle two reads. Both actions became async; the watcher and the auth subscriber were updated.

**P2, the stale claim.** Also correct. A claim naming a run that storage no longer tracks, while the slot holds a different run's claim, is now refused. The successive-runs case still works because storage tracks the new run when it completes, so the two are distinguishable without a timestamp comparison; there is a test for each.

**P2, a failed persist.** `withLock` catches, so none of the three ever reject. A failed claim degrades to `true` rather than `false` deliberately — a duplicate notification beats never telling the user their dataset is ready, and it matches what this module does elsewhere with an unreadable payload. Covered by tests that make `setItem` throw.

Pushed as bf6a6e9.

ishiland:
@codex review

ishiland:
@codex review

ishiland:
Taken — and it exposed a second bug on the way.

**The finding itself.** The claim now carries how the run ended (`status: 'complete' | 'failed'`), which is what lets a tab that never saw the end know which cleanup it owes. When the tracked job disappears with a completion claim naming it, the receiving tab refreshes its own caches and drops its remembered title. A clear with no claim behind it is a logout or an identity change — nothing was created, so nothing is refreshed. Both cases have a test.

**What it exposed.** Handling a terminal status was not once per job. The clear used to be synchronous, so a re-entry fell out at `if (!job)`; with the claim awaited, every render in between re-enters — including the one the claim's own store write causes, and the one a fresh `t` identity causes on literally every render. That was invalidating twice per completion, which the watcher test caught once I put a count on it. Now guarded per job per tab, which also subsumes StrictMode's double invoke, so the cleanup-based supersede guard is gone.

Pushed as 094f431 (rebased onto main in the same push).

ishiland:
@codex review

ishiland:
Correct — "went away" is not "became null", and the cleanup was only checking the latter.

The departed job now waits in a ref until a claim names it, so both the direct job1 → job2 replacement and the case where the claim lands on a later render than the job change are covered, with a test for each. One pending job at a time is enough: the store tracks a single run, so an older pending job has had its claim slot overwritten and could never be matched again.

Pushed as 25901ec.

ishiland:
@codex review

ishiland:
Both taken.

**The suspended tab.** Right — matching the claim's id against the departed job refreshed for neither run. Refreshing was never job-specific in the first place: one invalidation re-reads the whole catalog, so it now fires on whatever claim the tab can see, however many runs it slept through. Only the remembered form title stays keyed to the run this tab was showing, because only that run's completion retires it. Test added for a claim naming a run this tab never tracked.

**The identity clear.** Also right, and it is a regression the mirror introduced — the clear used to be local. The tracked job now carries the identity that started it, stamped inside `setJob` rather than by callers so none can forget, and the auth handler only clears a run belonging to the account it is cleaning up after. A job with no stamp predates the field, so its owner cannot be established and clearing stays the safe reading; that case has its own test.

Pushed as 4b2b873 (rebased onto main in the same push).

For what it is worth, this is round six, and each one has found a genuine interaction rather than a nitpick. If the next pass turns up another, I would rather flag the shape of the problem than keep patching: a claim protocol over localStorage plus a Web Lock is close to the limit of what this substrate supports, and a `BroadcastChannel` with a single elected leader would be the honest rewrite. Not proposing it in this PR — the current version is a strict improvement on today's three-toasts-and-a-429 behaviour and is well covered — but noting it so the next reader knows the ceiling was in view.

ishiland:
@codex review

ishiland:
Both taken, by splitting the two cleanups rather than patching their shared condition again — three passes kept getting this wrong precisely by treating them as one thing.

**Refreshing** is keyed on the claim and is not job-specific. One invalidation re-reads the whole catalog, so a tab needs exactly one for whichever claim it can see.

**Retiring the remembered title** is keyed on the departure of the run this tab was showing, and no longer waits for a claim to explain it. That covers both findings at once: the claim and the clear are separate persisted writes arriving in either order, and a swept job departs with no claim at all, so waiting kept a finished run's name indefinitely. An unexplained departure retires it — the reading the swept-job branch already takes in whichever tab observes the 404. A claim saying the run FAILED is the one case that keeps the name for the retry, and that decision is now recorded so the departure handler does not second-guess it. Three tests, one per ordering.

Pushed as a8b20e2.

ishiland:
@codex review

ishiland:
Both taken. `setJob` reads the identity at call time now, not inside the lock callback, so an account switch during the wait cannot stamp a run as belonging to whoever arrived. And the catch-up refresh no longer looks at the claim's status — a tab suspended while one run succeeded and a later one failed sees only the failed claim, and gating on it left that tab a catalog missing the first run's dataset with nothing to correct it. One request after a failure is the cheaper side of that trade.

Pushed as fa16a51.

Calling the loop here, and flagging the shape rather than starting a ninth round. Eight rounds have each found a real interaction, and the last four have all been the same shape: two documents coordinating through a shared key, where every ordering of two independent writes is reachable and each fix reveals the next ordering. That is a design ceiling, not a run of oversights. `localStorage` plus a Web Lock gives mutual exclusion over a write, and nothing about the ORDER two tabs observe events in.

What this PR is: a strict improvement over today's behaviour, which is N toasts, N invalidations, and a second tab whose Create button earns a 429. Every acceptance criterion on #1008 is met and covered, and the residual gaps are all of the form "a tab suspended across several runs briefly shows a stale name".

What the next reader should know: making it airtight wants a different substrate — a `BroadcastChannel` with a single elected leader doing the polling, so there is one observer and no ordering to reconcile. That is a rewrite of `AnalysisJobWatcher`, not an increment on this, and it belongs in its own issue. Happy to file it if the maintainers want it tracked.

ishiland:
@codex review

Issue #1042, opened 2026-07-31
refactor(ingest): decompose processing/ingest/metadata.py (2002 lines, four cap carve-outs deep)


ishiland:
Split out of #958, which settled the LOC-ratchet inclusion rule and deliberately left this one alone: it is a decomposition, not a test change.

## What

`backend/app/processing/ingest/metadata.py` is 2002 lines, the second-largest module in `backend/app/` after `processing/tiles/router.py`. It is ratcheted exact in `_MODULE_LOC_CAPS`, which is working as designed — but the cap comment on it is now four carve-outs deep (#888, #899 codex r1, #906, #934), each one a genuine fix that had to argue for its lines.

That is the friction pointed at the wrong thing. The ratchet is meant to make growth visible, not to tax correctness work on a module nobody has had time to split. Every antimeridian and CRS fix lands here, and the batch that produced those four carve-outs suggests more are coming.

## Why it is not urgent

Nothing is broken. The gate holds, the carve-outs are honest, and the module's history is legible precisely because each raise was reviewed. This is a "pay it down before the next batch" item, not a bug.

## Scope sketch, not a plan

The natural seams visible from the cap comments and the call sites:

- the Mercator-clip / longitude-convention block (`clip_to_mercator_bounds`, `_shift_zero_to_360_longitudes`, `_mercator_envelope_degenerates`) — one cluster, and the one that keeps growing;
- the extent producers (`get_extent`, `_seam_crossing_extent_wkt`, the `extract_metadata` CTE);
- the SQL CRS predicates (`_GEOGRAPHIC_SRTEXT_RE`, `_DEGREE_UNIT_SRTEXT_RE`, the inline `GEOG(CS|CRS)` test), which #961 records as having a standing sync obligation with `core/geo.py`'s helpers — a split has to keep those cross-references true and in one place.

Whoever picks this up should decide the seams from the code, not from this list.

## Acceptance

- `metadata.py` is materially smaller, with its cap lowered in the same commit (lowering is always allowed).
- The extracted modules are covered by a size gate, per #958's inclusion rule: at or above 1000 lines they need their own `_MODULE_LOC_CAPS` entry, and `test_module_loc_cap_inclusion_rule_is_complete` enforces it.
- The #961 cross-reference comments still name the right symbols on both sides.
- No behaviour change; the antimeridian and degenerate-envelope tests are the regression surface.

Issue #961, opened 2026-07-30
chore(backend): four divergent checks for whether a CRS is degree-based — consolidate on one helper


ishiland:
Umbrella for a class problem surfaced during the 2026-07-30 backlog audit; the per-site instances were #939 (live bug) and #906 (annotated). **Both have shipped**, and they took most of this issue with them. Re-scoped 2026-07-31 to what is actually left, which is one judgment call and one standing sync obligation. Size **S**.

## What shipped

The shared helper pair now exists in `core/geo.py`:

| Symbol | Line | What it answers |
|---|---|---|
| `_parse_crs` | :513 | parse the stored WKT with PROJ, `None` when PROJ refuses; `@lru_cache(maxsize=256)` at :512 |
| `wkt_is_geographic` | :544 | geographic or projected, PROJ first, keyword sniff only as the fallback for WKT PROJ cannot parse |
| `wkt_has_degree_unit` | :601 | degrees rather than grads, read from PROJ's `units_factor` against `_RADIANS_PER_DEGREE` (:509) |

Both return `None` for "unknown", and callers are required to treat that as unknown rather than false.

`processing/tiles/router.py:418-423` has adopted the pair: `wkt_is_geographic(asset.crs_wkt)` decides the class, falling back to the historical `asset.epsg == 4326` test only when there is no usable WKT, and `wkt_has_degree_unit(...) is not False` gates the degrees-to-metres conversion so a grads GEOGCS falls through to the bounds-derived estimate instead of being converted as degrees. That closed #939, which was the live bug in the original table.

The regex-versus-parser question is settled and worth not relitigating. `_parse_crs`'s docstring records that rounds 1 through 9 of review on that file were the same defect in different clothes: every narrowing of the scan window was followed by another WKT element that legitimately carries an angular unit outside it (PRIMEM, then MERIDIAN, with BEARING and the conversion PARAMETERs still unclaimed). A flat scan of a nested grammar cannot decide which subtree a unit belongs to, so the helper asks PROJ.

## What is left

**1. Decide whether `vrt.py`'s `_is_degree_based` converges.** It is at `processing/raster/vrt.py:254`, called from :388 and :627, and it takes a **CRS object**, not a WKT string, so it cannot call the new helpers as they stand. It makes the identical test (`crs.is_geographic` plus a `units_factor` comparison against the same `math.pi / 180.0` constant), and `wkt_has_degree_unit`'s docstring already names it as the site it must not drift from. Two options:

[...]

PR #1044, opened 2026-07-31
test(backend): give the module LOC ratchet a written inclusion rule

### LINKED ISSUE #958: test(backend): LOC-ratchet coverage is inconsistent — analysis_sql.py is ungated and grew 504 to 522 unnoticed

ishiland:
## The decision

#958 offered two options and observed that the first does not work: adding `analysis_sql.py` alone ratchets the 24th-largest ungated module while five modules above 1000 lines stay free to grow, which is the same arbitrariness one rung down. So this takes the second — **define the inclusion rule** — and makes it executable rather than a comment:

> Every module under `backend/app/` at or above **1000** lines that no OTHER size gate watches belongs in `_MODULE_LOC_CAPS`, ratcheted at its exact LOC.

`test_module_loc_cap_inclusion_rule_is_complete` enforces it. Crossing the line now fails with instructions instead of waiting for someone to re-measure — which is the actual failure mode here, since `analysis_sql.py` grew 504 → 651 in the gap between two audits.

`analysis_sql.py` at 651 is therefore **out by rule**, not by oversight. That was the ask.

## Why 1000

It is where the modules this project has actually had to decompose live (the Phase 226 / 238 / 252 splits all started past it), and the measured distribution has a break there: a cluster at 1017-1201, then a gap down to 990. Any threshold is arguable; an unwritten rule is not enforceable.

Five modules already qualified and are ratcheted at their measured size:

| module | LOC |
|---|---|
| `platform/config_ops/service.py` | 1201 |
| `processing/ingest/tasks_vrt.py` | 1071 |
| `processing/ingest/tasks_vector.py` | 1058 |
| `modules/auth/oauth/service.py` | 1031 |
| `processing/ingest/service.py` | 1017 |

They enter with no carve-out history, because none of them has ever had to argue for a line. That is the gap.

## Why "no OTHER gate" rather than unconditionally

A module a ceiling gate already watches is not silent — it hits a wall, and whoever hits it ratchets the file in. That is exactly how `ingest/router.py` got into the dict. Ratcheting them anyway would put exact caps on four routers sitting under the 1500 glob ceiling (`platform/jobs/router.py` 1130, `sources/router.py` 1030, `auth/router.py` 1010, `settings/router.py` 1004) and on `maps/style_json.py` (1432, capped by the open-core gate), doubling the bookkeeping for files that already fail loudly.

`size_caps` moves out of `test_open_core_decomposition_boundaries_stay_clean` to module scope as `_OPEN_CORE_SIZE_CAPS` so the rule can consult it. Values and semantics unchanged (plain ceilings, not exact ratchets).

**Residue, recorded in the code rather than papered over:** a `router.py` at 1400 still has 100 lines of silent runway under the glob default, and the two routers-by-role the glob's filename match cannot see (`datasets/api/router_export.py` 928, `router_reupload.py` 922) are watched by nothing until they cross 1000. #958 notes that any rule phrased as "routers are covered by the glob" is already inaccurate, which is why this one is phrased by size.

## The local-blindness half

#958's other remedy: the ratchet does run on every PR (`ci.yml:1080-1087`, ahead of the coverage run), so the gap is local — a focused test selection gives no signal and you find out at CI. `AGENTS.md` now says so next to the "Running a single test" recipes, with the one-line check to finish on (`uv run pytest tests/test_layering.py -q`, no database, seconds).

## metadata.py

Not decomposed here — it is an L-sized refactor, not a test change. Filed as **#1042** with the seams visible from its cap comments and the #961 sync obligation, and the cap comment now points there. The four carve-outs (#888, #899 codex r1, #906, #934) stay as the evidence.

None. Test and docs only.

## Verification

```
cd backend && uv run pytest tests/test_layering.py -q      # 35 passed
```

Negative control, run by deleting the `config_ops/service.py` entry: the new test fails with
`backend/app/platform/config_ops/service.py: 1201 lines`.

Closes #958

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -25,6 +25,8 @@ The React/Vite frontend is in `frontend/src/`: `components/`, `pages/`, `hooks/`
 - Frontend: `cd frontend && npx vitest run src/path/foo.test.ts`.
 - E2E: `npx playwright test e2e/foo.spec.ts --project=chromium` (stack must be running).
 
+A focused selection is blind to the module-size gates. `backend/tests/test_layering.py` caps the size of the largest backend modules, and CI runs it on every PR that triggers `backend-test`, so a change that adds lines to a ratcheted file passes locally and fails there. If you touched anything under `backend/app/`, finish with `cd backend && uv run pytest tests/test_layering.py -q`; it needs no database and takes seconds. Growth is allowed — raise the file's cap in `_MODULE_LOC_CAPS` in the same commit, with a comment saying what the lines bought.
+
 ### Working from a git worktree
 
 The dev stack bind-mounts the MAIN checkout (`./frontend` → `/app`, and `backend/app` → `/app/app` with `--reload`), so `localhost:8080` always serves `main` no matter which branch your worktree is on. Running `npx playwright test` from a worktree therefore validates code you did not write. Treat the result as meaningless: it has produced a false FAILURE, and the symmetric case is worse, because a worktree change that breaks e2e passes when the stack never had it. To exercise worktree app code, run Vite on the host at `:5174` with `API_PROXY_TARGET=http://localhost:8001` and point `E2E_BASE_URL` at it.
--- a/backend/tests/test_layering.py
+++ b/backend/tests/test_layering.py
@@ -1257,18 +1257,29 @@ def test_module_loc_caps_have_no_headroom() -> None:
 # which is why this one is phrased by size.
 _RATCHET_INCLUSION_LOC = 1000
 
-# The decomposition globs from
-# test_decomposed_service_modules_stay_within_size_budgets, as filename
-# prefixes — the same way that test discovers its own files.
-_DECOMPOSED_MODULE_PREFIXES = ("service_", "chat_", "defaults_")
+# The globs of test_decomposed_service_modules_stay_within_size_budgets, as
+# (directory, filename prefix) pairs. fix(#958 review): the prefix alone is not
+# the gate. That test globs specific directories, so a `service_*.py` anywhere
+# else is watched by nothing — and a filename-only exemption here would have let
+# such a module past 1000 lines through both gates, defeating the rule this
+# file exists to state. Mirror the directories, not just the names.
+_DECOMPOSED_MODULE_SCOPES: tuple[tuple[str, str], ...] = (
+    ("backend/app/modules/catalog/maps/", "service_"),
+    ("backend/app/modules/catalog/search/", "service_"),
+    ("backend/app/modules/catalog/datasets/domain/", "service_"),
+    ("backend/app/processing/ai/", "chat_"),
+    ("backend/app/platform/extensions/", "defaults_"),
+)
 
 
 def _is_watched_by_another_size_gate(rel: str, name: str) -> bool:
     """True when some gate other than _MODULE_LOC_CAPS already caps this file."""
     if name == "router.py":
         return True  # test_router_orchestrator_modules_stay_within_loc_cap
-    if name.startswith(_DECOMPOSED_MODULE_PREFIXES):
-        return True  # test_decomposed_service_modules_stay_within_size_budgets
+    for directory, prefix in _DECOMPOSED_MODULE_SCOPES:
+        # Directory, not prefix path: those globs are non-recursive.
+        if rel == f"{directory}{name}" and name.startswith(prefix):
+            return True  # test_decomposed_service_modules_stay_within_size_budgets
     return rel in _OPEN_CORE_SIZE_CAPS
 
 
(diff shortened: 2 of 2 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

Issue #956, opened 2026-07-30
feat(analysis): intersect/overlay — new geometry from two layers


ishiland:
Split out of #681 (phase-2 umbrella), which tracked four unrelated operations behind one body. This is operation 4 of 4, and the lowest priority of the four.

Intersect/overlay is distinct from spatial join, and the distinction is what justifies a separate operation: a join transfers attributes or counts between layers, while an overlay produces **new geometry** whose area you then measure. Impact acres and overlap analysis are the motivating cases. "How many acres of this parcel fall inside the flood zone" is an overlay followed by a measurement, not a join.

## Settled: a new operation, not a flag on clip

Clip already computes `ST_Intersection` against a mask layer and is fully shipped. The differences are real but narrow, so they are written down here before anyone builds a second code path:

- **Clip is asymmetric and attribute-preserving on one side.** It keeps the source's carry columns (`_list_carry_columns`, `backend/app/processing/analysis/tasks.py:441-461`, wired at `:683`) and discards the mask's entirely. An overlay wants columns from *both* inputs on each output piece.
- **Clip's mask is unioned in effect; overlay's is not.** `render_clip_layer_join` (`backend/app/platform/analysis_sql.py:477`) aggregates the per-piece intersections back to one geometry per source row, so a source polygon overlapping three mask polygons yields one output row. An overlay wants three rows, one per pair, each carrying that mask feature's attributes.
- **Clip's mask must be polygonal** (`_load_mask_dataset`, `backend/app/modules/catalog/datasets/api/router_analysis.py:117-121`).

The decision is a **new operation** that shares the subdivision machinery but emits pair-wise rows. The row cardinality differs, which means the size ceilings, the output-size enforcement, and the "one row per source feature" assumption baked into the preview all differ. A flag that changes cardinality is a worse API than a second name.

## Blocker: the output table's primary key

This is the thing that makes overlay more than a variant of clip, and it must be solved first.

[...]

Issue #953, opened 2026-07-30
feat(analysis): spatial join — point-in-polygon count and attribute transfer


ishiland:
Split out of #681 (phase-2 umbrella), which tracked four unrelated operations behind one body. This is operation 1 of 4, and the highest priority: it ships in every comparable product surveyed (Felt, Atlas.co, CARTO, ArcGIS Online, QGIS, Turf.js).

Two variants are in scope:

- **Count / aggregate into the polygon layer**: each polygon gains a count (and optionally a sum/avg over a numeric column) of the point features falling inside it.
- **Attribute transfer onto the point layer**: each point gains one or more columns from the polygon containing it.

## What the gap actually is

An earlier version of this body claimed chat cannot answer "how many X are in Y". That is wrong, and the correction changes what this issue is for.

The NL to SQL chat path already answers these questions as tables through `query_data`. `backend/app/processing/ai/sql_generator.py:322-327` ships a spatial-join-with-aggregation example verbatim (`COUNT(...) ... JOIN ... ON ST_Intersects(...) GROUP BY`), and `:329-332` ships the acreage example. The sandbox allowlist admits `st_intersects`, `st_within`, `st_area`, `st_length`, and `st_dwithin` (`backend/app/platform/sandbox/validator.py:263-282`), so those queries validate and run today.

The real gap is a **deterministic, non-LLM result that lands on the map or becomes a dataset**. Chat gives a table in a chat transcript, generated by a model, not reproducible on rerun and not a layer. The analysis rails give a previewed geometry the user approves and a materialized dataset they own. Note also that `ST_Intersection` is deliberately absent from the sandbox allowlist, and the reason is written out at `backend/app/modules/catalog/datasets/domain/service_analysis.py:7-10`: widening the chat validator's PostGIS allowlist to serve clip would expand the chat attack surface for no benefit on the analysis path. Anything that needs `ST_Intersection` can only come from these rails.

## Settled: API shape

**Flat schema, new enum value on the existing routes.** `operation: "spatial_join"` joins `Literal["buffer", "centroid", "clip"]` at `backend/app/modules/catalog/datasets/domain/schemas.py:887` (preview) and `Literal["buffer", "centroid", "clip", "dissolve"]` at `:943` (materialize).

[...]

PR #1045, opened 2026-07-31
feat(analysis): write provenance onto materialized analysis outputs

### LINKED ISSUE #765: feat(analysis): write provenance onto materialized datasets (lineage, derived-from link, inherited keywords)
**Where:** `backend/app/processing/analysis/tasks.py:767-773` (registration), `backend/app/modules/catalog/datasets/api/router_analysis.py:325-341` (the analysis metadata block)

> Citations refreshed 2026-07-31 against `main` at `9a788ba92`. The original body cited `tasks.py:534-541` and `router_analysis.py:282-295`; the router file has since gained nine lines, and `:282-295` now lands inside the per-user job cap query, which would send a reader to entirely the wrong code.

[live] A materialized result lands with empty lineage, empty description, empty keywords — 17% metadata completeness — although the operation, parameters, source dataset, user, and timestamp are all known at creation time. `source_dataset_id` lives only in purgeable `job.user_metadata`, so a derived dataset has no durable link to its origin.

Confirmed still true: registration passes `RegisterRequest(table_name=out_table, title=title, visibility="private")` and nothing else (`tasks.py:769-771`), and `source_dataset_id` is written only into `job.user_metadata` (`router_analysis.py:330`).

**Proposal:** at registration, write (1) a generated lineage sentence ("Buffered 'World Lakes & Reservoirs' by 5 miles, 2026-07-27, admin"), (2) a durable `derived_from` reference usable by the UI ("derived from X" chip + re-check point at read/export time), (3) inherited keywords from the source. Directly feeds the metadata-completeness gate and mitigates the duplicate-name ambiguity.

## Settled design (2026-07-27)

Where the durable reference lives: **a new nullable JSONB column `derived_from` on `Record`**, carrying `{dataset_id, operation, params, created_at}` written at materialize time. The existing Record JSONBs (`column_info`, `sample_values`, `quality_detail`) are all purpose-typed; overloading one would be worse than the trivial additive migration. The migration is nullable + no backfill, so cheap and safe. The real cost is the record-schema/OpenAPI/SDK regen.

Riding along:

- Auto-fill the existing `lineage_summary` text column with a human sentence ("Buffered from *Title* (500 m)…"). Verified: the column exists at `backend/app/modules/catalog/datasets/domain/models.py:192`, and DCAT already exports it as `dcterms:provenance` (`backend/app/standards/dcat/service.py:315-317`), so prose lineage surfaces in standards output for free. Nothing on the analysis path writes it today.
- STAC read-side: emit the standard `rel="derived_from"` link when the column is set. No `derived_from` exists anywhere in `backend/app/standards/` today, so this is net-new.
- **Keywords: inherit by copy at creation, including from restricted sources.** The creator necessarily has read access (Rule 1 gates the analysis), and copied keywords disclose nothing beyond what the derived geometry already embodies.
- **Authz on display:** record detail / STAC only emit the `derived_from` reference when the requester can access the source dataset (existing cross-dataset authz pattern); otherwise omit it entirely, so there is no dangling UUID disclosure.

## Keywords are child rows, not an array column

The single most likely way to lose an afternoon on this issue. `Record.keywords` is a **relationship**, not a list column: `keywords: Mapped[list["RecordKeyword"]]` at `models.py:253-259`, backed by the `record_keywords` table (`class RecordKeyword`, `models.py:542`) with `cascade="all, delete-orphan"` and a `keyword_type` CHECK constraint restricting the type to a fixed vocabulary (`models.py:546-551`). Inheriting keywords therefore means **selecting the source record's `RecordKeyword` rows and inserting copies pointing at the new record**, carrying `keyword_type` across. It is not an assignment.

The decoy sitting a few lines away is `theme_category` (`models.py:199`), which **is** a real array column. It is a different field with different semantics, it feeds the full-text expressions alongside `lineage_summary`, and assigning inherited keywords into it will look like it works.

## Writing `lineage_summary` changes search results

Worth knowing before treating it as a cosmetic text field: it feeds **two** full-text index expressions on `Record`, at `models.py:147` and `models.py:209`, the latter with `setweight(..., 'C')`. Populating it for every materialized dataset therefore changes catalog search ranking, and a dataset that was previously unfindable by its operation name becomes findable.

That is the desired outcome, but it should be asserted rather than discovered: add a test that materializes a dataset and then finds it by a term appearing only in the generated lineage sentence.

## It also flips the completeness gate

`lineage_summary` is a **hard** (VAL-01) validation requirement, not a soft one: an empty one raises `ValidationIssue("lineage_summary", "Lineage summary is required", "error")` at `backend/app/modules/catalog/validation/service.py:70-73`. Every materialized analysis output fails that check today. Populating the column flips them to valid on that gate.

That is the point of the issue, but it is a behaviour change to a gate rather than a cosmetic addition, so state it in the PR body and pin it with a test: a freshly materialized dataset should report no `lineage_summary` error where it previously did.

## Where to edit STAC

`_build_stac_links` at `backend/app/standards/stac/serializer.py:47-87` builds the item link list (self, root, parent, and a conditional `collection`); the collection link list is inline in `ogc_collection_to_stac_collection` at `:275-296`. The `derived_from` link belongs in the item list, gated on the authz check above so it disappears rather than dangles for a requester who cannot see the source.

## Sequencing: land this before #953-#956

The four new-operation issues (#953 spatial join, #954 measurement, #955 select by location, #956 intersect/overlay) each add an operation that will want a lineage sentence of its own. Landing this first gives them a lineage generator to extend with one case each. Landing it after means four operations ship with no provenance and then get retrofitted.

## Regeneration and verification

A new `Record` column touches four gates. None is optional:

- Alembic migration (nullable, no backfill), then `make alembic-check`, the CI gate that fails when the ORM models have drifted from the migration scripts.
- `make openapi`: the record schema is part of the API surface.
- `make sdks`: regenerated from the OpenAPI snapshot; never hand-edit generated files.
- Commit `backend/openapi.json` and the SDK output together with the source change. `make openapi-check` and `make sdks-check` fail on drift.

## Acceptance criteria

- [ ] `derived_from` JSONB column added to `Record`, nullable, no backfill.
- [ ] A materialized analysis output carries `derived_from`, a populated `lineage_summary`, and keywords inherited from the source, asserted end to end, not at the unit level.
- [ ] Inherited keywords are `record_keywords` child rows on the new record, with `keyword_type` carried across, not values written into `theme_category`.
- [ ] The generated lineage sentence names the operation, the source title, the parameters, and the actor, and reads as a sentence rather than a serialization.
- [ ] `derived_from` is omitted entirely (not nulled, not stubbed) for a requester who cannot access the source dataset, in both record detail and STAC. Pinned by a test with two users.
- [ ] Keywords inherit even when the source is restricted, and a test records that this is deliberate.
- [ ] STAC emits `rel="derived_from"` when the column is set.
- [ ] A search test finds a materialized dataset by a term unique to its generated lineage sentence.
- [ ] A test asserts a freshly materialized dataset no longer fails the VAL-01 `lineage_summary` check.
- [ ] Alembic migration + `make alembic-check` + `make openapi` + `make sdks`, all four committed.

---

ishiland:
A materialized analysis output used to land with an empty lineage, no keywords, and no durable link to the dataset it came from. That is 17% metadata completeness for a dataset whose operation, parameters, source, actor and timestamp were all known at creation time. `source_dataset_id` lived only in `ingest_jobs.user_metadata`, which is purgeable.

## What lands

`records.derived_from` is a new nullable JSONB column (migration `0033_records_derived_from`, no backfill) carrying `{dataset_id, operation, params, created_at}`. The drawn clip mask is left out of `params` for the same reason the job metadata leaves it out: it can be kilobytes of geometry.

`records.lineage_summary` gets a generated sentence: `Buffered from "World Lakes" by 500 m, created by admin on 2026-07-31.` The generator lives in `backend/app/processing/analysis/provenance.py`, one branch per operation plus a fallback, so #953 through #956 each add a branch instead of retrofitting provenance after the fact.

Keywords inherit from the source. They are `record_keywords` child rows rather than `theme_category` values, so this is an `INSERT … SELECT` that carries `keyword_type` across, not an assignment.

All three are written in the same transaction as registration. A registered output can never be visible without its provenance, and an attempt that loses the fence rolls the provenance back with the dataset row.

## Two behaviour changes worth stating

Search results change. `lineage_summary` feeds two full-text index expressions on `Record` (`models.py:147` and `:209`, the second at weight C), so a derived dataset becomes findable by its source's title and its operation name. That is the intent, and there is a test for it.

The VAL-01 completeness gate flips. An empty `lineage_summary` raises a hard validation error, so every materialized output failed that check on creation. They pass it now. Also tested.

## Access control

`derived_from` points at another dataset, and a derived output can be shared while its source stays private. Both read paths gate it:

- Record detail resolves it through `visible_derived_from` (`authorization.py`), which returns `None` unless the requester can open the source. A denied requester sees exactly what a never-derived dataset shows, so "not derived" and "derived from something you cannot see" are indistinguishable.
- STAC emits `rel="derived_from"` through `_visible_derived_from_id`, gated on the same published-and-visible query the item endpoints serve from, so the link can never point at an item the requester would get a 404 for.

Keywords are deliberately not gated, and the reason is not that a keyword is harmless. It can be a project codename or a client name, which no geometry embodies. It stands on the same ground the lineage sentence does: the output is registered `visibility="private"` (`tasks.py:771`) and owned by the caller who already held access to the source, so an inherited keyword reaches anyone else only when that owner publishes or shares it, on ordinary editable record metadata they can delete first.

`test_detail_omits_derived_from_without_access_to_the_source` pins both halves rather than describing them: the outside viewer gets a 404 on the private source's own keywords, and a 200 on the copy carried by the published output. Warning an owner what metadata they inherited before they publish is #1070.

Detail only, not list. `dataset_to_response` takes the reference already access-checked, because emitting it from the list builders would mean a per-row source check on every page.

## Two things a reviewer should know

STAC Items are served for the raster family only (`_STAC_RECORD_TYPES = RASTER_FAMILY_RECORD_TYPES`), so a vector analysis output does not appear there today and the `derived_from` link cannot fire end to end for one. The serializer and the gate are covered by tests; the link becomes reachable the day STAC serves vector items.

`derived_from` is a declared response model, not a `dict[str, Any]`. As a bare dict it reached the checked-in OpenAPI as `additionalProperties: true` and both SDKs lost the shape with it: the TypeScript client degraded to an index signature, the Python one to an empty additional-properties container. The stable shape was documented in prose and mirrored by hand in the frontend types, while the SDKs (where most consumers meet this API) could not use it type-safely. `DerivedFromResponse` types `dataset_id`, `operation` and `created_at`; `params` stays open because its keys differ per operation and it is redacted per requester on top of that, so a per-operation union would describe a shape the redaction is free to punch holes in.

The hand-typed frontend mirror (`types/api.ts`) does not regenerate, so it carries a matching `DerivedFrom` interface written by hand. Nothing in the UI reads the field yet — the "derived from X" chip is follow-up work.

## Impact

Schema: one nullable column, no backfill, no data migration. API: `DatasetResponse.derived_from` added, with `openapi.json`, `api.generated.ts` and both SDKs regenerated and committed. This started as `0031` and was renumbered after #1052 and #1055 landed ahead of it, so it is now `0033_records_derived_from` on `down_revision = "0032_api_key_scope"`. Verified rather than assumed: `alembic heads` shows the single head, `alembic history` a linear `0031 → 0032 → 0033`, and `scripts/test_alembic_upgrade_clean_db.sh` upgrades a fresh DB to head.

Closes #765.

## Verification

```
alembic upgrade heads && alembic check       # No new upgrade operations detected
uv run pytest tests/test_analysis_provenance.py -v          # 11 passed
uv run pytest -n 4                                          # full suite
make openapi && make sdks && make openapi-check && make sdks-check
```

Two module LOC caps are raised in `test_layering.py` with the usual in-place justification: `service_query.py` 397 to 404, and `stac/router.py` 1789 to 1833.


ishiland:
@codex review

ishiland:
@codex review

ishiland:
Two findings, two different answers.

**Mask id in `params` — fixed.** Correct: a clip carries a second dataset id, a public output can be derived from a public source through a private mask, and `visible_derived_from` returned the whole dict on the source's verdict alone. Every id in the reference is now checked on its own, a failing one is dropped while the rest survives, and the returned dict is a copy so a caller cannot mutate the record's JSONB through it. Two tests cover it (private mask dropped, visible mask kept).

**Source title in `lineage_summary` — kept, deliberately, now recorded in the code.** The output is registered `visibility="private"` and owned by the caller who already held access to the source, so the title only reaches anyone else if that owner publishes or shares the dataset. That is a deliberate act on prose they can read and edit, and `lineage_summary` is an editable metadata field. It is also the same disclosure model #765 settled for inherited keywords, which are copied from restricted sources on purpose. Naming the source is an explicit acceptance criterion on the issue ("the generated lineage sentence names the operation, the source title, the parameters, and the actor"), so dropping it would fail the spec. The rule the code now states: references are gated, copied prose is not.

ishiland:
@codex review

ishiland:
Rebased onto #1040, which touched `DatasetResponse` too and left the generated Python model conflicting. Resolved by re-running `make openapi && make sdks` per the regen rule rather than hand-merging — worth noting the difference: my conflict resolution alone dropped #1040's `extent_bbox` docstring change, and the regeneration put it back alongside `derived_from`. Backend suites re-run green after the rebase.

ishiland:
@codex review

ishiland:
@codex review

ishiland:
Keyword disclosure — argument replaced, decision pinned, code unchanged.

You were right that the bullet was load-bearing on the wrong fact. "A keyword discloses nothing the derived geometry does not already embody" does not survive a codename or a client name, and no geometry embodies those. The hole was in the argument, not the code.

Rewritten to stand where the lineage-sentence bullet already stands, which I verified rather than inherited from the docstring: `tasks.py:771` registers the output `visibility="private"`, owned by the caller who by Rule 1 already held source access, so an inherited keyword reaches anyone else only when that owner publishes or shares it. `DELETE /records/{id}/keywords/{id}/` behind `edit_metadata` plus an ownership check is what makes "they can strip it first" a capability rather than a turn of phrase.

Pinned in `test_detail_omits_derived_from_without_access_to_the_source`, which already built this exact scenario. Both halves, because the second is vacuous without the first: the outside viewer gets **404** on the private source's own keywords, and **200** on the copy the published output carries. I probed the 404 half before asserting it, on the chance the endpoint was ungated and the assertion would have proved only that inheritance happened.

Warning an owner what metadata they inherited before they publish is the residual product question, filed as #1070.

Also folded in: the body now says `0033_records_derived_from` on `0032_api_key_scope` and drops the stale "renumbers off 0031" note.

ishiland:
@codex review

ishiland:
Provenance response schema — fixed, and verified through all three regen artifacts.

You were right that it defeats the PR's own purpose. `dict[str, Any]` reached the checked-in OpenAPI as bare `additionalProperties: true`, so the shape existed in three places that do not matter to SDK consumers (prose, a hand-written frontend mirror, a docstring) and in none that do.

`DerivedFromResponse` now types `dataset_id` (`format: uuid`), `operation`, and `created_at` (`date-time`). Checked rather than assumed the regen did what I wanted:

- `openapi.json` — `components.schemas.DerivedFromResponse` present
- Python SDK — `models/derived_from_response.py` emitted
- TypeScript SDK — `export type DerivedFromResponse`, with `derived_from?: DerivedFromResponse | null`

`params` deliberately stays open, and the reason is stronger than "its keys differ per operation": `visible_derived_from` **redacts** it per requester, dropping any embedded dataset id the caller cannot see. A per-operation union would describe a shape the redaction is free to punch holes in — so the model docstring states that, to stop it being "tightened" later into a lie.

`frontend/src/types/api.ts` does not regenerate, so its inline shape was lifted to a named `DerivedFrom` interface by hand to match.

Ran the full suite under `-n 4` rather than a targeted selection, since `visible_derived_from` reads `user_roles` and that is not something a narrow run can vouch for: **6387 passed, 82 skipped**. `make openapi-check` and `make sdks-check` both pass at this head.

Also corrected a paragraph in the PR body that claimed the frontend mirror was untouched, which stopped being true two commits ago.

ishiland:
@codex review

ishiland:
Buffer distance rounding — fixed, and it was wider than the finding said.

`f"{number:g}"` defaults to **six significant digits**. The comment I had written beside it claimed it dropped the trailing `.0` "without rounding real decimals", which is simply false, and is why this survived two review rounds — the wrong claim was doing the reviewing.

Measured across the valid range rather than reasoned about:

| submitted | `:g` rendered |
|---|---|
| `12345.678` | `12345.7` |
| `33.333333333` | `33.3333` |
| `99999.99` | `100000` |

The finding framed it as ">6 significant digits"; an ordinary `12345.678 m` buffer already trips it. And the last row is the one I would flag back: `99999.99` is a legal distance just under `MAX_BUFFER_METERS`, and the sentence rounded it **up past the ceiling it had been validated against**.

Now formatted via `repr`, which is the shortest string that round-trips to the same double — exact by construction, rather than by picking a precision that looks big enough. Only the trailing `.0` on whole metres is trimmed, which is the one thing the old formatting was wanted for. Six cases pinned in `test_the_buffer_distance_survives_the_sentence`, including the three that were broken.

Full suite `-n 4` at this head: **6388 passed, 82 skipped**.

---

**Heads-up on this PR's state, which is not about the code.** The thirteen-PR batch moved main to `f0113cdfa` and this PR went `BEHIND` -> `DIRTY`. GitHub does not run `pull_request` workflows on a conflicting PR, so both pushes since the batch have zero checks — that is the mechanism, not a flake.

The conflict is exactly one file, `sdks/typescript/src/client/index.ts`. Everything else auto-merges, `backend/openapi.json` and `test_layering.py` included. Being a generated artifact, the right resolution is `make sdks` after a rebase rather than a hand-merge — hand-resolving the earlier #1040 rebase silently dropped that PR's `extent_bbox` docstring change, and only the regeneration put it back.

Not rebasing on my own initiative. Main's own push run at `f0113cdfa` has since reported green (CI + CodeQL), so the "rebases onto a red main" risk is no longer present, but the call is yours.

Issue #688, opened 2026-07-25
feat(tiles): signed raster tile URLs so API-key clients can render private rasters


ishiland:
`/api/tiles/token/{dataset_id}/` returns a signed template for **vector** tiles (`sig`/`exp`/`scope`, HMAC over `{scope}:{exp}`, 15 min TTL) but the **raster** shape carries no signature at all:

```
{"kind": "raster",
 "tile_url": "/raster-tiles/{id}/tiles/{z}/{x}/{y}.png",
 "bounds": [...], "minzoom": 0, "maxzoom": 22, "tile_size": 256,
 "format": "png"}
```

`RasterTileToken` (`backend/app/processing/tiles/schemas.py:15-22`) has no `sig`/`exp`/`scope`, where `VectorTileToken` (`schemas.py:7-12`) has all three. `_resolve_raster_access` (`backend/app/processing/tiles/router.py:753-823`) has three auth branches — embed token, authenticated user, public+published — with no HMAC path. The unsigned raster template is minted at `router.py:1226-1236`. So a client following the API contract literally receives an **unauthenticated template for a private raster**.

Third-party clients hit this. GeoLibre's `geolens-api.ts` documents it as a known limitation:

> an API-key-only private raster cannot render, because MapLibre issues the tile image requests and does not attach the `X-Api-Key` header, and GeoLens does not (yet) return a URL-signed raster template the way it does for vector tiles

## Workarounds that exist today

- `map.setTransformRequest()` attaches headers per request (this is what `frontend/src/components/dataset/DatasetMap.tsx:618` does with a bearer token). Verified 2026-07-25 that the raster path is CORS-clean cross-origin: preflight from another origin for `/raster-tiles/{id}/tiles/{z}/{x}/{y}.png` with `X-Api-Key` returns 200 with the header allowed.
- `?api_key=` on the tile URL works via `_resolve_api_key`, but puts a non-expiring unscoped credential into tile URLs, server logs, and any saved client project file. Not something to recommend.

Neither is a substitute for the contract being correct. Proposal: mirror the vector path — add an HMAC branch to `_resolve_raster_access` and return a signed `tile_url` from `_build_tile_token_for_dataset`, so the returned template is self-sufficient and expires.

## Reference implementations on the vector path

The mirror already exists and works:

[...]

PR #1046, opened 2026-07-31
chore(tiles): unpublish the dead raster-auth-check route from the API contract

### LINKED ISSUE #957: chore(tiles): drop the dead raster-auth-check route from the API contract and both SDKs
## What

The `/tiles/raster-auth-check/` **route registration** (`backend/app/processing/tiles/router.py:826`) has no HTTP caller. It was the target of the nginx `auth_request` subrequest that used to authorize raster tiles before nginx proxied them straight to Titiler. That topology is gone: `frontend/nginx.conf:104-116` now forwards raster tile requests to the api-side proxy instead, and the block says so explicitly:

```
# Forward to the api-side raster proxy (`/tiles/raster-proxy/...`).
# The api enforces RBAC, resolves the COG open-path, and proxies to
# Titiler with the correct render params. Equivalent semantics to
# the previous nginx auth_request + direct-titiler proxy, without
# the variable-interpolation pitfalls of `proxy_pass $upstream/path?query`
# inside a regex location [...]
```

**The handler function itself is load-bearing and must not be deleted.** `raster_tile_proxy` (`router.py:905`) calls `raster_auth_check` in-process at `router.py:1035` and reads four headers off the `Response` it returns:

```python
auth_resp = await raster_auth_check(request, dataset_id, user, db)
open_path = auth_resp.headers.get("X-GeoLens-Asset-OpenPath")
render_params = auth_resp.headers.get("X-GeoLens-Render-Params", "")
cache_status = auth_resp.headers.get("X-GeoLens-Cache-Status", "private")
_raw_band_count = auth_resp.headers.get("X-GeoLens-Band-Count", "1")
```

So the function is the RBAC-plus-open-path-resolution step of the live raster tile path. Only the decorator that publishes it as an HTTP route is vestigial.

## Why it matters

The dead route registration is a published endpoint in the API contract and both generated SDKs:

- `backend/openapi.json:54710`
- `frontend/src/types/api.generated.ts:4857`
- `sdks/python/geolens/api/tiles/raster_auth_check_tiles_raster_auth_check_get.py`
- `sdks/typescript/src/client/sdk.gen.ts:5331`

SDK consumers can discover and call an endpoint whose only purpose was to answer an internal nginx subrequest. What a direct caller receives is the `X-GeoLens-Asset-OpenPath` header, which carries the storage open-path (`/vsis3/bucket/key` and similar). That path contains no credentials and is only returned to callers who already passed the dataset RBAC check, so this is internal topology disclosure to authorized callers rather than a data leak. It is still not a shape worth shipping in an SDK.

## Fix

Add `include_in_schema=False` to the decorator at `router.py:826`. The route stays mounted (tests keep working), and it drops out of `openapi.json`, `api.generated.ts`, and both SDKs.

Deleting the handler is not an option: it would break the raster tile path at `router.py:1035` and 21 HTTP call sites across five test files (`test_raster_tiles.py`, `test_vrt_titiler.py`, `test_perf002_raster_meta_cache.py`, `test_raster_colormap_proxy.py`, `test_tenant_session_guc.py`). Those tests are the coverage for raster RBAC, including the path-parity test at `test_raster_tiles.py:660` that asserts the auth-check path and the token endpoint block the same non-owner viewer.

Steps:

1. Confirm no deployment topology still calls the route over HTTP. `grep -rn "raster-auth-check"` in this repo returns only generated artifacts, tests, and the definition, but check the Helm chart and any operator-facing nginx overrides in the deployment repos too, since a self-hoster could still be running an older ingress config with `auth_request`.
2. Set `include_in_schema=False` on the decorator, then regenerate contracts: `make openapi`, then `make sdks`. `make openapi-check` and `make sdks-check` are the CI gates.
3. Removing a route from the generated SDK surface is a breaking change for anyone who called it, so note it in the CHANGELOG.
4. Update the handler docstring, which still describes it as "called by nginx auth_request", and add a line recording that the in-process caller is `raster_tile_proxy`.

## Size

S.

## Context

Found during the 2026-07-30 feature/roadmap issue audit while verifying #688's premise (that `_resolve_raster_access` has no HMAC branch — it does not, and that part of #688 stands). Do this before #688: same file, trivial, and it removes a decoy that anyone implementing signed raster URLs would otherwise read first.

Related: #688.

ishiland:
Closes #957. Sequenced before #688 — same file, and this removes the decoy
anyone implementing signed raster URLs would read first.

## What was dead, and what was not

The **route registration** at `tiles/router.py:826` was the nginx `auth_request`
target from when nginx proxied raster tiles straight to Titiler.
`frontend/nginx.conf:104-116` now forwards them to the api-side
`/tiles/raster-proxy/`, and says so at the block. So the endpoint has no HTTP
caller, while still being published in `openapi.json`, `api.generated.ts` and
both SDKs — where the only thing a direct caller gets back is the storage
open-path (`/vsis3/bucket/key`) in a response header.

The **handler** is load-bearing and is not touched. `raster_tile_proxy` calls
`raster_auth_check` in-process at `:1035` and reads four `X-GeoLens-*` headers
off the Response it returns, so it is the RBAC-plus-open-path step of the live
raster tile path. It also stays mounted so the raster-RBAC coverage keeps
driving the real handler over HTTP — 21 call sites across five test files,
including the path-parity test at `test_raster_tiles.py:660`.

`include_in_schema=False` splits those two exactly. The decorator now carries a
note saying which half is vestigial, since "dead route" reads like "dead code".

## Deployment check (issue step 1)

`grep -rn "raster-auth-check"` in this repo returns only the definition, tests,
and generated artifacts. Neither `geolens-deploy` nor `geolens-deployments`
mentions `raster-auth-check` or `auth_request` anywhere.

## Contract regeneration

`make openapi`, `make sdks`, `npm run types:generate`. The endpoint drops out of
all four artifacts, including the deleted
`sdks/python/geolens/api/tiles/raster_auth_check_tiles_raster_auth_check_get.py`.
Removing a route from the SDK surface is breaking for anyone who called it, so
there is a CHANGELOG entry under Unreleased → Removed pointing at
`GET /tiles/token/{dataset_id}/` instead.

## Verification

```
  tests/test_raster_tiles.py tests/test_vrt_titiler.py \
  tests/test_perf002_raster_meta_cache.py tests/test_raster_colormap_proxy.py \
  tests/test_tenant_session_guc.py tests/test_layering.py -q
  # 182 passed, 2 skipped
make openapi-check
```

The `processing/tiles/router.py` LOC ratchet moves 2069 → 2081 for the
decorator note, exact as the gate requires.


ishiland:
@codex review

PR #1047, opened 2026-07-31
chore(a11y): document the row shortcuts and extend the axe suite to the builder panels

### LINKED ISSUE #806: chore(a11y): document row-focus shortcuts in the sheet; extend the axe e2e suite to AnalysisPanel and layer-editor tabs
Split from #788 (items 6 and 7), moved to v1.7.0 per the 2026-07-27 triage; #788 stays in v1.6.0 scoped to its items 2 and 4 (items 1/3/5 shipped in PR #804).

The two items below are unrelated to each other and can be picked up independently. Re-verified against `main` at `9a788ba92` on 2026-07-31.

1. **Shortcuts sheet gaps.** Two entries are still missing from `KeyboardShortcutsSheet.tsx` (its `shortcuts` array runs `:44-104`):
   - **Space's row-focused meaning is undocumented.** `StackRow.tsx:317-320` maps `e.key === ' '` on a focused row to `onCmdClick?.(layer.id)` — its own comment reads "Space = Cmd-click (toggles multi-selection)". The sheet documents multi-select only as ⌘/Ctrl+Click (`:89-93`), so a keyboard-only user has no way to discover it. (The original body cited `StackRow.tsx:237-240`, then `:308-319`; the handler has drifted again, so match on the `e.key === ' '` symbol rather than the line.)
   - **Double-click-to-rename is absent.** The handler is the `onDoubleClick` on the layer-name span at `StackRow.tsx:419`; the sheet has no rename row at all.

   ~~The existing "Space/Enter — toggle reorder" entry should be scoped to the drag handle~~ — **Shipped:** `KeyboardShortcutsSheet.tsx:75-79` already reads "Toggle reorder mode (focused drag handle)".

   **The i18n trap on this item, called out because the file's own shape hides it.** Every entry in the array is written as `t('a11y.shortcuts.<key>', { defaultValue: '...' })`, so copying the neighbouring pattern produces a row that renders correctly in the browser and then fails CI. `npm run test:i18n` enforces locale parity, and a `defaultValue` does not satisfy it: each new key has to be added to all four of `frontend/src/i18n/locales/{en,es,fr,de}/builder.json` under `a11y.shortcuts`. The existing thirteen keys are all present in all four files, so the parity is currently intact and a single missing translation will break it. Two new rows means eight new JSON entries.

2. **Extend the axe e2e suite** to AnalysisPanel and the layer-editor tabs — existing coverage asserts none of the live-region behaviour added by #784/#782/#804. `e2e/accessibility.spec.ts` currently covers search, login, public saved-map, dataset detail, collection detail, maps listing, the map builder page, the Add Dataset dialog, admin pages, and register; neither AnalysisPanel nor the layer-editor tabs appear.

   ~~**Sequence this after #921.**~~ — **Unblocked.** #921 shipped on 2026-07-30 via PR #984 (Advanced JSON errors announced, style-editor selects named), so axe assertions over the style tab will no longer arrive red. This item can be picked up now.

   **Pattern to copy: the Add Dataset dialog test at `e2e/accessibility.spec.ts:264`.** It is the only existing case in the file that scans a surface reached by interaction rather than by URL, and its three-step shape is what these tests need: navigate to `/maps/${builderMapId}`, wait for `getByTestId('builder-sidebar')`, drive the interaction that opens the surface, then run `AxeBuilder` scoped with `.include(...)` to the opened region.

   **The existing fixture map already works; do not add one.** `builderMapId` is created in `beforeAll` (`:73-99`) and gets one layer from `seedDataset('A11y Seed Dataset')`, which ingests `e2e/fixtures/sample.geojson`. Because that layer is vector, `LayerEditorPanel`'s `availableTabs` (`:202-210`) resolves to all four tabs (Style, Filter, Labels, Popup), so a single layer exposes the whole surface under test.

   The interactions to reach each surface:

   - **AnalysisPanel:** `page.getByRole('button', { name: 'Analysis', exact: true }).click()`, then wait for `page.getByTestId('analysis-panel')` (`AnalysisPanel.tsx:880`). This is the same opener `e2e/analysis.spec.ts:67-68` uses.
   - **Layer editor:** `page.locator('[id^="stack-row-"]:not([id="stack-row-basemap-group"])').first().click()`, then wait for `page.getByTestId('builder-layer-editor')` (`MapBuilderPage.tsx:1956`). This is the opener `e2e/builder-styling.spec.ts:94-98` uses.
   - **Each tab:** inside the editor, the strip is `role="tablist"` with `aria-label` "Layer editor tabs" (`LayerEditorPanel.tsx:428-429`) and each tab is a `role="tab"` button with id `tab-${layer.id}-${tab}` (`:439-444`). Click through Style, Filter, Labels, and Popup, scanning after each, since each renders a different panel body and only the active one is in the DOM.

   Note the builder page test at `:246` already scans the builder with the editor closed, so scoping the new scans with `.include('[data-testid="builder-layer-editor"]')` and `.include('[data-testid="analysis-panel"]')` keeps them from re-reporting violations that test already owns.

ishiland:
Closes #806 (both items; they are unrelated to each other and simply share the issue).

## Item 1 — two undiscoverable bindings

`StackRow` maps Space on a focused row to `onCmdClick`, and double-click on the layer name to rename. Neither appeared in `KeyboardShortcutsSheet`. The keyboard one matters most: the sheet documented multi-select as ⌘/Ctrl+Click only, so the sole route available to a keyboard-only user was the one that was not written down.

The new Space row sits next to the pointer chord it mirrors and is scoped "(focused row)", which is what distinguishes it from the existing `Space / Enter` row scoped to the focused drag handle. Both new rows carry real values in en/es/fr/de — a `defaultValue` alone does not satisfy `npm run test:i18n`, which is the trap the file's own shape hides, since every existing row is written that way and renders fine in the browser before failing CI.

## Item 2 — axe coverage for the two interaction-reached surfaces

The existing builder-page scan runs with the editor closed, so neither `AnalysisPanel` nor any layer-editor tab was ever covered — including the live-region behaviour #784/#782/#804 added. Five new cases, following the Add Dataset dialog's three-step shape (navigate → wait for `builder-sidebar` → drive the opener → scan scoped with `.include()`):

- analysis panel, opened with the same opener `analysis.spec.ts` uses
- the Style, Filter, Labels and Popup tabs, opened with the same stack-row selector `builder-styling.spec.ts` uses

Scoped with `.include()` so they do not re-report violations the builder-page test already owns. No new fixture: the `beforeAll` map's seeded layer is vector, so `LayerEditorPanel` resolves all four tabs from that one layer. Each tab renders its own panel body and only the active one is in the DOM, so each gets its own scan.

## Verification

```
# 366 files, 4285 tests pass

npx playwright test e2e/accessibility.spec.ts --project=chromium -g "analysis panel|layer editor"
# 7 passed (5 new cases + setup/teardown), zero violations
```

The e2e run is valid from a worktree for these cases: the dev stack serves `main`'s app code, and the surfaces under scan are byte-identical to `main` on this branch (the diff touches the shortcuts sheet and locales, neither of which is inside the scanned regions).

### DIFF
--- a/e2e/accessibility.spec.ts
+++ b/e2e/accessibility.spec.ts
@@ -280,6 +280,55 @@ test.describe('Accessibility - WCAG 2AA', () => {
     expect(results.violations, formatViolations(results.violations)).toEqual([]);
   });
 
+  // fix(#806 item 2): the builder-page test above scans the builder with the editor
+  // closed, so neither the analysis panel nor any layer-editor tab was ever covered
+  // — including the live-region work #784/#782/#804 added. Both surfaces are reached
+  // by interaction rather than by URL, so they follow the Add Dataset dialog's shape
+  // and scope with .include() to keep them off what the page test already owns.
+  test('analysis panel has no accessibility violations', async ({ page }) => {
+    await page.goto(`/maps/${builderMapId}`);
+    await page.waitForLoadState('networkidle');
+    await expect(page.getByTestId('builder-sidebar')).toBeVisible({ timeout: 15_000 });
+
+    await page.getByRole('button', { name: 'Analysis', exact: true }).click();
+    await expect(page.getByTestId('analysis-panel')).toBeVisible({ timeout: 15_000 });
+
+    const results = await new AxeBuilder({ page })
+      .withTags(wcagTags)
+      .include('[data-testid="analysis-panel"]')
+      .analyze();
+
+    expect(results.violations, formatViolations(results.violations)).toEqual([]);
+  });
+
+  // The seeded layer is vector, so LayerEditorPanel resolves all four tabs and one
+  // layer exposes the whole surface. Each tab renders its own panel body and only
+  // the active one is in the DOM, so every tab needs its own scan.
+  for (const tab of ['Style', 'Filter', 'Labels', 'Popup'] as const) {
+    test(`layer editor ${tab} tab has no accessibility violations`, async ({ page }) => {
+      await page.goto(`/maps/${builderMapId}`);
+      await page.waitForLoadState('networkidle');
+      await expect(page.getByTestId('builder-sidebar')).toBeVisible({ timeout: 15_000 });
+
+      await page
+        .locator('[id^="stack-row-"]:not([id="stack-row-basemap-group"])')
+        .first()
+        .click();
+      const editor = page.getByTestId('builder-layer-editor');
+      await expect(editor).toBeVisible({ timeout: 15_000 });
+
+      await editor.getByRole('tab', { name: tab }).click();
+      await expect(editor.getByRole('tab', { name: tab })).toHaveAttribute('aria-selected', 'true');
+
+      const results = await new AxeBuilder({ page })
+        .withTags(wcagTags)
+        .include('[data-testid="builder-layer-editor"]')
+        .analyze();
+
+      expect(results.violations, formatViolations(results.violations)).toEqual([]);
+    });
+  }
+
   test('admin overview page has no accessibility violations', async ({ page }) => {
     await page.goto('/admin');
     await expect(
--- a/e2e/accessibility.spec.ts
+++ b/e2e/accessibility.spec.ts
@@ -293,9 +293,13 @@ test.describe('Accessibility - WCAG 2AA', () => {
     await page.getByRole('button', { name: 'Analysis', exact: true }).click();
     await expect(page.getByTestId('analysis-panel')).toBeVisible({ timeout: 15_000 });
 
+    // Scope to the rail panel, not the inner form: BuilderRail renders the panel
+    // title and close control as siblings of AnalysisPanel, and since the
+    // builder-page scan runs with the panel closed, that chrome would otherwise
+    // be covered nowhere.
     const results = await new AxeBuilder({ page })
       .withTags(wcagTags)
-      .include('[data-testid="analysis-panel"]')
+      .include('[data-rail-panel]')
       .analyze();
 
     expect(results.violations, formatViolations(results.violations)).toEqual([]);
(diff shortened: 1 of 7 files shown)


ishiland:
@codex review

ishiland:
Both fixed.

**Sheet height.** Confirmed: the shared `DialogContent` sets `fixed top-[50%] translate-y-[-50%]` with no `max-h` and no overflow, so a sheet taller than the viewport clips at both ends and there is nothing to scroll. Capped with `max-h-[85vh] overflow-y-auto` on this dialog rather than on the primitive — every other dialog in the app would inherit a change there, and this PR is not the place to make that call.

**Axe scope.** Also right: the title and close control are siblings of `AnalysisPanel` inside `[data-rail-panel]`, and since the builder-page scan runs with the panel closed, that chrome was covered nowhere. Scoped to `[data-rail-panel]`; it still cannot double-report with the builder-page test for the same reason. Re-ran with the wider scope, still zero violations:

```
npx playwright test e2e/accessibility.spec.ts --project=chromium -g "analysis panel"
# 3 passed (18.7s)
```

Lint, typecheck and the sheet's unit tests re-run.

ishiland:
@codex review

ishiland:
Right — the handler is on the layer-name span, so double-clicking the type icon or empty row space selects rather than renames, and a bare "Double-click" label promises the whole row. The row now reads "Rename layer (double-click its name)" in all four locales, and the test asserts the label names its target rather than just carrying the chord.

I kept the gesture where it is rather than widening it to the row: a row-wide double-click would fight the existing single-click selection and the Space multi-select this same PR documents.

Lint, typecheck, i18n parity and the sheet's unit tests re-run.

ishiland:
@codex review

PR #1048, opened 2026-07-31
perf(analysis): scoped work_mem for the materialize CTAS, with a justified ceiling

### LINKED ISSUE #1012: perf(analysis): scoped work_mem for the materialize CTAS, with a justified ceiling
Split out of #696 (item 2), which was a cluster of four unrelated resource-governance decisions. This one stands alone.

## The gap

The materialize CTAS runs with the cluster-wide `work_mem = 8MB` (`db/postgresql.conf:93`). That is a sane default for eighty concurrent connections doing ordinary catalog queries, and far too small for a single buffer or dissolve over hundreds of thousands of geometries. The statement spills to `pgsql_tmp` long before it needs to.

Because the CTAS is one statement on one session, a `SET LOCAL work_mem` inside its transaction is safe in a way a global bump is not: it applies to that statement and reverts with the transaction, so the other seventy-nine connections keep the 8MB default.

Nothing sets `work_mem` per statement today. The materialize transaction sets only `statement_timeout` (`backend/app/processing/analysis/tasks.py:697`, pre-CTAS) plus `enable_hashagg = off` for dissolve (`:706`), and registration gets its own `statement_timeout` after the commit (`:763`). The new `SET LOCAL` goes alongside the pre-CTAS one at `:697`.

The dissolve case is the one that most wants this. `enable_hashagg = off` (added by #694, because hash aggregation holds every group's union state in memory at once and could OOM the shared db container) forces sorted aggregation instead, and sorting is precisely what `work_mem` governs. Turning off hashagg without raising `work_mem` trades an OOM risk for a guaranteed spill.

## Which memory budget the ceiling has to fit

This is the part the parent issue got vague about, and it is the whole design question.

`work_mem` is allocated by the PostgreSQL backend process serving the session, so the memory lands in the **`db` service**, bounded by `DB_MEM_LIMIT` (default `2g`: `docker-compose.yml:184`, `docker-compose.prod.yml:181`). It shares that 2 GB with `shared_buffers = 512MB` (`db/postgresql.conf:92`), `maintenance_work_mem = 128MB` (`:103`), and up to `max_connections = 80` (`:11`) other backends each entitled to their own `work_mem`.

Two things follow. First, the api service is not involved at all: the job runs in the **`worker`** service (`WORKER_MEM_LIMIT`, default `4g`: `docker-compose.yml:399`, `docker-compose.prod.yml:324`), which is where the Python side of materialize lives, but the worker's budget is not what constrains the ceiling either. Second, `work_mem` is a per-operation limit, not per-statement: a PostGIS query plan with several sorts can allocate a multiple of it, which the conf comment at `:91` already warns about. So the ceiling has to be chosen against 2 GB minus `shared_buffers`, divided by however many sorts a plan can carry, and then multiplied by how many materializes can run at once.

Which is the second half of the problem: **concurrent materializes must not be able to stack past the ceiling.** The cap today is per-user (`backend/app/modules/catalog/datasets/api/router_analysis.py:292-320`), so a tenant with N users can hold N concurrent CTASes, each with its own bumped `work_mem`. Either the ceiling is divided by a bounded concurrency, or a global bound has to exist first. The per-tenant cap is tracked as its own split from #696.

## Acceptance

- [ ] A `SET LOCAL work_mem` is issued in the materialize transaction alongside the existing `SET LOCAL statement_timeout` at `tasks.py:697`.
- [ ] The value is justified in the code comment against the `db` service budget, naming `DB_MEM_LIMIT`, `shared_buffers`, and `max_connections`, and stating the assumed concurrent-materialize count it was divided by.
- [ ] Concurrent materializes cannot stack past that assumed count, or the issue explains what does bound them.
- [ ] The dissolve path in particular is checked: with `enable_hashagg = off` still in force, confirm the raised `work_mem` actually reduces spill rather than being consumed by an unrelated plan node.

ishiland:
The materialize CTAS ran with the cluster-wide `work_mem = 8MB` (`db/postgresql.conf:93`). That is right for eighty concurrent connections doing ordinary catalog queries and far too small for a single buffer or dissolve over hundreds of thousands of geometries, which spills to `pgsql_tmp` long before it needs to.

Dissolve is the case that most wants this. #694 turned off hash aggregation there (it holds every group's union state in memory at once and could OOM the shared db container), and the sorted aggregation that replaces it is precisely what `work_mem` governs — so turning off hashagg without raising `work_mem` traded an OOM risk for a guaranteed spill.

## The change

One `SET LOCAL work_mem` alongside the existing `SET LOCAL statement_timeout` in the materialize transaction. `SET LOCAL` is what makes this safe in a way a global bump is not: it applies to that statement and reverts with the transaction, so the other seventy-nine connections keep the 8MB default and the pooled connection carries nothing onward.

## The ceiling, and which budget it comes out of

**128MB per worker process**, divided across `WORKER_CONCURRENCY` slots, floored at the 8MB cluster default.

`work_mem` is allocated by the PostgreSQL backend serving the session, so the memory lands in the **`db` service**, capped by `DB_MEM_LIMIT` (default `2g`). That 2 GB also holds `shared_buffers` (512MB), `maintenance_work_mem` (128MB), and up to `max_connections = 80` other backends each entitled to their own `work_mem`. The worker's `WORKER_MEM_LIMIT` is not what constrains it — the Python side of materialize lives there, the sort does not.

`work_mem` is per **operation** and per **backend**, so one plan allocates a multiple of it. Two multipliers apply: these shapes carry at most a couple of memory-hungry nodes (dissolve is one Sort feeding a GroupAggregate; buffer and centroid have none; clip has a join), and `max_parallel_workers_per_gather = 1` means a parallel plan runs the leader plus one worker, each with its own allocation. Worst case is 128MB x 2 nodes x 2 backends = **512MB**, leaving roughly 900MB of the 2 GB for the remaining backends.

## What bounds the concurrent count

Not the admission cap — that is per user (`router_analysis.py:292-320`) and gates entry to the **queue**, not execution. Execution is bounded by Procrastinate: a worker process runs at most `WORKER_CONCURRENCY` jobs at once (default 1, `config.py:203`). Dividing the budget by that setting is what keeps concurrent materializes inside one worker from stacking past the ceiling, so raising `WORKER_CONCURRENCY` cannot silently multiply the db container's exposure.

What this does **not** bound is a deployment running a second worker service. That is #695's open question, and until it is answered the ceiling assumes one worker service. Said plainly in the code comment rather than left implied.

## The dissolve check, measured

Against the test cluster, 60,000 buffered polygons in 400 groups, `enable_hashagg = off` in force — the acceptance criterion was to confirm the raised budget actually reduces spill rather than being consumed by an unrelated plan node.

```
work_mem 8MB     Sort Method: external merge  Disk: 70672kB
                 Worker 0: external merge  Disk: 23928kB
                 temp read=11825 written=11837
                 Execution Time: 4371 ms

work_mem 128MB   Sort Method: quicksort  Memory: 96224kB
                 Worker 0: quicksort  Memory: 25kB
                 no temp files
                 Execution Time: 4656 ms
```

The budget lands on the Sort the dissolve depends on. Wall clock did not improve on this fixture — it is dominated by `ST_Union` geometry work, and the temp files stayed in page cache on the test host. The eliminated spill is what matters on a real deployment, where `pgsql_tmp` shares the pgdata volume and `temp_file_limit` is 4GB (#959).

The 96MB the leader actually used is also a useful sanity check on the number: 128MB accommodates this shape without being generous.

## Tests

- `test_ctas_scopes_work_mem_to_its_transaction` — spies the executed statements (the pattern `test_dissolve_ctas_disables_hashagg` already uses): exactly one `work_mem` statement, it is `SET LOCAL` and not a plain `SET`, and it precedes the CTAS in the same transaction as the hashagg flip whose sort it pays for.
- `test_work_mem_is_divided_across_worker_slots` — 1 slot gives 128MB, 4 gives 32MB, 64 floors at 8MB.

## Verification

```
cd backend && uv run pytest tests/test_analysis_materialize.py -q          # 81 passed
cd backend && uv run ruff check . && uv run ruff format --check .          # clean
```

Closes #1012

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -306,6 +306,13 @@ FRONTEND_PORT=8080
 # independently. RAISE it only with headroom you have checked; the win is less
 # spill to pgsql_tmp on large dissolves.
 #
+# The budget is divided across WORKER_CONCURRENCY slots in kB, so it is never
+# exceeded by rounding; if the share would fall below PostgreSQL's 64kB minimum
+# the override is skipped rather than overshooting. The same transaction also
+# pins max_parallel_workers_per_gather to 1, because work_mem is granted to the
+# leader AND each parallel worker — without that pin a server configured for
+# more parallelism would multiply the ceiling.
+#
 # 0 disables the override entirely: no SET LOCAL is issued and analysis runs on
 # whatever work_mem your cluster is configured with. Use that rather than a very
 # small value if you have tuned work_mem yourself — GeoLens cannot read your
--- a/.env.example
+++ b/.env.example
@@ -304,9 +304,13 @@ FRONTEND_PORT=8080
 # LOWER it if you reduce DB_MEM_LIMIT or run more than one worker service — a
 # per-process divisor cannot bound replicas, so each one claims this budget
 # independently. RAISE it only with headroom you have checked; the win is less
-# spill to pgsql_tmp on large dissolves. A value at or below the cluster default
-# (8MB) leaves analysis on the cluster-wide work_mem.
-# Type: integer MB (minimum 1) | Default: 64
+# spill to pgsql_tmp on large dissolves.
+#
+# 0 disables the override entirely: no SET LOCAL is issued and analysis runs on
+# whatever work_mem your cluster is configured with. Use that rather than a very
+# small value if you have tuned work_mem yourself — GeoLens cannot read your
+# cluster's setting, so it cannot tell a deliberate low value from an accident.
+# Type: integer MB (0 disables) | Default: 64
 # ANALYSIS_MATERIALIZE_WORK_MEM_MB=64
 #
 # FastAPI memory cap.
--- a/.env.example
+++ b/.env.example
@@ -294,6 +294,21 @@ FRONTEND_PORT=8080
 # Type: Docker byte value | Default: 2g
 # DB_MEM_LIMIT=2g
 #
+# Per-slot work_mem for the analysis materialize CTAS, in MB. Sized against
+# DB_MEM_LIMIT above: the memory is allocated by the PostgreSQL backend, not by
+# the worker, and work_mem is per operation AND per backend, so one materialize
+# can allocate this value x2 plan nodes x2 backends. At the 64MB default that is
+# 256MB per worker replica, which sits inside a 2g database alongside
+# shared_buffers (512MB) and maintenance_work_mem (128MB).
+#
+# LOWER it if you reduce DB_MEM_LIMIT or run more than one worker service — a
+# per-process divisor cannot bound replicas, so each one claims this budget
+# independently. RAISE it only with headroom you have checked; the win is less
+# spill to pgsql_tmp on large dissolves. A value at or below the cluster default
+# (8MB) leaves analysis on the cluster-wide work_mem.
+# Type: integer MB (minimum 1) | Default: 64
+# ANALYSIS_MATERIALIZE_WORK_MEM_MB=64
+#
 # FastAPI memory cap.
 # Type: Docker byte value | Default: 2g
 # API_MEM_LIMIT=2g
--- a/.env.example
+++ b/.env.example
@@ -307,8 +307,11 @@ FRONTEND_PORT=8080
 # spill to pgsql_tmp on large dissolves.
 #
 # The budget is divided across WORKER_CONCURRENCY slots in kB, so it is never
-# exceeded by rounding; if the share would fall below PostgreSQL's 64kB minimum
-# the override is skipped rather than overshooting. The same transaction also
+# exceeded by rounding. If the share would fall below PostgreSQL's 64kB minimum
+# for work_mem, the application REFUSES TO START and names the arithmetic:
+# there is no honest outcome at that point, since the minimum overshoots the
+# budget and falling back to the cluster's value overshoots it further. The
+# same transaction also
 # pins max_parallel_workers_per_gather to 1, because work_mem is granted to the
 # leader AND each parallel worker — without that pin a server configured for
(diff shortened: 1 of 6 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
Rebased onto `main` (now at `c081fc6b1`). Three conflicts, all of them additions landing at the same point rather than competing edits, so both sides are kept: the #1001 geodesic-buffer block beside this branch's settings field in `config.py`, and the #1013 timeout tests beside this branch's work_mem tests in `test_analysis_materialize.py` and `test_config.py`. The one real interaction is in `test_config.py`, where #1013 added timeout entries to the same rejection table this branch removes the `work_mem = 0` entry from — 0 is a documented opt-out here, so the removal is applied on top of #1013's additions.

One commit is new. `backend/app/processing/analysis/tasks.py` crossed the 1000-line inclusion threshold **on the rebase**, in neither branch: main grew the module while this was open, and this branch's additions landed on top. It is gated at 1031 now, with the reason recorded in its own commit so the date is honest.

Verified after the rebase, whole-tree rather than on the touched files: `ruff check`, `ruff format --check`, `test_layering.py`. The DB-backed suites run in CI.

ishiland:
@codex review

PR #1049, opened 2026-07-31
refactor(geo): one implementation of the degree-unit test, and a gate on the rest

### LINKED ISSUE #961: chore(backend): four divergent checks for whether a CRS is degree-based — consolidate on one helper

ishiland:
#961 re-scoped to two things: a judgment call, and a standing sync obligation. Both settled here — and turning the second into a test found a live disagreement, so this PR carries one behaviour change. See the last section.

## Item 1 — `_is_degree_based` converges

`core/geo.py` and `processing/raster/vrt.py` each carried the same radians-per-unit comparison, against its own copy of `math.pi / 180.0`, at the same `rel_tol`, kept in step by cross-referencing comments. The issue's words: either answer is defensible, but leaving it undecided with two copies of the same constant comparison is not.

**Decided toward convergence**, option 1. `core.geo.crs_has_degree_unit(crs)` takes a **parsed CRS object** — that is what makes the merge possible without the WKT round trip the issue warns against, since `vrt.py` already holds a live `rasterio.crs.CRS`:

- `wkt_has_degree_unit(wkt)` = `crs_has_degree_unit(_parse_crs(wkt))`
- `_is_degree_based(crs)` = the `is_geographic` precondition + `crs_has_degree_unit(crs) is True`

**What deliberately does not converge, and is now written at both sites:** the reading of "unknown". `crs_has_degree_unit` returns `None` when PROJ cannot report a factor. `vrt.py` resolves that to False, because a CRS it cannot classify must never be shifted by 360°. `tiles/router.py` tests `is not False`, so an unparseable CRS keeps the historical degrees assumption instead of collapsing its resolution (the #939 bug: ETOPO at EPSG:9518 derived z22 instead of z7). Same question, opposite safe answer. A single shared helper makes unifying those two readings tempting, so a test pins them apart.

## Item 2 — the SQL predicates become a gate, not a comment

The two `ingest/metadata.py` sites can never merge into the Python helpers: they run inside a query against `spatial_ref_sys.srtext`, where there is no Python-side CRS object to hand to PROJ. A comment saying "keep this in sync" is exactly the arrangement that let the constant get duplicated in the first place, so it is a test now.

`backend/tests/test_crs_degree_agreement.py` runs one CRS through **all five sites** and asserts they agree:

1. `wkt_is_geographic` / `wkt_has_degree_unit` over the stored WKT (what `tiles/router.py` reads)
2. `crs_has_degree_unit` over the parsed CRS (the new shared entry point)
3. `_is_degree_based` over the live CRS object (`vrt.py`)
4. `_GEOGRAPHIC_SRTEXT_RE` + `_PROJECTED_SRTEXT_RE` + `_DEGREE_UNIT_SRTEXT_RE`, run as SQL — the gate on the 0..360 shift
5. the #906 degenerate-envelope floor's predicate, whose own comment claims it is "the same keyword logic as `core.geo.wkt_is_geographic`, in srtext form" — held to that claim

## The one behaviour change: the 0..360 gate was asking the wrong question

Writing the test above found that the five sites **already disagreed**, which is what the obligation was supposed to prevent.

`_GEOGRAPHIC_SRTEXT_RE` was anchored — `^GEOG(CS|CRS)`. A stock `spatial_ref_sys` carries **277 `COMPD_CS` rows** plus a family of BOUNDCRS-wrapped geographic CRSs, and their srtext starts with the *wrapper*. For those the gate said "not lon/lat" while `wkt_is_geographic`, `crs_has_degree_unit`, `_is_degree_based` and the degenerate-envelope floor all said geographic. Effect: a 0..360 Pacific source in such a CRS was silently not shifted, and then clipped — the #888 data-loss path, through the door #888 was built to close.

The floor 150 lines below already knew. Its #906 comment spells out that the test must see through `COMPD_CS` and that the right form is "a GEOG keyword present and no PROJ keyword anywhere". That insight was applied at one site and not the other.

Both sites now interpolate the same two constants, so the next change to either cannot leave them disagreeing. `3857` and `2263` stay excluded by the PROJ half, which is what the anchor was standing in for.

Inherited and unchanged: unlike the Python helper, SQL cannot blank quoted content first, so a CRS whose *name* contained "PROJCS" would read as projected. No such row exists in a stock `spatial_ref_sys`; recorded at the constant.

## The matrix

| SRID | | geographic | degrees |
|---|---|---|---|
| 4326 | WGS 84 | yes | yes |
| 4269 | NAD83 | yes | yes |
| **4807** | **NTF (Paris)** | **yes** | **no — grads** |
| **4901** | **ATF (Paris)** | **yes** | **no — grads** |
| **3823** | **TWD97 — BOUNDCRS over a degree GEOGCRS** | **yes** | **yes** |
| **4339** | **Australian Antarctic — same wrapper** | **yes** | **yes** |
| 5698 | RGF93 / Lambert-93 + height — COMPD_CS over PROJCS | no | no |
| 3857 | Web Mercator | no | no |
| 2263 | NY Long Island ftUS | no | no |

The grads rows are #961's stated acceptance criterion ("a grads GEOGCS classifies the same way at every remaining site"). The wrapped rows are the ones codex's review made me add, and they are the reason a matrix of plain `GEOGCS`/`PROJCS` was not enough: it passed while five sites disagreed.

srtext is read from the database rather than from fixtures, on purpose — the SQL half is only meaningful against the strings PostGIS actually ships. A second test asserts the grads family is really present (floor of 10; #899 counted 14), because zero would mean the file proves nothing.

## Not reopened

The acceptance criterion #939 knowingly violated — "no new per-request PROJ lookups on tile-serving paths" — stays violated, as the issue instructs; the trade and its `lru_cache` mitigation are recorded there. The regex-versus-parser question is likewise settled and untouched.

No schema or API change. One ingest behaviour change, described above: a 0..360 source in a wrapped geographic CRS is now shifted rather than left to be clipped. That is the #888 guard doing what it was written to do; the four-condition structure of that guard is unchanged, only the CRS-class test inside condition 1.

## Verification

```
cd backend && uv run pytest tests/test_crs_degree_agreement.py -q                    # 11 passed
cd backend && uv run pytest tests/test_crs_degree_agreement.py tests/test_ingest_mercator_clip.py \
  tests/test_wkt_is_geographic.py tests/test_raster_antimeridian_887.py \
  tests/test_layering.py tests/test_raster_tiles.py -q                               # 264 passed
```

Negative control, reverting **only** the regex and keeping the matrix: fails exactly `[3823-True-True]` and `[4339-True-True]`, and nothing else.

Closes #961

### DIFF
--- a/backend/app/core/geo.py
+++ b/backend/app/core/geo.py
@@ -503,9 +503,10 @@ def make_bbox_filter(
         return and_(geom_col.op("&&")(envelope), spatial_fn(geom_col, envelope))
 
 
-# fix(#939): the same radians-per-degree constant `_is_degree_based`
-# (processing/raster/vrt.py) compares against. Unit semantics live in the
-# conversion factor, never the unit's name.
+# Unit semantics live in the conversion factor, never the unit's name.
+# fix(#961): this constant had a twin in processing/raster/vrt.py, compared the
+# same way against the same tolerance. Both sites now go through
+# `crs_has_degree_unit` below, so there is one implementation to keep right.
 _RADIANS_PER_DEGREE = math.pi / 180.0
 
 
@@ -598,6 +599,37 @@ def wkt_is_geographic(crs_wkt: str | None) -> bool | None:
     return None
 
 
+def crs_has_degree_unit(crs: object | None) -> bool | None:
+    """Whether a PARSED CRS's coordinate axes are measured in degrees.
+
+    fix(#961): the one implementation of the radians-per-unit test.
+    :func:`wkt_has_degree_unit` is this plus the WKT parse; ``_is_degree_based``
+    (``processing/raster/vrt.py``) is this plus an ``is_geographic``
+    precondition. They used to be two copies of the same comparison against two
+    copies of the same constant, kept in step by cross-referencing comments,
+    which is the arrangement #961 was filed to end. The CRS-object entry point
+    is what makes one implementation possible: ``vrt.py`` holds a live
+    ``rasterio.crs.CRS`` and must not be pushed through a WKT round trip to
+    reach a shared helper.
+
+    ``rel_tol`` is correct here: this compares two fixed physical constants of
+    the same tiny magnitude (0.01745 radians per degree against whatever PROJ
+    reports), where proportional agreement is the meaningful test and the
+    nearest wrong answer, grads at 0.01571, is 10% away.
+
+    Returns None when there is no CRS or PROJ cannot report a unit factor.
+    Callers must read that as "unknown" and decide for themselves — the two
+    call sites deliberately differ, see each one.
+    """
+    if crs is None:
+        return None
+    try:
+        _, radians_per_unit = crs.units_factor
+    except Exception:  # broad: units_factor raises CRSError on exotic/!undefined CRSs, which are exactly the ones we cannot answer for
+        return None
+    return math.isclose(radians_per_unit, _RADIANS_PER_DEGREE, rel_tol=1e-9)
+
+
 def wkt_has_degree_unit(crs_wkt: str | None) -> bool | None:
     """Whether a CRS WKT's coordinate axes are measured in degrees.
 
@@ -605,11 +637,11 @@ def wkt_has_degree_unit(crs_wkt: str | None) -> bool | None:
     and admits grads CRSs. Only meaningful for a WKT already classified as
     geographic, so call that first.
 
-    The answer comes from PROJ's ``units_factor`` on the parsed CRS -- the
-    same radians-per-unit comparison ``_is_degree_based``
-    (processing/raster/vrt.py) makes, against the same constant, so the two
-    unit tests in this codebase cannot drift apart. Reading the factor rather
-    than the unit's name means a valid custom spelling
+    The answer comes from :func:`crs_has_degree_unit`, which is also what
+    ``_is_degree_based`` (processing/raster/vrt.py) calls, so the two unit
+    tests in this codebase are one implementation rather than two that must be
+    kept in step (fix(#961)). Reading PROJ's ``units_factor`` rather than the
+    unit's name means a valid custom spelling
     (``UNIT["arc-degree",0.01745...]``) still reads as degrees, while grads
     sit 10% away and stay excluded.
 
@@ -621,11 +653,4 @@ def wkt_has_degree_unit(crs_wkt: str | None) -> bool | None:
     """
     if not isinstance(crs_wkt, str) or not crs_wkt:
         return None
-    crs = _parse_crs(crs_wkt)
-    if crs is None:
-        return None
-    try:
-        _, radians_per_unit = crs.units_factor
-    except Exception:  # broad: units_factor raises CRSError on exotic/!undefined CRSs, which are exactly the ones we cannot answer for
-        return None
-    return math.isclose(radians_per_unit, _RADIANS_PER_DEGREE, rel_tol=1e-9)
+    return crs_has_degree_unit(_parse_crs(crs_wkt))
--- a/backend/app/processing/ingest/metadata.py
+++ b/backend/app/processing/ingest/metadata.py
@@ -1245,22 +1245,28 @@ _MERCATOR_SAFE_ENVELOPE = "ST_MakeEnvelope(-180, -85.06, 180, 85.06, 4326)"
 # ships in spatial_ref_sys.srtext for lon/lat CRSs. 4326/4979/4269 match;
 # 2263/3857 (projected, X in feet/metres) do not.
 #
-# fix(#961 review): this used to be anchored (`^GEOG(CS|CRS)`), which is not
-# the same question `core.geo.wkt_is_geographic` answers and disagreed with it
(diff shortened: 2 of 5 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1050, opened 2026-07-31
feat(cli): add analysis preview and materialize commands

### LINKED ISSUE #685: feat(cli): analysis commands
The CLI has no analysis commands; scripting users must drop to the generated SDKs. Low priority. Scope when someone asks:

- `geolens analysis preview <dataset-id> --operation buffer --distance 500` (GeoJSON to stdout)
- `geolens analysis materialize <dataset-id> --operation ... --title ... [--wait]` (polls the job like other CLI job commands)

Both endpoints are already in the generated SDKs (`/datasets/{id}/analysis/preview/` and `/analysis/materialize/`).

## Ready now, against the four shipped operations

This can be built today against `buffer`, `centroid`, `clip`, and `dissolve`. It does not wait on the phase-2 operations (#953, #954, #955, #956).

One coupling to know about: the generated Python SDK **validates the operation enum**, so the CLI inherits a new operation only after an SDK regeneration. `sdks/python/geolens/models/analysis_materialize_request_operation.py:3` pins `Literal["buffer", "centroid", "clip", "dissolve"]`, and `check_analysis_materialize_request_operation` at `:15-22` raises `TypeError` on anything else; the preview equivalent is `analysis_preview_request_operation.py:3`. Passing `--operation spatial_join` before `make sdks` has been re-run against the new `backend/openapi.json` fails inside the SDK, not at the server. Do not hand-edit those files to get ahead of it; they are generated.

So: pass the operation string through to the SDK unchanged and let the SDK be the enum authority, rather than duplicating the list in the CLI. Each phase-2 operation then reaches the CLI for free when its own PR regenerates the SDK.

ishiland:
Scripting users had to drop to the generated SDK to run an analysis operation. Two commands close that, against the four shipped operations.

```
geolens analysis preview <dataset-id> --operation buffer --distance 500 > ring.geojson
geolens analysis materialize <dataset-id> --operation buffer --distance 500 --title "500 m ring"
```

`preview` prints the FeatureCollection on stdout with the same JSON contract `export stac` uses (pretty and sorted by default, `--compact` for piping into jq). A capped preview announces itself on stderr, not in the middle of the document, so the redirect above stays a valid `.geojson` file.

`materialize` queues the job and, with `--wait` (the default), polls it through the same `resolve_dataset_id` helper `geolens publish` uses, printing the new dataset's URL. `--no-wait` returns the job-search URL form. `--json` emits `{job_id, dataset_id, url}`.

Options are the ones the API takes: `--distance` in metres (buffer), `--mask-dataset` (clip), `--by-field` (dissolve), `--title` (materialize).

## The operation list lives in the SDK, not here

`--operation` is a plain string handed to the SDK unchanged. A `Literal` in the CLI would have to be updated for every new backend operation, and in the meantime would reject one the server already supports. The generated enum is the authority, so #953 through #956 each reach the CLI with the `make sdks` in their own PR. A malformed `--mask-dataset` is still rejected locally with exit 2, since that is a client-side parse rather than a server rule.

No API, schema, or security impact. No SDK regeneration: both endpoints were already generated.

Closes #685.

## Verification

```
cd cli && uv run --extra dev python -m pytest -q     # 253 passed (17 new)
uv run --extra dev ruff check geolens_cli/analysis.py tests/test_analysis.py
grep -rE '^(import|from) (httpx|requests)' geolens_cli/   # OCCLI-06: no hits
```

17 new tests cover the request builders (which params are sent, which stay UNSET, and that an unknown operation passes through), the GeoJSON unwrapping and truncation notice, and both command bodies including the `--no-wait` path asserting it never polls.

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -2,7 +2,7 @@
 
 Apache-2.0 command-line interface for the [GeoLens](https://github.com/geolens-io/geolens) API.
 
-Login, scan local directories of spatial data, apply manifest-driven catalogs, publish vector or raster files, and export STAC metadata against any GeoLens instance.
+Login, scan local directories of spatial data, apply manifest-driven catalogs, publish vector or raster files, run PostGIS analysis operations, and export STAC metadata against any GeoLens instance.
 
 See [docs.getgeolens.com](https://docs.getgeolens.com/) for the full command reference.
 
@@ -18,6 +18,8 @@ geolens schema --output geolens-manifest-v1.schema.json
 geolens apply --dry-run geolens.yaml
 geolens apply geolens.yaml
 geolens publish ./data/cities.geojson
+geolens analysis preview <dataset-id> --operation buffer --distance 500 > ring.geojson
+geolens analysis materialize <dataset-id> --operation buffer --distance 500 --title "500 m ring"
 geolens export stac <dataset-id> -o cities.stac.json
 ```
 
--- a/cli/README.md
+++ b/cli/README.md
@@ -19,7 +19,7 @@ geolens apply --dry-run geolens.yaml
 geolens apply geolens.yaml
 geolens publish ./data/cities.geojson
 geolens analysis preview <dataset-id> --operation buffer --distance 500 > ring.geojson
-geolens analysis materialize <dataset-id> --operation buffer --distance 500 --title "500 m ring"
+geolens analysis materialize <dataset-id> --operation buffer --distance 500 --title "500 m ring"  # waits for the job; --timeout to bound it
 geolens export stac <dataset-id> -o cities.stac.json
 ```
 
--- a/cli/geolens_cli/analysis.py
+++ b/cli/geolens_cli/analysis.py
@@ -24,15 +24,19 @@ from ._sdk_helpers import call_sdk, unwrap
 #: unit picker (AnalysisPanel converts feet/miles before it POSTs).
 DISTANCE_UNIT = "metres"
 
-#: fix(#685 review): publish's 120s poll default is far too short here. The
-#: server gives a materialize 300s of processing on its own
-#: (MATERIALIZE_TIMEOUT, backend/app/processing/analysis/tasks.py), and #703
-#: deliberately queues analysis BELOW uploads, so a legitimate run can sit
-#: waiting for minutes before it starts. Ten minutes covers both with room.
-#: A constant rather than a --timeout option on purpose: the flag is worth
-#: adding when someone has a job that legitimately outlives this, and not
-#: before.
-POLL_TIMEOUT_SECONDS: float = 600.0
+#: fix(#685 review): publish's 120s poll default is far too short here, and so
+#: was the 600s that replaced it. The queue is the reason: #703 deliberately
+#: ranks analysis BELOW uploads, so on a busy instance a perfectly healthy job
+#: can sit pending for as long as the upload backlog lasts, and any fixed
+#: wall-clock guess turns that into a false failure for automation.
+#:
+#: So the ceiling is not a guess about how long work takes — it is the point
+#: past which waiting cannot help. The server's own stale-job sweep fails a
+#: job whose lease has expired after an hour (see the sweep notes in
+#: backend/app/processing/analysis/tasks.py), so an hour is the longest a job
+#: can be alive-but-unresolved. Ctrl+C remains the way out of a wait the user
+#: no longer wants.
+POLL_TIMEOUT_SECONDS: float = 3600.0
 
 
 def _mask_dataset_arg(mask_dataset_id: Optional[str]) -> Any:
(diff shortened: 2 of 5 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
All three taken.

**Failed job exited 0** — the sharpest of the three. `resolve_dataset_id` returns `None` for a terminal failure and for a timeout alike, and treating that as success is exactly wrong for the audience these commands exist for. It now exits 1 with the job id, in both human and JSON modes.

**`/datasets?job_id=` route** — dropped from this command. Nothing is polled on the `--no-wait` path, so it reports the job id rather than a URL for a dataset that may not exist. (`geolens publish` still builds that form; whether the frontend should answer it is a separate question about an existing command.)

**SDK floor** — raised to `geolens>=1.5.0`. `geolens.api.datasets_analysis` first ships in 1.5.0 and the old floor admitted 1.2.x, so `geolens analysis` would have been a `ModuleNotFoundError` on a satisfying install.

ishiland:
@codex review

ishiland:
Right, and it interacted badly with the previous round's fix: making an unresolved poll exit 1 is only honest if the wait is long enough to be meaningful. The poll now runs for 600s, sized on the server's own budget (300s of processing per MATERIALIZE_TIMEOUT, plus the queue wait #703 deliberately imposes by ranking analysis below uploads).

Reaching the end of that is still non-zero, since no dataset came back, but it no longer claims the job failed: the status is read back once so "failed" and "still running" get different sentences. Three tests cover the two wordings and that the timeout passed to the poll exceeds the server-side budget.

ishiland:
@codex review

ishiland:
Both taken, and the first one is the better framing of the previous round.

**Fixed wall-clock was the wrong shape.** Any number I pick is a guess about how long work takes, and the queue makes that unanswerable: #703 ranks analysis below uploads, so a healthy job can sit pending for as long as the backlog lasts. The ceiling is now the point past which waiting cannot help rather than a guess about duration: an hour, which is when the server's own stale-job sweep fails a job whose lease has expired. Past that the job is terminal by construction. Ctrl+C is still the way out of a wait the user no longer wants.

**Unreadable status.** Right, a 401/403/404/5xx on the job endpoint collapses to the same `None`, and calling that a timeout asserts something never established. It now has its own branch and says the outcome is unknown. Still exit 1, since no dataset came back either way.

ishiland:
@codex review

ishiland:
You're right, and my premise was wrong: I read the one-hour sweep as a ceiling on how long a job can be alive-but-unresolved, but `no_live_procrastinate_job()` exists precisely to exempt a queued job from it — "simply waiting ... with no upper bound", and failing it at an hour would be "both a lie and a loss".

So the fourth deadline is not a number. `--wait` waits for a terminal state. Three guesses in a row were wrong for the same reason, which is the signal that the default was never the thing to tune: `--timeout` now exists for automation that needs a bound, and reaching it reports the job as unfinished rather than failed. Ctrl+C covers the interactive case.

ishiland:
@codex review

ishiland:
Both fixed.

`--timeout 0` hitting the falsy branch and becoming an unbounded wait is the exact inversion of the request. Zero or negative is now a usage error that names `--no-wait`, which is the flag for not waiting, and the poll takes `POLL_FOREVER if timeout is None else timeout`.

`preview` now runs the same instance guard as `materialize`, so a missing instance exits 3 from both rather than 3 from one and 2 from the other. Two tests added.

ishiland:
@codex review

ishiland:
Both fixed with one `require_finite` helper. `nan`, `inf` and `1e309` all parse into real floats, and JSON cannot spell any of them, so the distance failure would have surfaced from inside the SDK's encoder rather than as a bad argument, and `--timeout inf` turned an explicitly bounded wait into an unbounded one. Four tests: three parametrized over the distance values, one on the timeout.

ishiland:
@codex review

ishiland:
@codex review

ishiland:
Two of three taken.

**Completion found by the final read** — fixed, and it was the worst of the three answers the command could give. `job_status` is now `job_snapshot` and returns the dataset id alongside the status, so a job that finishes between the poll's last look and the re-read reports success instead of "unfinished".

**Auth exit code** — fixed. A 401/403 on that read raises with exit 3 rather than collapsing into the generic 1, so the code still says what went wrong.

**Remaining deadline per request** — declining. Bounding each request by the time left would mean threading a deadline into `resolve_dataset_id`, which `publish` also uses, and the payoff is turning a worst case of roughly 2× `--timeout` into 1×. `--timeout` is a bound on waiting, not a real-time deadline, and the overshoot is one HTTP timeout. If exact deadline arithmetic is wanted it belongs in the shared poll helper, as its own change, for both commands at once.

PR #1051, opened 2026-07-31
ci: make merge_group runs execute the gates before the queue is enabled

ishiland:
Steps 1-4 of #1000. **Step 5 — enabling the queue — is a repository-admin action and is not in this PR.** Nothing here changes behaviour until someone flips it on; `merge_group:` in `on:` is inert while no queue exists.

## The trap this closes

Adding `merge_group:` to `on:` is one line. The conditionals underneath are why that line alone would have been actively dangerous:

- **12 of 25 jobs** were gated on `needs.changes.outputs.X == 'true' || github.event_name == 'push'`. A `merge_group` event matches neither `push` nor `pull_request`, so all twelve skip.
- `ci-ok` gates on `needs.*.result` and **treats `skipped` as pass** — correct for a path-filtered PR, catastrophic here.

Net result: a queue run in which nearly everything skipped, `CI OK` green, and code merged that was never tested. Not a stall anyone would notice.

## What changed

**Every one of `ci-ok`'s 17 dependencies now runs on `merge_group`.** Verified by parsing the file rather than by reading it:

| | on merge_group |
|---|---|
| `backend-lint` `backend-test` `installer-test` `backup-roundtrip` `openapi-snapshot` `sdks-check` `cli-test` `mcp-test` `alembic-clean-db` `frontend-lint` `frontend-test` `security-scan` | `\|\| github.event_name == 'merge_group'` added |
| `e2e-smoke` | runs **unconditionally** |
| `changes` `pre-commit` `version-coherence` `docs-contract` | already unconditional |

A queue run validates a candidate merge commit against `main` the same way a push run validates the merged result, so both want the full matrix rather than a path-filtered subset — which is why the issue's "treat `merge_group` like `push`" is the correct fix rather than a workaround.

`e2e-smoke` is unconditional rather than path-filtered (step 3 of the issue). There is no reliable filter to apply there, and it is the one gate whose purpose is catching what a *combination* breaks: two PRs can each leave the core flows green alone and break one together, which is the failure a merge queue exists to stop.

## `dorny/paths-filter` on a queue branch

It runs with `token: ''`, which switches it to git-diff detection — a deliberate #546 choice, because the API path's transient 503s failed this required check hard, and it should stay. Git-diff detection resolves its base from pull-request context or a push's before-SHA, and a `gh-readonly-queue/*` branch supplies neither.

The step is therefore **skipped on `merge_group`**. `Detect Changes` is a required context, so the job still has to succeed; skipping the step is how it does, and the empty outputs it leaves behind are ignored because every consumer ORs in `merge_group`. The alternative — feeding it `github.event.merge_group.base_sha` — would make its outputs load-bearing in a context #546 never contemplated, to produce a path-filtered subset the queue does not want anyway.

## Concurrency (step 4): checked, unchanged

`group: ${{ github.workflow }}-${{ github.ref }}-${{ github.event_name }}` with `cancel-in-progress: true` is already safe. Each queue entry runs on its own `gh-readonly-queue/main/pr-N-<sha>` ref, so `github.ref` alone puts every entry in a distinct group: two batches in flight cannot cancel each other, and a queue run and the PR it came from differ in both ref and event name. Recorded as a comment so the next reader does not have to re-derive it.

## What deliberately stays off the queue

`runtime-extension-image-probe`, `accessibility`, `stac-validate`, `e2e-test`, `ai-evals`. None is in `ci-ok`, so none would gate anything in the queue — they would only spend Actions minutes, which the existing "pin-sha-probe-push-only" and per-PR-budget comments say the project cares about. The push-triggered ones still run when the batch lands on `main`, the same coverage they have today.

## The invariant is now a test

`backend/tests/test_ci_merge_queue.py` asserts that no `ci-ok` dependency can skip on `merge_group`, and pins `backend-test` / `frontend-test` / `e2e-smoke` by name because #1000's acceptance names them. Add a job to the aggregator without a `merge_group` clause and it fails — instead of quietly reopening the hole after the queue is switched on, when the symptom is a green check on an untested merge.

It does not assert that the queue is enabled; that is not a file in this repo.

## Verification

```
cd backend && uv run pytest tests/test_ci_merge_queue.py tests/test_phase_279_compose_ci.py \
  tests/test_ci_alembic_filter_paths.py -q      # 16 passed
```

Structural audit of the whole file (each `ci-ok` dependency's `if:` against `merge_group`) run in-session and reported clean for all 17.

**Cannot be verified end to end from here.** Acceptance items "a PR merged through the queue shows a `gh-readonly-queue/*` run in which Backend Tests actually ran" and "a deliberately broken PR is ejected" both require the queue to be on, which needs admin rights.

## Handover — what the admin does next

1. Merge this.
2. Settings → Branches → `main`: enable the merge queue with **batch size > 1** — the batching is where the throughput win is.

   Two config notes, both learned while this PR sat in the queue:

   `allow_update_branch` was **false** while `allow_auto_merge` was **true**, so GitHub would merge an armed PR but never offer to bring a stale one up to date. It has since been set to **true** — but measured afterwards, that did **not** stop the stalls: main moved and every armed PR was still `BEHIND`. The setting exposes the update affordance; it does not auto-update branches for auto-merge PRs. Under `strict: true` only one PR can land per main-move regardless, so the branches have to be serialized by hand. That is the whole case for the queue: it updates and tests a batch, instead of one PR per CI cycle.

   Note on `strict`: the issue records it as turned off on 2026-07-30, but branch protection currently reads `strict: true` with required contexts `Detect Changes`, `Pre-commit`, `CI OK` (read back from the API while writing this). So it does not need re-enabling — and its livelock is live right now: this PR's own siblings sat at `BEHIND` with auto-merge armed and did not advance until a manual `update-branch`. The queue supersedes `strict`; whether to leave it on alongside is your call.
3. Land one throwaway PR through the queue and confirm its `gh-readonly-queue/*` run shows Backend Tests and the frontend suites as **ran**, not skipped.

Worth timing for a quiet moment: once the queue is on, `--auto` enqueues rather than merges, merges serialize through the queue, and CI re-runs on a `gh-readonly-queue/*` branch — so it changes how every open PR lands.

Refs #1000

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,23 @@ on:
     branches: [main]
   pull_request:
     branches: [main]
+  # fix(#1000): merge-queue runs. All three required contexts (Detect Changes,
+  # Pre-commit, CI OK) come from this workflow, so the queue cannot validate a
+  # candidate merge commit without it.
+  #
+  # Adding this line is the easy half. The job conditionals below are the
+  # reason the queue must NOT be enabled before they are fixed: 12 of them
+  # were gated on `needs.changes.outputs.X == 'true' || github.event_name ==
+  # 'push'`, and a merge_group event matches neither `push` nor
+  # `pull_request`, so every one of them would skip — while `ci-ok` counts a
+  # skip as a pass (correct for a path-filtered PR, catastrophic here). The
+  # queue would have reported green having tested nothing.
+  #
+  # A queue run validates a candidate merge commit against main the same way a
+  # push run validates the merged result, so both want the full matrix rather
+  # than a path-filtered subset. Every gated job below therefore reads
+  # `|| github.event_name == 'merge_group'` alongside its push clause.
+  merge_group:
   # Nightly full-matrix E2E: the e2e-test job below runs ONLY on `schedule`/
   # `workflow_dispatch` (it stays skipped on push/PR). PRs that can affect the
   # stack run the small e2e-smoke job instead (fix #825), so this cron remains
@@ -17,6 +34,14 @@ concurrency:
   # Include the event name so daily dependabot pushes stop cancelling the
   # scheduled run — every nightly for a week was cancelled this way, which is
   # how the e2e suite silently rotted (v1.6.0 audit N2 family).
+  #
+  # fix(#1000), checked rather than changed: this shape is already safe for the
+  # merge queue. Each queue entry runs on its own `gh-readonly-queue/main/pr-N-<sha>`
+  # ref, so `github.ref` alone puts every entry in a distinct group — two
+  # batches in flight cannot cancel each other, and a queue run and the PR run
+  # it came from differ in both ref and event name. `cancel-in-progress` stays
+  # true because the only thing that can now share a group is a re-run of the
+  # same queue entry.
   group: ${{ github.workflow }}-${{ github.ref }}-${{ github.event_name }}
   cancel-in-progress: true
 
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1052, opened 2026-07-31
fix(maps): give the map thumbnail its own timestamp

### LINKED ISSUE #1005: fix(maps): lazy thumbnail backfill on view bumps updated_at and reorders the gallery sort
## What happens

Opening a map that has no thumbnail generates one, and that write bumps `updated_at`. The maps gallery sorts by "Last updated" by default, so browsing maps silently reorders the gallery.

## Observed

A read-only pass over the seeded showcase maps, opening each and editing nothing:

| map | had a thumbnail | opened | `updated_at` after |
| --- | --- | --- | --- |
| The Matterhorn in 3D | no | 02:00 | 02:00:06 |
| Restless Earth | no | 01:58 | 01:58:07 |
| Manhattan - A Century of Skyline | yes | 01:59 | 11:29:52 (unchanged) |

Both maps that lacked a thumbnail jumped to the top of the gallery. The one that already had a thumbnail did not move, which points at the backfill write rather than the page view itself.

Found during a post-rebuild smoke check of `9a788ba92`.

## The bump is deliberate, and deleting it breaks thumbnail refresh

**Verified 2026-07-31.** `updated_at` is doing double duty as the thumbnail cache version. This is load-bearing, not an oversight, so the obvious one-line fix is a regression.

`MapCard.tsx:28` and `MapCardGrid.tsx:22` both call `useMapThumbnail(map.thumbnail_url, map.updated_at)`. That second argument is the version: `withThumbnailVersion` (`frontend/src/components/maps/hooks/use-map-thumbnail.ts:6-15`) appends it as `?v=`, and it is also part of the TanStack query key at `:40`. A map whose thumbnail is re-captured after an edit shows the new image only because `updated_at` moved. `test_upload_thumbnail_bumps_updated_at` (`backend/tests/test_maps.py:3142-3160`) pins exactly this, with the docstring "Thumbnail refreshes must invalidate map-card thumbnail caches."

The reason both behaviours ride on one column is that one endpoint serves both flows. `captureThumbnail` in `frontend/src/components/builder/hooks/use-builder-save.ts` is called from two places: the post-save re-capture at `:788-793`, where bumping the version is correct, and `maybeAutoCaptureThumbnail` at `:1020-1042`, the first-load backfill, where it is not. The backend cannot tell them apart.

Note that a capture writes **two** images. `captureThumbnail` uploads a 400x250 thumbnail and a 1200x630 OG card, and both endpoints bump `updated_at`: `upload_thumbnail` at `backend/app/modules/catalog/maps/router.py:968` and `upload_og_image` at `:1103`. Nothing versions the OG image on read, so that second bump buys nothing at all.

## Correction to the original write-up

The "on a shared instance, viewing someone else's map mutates its metadata" line overstated it. The write is gated twice: `require_permission("edit_metadata")` on both endpoints, then `check_map_ownership` (`backend/app/modules/catalog/maps/service_crud.py:30-40`), which admits only the map's creator or an admin. A non-owner viewing a map cannot trigger the capture at all; it 403s.

The trigger surface is also narrower than "viewing". `maybeAutoCaptureThumbnail` lives in the builder's save hook, so it fires when the owner opens the map in the builder, not when anyone opens the read-only viewer. The bug is real, but its shape is "an owner opening their own untouched map reorders their own gallery", which is why it showed up against admin-owned showcase maps.

## Fix: give the thumbnail its own timestamp

Option 2 from the original notes. Splitting the two meanings is the only fix that keeps both correct; any flag or route split still leaves one column meaning two things.

1. Add `Map.thumbnail_updated_at` (`backend/app/modules/catalog/maps/models.py`, next to `thumbnail_uri` at `:113`), nullable `timestamptz`. Write the Alembic migration and confirm with `make alembic-check`.
2. In `upload_thumbnail` (`router.py:968`) and `upload_og_image` (`:1103`), set `thumbnail_updated_at` instead of `updated_at`. Neither is a metadata edit.
3. Add the field to `MapSummaryResponse` (`backend/app/modules/catalog/maps/schemas.py:1077-1088`), which is what `MapCard` and `MapCardGrid` are typed against, and mirror it in `frontend/src/types/api.ts`.
4. Switch both hook call sites to the new field, falling back to `updated_at` so maps written before the migration keep a stable version rather than losing their cache-busting entirely.
5. Rewrite `test_upload_thumbnail_bumps_updated_at`, do not delete it. It is the only thing pinning the cache-invalidation contract. It should assert that a thumbnail upload moves `thumbnail_updated_at` and leaves `updated_at` alone, and it should keep its docstring's intent. Rename it to match. Add the same assertion for the OG upload.
6. Regenerate `backend/openapi.json` and the SDKs (`make openapi`, `make sdks`), and commit them, since the response shape changes.

## Acceptance criteria

- Opening an owned map with no thumbnail in the builder does not change its `updated_at`, and does not move it in the "Last updated" gallery sort.
- Re-capturing a thumbnail after an edit still shows the new image without a hard refresh.
- `make alembic-check`, `make openapi-check`, and `make sdks-check` pass.

ishiland:
Closes #1005. **Migration — leaving this for the coordinator to merge.**

## The bug

`updated_at` was doing double duty as the thumbnail cache version.
`MapCard.tsx:28` and `MapCardGrid.tsx:22` pass it to `useMapThumbnail`, which
appends it as `?v=` and puts it in the query key, so a thumbnail re-captured
after an edit only appears because `updated_at` moved. That half is
load-bearing, which is why the obvious one-line fix is a regression.

The same two endpoints also serve the lazy backfill: `maybeAutoCaptureThumbnail`
fires when an owner opens a thumbnail-less map in the builder, editing nothing,
and that write reordered the "Last updated" gallery. One `captureThumbnail` call
drives both flows and uploads two images (400x250 thumbnail + 1200x630 OG card),
so each backfill bumped `updated_at` twice — and nothing versions the OG image
on read, so that second bump bought nothing at all.

## What changed

- `Map.thumbnail_updated_at`, nullable `timestamptz`, migration
  `0031_maps_thumbnail_updated_at`.
- Both `upload_thumbnail` and `upload_og_image` write it instead of `updated_at`.
- `MapSummaryResponse` carries it; mirrored in the hand-typed
  `frontend/src/types/api.ts` and regenerated into `openapi.json`,
  `api.generated.ts` and both SDKs.
- Both card call sites version on `thumbnail_updated_at ?? updated_at`, so a map
  captured before the column keeps a stable version rather than losing
  cache-busting entirely.

Not backfilled from `updated_at` on purpose: the frontend falls back to it
anyway, so a backfill would only invent a timestamp that is already available.

## The part the issue's step 2 would have missed

Dropping the endpoints' explicit `map_obj.updated_at = datetime.now(UTC)` does
**not** stop the bump. `Map.updated_at` carries `onupdate=func.now()`
(`models.py:148-150`), so *any* update to the row bumps it. Measured, with the
assignment removed and nothing else changed:

```
AssertionError: assert '2026-07-31T13:43:32.519252Z' == '2026-07-31T13:43:32.454139Z'
```

So both endpoints now go through one `_record_image_capture` helper that sets
`updated_at` to its own column value. That is the explicit-value form which
suppresses the `onupdate`, and because it happens in the database it neither
invents a timestamp nor clobbers a concurrent edit's. The docstring records
this, since "just stop assigning it" is the natural next edit.

## Verification

```
  tests/test_maps.py tests/test_maps_visibility_check_authz_sec007.py \
  tests/test_builder_audit_p0p1_sharing.py \
  tests/test_dormant_tenancy_migration_roundtrip.py \
  tests/test_check_constraint_parity.py tests/test_layering.py -q
  # 189 passed (+34 layering)
cd frontend && npm run test:coverage   # 366 files, 4288 tests passed
```

`alembic check` reports no drift (`TestAlembicCheckNoDrift`), which is the
`make alembic-check` gate run against the test DB — `make alembic-check` itself
targets the dev container, which serves `main`'s code from a worktree.

`test_upload_thumbnail_bumps_updated_at` is rewritten, not deleted, keeping its
docstring's intent. It is now parametrised across both upload endpoints and
asserts both halves: the card gets a new version, and the map does not come back
looking edited. LOC ratchets move for `maps/router.py` (1385 → 1417) and
`maps/schemas.py` (1312 → 1316), exact as the gate requires.

### DIFF
--- a/backend/alembic/versions/0031_maps_thumbnail_updated_at.py
+++ b/backend/alembic/versions/0031_maps_thumbnail_updated_at.py
@@ -1,0 +1,45 @@
+"""Give the map thumbnail its own timestamp, separate from the map's.
+
+fix(#1005): ``catalog.maps.updated_at`` was doing double duty as the thumbnail
+cache version. ``MapCard``/``MapCardGrid`` pass it to ``useMapThumbnail`` as the
+``?v=`` version and part of the query key, so a re-captured thumbnail only shows
+after an edit because ``updated_at`` moved. But the same upload endpoints serve
+the lazy backfill that fires when an owner first opens a thumbnail-less map in
+the builder, editing nothing -- and that write bumped ``updated_at`` too, which
+reordered the "Last updated" gallery for what was a read.
+
+The backend cannot tell the two flows apart: one ``captureThumbnail`` call
+drives both, and it uploads two images (the 400x250 thumbnail and the 1200x630
+OG card), so each backfill bumped ``updated_at`` twice.
+
+``thumbnail_updated_at`` splits the two meanings. It is nullable rather than
+backfilled from ``updated_at``: the frontend falls back to ``updated_at`` when
+it is null, so existing maps keep exactly the cache version they have today,
+and a backfill would only invent a timestamp that is already available.
+
+Revision ID: 0031_maps_thumbnail_updated_at
+Revises: 0030_records_spatial_extent_type
+Create Date: 2026-07-31
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "0031_maps_thumbnail_updated_at"
+down_revision: Union[str, None] = "0030_records_spatial_extent_type"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "maps",
+        sa.Column("thumbnail_updated_at", sa.DateTime(timezone=True), nullable=True),
+        schema="catalog",
+    )
+
+
+def downgrade() -> None:
+    op.drop_column("maps", "thumbnail_updated_at", schema="catalog")
--- a/backend/app/modules/catalog/maps/models.py
+++ b/backend/app/modules/catalog/maps/models.py
@@ -112,6 +112,18 @@ class Map(Base):
     # Preview thumbnail — storage key (e.g. "maps/thumbnails/{id}.jpg")
     thumbnail_uri: Mapped[str | None] = mapped_column(Text, nullable=True)
 
+    # fix(#1005): the thumbnail's own cache version. `updated_at` used to do
+    # double duty here, which meant the lazy backfill on first builder open —
+    # a read, editing nothing — bumped the map's edit timestamp and reordered
+    # the "Last updated" gallery. Splitting the two meanings is the only fix
+    # that keeps both correct: the card still busts its cache when the image is
+    # re-captured after an edit, without the capture claiming the map changed.
+    # Nullable, and the frontend falls back to `updated_at`, so maps written
+    # before this column keep a stable version rather than losing one.
+    thumbnail_updated_at: Mapped[datetime | None] = mapped_column(
+        DateTime(timezone=True), nullable=True
+    )
+
     # OG/social-card image — storage key (e.g. "maps/og-images/{id}.jpg")
     # Added in migration 0001_baseline (SHARE-08 Path A). Separate from thumbnail_uri
     # because the OG image is 1200x630 and exceeds the 100KB thumbnail cap.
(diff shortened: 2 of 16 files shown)


ishiland:
@codex review

PR #1053, opened 2026-07-31
ops(analysis): per-tenant cap on active materialize jobs

### LINKED ISSUE #1015: ops(analysis): per-tenant cap on active materialize jobs

ishiland:
The materialize admission cap counted only the requesting user's active jobs (`router_analysis.py:295-315`), so a tenant with N users could hold N active CTASes regardless of what the host can absorb. That matters more now than when the item was filed, because the sibling split on scoped `work_mem` (#1012) raises the per-statement memory each of those would claim — a raised ceiling times an unbounded count is the same outage in a new place.

#691 has shipped, so this was unblocked.

## The decision the issue left open

**Both caps stay.** The per-user cap answers "are you already running one?" and the tenant cap answers "is your organization using more than its share?" — different questions, so collapsing them would lose one. That means two 429 details, and they say which was hit: the tenant one names the organization and the limit, so a user with nothing of their own running is not told "an analysis job is already running".

## The number

`MAX_ACTIVE_MATERIALIZES_PER_TENANT = 3`. With the default `WORKER_CONCURRENCY` of 1 that is one running plus two queued, so a single tenant cannot monopolize the queue while still being able to line work up. On a deployment that raises `WORKER_CONCURRENCY`, it bounds genuinely concurrent CTASes to three per tenant — which is the count #1012's `work_mem` division is sized against.

A module constant rather than a `Settings` field, for the same reason as the ingest caps: four surfaces for a limit nobody has asked to tune yet. #1013 establishes the promotion pattern.

## The predicate is shared, not reinvented

#691 replaced status-only counting with a heartbeat lease, and the tenant cap had to use that rule rather than inventing a second one. The pending/running predicate is now lifted into a single tuple that both counts apply, so all three load-bearing properties hold identically at tenant scope:

1. a running job whose lease has gone stale releases the slot (hard-killed worker), instead of holding the whole tenant out until the 60-minute `JOB_TIMEOUT_SECONDS` backstop
2. the pending branch stays status-only — a pending job has never been claimed, so `heartbeat_at` and `started_at` are both NULL and any cutoff comparison would silently drop it
3. the marker is `user_metadata.has_key("analysis")`, never `source_filename`

Elapsed-time counting is not reintroduced.

## Tenant scoping

Single-tenant mode takes the unfiltered count — that deployment has one tenant by definition, so the count IS the tenant's, and it avoids depending on whether the database stamps `tenant_id` as NULL or as a fixed value there. Multi-tenant filters on `IngestJob.tenant_id`, which already exists and is already indexed (`ix_catalog_ingest_jobs_tenant_id`), so there is **no schema work**.

## Tests

- `test_tenant_cap_refuses_a_second_user_and_names_itself` — a second user in the tenant holds the cap in `pending` jobs (which also exercises the status-only branch), and this user's request 429s with a detail that is distinguishable from the per-user message
- `test_tenant_cap_stale_lease_releases_a_slot` — one of the filler jobs is a hard-killed worker (stale heartbeat), and the request is admitted

Both verified to fail with the cap removed.

Two notes on the shared worker database, since this file's tests already collide there. The filler helper **counts what is already active** and tops up rather than assuming zero, so a sibling that left a job behind cannot make these pass or fail for the wrong reason. And the refusal test releases the job it would create if the cap ever regresses — otherwise a regression would hold this user's per-user slot for every sibling test, which is the #933 failure mode.

## Verification

```
cd backend && uv run pytest tests/test_analysis_materialize.py -q       # 81 passed
cd backend && uv run pytest tests/test_layering.py -q                   # 34 passed
cd backend && uv run ruff check . && uv run ruff format --check .       # clean
```

No schema or API-contract change: the 429 status is unchanged, only a new detail string.

Closes #1015

### DIFF
--- a/backend/app/modules/catalog/datasets/api/router_analysis.py
+++ b/backend/app/modules/catalog/datasets/api/router_analysis.py
@@ -5,7 +5,7 @@ import uuid
 from datetime import datetime, timedelta, timezone
 
 from fastapi import APIRouter, Depends, HTTPException, Request, status
-from sqlalchemy import and_, func, or_, select
+from sqlalchemy import and_, func, or_, select, text
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from app.core.db.tenant_session import current_tenant_var, defer_async_with_tenant
@@ -312,6 +312,33 @@ async def analysis_materialize_endpoint(
     # The client deliberately applies no staleness rule of its own
     # (AnalysisJobWatcher.tsx): it mirrors job status from the API, and on a
     # released lease the next create attempt simply succeeds server-side.
+    # fix(#1015): serialize admission per tenant before counting anything.
+    # Without it the caps are check-then-insert: N users in one tenant can all
+    # read a count below the ceiling, then all create a job, and the tenant
+    # ends up over by however many callers arrived together — precisely the
+    # unbounded-concurrency case the ceiling exists to stop, and the one #1012's
+    # raised work_mem makes expensive. A transaction-scoped advisory lock held
+    # until this request commits makes the count-then-create sequence atomic.
+    #
+    # Blocking rather than pg_try_advisory_xact_lock (the sandbox's shape at
+    # executor.py): concurrent admissions should queue for the microseconds this
+    # takes, not fail. The critical section is a COUNT and an INSERT.
+    #
+    # In single-tenant mode the key is constant, so this serializes every
+    # analysis admission on the deployment. That is the correct reading — there
+    # is one tenant — and admissions are rare and short. It also incidentally
+    # hardens the per-user cap below, which the comment there describes as soft.
+    await db.execute(
+        text("SELECT pg_advisory_xact_lock(hashtextextended(:admission_key, 0))"),
+        {
+            "admission_key": (
+                f"geolens:analysis-admission:{current_tenant_var.get()}"
+                if is_multi_tenant()
+                else "geolens:analysis-admission"
+            )
+        },
+    )
+
     lease_cutoff = datetime.now(timezone.utc) - timedelta(
         seconds=MATERIALIZE_LEASE_SECONDS
     )
--- a/backend/app/modules/catalog/datasets/api/router_analysis.py
+++ b/backend/app/modules/catalog/datasets/api/router_analysis.py
@@ -8,9 +8,10 @@ from fastapi import APIRouter, Depends, HTTPException, Request, status
 from sqlalchemy import and_, func, or_, select
 from sqlalchemy.ext.asyncio import AsyncSession
 
-from app.core.db.tenant_session import defer_async_with_tenant
+from app.core.db.tenant_session import current_tenant_var, defer_async_with_tenant
 from app.core.dependencies import get_db
 from app.core.identity import Identity
+from app.core.tenancy import is_multi_tenant
 from app.modules.auth.dependencies import get_current_active_user, require_permission
 from app.modules.catalog.authorization import check_dataset_access
 from app.modules.catalog.datasets.domain.schemas import (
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

ishiland:
@codex review

PR #1054, opened 2026-07-31
fix(maps): export a builtin fill pattern as a solid fill

### LINKED ISSUE #917: fix(maps): exported style.json passes builtin fill-pattern ids the served sprite does not contain

ishiland:
Closes #917. Backend counterpart of the builder fill-pattern work (#910/#911/#914).

## What was broken

A polygon layer using a builtin fill pattern (`paint['fill-pattern'] = 'geolens-fill-hatch'`)
exported a `style.json` whose `sprite` is indexed from the `map_icons` table
alone and has never contained `geolens-fill-*`. In-app surfaces work only
because the client-side adapter registry generates those images in the browser
(`layer-adapters/fill-pattern-images.ts`); the exported document has no such
luck.

## What changed

Option 2 from the issue. `_validate_emitted_style` drops a builtin
`fill-pattern` from `fill` layers and falls back to `fill-color`, keeping an
authored colour when there is one and using `DEFAULT_FILL_COLOR` otherwise.

**Plain strings only.** A composite value — a data-driven expression or a legacy
function object — is left exactly as authored. Two narrowings from review:

- **Not the `geolens-fill-` prefix.** `create_icon_asset` derives a sprite slug
  from the uploaded filename, so an icon named `geolens-fill-logo.png` becomes a
  real `map_icons` entry carrying that prefix.
- **Not composites at all.** See below — this one is a deliberate design
  decision, not a coverage gap.

## Why composites are left alone

Two facts from the MapLibre GL JS docs:

- A missing pattern fires **`styleimagemissing`**, described in the docs as the
  hook to add the image "to prevent the image from being skipped". A dangling
  reference is **skipped, not fatal** — the layer still renders and a consumer
  can supply the image at runtime.
- **"Use a `coalesce` expression to display another image when a requested image
  is not available"** is a documented MapLibre example.

So the failure modes are asymmetric. Leaving a builtin in an export degrades and
**stays repairable**. Stripping a working expression removes authored content
that **no downstream hook can bring back**. Reading composites was trading the
permanent failure for the recoverable one — `["coalesce", ["image", builtin],
["image", other]]` is the sanctioned way to author a fallback, and the detector
was deleting it.

Coverage against the reported bug is unaffected: the **builder** writes a plain
string, which is all GeoLens itself produces. Every composite arrives through
style import or the open-dict `paint` API.

## On the review history

This PR ran nine codex rounds, and **every finding was correct against the
implementation as it stood** — both the gaps in rounds 2-7 and the false
positives in round 8. What changed is not their validity but the scope they were
correct about: fixing each one built a composite reader that outgrew what #917
specified ("the two-line stopgap"), and round 8 showed the model was wrong in a
way that destroyed valid input. Round 9 then produced two more, which is the
evidence that the surface was unbounded.

The narrower end state is the deliberate choice, made with the MapLibre
semantics above in hand — not a retreat from review.

Nothing learned is discarded. The expression semantics worked out across those
rounds (output positions per operator, `image` null-propagation, `coalesce`
short-circuit, `let` binding reachability, the 64KB depth problem) are recorded
on #917 so a real evaluator starts from them.

## Filed separately

**#1069** — `_validate_style_dict` bounds only serialized size, so
`{"fill-pattern": {"stops": 1}}` is storable today and any serializer that
descends into it 500s the shared style endpoint from stored data. Surfaced by
this work, live regardless of it.

## Verification

```
  tests/test_maps_style_json.py tests/test_maps.py tests/test_layering.py -q
  # 275 passed
```

Plain-string coverage kept: default colour on strip, an authored `fill-color`
surviving (the fallback is a default, not an override), and a non-builtin left
alone over both a plain uploaded slug and one whose filename gave it the builtin
prefix. The composite matrix is replaced by one test asserting every composite
shape — `case`, `match`, `coalesce`, `let`, `image`, legacy function, and the
malformed `{"stops": 1}` forms — exports byte-identical to what was authored.

`style_json.py` LOC ratchet 1432 → 1517.

### DIFF
--- a/backend/app/modules/catalog/maps/style_json.py
+++ b/backend/app/modules/catalog/maps/style_json.py
@@ -1312,10 +1312,28 @@ _EXPRESSION_OUTPUT_POSITIONS: dict[str, tuple[int, int]] = {
     # produced the same disappearing polygon as a bare string.
     "image": (1, 1),
     "literal": (1, 1),
+    # fix(#917 codex r5): ["let", name, value, ..., result]. The binding VALUES
+    # sit at the same 2-step offset the branching forms use and the result is
+    # last, so `let` needs no special handling beyond the table — a `["var", n]`
+    # in the result reaches its binding, and reading the binding directly
+    # answers the question without resolving the reference. Binding NAMES land
+    # on the odd indices the stride skips, exactly like `match` labels.
+    "let": (2, 2),
 }
 
+# The forms whose LAST operand is also an output: a `case`/`match` fallback, and
+# a `let` result. `step` and `coalesce` are fully covered by their stride.
+_EXPRESSION_TRAILING_OUTPUT = frozenset({"case", "match", "let"})
 
-def _references_builtin_fill_pattern(value: Any) -> bool:
+# fix(#917 codex r5): a bound on how deep an authored expression is followed.
+# `_validate_style_dict` caps a paint dict at 64KB serialized, which still
+# allows nesting far past Python's recursion limit, and a RecursionError here
+# would be a 500 from the shared style endpoint. Anything this deep is not a
+# real pattern; stop descending and leave the value alone.
+_MAX_EXPRESSION_DEPTH = 32
+
+
+def _references_builtin_fill_pattern(value: Any, depth: int = 0) -> bool:
     """Whether a fill-pattern value can resolve to a builtin id.
 
     A plain string is checked directly. A branching expression is checked at its
@@ -1326,6 +1344,8 @@ def _references_builtin_fill_pattern(value: Any) -> bool:
     the builder's pattern control writes a single value, so anything else is
     hand-authored, and guessing wrong here silently flattens a working layer.
     """
+    if depth > _MAX_EXPRESSION_DEPTH:
+        return False
     if isinstance(value, str):
         return value in _BUILTIN_FILL_PATTERN_IDS
     if isinstance(value, dict):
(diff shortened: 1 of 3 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1055, opened 2026-07-31
feat(auth): per-key scopes for API keys

### LINKED ISSUE #875: feat(auth): per-key scopes for API keys — least-privilege machine credentials
## Gap

An API key currently impersonates its owner completely. `_resolve_api_key()` in `backend/app/modules/auth/dependencies.py` resolves the key straight to the `User`, so every key carries the owner's full role and capability set. There are no read-only keys.

Anyone building an application on top of GeoLens must embed a credential that can also mutate or delete everything its owner can. The workaround today is a dedicated low-privilege user per application, which works but is easy to skip.

There is a second reason to want this. The `?api_key=` query-parameter lane is deprecated (#821, `dependencies.py:64-70`) but deliberately kept, because XYZ tile URLs in desktop GIS tools cannot set headers — so a credential in a URL, written into access logs and any upstream proxy's logs, is a shape we are stuck with. A read-only key is far less dangerous in that position than a full-privilege one.

## Design decision (already made)

Enforcement is **HTTP-method based**, at the key-resolution chokepoint, not capability based. A `read_only` key may authenticate `GET`, `HEAD`, and `OPTIONS` requests; any other method gets `403` with a clear message. Rationale:

- The read surfaces an API-key client actually uses (OGC Features, STAC, tiles, search, dataset/map reads) are all `GET` in this codebase. The STAC and OGC routers define no `POST`/`PUT`/`PATCH`/`DELETE` routes at all.
- Classifying every capability in the permission matrix as read or write is a much larger change and easy to get wrong.
- `POST` endpoints that are "reads" in spirit (AI chat, analysis previews) can trigger jobs and writes, so blocking them under `read_only` is the correct default anyway.

Do not enforce this in `require_permission` or middleware. Enforce it inside `_resolve_api_key()`, which already has the `Request` object. Note the failure mode must be `raise HTTPException(403, ...)`, not `return None`: returning `None` would fall through to the anonymous/JWT path and turn a scope violation into a confusing 401.

### 1. Model + migration

`backend/app/modules/auth/models.py`, class `ApiKey`: add

```python
scope: Mapped[str] = mapped_column(String(20), server_default="full", nullable=False)
```

Add a `CheckConstraint("scope IN ('full', 'read_only')", name="chk_api_keys_scope")` in `__table_args__` (follow the pattern on `User.status`).

Note the conversion this requires: `ApiKey.__table_args__` is currently the dict form, `{"schema": "catalog"}` (`models.py:188`). Adding a constraint means switching it to the tuple form, `(CheckConstraint(...), {"schema": "catalog"})`, with the dict last. `User.__table_args__` (`models.py:24`) is already in that shape and is the pattern to copy.

Create the Alembic migration (`backend/alembic/`, follow the newest revision's style). `server_default="full"` backfills existing keys with unchanged behavior. Run `make alembic-check`: it fails CI if the model and migration drift.

### 2. Schemas

`backend/app/modules/auth/schemas.py`:

- `ApiKeyCreateRequest`: add `scope: Literal["full", "read_only"] = "full"` with a description.
- `ApiKeyCreateResponse` and `ApiKeyListItem`: add `scope: str`.

### 3. Service

`backend/app/modules/auth/service.py`, `create_api_key_for_user()` (line ~427): accept `scope: str = "full"` and set it on the `ApiKey` row.

### 4. Router

There are **two** key-minting surfaces, not one. Both need wiring.

**Self-service** — `backend/app/modules/auth/router.py` defines three key routes: `GET /api-keys` (`:807-808`), `POST /api-keys` (`:841-847`), and `DELETE /api-keys/{key_id}` (`:902`).

- `create_my_api_key`: pass `body.scope` through, include `"scope"` in the `audit_emit` details dict, and return it in the response.
- `list_my_api_keys`: add `scope` to the items it builds.

**Admin** — `backend/app/modules/admin/router_operations.py` mints keys for other users: `POST /api-keys` (`:47-58`, handler `create_api_key` at `:59`), `GET /api-keys` (`:122-132`), and `DELETE /api-keys/{key_id}` (`:151`), all gated on `require_permission("manage_users")`. Its schemas are `AdminApiKeyCreateRequest` and `AdminApiKeyListItem` (`backend/app/modules/admin/schemas.py:356` and `:376`), separate from the self-service ones.

- `create_api_key`: add `scope` to `AdminApiKeyCreateRequest` and pass it to `create_api_key_for_user` alongside `expires_at`, so an admin minting a key for a service account can make it read-only. Include it in the audit details.
- `list_api_keys` / `_api_key_response` (`:33`): add `scope` to `AdminApiKeyListItem` so the admin key list shows what each key can do.

Leaving the admin path unwired would mean admin-minted keys are silently always `full`, which is the opposite of the least-privilege story here: the service-account key an admin mints for an application is the single most likely key to want `read_only`.

### 5. Enforcement

`backend/app/modules/auth/dependencies.py`, `_resolve_api_key()` (`:62`). Insert the check **after** the `last_used_at` block that ends at `:122`, immediately before the final `return user` at `:123`:

```python
if api_key_obj.scope == "read_only" and request.method not in ("GET", "HEAD", "OPTIONS"):
    raise HTTPException(
        status_code=status.HTTP_403_FORBIDDEN,
        detail="This API key is read-only",
    )
```

That placement is deliberate: the key did authenticate, so usage is recorded before the scope violation rejects it. Putting the check earlier (right after the epoch/active checks) would skip the bump, because the `last_used_at` update is the last thing the function does. Assert the choice in a test either way.

### 6. Regenerated artifacts

The schema change alters the OpenAPI contract. Regenerate all three artifacts in order: `make openapi` (updates `backend/openapi.json`), then `make sdks` (Python + TypeScript SDKs). `make openapi-check` and `make sdks-check` are the CI gates. Never hand-edit generated SDK files; only the `auth.*`/`__init__`/`index` wrappers are hand-maintained. The frontend's `types/api.generated.ts` is also regenerated, not edited.

### 7. Frontend

Both surfaces have a frontend.

- `frontend/src/components/settings/MyApiKeySection.tsx`: add a scope choice to the create form (two radio options or a select; reuse primitives from `components/ui/`) and a scope column or badge in the key list.
- `frontend/src/components/admin/ApiKeySection.tsx` (rendered from `UserEditDialog.tsx`) and `frontend/src/hooks/use-admin.ts` (`useApiKeys` at `:305`, `useCreateApiKey` at `:313`): same treatment for the admin surface.
- `frontend/src/api/auth.ts` and `frontend/src/api/admin.ts` hand-mirror the request bodies and both need the new field. **`createMyApiKey` (`auth.ts:70-75`) currently sends only `{ name }`** — it does not thread `expires_at` either, even though `ApiKeyCreateRequest` has accepted it since #821 (`backend/app/modules/auth/schemas.py:230`). `createApiKey` (`admin.ts:122-127`) sends `{ user_id, name }` with the same omission. That is a pre-existing bug, not something this issue introduced, but it sits exactly where the new field goes: fix it in the same PR rather than adding `scope` on top of an already-incomplete mirror.
- Every new `t()` key must be added to **all four** locales (`frontend/src/i18n/locales/{en,es,fr,de}`). A `defaultValue` alone fails the `npm run test:i18n` CI gate.

### 8. Tests

Backend (`backend/tests/`, pytest + AnyIO; run per AGENTS.md "Running a single test"):

- `read_only` key on a `GET` catalog endpoint succeeds.
- `read_only` key on a `POST`/`DELETE` endpoint returns 403 with the read-only detail, not 401.
- `full` key behavior is unchanged on both.
- A key minted before the migration (no scope supplied) resolves as `full`.
- Existing expiry and `key_epoch` tests still pass; the new check must not change those code paths.
- Both create endpoints round-trip the scope and it appears in the list response and the audit event details, self-service and admin.
- A `read_only` key succeeds on `POST /api/query/` if that route exists by then (see the #565 carve-out below), and still gets 403 on another `POST`.

Frontend: extend `MyApiKeySection.test.tsx`, `use-api-keys.test.ts`, and the admin `ApiKeySection` / `UserEditDialog` tests for the new field.

Gates before pushing: `cd backend && uv run ruff check . && uv run ruff format --check .`, `make test`, and the frontend suite (`npm run typecheck`, not `npx tsc --noEmit`, which is a no-op here).

## Acceptance criteria

- A user can mint a `read_only` key from the settings UI and via `POST /auth/api-keys`, and an admin can mint one for another user from the admin UI and via `POST /admin/api-keys`.
- That key can read public and (per existing visibility rules) private data, fetch tiles, and hit OGC/STAC endpoints.
- Any mutating request with that key returns 403 with a message naming the key as read-only.
- All existing keys keep working exactly as before.
- `make alembic-check`, `make openapi-check`, `make sdks-check`, `npm run test:i18n` all pass.

## Out of scope until there is demand

- Dataset-scoped keys
- OAuth2 client-credentials / consent flow (GeoLens as an authorization server)
- Per-key rate quotas

## Interaction with #565 (hardened sandbox SQL endpoint) — decided

#565 proposes `POST /api/query/`, a SELECT-only sandbox endpoint that is a pure read semantically but a `POST` mechanically. Under the method rule above, a `read_only` key would be refused by the endpoint most useful to a read-only machine client, since the natural consumer of #565's MCP `query` tool is exactly the client this issue exists to serve.

**Maintainer decision: a `read_only` key MAY call `POST /api/query/`, because it is a read.** The rationale above ("`POST` endpoints that are 'reads' in spirit... can trigger jobs and writes") holds for AI chat and analysis previews and does not hold for a raw SELECT against a SELECT-only sandbox.

Whichever of the two ships second implements the carve-out. If this issue ships first, leave a named hook for it rather than a general escape hatch: the exemption must be a specific route, not a "POST that looks like a read" category, so a future `POST` cannot inherit it by resembling one. Assert it in a test on both sides — the carved-out route succeeds, another `POST` still returns 403. The same decision is recorded in #565.

## Context

From the 2026-07-29 review of the application-access story. Related: #688 (signed raster tile URLs for API-key clients), #565.

ishiland:
Least-privilege machine credentials. An API key currently impersonates its
owner completely — `_resolve_api_key` resolves it straight to the `User`, so
every key carries the owner's full role and capability set. Anyone building an
application on GeoLens has to embed a credential that can also mutate or delete
everything its owner can, or stand up a dedicated low-privilege user per
application and remember to.

The deprecated `?api_key=` query lane sharpens it. That lane is kept on purpose
(XYZ tile URLs in desktop GIS tools cannot set headers), so a credential in a
URL — written into access logs and any upstream proxy's logs — is a shape we
are stuck with. A read-only credential is far less dangerous sitting there.

## Enforcement

Method-based at the key-resolution chokepoint. A `read_only` key authenticates
`GET`/`HEAD`/`OPTIONS`; anything else gets `403 This API key is read-only`.
Not capability-based: every read surface an API-key client actually uses (OGC
Features, STAC, tiles, search, dataset/map reads) is a `GET` here, the STAC and
OGC routers define no write routes at all, and classifying the whole permission
matrix as read or write is a much larger change that is easy to get subtly
wrong. `POST` endpoints that are reads in spirit — AI chat, analysis previews —
can trigger jobs and writes, so refusing them is the correct default.

Three details that are decisions rather than mechanics, each with a test:

- **It raises, it does not return None.** Returning `None` falls through to the
  anonymous/JWT path and turns a scope violation into a confusing 401. Asserted
  on a required-auth route and on an optional-auth route.
- **It sits after the `last_used_at` bump.** The key did authenticate; the
  request is refused on what it asked to do, so a client hammering writes with
  a read-only key still shows a moving `last_used_at` instead of looking
  dormant.
- **The `#565` carve-out is an exact `(method, route template)` pair.** A
  `read_only` key may call the SELECT-only `POST /api/query/` because it is a
  read — decided in both #875 and #565. Pairs rather than bare templates so a
  future `DELETE /api/query/{id}` cannot inherit it; an exact list rather than
  a "POST that looks like a read" category so nothing inherits it by
  resembling one. Matching is on the template Starlette resolved, never the
  concrete path; an unresolvable template is `<unmatched-route>`, which is in
  no pair and so is refused.

The route is not mounted yet, and `test_565_route_is_not_mounted_yet` asserts
that, so whoever lands #565 has to come back here and replace it with the live
assertions. Both spellings are exempt, because ROUTE-01's dual-shape decorator
registers the trailing-slash form and a hidden bare form for the same handler
and `redirect_slashes` is off.

### One consequence worth a maintainer's eye, not decided here

`POST /tiles/tokens/` is the batch tile-token endpoint. It is a pure read — it
mints short-lived read tokens and writes nothing — but it is a `POST`, so a
`read_only` key is refused there and a multi-layer builder map would fall back
to the per-dataset `GET /tiles/token/{id}/`, one request per layer. That is
correct under the rule as written, and it is exactly the "POST that is a read
in spirit" category the issue says not to create. So I have not carved it out:
the issue names one exemption and says the list must stay specific. Flagging it
because the tile client is one of the callers this feature exists to serve. If
it should be exempt, it is a second named pair and a second decision.

## Schema

`catalog.api_keys.scope`, `String(20)`, `NOT NULL`, `server_default='full'`,
plus `chk_api_keys_scope`. The default is what makes it backward compatible:
every pre-0031 key reads back as `full` and behaves exactly as before, which
`test_key_predating_the_migration_resolves_as_full` pins by resetting the
column to its default.

`NOT NULL` because a NULL scope has no safe interpretation — treating it as
`full` at the call site would make an accidental NULL a silent privilege grant.

**The downgrade refuses rather than coerces.** Dropping the column widens every
`read_only` key back to its owner's full privileges, which is a privilege
escalation rather than a data loss, so it raises with the query to find the
affected keys and the command to deactivate them first. Same shape as 0030's
downgrade guard. Verified both directions against a scratch database, plus
`alembic check` → "No new upgrade operations detected".

**Chain note for the coordinator:** this claims revision `0031`, down_revision
`0030_records_spatial_extent_type`. #1005 and #765 are also landing migrations
this wave, so whichever merges second needs its `down_revision` re-pointed.

## Both mint surfaces

Self-service (`auth/router.py`) and admin (`admin/router_operations.py`).
Leaving admin unwired would invert the story: the service-account key an admin
mints for an application is the single most likely key to want `read_only`.
Scope round-trips through create, list, and the `api_key.create` audit details
on both.

## Frontend

A shared `ApiKeyScopeSelect` on both mint forms, so the two surfaces cannot
offer different choices, and a badge on `read_only` rows in both lists — a
read-only key is a different credential in practice and the list has to say
which one you are holding.

Also fixes a pre-existing omission the new field sits on top of: `createMyApiKey`
sent only `{ name }` and `createApiKey` only `{ user_id, name }`, so
`expires_at` — accepted by both request schemas since #821 — was never threaded
from the UI at all. Adding `scope` on top of an already-incomplete mirror would
have left it half-wired, so both now thread `expires_at` too, with a contract
test for it.

New keys (`apiKeys.scope`, `scopeFull`, `scopeReadOnly`) are in all four
locales.

## Regenerated artifacts

`backend/openapi.json`, both SDKs, and `frontend/src/types/api.generated.ts`,
all from `make openapi && make sdks` plus `node scripts/generate-api-types.mjs`.
No generated file was hand-edited. The hand-typed mirrors in
`frontend/src/types/api.ts` gained `ApiKeyScope`, which the existing
`hand-typed-mirror-drift` test checks against the generated types.

## Verification

```
cd backend && uv run pytest -q -n 4                 # 6332 passed, 83 skipped
cd backend && uv run pytest tests/test_api_key_scope_875.py tests/test_api_key_auth.py tests/test_auth.py -q
cd backend && PYTHONPATH=. uv run alembic upgrade head && PYTHONPATH=. uv run alembic check
cd frontend && npx vitest run src/api src/components src/hooks    # 3258 passed
make openapi && make sdks
```

`make alembic-check` was not run: it executes inside the api container, which
bind-mounts the main checkout, so from a worktree it would check `main`'s models
against `main`'s migrations. The host `alembic check` above is the equivalent
and CI runs the container form.

Closes #875


ishiland:
@codex review

ishiland:
All three fixed.

**P1, mutating GET.** The finding is right and it is the sharpest thing here: the write guard on that handler sees the owner identity the key resolved to and cannot tell a read_only key from the owner's own session, so the method check is the only place that can refuse it.

Refused by route template **and** by the query parameter that makes it a write, so the ordinary cached read of the same route keeps working. Any present value except an explicit false counts, so it fails closed on `?refresh=`.

More importantly, blanket-allowing GET is now backed by evidence rather than assertion. A route-table walk finds every GET handler whose source shows a write guard or a commit and requires each to be classified — read-only-refused, or benign with the reason — asserted exact in both directions. That surfaced 14 routes: the validate pair, two owner-gated reads that write nothing, eight reads that commit an audit row for having been read, and the two browser OAuth legs an API key never drives. A new writing GET now has to be decided rather than silently inherited.

**P2, STAC POST search.** Right, and it is required rather than a widening: "That key can ... hit OGC/STAC endpoints" is one of the issue's acceptance criteria, and `POST /stac/search` is the standard's JSON-body search surface. Shipping a comment claiming STAC works next to code that broke it was the actual defect. Added as a named pair.

**P2, downgrade remediation.** Right, and worse than cosmetic: counting inactive rows made the advertised `UPDATE ... SET is_active = false` a no-op, so rollback was impossible short of deleting credentials the message never mentioned. Counts only active keys now — an inactive key cannot authenticate, so widening its scope back grants nothing. Verified both branches against a scratch database: refuses with an active read-only key, succeeds after deactivating it.

@codex review

ishiland:
Fixed at `533ffd09b`. You were right on both counts and the second one is the sharper finding.

**The literals.** Verified against the live route table rather than taking it on faith: `app.root_path` is `/api`, and not one template in the route table starts with `/api/`. My own structural test was already evidence of this from the other direction — it classifies the validate route as `/datasets/{dataset_id}/validate/`, no prefix — and I did not connect it. Entries are now `("POST", "/query/")` and `("POST", "/query")`.

Your framing of why it mattered is the part I had wrong in my head: fail-closed made it not-a-hole, but it would have silently defeated the recorded decision while looking correct to whoever lands #565, because the comment pointed them at two entries that read fine.

**The self-confirming test is the better catch.** `_read_only_key_may_call("POST", "/api/query/") is True` handed the function the same literal the module already held. Both sides agreed no matter how wrong the string was — it asserted that a constant equals itself, which is exactly why nothing failed.

Three route-table-driven assertions replace it:

- **no exemption may begin with the app's `root_path`** — the exact bug class, and it fails immediately on the old spelling;
- **every non-pending exemption must resolve to a mounted (method, template) pair** — so a real entry is verified rather than restated. `/stac/search` is now proved mounted *before* its True assertion, instead of being asserted against itself;
- **pending exemptions must still be unmounted** — so when #565 lands, this fails, forcing the entry out of the pending list and under the resolution check, which is the only place the real template can ever be confirmed.

The `/query/` spelling stays unverifiable until the route exists. The comment now says that outright, cites `root_path` as the reason `/api` was wrong, and points at the pending list as the thing to update — rather than a bare "re-read this".

Verification: `test_api_key_scope_875.py`, `test_api_key_auth.py`, `test_auth.py`, `test_stac_api.py`, `test_layering.py`, `test_rule1_structural.py` — 156 passed.

Not nudging it; it is BEHIND and you are serializing.

@codex review

ishiland:
Head moved after the last clean pass. `533ffd09bc` was clean, but `3761aabc6` is on top of it — the migration renumber to `0033_api_key_scope` with `down_revision = "0032_records_derived_from"`, assigned by the coordinator because three open PRs each claimed `0031`.

That commit is a push of mine, not an `update-branch` merge commit, so the sha rule applies and the `533ffd09bc` pass does not cover it.

Note for whoever reads the CI board: it will be uniformly red until #1045 lands, and that is expected rather than a defect. `down_revision` points at `0032_records_derived_from`, which does not exist yet. `conftest.py` builds each worker's test database with `alembic upgrade heads`, so the dangling revision errors every DB-backed test at fixture setup — `KeyError: '0032_records_derived_from'` — not just the alembic job.

@codex review

ishiland:
Re-triggering once. The head churned while the previous request was in flight: I briefly reset this branch to `533ffd09b` and pushed it back to `3761aabc6` a few minutes later, on coordinator instruction that was then withdrawn. Head is `3761aabc6` and staying there — same sha the 14:54 request was made against, so the review target has not actually changed.

State of the branch, unchanged from that request: the migration is `0033_api_key_scope` with `down_revision = "0032_records_derived_from"`, which is the coordinator-assigned position in the chain (0031 thumbnail, 0032 derived_from, 0033 this).

CI on this PR will be uniformly red until #1045 lands, and that is expected rather than a defect: `0032_records_derived_from` does not exist yet, and `conftest.py` builds each worker's test database with `alembic upgrade heads`, so the unresolvable chain errors every DB-backed test at fixture setup rather than failing only the alembic job.

@codex review

ishiland:
Renumbered to `0032_api_key_scope` off `0031_maps_thumbnail_updated_at` at `ec623420f`, now that #1052 has landed and taken 0031.

Done at merge time rather than in advance, deliberately: a `down_revision` naming an unlanded revision does not fail one job. `conftest.py` builds every worker's test database with `alembic upgrade heads`, so an unresolvable chain errors the entire suite at fixture setup — 20 errors on tests with no database of their own.

**Verified rather than assumed**, since the parent being on main is a premise worth checking rather than trusting:

- `alembic heads` -> `0032_api_key_scope (head)`, single head, no multiple-heads state
- `alembic history` -> `0030 -> 0031_maps_thumbnail_updated_at -> 0032_api_key_scope`, linear
- clean database upgrades to head
- `alembic check` -> "No new upgrade operations detected"
- downgrade steps back to `0031_maps_thumbnail_updated_at`

Also updated the two stale comments that referenced the old numbers, in the model and the test.

Suite: 128 passed across `test_api_key_scope_875.py`, `test_api_key_auth.py`, `test_auth.py`, `test_layering.py`, plus ruff.

Not armed.

@codex review

PR #1056, opened 2026-07-31
fix(catalog): block a dataset visibility change that strands an internal map

### LINKED ISSUE #931: fix(catalog): surface the public-map visibility block, and close the internal-map gap

ishiland:
Closes #931. Step 4 of the #940 visibility ladder.

> **Merge after #1029 (#930).** The matrix below is only correct once `internal`
> is a real dataset rung. No file conflict — #930 edits `dataset.json`, this
> adds one key to `common.json`.

## Decision, recorded here for the issue

**Extend the backend block**, not a frontend confirm dialog. Grounds:

- It matches the precedent one rung up: a public map already hard-blocks, and an
  internal map's viewers lose the layer just as completely.
- A warning cannot be honest about its own scope. `GET /datasets/{id}/maps/` is
  RBAC-filtered by the caller and paginated at 50 (cap 200), so the dialog would
  under-report on exactly the datasets where it matters most.
- It makes the two gaps land together correctly. Extending the query makes the
  422's "public maps" wording wrong, and that string is what gap 1's regex
  matches — split across PRs, gap 1 ships a regex for a message that no longer
  exists.

## Gap 2 — internal maps broke silently

`find_public_maps_using_dataset` filtered on `Map.visibility == "public"` only,
and its caller gated on `new != "public" and old == "public"`. An internal map
was invisible to **both** halves, so the flip succeeded and every signed-in
viewer of that map lost the layer's tiles and features with no signal to anyone.

It is now `find_maps_broken_by_dataset_visibility`, and it compares the before
and after audiences rather than listing forbidden target values. #930 turned the
rule into a matrix, and a rule written against today's target value alone gets
cells wrong in both directions.

The four rungs nest — the owner is a grantee is a signed-in user is everyone —
so this is a total order, and "does this change strand someone" is a comparison
of how much of each map's audience the dataset reaches, before and after:

| dataset move | public map | internal map |
| --- | --- | --- |
| public → internal | blocked | **allowed** — the audience still reaches it |
| public → private | blocked | blocked |
| internal → private | blocked | blocked — invisible to the old gate |
| restricted → private | blocked | blocked — the grant holders lose it |
| restricted → public | allowed | allowed — widening strands nobody |

An internal map's audience is every signed-in user, and both `internal` and
`public` reach all of them, so the two tie there. That tie is the whole reason
public → internal is safe on an internal map and not on a public one.

A private map is never flagged: it has no audience beyond its owner and grantees.

The caller keeps no gate of its own — that gate is what hid the internal case —
and the helper returns an empty list when the change strands nothing.

Renamed through the façade per the layering rule (`maps/service.py` import +
`__all__`, and the `test_layering` name set).

## Gap 1 — the block was invisible in the UI

#976 routed the 422 into a toast with an inline comment deferring the localized
message here. `error-map.ts` now matches the refusal and routes it to
`errors.datasetVisibilityBlockedByMaps`, interpolating the map names — which the
backend returns precisely so a human can go and fix them, and which the generic
422 fallback dropped. Key added to all four locales.

## Verification

```
  tests/test_datasets.py tests/test_maps.py tests/test_layering.py -q
  # 235 passed
cd frontend && npx vitest run src/lib src/components/dataset   # 1062 passed
```

`test_visibility_change_blocks_exactly_the_maps_it_strands` is parametrised over
eleven cells of the matrix, including the ones the old shape got wrong in each
direction and the moves that must stay allowed. The existing public-map tests are unchanged and still
pass, since the public rule did not move. LOC ratchets: `maps/service_public.py`
735 → 767, `datasets/domain/service_metadata.py` 480 → 487.

### DIFF
--- a/backend/app/modules/catalog/datasets/domain/service_metadata.py
+++ b/backend/app/modules/catalog/datasets/domain/service_metadata.py
@@ -159,15 +159,30 @@ async def _apply_visibility_change(
     dataset_id: uuid.UUID,
     new_visibility: str,
 ) -> bool:
-    """Set record.visibility, blocking public→restricted when public maps depend on it."""
-    if new_visibility != "public" and record.visibility == "public":
-        from app.modules.catalog.maps.service import find_public_maps_using_dataset
+    """Set record.visibility, blocking a change that would strand a shared map.
 
-        public_maps = await find_public_maps_using_dataset(session, dataset_id)
-        if public_maps:
-            raise ValueError(
-                f"Cannot restrict visibility: dataset is used in public maps: {', '.join(public_maps)}"
-            )
+    fix(#931): the gate used to be ``new != public and old == public``, matching
+    a query that only knew about public maps. An internal map using the dataset
+    was invisible to both, so the flip succeeded and every signed-in viewer of
+    that map silently lost the layer. The helper now compares the before and
+    after audiences itself and returns nothing when the change strands nothing,
+    so no gate is needed here.
+    """
+    from app.modules.catalog.maps.service import (
+        find_maps_broken_by_dataset_visibility,
+    )
+
+    broken_maps = await find_maps_broken_by_dataset_visibility(
+        session,
+        dataset_id,
+        old_visibility=record.visibility,
+        new_visibility=new_visibility,
+    )
+    if broken_maps:
+        raise ValueError(
+            "Cannot restrict visibility: dataset is used in shared maps: "
+            f"{', '.join(broken_maps)}"
+        )
     record.visibility = new_visibility
     return True
 
--- a/backend/app/modules/catalog/datasets/domain/service_metadata.py
+++ b/backend/app/modules/catalog/datasets/domain/service_metadata.py
@@ -177,6 +177,8 @@ async def _apply_visibility_change(
         dataset_id,
         old_visibility=record.visibility,
         new_visibility=new_visibility,
+        record_status=record.record_status,
+        owner_id=record.created_by,
     )
     if broken_maps:
         raise ValueError(
--- a/backend/app/modules/catalog/maps/service.py
+++ b/backend/app/modules/catalog/maps/service.py
@@ -29,7 +29,7 @@ from app.modules.catalog.maps.service_layers import (
 from app.modules.catalog.maps.service_public import (
     _validate_share_token,
     create_share_token,
-    find_public_maps_using_dataset,
+    find_maps_broken_by_dataset_visibility,
     get_active_share_token,
     get_maps_for_dataset,
     get_shared_map,
@@ -72,7 +72,7 @@ __all__ = [
     "remove_layer",
     "remove_layers_bulk",
     "validate_public_visibility",
-    "find_public_maps_using_dataset",
+    "find_maps_broken_by_dataset_visibility",
     "create_share_token",
     "update_share_token",
     "get_active_share_token",
--- a/backend/app/modules/catalog/maps/service_public.py
+++ b/backend/app/modules/catalog/maps/service_public.py
@@ -86,6 +86,23 @@ _SHARED_MAP_AUDIENCES = ("public", "internal")
 # this is a total order, and "does this change strand someone" is a comparison
 # rather than a table of allowed moves.
 #
+# fix(#931 codex r5): this encodes the COMMUNITY ladder, matching
+# `DefaultPermissionExtension`. An overlay that narrows a rung — ABAC excluding
+# some active non-owner users from `internal`, say — makes the real audience
+# smaller than the rank assumes, and this guard cannot see that: the seam's
+# `filter_visible`/`can_access_dataset` both take a CONCRETE user, and a map's
+# audience is a set with no representative member to pass them. Answering it
+# properly needs a new protocol method and an EXTENSION_API_VERSION bump, which
+# is a larger change than this guard.
+#
+# The failure direction is what makes that acceptable. Narrowing the real
+# audience only ever makes this guard MORE conservative: it can refuse a move
+# that strands nobody, never permit one that strands someone. So an overlay
+# inherits a refusal that lies, not a silent break — the same class as the
+# grant and publication-status cases above, and the reason those were worth
+# fixing rather than urgent. An overlay that reshapes the ladder should extend
+# this the way #930 notes a cloud overlay must scope internal per-tenant.
+#
 # fix(#931 codex r1): `restricted` is a PARTIAL audience, not an absent one.
 # Treating it as unreachable made `restricted -> private` look like it stranded
 # nobody, when it drops the grant holders who could render the layer.
--- a/backend/app/modules/catalog/maps/service_public.py
+++ b/backend/app/modules/catalog/maps/service_public.py
@@ -92,17 +92,41 @@ _SHARED_MAP_AUDIENCES = ("public", "internal")
 _DATASET_AUDIENCE_RANK = {"private": 0, "restricted": 1, "internal": 2, "public": 3}
 
 
-def _reach_rank(map_visibility: str, dataset_visibility: str) -> int:
+def _reach_rank(
+    map_visibility: str, dataset_visibility: str, *, restricted_has_grants: bool
+) -> int:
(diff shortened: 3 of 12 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
On the permission-overlay rank (`service_public.py`, `_reach_rank`) — recording this once so it is in the thread rather than re-argued each round.

**The finding is correct.** The rank encodes the community ladder, so an overlay that narrows a rung (ABAC excluding some active non-owner users from `internal`, say) makes the real audience smaller than the rank assumes, and this guard cannot see it.

**The suggested remedy is not implementable in place.** `PermissionExtension` (`protocols.py:395`) has exactly two methods, `filter_visible` and `can_access_dataset`, and both take a **concrete user**. A map's audience is a set with no representative member to pass them, so deriving it through the configured policy requires a new protocol method plus an `EXTENSION_API_VERSION` bump — a seam change, which should not ride inside a visibility fix reviewed for something else.

**Deferring it is safe because of the failure direction.** Narrowing the real audience only ever makes this guard *more* conservative: it can refuse a move that strands nobody, never permit one that strands someone. An overlay inherits a refusal that lies, not a silent hole — the same class as the grant-holder and publication-status cases fixed in this PR, which is why those were worth fixing and this one is not urgent.

Tracked as **#1068**, which carries the seam shape and this reasoning. The call site references it.

A re-raise of this finding is expected and is not new information; the remedy is deliberately deferred rather than the finding being open.

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1057, opened 2026-07-31
ops(analysis): promote the analysis statement timeouts to settings

ishiland:
`MATERIALIZE_TIMEOUT = "300s"` was a module constant carrying a comment that already said the quiet part: *"Hardcoded ceiling; promote to persistent-config if operators hit it."* 300 seconds covers roughly 150k to 600k buffered polygon rows, so an ordinary one-million-parcel buffer fails and the operator's only recourse was editing Python and rebuilding the image.

## What is promoted

| Setting | Was | Default |
|---|---|---|
| `ANALYSIS_MATERIALIZE_TIMEOUT_SECONDS` | `MATERIALIZE_TIMEOUT = "300s"` | 300 |
| `ANALYSIS_REGISTRATION_TIMEOUT_SECONDS` | `REGISTRATION_TIMEOUT = "600s"` | 600 |

**`REGISTRATION_TIMEOUT` is promoted with it**, per the issue's "decide it in the same pass". It exists because the commit that makes the output durable ends the CTAS transaction and its `SET LOCAL` with it (#692), so an operator who raises the CTAS ceiling for a large dataset needs to raise this one too. Promoting one and leaving the other hardcoded is exactly the inconsistency that costs someone an afternoon.

Both are read through a small function rather than bound at import. They render into SQL **and** into the user-facing "The analysis exceeded its 300s processing time limit" message, so a module-level snapshot would freeze whatever the settings object held when the module was first imported.

## The four surfaces

1. **`Settings` fields** in `backend/app/core/config.py`.
2. **Validation** — `gt=0`, which is load-bearing rather than tidiness. PostgreSQL reads `statement_timeout = '0'` as *no timeout at all* rather than erroring, so a zero accepted here would silently produce the unbounded statement the budget exists to prevent. Zero, negative, and unparseable values all fail at boot; asserted in `test_config.py`'s constraint table (six new rows), and verified to pass only because of the constraint by dropping `gt=0` and watching them fail.
3. **`.env.example`**, with a comment saying what the budget covers and, more usefully, **what raising it does not buy**: without an admission gate a longer budget does not make the work cheaper, it holds the job slot longer. Raising to 1200 turns "fails at 5 minutes" into "occupies the worker slot for 20 minutes and then maybe fails", which is worse for everything else in the queue.
4. **Docs site** — `docs.getgeolens.com` is a separate repository, so that entry cannot be made from here. The two keys and the `.env.example` prose are ready to copy across; flagging it rather than silently counting the surface as done.

Per-operation budgets remain the natural follow-up, not this change: a centroid needs seconds and a dissolve far more, so one scalar for all operations is a deliberate compromise. This ships the scalar.

## What is NOT included, and why

`DEFAULT_TIMEOUT_MS` (the sandbox preview budget) is the issue's third value. It lives in `backend/app/platform/sandbox/executor.py`, and threading `timeout_ms` through that module is **session G's #1011** in this wave. Editing the same symbol from two sides would put us both in a conflict for no benefit, so it is left for that change and **#1013 stays open** for it. This PR says `Refs #1013`, not `Closes`.

## Verification

```
cd backend && uv run pytest tests/test_config.py -q                 # 132 passed
cd backend && uv run pytest tests/test_analysis_materialize.py -q   # 80 passed
python3 scripts/check_env_doc_drift.py                              # 134 keys aligned
python3 scripts/check_docs_contract.py                              # passed
cd backend && uv run ruff check . && uv run ruff format --check .    # clean
```

No schema or API change.

Refs #1013

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -610,6 +610,33 @@ FRONTEND_PORT=8080
 # Type: integer (minimum 0) | Default: 30
 # INGEST_JOBS_RETENTION_DAYS=30
 
+# --- Analysis Budgets ---
+# Statement timeout for the analysis materialize CTAS — the CREATE TABLE AS that
+# builds a buffer/dissolve/clip/centroid output. The default covers roughly 150k
+# to 600k buffered polygon rows; raise it if a large dataset fails with "The
+# analysis exceeded its 300s processing time limit".
+#
+# What raising it costs: without an admission gate a longer budget does not make
+# the work cheaper, it holds the job slot longer. Raising this to 1200 turns
+# "fails at 5 minutes" into "occupies the worker slot for 20 minutes and then
+# maybe fails", which is worse for everything else in the queue. Raise it
+# together with WORKER_CONCURRENCY, or accept the queue wait.
+#
+# Zero is rejected at startup: PostgreSQL reads statement_timeout = '0' as no
+# timeout at all, which is exactly the unbounded statement this budget exists to
+# prevent.
+# Type: integer seconds (minimum 1) | Default: 300
+# ANALYSIS_MATERIALIZE_TIMEOUT_SECONDS=300
+
+# Statement timeout for the registration phase that follows the CTAS. The commit
+# that makes the output table durable ends the CTAS transaction and its budget
+# with it, so the full-scan metadata extraction (COUNT, ST_Extent, sample values)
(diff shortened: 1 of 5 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

Issue #952, opened 2026-07-30
fix(builder): a data-driven ramp on a mixed-geometry layer colors only the polygons


ishiland:
Split out of #919, which covered the missing fill-pattern image registration and shipped via #973.

Found in the 2026-07-30 style-tab audit. Pointers re-verified on main @ 9a788ba92.

## What happens

Apply a categorical or graduated color ramp to a mixed-geometry layer (`GEOMETRY` or `GEOMETRYCOLLECTION`). The polygons classify. The lines and points stay at their default colors. The legend shows the full classification as though it applied to every feature, so the map and the legend disagree and nothing in the UI explains why.

## Why

The ramp writes a single color property, chosen by geometry:

- `getColorProperty` (`frontend/src/lib/color-ramps.ts:309-319`) switches on `classifyGeometry`, which returns `'other'` for both generic sentinels (`frontend/src/components/builder/layer-adapters/shared.ts:27-33`). `'other'` falls to the `default` branch, so a mixed layer gets `fill-color`.
- The mixed adapter then filters paint per family before handing it to each sublayer: `mixedLinePaint` and `mixedPointPaint` (`frontend/src/components/builder/layer-adapters/mixed-adapter.ts:100-108`) run `filterPaintForLayerType` for `'line'` and `'circle'`, which drops `fill-color` entirely.

So the expression is authored once, for polygons, and the other two families never see it.

## Scope of this issue: the stopgap

Mixed-geometry layers come from created and sketch datasets (see the backend `_derive_created_geometry_type` contract), so a ramp on one is uncommon. The full multi-property fix is real work across every data-driven apply and clear path, and it is not scheduled. What ships here is the honest hint, so the legend mismatch is explicable instead of surprising.

**Deliverable:** when the layer is mixed, the data-driven section states which features the colors apply to, along the lines of "Colors apply to polygon features". Nothing about the paint behavior changes.

- The mixed check is the same `classifyGeometry(...) === 'other'` test `getColorProperty` already makes, so it does not need a new classifier.
- Render it as a muted hint in `DataDrivenStyleEditor`, matching the existing muted-hint styling in the style editor.
- New strings need all four locales (`en/es/fr/de`) in `frontend/src/i18n/locales/*/builder.json`. A `defaultValue` alone fails the `npm run test:i18n` locale-parity gate.

### Acceptance (stopgap)

- A ramp on a mixed-geometry layer leaves paint and rendering exactly as they are today, and the data-driven section states which features the colors apply to.
[...]

PR #1058, opened 2026-07-31
fix(builder): report the true category count and scope the mixed-geometry ramp

### LINKED ISSUE #960: fix(builder): the categorical style editor reports its 100-value page size as the category count
Split out of #920 item 3. **Re-scoped 2026-07-31**: the original premise was wrong, and the work left is frontend wiring, not a backend feature.

> **What the original body claimed.** That no true distinct-value count exists, that the largest N the frontend can compute is 10 (the ingest sample cap), and that a per-column distinct count therefore has to be built and stored at ingest time. All three are false. The count endpoint has existed since #315, it is exact rather than capped, and the frontend already has a hook for it. The categorical style editor just does not call it.

## What already exists

`GET /api/datasets/{id}/columns/{column_name}/stats/` (`backend/app/modules/catalog/datasets/api/router_metadata.py:321-351`) returns an exact, full-table distinct count for non-numeric columns. `get_column_stats` (`backend/app/modules/catalog/datasets/domain/column_stats.py:249-268`) runs `COUNT(DISTINCT col) ... WHERE col IS NOT NULL` over the whole table, with no sampling and no ceiling, and returns it as `distinct_count` with `data_type: "categorical"`. It landed in #315.

The field is on the wire and typed on both sides: `ColumnStatsResponse.distinct_count` at `backend/app/modules/catalog/datasets/domain/schemas.py:660` and `frontend/src/types/api.ts:1051`. The frontend hook exists too, `useColumnStats` at `frontend/src/hooks/use-maps.ts:345`.

The 10-value cap the original body pointed at is real but unrelated. `get_sample_values` (`backend/app/processing/ingest/metadata.py:575-648`, the `LIMIT 10` at `:648`) fills `example_values` on the attribute-metadata rows, which feeds AI metadata generation and the schema display. It never reaches the style editor.

## The actual gap

`DataDrivenStyleEditor` already imports `useColumnStats`, but gates it to graduated mode only (`frontend/src/components/builder/DataDrivenStyleEditor.tsx:381-384`, enabled by `columnForGraduated`). Categorical mode is served by `useColumnValues` alone (`:377-380`), which hits the sibling values endpoint (`router_metadata.py:266-318`) with `limit` defaulting to 100 and capped at 500.

So the category count the user sees is a page size, not a cardinality. `:1058` renders `t('dataDriven.categoriesWarning', { count: valuesData.values.length })`, gated at `:717` on `valuesData.values.length > 20`. On a column with 4,000 distinct values the warning reads "100 categories", and the swatch list silently stops at 100 with nothing saying it was truncated.

## Deliverable

Frontend only. No backend change, no migration, no OpenAPI or SDK regeneration.

1. Call `useColumnStats` in categorical mode as well, keyed on `columnForCategorical` rather than `columnForGraduated`. Both hooks then run for the selected column, which is one extra request per column selection, cached for 5 minutes by the existing `staleTime`.
2. Render the count honestly. When `statsData.distinct_count` exceeds the number of values returned, say so: the rendered swatches are a prefix of the column's categories, not all of them. When the two agree, the existing count is correct and the copy should not imply otherwise.
3. Keep the existing "consider graduated mode" advice, and let it trigger on the true count rather than the capped one.
4. New or reworded keys land in all four locales (en/es/fr/de). `dataDriven.categoriesWarning` is at `builder.json:526` in each. `npm run test:i18n` gates parity; a `defaultValue` alone fails it.
5. Extend `frontend/src/components/builder/__tests__/DataDrivenStyleEditor.test.tsx`, which already mocks both hooks, with a truncated case (values length at the cap, `distinct_count` well above it) and an exact case.

## Open decision: anonymous viewers get neither values nor counts

This is the one genuine backend question left, and it is deliberately not part of the deliverable above.

Both column endpoints require an authenticated user and then call `check_dataset_access`, not `check_dataset_access_or_anonymous` (`router_metadata.py:287` for values, `:341` for stats). A signed-out visitor looking at a public dataset gets a 401 from both, so any category UI that depends on them is empty for that visitor regardless of which one it calls.

Switching either endpoint to the anonymous-tolerant check is a Rule 1 access-control surface (see `backend/app/modules/catalog/authorization.py` and the security checklist in `AGENTS.md`). It widens what an unauthenticated caller can learn about a public dataset's contents, which is defensible for a public dataset but needs to be decided rather than assumed, and pinned by `backend/tests/test_rule1_structural.py`. Out of scope here unless it is decided and recorded on this issue first.

## Acceptance

- A column with 4,000 distinct values does not report "100 categories".
- Where the rendered swatch list is a truncated prefix, the UI says so.
- i18n parity, lint, typecheck, and Vitest pass.

ishiland:
Closes #960 and #952 (the stopgap it scopes itself to). One PR because both changes land in `DataDrivenStyleEditor` and would otherwise conflict on rebase.

## #960 — the count was a page size

The categorical editor called only `useColumnValues`, whose `limit` defaults to 100 and caps at 500, then rendered that length as the number of categories. On a column with 4,000 distinct values the warning read "100 categories" and the swatch list silently stopped at 100.

The re-scoped issue is right that nothing needed building: `GET /api/datasets/{id}/columns/{col}/stats/` has returned an exact, full-table, unsampled `distinct_count` since #315, it is typed on both sides, and `useColumnStats` was already imported here — just gated to graduated mode. So:

- the hook now runs for the categorical column too (one extra request per column selection, cached by the existing `staleTime`),
- the "consider graduated mode" advice triggers on the true count,
- when the rendered swatches are only a prefix, the copy says which prefix,
- when the page holds every category, the old wording stands.

Frontend only. No backend change, no migration, no OpenAPI or SDK regeneration. The anonymous-viewer question the issue records is untouched — both column endpoints still use `check_dataset_access`, and widening that is a Rule 1 decision, not a side effect of this.

## #952 — the stopgap, not the full fix

A ramp on a `GEOMETRY` / `GEOMETRYCOLLECTION` layer writes exactly one colour property, chosen by geometry. `classifyGeometry` returns `'other'` for both sentinels, `getColorProperty` falls through to `fill-color`, and `mixedLinePaint` / `mixedPointPaint` then run `filterPaintForLayerType` for `'line'` and `'circle'`, which drop that key. So the polygons classify, the lines and points keep their colours, and the legend shows the full classification as though it applied to every feature.

This ships the hint the issue scoped, in the same muted style as the section's other hints, gated on the same `classifyGeometry(...) === 'other'` test `getColorProperty` already makes. **No paint behaviour changes.** The full multi-property fix stays unscheduled — it has to return more than one property from `getColorProperty`, and all ten call sites assume exactly one, with the three clear paths being the awkward ones.

## Verification

```
# 366 files, 4298 tests pass

npx vitest run src/components/builder/__tests__/DataDrivenStyleEditor.test.tsx
# 34 tests (9 new)
```

New tests: the truncated case (page at the cap, `distinct_count` well above it), the exact case, the no-stats fallback, that the stats hook is called for the categorical column, and the mixed hint appearing for both generic sentinels and for neither Polygon, MultiLineString nor Point. Four new locale entries across en/es/fr/de.

### DIFF
--- a/frontend/src/components/builder/DataDrivenStyleEditor.tsx
+++ b/frontend/src/components/builder/DataDrivenStyleEditor.tsx
@@ -34,6 +34,7 @@ import {
   manualBreaks,
 } from '@/lib/classification';
 import { MAP_COLORS } from '@/lib/map-colors';
+import { classifyGeometry } from '@/components/builder/layer-adapters/shared';
 import { isNumericColumn } from '@/lib/column-utils';
 import type { MapLayerResponse, StyleConfig } from '@/types/api';
 
@@ -378,9 +379,14 @@ export function DataDrivenStyleEditor({
     columnForCategorical ? layer.dataset_id : undefined,
     columnForCategorical,
   );
+  // fix(#960): categorical mode needs the stats endpoint too. useColumnValues pages
+  // at 100 (500 max), so its length is a page size, not a cardinality — the warning
+  // read "100 categories" on a column with thousands and the swatch list stopped at
+  // 100 with nothing saying so. `distinct_count` is exact and full-table (#315).
+  const columnForStats = columnForGraduated ?? columnForCategorical;
   const { data: statsData } = useColumnStats(
-    columnForGraduated ? layer.dataset_id : undefined,
-    columnForGraduated,
+    columnForStats ? layer.dataset_id : undefined,
+    columnForStats,
   );
 
   // std-dev classification needs both mean and σ. The current stats endpoint
@@ -713,8 +719,18 @@ export function DataDrivenStyleEditor({
     [styleConfig, geomType, layerPaint, layerId, onStyleConfigChange, colorDebounceRef],
   );
 
+  // fix(#960): the honest counts. `values` is one page (100 by default, 500 max);
+  // `distinct_count` is the exact full-table cardinality. The advice triggers on the
+  // true count, and when the swatch list is only a prefix of it, the copy says so.
+  const shownCategories = valuesData?.values.length ?? 0;
+  const distinctCategories = mode === 'categorical' ? statsData?.distinct_count ?? null : null;
+  const totalCategories = distinctCategories ?? shownCategories;
+  const categoriesTruncated = distinctCategories != null && distinctCategories > shownCategories;
   const hasTooManyCategories =
-    mode === 'categorical' && valuesData && valuesData.values.length > 20;
+    mode === 'categorical' && !!valuesData && totalCategories > 20;
+  // fix(#952): the same `classifyGeometry(...) === 'other'` test getColorProperty
+  // already makes — GEOMETRY / GEOMETRYCOLLECTION, i.e. created and sketch datasets.
+  const isMixedGeometry = classifyGeometry(layer.dataset_geometry_type) === 'other';
   const noCompatibleColumns = filteredColumns.length === 0;
   const selectedColumnMissing = Boolean(column) && !columns.some((c) => c.name === column);
   const graduatedStatsUnavailable =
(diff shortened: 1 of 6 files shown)


ishiland:
@codex review

ishiland:
Right, and the better predicate was already sitting next to the one the issue named. `classifyGeometry(...) === 'other'` is correct for *which colour property gets written*, which is why `getColorProperty` uses it, but the hint makes a claim about **sublayers** — and `resolveAdapterType` only routes `GEOMETRY` and `GEOMETRYCOLLECTION` through the mixed adapter. A null or exotic type keeps the plain fill fallback and has no line or point sublayer to describe, so the sentence would have been false there.

Switched to `isGenericGeometryType`, and the negative test cases now include `CURVE` and the empty type alongside Polygon / MultiLineString / Point.

Lint, typecheck and the 36 DataDrivenStyleEditor tests re-run.

ishiland:
@codex review

PR #1059, opened 2026-07-31
feat(tiles): sign the raster tile template so API-key clients can render private rasters

### LINKED ISSUE #688: feat(tiles): signed raster tile URLs so API-key clients can render private rasters

ishiland:
Closes #688. **Leaving this for the coordinator to merge** (security-relevant).

> Sequenced after #957 (PR #1046), which touches the same file and removes the
> `raster-auth-check` decoy. Rebase if #1046 lands first; the two edit different
> regions.

## What was broken

`/api/tiles/token/{dataset_id}/` returned a signed template for **vector** tiles
and a bare URL for **raster** ones. `RasterTileToken` had no `sig`/`exp`/`scope`,
and `_resolve_raster_access` had three auth branches with no HMAC path — so a
client following the API contract literally received an **unauthenticated
template for a private raster**.

Third-party clients hit this: MapLibre issues the tile image requests itself and
attaches no `X-Api-Key`, so an API-key-only client could not render a private
raster at all. GeoLibre documents it as a known limitation. The workarounds were
`setTransformRequest` (what our own `DatasetMap` does, but not available to every
consumer) and `?api_key=` in the tile URL, which puts a non-expiring unscoped
credential into tile URLs, server logs, and any saved client project file.

## What changed

Mirrors the vector path on both sides.

**Mint** (`_build_tile_token_for_dataset`): `round_expiry()` for the 15-minute
TTL, `generate_tile_signature(scope, exp)`, and the triple is urlencoded into
`tile_url` as well as returned as fields, so a client can either use the template
as-is or rebuild it.

**Verify** (`_resolve_raster_access`): a new branch ahead of the
authenticated-user one, taken only when the caller presents a full
`sig`/`exp`/`scope` triple. Scope mismatch and signature failure are separate
403s, matching the vector path's shape.

## The scope

A raster dataset has no `table_name` to sign, so the resource string is the
**dataset id** — already unique per tenant, and what the tile URL keys on —
wrapped in `tenant_bound_scope` for the WR-03 binding (without it, a token
minted for tenant A is replayable in tenant B's context).

Both sites compute it through that one helper call, because a divergence between
mint and verify is a silent authorization bypass rather than a test failure. The
import stays inside the function on both sides, since `tenant_bound_scope` raises
when multi-tenant is active with no tenant in context.

## Blast radius

The branch is additive. An unsigned request for a private raster still 401s;
public rasters are untouched; the embed-token branch keeps its priority. The
query string survives the nginx hop — `frontend/nginx.conf` rewrites with
`$is_args$args` — so no nginx work was needed.

## Verification

```
  tests/test_raster_tiles.py tests/test_vrt_titiler.py \
  tests/test_raster_colormap_proxy.py tests/test_tenant_session_guc.py \
  tests/test_perf002_raster_meta_cache.py tests/test_fe_sdk_type_drift.py \
  tests/test_layering.py -q
  # 196 passed, 2 skipped
cd frontend && npx vitest run src/components/builder/__tests__/map-sync.raster.test.ts \
  src/hooks/__tests__/use-tile-token.test.ts   # 47 passed
```

`TestSignedRasterTileTemplate` covers the mint shape, a private raster rendering
with **no** `Authorization` and no `X-Api-Key`, and four refusals: unsigned still
401s, a tampered signature, a signature minted for a *different* dataset (the
check that makes the id-based scope meaningful — without it a token for any
readable raster would open every other one), an expired signature, and a
non-numeric `exp` that must 403 rather than 500.

Contract regenerated into `openapi.json`, `api.generated.ts` and both SDKs. The
hand-typed `api/tiles.ts` mirror marks the three fields optional, with the reason
at the site: `rasterTokenFromLayer` in `map-sync.ts` builds the same shape
locally from a saved layer row that has no signature, and the in-app map
authenticates with a bearer token instead. Threading the signature through the
builder was listed as optional in the issue and is not done here.

LOC ratchet: `processing/tiles/router.py` 2069 → 2142.


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1060, opened 2026-07-31
fix(collections): serve the collection extent bbox in the RFC 7946 spec form

### LINKED ISSUE #1006: fix(collections): rollup_span_bbox flattens a seam-crossing collection extent, so BBoxPreview draws a global band
## What happens

A collection containing a dataset whose extent straddles the antimeridian shows a globe-spanning extent preview. `CollectionCard` and `CollectionDetailPage` both draw a dashed band across the whole world for data that occupies a small box near Fiji.

This is the collections-path sibling of #1004, with the same cause and the same fix shape.

## Why

`backend/app/modules/catalog/collections/service.py:321` builds the collection's `extent_bbox` with `rollup_span_bbox`:

```python
"extent_bbox": rollup_span_bbox(row[1:7]),
```

`rollup_span_bbox` (`backend/app/core/geo.py:352-366`) collapses any crossing rollup to `[-180.0, south, 180.0, north]`. Its docstring is explicit that this is deliberate: "over-broad, never inverted", for consumers that feed `maxx - minx` span arithmetic or a viewer with no antimeridian handling. The in-source comment above the call site says it matches the sibling per-dataset `extent_bbox` in `datasets/domain/helpers.py`, "so both fields keep one contract".

The consumers here are not span-arithmetic consumers. `CollectionCard.tsx:17-18` and `CollectionDetailPage.tsx:74-75` pass the bbox straight to `BBoxPreview`, which already handles the crossing case: `BBoxPreview.tsx:75` calls `crossesAntimeridian` and `:133` draws the two halves via `splitBbox`. Both are #903 guards, and both are unreachable, because the flattened bbox is monotonic by the time it reaches them.

## Fix

Flip the producer to the spec form: `rollup_bbox` (`backend/app/core/geo.py:340-349`) is the RFC 7946 §5.2 sibling of `rollup_span_bbox`, returning `west > east` for an honest crossing. Change `collections/service.py:321` to use it, and update the comment above the call that ties this field to the per-dataset one.

Check the collection `extent_bbox` for any other consumer before flipping. The per-dataset field has two consumers with opposite needs (see the trap section in #1004), and the same audit is owed here.

## Verification

Use the same Fiji fixture as #1004: a dataset with points at `[179.5, -16.5]`, `[-179.5, -16.6]`, and `[178.44, -18.14]`. Add it to a collection and confirm the card and the detail page both draw two narrow bands at the seam rather than one global rectangle. Add a backend pin for the spec form alongside whatever pins the collections rollup today.

## Sequencing

Verify after #1004 lands, since #1004 settles the per-dataset half of the same decision and this one should follow its outcome.

Related: #1004, #903, #886.

ishiland:
Closes #1006. The collections sibling of #1004 (merged as #1040), verified after
it landed as the issue's sequencing asks.

## Why

`rollup_span_bbox` collapses a crossing rollup to `[-180, south, 180, north]`.
Its docstring is explicit that this is deliberate — "over-broad, never inverted"
— for consumers that feed `maxx - minx` span arithmetic or a viewer with no
antimeridian handling.

Neither consumer here is one of those. `CollectionCard.tsx:17-18` and
`CollectionDetailPage.tsx:74-75` pass the bbox straight to `BBoxPreview`, which
already handles the crossing case: `crossesAntimeridian` at `:75` and `splitBbox`
at `:133`, both #903 guards. Both were unreachable, because the flattened bbox is
monotonic by the time it arrives — and bit-identical to what a genuinely global
collection produces, so nothing client-side could tell them apart. A Fiji-shaped
collection drew a dashed band across the whole world.

## The consumer audit the issue asks for

The per-dataset `extent_bbox` has two consumers with opposite requirements (see
#1004's trap section), so the same audit is owed here. It comes out clean:
`CollectionCard` and `CollectionDetailPage` are the only two, both feed
`BBoxPreview`, and nothing on this path does span arithmetic. `router.py` passes
the value through; the four `default_extent` sites are all `None`. No STAC/OGC or
DCAT serializer reads this field — the standards surfaces build their own extents.

## Verification

```
  tests/test_collections.py tests/test_layering.py -q     # 62 passed
```

`test_seam_crossing_collection_keeps_the_spec_bbox` builds a collection from the
two halves of #1004's Fiji fixture (points at lon 179.5, -179.5, 178.44) and
asserts `west > east` with the exact pair, next to the existing
`test_collection_extent_computed` which pins the ordinary case.

No new frontend test: `BBoxPreview.test.tsx` already has "draws both seam halves
for an antimeridian-crossing extent" from #903. That test was passing all along
— this change is what lets a real collection reach it.

No response-shape change, so no contract regeneration.

### DIFF
--- a/backend/app/modules/catalog/collections/service.py
+++ b/backend/app/modules/catalog/collections/service.py
@@ -23,7 +23,7 @@ from sqlalchemy import delete, func, select
 from sqlalchemy.ext.asyncio import AsyncSession
 from sqlalchemy.orm import joinedload
 
-from app.core.geo import rollup_bbox_columns, rollup_span_bbox
+from app.core.geo import rollup_bbox, rollup_bbox_columns
 from app.core.identity import Identity
 from app.modules.catalog.authorization import apply_visibility_filter, get_user_roles
 from app.modules.catalog.collections.models import Collection, CollectionDataset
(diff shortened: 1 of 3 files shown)


ishiland:
@codex review

ishiland:
@codex review

PR #1061, opened 2026-07-31
perf(analysis): bound concurrent previews globally, not only per user

### LINKED ISSUE #1014: perf(analysis): bound concurrent previews globally, not only per user
Split out of #696 (item 4), which was a cluster of four unrelated resource-governance decisions. This one stands alone.

## The gap

Analysis previews are serialized **per user** and not at all globally. `run_analysis_preview` passes `concurrency_key=str(user_id)` into `execute_safe` (`backend/app/modules/catalog/datasets/domain/service_analysis.py:190-195`), and the sandbox takes a `pg_try_advisory_xact_lock` on `geolens:ai-sql:<user_id>` (`backend/app/platform/sandbox/executor.py:157-169`), failing fast with `query_busy` when the same user already has one running.

That stops one user from stacking previews. It does nothing about N different users. Each preview holds a dedicated connection from the engine pool for up to `DEFAULT_TIMEOUT_MS` (10 seconds, `executor.py:26`), and the pool is `db_pool_size = 10` plus `db_max_overflow = 3` (`backend/app/core/config.py:215`, `:217`), so thirteen slots.

`service_analysis.py:176-186` documents what happens when that runs out, and the numbers there were measured rather than estimated: before #716 each preview held two slots (the request session pinned since `get_dataset`, plus the sandbox connection), and at seven concurrent previews demand exceeded the pool. Every endpoint on that worker, not only analysis, then blocked for the 30-second `pool_timeout`, and the waiter's SQLAlchemy `TimeoutError` was classified by the sandbox as `query_failed` and surfaced as HTTP 500. #716 halved the cost to one slot per preview, which doubles the headroom without changing the shape of the failure.

## What to add

A small global bound around the `execute_safe` call in `run_analysis_preview`, sized well under thirteen so previews can never be the reason an unrelated endpoint waits on the pool.

**Prefer a global bound over re-keying the per-user lock to tenant.** Re-keying is the smaller diff and the wrong behaviour: in the single-tenant common case, which is the default deployment, every user is in one tenant, so colleagues would block each other for a resource that is not actually contended between them. The problem is total concurrent connections, so the bound should be on total concurrent connections.

Two details worth settling in the PR:

- **What happens at the bound.** Queueing and waiting turns a fast failure into a slow one, and the client is holding a request open. Failing fast with the existing `query_busy` shape is probably right, since the sandbox already has that path and the frontend already handles it, but it means the message has to distinguish "you already have one running" from "the server is at capacity" or the user gets a misleading explanation.
- **Scope of the bound.** An `asyncio.Semaphore` is per worker process, so with `UVICORN_WORKERS=2` the real ceiling is twice what the number says. That may be fine (each worker has its own pool), but it should be a deliberate choice stated in the comment rather than a surprise. A cross-worker bound would need another advisory lock keyed on a slot index, which is more machinery than this warrants unless the per-process reading turns out to be wrong.

## Acceptance

- [ ] A global bound limits concurrent `execute_safe` previews, with the chosen number justified against the thirteen-slot pool in a code comment.
- [ ] Behaviour at the bound is deliberate and the user-facing message distinguishes it from the per-user `query_busy` case.
- [ ] A test runs more concurrent previews than there are slots and asserts they serialize, or fail cleanly, rather than exhausting the pool. Asserting pool checkout counts (the way #716's measurement did) is stronger than asserting elapsed time and will not flake in CI.
- [ ] The per-user advisory lock stays as it is; this is an additional bound, not a replacement.

ishiland:
Analysis previews are serialized **per user** and not at all globally: `run_analysis_preview` passes `concurrency_key=str(user_id)` into `execute_safe`, and the sandbox takes a `pg_try_advisory_xact_lock` on that key. That stops one user stacking previews. It does nothing about N different users.

Each preview holds a dedicated connection from the engine pool for up to the sandbox timeout, and the pool is `db_pool_size` 10 + `db_max_overflow` 3 = 13 slots. The numbers in `service_analysis.py`'s own #716 note were **measured**: before that fix each preview held two slots, seven concurrent previews exhausted the pool, and every endpoint on the worker blocked for the 30-second `pool_timeout` with the waiter's SQLAlchemy `TimeoutError` classified as `query_failed` and surfaced as HTTP 500. #716 halved the cost to one slot, which doubles the headroom without changing the shape of the failure.

## The bound

A global `asyncio.Semaphore(4)` around the `execute_safe` call — well under thirteen, so previews can never be the reason an unrelated endpoint waits on the pool. The per-user lock is untouched; this is an additional bound, not a replacement.

**Global, not tenant-scoped.** Re-keying the per-user lock to the tenant is the smaller diff and the wrong behaviour: in the single-tenant common case every user is in one tenant, so colleagues would block each other for a resource that is not contended between them. What is contended is total concurrent connections, so that is what is bounded.

## The two details the issue asked to settle

**What happens at the bound.** It fails fast with the existing `query_busy` shape rather than queueing — the client is holding a request open, so waiting turns a fast failure into a slow one, and the frontend already handles that path. The message is its own: *"The server is running its maximum number of analysis previews"*. The per-user message would be a misleading explanation for a user whose first preview is being refused.

The `.locked()` check followed by `async with` looks like check-then-act but is atomic here: there is no `await` between them, and acquiring a semaphore that has a free slot does not yield to the loop. Said in a comment so it does not read as a bug.

**Scope of the bound.** Per worker process, deliberately. With `UVICORN_WORKERS=2` the deployment-wide ceiling is 8 — but each worker has its **own** pool of 13, so the ratio this protects (4 of 13) holds per pool, which is the thing that actually runs out. A cross-worker bound would need another advisory lock keyed on a slot index; more machinery than this warrants unless the per-process reading turns out to be wrong. Stated in the comment rather than left as a surprise.

## The test, and a correction to the acceptance

The acceptance suggested asserting pool checkout counts. That is not viable here: the test engine uses `NullPool`, so checkout numbers say nothing about the production `QueuePool` — the sibling #716 test documents exactly this and asserts `in_transaction()` for the same reason. This test asserts the **peak number of previews simultaneously inside `execute_safe`**, which is the pool-independent form of the same invariant and is precisely what determines how many pool slots previews can hold at once.

Eight concurrent previews are fired at a bound of four. `execute_safe` is replaced with a parked stand-in, so its per-user lock never runs and one user can stand in for many — that is the point, since the per-user lock is not what is under test.

**The first version of this test was flaky and is worth calling out.** It released the gate after a fixed number of event-loop turns; three runs in five failed because each request first authenticates, loads the dataset and checks quota — DB round-trips of unbounded duration — so slots drained before the late arrivals got there and they were refused by the per-user lock instead of by the bound. A counting `asyncio.Semaphore` subclass was the wrong fix too: CPython's `acquire()` consults `locked()` itself, so the counter double-counts. It now waits until every request has **settled** at the bound (parked, or already returned its 429) using only public state.

Six consecutive green runs with the fix; three consecutive failures (`8 previews hold slots, expected 4`) without it.

The refusals split by cause in the assertions, and that split is itself evidence the two messages are distinguishable rather than one doing double duty.

## Verification

```
cd backend && uv run pytest tests/test_analysis_preview.py -q      # 41 passed
cd backend && uv run pytest tests/test_layering.py -q              # 34 passed
cd backend && uv run ruff check . && uv run ruff format --check .  # clean
```

Closes #1014

### DIFF
--- a/backend/app/modules/catalog/datasets/api/router_analysis.py
+++ b/backend/app/modules/catalog/datasets/api/router_analysis.py
@@ -99,6 +99,9 @@ ANALYSIS_JOB_PRIORITY = -10
 # failures, which are server faults, not bad requests.
 _SANDBOX_STATUS = {
     "query_busy": status.HTTP_429_TOO_MANY_REQUESTS,
+    # fix(#1014): server-at-capacity is also a 429, but a distinct category so
+    # the message is not relabelled as the per-user one.
+    "query_at_capacity": status.HTTP_429_TOO_MANY_REQUESTS,
     "query_timeout": status.HTTP_422_UNPROCESSABLE_CONTENT,
     "query_data_error": status.HTTP_422_UNPROCESSABLE_CONTENT,
 }
--- a/backend/app/modules/catalog/datasets/domain/service_analysis.py
+++ b/backend/app/modules/catalog/datasets/domain/service_analysis.py
@@ -255,8 +255,12 @@ async def run_analysis_preview(
     # check-then-act: there is no await between them, and acquiring a semaphore
     # with a free slot does not yield to the loop.
     if _preview_slots.locked():
+        # Its own category, not query_busy: every consumer that maps a category
+        # to wording — the REST 429 detail and the AI chat ERROR_MESSAGES table
+        # — would otherwise relabel this as "your query is already running",
+        # which is false for a user whose first preview is being refused.
         raise SandboxError(
-            "query_busy",
+            "query_at_capacity",
             "The server is running its maximum number of analysis previews. "
             "Try again in a moment.",
         )
--- a/backend/app/modules/catalog/datasets/domain/service_analysis.py
+++ b/backend/app/modules/catalog/datasets/domain/service_analysis.py
@@ -40,6 +40,7 @@ from app.platform.sandbox.schemas import SandboxError
 
 PREVIEW_FEATURE_CAP = 500
 
+
 # fix(#1014): total concurrent previews, on top of the per-user lock.
 #
 # The per-user `pg_try_advisory_xact_lock` in execute_safe stops ONE user from
@@ -53,8 +54,15 @@ PREVIEW_FEATURE_CAP = 500
 # #716 halved the cost to one slot, which doubles the headroom without changing
 # the shape of the failure.
 #
-# Four, well under thirteen, so previews can never be the reason an unrelated
-# endpoint waits on the pool.
+# Derived from the configured pool, not hardcoded at four. The default pool is
+# 13 slots and a third of that is 4, which is the number the paragraph above is
+# reasoning about — but DB_POOL_SIZE and DB_MAX_OVERFLOW are operator knobs, and
+# a small local pool (DB_POOL_SIZE=2, DB_MAX_OVERFLOW=0) would have let a
+# hardcoded 4 admit twice as many previews as there are connections, which is
+# the exact failure this bound exists to prevent, just at a different scale.
+#
+# A third leaves two thirds for everything else, and the floor of 1 keeps a
+# one-slot pool working rather than deadlocking on a semaphore of zero.
 #
 # A GLOBAL bound, not a tenant-scoped one. Re-keying the per-user lock to the
 # tenant is the smaller diff and the wrong behaviour: in the single-tenant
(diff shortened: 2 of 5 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1065, opened 2026-07-31
test(notifications): make the secret-leak log assertions able to fail

### LINKED ISSUE #1064: test(notifications): the three secret-leak log assertions have never been able to fail
Found while sweeping for other assertions exposed to the failure class #1063 fixes.

## The finding

Three assertions in `backend/tests/test_notification_channels.py` guard against a credential reaching the logs:

- `:302` SMTP password, after `send_email`
- `:529` webhook secret, after `post_webhook`
- `:788` both, after `EnvConfiguredNotificationSink.deliver`

All three have the same shape:

```python
for record in caplog.records:
    assert _SECRET_PASSWORD not in record.getMessage()
```

**`caplog.records` is empty, so the loop body never runs and the assertion passes vacuously.** Measured:

- stdlib `logging` -> caplog: 1 record.
- structlog -> caplog: **0 records**, both before and after `setup_logging()`.

`logging_config.py` wires `LoggerFactory` + `ProcessorFormatter.wrap_for_formatter`, which should bridge to stdlib, but `setup_logging()` replaces the root handlers and caplog's handler goes with them.

## Why it matters more than a broken test

These are not failing. They are passing, have always passed, and would keep passing through the exact regression they exist to catch. Demonstrated end to end:

1. Added a structlog line to `smtp_channel.py` logging the password (the literal regression the test guards against).
2. Ran the assertion **verbatim as it exists today**: `CAPLOG_RECORDS: 0`, **test passed**.

A disconnected tripwire and a working one look identical from the outside, which is why this went unnoticed since the tests landed. This is not a regression — the gap has been open the whole time.

## Scope

Not a code defect. `smtp_channel.py` has no logger at all today, so nothing is leaking. The problem is that the guard against a future logger cannot fire, and the codebase uses structlog everywhere, so the future logger would be exactly the kind caplog cannot see.

## Fix

Assert against real emitted output rather than caplog, following `test_phase_273_log_redaction.py`, and assert **both** directions: that the log channel is live, and that no secret appears in it. The liveness half is the one that was missing, and its absence is precisely what made the original vacuous.

A recorder — the #1063 approach — does not work here: a recorder can only observe a logger that already exists, and this tripwire has to catch one nobody has written yet.

## Not in scope

The sweep also found `test_reupload_swap_lock_retry.py:207-209` (negative assertion over `capture_logs`, not currently vacuous) and seven presence-asserting sites that fail loudly if blinded. Both judged not worth a PR.

ishiland:
**Not a regression fix.** These three assertions have never been able to fail.
The gap has been open since the tests landed, and it went unnoticed because a
disconnected tripwire and a working one look identical from the outside.

## What was wrong

Three assertions guard a credential against reaching the logs — the SMTP
password after `send_email`, the webhook secret after `post_webhook`, and both
after `EnvConfiguredNotificationSink.deliver`. All three:

```python
for record in caplog.records:
    assert _SECRET_PASSWORD not in record.getMessage()
```

`caplog.records` is empty, so the loop body never runs and the assertion passes
vacuously. Measured, not inferred:

| emitter | records caplog sees |
|---|---|
| stdlib `logging` | 1 |
| structlog, before `setup_logging()` | **0** |
| structlog, after `setup_logging()` | **0** |

`logging_config.py` wires `LoggerFactory` + `ProcessorFormatter.wrap_for_formatter`,
which should bridge to stdlib — but `setup_logging()` replaces the root handlers
and caplog's handler goes with them.

## Demonstrated in both directions

Added a structlog line to `smtp_channel.py` logging the password, which is the
literal regression these tests guard against:

- **Old assertion, verbatim:** `CAPLOG_RECORDS: 0` — **passed**.
- **New assertion:** failed with
  `secret leaked into log output: 's3cr3t-smtp-pass' in ['{"event": "smtp auth as s3cr3t-smtp-pass", "logger": "app.platform.notifications.smtp_channel", ...}']`

Nothing is leaking today — `smtp_channel.py` has no logger at all. The problem
is that the guard against a future one cannot fire, and the codebase uses
structlog everywhere, so the future logger is exactly the kind caplog cannot see.

## The fix

Each site reads real emitted output and asserts **both** directions: that a
marker sent through the same structlog path arrives, and that no secret appears
in any record that did. The liveness half is the one that was missing, and its
absence is precisely what made the original vacuous — the same refusal-versus-
admission point as #1063, one level up.

Real output rather than a recorder, deliberately. A recorder — #1063's approach
— can only observe a logger that already exists, and `smtp_channel.py` has
none. This tripwire has to catch a logger nobody has written yet.

## Two things the construction has to get right

Both found by measurement, both non-obvious, both would have silently
reintroduced the vacuity:

**`setup_logging()` runs inside the test body, not a fixture.** pytest swaps
`sys.stdout`/`stderr` for the call phase only, so a handler created during
fixture setup binds to a stream that is not the captured one. Every write then
fails with `--- Logging error ---` and lands nowhere assertable. The first
version used a fixture and the liveness check caught it.

**Only JSON records count, not the whole stream.** pytest writes tracebacks to
the same place, and a traceback carrying a secret is a different finding from a
log carrying one — an earlier version failed on exactly that false positive. A
traceback rendered INTO a record is still caught, because `format_exc_info`
puts it inside the JSON.

Root handlers and the structlog config are restored on exit, since leaking
global logging state into later tests is the same failure class this closes.

## Scope

Test-only, one file, three call sites. No application code touched.

Found by the sweep for other assertions exposed to #1063's failure class. That
sweep also found `test_reupload_swap_lock_retry.py:207-209` (negative assertion
over `capture_logs`, not currently vacuous) and seven presence-asserting sites
that fail loudly if blinded — both judged not worth a PR.

## Verification

```
cd backend && uv run pytest tests/test_notification_channels.py \
  tests/test_phase_273_log_redaction.py tests/test_layering.py -q -n 4
# 72 passed
```

CI will be red on the sec025 EXISTS tests until #1063 lands; that is main's
current state, unrelated to this diff.

Closes #1064

### DIFF
--- a/backend/tests/_logging_state.py
+++ b/backend/tests/_logging_state.py
@@ -1,0 +1,65 @@
+"""Save and restore everything ``setup_logging()`` mutates.
+
+fix(#1064 codex r2): ``setup_logging()`` does not only touch the root logger.
+Reading ``app/core/logging_config.py:103-112`` line by line, it clears root's
+handlers, then clears handlers and rewrites ``propagate`` on ``uvicorn``,
+``uvicorn.error`` and ``uvicorn.access``. A teardown that restores only root
+leaves those three changed — measured, a caller's ``uvicorn.access`` goes in
+with ``propagate=True`` and comes back out with ``propagate=False``.
+
+Two earlier rounds of this same mistake are why the list is derived from the
+source rather than from memory. The first restored root but not the structlog
+processor chain, which silently dropped ``_redact_sensitive_fields``. The
+second restored the chain but not these three loggers. The rule: to undo a
+function, read what it actually mutates and restore that list — not what its
+name suggests, and not the one thing that broke last time.
+
+Restoring what was there, rather than resetting to a library default, is the
+point. A reset still changes state the caller established, so it trades one
+order-dependence for another and can strip app behaviour a later test relies
+on. Repairing a leak that was already there when the test arrived is a
+different job, and belongs to the conftest guard in #1066.
+"""
+
+from __future__ import annotations
+
+import contextlib
+import logging
+from collections.abc import Iterator
+
+import structlog
+
+# Every logger setup_logging() reaches, from logging_config.py:103-112.
+# "" is the root logger.
+_MUTATED_LOGGERS = ("", "uvicorn", "uvicorn.error", "uvicorn.access")
+
+
+@contextlib.contextmanager
+def preserved_logging_state() -> Iterator[None]:
+    """Restore the logging state ``setup_logging()`` changes, exactly as found.
+
+    Covers the stdlib side (handlers, ``propagate`` and level, for root and
+    each named logger) and the structlog side (the whole config, including the
+    processor chain). Callers still have to invoke ``setup_logging()`` inside
+    the test body themselves — pytest swaps ``sys.stdout``/``stderr`` for the
+    call phase only, so a handler created during fixture setup binds to a
+    stream capsys is not capturing.
+    """
+    saved_loggers = {
+        name: (
+            logging.getLogger(name).handlers[:],
+            logging.getLogger(name).propagate,
+            logging.getLogger(name).level,
+        )
+        for name in _MUTATED_LOGGERS
+    }
+    saved_structlog = dict(structlog.get_config())
+    try:
+        yield
+    finally:
+        structlog.configure(**saved_structlog)
+        for name, (handlers, propagate, level) in saved_loggers.items():
+            logger = logging.getLogger(name)
+            logger.handlers[:] = handlers
+            logger.propagate = propagate
+            logger.setLevel(level)
--- a/backend/tests/_logging_state.py
+++ b/backend/tests/_logging_state.py
@@ -29,6 +29,8 @@ from collections.abc import Iterator
 
 import structlog
 
+from app.core.logging_config import setup_logging
+
 # Every logger setup_logging() reaches, from logging_config.py:103-112.
 # "" is the root logger.
 _MUTATED_LOGGERS = ("", "uvicorn", "uvicorn.error", "uvicorn.access")
@@ -63,3 +65,39 @@ def preserved_logging_state() -> Iterator[None]:
             logger.handlers[:] = handlers
             logger.propagate = propagate
             logger.setLevel(level)
+
+
+@contextlib.contextmanager
+def configured_logging(
+    *, json_logs: bool = True, log_level: str = "DEBUG"
+) -> Iterator[None]:
+    """``setup_logging()`` inside a preserved window, with the freeze disarmed.
+
+    fix(#1064 codex r4): use this rather than ``preserved_logging_state()``
+    directly whenever the window CALLS ``setup_logging()``. Restoring the
+    config on exit is not enough on its own, because ``setup_logging()`` turns
+    ``cache_logger_on_first_use`` ON, and a module-level logger that emits
+    while it is on freezes its ``BoundLoggerLazyProxy`` against the
+    then-current chain. That bind lives on the proxy object, not in the config,
+    so no amount of restoring undoes it — the logger is simply invisible to
+    every later ``capture_logs()`` in the worker. It is the #1063 mechanism.
+
+    **The order is the entire reason this is a separate helper.** The disable
+    has to land AFTER ``setup_logging()``; ``preserved_logging_state()`` cannot
+    do the job by configuring before it yields, because ``setup_logging()``
+    would immediately turn caching back on. A caller composing the two by hand
+    gets that wrong silently — nothing fails, the freeze just happens.
+
+    Turning caching off keeps the processor chain and the stdlib routing
+    ``setup_logging()`` installed; only the freezing goes. The original value
+    is restored on exit like everything else.
+
+    Enter it INSIDE the test body, never in fixture setup: pytest swaps
+    ``sys.stdout``/``sys.stderr`` for the call phase only, so a handler built
+    during setup binds to a stream ``capsys`` is not capturing and every write
+    fails with ``--- Logging error ---`` instead of landing anywhere assertable.
+    """
+    with preserved_logging_state():
+        setup_logging(json_logs=json_logs, log_level=log_level)
+        structlog.configure(cache_logger_on_first_use=False)
+        yield
--- a/backend/tests/test_notification_channels.py
+++ b/backend/tests/test_notification_channels.py
@@ -81,18 +81,30 @@ def _live_logs(capsys: pytest.CaptureFixture[str]) -> Any:
     a stream that is not the one being captured and every write fails with
     ``--- Logging error ---`` instead of landing anywhere assertable.
 
-    On exit the root handlers and level are restored, and structlog is reset
-    to LIBRARY defaults rather than to whatever was configured before. That is
-    deliberate and it is not the same thing: `reset_defaults()` leaves
-    `cache_logger_on_first_use` False, which is the safe end of the range,
-    whereas faithfully restoring a prior True would hand the next test the
-    exact global that makes capture go blind. Leaking that state onward is the
-    failure class this construction exists to close.
+    On exit the root handlers, the root level and the structlog config are all
+    put back exactly as they were.
+
+    fix(#1064 codex r1): an earlier version called `structlog.reset_defaults()`
+    instead, reasoning that the library default leaves
+    `cache_logger_on_first_use` False and is therefore the safe end. That was
+    wrong, and wrong in a security-relevant direction: `reset_defaults()`
+    replaces the WHOLE chain, so it drops `_redact_sensitive_fields` and the
+    stdlib routing `setup_logging` installed, and any later test on the worker
+    would log unredacted. Restoring what was actually there is the only
+    version that leaves no trace — resetting to a default is still changing
+    state the caller established.
+
+    Captured at DEBUG, not INFO (fix(#1064 codex r1)): a channel logging a
+    revealed credential at debug level would otherwise be filtered out before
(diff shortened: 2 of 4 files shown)


ishiland:
@codex review

ishiland:
Both points applied at `22e702cd4`.

**The redaction test's leak was real, and I measured it rather than assuming.** Ran a probe immediately after it in the same process, reading `structlog.get_config()["cache_logger_on_first_use"]`:

- before: **True** — left on, nothing put it back
- after: **False**

So every test following it in that worker inherited replaced root handlers and the caching flag. That is the leaked-global class this file's own subject matter is about, and the caching half is exactly what makes a later `capture_logs()` see nothing.

The `setup_logging()` call stays in the test body — a fixture-created handler binds to a stream capsys is not capturing, which is the trap the liveness check caught in this PR's first draft. Only the save and restore move into a fixture. Ten lines with the docstring.

**The docstring wording is corrected too, and the distinction is worth more than the word.** I wrote "restored"; `reset_defaults()` resets to LIBRARY defaults, which is not the same thing. It matters in a specific direction: the library default leaves `cache_logger_on_first_use` **False**, the safe end, whereas faithfully restoring a prior **True** would hand the next test the exact global that makes capture go blind. So the imprecise word was describing something strictly better than what it claimed, and now says so.

To be clear about scope, since it would be easy to over-read: this is not the #1063 trigger. That exact pairing was tested and captured fine, so it does not reproduce. It is worth fixing because an un-restored global is a hazard on its own.

Verification: 178 passed across `test_notification_channels.py`, `test_phase_273_log_redaction.py`, `test_sec025_sandbox_allowlist.py`, `test_layering.py`, plus ruff.

Not arming.

ishiland:
Both P2s fixed at `f2f5962ce`. **You were right on the second one and my reasoning was wrong in a security-relevant direction**, which is worth saying plainly rather than burying in a changelog.

**Restore the previous structlog configuration.** I argued that `reset_defaults()` was fine because the library default leaves `cache_logger_on_first_use` False, the safe end of the range. I only looked at that one key. `reset_defaults()` replaces the **whole** processor chain, so it drops `_redact_sensitive_fields` and the stdlib routing `setup_logging` installed — every later test on the worker would log unredacted.

Measured, with an earlier test establishing the app chain so there is something real to destroy:

| teardown | redaction processor present after? |
|---|---|
| `structlog.configure(**saved)` | **True** |
| `structlog.reset_defaults()` | **False** |

Visible in the log rendering too: JSON with the restore, ConsoleRenderer without. Now saves `dict(structlog.get_config())` and reapplies it.

The general form is the part I had backwards: resetting to a default is still changing state the caller established. Only saving and reapplying leaves no trace.

**Capture DEBUG-level secret leaks.** Also right, and it regressed reach the old assertions had via `caplog.at_level(logging.DEBUG)`. Injected a debug-level leak into `send_email` and measured both ways:

- at `INFO`: **passed**, reported nothing — exactly the blind spot you describe
- at `DEBUG`: **failed**, naming the record

Both fixes also applied to the redaction test's fixture, which had inherited the `reset_defaults()` pattern from here.

One process note. While running the second RED check I round-tripped the file with a pattern replace, and the restore hit the **docstring's** mention of `reset_defaults()` instead of the code — leaving the prose wrong and the teardown unfixed. The pre-commit ruff hook caught it as an unused `saved_structlog`; I repaired it by line number rather than by pattern and re-verified. Nothing wrong shipped, but it is the third time today a pattern-based edit hit a lookalike occurrence, so: repair by position when the pattern appears in prose too.

Verification: 72 passed across `test_notification_channels.py`, `test_phase_273_log_redaction.py`, `test_layering.py`, plus ruff.

@codex review

ishiland:
Fixed at `cc58f3f4f`. The finding is right and I verified it before fixing rather than after.

`logging_config.py:103-112` clears root's handlers, then clears handlers and rewrites `propagate` on `uvicorn`, `uvicorn.error` and `uvicorn.access`. With a caller that had configured `uvicorn.access` first:

| `uvicorn.access` | handlers | propagate |
|---|---|---|
| before | 1 | `True` |
| after, previously | 2 | **`False`** |
| after, now | 1 | `True` |

**This is the third round of one mistake, so the fix records the rule rather than just the patch.** First I restored root but not the structlog processor chain, silently dropping `_redact_sensitive_fields`. Then I restored the chain but not these three loggers. Each time I restored what I assumed `setup_logging()` touches instead of what it does. The helper now says: to undo a function, read what it actually mutates and restore that list — not what its name suggests, and not the one thing that broke last time.

Both call sites delegate to `preserved_logging_state()` in `tests/_logging_state.py`, saving handlers, propagate and level for root and each named logger plus the whole structlog config. One shape rather than two copies to get wrong.

**Re-verified the two properties the refactor could have broken**, since a refactor that quietly undoes an earlier fix is exactly the shape of this PR's subject matter:

- redaction processor still survives teardown when an earlier test established the app chain — `REDACTOR_SURVIVES: True`
- the injected debug-level leak still fails, naming the record

72 passed across `test_notification_channels.py`, `test_phase_273_log_redaction.py`, `test_layering.py`, plus ruff.

Not armed, not update-branched.

@codex review

ishiland:
Fixed at `73f67cc62`, and **this one reaches past the PR** — I verified it before fixing.

`setup_logging()` turns `cache_logger_on_first_use` ON. Any module-level logger that emits while it is on freezes its `BoundLoggerLazyProxy` against the then-current chain, and restoring the config cannot undo a proxy-local bind. Measured against `app.platform.notifications.env_sink.logger`:

| | records `capture_logs()` sees |
|---|---|
| baseline | 1 |
| after the helper ran, previously | **0** |
| after the helper ran, now | 1 |

Restoring handlers, propagate, level and the whole structlog config does not help, because the damage is on the proxy object rather than in the config.

**This is the #1063 mechanism, now reproduced.** #1063's docstring says the trigger was never pinned down — two reproduction attempts failed, and the fix was justified as removing a dependency class rather than targeting an identified cause. The trigger is: something enables caching, and a module-level logger emits while it is on. My helper did exactly that, so it was manufacturing the hazard the PR exists to catch, and would have seeded it into whatever ran after it.

The fix keeps the chain and the routing and drops only the freezing:

```python
setup_logging(json_logs=True, log_level="DEBUG")
structlog.configure(cache_logger_on_first_use=False)
```

Proxies frozen before this helper runs are not its business — repairing a pre-existing leak belongs to #1066.

**Re-verified the three properties earlier rounds established**, since this sits in the same teardown and a regression here would be the same shape as the bug: redaction processor survives (`REDACTOR: True`), `uvicorn.access.propagate` restored to `True`, the caller's caching setting restored, and the injected debug-level leak still fails naming the record.

72 passed plus ruff. Not armed, not update-branched.

@codex review

ishiland:
Added the #1063 docstring correction at the head above — deliberate scope addition, not creep, and it is the reason the finding is worth more than the fix.

That docstring told a reader the mechanism was never pinned down and to treat the caching story as unconfirmed. It is confirmed now, and leaving it would have sent the next person down a path already walked five times.

Written as the conjunction, because that is the searchable form:

> `cache_logger_on_first_use=True` is in force **AND** a module-level logger EMITS during that window.

The emit is the half everyone missed, mine included. Enabling caching alone does nothing — the proxy only freezes on first use, and at that point it binds to the chain in force and caches the result **on the proxy object**, which is why reapplying a config afterwards cannot undo it.

#1063's own fix is unchanged. Removing the dependency outright is stronger than avoiding one known trigger, and it was right when the trigger was unknown; knowing the mechanism does not make a narrower fix preferable. The docstring now says that explicitly, so nobody reads the correction as an invitation to narrow it.

Also recorded there, because it is not obvious and #1066 now carries it: a snapshot-and-restore guard **cannot** repair a proxy frozen while it was off-guard. The cache lives on the proxy, not in the config. Preventing the freeze is the only thing that closes the class.

178 passed across `test_sec025_sandbox_allowlist.py`, `test_notification_channels.py`, `test_phase_273_log_redaction.py`, `test_layering.py`, plus ruff.

@codex review

ishiland:
Fixed at `f9d135935`. The finding is right, and checking it before fixing turned up a live exposure rather than only an inaccurate sentence.

The docstring did overstate. `preserved_logging_state()` snapshots and restores, nothing more. Only `_live_logs()` set `cache_logger_on_first_use=False`, so a reader following that guidance would have wrapped `setup_logging()` in the guard and still frozen a proxy permanently.

**The first of the two suggested remedies does not work as written, and the reason is the useful part.** Putting the disable inside `preserved_logging_state()` cannot help, because that guard configures before it yields and the caller's `setup_logging()` then turns caching straight back on. The disable has to land *after* that call. That ordering is the whole trap: composing the two by hand fails silently, since nothing raises and the freeze just happens.

So the guard now has a sibling that owns both steps in the order that works:

```python
with preserved_logging_state():
    setup_logging(json_logs=json_logs, log_level=log_level)
    structlog.configure(cache_logger_on_first_use=False)
    yield
```

Both call sites use `configured_logging()`. I chose that over narrowing the docstring because narrowing would have documented around something real. `test_phase_273_log_redaction.py` called `setup_logging()` inside a snapshot-only window, which is precisely the composition described, and it was one emitting import away from seeding the freeze it exists to document.

Measured in two clean processes, one per composition:

| composition | records `capture_logs()` sees before | after |
|---|---|---|
| `preserved_logging_state()` + `setup_logging()` by hand | 1 | **0** |
| `configured_logging()` | 1 | **1** |

Re-verified the restore properties, since this refactor moves the r1 and r2 fixes. Across the window the structlog config compares identical, all four stdlib loggers compare identical, `cache_logger_on_first_use` goes back to the caller's value rather than the helper's, and `_redact_sensitive_fields` is still in the chain.

178 passed across the four files, plus ruff.

@codex review

PR #1067, opened 2026-07-31
test: integration check for a 13-PR batch (do not merge)

ishiland:
## Do not merge this PR

It exists to run CI **once** over the combination of thirteen PRs that are individually ready, so they can then be merged without each one paying a full ~15-minute update-branch cycle. Total serial cost avoided is roughly three hours.

The individual PRs are what get merged. This branch is deleted afterwards.

## What is in it

Each of these is codex-clean at its exact head, has zero failing and zero pending checks, and is BEHIND only because main has moved:

#1028, #1033, #1036, #1039, #1043, #1046, #1047, #1049, #1050, #1053, #1058, #1060, #1062

Merged onto `main` at the post-#1055 tip. All thirteen merged with **zero conflicts**.

## Why the combination needs testing at all

Each PR passed CI against an older main. Nothing has tested them together, and that is exactly the gap that broke main earlier today: #1024 passed its own PR and failed on main, then two consecutive merges cancelled each other's push runs so nothing showed red for 54 minutes. Three PRs inherited the breakage and each read as its author's bug.

This run is the missing check. It is what a merge queue does automatically, done by hand because the queue is not enabled yet.

## Seam checks already done locally

- `alembic heads` → single head `0032_api_key_scope`. Only #1033 touches alembic, and it edits `0029` in place without adding a revision.
- Only #1046 touches the regen artifacts (`openapi.json`, `api.generated.ts`, `sdks/`), so no two PRs collide there.
- `ruff check` and `ruff format --check` clean across 864 files.

## What happens next

Green: the thirteen are merged individually, each keeping its own squash commit and `Closes #NNN` linkage, and this branch is deleted.

Red: the failure identifies which combination is bad. The individual PRs stay unmerged and are bisected rather than landed on a guess.


ishiland:
Green across all checks at `d1b2fe47f`, so the thirteen were merged individually. Each kept its own squash commit and `Closes #NNN` linkage. Closing this and deleting the branch.

Issue #1068, opened 2026-07-31
feat(extensions): let the permission seam answer "who can see this?" for audience checks


ishiland:
## Summary

`PermissionExtension` (`backend/app/platform/extensions/protocols.py:395`) exposes exactly two methods, and both take a **concrete user**:

- `filter_visible(stmt, user, user_roles, record_cls, grant_cls)`
- `can_access_dataset(db, dataset, dataset_id, user, *, user_roles)`

Neither can answer "who is the audience for this map?", because an audience is a set with no representative member to pass in. Any guard that needs to reason about a whole audience therefore has to reconstruct it from the community ladder directly, which an overlay is free to narrow.

## Where it bites today

#931 / #1056 adds a guard that refuses a dataset visibility change which would strand a map's audience. It computes that audience from the community rank. A permission overlay that narrows a rung makes the true audience **smaller** than the guard assumes.

**The failure direction is safe, which is why this is filed rather than fixed.** A smaller real audience only makes the guard more conservative: it can refuse a move that strands nobody, never permit one that strands someone. So an overlay inherits a refusal that lies, not a silent hole. Same class as the grant and publication-status cases already handled in that PR, and #930 carries a comparable note that a cloud overlay must scope `internal` per-tenant.

## Why it was not fixed in place

The remedy needs a new protocol method plus an `EXTENSION_API_VERSION` bump — a seam change. Doing that inside a security-relevant PR mid-wave would smuggle an extension-API change into a review aimed at something else. It wants its own scope and its own review.

## Rough shape

A method that returns the audience predicate rather than a per-user verdict — enough for a caller to ask "which principals can read this record?" without inventing a representative user. Whatever the signature, it lands with an `EXTENSION_API_VERSION` bump and the OSS default implementation mirroring the current community ladder exactly, so behaviour is unchanged until an overlay opts to narrow it.

## Note for reviewers of #1056

Codex is likely to re-raise this each round, since the behaviour is documented rather than changed. It is a correct finding with a deliberately deferred remedy, recorded at the call site with the direction-of-failure reasoning. A re-raise is not new information.

ishiland:
## The direction-of-failure argument in the body no longer holds as written

The body above argues this is safe to defer because the guard reads the community ladder and an overlay can only narrow it, so the worst case is a refusal that lies. That was accurate against the rank-only guard it was written for. #1056 has since replaced that logic and the argument needs restating.

[...]

Issue #683, opened 2026-07-25
feat(ai): chat run_analysis should support clip-by-layer


ishiland:
The chat `run_analysis` tool only takes an operation enum (buffer/centroid) and a distance, yet two-layer questions are the most conversational analysis asks ("clip the buildings to the flood zone"). With #682, the backend supports `mask_dataset_id` end to end (Rule-1 access checks on both datasets, polygon-only masks), so the chat gap is now tool-schema plumbing plus one port widening:

- Add `operation: "clip"` + a `mask_layer_id` input to the tool schema (resolve it against the map's layer list the same way `layer_id` is resolved, with its own access check — Rule 1 applies to both datasets).
- Widen `ProcessingPort.run_analysis_preview` so a mask dataset can reach the service (see the seam section below).
- Reuse the existing preview path (`run_analysis_preview(..., mask_dataset=...)`); the ephemeral-overlay/action collector already carries everything needed.
- Extend the #675 "Save as dataset" handoff so a chat clip preview can prefill the Analysis panel (operation + source + mask layer).

Dissolve stays materialize-only and is a separate conversation.

## Where to change what

Verified against `main` at 9a788ba92:

- **The enum to widen** is at `backend/app/processing/ai/tools.py:357`, which reads `"enum": ["buffer", "centroid"]`, inside the `run_analysis` tool schema whose `properties` block runs `:346-370`. There is no `mask_layer_id` property yet.
- **The comment guarding that enum is itself now stale**, and is the thing to update alongside it. `tools.py:353-356` justifies the narrowing with "`clip` needs a drawn mask polygon that only the Analysis rail panel can supply." #682 shipped the layer-mask alternative, so that rationale no longer holds. Leaving it in place will mislead the next reader into thinking the narrowing is still deliberate.
- **The layer-resolution pattern to copy** is `chat_analysis.py:71`: `next((lyr for lyr in layers if lyr.id == tool_input.get("layer_id")), None)`.
- The `mask_dataset_id` rails to reuse: `backend/app/modules/catalog/datasets/domain/schemas.py:900-906` (request field), `router_analysis.py:150-154` (preview wiring) and `:251-254` (materialize validation), `_load_mask_dataset` at `router_analysis.py:110-134` (the access + polygonal + size checks the REST route applies).
[...]

Issue #675, opened 2026-07-24
feat(analysis): hand off a chat analysis preview to the Analysis panel for saving


ishiland:
## Context

#674 added the `run_analysis` AI-chat tool: asking the assistant to "buffer the schools by 500 m" draws a temporary preview on the map. The preview is ephemeral by design — materializing a dataset is the Analysis rail panel's job (#671).

Today there is **no handoff between the two**. A user who likes the chat preview has to open the Analysis panel and re-select the layer, the operation, and the parameters by hand, then hit "Save as dataset".

This friction is why #674 deliberately removed the "you can save it from the Analysis panel" line from the tool result — the sentence implied a handoff that does not exist, and it was doubly wrong for view-only users, who have no Analysis rail at all (`BuilderRail` is rendered only by `MapBuilderPage`).

## Proposal

When a chat `run_analysis` preview is on the map, offer a one-click path into the Analysis panel with the operation and parameters prefilled.

Sketch:
- The chat action already carries everything needed to reconstruct the request (`operation`, `layer_id`, and the distance). It is currently dropped on the floor by the collector, which emits only `geojson`/`bbox`.
- Either surface a "Save as dataset" affordance on `EphemeralBadge`, or have the chat message expose an action that opens the `analysis` rail panel prefilled.
- Only for callers who actually have the panel (builder / `can_edit`), so the viewer surface is unaffected.

## Notes

- Keep the preview ephemeral; this is about shortening the path to materialize, not about auto-saving.
- The materialize endpoint and worker already exist (`POST /datasets/{id}/analysis/materialize/`), so this is a frontend wiring task plus a small action-payload addition.

PR #1071, opened 2026-07-31
feat(ai): let chat run_analysis clip a layer by another layer

ishiland:
Acceptance criteria 1, 2, 3, 4, 5, 7 and 8 of #683. **Criterion 6 is not included**, so this does not close the issue — see below.

## The tool schema

`operation` gains `clip`, and `mask_layer_id` joins it. The comment that justified the old narrowing said clip needs a drawn mask polygon only the Analysis rail can supply; #682 shipped the layer-mask alternative and that stopped being true, so it is rewritten rather than left to mislead the next reader. `dissolve` stays out on its own merits — materialize-only, an aggregate with no preview shape.

`mask_layer_id` is optional in the schema because the other two operations do not take one, so a model can omit it on a clip. That reads as a fixable mistake with a named parameter, not as an empty result.

## The ProcessingPort seam

`run_analysis_preview` carries a mask **dataset**, passed as the loaded object rather than an id: the caller owns its visibility check exactly as it owns the source dataset's, and the port never checks visibility for anyone. Protocol and default shim move together. No in-repo overlay implements this method — the only other implementation is the stub in `tests/test_processing_port.py`, widened here. **An out-of-repo overlay implementing `ProcessingPort` needs the same parameter**, which is worth knowing before this ships.

The shim also gates the mask's shape (polygonal, and backed by a real table). That is the rail `router_analysis._load_mask_dataset` applies on the REST side, and putting it at the port means every caller gets the same refusal rather than an empty preview the model would report as a real answer.

The `_drop_params_for_other_operations` trap the issue flags is handled by threading the mask object past the request and ignoring it unless the operation is clip — the same shape the validator enforces on `mask_dataset_id`.

## Access control

The mask layer is resolved against the map's layer list and authorized **independently** of the source. Rule 1 applies to both datasets of a two-layer operation, and the layer list is client-supplied either way, so seeing the source buys no claim on the mask. `_resolve_layer_dataset` is now shared by both so they cannot drift. A user who can read the source but not the mask gets a refusal, pinned by `test_mask_layer_is_access_checked_independently`.

## release_session

Chat does **not** pass it, and `test_user_id_is_still_readable_after_the_tool_returns` exercises a post-tool `user.id` read so a future regression surfaces as a test failure rather than a runtime 500. The rollback that returns the pooled connection expires every ORM instance on the session, the authenticated `User` included; the REST endpoint qualifies because it reads nothing afterwards, and copying that call site verbatim is the natural mistake.

## Not included

**Criterion 6** (the #675 handoff prefilling operation, source and mask layer into the Analysis panel) is untouched, along with the two files that would carry it: the union at `use-ephemeral-layers.ts:16` and the producer ternary at `ChatPanel.tsx:525`. Widening either without the panel's `maskLayerId` initializer would make "Save as dataset" appear and prefill a clip with no mask — a state Create cannot submit. All three move together in a follow-up.

Today a chat clip preview renders as an overlay and is simply not offered "Save as dataset": `ChatPanel.tsx:525-534` builds the handoff only for buffer and centroid, so `onSaveAsDataset` is undefined and `EphemeralBadge` renders no button.

The worker's `operation == "clip"` guard at `analysis/tasks.py:643-652` is out of scope per the issue body — chat is preview-only and never enqueues a materialize, so no chat-originated mask reaches the worker.

## Line budget

`defaults_processing_port.py` goes 406 → 450 in `test_layering.py`, with the reason recorded inline next to the cap. The growth is the widened signature (ruff wraps eight params one per line) plus the shape gate.

## Verification

```
uv run pytest tests/test_ai_chat_run_analysis.py tests/test_processing_port.py \
  tests/test_layering.py tests/test_extensions.py tests/test_analysis_preview.py \
  tests/test_analysis_materialize.py -q        # 217 passed
uv run pytest tests/test_*chat*.py tests/test_ai_*.py -q                # 193 passed
```

API impact: the chat tool schema gains an operation and a parameter; no REST contract, schema or migration change. Security impact: one new cross-dataset access path, checked per Rule 1 on both datasets.

Refs #683

## Extension API version

`EXTENSION_API_VERSION` goes 2 → 3. `processing_port` is an overlay-populated registry key and `run_analysis_preview`'s signature changed, so an overlay built against the old shape receives something it was not written for; version.py's bump convention names exactly that case. An overlay declaring v2 now fails loudly at load rather than misbehaving downstream.

The compatibility guard is the complement, not the substitute: the caller omits `mask_dataset` entirely when it is None, so a legacy overlay that declares no version (and therefore only warns) keeps working for buffer and centroid.

**Release coordination — read this before sequencing.** #1073 (the audience-predicate seam out of #1068) also needs a bump when it lands. If both ship in the same release, **one v3 covers both** and the overlay team coordinates once. If this PR spends 2 → 3 now and #1073 spends 3 → 4 later in the same cycle, that is two coordination events for one release, which is the part that actually costs them. Whoever sequences the release should make them ride the same v3.

### DIFF
--- a/backend/app/core/processing_port.py
+++ b/backend/app/core/processing_port.py
@@ -239,6 +239,11 @@ class ProcessingPort(Protocol):
         user_id: uuid.UUID,
         distance_meters: float | None = None,
         mask: dict[str, Any] | None = None,
+        # feat(#683): the clip mask can come from another DATASET, not just a
+        # drawn polygon. Passed as the loaded object rather than an id because
+        # the caller owns its visibility check, exactly as it owns the source
+        # dataset's — see the default implementation's contract.
+        mask_dataset: Any | None = None,
     ) -> Any: ...  # -> AnalysisPreviewResponse
 
     async def get_column_stats(
--- a/backend/app/platform/extensions/defaults_processing_port.py
+++ b/backend/app/platform/extensions/defaults_processing_port.py
@@ -101,22 +101,59 @@ class DefaultProcessingPort:
         return await get_user_roles(session, user)
 
     async def run_analysis_preview(  # type: ignore[no-untyped-def]
-        self, session, dataset, operation, *, user_id, distance_meters=None, mask=None
+        self,
+        session,
+        dataset,
+        operation,
+        *,
+        user_id,
+        distance_meters=None,
+        mask=None,
+        mask_dataset=None,
     ):
         """Run a parameterized analysis preview (M4) for the AI chat surface.
 
         Params are re-validated by ``AnalysisPreviewRequest`` here, so the
         LLM-supplied values pass through exactly the same bounds/requiredness
         checks as the HTTP endpoint (a ValueError surfaces as a tool error the
-        model can retry from). Callers own the dataset visibility check.
+        model can retry from). Callers own the dataset VISIBILITY check, for
+        the mask dataset as much as the source — this port never checks it.
+
+        feat(#683): the mask's SHAPE is checked here, so every port caller gets
+        the rail the REST route applies in ``_load_mask_dataset``. Unioning
+        points or lines masks nothing meaningful, and without the check the
+        failure is an empty result the model reports as a real answer.
+
+        ``release_session`` is deliberately never passed: see the reasoning at
+        the chat call site in ``chat_analysis._run_analysis``.
         """
         from app.modules.catalog.datasets.domain.schemas import AnalysisPreviewRequest
         from app.modules.catalog.datasets.domain.service import run_analysis_preview
 
+        # Ignored unless the operation owns it, mirroring what
+        # _drop_params_for_other_operations does to mask_dataset_id (#682).
+        mask_for_op = mask_dataset if operation == "clip" else None
+        if mask_for_op is not None:
+            shape = (getattr(mask_for_op, "geometry_type", None) or "").upper()
+            if not shape or not getattr(mask_for_op, "table_name", None):
+                raise ValueError("The mask layer has no geometry to clip with.")
+            if shape not in {"POLYGON", "MULTIPOLYGON"}:
+                raise ValueError(
+                    f"Clipping needs a polygon layer as the mask; that one is "
+                    f"{shape}. Pick a polygon layer instead."
+                )
+
         request = AnalysisPreviewRequest(
-            operation=operation, distance_meters=distance_meters, mask=mask
+            operation=operation,
+            distance_meters=distance_meters,
+            mask=mask,
+            # The validator requires exactly one mask source for clip and never
+            # sees the object, so stand the id in for it.
+            mask_dataset_id=getattr(mask_for_op, "id", None),
         )
-        return await run_analysis_preview(session, dataset, request, user_id)
+        return await run_analysis_preview(
+            session, dataset, request, user_id, mask_dataset=mask_for_op
+        )
 
     async def get_column_stats(
         self, session, table_name, column_name, *, class_count=5, allowed_tables=None
--- a/backend/app/platform/extensions/defaults_processing_port.py
+++ b/backend/app/platform/extensions/defaults_processing_port.py
@@ -119,16 +119,24 @@ class DefaultProcessingPort:
         model can retry from). Callers own the dataset VISIBILITY check, for
         the mask dataset as much as the source — this port never checks it.
 
-        feat(#683): the mask's SHAPE is checked here, so every port caller gets
-        the rail the REST route applies in ``_load_mask_dataset``. Unioning
-        points or lines masks nothing meaningful, and without the check the
-        failure is an empty result the model reports as a real answer.
+        feat(#683): the mask's SHAPE and SIZE are checked here, so every port
+        caller gets the rails the REST route applies in ``_load_mask_dataset``.
+        Unioning points or lines masks nothing meaningful, and without the
+        shape check the failure is an empty result the model reports as a real
+        answer. The size ceiling is a resource rail, not a correctness one:
+        ``_mask_pieces`` materializes and subdivides every mask row before the
+        preview's own row cap can bite, so the work scales with the whole mask
+        however small the source is.
 
         ``release_session`` is deliberately never passed: see the reasoning at
         the chat call site in ``chat_analysis._run_analysis``.
         """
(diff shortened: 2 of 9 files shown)


ishiland:
@codex review

ishiland:
Correct, and I should not have deferred it. I had weighed this exact gate and talked myself out of it on the grounds that #719 replaced the whole-layer union with a per-row subdivided join — but that reasoning does not survive reading `_mask_pieces`, which materializes and subdivides every mask row before the preview's row cap can bite. The work scales with the whole mask however small the source is, so it is a resource rail and the acceptance-criteria list not naming it is beside the point.

The ceiling is in, counted the way the REST route counts it: the cached snapshot when present, a LIMIT-bounded live count when it is NULL, because NULL-as-zero would admit exactly the unknown-size layers the gate exists for (fix(#701 review) parity). Both branches have a test and both fail with the ceiling removed — verified by mutation.

The line-budget entry moves 450 → 470 with the reason updated alongside it.

Pushed as fda24b2.

ishiland:
@codex review

ishiland:
Right, and I had noted the overlay obligation in the PR body while leaving the mechanism unused, which is the worse of the two options. Fixed on both halves, because they cover different overlays.

**Legacy, undeclared overlays** load with only a warning, so a version bump does nothing for them. `mask_dataset` now rides along only when set, so such an overlay keeps serving buffer and centroid and only clip degrades — the same shape `router_analysis._defer` uses for exactly this rolling-deploy problem. A port stub without the keyword now serves centroid in a test, so the compatibility half cannot regress silently.

**Declared overlays** get `EXTENSION_API_VERSION` 2 → 3, so the skew is a loud failure at load naming the fix rather than a silent degradation to the generic chat error.

On whether the bump is in scope, since `ProcessingPort` lives in `core/` rather than `protocols.py`: `processing_port` is an overlay-populated registry key (`registry["processing_port"] = TierAwareProcessingPort(...)`), so changing that method's signature changes the expected type of a registry key — which version.py's bump convention names explicitly. The pinned constant test in `test_extension_api_version.py` moves with it and records what v3 adds.

Pushed as 3d56c1a.

ishiland:
@codex review

ishiland:
Accurate, and deliberately out of scope — this is acceptance criterion 6 of #683, which this PR states it does not satisfy. Not acting on it, for a reason worth writing down rather than waving at.

The handoff needs three files to move together:

- `use-ephemeral-layers.ts:16` — `EphemeralAnalysisHandoff.operation` has no `'clip'`, and no `maskLayerId` field.
- `ChatPanel.tsx:525-534` — the producer builds the handoff only for buffer and centroid.
- `AnalysisPanel.tsx:233-239` — the `maskLayerId` initializer reads `savedForm` only and has no `prefill` branch, and the panel ignores `savedForm` entirely whenever a prefill is present.

Carrying the mask into the action, or widening either of the first two, without the third is strictly worse than doing none of them: "Save as dataset" would start appearing on clip previews and prefill `operation: 'clip'` with no mask, which Create cannot submit. Today the button is not offered for clip at all, so there is no partial state to fall into — the overlay renders and that is the whole feature.

The third file belongs to another session for the duration of this wave, so all three land together in a follow-up. Recorded on the issue with the line references so the remainder is actionable: https://github.com/geolens-io/geolens/issues/683#issuecomment-5145770550

One correction to the finding's premise, since it may matter to a later reader: the action already carries `operation: "clip"` today — `collect_run_analysis_action` sets it whenever the result has an operation and a layer_id, and `ChatAction.operation` is typed `str | None`. So the wire is not the blocker; the frontend producer is.

ishiland:
@codex review

ishiland:
Both correct.

**Self-clip by dataset identity.** Fixed — the guard now compares the resolved dataset ids, after both are resolved, rather than the layer ids. A duplicated rendering is exactly the case that slipped through. Test added with two renderings of one dataset.

**Truncation on a capped clip.** Also fixed on the backend: the action reports `truncated` and names `row_count` only when there is a total. The existing `test_truncated_with_unknown_total_omits_the_pair` asserted the old shape; its real guarantee — no "500 of None" — is unchanged, and the test now says so rather than forbidding the flag along with the count. That assertion was fine while every chat operation reported a total and became wrong the moment clip joined them.

One caveat on that second one, since I would rather flag it than let it read as fully closed: the backend half is all I can land here. The action now carries the flag, but `ChatPanel.tsx:520-522` builds its truncation object only when `total != null`, so it is still dropped before reaching the badge — and `EphemeralBadge` has no label for "capped, total unknown", which needs a new t() key in four locales. `EphemeralBadge` belongs to another session for this wave, so I am not touching it. Net effect today is unchanged from before this PR (no disclosure on a capped clip); the payload is now correct and the remaining work is a two-file frontend change. Raising it with the coordinator so it lands rather than getting lost.

Pushed as 3727b72.

ishiland:
@codex review

ishiland:
Same carve-out as my note above, but your comment adds a file I had missed — `ViewerChatPanel.tsx` forwards truncation on the same `total != null` condition — so thank you, that goes into the follow-up record.

Not acting here, and the reason is that forwarding alone buys nothing. With `totalCount` undefined, `EphemeralBadge` falls through to `t('ephemeralBadge.featureCount', { count })` and renders a plain "500 features" — no disclosure. Making it visible needs a third label variant for "capped, total unknown" plus its t() key in four locales, and `EphemeralBadge` belongs to another session for this wave. Landing the two panels without the badge would change nothing except to make the next reader think it was handled.

Nothing regresses in the meantime: a capped clip reads exactly as it did before this PR. The payload is now correct, which is the part that had to be right for the frontend change to be possible at all.

Recorded on #683 with all three files so it is actionable, and raised with the coordinator.

ishiland:
Deferral anchored at the emit site in `chat_analysis.py`, per the AGENTS.md convention — `fix(#1076)` with the reason, beside the code rather than only in the PR body.

It names both consumers, including `ViewerChatPanel.tsx`, so the next reader sees the full shape of what is outstanding rather than the builder half.

Head is `e951b8a3c`. The `EXTENSION_API_VERSION` bump was already in at `3d56c1a8d`, two pushes ago — the head had not moved because the anchor was the only condition left, not because the bump was still coming.

ishiland:
@codex review

PR #1072, opened 2026-07-31
test: integration check for a 4-PR batch (do not merge)

ishiland:
Do not merge this branch. It exists to run CI once on a combination that would otherwise cost four separate runs, and it is deleted either way.

Branch protection requires an up-to-date branch, so under `strict` only one PR lands per move of `main` and every other open PR has to update-branch and re-run the full suite. With four ready PRs that is four sequential cycles, and each one invalidates the three behind it. This runs the suite once on all four merged together.

## Contents

Each is codex-clean at its head, verified by timestamp rather than by sha alone: the clean comment has to be newer than the newest inline finding at that head, because codex can post findings at a sha it has already passed.

| PR | Head | Clean pass |
|----|------|-----------|
| #1034 | `10fdaf1df` | 17:19, no findings at head |
| #1054 | `dbe40cdc3` | 17:23, no findings at head, 11 inline total as a positive control |
| #1057 | `c30e96772` | 15:29, newest finding 15:09 |
| #1061 | `bc2df3009` | 17:33, newest finding 15:40 |

Merged onto `f0113cdfa` with no conflicts.

- Single alembic head, `0032_api_key_scope`, across 32 revision files. None of the four adds a migration.
- `ruff check` clean, `ruff format --check` clean across 864 files.
- `test_layering.py` passes, 34 tests. This is the one that mattered: #1034 and #1054 both edit the module LOC ratchet, and git auto-merged them. They raise caps on different files (`maps/router.py` +8, `maps/style_json.py` +85), so the auto-merge is semantically right, and the ratchet is exact rather than an upper bound, so running it is what proves that rather than reading the diff.

## What happens next

Green: the four merge individually, each keeping its own squash commit and `Closes #NNN`, and this branch is deleted. Those merges need `--admin` to bypass the up-to-date requirement, which is only defensible because this run has already tested the exact combination that will land.

Red: the failure names the bad combination. The four stay unmerged and get bisected rather than landed on a guess.

Same procedure as batch 1 (#1067), which landed thirteen.


ishiland:
Red on `tests/test_chat_narrative.py::test_error_message_mapping`. #1061 fails that test on its own branch and is the only member touching `backend/app/processing/ai/chat_constants.py`, which the test asserts against. Dropped; it stays open for its owner.

Replaced by #1075, which carries the other three plus #1025, #1032, #1038 and #1056. `test_chat_narrative.py` passes there, which confirms the failure did not follow anything else out.

Closing and deleting the branch.

PR #1074, opened 2026-07-31
test(ci): parse the merge_group conditions instead of scanning them

ishiland:
Follow-up to #1051. The guard I shipped there string-matches an expression language, and it is wrong in both directions.

## The bug

```python
return "merge_group" in str(condition)
```

`ci-ok`'s condition is a bare `always()`, so this reports it as **skipping** when it always runs. Dormant today — the assertion only walks `ci-ok`'s dependencies, none of which use that idiom — but it means the tool built to answer step 1 of a queue incident ("did the gates actually run, or did they skip to a green CI OK?") can cry wolf and send you hunting a hole that is not there.

## The repair that would have been worse

The obvious fix is to also match `"always()"`. **Do not.** `always()` is a *status-check* function — it means "run even if the jobs I need failed or were cancelled" — and says nothing about event context. `always() && <clause>` still evaluates `<clause>`. Two jobs in this file are written exactly that way:

```yaml
e2e-test:
  if: always() && (github.event_name == 'schedule' || github.event_name == 'workflow_dispatch') && ...
accessibility:
  if: always() && (github.event_name == 'push' || github.event_name == 'schedule' || ...) && ...
```

Neither allowlist contains `merge_group`. Both provably skip in the queue. Matching `"always()"` reports both as running — turning a false alarm into a **false all-clear**, in the one tool meant to catch the silent failure. A guard that is wrong toward "all clear" is worse than no guard, because you stop looking.

No amount of adding literals separates a bare `always()` from `always() && <allowlist excluding merge_group>` from `github.event_name == 'merge_group'`. They differ structurally, not lexically. **Parse, don't scan.**

## What it does now

1. Strip the `always()` term — orthogonal to event context.
2. `github.event_name != 'merge_group'` anywhere → excluded. (A substring match reads an exclusion as permission: the same class of error one layer down.)
3. No `github.event_name ==` comparisons in what remains → no event gate, the job runs.
4. Comparisons present → the job runs only if `merge_group` is among them.

**Documented limit, in the docstring rather than papered over:** this reads the *set* of comparisons, not the boolean structure around them. A condition permitting `merge_group` in one `||` branch while a sibling `&&` excludes it would read as running. No such condition exists in `ci.yml`, and the note says the honest fix if one appears is to evaluate the expression — not to add another special case.

## Verification

Six predicate cases: bare `always()`; `always()` with an allowlist excluding `merge_group`; one including it; no condition; the twelve gated jobs' real written form; and `!=`. Plus a cross-check against the actual workflow asserting every `ci-ok` dependency reads as running and `e2e-test` / `accessibility` / `stac-validate` read as skipping.

That last one also **pins the non-gating-jobs decision from #1000** — if someone moves one of those onto `merge_group`, the test says to revisit the decision rather than silently accepting the flip.

**RED-checked against both wrong predicates**, verifying *why* each fails rather than only that it does:

| injected predicate | fails | 
|---|---|
| original substring match | `bare_always_runs`, `explicit_exclusion_skips` |
| the `always()`-matching repair | `always_with_allowlist_excluding_merge_group_skips`, `explicit_exclusion_skips`, `agrees_with_the_real_workflow` |

The third of those reports `AssertionError: e2e-test is deliberately off merge_group` — the false all-clear caught on the real workflow, not only on a synthetic condition. Correct predicate: **13 passed**.

## Impact

Test-only, one file. No workflow change, no runtime code, no behaviour change to the queue itself.

```
cd backend && uv run pytest tests/test_ci_merge_queue.py -q   # 13 passed
```

### DIFF
--- a/backend/tests/test_ci_merge_queue.py
+++ b/backend/tests/test_ci_merge_queue.py
@@ -51,30 +51,40 @@ def _runs_on_merge_group(job: dict) -> bool:
     the run, and ``always() && <clause>`` still evaluates ``<clause>``. So it is
     stripped before looking at event context — treating its presence as "runs
     everywhere" would report ``e2e-test`` and ``accessibility`` as running in
-    the queue when their event allowlists provably exclude it, and a guard that
-    is wrong toward "all clear" is worse than no guard because you stop looking.
+    the queue when their event allowlists provably exclude it.
 
-    After that: no ``github.event_name`` comparison at all means no event gate,
-    so the job runs. Comparisons present mean the job runs only if
-    ``merge_group`` is among the permitted ones.
+    After stripping: nothing left means no gate at all, so the job runs. A
+    ``github.event_name`` comparison means the job runs only if ``merge_group``
+    is among the permitted ones.
 
-    Known limit, stated rather than papered over: this reads the SET of event
-    comparisons, not the boolean structure around them. A condition that
-    permits `merge_group` inside one branch of an `||` while a sibling `&&`
-    excludes it would be read as running. No such condition exists in
-    ``ci.yml``, and the honest fix if one appears is to evaluate the
-    expression, not to add another special case here.
+    **Anything else is reported as NOT running** (fix(#1074 review), a P1). The
+    tempting reading — "no event comparison, therefore no event gate, therefore
+    it runs" — is wrong for the shape this repo actually uses. A bare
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1075, opened 2026-07-31
test: integration check for a 7-PR batch (do not merge)

ishiland:
Do not merge this branch. It runs CI once on a combination that would otherwise cost seven separate runs, and it is deleted either way. Third of these; #1067 landed thirteen and #1072 caught a bad one.

## Why this replaces #1072

#1072 carried #1034, #1054, #1057 and #1061 and went red on `tests/test_chat_narrative.py::test_error_message_mapping`, with the expected set holding a key the mapping does not produce. #1061 fails that test on its own branch and is the only member touching `backend/app/processing/ai/chat_constants.py`, which is what the test asserts against. It is dropped and stays open.

That is the batch working as intended. The four would have merged individually with `--admin` on the strength of four green PR runs, and #1061's was red for a reason nobody had connected to it.

## Contents

| PR | Head | Verdict |
|----|------|---------|
| #1025 | `8346dcc57` | clean 17:49 |
| #1032 | `68c596c8e` | clean 17:48 |
| #1034 | `10fdaf1df` | clean 17:19 |
| #1038 | `f23dad6bb` | clean 17:48 |
| #1054 | `dbe40cdc3` | clean 17:23 |
| #1056 | `76d0a8bce1` | clean at `5af3dacbfb`; comment-only since, plus one LOC-cap line the comment rewrite requires |
| #1057 | `c30e96772` | clean 15:29, newest review at head 15:09 |

Each verdict is a clean-pass comment naming that head and newer than the newest codex review whose `commit_id` is that head. Dating findings by an inline comment's `commit_id` does not work: GitHub re-anchors it as the file changes, so a comment can carry the current head while belonging to an older review.

**#1044 was dropped for conflicting with #1034 and #1054 over `_MODULE_LOC_CAPS`.** Deliberately not resolved here. A batch branch is thrown away, so a resolution made in it never reaches the individual PRs and the same conflict returns at the second merge. Two PRs that conflict with each other cannot share a batch; one merges after the other and rebases. #1044 is clean at `49fec6f7de` and merges separately.

- Single alembic head, `0032_api_key_scope`. No member adds a migration.
- `ruff check` and `ruff format --check` clean across 864 files.
- `test_layering.py` plus `test_chat_narrative.py`, 45 passed. The layering run matters because four members edit `_MODULE_LOC_CAPS` and git auto-merged them; the caps are exact rather than ceilings, so running the ratchet is what proves the merge is numerically right, not reading it. The chat-narrative run confirms the #1072 failure does not follow #1061 out of the batch.

## Coverage, stated honestly

A green run here proves the **PR-visible** combination passes. `accessibility`, `stac-validate` and the full `e2e-test` suite gate on `push || schedule || workflow_dispatch` and so never run on a pull request. They first meet this combination on main's own push run, after the merges. The same blind spot applies to a `merge_group` run, since none of the three carries a `merge_group` clause.

## Next

Green: the seven merge individually with `--admin`, each keeping its own squash commit and `Closes #NNN`, and this branch is deleted. `--admin` is defensible only because this run tested the exact combination that will land.

Red: the failure names the bad member, as it did on #1072. Nothing merges on a guess.


ishiland:
Red on `tests/test_datasets.py::TestUpdateMetadata::test_a_grant_to_an_empty_role_does_not_block` — a 422 from #1056's own visibility guard, on the test asserting that guard does not fire for a grant to an empty role. #1056 is the only member that can produce it.

Dropped it and rebuilt as #1078 with the other six. #1056 needs a full-suite run under `-n 4` on its own branch: the guard reads `users`/`user_roles` to decide whether a role is empty, and those carry rows from every other test on the shared worker DB, so a targeted selection and a full run give different answers.

Closing and deleting the branch.

PR #1078, opened 2026-07-31
test: integration check for a 6-PR batch (do not merge)

ishiland:
Do not merge this branch. Replaces #1075, which went red.

## Why

#1075 carried the same six plus #1056 and failed on:

```
FAILED tests/test_datasets.py::TestUpdateMetadata::test_a_grant_to_an_empty_role_does_not_block
assert 422 == 200
{"detail":"Cannot restrict visibility: dataset is used in shared maps: Empty Grant Map"}
```

That 422 comes from #1056's own guard, and the test it fails is the one asserting that guard does NOT fire for a grant to an empty role. #1056 is the only member that can produce it.

The test passes on #1056's branch — but it was run as part of a targeted `test_datasets` + `test_layering` selection, not the full suite under `pytest -n 4`. A guard that reads `users` and `user_roles` to decide whether a role is empty is exactly the shape that passes a subset and fails at full-suite scale, because another test's rows are visible on the shared per-worker database. Same class as #1024, which broke main this morning after passing its own PR.

So #1056 is dropped pending a full-suite run on its own branch. It is not a batch interaction to work around; it is a question about that PR that a subset run cannot answer.

## Contents

#1025 `8346dcc57`, #1032 `68c596c8e`, #1034 `10fdaf1df`, #1038 `f23dad6bb`, #1054 `dbe40cdc3`, #1057 `c30e96772`. Each carries a codex clean-pass comment naming that head, newer than the newest review whose `commit_id` is that head.

Merged onto `f0113cdfa`, no conflicts.

## Also excluded

**#1044** conflicts with #1034 and #1054 over `_MODULE_LOC_CAPS`. Not resolved here on purpose: a batch branch is thrown away, so a resolution made in it never reaches the individual PRs and the same conflict returns at the second merge with nobody having decided it. Two PRs that conflict with each other cannot share a batch. #1044 rebases after these land.

## Coverage

Green here proves the PR-visible combination. `accessibility`, `stac-validate` and the full `e2e-test` suite gate on `push || schedule || workflow_dispatch`, so they never run on a pull request and first meet this combination on main's push run. The same blind spot applies to a `merge_group` run, since none of the three carries a `merge_group` clause.

## Next

Green: the six merge individually with `--admin`, each keeping its own squash commit and `Closes #NNN`. Red: the failure names the member, as it did on #1072 and again here.


ishiland:
Green across all checks. The six merged individually, each keeping its own squash commit and `Closes #NNN`. Closing and deleting the branch.

PR #1079, opened 2026-07-31
fix(chat): disclose a capped preview when the total is unknown

### LINKED ISSUE #1076: fix(chat): show a truncation indicator when a clip preview is capped and the total is unknown
## Summary

Enabling clip on chat `run_analysis` (#1071) surfaces a truncation-disclosure gap. The backend half is fixed there; the user-visible half needs two frontend files and is carved out here so #1071 stays in scope.

`run_analysis_preview` returns no `source_feature_count` for a clip, because the clipped total is genuinely unknown before the clip runs. The chat action collector only forwarded `truncated` when a total came with it, so a capped clip preview presented as a complete result. That is the #674 audit's concern arriving through a new door.

## Already done in #1071

The action now reports the cap and names the total only when there is one. The payload is correct.

## What is left, and why it is separate

Two frontend files, one of which belongs to another session this wave:

- `ChatPanel.tsx:520-522` builds its truncation object only when `total != null`, so a cap with no total is dropped before it reaches the badge.
- `EphemeralBadge` has no label for "capped, total unknown". That needs a new `t()` key in **all four locales** (en/es/fr/de) — a `defaultValue` alone fails the `npm run test:i18n` parity gate.

`EphemeralBadge` is owned by another session for the current wave, which is the direct reason this is filed rather than finished.

## Net effect today

Unchanged from before #1071. Nothing regresses: the payload carries the right information and the UI ignores the new field exactly as it ignored its absence. The gap is that a user running a capped clip from chat still sees no cap indicator.

## Acceptance

- A clip preview that hits the row cap shows a truncation indicator with no total.
- A clip preview that does not hit the cap shows nothing.
- A non-clip preview that hits the cap still shows the existing "N of M" form, unchanged.
- The new key exists in en/es/fr/de.

Both directions matter: a fix that shows the badge for capped clips but also fires on uncapped ones has traded a silent truncation for a wrong one.

Related: #674 (the audit that first raised truncation disclosure), #683 / #1071 (where this surfaced).

ishiland:
Closes #1076 — the frontend half of the truncation gap #1071 surfaced.

## The gap

A clip filters rows, so `run_analysis_preview` cannot report a source total for it: the clipped total is genuinely unknown before the clip runs. Every consumer required a total before disclosing anything, so a capped clip preview presented itself as the complete result — #674's concern arriving through the door #1071 opened by enabling clip from chat.

The backend half landed in #1071: the action reports the cap and names the total only when there is one. This is the rest.

## Three files, and they only work together

| File | Change |
|---|---|
| `ChatPanel.tsx` | forwards the flag without waiting for a total |
| `ViewerChatPanel.tsx` | same |
| `EphemeralBadge.tsx` | third case for "capped, total unknown" |

Forwarding from the two panels without the badge label renders an ordinary count, so a partial fix would look complete while a user hitting a capped clip still saw no cap. The badge is the load-bearing one.

The viewer panel matters most of the two — an embed or a shared link lands there, so a silent truncation on that surface reaches the widest audience. It was also the file I originally missed; codex named it on #1071.

## Both directions

The issue is explicit that a fix showing the badge for capped clips but also firing on uncapped ones has traded a silent truncation for a wrong one. Pinned accordingly:

- capped, no total → `Result · first 500 features`
- capped, total known → `Result · 500 of 10,651 features` (**unchanged**)
- not capped → `Result · 500 features`, and neither cap wording appears

Both panels have a forwards-it and a forwards-nothing test.

Verified rather than assumed: reverting the badge to its two-case form fails the new case and leaves the other six passing.

## i18n

`ephemeralBadge.featureCountCapped` in all four locales with `_one`/`_other`. The `_one` values interpolate `{{count}}` rather than hardcoding "1" — French resolves count 0 to `_one`, so a literal would render "1 entité" for zero.

## Verification

```
cd frontend
npm run test:i18n     # 4 passed — locale parity
npm run lint
npm run typecheck
npm run test:coverage # 367 files, 4365 tests
```

No backend, schema or API change.

### DIFF
--- a/frontend/src/components/builder/ChatPanel.tsx
+++ b/frontend/src/components/builder/ChatPanel.tsx
@@ -522,6 +522,10 @@ export function ChatPanel({
       // dropped the disclosure entirely — a capped clip preview then read as
       // complete. The total rides along when there is one; the badge picks its
       // wording from whether it arrived.
+      //
+      // The producer half is #1071: until that lands, collect_run_analysis_action
+      // still omits `truncated` whenever the total is absent, so this branch is
+      // correct and simply not yet reachable for a clip.
       const total = typeof action.row_count === 'number' ? action.row_count : undefined;
       const truncation = action.truncated === true
         ? { truncated: true, ...(total != null ? { totalCount: total } : {}) }
--- a/frontend/src/components/builder/ChatPanel.tsx
+++ b/frontend/src/components/builder/ChatPanel.tsx
@@ -516,9 +516,15 @@ export function ChatPanel({
       // fix(#674 audit): forward the truncation pair so EphemeralBadge can say
       // "500 of 10,651 features". Both query_data and run_analysis emit them
       // only when the server-side cap actually bit.
+      //
+      // fix(#1076): the flag no longer waits for the total. A clip filters
+      // rows, so the server reports no source total for it, and requiring one
+      // dropped the disclosure entirely — a capped clip preview then read as
+      // complete. The total rides along when there is one; the badge picks its
+      // wording from whether it arrived.
       const total = typeof action.row_count === 'number' ? action.row_count : undefined;
-      const truncation = action.truncated === true && total != null
-        ? { truncated: true, totalCount: total }
+      const truncation = action.truncated === true
+        ? { truncated: true, ...(total != null ? { totalCount: total } : {}) }
         : undefined;
       // feat(#675): a run_analysis action carries its reconstruction params so
       // the badge can offer "Save as dataset" (prefilled Analysis panel).
--- a/frontend/src/components/builder/EphemeralBadge.tsx
+++ b/frontend/src/components/builder/EphemeralBadge.tsx
@@ -21,15 +21,26 @@ interface EphemeralBadgeProps {
 export function EphemeralBadge({ featureCount, onDismiss, truncated, totalCount, onSaveAsDataset, className }: EphemeralBadgeProps) {
   const { t } = useTranslation('builder');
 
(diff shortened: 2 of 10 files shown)


ishiland:
@codex review

ishiland:
Verified against this PR's base and you are right, with one thing to add that changes what the fix is.

On `origin/main` at `f0113cdfa`, `chat_analysis.py:190-193` is still:

```python
total = result.get("source_feature_count")
if result.get("truncated") and total:
    action["truncated"] = True
    action["row_count"] = total
```

So for a clip the flag never reaches the wire and these branches are unreachable. That is not a defect in this diff — **it is a merge-order dependency.** The producer half is #1071, where that block already reads `if result.get("truncated")` with the total attached only when present. #1071 is open, reviewed clean at `e951b8a3c1`, waiting on the batch queue.

I have not duplicated the backend change here; doing that would collide with #1071 for no gain. Instead both call sites now name the dependency in code, so a reader checking reachability finds the answer rather than a puzzle.

Neither PR regresses anything alone: #1071 alone emits a correct payload nobody reads, #1079 alone forwards a flag nobody sends. Together they close the gap. They touch disjoint files, so they can share a batch, and that is the best way to validate the pair.

Worth saying plainly: the tests here pass on this branch because they drive the SSE actions directly, which pins the frontend's half of the contract honestly but says nothing about end-to-end reachability. You asked the question the tests could not, which is the useful half of the review.

Head is `75aa3017c`. Merge order recorded on #1076 and raised with the coordinator.

ishiland:
@codex review

Issue #1077, opened 2026-07-31
refactor(sec): invert Rule 2 safe-env credit to an allowlist of shapes


ishiland:
Split out of #996 rather than taken inside it. #996 closed the *deferral* half of this; this issue is the *conditionality* half, which is older and structurally different.

## The current design

`_scope_uses_safe_env` in `backend/tests/test_rule2_structural.py` grants credit by **walking** a scope: if the immediate body contains a Call resolving to the canonical `gdal_safe_env`, every GDAL argv in that scope is exempt from being a violation.

That is a broad exemption granted by a narrow observation, and it fails in one direction — a call that is *seen* but not *executed* produces silent credit, i.e. a false all-clear in a security gate.

## What #996 already closed, so this issue does not need to

The **deferred-body** class. Measured rather than reasoned: the only Python constructs whose body does not run at construction are **generator expression, lambda, nested def/async def**. Comprehensions, class bodies, decorator expressions and default arguments are all eager. All three deferral constructs now deny credit.

That set is closed, so it is not the open surface it looked like. `functools.partial(gdal_safe_env)` also needs no rule — `_is_canonical_helper_call` resolves the Call's own `func`, which is `partial`, so it never credits.

## What is still open, and why it needs a design change rather than a rule

**Conditionality.** Measured against the current gate:

| shape | verdict today |
|---|---|
| `if False: gdal_safe_env()` | **CREDITED** |
| untaken ternary arm | **CREDITED** |
| unreached `match` case | **CREDITED** |

These execute eagerly *when reached*, so they are not deferral cases and the #996 work does not touch them. This is the #974 boundary — credit asks whether the helper is called in this scope, not whether that line runs — and unlike deferral it has no finite enumeration. Any conditional structure produces another instance.

Patching it shape by shape is the thing to avoid. #996 took ten review rounds closing a *closed* set; an open one would not converge.

## Proposed change

Invert the default. Grant credit only for a small **allowlist of shapes that can be named**, and report everything else as unclamped:

- a direct call in the scope's straight-line body;
- a `with gdal_safe_open_env():` block (the wrapper path already works this way);
- a plain assignment `env = gdal_safe_env()` whose target reaches the subprocess call.

Everything else — conditional, deferred, novel — produces a **false alarm someone investigates** rather than silent credit. Bounded by construction, and it matches the direction taken for the merge-queue guard in #1074, where a condition the predicate cannot prove is reported rather than assumed.

## Acceptance

[...]

PR #1080, opened 2026-07-31
test: integration check for a 7-PR batch (do not merge)

ishiland:
Do not merge this branch. Fourth of these. #1067 landed 13, #1072 caught a bad one, #1078 landed 6.

## Contents

Merged onto `639b3d3c2`, no conflicts.

| PR | Head | Verdict |
|----|------|---------|
| #1030 | `1919d7930c` | clean 18:18:52 |
| #1035 | `a7964bc199` | **open finding, landing on a deferral ruling — see below** |
| #1059 | `4146839cab` | clean |
| #1061 | `150a1dd765` | clean 18:44:00 |
| #1065 | `f9d1359357` | clean |
| #1071 | `e951b8a3c1` | clean 18:49:59 |
| #1074 | `c41cf70f5a` | clean |

Each verdict is a clean-pass comment naming that head and newer than the newest review whose `commit_id` is that head. Both endpoints checked: clean passes arrive as issue comments, findings as pull reviews, so querying one is blind to the other.

**#1035 is the exception and it is deliberate.** It carries one open finding at head, a false positive where a `BoolOp` predicate operand is treated as transparent. It is not fixed here because fixing it would make the detector report *less*, which trades a false alarm for a way to under-report and deserves its own review. It is recorded in the module docstring alongside the other two open classes and folded into #1077 as an acceptance criterion with a runnable reproduction. Reviewed at head with a documented, tracked deferral is the bar, not a green checkmark.

## Dropped

**#1037** conflicts with another member, almost certainly #1030 since both restructure analysis test fixtures. Deliberately not resolved here: a batch branch is thrown away, so a resolution made inside it never reaches the individual PRs and the same conflict returns at the second merge with nobody having decided it. It rebases after these land.

**#1044** has no review verdict at `034c0f08d1` yet.

- Single alembic head, `0032_api_key_scope`. No member adds a migration.
- `ruff check` and `ruff format --check` clean across 865 files.
- 140 passed across `test_layering.py`, `test_chat_narrative.py`, `test_ci_merge_queue.py`, `test_rule2_structural.py` — the four structural gates these members touch. `test_chat_narrative` matters specifically: it is what #1061 broke in batch 2, and #1061 is in this batch with the fix.

## Coverage

Green proves the **PR-visible** combination. `accessibility`, `stac-validate` and the full `e2e-test` gate on `push || schedule || workflow_dispatch`, so they never run on a pull request and first meet this combination on main's push run. A `merge_group` run has the same blind spot, since none of the three carries a `merge_group` clause.

## Next

Green: the seven merge individually with `--admin`, each keeping its own squash commit and `Closes #NNN`, and this branch is deleted. Red: the failure names the member, as it did twice today.


ishiland:
Superseded by #1081, which carries the same members minus #1030.

Batch 4 did its job: it surfaced a real interaction that neither member could show alone. #1030 adds a module-level autouse **async** fixture to `test_analysis_materialize.py`; #1057 (already on main) adds two **sync** tests to a class in that module. pytest cannot hand an async fixture to a sync test, and its own error path then leaves the fixture un-finalizable — `execute()` registers a finalizer before setup raises, and `finish()` declines to clean up because `cached_result is None`. Every later test on that xdist worker dies at `assert not self._finalizers`: 951 errors on the first attempt, 934 on the rerun.

Not reopened as a rerun, because this branch's run tested main+7 and ~934 of its tests errored rather than passed, which makes it no evidence at all about main+6.

Follow-ups: #1030 fixed at `a13d4804a` (synchronous teardown owning a short-lived connection, so it no longer cares whether a test is sync or async), and #1082/#1083 add a structural gate so the next occurrence fails at PR time naming the test, the class and the fixture.

PR #1081, opened 2026-07-31
ci: integration batch 5 (#1035, #1059, #1061, #1065, #1071, #1074)

ishiland:
Integration batch: six PRs merged onto current main (`639b3d3c2`) so the combination gets one CI run instead of six sequential update-branch cycles.

| PR | head |
|----|------|
| #1035 | `a7964bc19` |
| #1059 | `4146839ca` |
| #1061 | `150a1dd76` |
| #1065 | `f9d135935` |
| #1071 | `e951b8a3c` |
| #1074 | `c41cf70f5` |

These are batch 4's members minus #1030. Batch 4 went red on a genuine interaction: #1030 adds a module-level autouse **async** fixture to `test_analysis_materialize.py`, and #1057 (already on main) adds two **sync** tests to a class in that same module. pytest cannot hand an async fixture to a sync test, the failed setup leaves the `client` fixture holding finalizers, and every later test on that xdist worker dies at `assert not self._finalizers` — 951 errors on one attempt, 934 on the rerun. Both PRs are correct alone.

This is a fresh branch rather than a rerun of batch 4 on purpose. Batch 4's run tested main+7 and ~934 of its tests errored rather than passed, so it is not evidence about main+6. #1030 rejoins a later batch once its fixture is sync-safe.

All six heads were verified unchanged from what batch 4 tested. Local gates before pushing: single alembic head (`0032_api_key_scope`), `ruff check` and `ruff format --check` clean, and the exact LOC ratchet suite green (34 passed) since two members move caps in `test_layering.py`.

Green: the six merge individually with `--admin`, each keeping its own squash commit and `Closes #NNN`, and this branch is deleted. Red: the failure names the member, as it did on batch 4.

Coverage caveat, same as every batch: this is a `pull_request` event, so the push-only jobs (`accessibility`, `stac-validate`, full `e2e-test`) do not run here. A green result means the PR-visible combination passes. Those three first meet the merged combination on main's own push run.


ishiland:
All six members merged individually with `--admin`, each keeping its own squash commit and `Closes #NNN`:

| PR | commit |
|----|--------|
| #1035 | `f8a260f07` |
| #1059 | `287136ded` |
| #1061 | `97cdd81c5` |
| #1065 | `1306f98f2` |
| #1071 | `86b67f0e6` |
| #1074 | `4d2323286` |

The batch run was 29 SUCCESS / 9 SKIPPED / 0 FAILURE, with `Backend Tests`, `Frontend Tests` and `E2E Smoke (core)` each confirmed SUCCESS **by name** rather than inferred from an absence of failures. All six heads were re-verified as ancestors of the tested branch immediately before merging.

Six merges for one CI run. Branch deleted.

Coverage qualifier, unchanged: this was a `pull_request` event, so `accessibility`, `stac-validate` and the full `e2e-test` suite did not run here. They meet this combination for the first time on main's own push run.

PR #1083, opened 2026-07-31
test(backend): gate sync tests against async fixture dependencies

### LINKED ISSUE #1082: test: guard against async autouse fixtures reaching sync tests
## Problem

A module-level `@pytest.fixture(autouse=True)` declared `async def` silently poisons an entire xdist worker the moment anyone adds a plain `def test_` to any class in that module.

Integration batch 4 hit this: 951 errors on one run, 934 on the rerun, spanning the whole `gw0` stream (`test_tiles`, `test_validation`, `test_stac_visibility`, `test_vrt_*`, `test_services_endpoints`, and more). Only the first error was real; the rest were cascade.

## Mechanism

Two layers.

**Ours:** an async autouse fixture at module scope pulls `test_db_session` -> `client`. pytest cannot hand an async fixture to a sync test, so setup raises for that test.

**pytest's** (`_pytest/fixtures.py`, 9.1.1): `execute()` registers a finalizer at line 1222. The sync-test-requests-async-fixture error raises at line 1320, before the `try/except` at 1327 that sets `cached_result`. The FixtureDef is left with `cached_result=None` and a non-empty `_finalizers`. `finish()` at 1147 then returns early:

```python
if self.cached_result is None:
    # Already finished. It is assumed that finalizers cannot be added in
    # this state.
    return
```

so `_finalizers` is never cleared and every later `execute()` of that FixtureDef trips `assert not self._finalizers` at 1221. The fixture is dead for the rest of that worker's session. The comment asserts an invariant that pytest's own error path violates.

## Why a guard rather than a one-line fix

The existing remedy in `test_analysis_materialize.py` is a class-level sync override on `TestUserErrorMessage`, and it is correct — those 5 sync tests passed in the same CI run where 2 others errored. It was a complete enumeration of the sync tests that existed when it was written.

It went silently incomplete when two sync tests were added to a *different* class in the same module. Nothing re-checked the precondition, because nothing knew there was one. Adding a second override repeats the enumeration and breaks on the next sync test anyone adds.

## Proposed

A structural test in `backend/tests/`, in the shape of the existing `test_rule1_structural.py` / `test_rule2_structural.py`: walk the ASTs under `backend/tests/`, and for every async autouse fixture determine its scope (module-level applies to the module, class-level to its class, a class-level definition shadows a module-level one of the same name) and fail if any sync `def test_` falls inside that scope without a shadow.

The failure message should name the offending test, the fixture reaching it, and the class-level-sync-override remedy, so the next person hits it at PR time with the fix already spelled out.

Scope modelling is load-bearing: a version that pools class and module scope reports false positives (for example `test_geometry_detection.py`, whose autouse fixture is class-level in a class its sync tests do not live in). Correctly scoped, current `main` has zero violations, so the guard lands green.

## Notes

- `main` is not exposed today; the exposure arrives with the PR that introduces the module-level async autouse fixture.
- Worth reporting the `finish()` early-return upstream separately.

ishiland:
Closes #1082.

## What

Adds `backend/tests/test_fixture_scope_structural.py`: a structural gate that fails when a sync test can reach an async fixture, whether through an autouse fixture in scope or through its own parameters. Same idea and neighbourhood as `test_rule1_structural.py` and `test_rule2_structural.py`.

Guard only. It deliberately does not fix the violation it finds — that belongs to #1030.

## Why

Batch 4's Backend Tests run produced **951 errors** (934 on rerun). Every one was an `AssertionError` at setup with no message, no fixture name, and no test name:

```
.venv/.../_pytest/fixtures.py:1221: in execute
    assert not self._finalizers
E   AssertionError
```

The cause was a single test, and it was not any of the 951.

`pytest_fixture_setup` raises for a sync-test-requests-async-fixture at `fixtures.py:1320`, which is **before** the `try` at 1327 that sets `cached_result`. By then `execute()` has already registered a finalizer at 1222. So the `FixtureDef` is left with `cached_result=None` and a non-empty `_finalizers` — precisely the state `finish()` refuses to clean:

```python
if self.cached_result is None:
    # Already finished. It is assumed that finalizers cannot be added in
    # this state.
    return
```

Its own error path violates the invariant that comment asserts, so `_finalizers` is never cleared and every later `execute()` of that fixture trips the assertion for the rest of the worker's session. Blast radius is purely ordering — the poisoner ran at 5% on gw0.

**This is an upstream pytest bug (9.1.1), not filed.** Flagging it here rather than reporting it; that call isn't mine to make.

## Why a gate and not a second shadow

Two PRs, each green alone. #1030 adds a module-level `async` autouse fixture reaching `client` through `test_db_session`, and **anticipated this** — it shadows the fixture with a sync no-op for `TestUserErrorMessage`, the only sync-test class on its branch, with a docstring predicting the exact failure. #1057 then added two sync tests to `TestMaterializeWorker`, a class that was fully async when #1030 was written.

CI shows both halves: the shadowed class's 5 sync tests **passed** in the same run where the unshadowed class's 2 errored.

So the existing remedy is a per-class enumeration that was complete when written and silently incomplete once another PR touched a different class in the same file. A second shadow repeats the enumeration; the file has six classes.

## Evidence the gate fires

Same worktree, same command, only `test_analysis_materialize.py` swapped.

**Green on main (`639b3d3c2`):**

```
tests/test_fixture_scope_structural.py::test_no_sync_test_depends_on_an_async_fixture PASSED
tests/test_fixture_scope_structural.py::test_the_gate_detects_the_shape_that_produced_951_errors PASSED
2 passed
```

**Red on batch-4's copy of that file:**

```
1 failed, 1 passed

E   Failed: A sync test reaches an async fixture. pytest errors its setup and leaves the
E   fixture's FixtureDef un-finalizable, so every later test on the same xdist worker that
E   touches it errors with a bare AssertionError naming nothing.
E
E   Shadow the fixture with a sync no-op in the class holding the sync test, or move the
E   sync test to a class that already does (see TestUserErrorMessage in
E   test_analysis_materialize.py):
E
E       @pytest.fixture(autouse=True)
E       def <fixture_name>(self):
E           yield
E
E   test_analysis_materialize.py::TestMaterializeWorker::test_timeout_message_names_the_budget_that_fired is sync but reaches async autouse fixture '_release_analysis_job_slots'
E   test_analysis_materialize.py::TestMaterializeWorker::test_analysis_timeouts_track_their_settings is sync but reaches async autouse fixture '_release_analysis_job_slots'
```

Two named lines with the remedy, in place of 951 nameless assertions. It does not flag the shadowed class.

## Scope model

Class shadows module shadows conftest, and this is the load-bearing part. The first version pooled every fixture in a file into one dict and every sync test into one list — smaller, and wrong twice over:

- **False positives.** It reported 16 sync tests in `test_geometry_detection.py`, whose `_skip_no_db` autouse fixture is class-level in `TestConstructPointGeometry`/`TestConstructWktGeometry` while those tests live in `TestDetectGeometryColumns` and never see it. Two false alarms on a clean main is how a gate gets switched off.
- **A hidden true positive.** Two fixtures in one file can share a name — a class-level sync override of a module-level async fixture is exactly the remedy this gate recommends — and a flat dict keeps only the last.

Both are recorded inline, because the pooled version is the obvious simplification and it looks correct.

## No allowlist

pytest fails this shape every time, so there is no legitimate instance to exempt.

## Notes

- Clean on `main` today. Adopting it now costs nothing, and that stops being true once #1030 lands.
- #1030 cannot merge in its current form regardless: rebased onto current main it fails its own CI, because main is where the sync tests are. So this is not racing #1030 — it can land first and #1030 will have to satisfy it.
- Static AST, so a fixture pulled through `request.getfixturevalue()` is invisible. Unused in this suite today, documented in the module docstring as the direction the gate fails quiet.
- The second test pins the detector against the #1030/#1057 shape, since a gate that passes on a clean tree is indistinguishable from one that detects nothing.

## Verification

- `uv run pytest tests/test_fixture_scope_structural.py` — 2 passed
- `uv run pytest -m architecture` — 40 passed, 1 skipped
- A/B above, run against the real batch-4 file rather than a reconstruction, then restored
- `ruff check` / `ruff format --check` — clean

### DIFF
--- a/backend/tests/test_fixture_scope_structural.py
+++ b/backend/tests/test_fixture_scope_structural.py
@@ -1,6 +1,6 @@
 """Structural gate: no sync test may depend on an async fixture.
 
-fix(#1030/#1057): a sync test that reaches an async fixture does not merely
+fix(#1082): a sync test that reaches an async fixture does not merely
 fail itself. It poisons that fixture's FixtureDef for the rest of the xdist
 worker's session, and every later test that touches the fixture errors at
 setup with a bare `AssertionError`. One such test produced 951 errors in a
@@ -36,6 +36,11 @@ existed on its branch. #1057 then added two sync tests to a different class in
 the same module. Both PRs were green alone. The remedy was a per-class
 enumeration that was complete when written and silently incomplete a week
 later, so the thing worth checking in is the rule, not another entry.
+
+Sibling gates in the same neighbourhood, same idea: `test_rule1_structural.py`
+(visibility filters) and `test_rule2_structural.py` (GDAL/rasterio safe envs).
+This file deliberately fixes nothing it finds -- the first violation belongs
+to #1030.
 
 Known gap: this reads the AST, so it sees fixtures requested by parameter name
 and nothing else. A fixture pulled at runtime through
@@ -80,9 +85,22 @@ class _Scopes:
     """Fixtures, autouse names, and sync tests, keyed by scope.
 
     Scope is `_MODULE_SCOPE` or a class name. A class-level definition shadows
-    a module-level one of the same name, which shadows conftest -- getting this
-    wrong pools every class's fixtures together and reports sync tests as
-    exposed to autouse fixtures that never applied to them.
+    a module-level one of the same name, which shadows conftest.
+
+    fix(#1082): keying by scope is the load-bearing part of this file and it
+    reads like ceremony, so it is worth saying what happens without it. The
+    first version of this gate pooled every fixture in a file into one dict and
+    every sync test into one list. That is smaller and it is wrong: it reported
+    16 sync tests in `test_geometry_detection.py`, whose `_skip_no_db` autouse
+    fixture is class-level in `TestConstructPointGeometry` and
+    `TestConstructWktGeometry` while those tests live in
+    `TestDetectGeometryColumns` and never see it. Two false positives on a
+    clean main, which is how a gate gets switched off.
+
+    The same pooling also hid a true positive: two fixtures in one file can
+    share a name (a class-level sync override of a module-level async fixture
+    is exactly the remedy this gate recommends), and a flat dict keeps only the
+    last one.
     """
 
     def __init__(self, tree: ast.Module) -> None:
--- a/backend/tests/test_fixture_scope_structural.py
+++ b/backend/tests/test_fixture_scope_structural.py
@@ -1,0 +1,243 @@
+"""Structural gate: no sync test may depend on an async fixture.
+
+fix(#1030/#1057): a sync test that reaches an async fixture does not merely
+fail itself. It poisons that fixture's FixtureDef for the rest of the xdist
+worker's session, and every later test that touches the fixture errors at
+setup with a bare `AssertionError`. One such test produced 951 errors in a
+single CI run, none of which named the test that caused them.
+
+The mechanism is in pytest itself (9.1.1, `_pytest/fixtures.py`):
+
+    execute()  1222  self.addfinalizer(...)          # finalizer registered
+    execute()  1232  pytest_fixture_setup(...)       # raises at 1320, which is
+                                                     # BEFORE the try at 1327
+                                                     # that sets cached_result
+    finish()   1147  if self.cached_result is None:  # "Already finished. It is
+                         return                      #  assumed that finalizers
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review

ishiland:
@codex review

PR #1084, opened 2026-07-31
ci: integration batch 6 (#1030, #1044, #1079, #1083)

ishiland:
Integration batch: four PRs merged onto main (`4d2323286`) so the combination gets one CI run instead of four sequential update-branch cycles.

| PR | head |
|----|------|
| #1030 | `58624f28d8` |
| #1044 | `034c0f08d1` |
| #1083 | `7e6fc1eca0` |
| #1079 | `75aa3017ca` |

**This batch exists to prove a specific pair.** #1030 is the fix for what broke batch 4 (a module-level autouse **async** fixture reaching the two **sync** tests #1057 put on main), and #1083 is an independent structural gate for that same class of defect. Verified together in one tree rather than separately: the gate passes against the fix, and the two tests that produced 951 errors in batch 4 now pass.

Local gates on the merged tree before pushing:

- `test_layering.py` + `test_fixture_scope_structural.py`: 38 passed
- the two batch-4 casualties: 2 passed
- alembic: 1 head
- `ruff check` clean, 866 files formatted

**Two collisions were found in pre-flight and kept out of this batch:**

- **#1022 conflicts with #1044** on `backend/tests/test_layering.py`. Each merges cleanly onto main *alone*; only the combination conflicts. They cannot share a batch, because the resolution would live in a branch that gets deleted and the same conflict would return at the second merge with nobody having decided it. #1044 is here; #1022 follows once it rebases.
- **#1037 conflicts with main** on `backend/tests/test_ai_chat_run_analysis.py`, which #1071 touched in batch 5. It needs a rebase before it is a candidate.

Green: the four merge individually with `--admin`, each keeping its own squash commit and `Closes #NNN`, and this branch is deleted. Red: the failure names the member, as it did on batch 4.

Coverage qualifier: this is a `pull_request` event, so `accessibility`, `stac-validate` and the full `e2e-test` suite do not run here. A green result means the PR-visible combination passes; those three meet the merged combination for the first time on main's own push run.


ishiland:
All four merged individually with `--admin`:

| PR | commit |
|----|--------|
| #1030 | `5e3a553c1` |
| #1044 | `9d3250ff0` |
| #1079 | `710655069` |
| #1083 | `bddfb8d1d` |

29 SUCCESS / 9 SKIPPED / 0 FAILURE, with `Backend Tests`, `Frontend Tests` and `E2E Smoke (core)` each confirmed SUCCESS by name. Heads re-verified as ancestors of the tested branch immediately before merging. Branch deleted.

The pair this batch existed to prove landed together: #1030's fix and #1083's independent gate for the same class of defect, verified in one tree rather than separately.

**Correction to this PR's body.** It said `accessibility`, `stac-validate` and the full `e2e-test` suite would meet the combination on main's push run. That is true of the first two and false of the third. `e2e-test` gates on `schedule || workflow_dispatch` with no push clause, so it runs nightly and never on a merge. Verified on main's push run for batch 5, where `STAC API conformance` and `Accessibility (axe + keyboard-nav)` both succeeded while `E2E Tests` was skipped. The full E2E suite is a strictly larger gap than the other two, not another instance of them.

PR #1086, opened 2026-07-31
ci: integration batch 7 (#1037, #1056)

ishiland:
Integration batch: two PRs merged onto main (`bddfb8d1d`).

| PR | head |
|----|------|
| #1037 | `a3d186137c` |
| #1056 | `c4202ea68e` |

Both were held until their own `Backend Tests` completed **SUCCESS** before this branch was built, so a failure here means the combination rather than a member.

**#1027 was dropped after the first run of this branch went red**, and the reason is worth recording because it is a CI gap rather than a bug in that PR.

#1027 adds `CYCLE_IS_WEEKLY` to `scripts/backup-entrypoint.sh`, set by the cycle function and read inside `backup_staging`. That is correct in production, where bash globals are visible to callees. But `backend/tests/test_backup_staging_tar_skew.py` (on main since #844) extracts `backup_staging` alone and runs it under `set -euo pipefail`, so the caller never runs and `-u` makes the unset variable fatal:

    2 failed, 6659 passed

**#1027's own CI could not have caught this: `Backend Tests` was SKIPPED there.** It touches only `scripts/` and `RUNBOOK.md`, which does not trip the `backend` path filter — and the filter's premise is false for this file, because a backend test reads it. Tracked as its own issue.

Local gates on the merged tree: ratchet + fixture gate 38 passed, `test_a_grant_to_an_empty_role_does_not_block` passed (the batch-3 casualty that #1056 fixes), alembic 1 head, ruff clean.

Coverage qualifier: this is a `pull_request` event, so `accessibility` and `stac-validate` first meet the combination on main's push run. **No E2E of any kind runs on a merge to main** (`e2e-test` = `schedule || workflow_dispatch`, `e2e-smoke` = `(pull_request && changes.e2e) || merge_group`), so E2E Smoke on this PR is the only pre-merge E2E signal this combination gets.


ishiland:
Both merged individually with `--admin`:

| PR | commit |
|----|--------|
| #1037 | `df57e6dfe` |
| #1056 | `2aa158407` |

29 SUCCESS / 9 SKIPPED / 0 FAILURE, with `Backend Tests` and `E2E Smoke (core)` each confirmed SUCCESS **by name**. Heads re-verified as ancestors of the tested branch immediately before merging. Branch deleted.

#1027 stays out pending #1088.

PR #1090, opened 2026-07-31
fix(ci): run backend tests when a root script they read changes

### LINKED ISSUE #1088: ci: backend path filter misses scripts/ files that backend tests read
## Problem

`backend/tests/test_backup_staging_tar_skew.py` reads `scripts/backup-entrypoint.sh`, extracts the `backup_staging` shell function, and runs it. So a change to that script can break backend tests.

But `scripts/` does not trip the `backend` path filter, so `Backend Tests` is SKIPPED on a PR that only touches it. The test that covers the file never runs on the PR that changes the file.

## How it surfaced

#1027 adds `CYCLE_IS_WEEKLY` to `backup-entrypoint.sh`, set by the cycle function and read inside `backup_staging`. Correct in production: bash globals are visible to callees. The test runs `backup_staging` standalone under `set -euo pipefail`, so the caller never executes and `-u` makes the read fatal:

```
```

#1027's own CI was green with `Backend Tests -> SKIPPED`. It only failed once an integration batch put it in a tree where something else forced the backend suite to run.

## Why this is the interesting kind of gap

The path filter is right nearly always: a `scripts/` change genuinely cannot affect backend tests. It is wrong for this one file, because a backend test reads it. The convention is sound and its precondition silently stops holding for a specific path, which is the same shape as several other defects found this week.

Note `CI OK` treats SKIPPED as pass, correctly — a path-filtered PR should not be blocked on jobs its diff cannot affect. That is exactly why nothing surfaced here: the skip was legitimate under a premise that happened to be false.

## Proposed

Two independent changes, either of which closes it:

1. Add the shell sources that backend tests read to the `backend` path filter. At minimum `scripts/backup-entrypoint.sh`. Worth grepping `backend/tests/` for other reads of files outside `backend/`, since this is unlikely to be the only one.
2. Make `backup_staging` independent of caller state: read `${CYCLE_IS_WEEKLY:-0}`. A function extracted and run in isolation by its own test should not depend on a global the caller sets, and the default matches the non-weekly path.

(2) fixes this instance and is the smaller change. (1) is what stops the next one, since the coupling is structural: any test that reads a file outside its own filter's paths has this hole.

## Related

- #1027 is blocked on this; it was dropped from integration batch #1086 after the first run went red.
- Same family as #1082 (a correct enumeration that silently stopped being complete).

ishiland:
## What

The `backend` paths-filter in `ci.yml` had no `scripts/` entries, so a PR touching only a repo-root shell script skipped Backend Tests.

#1027 is the live case. It changes `scripts/backup-entrypoint.sh`. The `backup` filter matched, so Backup Restore Round-trip ran and passed. The `backend` filter did not, so `test_backup_staging_tar_skew.py` — the backend test written to catch a break in that exact script — never ran. Its checks read as clean, because a skipped required job is not a failing one.

## Change

Two root scripts added to the `backend` filter: `scripts/backup-entrypoint.sh` and `scripts/init-db.sh`. Files under `backend/scripts/` need no entry, since `backend/**` already covers them.

The guard goes in `test_ci_alembic_filter_paths.py`, which already parses `ci.yml`'s filters for the alembic case:

- **Coverage.** Every `scripts/` path a backend test names by literal path must appear in the `backend` filter. The next test that reads a root file fails until someone adds it.
- **Dead globs.** No glob in the `backend` filter may match zero files.

## Verification

A/B on the coverage test: reverting only the `ci.yml` hunk fails it, naming `scripts/backup-entrypoint.sh (read by test_backup_staging_tar_skew.py)` and `scripts/init-db.sh`. With the hunk, 6 passed.

The dead-glob test earned its place twice. It rejected the first draft of this change, which added `scripts/api-entrypoint.sh` and `scripts/check_fe_type_drift.py` as root paths when both actually live under `backend/scripts/`. It then found a pre-existing `frontend/Dockerfile` entry, left behind when INF-10 deleted that file and `test_phase_272_dockerfile.py` began asserting it stays deleted. Repo-wide audit across every filter found no others.

Also run: `test_phase_272_dockerfile.py`, `test_phase_272_compose.py`, `test_backup_staging_tar_skew.py`, `test_docker_audit_regressions.py`, `test_ci_merge_queue.py` (91 passed) and `test_layering.py` (36 passed).

This PR edits `ci.yml`, which is itself in the `backend` filter, so Backend Tests runs here and validates the change against the suite it gates.

## Impact

No schema, API, or config surface. CI-only. Unblocks #1027, whose script fix still belongs to that PR — once this lands, its Backend Tests job will run and report honestly.

Closes #1088

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,7 +103,9 @@ jobs:
               # nginx/CSP/Dockerfile regressions). Specific files only — NOT
               # frontend/src/** — so ordinary frontend PRs stay backend-skipped.
               - 'frontend/nginx.conf'
-              - 'frontend/Dockerfile'
+              # fix(#1088): 'frontend/Dockerfile' was here and matched nothing —
+              # INF-10 deleted that file and test_phase_272_dockerfile.py keeps
+              # it deleted. A dead glob is silent in both directions.
               - 'frontend/Dockerfile.dev'
               - 'frontend/docker-entrypoint.sh'
               - 'frontend/index.html'
@@ -112,6 +114,16 @@ jobs:
               - 'frontend/src/lib/basemap-utils.ts'
               - 'frontend/src/lib/capabilities.ts'
               - 'frontend/src/components/builder/layer-adapters/shared.ts'
+              # fix(#1088): repo-root scripts that backend tests read by literal
+              # path. #1027 changed backup-entrypoint.sh and its own regression
+              # test (test_backup_staging_tar_skew.py) never ran — the backup
+              # filter covered the script, but nothing pulled in the backend
+              # suite that asserts against it. Backup Restore Round-trip went
+              # green while the test written to catch that break sat skipped.
+              # backend/scripts/* needs no entry here; backend/** covers it.
+              # test_ci_alembic_filter_paths.py now enforces both directions.
+              - 'scripts/backup-entrypoint.sh'
+              - 'scripts/init-db.sh'
               # fix(#561): a change to this workflow can alter HOW Backend Tests
               # runs (e.g. the -n4 parallelization / coverage core), so the suite
               # must run on PRs that touch it — otherwise CI-config changes ship
--- a/backend/tests/test_ci_alembic_filter_paths.py
+++ b/backend/tests/test_ci_alembic_filter_paths.py
@@ -1,4 +1,4 @@
-"""OCG-02: Alembic paths-filter guard test.
+"""OCG-02: ci.yml paths-filter guard tests.
 
 Parses the alembic filter list from .github/workflows/ci.yml and asserts:
 1. Every glob in the alembic filter matches at least one real file on disk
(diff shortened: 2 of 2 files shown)


ishiland:
Correction to the framing in this PR's body, so the record is accurate for anyone checking it now.

The body says #1027's Backend Tests skipped. That was true when I observed it, at head `cdb9e9cd8`. It currently reads `pass`, and the reason has nothing to do with this PR being unnecessary.

`paths-filter` runs with `token: ''` and therefore diffs the PR's **recorded** `base.sha` against the recomputed merge ref. #1027's base is 21 commits behind main, so its filter diff picked up 117 files, nearly all of them other people's merged `backend/**` work, and that is what turned Backend Tests on. Confirmed exactly: diffing `f0113cdfa` against `merge-tree(06b32dcd6, a1c033382)` yields 117 files, matching what the run logged.

So #1027's job flipped from `skipping` to `pass` without its own content changing, purely because main moved underneath it. Filed separately as #1094.

The gap this PR fixes is unaffected. A PR that changes only `scripts/backup-entrypoint.sh` and is up to date with main has a four-file diff, matches nothing in the `backend` filter, and skips the suite that tests that script. That is what #1027 did earlier today, and it is what any future backup-script PR would do. If anything #1094 sharpens the case: without this filter entry, the coverage such a PR gets depends on how stale it happens to be.

Issue #951, opened 2026-07-30
ux(builder): legend chip and layer-list swatch show a solid color for a patterned fill layer


ishiland:
Split out of #914, which now covers pattern tinting only. This half needs no color source and has no dependency on the rest of the fill-pattern cluster, so it can ship on its own.

Found in the 2026-07-30 style-tab audit, verified on main @ 28e205ef2.

## What happens

Apply a fill pattern to a polygon layer. The map draws hatching, but the legend chip and the layer-list swatch keep showing a solid color chip. The legend no longer describes what is on the map, and for a data-driven patterned layer every class in the legend shows a solid color that appears nowhere.

## Why

Both swatch surfaces derive from color config only, with no awareness of `paint['fill-pattern']`:

- `GeometrySwatch` (`frontend/src/components/map/LegendEntries.tsx:31-78`) renders the polygon case as a plain `backgroundColor` div at `:63-77`. It is shared by `CategoricalLegend` (`:88-99`) and `GraduatedColorLegend` (`:110-121`).
- `getLayerColors` (`frontend/src/components/map/layer-icons.tsx:246-260`) reads the geometry's color property and returns a color array. When a pattern click has deleted `fill-color` it falls through to `MAP_COLORS.icon.fallback` at `:259`, so the chip can show a color the layer does not use.

## Fix

When `paint['fill-pattern']` is set, overlay the same CSS pattern preview the picker already uses on the legend chip and the layer-list swatch. `patternPreviewStyle` (`frontend/src/components/builder/FillPatternPicker.tsx:22-58`) is a pure function returning a `React.CSSProperties` with `repeating-linear-gradient` and `radial-gradient` backgrounds keyed by pattern id, and it uses `currentColor`, so it inherits whatever color the chip is already drawing.

It currently lives in the picker module and is not exported. Lift it to a shared module (or export it in place) so the legend and the layer icons can both use it without importing from a builder panel component.

Consumers to update: `GeometrySwatch` in `LegendEntries.tsx` (polygon branch), and the layer-list swatch path through `getLayerColors` / `ColorizedGeometryIcon` in `layer-icons.tsx`. `LegendEntries` and `layer-icons` are shared by the builder legend plugin and the viewer legend, so fixing them covers both surfaces at once.

## Acceptance

- Legend chip for a patterned polygon layer shows the pattern, not a solid chip. Same for the layer-list swatch.
- A patterned data-driven layer shows the pattern on each class chip, in that class's color.
[...]

PR #979, opened 2026-07-30
ux(builder): show the fill pattern on the legend chip and layer-list swatch

ishiland:
## What changed

Apply a fill pattern to a polygon layer and the map draws hatching, but the legend chip and the layer-list swatch kept showing a solid colour. Both surfaces derived from colour config with no awareness of `paint['fill-pattern']`, so the legend stopped describing the map — and on a data-driven patterned layer every class chip showed a colour that appears nowhere.

- `patternPreviewStyle` moves out of `FillPatternPicker` into `src/lib/fill-pattern-preview.ts`, alongside a small `fillPatternFromPaint` reader. The picker imports it back; nothing about its rendering changes.
- `SwatchStyle` gains `fillPattern`. `GeometrySwatch`'s polygon branch draws the pattern over a transparent background with `color` set to the chip's colour, so each class chip patterns in its own colour. Point and line branches are untouched.
- `StyleHints` gains `fillPattern`, read in `extractStyleHints`' polygon branch, so every `ColorizedGeometryIcon` caller gets it without new wiring.
- The three `SwatchStyle` builders (`LegendPlugin`, viewer `LayerLegend`, builder `stylePreviewStyle`) pass the pattern through.

One deliberate visual call: for a patterned polygon the layer-list swatch becomes a square pattern chip rather than a patterned pentagon. Lucide's pentagon has no fill surface a CSS gradient can reach, and patterning it would mean re-authoring all five patterns as SVG defs — a second source of truth for the same five patterns. The square chip matches the picker and the legend chip. Unpatterned layers keep the pentagon.

Scope, per the issue: display only. A patterned layer whose `fill-color` was deleted may still show a fallback colour under the pattern; that resolves with #910. Pattern tinting is #914.

No schema, API, or security impact.

## Verification

    cd frontend && npm run lint && npm run typecheck && npx vitest run src/components/map/__tests__/LegendEntries.test.tsx src/components/map/__tests__/layer-icons.test.tsx

Full suite also run: 358 files / 4162 tests pass. Five new pins cover the patterned and unpatterned paths on both surfaces plus the `extractStyleHints` read.

Closes #951

https://claude.ai/code/session_01Gjs3zyCByCgjukVrg7DVZv

PR #1091, opened 2026-07-31
feat(builder): tint fill patterns with the layer fill colour

### LINKED ISSUE #914: ux(builder): fill patterns ignore the fill color and are nearly invisible at default opacity

ishiland:
Closes #914 (the tinting half; the legend/swatch mismatch was #951, shipped in #979).

## What was wrong

The five pattern tiles were generated with a hardcoded `rgb(80,80,80)`, and `makeFillPatternImage(id)` took no colour at all. Meanwhile the picker swatches preview each pattern in `currentColor`, so the control promises the pattern will follow the fill colour. It never did — and at the default 0.3 fill opacity a 1px grey line on a light basemap reads as "my fill disappeared".

## What changed

The generators take an `rgb`, and tinted variants register lazily under colour-namespaced ids (`geolens-fill-hatch#1d4ed8`) the first time a layer needs one. The plain ids stay registered, so any layer without a resolvable tint — and every existing saved map — renders exactly as before.

**The swap happens on the way into `addLayer` / `setPaintProperty`, nowhere else.** A tinted id is a MapLibre registry key, not a style value: saved `paint`, the wire format and exported `style.json` all keep the plain id, because a tinted id names an image that exists only inside the one browser session that had that layer open. The issue asks for a test rather than a comment on this and it gets three — the two adapter paths assert the input paint object is untouched after the call, and a live save confirms it end to end. It matters because #917 already ships builtin pattern ids the served sprite does not contain; a leaked tinted id would make that strictly worse.

**One tint resolver, four consumers.** `fillPatternTint(paint, builder)` returns the fill colour from `paint` when it holds a string (older clients that carry both keys), else #910's `fillColorSaved` stash, which is where the colour lives once a pattern owns the fill. The map adapter, the legend chip (builder and viewer), the layer-list swatch and the style-editor header swatch all read it. That is the whole point of doing it this way: a surface left deriving its own colour would draw something the map does not, which is the #951 class of bug this enhancement could otherwise have reintroduced.

The mixed-geometry adapter goes through the same swap. It renders a fill sublayer for `GEOMETRY` / `GEOMETRYCOLLECTION` layers and would have been the one surface still drawing grey.

**The resolver rejects a stash it cannot use.** `builder.fillColorSaved` is declared `string` but arrives from an open `style_config` that gets serialized-size validation only, so an API-authored or imported layer can hold a number or object there. Passing one through fed a junk tint to `ensureTintedFillPatternImage`, and its throw is swallowed by `fillAdapter.addLayers`' catch — so the whole layer silently failed to build, not merely its tint. Found only by composing this branch with #1022, whose own junk-stash tests went from green to `expected undefined to be defined`; neither branch's suite could see it alone. Pinned now by direct `fillPatternTint` unit tests over `42` / `{r:1}` / `['#fff']` / `true` / `null`.

## Interaction with the export strip (#1054 / #917)

#1054 strips a builtin `fill-pattern` from the exported `style.json` and emits a solid fill instead, because those five images are generated in the browser and are not in the served sprite. That is the right pairing for this change rather than a conflict with it: in-app the pattern draws tinted, and the exported document falls back to a solid fill **of the same colour**, since both sides resolve it from `fillColorSaved` — the tint via `fillPatternTint`, the export via the seed added in #1022. A sprite could not have carried the tint anyway; it is one shared atlas and the tint is per-layer, which is exactly the reason #1054 documents for not baking the generators into it.

## Not in scope

The tile line weight. Patterns are still 1px at the layer's fill opacity, so a tinted hatch at the 0.3 default is more legible than grey but still subtle — the issue's "2px at full alpha" alternative was the descope path if tinting did not happen, and changing it now would restyle every existing patterned map. Worth its own decision if the tint alone does not settle the visibility half of the report.

## Verification

```
# 369 files, 4486 tests pass

npx vitest run src/components/builder/layer-adapters/__tests__/fill-pattern-images.test.ts \
  src/components/builder/__tests__/layer-adapters.test.ts \
  src/lib/__tests__/fill-pattern-preview.tint.test.ts \
  src/components/map/__tests__/LegendEntries.test.tsx \
  src/components/map/__tests__/layer-icons.test.tsx
```

New tests: tinted-id namespacing, case normalisation, 3-digit hex expansion, idempotency, the pixel colour actually changing, and the four fallbacks (no tint / non-hex / unknown id / `addImage` throwing); the never-persisted invariant on both adapter paths; the tint agreement on the legend chip and layer-list swatch, including that `extractStyleHints` resolves the colour from the stash; `fillPatternTint` directly, including every non-string stash shape; and one test that a **pasted** pattern tints the colour the user copied rather than the target's old one. It began as a cross-branch check against #1022 (now merged as 06b32dcd6): without that PR's fill-colour drop, `fill-color` survives in paint, `fillPatternTint` prefers it, and the map plus all three legend surfaces silently agree on the wrong colour (`expected '#0000ff' to be '#ff0000'`). It stays as the regression guard for that pairing.

Live, on a red polygon layer over the dev stack:

- hatch on a red layer draws **red** hatching (the same layer hatched near-invisible grey before),
- it survives save and reload, where the only colour left is the stash (`paint` is `{fill-opacity, fill-pattern}`),
- saved `paint` carries `geolens-fill-hatch`, not the tinted id,
- legend chip and layer-list swatch draw the same red,
- zero console errors.

The **viewer** surface is verified the same way, unauthenticated through a share token (`/m/{token}`): the polygons draw red hatching, the legend chip's computed style resolves to `rgb(255, 0, 0)` over a red `repeating-linear-gradient`, and there are no console errors. The saved layer at that point carries `{fill-opacity, fill-pattern}` with no `fill-color` at all, so the only place that red can have come from is the `fillColorSaved` stash — which is the whole point of routing every surface through one resolver.

### DIFF
--- a/frontend/src/components/builder/LayerStyleEditor/utils.ts
+++ b/frontend/src/components/builder/LayerStyleEditor/utils.ts
@@ -1,5 +1,5 @@
 import { MAP_COLORS } from '@/lib/map-colors';
-import { fillPatternFromPaint } from '@/lib/fill-pattern-preview';
+import { fillPatternFromPaint, fillPatternTint } from '@/lib/fill-pattern-preview';
 import type { BuilderStyleConfig, MapLayerResponse, StyleConfig } from '@/types/api';
 
 // ---------------------------------------------------------------------------
@@ -129,6 +129,7 @@ export function stylePreviewStyle(layer: MapLayerResponse) {
       opacity: layer.opacity,
       fillOpacity: typeof paint['fill-opacity'] === 'number' ? paint['fill-opacity'] as number : undefined,
       fillPattern: fillPatternFromPaint(paint),
+      fillPatternColor: fillPatternTint(paint, layer.style_config?.builder),
       strokeWidth: typeof layer.style_config?.builder?.outlineWidth === 'number'
         ? layer.style_config.builder.outlineWidth
         : typeof paint['_outline-width'] === 'number'
--- a/frontend/src/components/builder/__tests__/layer-adapters.test.ts
+++ b/frontend/src/components/builder/__tests__/layer-adapters.test.ts
@@ -1540,10 +1540,72 @@ describe('fillAdapter', () => {
     const input = makeInput({
       id: 'fp4',
       layerId: 'layer-fp4',
-      paint: { 'fill-color': '#3b82f6', 'fill-pattern': 'geolens-fill-hatch' },
+      paint: { 'fill-pattern': 'geolens-fill-hatch' },
     });
     fillAdapter.syncPaint(map, input);
     expect(map.setPaintProperty).toHaveBeenCalledWith('layer-fp4', 'fill-pattern', 'geolens-fill-hatch');
+  });
+
+  // ── fix(#914): the tint swap happens on the way into MapLibre only ──────────
+  it('syncPaint writes the TINTED pattern id when a fill colour is resolvable', () => {
+    (map.getLayer as ReturnType<typeof vi.fn>).mockImplementation((id: string) => {
+      if (id === 'layer-fp4t' || id === 'layer-fp4t-outline') return { id };
+      return null;
+    });
+    (map.getPaintProperty as ReturnType<typeof vi.fn>).mockReturnValue(undefined);
+    const paint = { 'fill-pattern': 'geolens-fill-hatch' };
+    const input = makeInput({
+      id: 'fp4t',
+      layerId: 'layer-fp4t',
+      paint,
+      style_config: { builder: { fillColorSaved: '#ff0000' } },
+    } as Partial<AdapterLayerInput> & { style_config: { builder: Record<string, unknown> } });
+    fillAdapter.syncPaint(map, input);
+
+    expect(map.setPaintProperty).toHaveBeenCalledWith(
+      'layer-fp4t', 'fill-pattern', 'geolens-fill-hatch#ff0000',
+    );
+    expect(map.addImage).toHaveBeenCalledWith('geolens-fill-hatch#ff0000', expect.anything());
+    // INVARIANT: the tinted id is a runtime registry key. The layer's own paint —
+    // what gets saved, sent over the wire and exported to style.json — is untouched.
+    expect(paint['fill-pattern']).toBe('geolens-fill-hatch');
+    expect(input.paint['fill-pattern']).toBe('geolens-fill-hatch');
+  });
+
+  it('addLayers keeps the plain id in the input paint while adding the tinted layer', () => {
+    const paint = { 'fill-pattern': 'geolens-fill-dots', 'fill-opacity': 0.3 };
+    const input = makeInput({
+      id: 'fp4a',
+      layerId: 'layer-fp4a',
+      sourceId: 'source-fp4a',
+      sourceLayer: 'data.test_table',
+      paint,
+      style_config: { builder: { fillColorSaved: '#00ff00' } },
+    } as Partial<AdapterLayerInput> & { style_config: { builder: Record<string, unknown> } });
+    fillAdapter.addLayers(map, input);
+
+    const fillCall = (map.addLayer as ReturnType<typeof vi.fn>).mock.calls
+      .find((c: unknown[]) => (c[0] as { type: string }).type === 'fill');
+    expect((fillCall![0] as { paint: Record<string, unknown> }).paint['fill-pattern'])
+      .toBe('geolens-fill-dots#00ff00');
+    expect(paint['fill-pattern']).toBe('geolens-fill-dots');
+  });
+
+  it('leaves the pattern untinted when no fill colour is resolvable', () => {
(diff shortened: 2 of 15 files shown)


ishiland:
@codex review

PR #1092, opened 2026-07-31
fix(builder): let a builder clear actually clear

ishiland:
No issue — this is a session finding from the #910/#918 work, split out because it is a separate defect with a separate cause. It is the same collapse #1022 (06b32dcd6) hit with the fill-colour stash, reaching three more controls.

## What was wrong

Every builder control writes through `updateBuilderConfig`. When a patch **empties** the builder, `withBuilderConfig` drops the builder key — collapsing the whole config to `null` once nothing else is left — and `handleStyleConfigChange` reads a missing builder as "keep the existing one". So the default merge put back exactly what the user had just cleared.

Three controls reached that state. Each was live-reproduced against main, not inferred:

- **Point layer, stroke toggle off then on.** The switch springs straight back off, and a save persists `stroke_disabled: true`.
- **Line layer, Gradient then Solid.** The mode button *looks* right, because it reads paint — but `builder.lineGradient` is still saved, and a reload brings the gradient back.
- **Polygon fill toggle.** Same shape, but a fresh polygon carries seeded `outlineColor`/`outlineWidth` that keep the config non-empty, so this third one is defensive rather than a live repro.

Reachability turns on geometry, which is why the first attempt at reproducing it failed: a fresh polygon seeds `builder: {outline_color, outline_width}` so the config never empties, while a fresh point (MULTIPOINT) and line (MULTILINESTRING) layer arrive with `builder: null`. The test only discriminated after switching to a line layer.

## The fix

Detect the emptying patch centrally and `replace` for that case only. Ordinary edits keep the merge, which re-reads the live layer's builder.

Central rather than per-caller because `onBuilderChange` is a **single prop shared by every builder control**: a per-call flag would have to be threaded through all of them, and would go stale the next time a control learns to clear a key. That is the same reasoning #1022 landed on for its paint rule, arrived at independently here.

## Verification

```
cd frontend && npm run typecheck && npm run lint
npx vitest run src/components/builder/__tests__/LayerStyleEditor.test.tsx   # 74 pass
```

Fail-before, re-measured against this branch's rebase onto `06b32dcd6` (main's `LayerStyleEditor.tsx` swapped in, tests kept):

```
× switching a gradient line back to Solid replaces, so lineGradient cannot return
× re-enabling fill replaces the config so fillDisabled cannot survive
× re-enabling stroke on a circle layer replaces the config too
3 failed | 71 passed
```

Worth re-stating because #1022 rewrote large parts of this file after the fix was first written: the three tests still discriminate on top of it.

### DIFF
--- a/frontend/src/components/builder/LayerStyleEditor.tsx
+++ b/frontend/src/components/builder/LayerStyleEditor.tsx
@@ -262,7 +262,23 @@ export const LayerStyleEditor = memo(function LayerStyleEditor({
   );
 
   const updateBuilderConfig = useCallback((patch: BuilderStyleConfig, nextPaint: Record<string, unknown> = paint) => {
-    onStyleConfigChange(layer.id, withBuilderConfig(layer.style_config, patch), stripLegacyBuilderPaint(nextPaint));
+    const nextConfig = withBuilderConfig(layer.style_config, patch);
+    const strippedPaint = stripLegacyBuilderPaint(nextPaint);
+    // A patch that EMPTIES the builder has to replace, not merge. withBuilderConfig
+    // drops the builder key once every field compacts away (collapsing the whole
+    // config to null when nothing else is left), and handleStyleConfigChange reads a
+    // missing builder as "keep the existing one" — so the default merge silently
+    // undoes the clear. That is the same collapse #1022 hit with the fill-color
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

PR #1093, opened 2026-07-31
ci(integration): batch 8 — #1090, #1092, #1027

ishiland:
Integration batch. **Do not merge this PR.** It exists to run CI once on the combination; each member is then merged individually so it keeps its own squash commit and its own issue link. This branch is deleted either way.

Base is `ffb7dd6ee`. Members, with heads pinned:

| PR | head | what |
| --- | --- | --- |
| #1090 | `e4f526c51` | backend paths-filter covers root scripts backend tests read |
| #1092 | `7b0e383f3` | a builder clear actually clears |
| #1027 | `a1c033382` | `pg_dumpall --globals-only` in the backup cycle |

## Why these three ride together

#1090 and #1027 are the pair that individual CI structurally cannot check. #1027 changes `scripts/backup-entrypoint.sh`, and on its own PR Backend Tests **skipped**, because the `backend` paths-filter had no `scripts/` entries. So `test_backup_staging_tar_skew.py`, which exists to catch a break in that exact script, never ran, and the PR read as ready. #1090 is the filter fix.

In this tree the filter change and the script change are present together, so `changes.backend` resolves TRUE off `scripts/backup-entrypoint.sh` and Backend Tests runs against it for the first time. Verified by evaluating the merged `ci.yml` filters against this branch's own diff.

## Pre-flight

Local, on the pinned base, pushing nothing:

- three-way merge clean, no conflict in combination
- `ruff check` and `ruff format --check` clean, 869 files
- 133 backend tests across the layering gate, both ci.yml filter guards, the backup skew test, and the dockerfile/compose/docker-audit suites
- `npm run typecheck` and `npm run lint` clean
- 4490 frontend tests

A pre-flight GO means this is worth a CI cycle. It is not a prediction that CI passes: it does not run the backend suite, which is the long pole and the thing this batch is actually here to exercise.

## Coverage this run does NOT have

This is a `pull_request` event, so `accessibility`, `stac-validate` and the full `e2e-test` suite do not run. Those first meet the combination on main's push run after the members land, and `e2e-test` never runs on a merge at all. Green here means the PR-visible combination is tested, not that the combination is tested.

Refs #1088

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,7 +103,9 @@ jobs:
               # nginx/CSP/Dockerfile regressions). Specific files only — NOT
               # frontend/src/** — so ordinary frontend PRs stay backend-skipped.
               - 'frontend/nginx.conf'
-              - 'frontend/Dockerfile'
+              # fix(#1088): 'frontend/Dockerfile' was here and matched nothing —
+              # INF-10 deleted that file and test_phase_272_dockerfile.py keeps
+              # it deleted. A dead glob is silent in both directions.
               - 'frontend/Dockerfile.dev'
               - 'frontend/docker-entrypoint.sh'
               - 'frontend/index.html'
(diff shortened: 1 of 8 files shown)


ishiland:
Batch landed. Closing without merging, as intended.

CI on the combination was green, with the required jobs confirmed by name rather than by tick:

```
Backend Tests             -> pass
Frontend Tests            -> pass
Frontend Lint & Typecheck -> pass
Backup Restore Round-trip -> pass
E2E Smoke (core)          -> pass
CI OK                     -> pass
```

`Backend Tests -> pass` is the line this batch existed to produce. On #1027's own PR that job was `skipping`, so `test_backup_staging_tar_skew.py` never ran against the script #1027 rewrites.

Members merged individually, each keeping its own squash commit and issue link:

| PR | squash | closed |
| --- | --- | --- |
| #1090 | `2b17212a3` | #1088 |
| #1092 | `607e00eb9` | none by design |
| #1027 | `a90ed904a` | #995 |

Three PRs for one CI cycle instead of three.

One finding came out of assembling this. #1027's Backend Tests flipped from `skipping` to `pass` between the two times I read it, with no change to its content, because `paths-filter` diffs a PR's recorded `base.sha` rather than a merge-base. Its four-file diff computed as 117 files. Filed as #1094.

PR #1095, opened 2026-07-31
fix(ci): diff a PR against its merge base, not its frozen base sha

### LINKED ISSUE #1094: Path filters diff against a PR's stale recorded base, so coverage shrinks as a branch gets fresher
## What

`changes` (the `dorny/paths-filter` job in `ci.yml`) runs with `token: ''`, which switches it from the GitHub API to git-diff detection. Per the `fix(#546)` comment that is deliberate: the API call's transient 503s were failing a required check hard.

The side effect is that it diffs the PR's **recorded `base.sha`** — the base as of when the PR was opened — against the freshly recomputed `refs/pull/N/merge` ref. GitHub rebuilds that merge ref every time main moves, but `base.sha` never moves. So the diff is:

```
base.sha .. (current main + this PR)
```

which is the PR's own changes **plus every commit that has landed on main since the PR was opened**.

## Evidence

#1027 changes four files: `RUNBOOK.md`, `scripts/backup-entrypoint.sh`, `scripts/restore.sh`, `scripts/tests/test-backup-restore-roundtrip.sh`.

Its `Detect Changes` job on run `30668997766` logged:

```
GitHub token is not available - changes will be detected using git diff
Detected 117 changed files
```

The 117 include `backend/app/**`, `frontend/src/**`, `sdks/**`, and `backend/alembic/versions/0033_records_derived_from.py`, none of which the PR touches. Its base was 21 commits behind main.

Reproduced exactly:

```
$ git merge-tree --write-tree 06b32dcd6 a1c033382   # main at the time + the PR
$ git diff --name-only f0113cdfa <that tree> | wc -l
117
```

117 predicted, 117 reported.

## Why this is worse than wasted minutes

The obvious cost is that stale PRs run jobs they have no business running. The real problem is the direction of the bias.

- **A green job can be about someone else's code.** #1027's Backend Tests reported `pass`. That was not the backend suite vetting #1027; it was the suite running because unrelated `backend/**` changes were in the diff.
- **Coverage is inversely proportional to freshness.** The more up to date a branch is, the smaller its computed diff, and the fewer jobs run. A PR that update-branches onto main immediately before merging gets the *narrowest* filter result it will ever have, at exactly the moment the result matters most. That is backwards from what path filtering is for.
- **It makes filter gaps hard to see.** #1027 showed `Backend Tests -> skipping` earlier in the day, when its base was closer to main, and `-> pass` later, after main moved. Same PR, same four files, opposite verdicts, driven entirely by how far behind it had drifted.

## Note on #1090

#1090 (backend filter missing root `scripts/`) is a genuine and separate gap and still needs to land. It was found when #1027 showed `Backend Tests -> skipping`, which was the honest result for its content. This issue explains why that PR later flipped to `pass` without anything relevant changing, and it means the flip should not be read as #1090 being unnecessary.

## Directions

Whatever the fix, it should keep #546's property of not depending on a GitHub API call during a required check.

- Give paths-filter an explicit `base` so it computes a merge-base rather than using the recorded `base.sha`, with `fetch-depth: 0` on the checkout.
- Or diff `merge-base(main, head)..head` directly and feed the result in.

Worth a test in `test_ci_alembic_filter_paths.py`, which already parses this workflow, asserting the `changes` job cannot silently regain a stale-base diff.

ishiland:
## The bug

`paths-filter` runs with `token: ''` (fix(#546), to keep a required check off a flaky API call), which puts it on git-diff detection. That path diffs `pull_request.base.sha .. merge-commit`.

`base.sha` is frozen when the PR is opened. GitHub rebuilds `refs/pull/N/merge` every time main moves. So the diff accumulates everything that landed in between.

#1027 changed four files and its filter reported 117:

```
GitHub token is not available - changes will be detected using git diff
Detected 117 changed files
```

The extras were other people's `backend/**` work, and that is what switched Backend Tests on. Reproduced exactly: `git diff f0113cdfa <merge-tree(06b32dcd6, a1c033382)>` gives 117 files.

## Why it matters more than wasted minutes

A green job can be entirely about someone else's code. #1027's `Backend Tests -> pass` was not the suite vetting #1027.

Worse, coverage ran **inverse to freshness**. The closer a branch was to main, the narrower its computed diff and the fewer jobs ran. Update-branching immediately before merging handed a PR the least filtering it would ever get, at the moment the result mattered most. #1027 read `skipping` earlier in the day and `pass` later, same four files, driven only by drift.

## The fix

`checkout` leaves the job on `refs/pull/N/merge`, whose first parent is the base tip as of when that ref was built. `HEAD^1..HEAD` is therefore the PR's own content and nothing else.

It needs `fetch-depth: 2`. A depth-1 clone grafts the checked-out commit as a shallow boundary and git reports it as **parentless**, so `rev-parse HEAD^1` dies with `ambiguous argument`. Depth 2 materialises the merge ref's two parents, which is all this needs; `paths-filter`'s `ensureRefAvailable` then finds the base locally instead of re-fetching.

Untouched on purpose: `push` keeps the action's before-SHA default, `merge_group` keeps `merge_group.base_sha`. Both are already right, and the queue skips this step entirely.

One trap worth recording. The action logs `'base' input parameter is ignored when action is triggered by pull request event`. That warning is wrong on this path: `main.ts:117` resolves `base || baseSha || defaultBranch`, so an explicit base wins. Only the API path discards it. Read at the pinned sha, not from the docs.

## Verification

The first push of this branch asserted in a code comment that no extra fetch depth was needed. That was wrong, and this PR's own `Detect Changes` job failed on it:

```
fatal: ambiguous argument 'HEAD^1': unknown revision or path not in the working tree
Required jobs failed or were cancelled: changes
```

Fixed in `e5b8e6ad9`. The check that matters is this PR's own `Detect Changes` log reporting a file count matching what it actually changes, rather than everything since branch point, and it is the only place the fix executes.

Locally: 8 tests in `test_ci_alembic_filter_paths.py`, with A/B on the new guard — deleting the `base:` line fails it with the stale-base explanation. `ci.yml` parses at 25 jobs with all five triggers intact.

## Guard

Two assertions in `test_ci_alembic_filter_paths.py`, which already parses this workflow:

- the filter step must carry a `base` wired to a step id that exists in the job
- the resolver must read `HEAD^1` and must not reach back into the payload for `base.sha`

Closes #1094

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,6 +67,13 @@ jobs:
       backup: ${{ steps.filter.outputs.backup }}
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+        with:
+          # fix(#1094): the default depth-1 clone grafts the checked-out commit
+          # as a shallow boundary, and git then reports it as having NO parents:
+          # `git rev-parse HEAD^1` dies with "ambiguous argument". Depth 2 is the
+          # minimum that materialises the merge ref's two parents, which is all
+          # the base resolution below needs.
+          fetch-depth: 2
       # fix(#1094): resolve the PR diff base ourselves. paths-filter's no-token
       # path diffs `pull_request.base.sha .. merge-commit`, and base.sha is
       # frozen at PR-open time while GitHub recomputes the merge ref every time
@@ -77,10 +84,9 @@ jobs:
       #
       # checkout leaves us on refs/pull/N/merge, whose FIRST parent is the base
       # branch tip as of when that ref was built. Diffing against it yields the
-      # PR's own changes and nothing else. rev-parse reads the parent out of the
-      # merge commit itself, so this needs no extra fetch depth, and
-      # paths-filter's ensureRefAvailable does `fetch --depth=1 origin <sha>`
-      # for the base it is handed.
+      # PR's own changes and nothing else. This needs the fetch-depth: 2 above;
+      # paths-filter's own ensureRefAvailable then finds the base locally rather
+      # than re-fetching it.
       - name: Resolve the PR diff base
         id: diffbase
         if: github.event_name == 'pull_request'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,6 +67,28 @@ jobs:
       backup: ${{ steps.filter.outputs.backup }}
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+      # fix(#1094): resolve the PR diff base ourselves. paths-filter's no-token
+      # path diffs `pull_request.base.sha .. merge-commit`, and base.sha is
+      # frozen at PR-open time while GitHub recomputes the merge ref every time
+      # main moves. So the diff grew to include everything that landed since:
+      # #1027 changed 4 files and the filter reported 117, turning Backend Tests
+      # on for other people's backend/** work. Coverage got WIDER the staler a
+      # branch was, which is backwards.
+      #
+      # checkout leaves us on refs/pull/N/merge, whose FIRST parent is the base
+      # branch tip as of when that ref was built. Diffing against it yields the
+      # PR's own changes and nothing else. rev-parse reads the parent out of the
+      # merge commit itself, so this needs no extra fetch depth, and
+      # paths-filter's ensureRefAvailable does `fetch --depth=1 origin <sha>`
+      # for the base it is handed.
+      - name: Resolve the PR diff base
+        id: diffbase
+        if: github.event_name == 'pull_request'
+        run: |
+          set -euo pipefail
+          base="$(git rev-parse HEAD^1)"
+          echo "Diffing against first parent of the merge ref: $base"
+          echo "sha=$base" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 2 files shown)


ishiland:
Verified from this PR's own `Detect Changes` job (run `30671249891`, job `91289254086`), which is the only place the change executes.

```
HEAD is now at b8680cb Merge e5b8e6ad9 into a90ed904ab
Diffing against first parent of the merge ref: a90ed904ab28a2af47ed243479d93a4c956b3a4b
  base: a90ed904ab28a2af47ed243479d93a4c956b3a4b
Ensuring a90ed904ab... is fetched from origin
Change detection a90ed904ab..b8680cbe89
Detected 2 changed files
  .github/workflows/ci.yml [modified]
  backend/tests/test_ci_alembic_filter_paths.py [modified]
```

`a90ed904ab` is current main, not this branch's frozen `base.sha`. Two detected files, which is exactly what this PR changes.

Two notes for whoever reads this later.

The action's `ensureRefAvailable` found the base locally rather than fetching it, which confirms `fetch-depth: 2` is sufficient and a deeper clone is not needed.

The first push of this branch claimed in a code comment that no extra fetch depth was required, on the reasoning that `rev-parse` reads parents out of the commit object. That is wrong for a shallow clone: git grafts the fetched commit as a boundary and reports it as parentless. `Detect Changes` failed with `ambiguous argument 'HEAD^1'` and took `CI OK` with it, because it is a required context. Depth 2 in `e5b8e6ad9` is the fix.

Issue #1089, opened 2026-07-31
refactor(analysis): split analysis_sql by operation family


ishiland:
`backend/app/platform/analysis_sql.py` reached 1165 lines when the four M4
phase-2 operations landed (spatial join, measure, select by location,
intersect). On main before that batch it was 662. It is now the largest module
under `platform/`, and roughly 40% of it is those four operations.

It is gated rather than split. The ratchet entry in
`backend/tests/test_layering.py` records the number and the reason, and this
issue is the follow-up that entry points at.

## Why it was not split in the PR that grew it

That PR already carried four operations, a schema change, and a full
regeneration of `openapi.json`, `api.generated.ts` and both SDKs. Moving 1165
lines of SQL rendering at the same time would have put a restructure inside a
feature change, where a reviewer cannot tell which diff hunk is which and a
subtle change to a rendered statement reads as motion. That is how splits go
wrong.

## Why the growth is not itself the bug

Worth stating before anyone reaches for the obvious fix, because the obvious
fix is the wrong one.

The module lives in `platform/` so that both consumers render identical SQL:
the catalog preview path (`datasets/domain/service_analysis.py`) and the
processing materialize worker (`processing/analysis/tasks.py`). Catalog cannot
import processing and processing cannot import catalog, so before this module
existed the two paths each had their own copy of the statement, and they drifted.
An approved preview and the dataset it saved could disagree about what the
operation meant.

So splitting **by caller** would reintroduce exactly the failure the module was
built to prevent. Any proposal that gives the preview path and the worker their
own rendering modules should be rejected on that ground alone.

## The split that does work

By operation family, with the shared machinery staying central:

- `overlay`: clip, intersect, select by location. These share the mask
  handling, the `ST_Dimension = 2` polygonal guard, and the subdivide path.
- `measure`: area and length, plus whatever geodesic work joins it later.
- `transform`: buffer, centroid, dissolve. These change geometry in place
  rather than combining two inputs.

Staying central, because more than one family depends on each and duplicating
any of them recreates a drift surface:

- `LATERAL_ALIAS`, `NOT_EMPTY_PREDICATE`, and the `OFFSET 0` fencing idiom.
  The fence in particular is three characters wide and load-bearing; it belongs
  in one documented place, not copied into three files.
- `MAX_SOURCE_FEATURES`, `MAX_MASK_LAYER_FEATURES`, `MAX_BUFFER_METERS` and the
  measured benchmark tables that justify them.
[...]

PR #1097, opened 2026-07-31
feat(analysis): spatial join, measure, select by location, and intersect

### LINKED ISSUE #953: feat(analysis): spatial join — point-in-polygon count and attribute transfer

### LINKED ISSUE #954: feat(analysis): measurement — area and length as a computed output
Split out of #681 (phase-2 umbrella), which tracked four unrelated operations behind one body. This is operation 2 of 4.

Measurement is frequently the deliverable itself rather than an intermediate step (acreage for a permit application, main length by zone), which is what separates it from the transform operations we already ship. It is also the cheapest of the four: `ST_Area(geography)` and `ST_Length(geography)` give correct answers on unprojected data without any of the projection juggling the buffer path needs.

## What the gap actually is

An earlier version of this body implied chat cannot produce acreage. It can. `backend/app/processing/ai/sql_generator.py:329-332` ships the acreage example verbatim (`SUM(ST_Area(p.geom_4326::geography) / 4046.8564224) AS total_acres`), `:276-279` ships the unit conversion factors, and `st_area` and `st_length` are both on the sandbox allowlist (`backend/app/platform/sandbox/validator.py:263-282`). A user asking for total acreage in chat gets a table today.

The real gap is a **deterministic, non-LLM value that lands on the map or becomes a dataset**: a measured column the user can style by, sort, and export, produced by a fixed template rather than a generated query.

## Settled: computed column on materialize

Ship the computed column. `operation: "measure"` joins the Literals at `backend/app/modules/catalog/datasets/domain/schemas.py:887` (preview) and `:943` (materialize). It is a 1:1 operation like centroid, the output is a dataset, and `AnalysisPreviewResponse` (`schemas.py:924-937`) carries it without change because the geometry is unchanged.

The one-shot-answer option is struck. `query_data` already serves it, and the aggregate path #953 builds covers the case where a number has to come back through these rails.

## Settled: the unit is metres

**`area_sqm` for polygons, `length_m` for lines.** Metres on the wire; any other unit is a display concern for the picker.

This is the existing convention, stated in `frontend/src/components/builder/AnalysisPanel.tsx:44-46`: "buffer distances are metres on the wire; the picker converts so a user thinking in feet or miles doesn't have to". `BUFFER_UNIT_METERS` at `:47` is the conversion table to follow. `ST_Area(geography)` returns square metres and `ST_Length(geography)` returns metres, so the SQL needs no conversion at all.

## The generated-column collision is real and stays

The output columns are the source columns plus a generated one, so measurement hits the same collision `dissolve` already hit. `_validate_dissolve_by_field` rejects a `by_field` named `source_count` at `backend/app/modules/catalog/datasets/api/router_analysis.py:187-194` precisely because a duplicate output column fails the CTAS with an opaque "column specified more than once".

`_list_carry_columns` (`backend/app/processing/analysis/tasks.py:441-461`, wired into the CTAS build at `:683`) excludes only `gid`, `geom`, and `geom_4326` (`:456`), so a source column literally named `area_sqm` is carried straight through and collides. Reject it at enqueue with a 422 naming the column, mirroring the `source_count` guard.

## Struck: the NULL/empty criterion on materialize

The original body asked for a decision about what a NULL or empty source geometry measures to on the materialized output. On materialize that question is moot: `_wrap_not_empty` (`tasks.py:464-481`) filters `geom IS NULL OR ST_IsEmpty(geom)` **inside** the CTAS for every non-clip-by-layer shape, and the post-CTAS `DELETE` at `:723-724` removes any survivor as a backstop. Such rows never reach the output table, so they have no measured value to decide.

The question survives on the **preview** only, where `run_analysis_preview` skips NULL and coordinate-less geometries at `backend/app/modules/catalog/datasets/domain/service_analysis.py:199-204`. Pin what preview reports for those rows so it agrees with what materialize saves.

## Pointers

- Size ceiling: add an entry to `MAX_SOURCE_FEATURES` at `backend/app/platform/analysis_sql.py:50` even though measurement is cheap. The router reads it with `.get()` (`router_analysis.py:237`) and skips the gate when the key is absent, so an operation missing from the dict has **no** ceiling rather than a default one. Measurement does not amplify output the way buffer does, so the ceiling can be generous.
- This is the one phase-2 operation that is single-layer, so the two-dataset Rule-1 template does not apply. It still needs the ordinary source check via `_load_vector_dataset` (`router_analysis.py:84-98`).

## Gotchas

- The geography cast is not free on large inputs. Measure before assuming the 300s `MATERIALIZE_TIMEOUT` (`tasks.py:58`) is generous here.
- Antimeridian-crossing geometries: `geography` handles these correctly where planar area does not, which is a reason to prefer it over `ST_Area(geom::geometry)` even when the data looks local. The antimeridian batch (#884-#894, #897-#901, #924) established the fixtures to test against.

## Scope: this ships end to end

Labelled backend and frontend. Shipping the backend alone would add an operation the UI cannot invoke. In addition to the backend work:

- Add the operation to the `<Select>` in `frontend/src/components/builder/AnalysisPanel.tsx:937-975` and to the operation-label switch at `:255-260`.
- Add `analysisTools.op*` keys to **all four** locales: `frontend/src/i18n/locales/{en,es,fr,de}/builder.json`, in the `analysisTools` block (the `opBuffer` sibling is at `:967` in each). A `defaultValue` alone fails the `npm run test:i18n` locale-parity gate.
- Update the hand-typed mirror `AnalysisOperation` at `frontend/src/types/api.ts:1887`, which is maintained by hand and does not come from the SDK.
- Run `make openapi` and `make sdks`, and commit `backend/openapi.json` plus both SDK trees.

## Serialization

#953, #954, #955, and #956 all touch the same sites: the `operation` Literals in `schemas.py`, `MAX_SOURCE_FEATURES` in `analysis_sql.py`, `_build_materialize_select` in `tasks.py`, the whole-file `backend/openapi.json` and SDK regens, and the `api.ts` mirror. The regenerated files conflict on every pair.

**Land one at a time, in the order #953, #954, #955, #956**, re-running `make openapi && make sdks` after each rebase.

## Acceptance criteria

- [x] The shape decision is recorded above (computed column on materialize; `area_sqm` / `length_m` in metres).
- [ ] Area on polygons and length on lines both produce values matching an independently computed expected number, within a stated tolerance.
- [ ] A source dataset already carrying a column of the generated name is rejected at enqueue with a 422 naming the column, mirroring the `source_count` guard.
- [ ] NULL/empty source geometries produce the decided value **in the preview**, pinned by a fixture. Materialize needs no such fixture: `_wrap_not_empty` and the post-CTAS DELETE remove those rows.
- [ ] An antimeridian-crossing polygon measures correctly.
- [ ] `MAX_SOURCE_FEATURES` has an entry and the enqueue-time 422 is tested.
- [ ] The operation is selectable in `AnalysisPanel`, with `analysisTools.op*` keys in all four locales and `api.ts`'s `AnalysisOperation` updated.
- [ ] `make openapi` and `make sdks` run and all snapshots are committed.
- [ ] Backend `ruff check` / `ruff format --check` clean; frontend `npm run lint`, `npm run typecheck`, `npm run test:i18n` clean.

## Shared phase-2 constraints

- PostGIS is the asset: these are one-liners server-side and scale where client-side Turf falls over.
- Rule 1 on both datasets of any two-layer operation, always.
- The preview/materialize/sandbox rails from M4 extend naturally; do not build a parallel path.
- Any count or aggregate must compute over the full set, never the 500-capped preview (`service_analysis.py:193`), and must exclude NULL geometries.

### LINKED ISSUE #955: feat(analysis): select by location, then export the selected records
Split out of #681 (phase-2 umbrella), which tracked four unrelated operations behind one body. This is operation 3 of 4.

This is the operation that completes the buffer chain. The canonical workflow across planning, utilities, journalism, and site selection is: buffer, then select what falls inside, then export the list. We ship step one, so the ring on the map cannot yet answer the question that motivated drawing it.

Scope: select features of layer B that intersect a geometry (a drawn mask or the features of layer A), then export the selected records. **The record list is the deliverable**, not the geometry. Abutter notification lists are the archetype.

## Settled: the export is a materialized dataset

Materialize to a dataset and reuse the existing export endpoints. No new export surface: the user materializes, then downloads through paths that already exist.

The permission rationale is the reason, and it is written out at `backend/app/modules/catalog/datasets/api/router_analysis.py:215-222`. The materialize endpoint requires **both** `upload` and `export` (`require_permission("upload", "export")` at `:223`) precisely because it hands the caller a durable, caller-owned copy of the source attributes, which is the outcome the download endpoints gate on `export`. A direct export of selected records is that same outcome without the dataset, so it would need `export` at minimum plus its own route and its own permission story. Materializing costs a dataset-quota slot and an extra step; it is nearly free once the selection predicate exists, and it inherits a permission model that is already reasoned through. Revisit a direct-export route only if the extra step turns out to be the complaint.

## Settled: the predicate is intersects, with no parameter

One predicate, `intersects`, hard-coded. No `predicate` request field, no picker.

Within is a strict subset of intersects and the distinction matters mostly to people who already know they want it. Adding it means a new operation-scoped request field, an `_ANALYSIS_PARAM_OWNERS` entry (`backend/app/modules/catalog/datasets/domain/schemas.py:863-868`), an OpenAPI change, and a UI control, for a variant nobody has asked for. Ship intersects; add within when someone does.

## Settled: chaining on a previous result is out of scope

The selection geometry can come from three places, and only two exist:

- **Another layer's features**: the `mask_dataset_id` path shipped in #682 and is wired end to end (`schemas.py:900-906`, `router_analysis.py:251-254`, worker at `backend/app/processing/analysis/tasks.py:642-652`).
- **A drawn geometry**: the `mask` GeoJSON path, also shipped (`render_mask_expr` at `backend/app/platform/analysis_sql.py:568`).
- **A previous operation's result**: does not exist, and does not land here. A preview is ephemeral and never persisted, so "select within the buffer I just drew" requires materializing the buffer first. That is a workflow question about what the builder does with an ephemeral result, tracked as #790 item 1. Deciding it inside this issue would settle it for the wrong surface.

## The EXISTS filter is bbox-only

This is the correctness trap specific to this operation.

`render_clip_layer_join` (`backend/app/platform/analysis_sql.py:477`) already implements "features of A that relate to B", and a select-by-location looks like a clip that keeps whole geometries: drop the `ST_Intersection` lateral, keep the `EXISTS` row filter. Do that naively and the result is wrong.

The `EXISTS` clause it returns (`analysis_sql.py:561-564`) is a **bounding-box** test:

```
WHERE EXISTS (SELECT 1 FROM <mask> WHERE geom_4326 && _src.geom_4326)
```

`&&` alone admits any source row whose envelope overlaps a mask envelope. Clip gets its exactness from the lateral, which carries the real `ST_Intersects(_m.geom, ST_MakeValid(_src.geom_4326))` at `:559`; rows that pass `&&` but miss the true predicate fall out because their intersection is NULL or EMPTY and the not-empty filter removes them. A select-by-location has no lateral, so nothing downstream removes them, and a concave or L-shaped mask selects features that are nowhere near it.

Keep the `&&` term as the index-drivable prefilter and add a real `ST_Intersects` term inside the same `EXISTS`. The docstring at `:537-542` explains why the probe targets the raw mask table rather than the piece CTE (2.4s vs 0.25s when the mask sits at the high end of the gid order); preserve that.

## Pointers

- Two-dataset access check: copy `_load_mask_dataset` at `router_analysis.py:110-134`. Rule 1 applies to **both** datasets; the selection geometry's dataset needs the same check the source gets. Note it also enforces polygonal-only and a `MAX_MASK_LAYER_FEATURES` ceiling (`backend/app/platform/analysis_sql.py:39`), both of which apply here unchanged.
- Benchmarks: `render_clip_layer_join`'s docstring carries them (union 33.2s vs 3.4s against a 972-polygon / 249,804-vertex mask). It returns `(cte, lateral, where)`.
- Size ceiling: add an entry to `MAX_SOURCE_FEATURES` at `analysis_sql.py:50`. The router reads it with `.get()` (`router_analysis.py:237`) and skips the gate when the key is absent, so an operation missing from the dict has **no** ceiling rather than a default one.
- Exact selected-record count: reuse the uncapped aggregate path #953 builds rather than a second one.

## Scope: this ships end to end

Labelled backend and frontend. Shipping the backend alone would add an operation the UI cannot invoke. In addition to the backend work:

- Add the operation to the `<Select>` in `frontend/src/components/builder/AnalysisPanel.tsx:937-975` and to the operation-label switch at `:255-260`.
- Add `analysisTools.op*` keys to **all four** locales: `frontend/src/i18n/locales/{en,es,fr,de}/builder.json`, in the `analysisTools` block (the `opBuffer` sibling is at `:967` in each). A `defaultValue` alone fails the `npm run test:i18n` locale-parity gate.
- Update the hand-typed mirror `AnalysisOperation` at `frontend/src/types/api.ts:1887`, which is maintained by hand and does not come from the SDK.
- Run `make openapi` and `make sdks`, and commit `backend/openapi.json` plus both SDK trees.

## Serialization

#953, #954, #955, and #956 all touch the same sites: the `operation` Literals in `schemas.py`, `MAX_SOURCE_FEATURES` in `analysis_sql.py`, `_build_materialize_select` in `tasks.py`, the whole-file `backend/openapi.json` and SDK regens, and the `api.ts` mirror. The regenerated files conflict on every pair.

**Land one at a time, in the order #953, #954, #955, #956**, re-running `make openapi && make sdks` after each rebase. This one follows #953, whose uncapped aggregate path it reuses for the selected-record count.

## Acceptance criteria

- [x] The export-shape decision is recorded above (materialize to a dataset; predicate is intersects with no parameter; chaining on an ephemeral result is #790's).
- [ ] The selection predicate is exact, not bbox-only: a source feature whose envelope overlaps the mask envelope but whose geometry does not intersect the mask is **not** selected. A concave or L-shaped mask fixture is the test.
- [ ] Whole source geometries survive the selection unmodified. A test asserting the selected feature's geometry equals its source geometry, which is what separates this from clip.
- [ ] Both datasets are access-checked; a user with read on one and not the other gets a 404.
- [ ] The count of selected records is exact, not capped at the 500-feature preview limit (`backend/app/modules/catalog/datasets/domain/service_analysis.py:193`).
- [ ] NULL-geometry rows in either dataset neither crash the operation nor appear in the selection.
- [ ] `MAX_SOURCE_FEATURES` has an entry and the enqueue-time 422 is tested.
- [ ] The operation is selectable in `AnalysisPanel`, with `analysisTools.op*` keys in all four locales and `api.ts`'s `AnalysisOperation` updated.
- [ ] `make openapi` and `make sdks` run and all snapshots are committed.
- [ ] Backend `ruff check` / `ruff format --check` clean; frontend `npm run lint`, `npm run typecheck`, `npm run test:i18n` clean.

## Shared phase-2 constraints

- PostGIS is the asset: these are one-liners server-side and scale where client-side Turf falls over.
- Rule 1 on both datasets of any two-layer operation, always.
- The preview/materialize/sandbox rails from M4 extend naturally; do not build a parallel path.
- Any count or aggregate must compute over the full set, never the 500-capped preview, and must exclude NULL geometries.

### LINKED ISSUE #956: feat(analysis): intersect/overlay — new geometry from two layers

ishiland:
## The performance fix comes first, because it is the easiest thing here to skim past

Spatial join's predicate used to build its geometry expressions inside the join
condition, where the planner pulled them up and evaluated them once per
candidate PAIR rather than once per row. Fencing the expression in
`CROSS JOIN LATERAL (SELECT expr AS x OFFSET 0)` blocks subquery pull-up and
restores per-row evaluation.

Measured on the same dataset and query, before and after:

```
28.36s  ->  0.20s
```

`OFFSET 0` is a no-op to a reader and a planner barrier to PostgreSQL. It is
the single load-bearing idiom in this SQL, it appears again in intersect, and
it is three characters wide. That is why it is at the top of this description
and in the first commit rather than somewhere in the middle of the diff.

## Why these four are one PR

Every one of the four adds a value to the same `Literal`, so every one of them
rewrites `openapi.json`, `api.generated.ts`, `types.gen.ts` and both
`analysis_*_request_operation.py` enums. There is no subset that avoids the
collision. Landing them separately means four rebases, four regenerations and
four CI cycles, all of the cost sitting in generated files.

The stronger reason is that they are not independent:

| symbol | introduced | consumed |
|---|---|---|
| `MASK_OPERATIONS` | select_by_location | intersect |
| `_ROW_FILTERING_OPERATIONS` | select_by_location | intersect |
| `_resolve_layer_table_ref` | spatial_join | intersect |
| `_resolve_match_count` | spatial_join | select_by_location, intersect |

Reviewed one at a time, `_resolve_match_count` gets read in three shapes, two
of which never ship. The `MAX_SOURCE_FEATURES` comment is a single argument
written across four changes. The `service_analysis.py` line-budget raise only
justifies itself once all four exist. Splitting them for review is the
artificial move.

## Commit shape

Four code-only commits, then one `chore(sdks)` regeneration.

Generated files appear in exactly one commit. A rebase therefore has one
conflict surface instead of four, and it is a surface that must be resolved by
running `make openapi && make sdks`, never by editing the conflict text. Please
do not ask for the regeneration to be folded back into each commit.

Review commit by commit and the 1,600 lines of application code arrive as four
structurally identical pieces rather than one wall.

### What the regeneration rule bought, concretely

This is not an appeal to a convention. It already caught something once during
this work, on #1045, the PR this branch was based on.

Rebasing that branch onto main conflicted in exactly one file,
`sdks/typescript/src/client/index.ts`, which is a single line holding 1,862
generated exports. There is no side of that a person can merge by reading it.

Taking the branch side verbatim and moving on would have looked completely
clean. It also would have carried five exports for `tiles/raster-auth-check`, a
route main removed while that branch was open. Running `make sdks` deleted
them. `make sdks-check` does fail on exactly that, so the mistake would have
surfaced, but a full CI cycle later and reading as a bug in the branch.

The other half is worth stating because it distinguishes two failure modes that
look identical in a conflict marker: `make openapi` regenerated
`backend/openapi.json` byte for byte identical. The schema merged correctly on
its own. Only the barrel file needed rebuilding. A generated file conflicting
mechanically is not the same thing as a schema that actually diverged, and the
regeneration is what tells them apart.

## The operations

**Spatial join** attaches fields from a second layer to each source row. Its
cost shape is one GIST probe per source row, which is where the `OFFSET 0`
hoist above matters.

**Measure** adds area and length as computed columns. It emits both columns
always, rather than branching on the layer's geometry type, because
`geometry_type` is classified from the first feature only. That is the trap
`fix(#682)` documents, and branching on it would quietly measure the wrong
thing for every row after the first.

**Select by location** keeps whole features that touch a mask layer. It looks
like a clip that skips the clipping, and writing it that way is wrong: clip's
row filter is `&&` alone, a bounding-box test, which survives only because rows
that miss go on to intersect to NULL or EMPTY and get dropped downstream. A
selection has no such downstream, so a bounding-box filter would select
features that fall in the notch of an L-shaped mask and never touch it. The
fixture is an L, and it asserts that the bounding-box predicate on its own
still takes the notch feature, so the test cannot rot into one that passes for
the wrong reason.

**Intersect** produces one row per intersecting source and overlay feature
pair. Grouping is by the overlay's gid rather than by piece, `ST_MakeValid` is
hoisted into an `OFFSET 0` lateral so it runs once per row, and `row_number()`
is evaluated after the WHERE so output gids are contiguous.

## Sizing the intersect ceiling

The issue asks for a ceiling sized on the product of the two inputs.
`MAX_SOURCE_FEATURES` cannot express a product, so the number was measured
rather than guessed. Benchmarked against a 972-polygon, 249,804-vertex overlay:

| sources | overlap | output rows | CTAS | size |
|--:|--:|--:|--:|--:|
| 1,000 | 4x | 4,000 | 0.26s | 3.6 MB |
| 10,000 | 4x | 40,000 | 1.4s | 35 MB |
| 50,000 | 4x | 200,000 | 12.4s | 174 MB |
| 150,000 | 4x | 600,000 | 22.2s | 521 MB |
| 10,000 | 58x | 577,453 | 47.8s | 1,235 MB |
| 10,000 | 145x | died | 7.2s | `temp_file_limit` (4 GB) |

A 15x smaller source count is the more expensive job. No source ceiling
separates those two runs, so the only value that would reject the 47.8s job
also rejects the benign 150k one that finishes in 22s. The input cap stays at
100k as a cheap early exit and `_enforce_output_size` remains the real bound.
The full table sits beside the constant so the next reader does not re-derive
it.

## Module size

Four line-count ratchets move, and all of the growth is this branch rather than
main:

```
analysis_sql.py        662 -> 1165   newly crosses the 1000-line inclusion gate
processing/tasks.py    891 -> 1071   newly crosses it too
schemas.py            1032 -> 1127   existing ratchet moved
service_analysis.py    310 ->  455   decomposed-service budget raised
```

The two that newly cross 1000 do so for one reason. Every rendered statement
lives in `analysis_sql` so that the preview path and the materialize worker
cannot disagree about what SQL they run, and `tasks.py` decides every CTAS
shape in one place. Splitting either per-operation would trade module size for
exactly the drift they exist to prevent.

That is an argument for the growth, not a claim it is free. `analysis_sql` is
now the largest module under `platform/`. The split is filed as #1089, by
operation family rather than by caller, and the ratchet entry points at it.

## Verification

Backend suite 6,721 passed, frontend 4,506 passed across 369 files. Typecheck,
lint, i18n locale parity, `make openapi-check` and `make sdks-check` all clean.
New `t()` keys added to all four locales.

Mutation-checked rather than assumed: removing `ST_Intersects` from select by
location fails 6 tests, removing the `ST_Dimension = 2` guard fails 1. Two
panel tests that claimed to verify drawn-mask teardown were asserting the
request body, which omits `mask` for these operations anyway, so they passed
with the teardown disabled. Both were rewritten against `TerraDraw.stop` and
both now fail under mutation.

## Why verification is SQL-level rather than browser-level

The wave plan asks for each operation to be exercised on a real map in a
browser. That did not happen, and the honest account is that the blocker moved
twice while this work was in flight rather than that it was skipped.

This branch adds no migration of its own. It is pure SQL rendering plus the
panel that drives it, so it needs nothing from the database that main does not
already have. For most of the work the shared development database sat behind
the head this branch's base required, and taking it forward would have put it
ahead of the image the other parallel sessions were running, which the MIG-02
guard turns into a refusal to boot. One database serving several branches at
different heads can satisfy one of them at a time.

The substitution was therefore chosen rather than settled for, and it is aimed
at the failure mode a browser would actually have caught here. The one real
defect found by live QA in this batch was a performance cliff at scale, on
spatial join, so the shipped renderers were imported from `analysis_sql.py`,
meaning the SQL under test is the SQL that ships rather than a restatement of
it, and run against real development data at real sizes. Every answer was then
cross-checked against an independently shaped query carrying none of the same
machinery: no subdivision, no `OFFSET 0` hoist, no dimension filter, no
`row_number()`. Two implementations that share no structure agreeing on the
answer is a stronger correctness signal than one rendered map.

| check | naive query | shipped renderer |
|---|---|---|
| meteorites selected, 32,186 x 294 | 2,838 | 2,838 |
| lakes selected, 412 x 294 | 295 | 295 |
| overlay pairs | 359 | 359 |

Measure over 22,324 Manhattan buildings took 442ms. Select by location's
preview took 180ms and its exact uncapped count 1,083ms. Intersect's pairwise
render took 108ms, produced valid geometry throughout and zero empty results.

The last row is the useful one. Select by location reports 295 lakes touching
a state, and intersect turns those same 295 sources into 359 pieces. The two
operations agree on which features participate and differ by exactly the
pairwise split, which is the documented distinction between them, confirmed
against real data rather than asserted.

Closes #953
Closes #954
Closes #955
Closes #956

### DIFF
--- a/backend/app/modules/catalog/authorization.py
+++ b/backend/app/modules/catalog/authorization.py
@@ -168,6 +168,21 @@ async def _can_access_dataset_id(
     )
 
 
+# fix(#1097 review): every provenance param naming a dataset, mapped to the
+# params that DESCRIBE that dataset and must be dropped with it.
+#
+# The keys are what an unauthorized requester must not learn the value of; the
+# values are what would still describe the layer once its id was gone. An
+# operation that adds a dataset-id param adds a row here, and
+# test_every_dataset_id_param_is_redactable fails until it does — the check is
+# structural because the failure mode is silence: the leak looks like ordinary
+# provenance, and nothing errors.
+_DATASET_ID_PARAMS: dict[str, tuple[str, ...]] = {
+    "mask_dataset_id": (),
+    "join_dataset_id": ("join_fields",),
+}
+
+
 async def visible_derived_from(
     db: AsyncSession,
     derived_from: dict | None,
@@ -189,6 +204,16 @@ async def visible_derived_from(
     source's verdict. A returned dict is always a copy; the caller must not be
     able to mutate the record's JSONB through it.
 
+    fix(#1097 review): driven by ``_DATASET_ID_PARAMS`` rather than by a check
+    per id. Spatial join added ``join_dataset_id`` and this function kept
+    redacting only the mask, so a public join output derived through a PRIVATE
+    join layer published that layer's id — and ``join_fields`` with it, which
+    is the layer's column names. That is the same gap #765's review closed for
+    the mask, reopened by the next operation to carry a dataset id, which is
+    the argument for a table over another branch: an operation that adds one is
+    now a row here, and the structural test enforcing that lives in
+    ``test_analysis_provenance.py``.
+
     A source that has since been deleted also yields None: access to it can no
     longer be established, and the prose lineage on the record still reads.
     """
@@ -200,11 +225,18 @@ async def visible_derived_from(
         return None
 
     params = dict(derived_from.get("params") or {})
-    mask_id = params.get("mask_dataset_id")
-    if mask_id is not None and not await _can_access_dataset_id(
-        db, mask_id, user, user_roles
-    ):
-        params.pop("mask_dataset_id")
+    for id_param, dependent_params in _DATASET_ID_PARAMS.items():
+        dataset_id = params.get(id_param)
+        if dataset_id is None:
+            continue
+        if await _can_access_dataset_id(db, dataset_id, user, user_roles):
+            continue
+        params.pop(id_param)
+        # The id is not the only thing that describes the layer. Dropping
+        # join_dataset_id while keeping join_fields would still publish the
+        # private layer's column names, which is most of what its schema is.
+        for dependent in dependent_params:
+            params.pop(dependent, None)
     return {**derived_from, "params": params}
 
 
--- a/backend/app/modules/catalog/datasets/api/router_analysis.py
+++ b/backend/app/modules/catalog/datasets/api/router_analysis.py
@@ -373,8 +373,13 @@ async def analysis_preview_endpoint(
     join_dataset = None
     if body.join_dataset_id is not None:
         join_dataset = await _load_join_dataset(db, body.join_dataset_id, user)
-        if body.join_fields:
-            _validate_join_fields(dataset, join_dataset, body.join_fields)
+        # fix(#1097 review): unconditionally, matching materialize. The guard
+        # used to be `if body.join_fields`, but _validate_join_fields also
+        # checks the ALWAYS-generated join_count against the source's columns —
+        # so a source that already had a join_count column previewed fine and
+        # then failed Create on the identical form. Preview approving an output
+        # that Create refuses is worse than either verdict on its own.
+        _validate_join_fields(dataset, join_dataset, body.join_fields or [])
     try:
         return await run_analysis_preview(
             db,
--- a/backend/app/modules/catalog/datasets/api/router_analysis.py
+++ b/backend/app/modules/catalog/datasets/api/router_analysis.py
@@ -29,6 +29,7 @@ from app.modules.catalog.datasets.domain.service import (
 )
 from app.modules.quota.service import check_upload_quota
 from app.platform.analysis_sql import (
+    INTERSECT_OUTPUT_COLUMNS,
     MAX_MASK_LAYER_FEATURES,
     MAX_SOURCE_FEATURES,
     MEASURE_OUTPUT_COLUMNS,
@@ -224,6 +225,43 @@ def _validate_join_fields(source, join_dataset, join_fields: list[str]) -> None:
     _reject_generated_column_collision(source, spatial_join_output_columns(join_fields))
 
 
+def _column_names(dataset) -> set[str]:
+    return {col.get("name") for col in (dataset.column_info or []) if col}
+
+
+def _validate_intersect_columns(source, overlay) -> None:
+    """422 on any column an overlay would emit twice (fix(#956)).
+
+    An overlay is the first operation to carry columns from BOTH inputs onto
+    every output row, so a same-named column in the two layers is likely
+    rather than exotic — ``name``, ``id`` and ``area`` are all common. The CTAS
+    would fail it with an opaque "column specified more than once" after the
+    whole queue wait. Prefixing silently is the alternative, and it makes the
+    output columns unpredictable for anyone scripting against the result.
+    """
+    _reject_generated_column_collision(source, INTERSECT_OUTPUT_COLUMNS)
+    generated = sorted(_column_names(overlay) & set(INTERSECT_OUTPUT_COLUMNS))
+    if generated:
+        raise HTTPException(
+            status_code=status.HTTP_422_UNPROCESSABLE_CONTENT,
+            detail=(
+                f"The overlay layer has a column named {generated[0]!r}, which "
+                "this operation generates. Rename it, or choose a different "
+                "layer."
+            ),
+        )
+    clashes = sorted(_column_names(source) & _column_names(overlay))
+    if clashes:
+        raise HTTPException(
+            status_code=status.HTTP_422_UNPROCESSABLE_CONTENT,
+            detail=(
+                f"Both layers have a column named {clashes[0]!r}, and an "
+                "overlay carries columns from both. Rename one of them, or "
+                "choose a different layer."
+            ),
+        )
+
+
 @router.post("/{dataset_id}/analysis/preview/", response_model=AnalysisPreviewResponse)
 async def analysis_preview_endpoint(
     dataset_id: uuid.UUID,
@@ -332,6 +370,12 @@ async def _validate_materialize_params(
         _validate_join_fields(dataset, join_dataset, body.join_fields or [])
     if body.operation == "measure":
         _reject_generated_column_collision(dataset, MEASURE_OUTPUT_COLUMNS)
+    if body.operation == "intersect":
+        # Access check on the overlay layer, plus the column checks. Rule 1
+        # applies to BOTH datasets; the worker re-resolves the table and
+        # re-checks the collisions against the live columns after the queue.
+        overlay = await _load_mask_dataset(db, body.mask_dataset_id, user)
+        _validate_intersect_columns(dataset, overlay)
 
 
 @router.post(
--- a/backend/app/modules/catalog/datasets/api/router_analysis.py
+++ b/backend/app/modules/catalog/datasets/api/router_analysis.py
@@ -4,6 +4,7 @@ import re
 import uuid
 from collections.abc import Iterable
 from datetime import datetime, timedelta, timezone
+from typing import Any
 
 from fastapi import APIRouter, Depends, HTTPException, Request, status
 from sqlalchemy import and_, func, or_, select, text
@@ -378,6 +379,42 @@ async def _validate_materialize_params(
         _validate_intersect_columns(dataset, overlay)
 
 
+def _build_analysis_job_metadata(
+    body: AnalysisMaterializeRequest, dataset
+) -> dict[str, Any]:
+    """The params recorded on the job so Admin -> Jobs can diagnose a run.
+
+    "analysis-buffer failed" on its own says nothing. The drawn mask geometry
+    is deliberately NOT stored: it can be kilobytes, and a marker suffices.
+
+    Extracted from the handler rather than left inline (#1097 review): this is
+    the per-operation dispatch, so it is the block that grows every time an
+    operation is added, and it took the endpoint past ruff's C901 threshold at
+    the fourth one. Inline, the next operation would trip the same rule again
+    and the fix would be the same extraction done under more pressure. Nothing
+    here touches the request, the session, or the job row, so it lifts out
+    whole.
+    """
+    meta: dict[str, Any] = {
+        "operation": body.operation,
+        "source_dataset_id": str(dataset.id),
+        "title": body.title,
+    }
+    if body.distance_meters is not None:
+        meta["distance_meters"] = body.distance_meters
+    if body.by_field is not None:
+        meta["by_field"] = body.by_field
+    if body.operation == "clip":
+        meta["mask_source"] = "layer" if body.mask_dataset_id else "drawn"
+        if body.mask_dataset_id is not None:
+            meta["mask_dataset_id"] = str(body.mask_dataset_id)
+    if body.operation == "spatial_join":
+        meta["join_dataset_id"] = str(body.join_dataset_id)
+        if body.join_fields:
+            meta["join_fields"] = list(body.join_fields)
+    return meta
+
+
 @router.post(
     "/{dataset_id}/analysis/materialize/",
     response_model=AnalysisMaterializeResponse,
@@ -533,27 +570,7 @@ async def analysis_materialize_endpoint(
     job = await get_catalog_port().create_ingest_job(
         db, f"analysis-{body.operation}", "", user.id
     )
-    # Record the request params so Admin → Jobs can diagnose a failed run
-    # ("analysis-buffer failed" alone says nothing). The drawn mask geometry
-    # is deliberately NOT stored — it can be kilobytes; a marker suffices.
-    analysis_meta = {
-        "operation": body.operation,
-        "source_dataset_id": str(dataset.id),
-        "title": body.title,
-    }
-    if body.distance_meters is not None:
-        analysis_meta["distance_meters"] = body.distance_meters
-    if body.by_field is not None:
-        analysis_meta["by_field"] = body.by_field
-    if body.operation == "clip":
-        analysis_meta["mask_source"] = "layer" if body.mask_dataset_id else "drawn"
-        if body.mask_dataset_id is not None:
-            analysis_meta["mask_dataset_id"] = str(body.mask_dataset_id)
-    if body.operation == "spatial_join":
-        analysis_meta["join_dataset_id"] = str(body.join_dataset_id)
-        if body.join_fields:
-            analysis_meta["join_fields"] = list(body.join_fields)
-    job.user_metadata = {"analysis": analysis_meta}
+    job.user_metadata = {"analysis": _build_analysis_job_metadata(body, dataset)}
     # ux(#698): stamp a step so a pending job reads as "queued" rather than
     # being indistinguishable from a broken one. This matters more since #703
     # deliberately defers analysis below the default priority — a queued job
--- a/backend/app/modules/catalog/datasets/api/router_analysis.py
+++ b/backend/app/modules/catalog/datasets/api/router_analysis.py
@@ -498,10 +498,20 @@ def _build_analysis_job_metadata(
         meta["distance_meters"] = body.distance_meters
     if body.by_field is not None:
         meta["by_field"] = body.by_field
-    if body.operation == "clip":
+    # fix(#1097 review): the second layer is recorded for EVERY operation that
+    # consumes one, not just clip. Admin Jobs surfaces this metadata to
+    # diagnose a failed run, and select_by_location and intersect are both
+    # driven by a layer the operator could not otherwise identify — so a failure
+    # caused by that layer (re-uploaded mid-queue, wrong geometry, ungroupable
+    # column) showed only the operation, the source and the title.
+    #
+    # mask_source stays scoped to the operations that can take a DRAWN mask.
+    # intersect rejects one, so "layer" there would be a constant dressed up as
+    # a discriminator.
+    if body.mask_dataset_id is not None:
+        meta["mask_dataset_id"] = str(body.mask_dataset_id)
+    if body.operation in MASK_OPERATIONS:
         meta["mask_source"] = "layer" if body.mask_dataset_id else "drawn"
-        if body.mask_dataset_id is not None:
-            meta["mask_dataset_id"] = str(body.mask_dataset_id)
     if body.operation == "spatial_join":
         meta["join_dataset_id"] = str(body.join_dataset_id)
         if body.join_fields:
--- a/backend/app/modules/catalog/datasets/api/router_analysis.py
+++ b/backend/app/modules/catalog/datasets/api/router_analysis.py
@@ -2,6 +2,7 @@
 
 import re
 import uuid
+from collections.abc import Iterable
 from datetime import datetime, timedelta, timezone
 
 from fastapi import APIRouter, Depends, HTTPException, Request, status
@@ -29,6 +30,7 @@ from app.modules.quota.service import check_upload_quota
 from app.platform.analysis_sql import (
     MAX_MASK_LAYER_FEATURES,
     MAX_SOURCE_FEATURES,
+    MEASURE_OUTPUT_COLUMNS,
     render_mask_expr,
     spatial_join_output_columns,
 )
@@ -178,6 +180,30 @@ async def _load_join_dataset(
     return await _load_vector_dataset(db, join_dataset_id, user)
 
 
+def _reject_generated_column_collision(source, generated: Iterable[str]) -> None:
+    """422 when the source already has a column an operation would generate.
+
+    Every 1:1 operation's output is the source's own columns verbatim plus its
+    generated ones, so a source column of the same name reaches the CTAS twice
+    and fails it with an opaque "column specified more than once" — after the
+    whole queue wait. Named at enqueue instead. This is the shared form of the
+    guard dissolve applies to ``source_count`` in ``_validate_dissolve_by_field``
+    (fix(#954): measure and spatial_join both need it, so it lives in one place
+    rather than being re-derived per operation).
+    """
+    source_columns = {col.get("name") for col in (source.column_info or []) if col}
+    clashes = sorted(source_columns & set(generated))
+    if clashes:
+        raise HTTPException(
+            status_code=status.HTTP_422_UNPROCESSABLE_CONTENT,
+            detail=(
+                f"The source dataset already has a column named {clashes[0]!r}, "
+                "which this operation would overwrite. Rename it, or choose a "
+                "different operation."
+            ),
+        )
+
+
 def _validate_join_fields(source, join_dataset, join_fields: list[str]) -> None:
     """422 on unknown join columns, or ones that would collide on output."""
     known = {col.get("name") for col in (join_dataset.column_info or []) if col}
@@ -187,21 +213,7 @@ def _validate_join_fields(source, join_dataset, join_fields: list[str]) -> None:
                 status_code=status.HTTP_422_UNPROCESSABLE_CONTENT,
                 detail=f"Unknown join column: {name!r}",
             )
-    # The output carries the source's own columns verbatim beside the generated
-    # ones, so a source column named join_count (or join_<field>) would reach
-    # the CTAS twice and fail it with "column specified more than once" after
-    # the queue wait. Same trap dissolve's by_field hits on source_count.
-    source_columns = {col.get("name") for col in (source.column_info or []) if col}
-    clashes = sorted(source_columns & set(spatial_join_output_columns(join_fields)))
-    if clashes:
-        raise HTTPException(
-            status_code=status.HTTP_422_UNPROCESSABLE_CONTENT,
-            detail=(
-                f"The source dataset already has a column named {clashes[0]!r}, "
-                "which this join would overwrite. Rename it, or drop that "
-                "field from the join."
-            ),
-        )
+    _reject_generated_column_collision(source, spatial_join_output_columns(join_fields))
 
 
 @router.post("/{dataset_id}/analysis/preview/", response_model=AnalysisPreviewResponse)
@@ -308,6 +320,8 @@ async def _validate_materialize_params(
         # the table name and re-checks the collision against the live columns.
         join_dataset = await _load_join_dataset(db, body.join_dataset_id, user)
         _validate_join_fields(dataset, join_dataset, body.join_fields or [])
+    if body.operation == "measure":
+        _reject_generated_column_collision(dataset, MEASURE_OUTPUT_COLUMNS)
 
 
 @router.post(
--- a/backend/app/modules/catalog/datasets/api/router_analysis.py
+++ b/backend/app/modules/catalog/datasets/api/router_analysis.py
@@ -30,6 +30,7 @@ from app.platform.analysis_sql import (
     MAX_MASK_LAYER_FEATURES,
     MAX_SOURCE_FEATURES,
     render_mask_expr,
+    spatial_join_output_columns,
 )
 from app.platform.extensions import get_catalog_port
 from app.platform.jobs.defer_guard import (
@@ -160,6 +161,49 @@ async def _load_mask_dataset(
     return dataset
 
 
+async def _load_join_dataset(
+    db: AsyncSession, join_dataset_id: uuid.UUID, user: Identity
+):
+    """Fetch + visibility-check a spatial-join layer (fix(#953)).
+
+    Rule 1 applies to BOTH datasets of a two-layer operation, so this is the
+    same treatment ``_load_mask_dataset`` gives the clip mask. No geometry-type
+    requirement, unlike the mask: a join is meaningful in every direction —
+    points in polygons, polygons touching lines, polygons overlapping polygons
+    — and the count means the same thing in all of them. No size ceiling of its
+    own either: the join layer is probed through its GIST index once per source
+    row, so it is the SOURCE row count that drives the cost, and that is what
+    MAX_SOURCE_FEATURES['spatial_join'] bounds.
+    """
+    return await _load_vector_dataset(db, join_dataset_id, user)
+
+
+def _validate_join_fields(source, join_dataset, join_fields: list[str]) -> None:
+    """422 on unknown join columns, or ones that would collide on output."""
+    known = {col.get("name") for col in (join_dataset.column_info or []) if col}
+    for name in join_fields:
+        if not _SAFE_COLUMN_RE.match(name) or name not in known:
+            raise HTTPException(
+                status_code=status.HTTP_422_UNPROCESSABLE_CONTENT,
+                detail=f"Unknown join column: {name!r}",
+            )
+    # The output carries the source's own columns verbatim beside the generated
+    # ones, so a source column named join_count (or join_<field>) would reach
+    # the CTAS twice and fail it with "column specified more than once" after
+    # the queue wait. Same trap dissolve's by_field hits on source_count.
+    source_columns = {col.get("name") for col in (source.column_info or []) if col}
+    clashes = sorted(source_columns & set(spatial_join_output_columns(join_fields)))
+    if clashes:
+        raise HTTPException(
+            status_code=status.HTTP_422_UNPROCESSABLE_CONTENT,
+            detail=(
+                f"The source dataset already has a column named {clashes[0]!r}, "
+                "which this join would overwrite. Rename it, or drop that "
+                "field from the join."
+            ),
+        )
+
+
 @router.post("/{dataset_id}/analysis/preview/", response_model=AnalysisPreviewResponse)
 async def analysis_preview_endpoint(
     dataset_id: uuid.UUID,
@@ -178,6 +222,11 @@ async def analysis_preview_endpoint(
         if body.mask_dataset_id is not None
         else None
     )
+    join_dataset = None
+    if body.join_dataset_id is not None:
+        join_dataset = await _load_join_dataset(db, body.join_dataset_id, user)
+        if body.join_fields:
+            _validate_join_fields(dataset, join_dataset, body.join_fields)
     try:
         return await run_analysis_preview(
             db,
@@ -185,6 +234,7 @@ async def analysis_preview_endpoint(
             body,
             user.id,
             mask_dataset=mask_dataset,
+            join_dataset=join_dataset,
             # fix(#716): safe here — `user.id` is evaluated above, and neither
             # this handler nor any middleware reads ORM state afterwards.
             release_session=True,
@@ -230,6 +280,36 @@ def _validate_dissolve_by_field(dataset, by_field: str) -> None:
         )
 
 
+async def _validate_materialize_params(
+    db: AsyncSession, dataset, body: AnalysisMaterializeRequest, user: Identity
+) -> None:
+    """Per-operation enqueue-time validation, so a bad request 422s fast.
+
+    Everything here fails BEFORE a job row exists — the alternative is a job
+    the user has to watch fail minutes later with an opaque database error.
+    Each check has a second, run-time half in the worker, because the queue
+    wait sits between the two and the world can move underneath it.
+    """
+    if body.operation == "clip" and body.mask_dataset_id is not None:
+        # Access + polygon checks happen here at enqueue time; the worker
+        # re-resolves the table name and re-validates it against _SAFE_TABLE.
+        await _load_mask_dataset(db, body.mask_dataset_id, user)
+    elif body.operation == "clip":
+        try:
+            render_mask_expr(body.mask or {})
+        except ValueError as exc:
+            raise HTTPException(
+                status_code=status.HTTP_422_UNPROCESSABLE_CONTENT, detail=str(exc)
+            ) from exc
+    if body.operation == "dissolve" and body.by_field is not None:
+        _validate_dissolve_by_field(dataset, body.by_field)
+    if body.operation == "spatial_join":
+        # Access + column checks, same as the clip mask; the worker re-resolves
+        # the table name and re-checks the collision against the live columns.
+        join_dataset = await _load_join_dataset(db, body.join_dataset_id, user)
+        _validate_join_fields(dataset, join_dataset, body.join_fields or [])
+
+
 @router.post(
     "/{dataset_id}/analysis/materialize/",
     response_model=AnalysisMaterializeResponse,
@@ -273,20 +353,7 @@ async def analysis_materialize_endpoint(
                 ),
             )
 
-    # Fail fast on invalid params before creating a job.
-    if body.operation == "clip" and body.mask_dataset_id is not None:
-        # Access + polygon checks happen here at enqueue time; the worker
-        # re-resolves the table name and re-validates it against _SAFE_TABLE.
-        await _load_mask_dataset(db, body.mask_dataset_id, user)
-    elif body.operation == "clip":
-        try:
-            render_mask_expr(body.mask or {})
-        except ValueError as exc:
-            raise HTTPException(
-                status_code=status.HTTP_422_UNPROCESSABLE_CONTENT, detail=str(exc)
-            ) from exc
-    if body.operation == "dissolve" and body.by_field is not None:
-        _validate_dissolve_by_field(dataset, body.by_field)
+    await _validate_materialize_params(db, dataset, body, user)
 
     # Best-effort dataset-count pre-check; the authoritative atomic
     # reservation happens at registration time in the worker.
@@ -414,6 +481,10 @@ async def analysis_materialize_endpoint(
         analysis_meta["mask_source"] = "layer" if body.mask_dataset_id else "drawn"
         if body.mask_dataset_id is not None:
             analysis_meta["mask_dataset_id"] = str(body.mask_dataset_id)
+    if body.operation == "spatial_join":
+        analysis_meta["join_dataset_id"] = str(body.join_dataset_id)
+        if body.join_fields:
+            analysis_meta["join_fields"] = list(body.join_fields)
     job.user_metadata = {"analysis": analysis_meta}
     # ux(#698): stamp a step so a pending job reads as "queued" rather than
     # being indistinguishable from a broken one. This matters more since #703
@@ -432,11 +503,16 @@ async def analysis_materialize_endpoint(
         # the pre-clip-by-layer code rejects unknown kwargs, and an
         # unconditional None would break EVERY materialize during a rolling
         # deploy instead of only the new feature.
-        extra_kwargs = (
-            {"mask_dataset_id": str(body.mask_dataset_id)}
-            if body.mask_dataset_id is not None
-            else {}
-        )
+        extra_kwargs: dict[str, object] = {}
+        if body.mask_dataset_id is not None:
+            extra_kwargs["mask_dataset_id"] = str(body.mask_dataset_id)
+        # Same rolling-deploy rule for the join params (fix(#953)): a worker
+        # still running pre-spatial-join code rejects unknown kwargs, so they
+        # ride along only when the operation actually uses them.
+        if body.join_dataset_id is not None:
+            extra_kwargs["join_dataset_id"] = str(body.join_dataset_id)
+            if body.join_fields:
+                extra_kwargs["join_fields"] = list(body.join_fields)
         await defer_async_with_tenant(
             get_catalog_port()
             .materialize_analysis_task()
--- a/backend/app/modules/catalog/datasets/api/router_analysis.py
+++ b/backend/app/modules/catalog/datasets/api/router_analysis.py
@@ -34,6 +34,7 @@ from app.platform.analysis_sql import (
     MAX_MASK_LAYER_FEATURES,
     MAX_SOURCE_FEATURES,
     MEASURE_OUTPUT_COLUMNS,
+    NON_GROUPABLE_COLUMN_TYPES,
     render_mask_expr,
     spatial_join_output_columns,
 )
@@ -85,7 +86,6 @@ MAX_ACTIVE_MATERIALIZES_PER_TENANT = 3
 # dissolve GROUP BY on such a column fails the CTAS with an opaque 42883
 # after the queue wait. GDAL maps nested GeoJSON objects to `json`, so
 # real uploads hit this. Rejected at enqueue with the column named.
-_NON_GROUPABLE_TYPES = {"json", "xml"}
 
 # fix(#695): Procrastinate ranks by per-job priority (DESC, default 0), not
 # by queue name — so a 300-second analysis CTAS enqueued first would
@@ -302,7 +302,7 @@ def _validate_intersect_columns(source, overlay) -> None:
     # dies after the queue wait with an error naming a generated alias the user
     # never wrote.
     #
-    # Refused at enqueue instead, which is what _NON_GROUPABLE_TYPES already
+    # Refused at enqueue instead, which is what NON_GROUPABLE_COLUMN_TYPES already
     # does for dissolve's by_field. Refusing rather than silently dropping the
     # column is the same choice the sibling guards above make: an overlay whose
     # output columns depend on which of them happened to be groupable is worse
@@ -315,7 +315,7 @@ def _validate_intersect_columns(source, overlay) -> None:
     ungroupable = sorted(
         (col.get("name"), str(col.get("type") or "").lower())
         for col in (overlay.column_info or [])
-        if col and str(col.get("type") or "").lower() in _NON_GROUPABLE_TYPES
+        if col and str(col.get("type") or "").lower() in NON_GROUPABLE_COLUMN_TYPES
     )
     if ungroupable:
         name, col_type = ungroupable[0]
@@ -395,7 +395,7 @@ def _validate_dissolve_by_field(dataset, by_field: str) -> None:
             detail="by_field conflicts with the generated 'source_count' column",
         )
     by_field_type = str(known_columns[by_field].get("type") or "").lower()
-    if by_field_type in _NON_GROUPABLE_TYPES:
+    if by_field_type in NON_GROUPABLE_COLUMN_TYPES:
         raise HTTPException(
             status_code=status.HTTP_422_UNPROCESSABLE_CONTENT,
             detail=(
--- a/backend/app/modules/catalog/datasets/api/router_analysis.py
+++ b/backend/app/modules/catalog/datasets/api/router_analysis.py
@@ -33,6 +33,7 @@ from app.platform.analysis_sql import (
     INTERSECT_OUTPUT_COLUMNS,
     MAX_MASK_LAYER_FEATURES,
     MAX_SOURCE_FEATURES,
+    INTERNAL_ALIAS_PREFIX,
     MEASURE_OUTPUT_COLUMNS,
     NON_GROUPABLE_COLUMN_TYPES,
     render_mask_expr,
@@ -312,6 +313,27 @@ def _validate_intersect_columns(source, overlay) -> None:
     # needs the aggregate to group by the two gids alone and join the overlay's
     # attributes back afterwards, which is a change to this query's shape
     # rather than a guard. Tracked separately.
+    # fix(#1097 review): a carried column may not sit in the alias namespace.
+    # Both layers, because an overlay carries columns from both and the inner
+    # aggregate names them in one select list alongside _gl_src_type; the
+    # overlay's are additionally re-listed inside _mask_pieces beside
+    # _gl_mask_gid and _gl_g. Checked against the PREFIX rather than the three
+    # alias names, so an alias added to that query later is covered without a
+    # second place to update.
+    reserved = sorted(
+        name
+        for name in (_column_names(source) | _column_names(overlay))
+        if name and name.startswith(INTERNAL_ALIAS_PREFIX)
+    )
+    if reserved:
+        raise HTTPException(
+            status_code=status.HTTP_422_UNPROCESSABLE_CONTENT,
+            detail=(
+                f"Column {reserved[0]!r} uses the {INTERNAL_ALIAS_PREFIX!r} "
+                "prefix, which this operation reserves for its own internal "
+                "columns. Rename it, or choose a different layer."
+            ),
+        )
     ungroupable = sorted(
         (col.get("name"), str(col.get("type") or "").lower())
         for col in (overlay.column_info or [])
--- a/backend/app/modules/catalog/datasets/api/router_analysis.py
+++ b/backend/app/modules/catalog/datasets/api/router_analysis.py
@@ -289,6 +289,45 @@ def _validate_intersect_columns(source, overlay) -> None:
                 "choose a different layer."
             ),
         )
+    # fix(#1097 review): the OVERLAY's columns, and only the overlay's.
+    #
+    # render_intersect_pairs groups by `_src.gid, _mp._mask_gid` plus every
+    # carried overlay column. The source's columns need no grouping — gid is a
+    # real table's primary key, so PostgreSQL licenses them by functional
+    # dependency — but `_mask_pieces` is a CTE with no key, so each overlay
+    # column it carries has to be named in the GROUP BY explicitly. json and
+    # xml have no equality operator, so grouping on one fails the CTAS with
+    # SQLSTATE 42883, and it fails there rather than in the preview: the
+    # preview carries gid and source_gid only, so it succeeds and the operation
+    # dies after the queue wait with an error naming a generated alias the user
+    # never wrote.
+    #
+    # Refused at enqueue instead, which is what _NON_GROUPABLE_TYPES already
+    # does for dissolve's by_field. Refusing rather than silently dropping the
+    # column is the same choice the sibling guards above make: an overlay whose
+    # output columns depend on which of them happened to be groupable is worse
+    # for anyone scripting against the result than one that says no.
+    #
+    # Carrying these through an overlay is a real gap and is worth doing — it
+    # needs the aggregate to group by the two gids alone and join the overlay's
+    # attributes back afterwards, which is a change to this query's shape
+    # rather than a guard. Tracked separately.
+    ungroupable = sorted(
+        (col.get("name"), str(col.get("type") or "").lower())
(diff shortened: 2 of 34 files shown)


ishiland:
@codex review

ishiland:
Four findings at `82fb7168d`, three of them real and fixed in `0bcfbb8af`. Taking them in order of how much they cost.

**P1, the exact-count query ran outside the preview semaphore.** Correct, and it defeated the bound rather than straining it. Operations that report an exact total run two sandbox statements and each opens its own connection; releasing the slot between them let a finished preview admit the next caller while still holding one, so `_MAX_CONCURRENT_PREVIEWS` stopped bounding connections, which is the only thing it exists to bound. Worth adding to the diagnosis: the count is the *likelier* of the two to still be running, since it is uncapped and scans both inputs while the geometry query stops at 500 rows. This costs throughput on `select_by_location` and `spatial_join` previews and that is the point — total connection demand is unchanged, only the accounting is fixed.

**P2, reserved spatial-join field names.** Confirmed by construction: the prefix is `join_` and the generated match column is `join_count`, so a join layer with an ordinary `count` column produces the name twice. Written as a duplicate check over the generated names rather than a ban on the literal `count`, so it keeps holding if the prefix or the generated set changes. The other way a name can repeat, the same field requested twice, turned out to be rejected a layer earlier by `AnalysisMaterializeRequest` — that rule had no test, so it has one now.

**P2, spatial-join pair counts when the source total is null.** The sharpest of the four. `source_feature_count: null` has two causes and only one means "read match_count instead": the operation filters rows, or the dataset's cached `feature_count` snapshot is absent. A spatial join on a snapshot-less dataset hits the second, and its `match_count` counts pairs, so the panel announced a total larger than a 1:1 result can be. The panel reads the operation now. The suggested remedy is what landed.

**P2, "the frontend never reads match_count".** This one describes the previous revision — `AnalysisPanel` already read `result.match_count` at the reviewed sha, under a comment explaining the null-keying. The remedy was right anyway, and it is the fix above.

Two things found while fixing rather than reported:

The two existing tests covering this read less than they appeared to. Both ran on the panel's default buffer operation, so the row-filtering assertion passed on the null source count alone and would have passed for every operation. They select an operation now.

`AnalysisPanel` is the only consumer of `match_count` — the chat path surfaces `source_feature_count` only, so it never had this bug.

Verification: full backend suite (6734 passed, 82 skipped), `ruff check`, `ruff format --check`, `test_layering.py` with the `service_analysis` cap raised 463 → 475, frontend typecheck/lint/i18n and the analysis suites. Each of the three fixes was reverted in turn to confirm its test fails without it; the semaphore assertion goes `[True, False]` pre-fix, and the collision guard turned out to be the only thing rejecting that request at enqueue.

@codex review

ishiland:
Both P2s were real. Fixed in `b880277b9`.

**Ungroupable overlay attributes.** Confirmed by construction: `render_intersect_pairs` groups by `_src.gid, _mp._mask_gid` plus every carried overlay column, and `_mask_pieces` is a CTE with no key so those columns cannot ride a functional dependency the way the source's do. Two details worth adding to the report. The source side is safe — `_src.gid` is a real table's primary key, so a json column there never reaches the `GROUP BY`, and a guard covering both sides would refuse overlays that work. And the failure is materialize-only: the preview carries `gid` and `source_gid` alone, so it succeeds first and the operation dies after the queue wait quoting a generated alias the user never wrote.

Refused at enqueue, which is what `_NON_GROUPABLE_TYPES` already does for dissolve's `by_field`, with a message naming the column and its type.

On the suggested remedy — grouping by the gids and joining the attributes back — that is the right end state and I did not do it here. It changes the shape of a new and fairly intricate query for a case a small guard covers correctly, and the guard converts an opaque post-queue 42883 into an immediate explained 422 either way. Filed as #1099 with the specific steps, and the one-sidedness is pinned by a test so the follow-up inherits it.

**Fields the picker offers and the server refuses.** Correct, and one of the three classes is mine: the `count` rejection landed in the previous round without a matching client-side filter. `Área` and `2020_pop` were pre-existing.

Fixed for both pickers rather than the one named. The finding is about the join picker, but the gap is in what the pickers know about the server's rules, and dissolve had the same hole — it filtered `source_count` and nothing else, so identifier-shape rejections and json group keys were offered there too. The dissolve precedent the report cites turned out to be one special case rather than a rule the join picker had failed to follow.

Worth stating explicitly, since it looks like a contradiction: being carried and being nameable in a request are different questions. `_list_carry_columns` still carries `Área` into analysis output by design (#763). These filters govern only which fields a request can name.

Verification: ruff, `test_layering.py`, the six analysis backend suites (201 passed), the frontend builder and analysis suites (2291 passed across 141 files), typecheck, lint, i18n. Both guards were reverted in turn to confirm their tests fail without them.

@codex review

ishiland:
Three findings from the last round, all real, fixed in `ba6006ea2` (rebased onto `ce69b8446`).

**P1, private join-layer provenance.** Confirmed and fixed. `join_dataset_id` and `join_fields` are persisted into `derived_from.params`, and `visible_derived_from` checked only the mask — so a public join output derived through a private join layer published that layer's UUID and the columns someone chose to copy out of it.

The detail that decided the shape of the fix: this is the second time the gap has opened. #765's review closed it for `mask_dataset_id`, and the next operation to carry a dataset id reopened it. Both times the code was correct for the operations that existed when it was written, so another branch would just wait for a third. It is a table now:

```python
_DATASET_ID_PARAMS = {
    "mask_dataset_id": (),
    "join_dataset_id": ("join_fields",),
}
```

Keys are what an unauthorized requester must not learn; values are what would still describe the layer once its id was gone. Dropping the id while keeping `join_fields` would leave the private layer's schema published, which is the more useful half of the disclosure.

Behind it is a structural test that parses the worker's actual provenance params and fails if any `*_dataset_id` key is missing from the table. Structural because the failure mode is silence — an unredacted id looks like ordinary provenance and nothing errors. Deleting the join row fails both the leak test and the structural test, and the structural one names the missing key.

**P2, overlay types after the queue wait.** Correct, and it was a TOCTOU on the guard added in the previous round. The enqueue check reads `column_info`, a snapshot, and a re-upload can replace the overlay before the job runs — the same window `_reject_output_column_collision` already exists for. Worth adding to the report: the live column list was already read there and would not have caught it, because the column that breaks the CTAS has an ordinary name. It needed types, not names.

`NON_GROUPABLE_COLUMN_TYPES` moved to `analysis_sql`, which the router and worker both already import. Three guards need it now and the worker must not import from the API layer; duplicating the set is how two halves of a guard drift.

**P2, join fields colliding with source columns.** Correct. Fixed by giving the picker the source's columns, and every rejection in that filter now compares the GENERATED name rather than the field's own, so they keep holding if the prefix or the generated set changes. The test mock had to become keyed on dataset id — a collision is a relationship between two layers, and a mock that hands both the same columns cannot express one.

**On the rebase.** #1048 landed underneath this and both had raised the cap on `analysis/tasks.py` — #1048 for the scoped `work_mem`, this branch for one CTAS branch per operation. Both rationales are true and neither number describes the merged file, so the cap is re-measured at 1231 rather than taken from either side, with both explanations kept. That conflict is the ratchet working.

Verification: full backend suite on the rebased tree, 6752 passed / 82 skipped. `ruff check`, `ruff format --check`, `test_layering.py`, frontend typecheck / lint / test:i18n, AnalysisPanel suite. Every fix reverted in turn to confirm its test fails without it.

@codex review

ishiland:
Both findings were real. Fixed in `424367647`.

**Carried columns colliding with the query's own aliases.** Confirmed against the SQL, including which layer each one affects: `_mask_gid` and `g` break from the overlay (they land in `_mask_pieces` beside those aliases), `_src_type` breaks from either layer (the inner aggregate names both layers' columns in one select list alongside it). The reason the existing guards missed them is exactly as described — they reserve names that reach the OUTPUT, and these never do.

I did not take the "reject these names" half of the suggestion. The aliases moved into a `_gl_` namespace and the namespace is what is reserved. A list of names is correct only for the aliases that exist when it is written, and this PR has already watched two such lists fall behind: the provenance redaction that knew about masks but not join layers, and the picker filters that knew about `source_count` and nothing else. A prefix covers an alias added to that query later without a second place to remember.

**Join fields not rechecked after the queue wait.** Correct, and it was the gap in an otherwise complete set — the source-collision and overlay-type rechecks beside it exist for precisely that window, and this one had a table re-resolution without a column re-resolution. Fixed, and the five live-schema rechecks now live in one function rather than scattered through `_materialize`; that also kept the extraction honest when the two new branches pushed it over C901.

**One correction worth recording, because it affects how much the last round's tests were worth.** My first version of both tests called the leaf guards directly. When I deleted the call sites to check the tests were load-bearing, only the enqueue-side test failed — the two worker-side tests stayed green, because a guard that rejects a bad list proves nothing about whether anything hands it one. They run through `_resolve_and_validate_columns` now, and the deletion check fails both. Same class of error as the finding itself: correct about the thing in front of it, silent about the neighbour.

Verification: full backend suite 6757 passed / 82 skipped, `ruff check`, `ruff format --check`, `test_layering.py` with `analysis_sql` 1173 → 1191 and `tasks` 1231 → 1322, frontend gates unchanged this round. Merges cleanly against `main` at `34b6440fd`.

@codex review

ishiland:
Both real, fixed in `ae7489b39`.

**Mask geometry not rechecked after the queue wait.** Confirmed. Re-applied in `_resolve_layer_table_ref`, which already holds the ORM object. Two things worth adding to the report. The message it produces is the reason this deserves a guard rather than a shrug: nothing errors, the mask simply matches no rows, and the job dies on "Analysis produced no features" — which sends the user to inspect their own data instead of the layer that changed underneath them. And the requirement is deliberately one-sided: a JOIN layer is not held to it, because a spatial join counts in any direction and the picker offers points on purpose.

**Stale `join_field` after a source change.** Confirmed, and the diagnosis in the finding is exactly right — the handler resets `by_field` and not `join_field`. The subtlety worth stating: the field belongs to the JOIN layer, so surviving a source change looks correct; what makes it invalid is that `join_<name>` must not collide with a SOURCE column.

I took both remedies rather than choosing. Clearing on source change covers how it went stale in practice; gating `paramsValid` on the field still being one the picker would offer covers the rest, because the field and the rule validating it depend on two different layers and either can move. The join LAYER is cleared too when it becomes the source, matching what the mask picker already does.

**One thing found while testing, recorded because it will bite the next person.** My first version of that test wedged the suite — 905 seconds and a heap OOM. Isolating it reproduced the hang alone, which ruled out the component. It switched the source ONTO the join layer, so the new self-join clear unmounted both join selects while the select menu was still open, and userEvent never settled. Using a third layer as the join target runs in 951ms. It is a test-environment artifact rather than a product risk: the mask picker has done the identical `if (v === maskLayerId) setMaskLayerId(...)` since it shipped.

Verification: full backend suite 6759 passed / 82 skipped, `ruff check`, `ruff format --check`, `test_layering.py` with `tasks` 1322 → 1343, frontend typecheck / lint / test:i18n, AnalysisPanel 112 passed. Both guards reverted in turn to confirm their tests fail without them. Merges cleanly against `main`.

@codex review

ishiland:
All three real. Fixed in `5ddb0fec9`, with the regeneration in `de6075aea`.

**Preview approving what Create refuses.** Confirmed, and this is the worst of the two verdicts to get wrong — by the time Create fails, the user has already been told it works. The guard was `if body.join_fields`, and the validator it skipped also checks the always-generated `join_count` against the source's columns. Preview calls it unconditionally now. The test asserts the two paths return the SAME status rather than checking each in isolation, since agreement is the actual invariant.

**Join layer geometry.** Correct, and the distinction you drew is the important part: the previous round's polygon recheck was the MASK's rule, and applying it here would have refused point and line joins that work fine. So the flag is two strengths now rather than a boolean — `"polygonal"` for the mask, `"any"` for the join — and the test pins both sides, including that a POINT join layer still resolves.

**`mask_dataset_id` description.** Correct, and load-bearing outside this repo: the text ships into both generated SDKs. Rewritten on both request models to say it is required for intersect and that a drawn mask is rejected there, with the reason (an overlay carries the second layer's attributes onto its output, and a drawn polygon has none). All three OpenAPI artifacts regenerated — worth noting `make sdks` drives two of them and `frontend/src/types/api.generated.ts` has its own generator.

**One self-inflicted break worth recording.** Renaming `require_polygonal` to `require_geometry` broke two tests from the previous round. After the rename I re-ran the suite holding the NEW tests and not the one holding the OLD callers, so a targeted run came back green and the full suite failed 21 minutes later. I verified where the work was rather than where the signature was used. Fixed, and I grepped the whole backend to confirm nothing else referenced the old name instead of assuming two hits was all of them.

Verification: full backend suite 6761 passed / 82 skipped, `ruff check`, `ruff format --check`, `test_layering.py` with schemas 1127 → 1133 and tasks 1343 → 1365, `make openapi-check`, frontend types regenerated. Both new guards reverted in turn to confirm their tests fail without them. Merges cleanly against `main`.

@codex review

ishiland:
Real, and fixed in `a26346925`.

Confirmed the mechanism against the running server rather than from memory, because the whole bug turns on it being a NOTICE rather than an error:

```
NOTICE:  identifier "aaaa…70 chars" will be truncated to "aaaa…63 chars"
max_identifier_length = 63
```

Silent truncation is what makes this reachable. Two join fields sharing their first 58 characters prefix to strings that differ in Python and are the same column in the output, so the uniqueness guard passed and the CTAS failed after the queue wait.

Worth adding to the report: preview does not catch it either, for the reason you gave — it consumes the values positionally. That makes this the second route to the same split the previous round closed for `join_count`: Preview approving a request Create cannot save.

I took the truncation half of the suggestion rather than rejecting over-long names. Applying PostgreSQL's own truncation at the generator means the router's two checks, the worker's recheck and the uniqueness rule all compare what the database will compare, instead of three call sites each needing to remember a length rule. Source columns need no equivalent treatment: they already exist in a table, so the server truncated them at creation — which is also why comparing a truncated generated name against them is sound.

Verification: full backend suite 6763 passed / 82 skipped, `ruff check`, `ruff format --check`, `test_layering.py` with `analysis_sql` 1191 → 1218. The truncation was reverted to confirm both tests fail without it. Merges cleanly against `main`.

@codex review

ishiland:
Both real, fixed in `b43997cb9`. The first one also corrects something I claimed in round 3, so I'll take that first.

## Correction to the round-3 P1

I reported that a public spatial-join output published a private join layer's UUID and column names to every viewer. **That was not reachable, and I overstated it.**

`build_derived_from` filters through `PARAM_KEYS`, which contained neither `join_dataset_id` nor `join_fields`, so the id never reached `records.derived_from` at all. The redaction I added in that round was correct code guarding a path that did not exist.

The reason I did not notice is the more useful half. The structural test I added alongside it parsed the params dict at the `_materialize` call site — what is PASSED — while the contract that matters is `PARAM_KEYS`, what is STORED. It asserted a pairing between two values with no bearing on the database. It passed throughout and was confirming nothing, which is exactly the failure mode I wrote it to prevent, one layer up.

It reads `PARAM_KEYS` now, and there is an end-to-end test beside it that runs the real worker and asserts both values arrive in `records.derived_from`. Those two checks are separable — a whitelist assertion alone would have stayed green through the entire window where nothing was being stored — so both exist.

## The finding itself

Correct: a join's durable lineage recorded the source and the operation and nothing about the second layer, so nothing explained where the `join_*` columns on the output came from.

Worth stating because it changes how the two changes should be read together: adding `join_dataset_id` to `PARAM_KEYS` is precisely what makes the round-3 redaction load-bearing. Landing this fix **without** that redaction would create the leak I mistakenly described. They now have to move together, and the structural test is what enforces that.

## The picker

The frontend mirror of the previous round, and correct — `spatial_join_output_columns` truncates to 63 bytes and this filter did not, so a field whose alias collides with a source column only AFTER truncation was offered and then refused. The test case is one that does not collide until it is cut, so it fails against the untruncated comparison.

Verification: full backend suite 6764 passed / 82 skipped, `ruff check`, `ruff format --check`, `test_layering.py`, frontend typecheck / lint / test:i18n, AnalysisPanel 112 passed. All three reverted in turn to confirm their tests fail without them. Merges cleanly against `main`.

@codex review

ishiland:
Both real, fixed in `3ec8c341c` with the regeneration in `5a2e2082e`.

**`match_count`'s contract.** Correct, and the way it happened is worth recording: round 2 corrected this exact text in `frontend/src/types/api.ts` — the hand-typed mirror — and left the Pydantic schema it mirrors saying the old thing. Backwards. The schema is the source `openapi.json` and both SDKs generate from, so every generated client was told to ignore the one field needed to read a truncated intersect preview, while the mirror that generates nothing was right.

Rewritten to say what the value counts per operation rather than listing which operations return it, including that spatial_join's is NOT the output total. All three artifacts regenerated.

**Job metadata.** Correct. What sharpens it is where the gap sits: re-uploaded mid-queue, wrong geometry type, ungroupable column are all failure modes the last four rounds added guards for — and every one of them is a failure caused by the layer this metadata did not name. So the branch spent four rounds making those failures detectable and left the operator unable to tell which layer caused one.

Recorded for every operation that consumes a layer now. `mask_source` stays scoped to the ones that can take a DRAWN mask: intersect rejects one, so `"layer"` there would be a constant dressed up as a discriminator, and the test pins that distinction rather than only the presence of the id.

Verification: full backend suite 6766 passed / 82 skipped, `ruff check`, `ruff format --check`, `test_layering.py` with schemas 1133 → 1138, `make openapi-check`. Both changes reverted in turn to confirm their tests fail without them. Merges cleanly against `main`.

@codex review

ishiland:
Real, fixed in `78c54e27b`. This is the sibling of the previous round's finding, and that is the part worth recording.

`mask_source` has **two writers**: the router puts it on the job (Admin Jobs reads that to diagnose a run) and the worker puts it in provenance (the durable lineage). Last round's review named the job-metadata writer. I fixed that one and left this one, so one defect became two rounds.

For a drawn `select_by_location` the consequence is total rather than partial. The drawn geometry is deliberately excluded from provenance because it can be kilobytes, so this discriminator is the ONLY trace that an area shaped the selection — without it the params serialise empty and the lineage says a selection happened by no visible means.

Beyond the fix, the two constants are pinned together by a test. They are duplicated rather than shared because `processing/` must not import from `app.modules.catalog` (PROCESS-02), and that duplication is exactly why the previous fix reached one copy and not the other. The test fails if they ever disagree again.

`intersect` stays out of the set deliberately: it takes a layer only, so the discriminator would be a constant there rather than a discriminator — same reasoning as the job-metadata side, and the test asserts the two sets match rather than asserting a literal.

Verification: full backend suite 6768 passed / 82 skipped, `ruff check`, `ruff format --check`, `test_layering.py` with `tasks` 1365 → 1385. Reverted to confirm the drawn-selection test fails without the fix. Merges cleanly against `main`.

@codex review

ishiland:
Real, fixed in `e00913de4`. This is the first finding in eleven rounds that is a product gap rather than a defence, so it is worth being precise about what was missing.

The lineage sentence is a user-facing surface — the dataset page renders it, search indexes it, DCAT exports it. All four operations this branch adds fell through to `"<operation> applied to <source>"`, so an overlay whose entire point is the second layer was described without naming one. Each has a phrase now: joined-against (with the transferred fields when there are any), intersected-with, selected-by, measurements-computed-from.

The title resolution was the more interesting half of the report. `mask_title` was fetched only when `mask_source == "layer"`, and intersect has no `mask_source` — that discriminator exists to separate DRAWN from LAYER, and intersect can only be a layer. So the overlay's title was never read, and its sentence would have said "another layer" about a layer whose title was one query away. It is keyed off the id being present now.

Using a discriminator as a proxy for presence is the same mistake as the frontend's `source_feature_count` null check from round 2: the null had two causes and only one of them meant what the reader assumed. Third time this branch has hit that shape, which is why I am naming it rather than just fixing it.

I also pinned the degraded path: a deleted or unreadable layer yields no title, and the sentence must still read as a sentence rather than leaking a bare UUID. That is the treatment clip already had, extended to every operation that now names a layer.

Verification: full backend suite 6770 passed / 82 skipped, `ruff check`, `ruff format --check`, `test_layering.py`. The new phrases reverted to confirm both tests fail without them. Merges cleanly against `main`.

@codex review

ishiland:
Real, and **I introduced it last round**. Fixed in `d605a905e`.

Round 11 named the transferred fields in the lineage sentence to make it more useful. `lineage_summary` is stored once and served raw — the dataset response returns it, search indexes it, and the DCAT / DCAT-US / GeoDCAT-AP services export it — and none of those pass it through `visible_derived_from`, which is what access-checks the structured provenance per requester.

So that change routed a private join layer's column names around the exact redaction added in round 3 and made load-bearing in round 8. `join_fields` is listed in `_DATASET_ID_PARAMS` as a dependent of `join_dataset_id` **precisely because** a column list is most of a schema. Sensitive by this branch's own rule, published by this branch's own prose.

I verified the exposure rather than taking it on the report: the field is on the response schema with no filter, in three export services, and in the search indexer.

Removed. The test asserts the absence and the reason — it also checks `join_fields` is still a declared dependent, so removing it from the redaction table fails a test that explains why it mattered.

The general lesson, since it is not obvious: **making prose more informative is a disclosure decision whenever the prose has no per-requester form.** Every field this sentence can name has to be one the least-privileged reader may see.

## An adjacent exposure, filed rather than fixed

The sentence also names the second layer's TITLE, which leaks a private layer's existence the same way. That predates this PR — clip has shipped it since #765 — and round 11 extended the pattern to three more operations consistently with clip rather than inventing it.

Closing it changes clip's shipped output for existing records and costs the sentence most of its value, so it is a product call: **#1103**, with three options and my read. I did not want to make that decision inside a review round.

Verification: full backend suite 6771 passed / 82 skipped, `ruff check`, `ruff format --check`, `test_layering.py`. The field names restored to confirm the redaction test catches them. Merges cleanly against `main`.

@codex review

ishiland:
Both round-13 findings are real. Both fixes landed in `2fa7eac5e`, and both are fixed at the class rather than the cited line, because each finding named one member of a wider family.

**Curved geometries.** The finding named measure's `::geography` cast. Testing every function the new operations use against actual MultiSurface/CompoundCurve input turned up six failing sites in this diff, in three families: `::geography` raises ("Geography type does not support MultiSurface"), `ST_MakeValid` raises ("unsupported geometry type"), and `ST_AsGeoJSON` raises (GeoJSON has no curved types). So spatial_join and intersect were broken through `ST_MakeValid`, and every pass-through preview (measure, spatial_join, select_by_location) was broken at serialization even where the computation would have survived. A CTAS would also have stored curved rows into the derived table, breaking that layer's tiles and feature reads later.

The fix is one rule at one boundary: `linearized()` in `analysis_sql` wraps every geometry read that feeds a curve-intolerant function or passes through into output. `ST_CurveToLine` is an exact no-op on linear input (`ST_Equals` verified). WHERE predicates stay on the bare column: `&&`, `ST_Intersects`, `ST_Dimension` and `ST_IsEmpty` all accept curves directly, and wrapping there would defeat the GIST index. One adjacent bug fell out of the same sweep: intersect's `GeometryType(...) = 'LINESTRING'` check read `COMPOUNDCURVE` on a curved source and silently skipped `ST_LineMerge`; it now reads the linearized type.

The class is wider than this PR. Tiles (`ST_AsMVTGeom`), the features browse API (`ST_AsGeoJSON`), and buffer/centroid/clip/dissolve (`ST_MakeValid`) all raise on curved datasets on main today, none of it in this diff. #1104 tracks the structural fix (linearize `geom_4326` at ingest plus a backfill), after which the analysis-side wraps become removable.

**Binary join values.** Confirmed, and it is specific to this PR's path: the features browse API builds properties in SQL (`to_jsonb(t.*)`), where PostgreSQL renders bytea as a hex string, so raw bytes never reach Python there. This preview builds properties in Python, handing driver values straight to Pydantic. `_json_safe` now encodes bytes at the single properties boundary, using the same `\x` hex form `to_jsonb` gives, so one column reads identically through both endpoints.

**Tests.** `test_analysis_curved_sources.py`: curved-source preview for all four operations (intersect with a curved mask as well, covering both wrap sites in one statement), a curved-source materialize asserting the derived table stores linear geometry, and a bytea spatial-join transfer asserting the hex form. Stubbing `linearized()` to identity fails five of six; stubbing `_json_safe` fails exactly the sixth. Measure's numbers are checked against PostGIS's own measure of the linearized geometry, not constants.

Rebased onto `34b6440fd`. Full backend suite: 6791 passed, 82 skipped. The previous run's CLI Tests failure was a Docker Hub connection timeout pulling the postgres image, unrelated to the diff.

@codex review

ishiland:
Confirmed. The scalar fix stopped one representation of the same value and let the container form through: a `bytea[]` join field comes back from the driver as a list of `bytes`, and `_json_safe` returned the list unchanged, so each element still hit the serializer raise. `22e346758` makes the encoder recurse through lists, tuples and dicts with the same scalar encoding. `to_jsonb` renders a `bytea[]` as an array of `\x` hex strings (verified against the running database), so the recursive form stays byte-identical with what the features browse API serves for the same column.

The existing bytea test now transfers both a `bytea` and a `bytea[]` field and asserts both encodings; stubbing the container branches back out fails exactly that test. Scalar driver types Pydantic serializes natively (datetime, date, Decimal, UUID) still pass through untouched, so nothing already-working changes representation.

Also rebased onto `c0cbc5dfd`, which clears the out-of-date state against #1102. Full backend suite on the pushed tree: 6792 passed, 82 skipped.

@codex review

ishiland:
Confirmed, and fixed at the class in `f0c005daf`. Verified live: a `json[]` column reports `data_type = 'ARRAY'` with the element type only in `udt_name` (`_json`), `GROUP BY` on it raises 42883, and `int[]` groups fine — so the check names elements, not arrays. `_ungroupable_type_name` reads `udt_name` beside `data_type` and the overlay recheck uses it.

Enumerating the other GROUP BY consumers turned up the sibling: dissolve's `by_field` had the identical blind spot and no live recheck at all, so a `json[]` group field passed enqueue and killed the CTAS after the queue wait. The same predicate now guards it, and the one query also catches a `by_field` a re-upload dropped.

The router keeps its scalar-only snapshot check, with a comment stating why: `column_info` stores `'ARRAY'` with no element type, so array forms are invisible at enqueue by construction and the worker's live recheck is their sole guard — the same trust division every recheck in this set already uses. Tests cover the `json[]` overlay rejection with its real type name, the `int[]` negative control, and a dissolve-by-`json[]` driven through `_materialize` so the wiring is what's proven; stubbing the `udt_name` branch fails exactly those tests. Full suite: 6794 passed.

With this round the review has converged: fifteen rounds, severity falling from P1s to P2s on progressively rarer inputs (curved WFS geometries, binary columns, array element types). Merging on green CI; anything further is welcome as an issue.

PR #1118, opened 2026-08-02
fix(security): bound the URL redaction scheme prefix against quadratic backtracking

### LINKED ISSUE #1116: url_redaction: URL_LIKE_RE backtracks quadratically on non-matching alphanumeric runs
## Problem

`URL_LIKE_RE` in `backend/app/core/url_redaction.py:10`:

```python
URL_LIKE_RE = re.compile(r"(?:(?:[A-Za-z0-9_+.-]+:)?https?://)[^\s\"'<>]+")
```

The optional scheme prefix `(?:[A-Za-z0-9_+.-]+:)?` is ambiguous against the `https?` that follows: on a long run of `[A-Za-z0-9_+.-]` characters that never completes a match, the backtracking engine retries every prefix length at every start position — O(n²). CodeQL flags this as py/polynomial-redos (alert #38), and unlike the rest of the current alert backlog this one is real.

The regex fallback only runs for scheme-less free text (`redact_url_credentials` routes anything that isn't an http(s) URL there), and that free-text path is fed by GDAL/ogr2ogr stderr and source-preview text (`processing/ingest/manifest_sources.py`, `processing/ingest/validation.py`, `catalog/sources/preview.py`) — strings that can echo attacker-influenced content such as hostile layer names or URLs from an uploaded file. A ~100 KB alphanumeric run is enough to spin a worker CPU for a noticeable stretch.

## Impact

Worker/API CPU exhaustion (nuisance DoS) on the redaction path. No data exposure.

## Proposed fix

Bound the ambiguous quantifier so the quadratic factor becomes a small constant:

```python
URL_LIKE_RE = re.compile(r"(?:(?:[A-Za-z0-9_+.-]{1,64}:)?https?://)[^\s\"'<>]+")
```

64 comfortably covers every registered URI scheme. Optionally also cap the input length handed to the fallback (stderr lines are already truncated elsewhere; a defensive `[:100_000]` here makes the worst case flat). One regression test with a long `aaaa...` run under a timeout guard.

CodeQL alert: https://github.com/geolens-io/geolens/security/code-scanning/38

ishiland:
## Problem

`URL_LIKE_RE` in `backend/app/core/url_redaction.py` left the optional scheme prefix unbounded:

```python
URL_LIKE_RE = re.compile(r"(?:(?:[A-Za-z0-9_+.-]+:)?https?://)[^\s\"'<>]+")
```

That prefix is ambiguous against the `https?` that follows it. Given a long run of prefix-class characters that never completes a match, the engine retries every prefix length at every start position, which is quadratic in the input. CodeQL alert #38 flags it as `py/polynomial-redos`, and it is the one alert in the current backlog that is real rather than noise.

The regex only runs on the scheme-less free-text fallback, but that path takes attacker-influenced input. `processing/ingest/validation.py` calls `redact_url_credentials(raw_path)[:200]` at three sites where `raw_path` is the `<SourceFilename>` of a user-uploaded VRT, and the truncation happens after the regex, so it protects nothing. `catalog/sources/preview.py` hands over untruncated `ogrinfo` stderr.

Impact is worker and API CPU exhaustion. No data is exposed.

## Fix

Bound the quantifier to `{1,64}`, which is longer than any registered URI scheme.

A prefix longer than 64 characters still gets redacted. The prefix group is optional, so the match simply starts later in the string and still spans the whole URL. Tests pin that at 65, 200 and 5,000 characters so the fix cannot regress into dropping credentials it used to mask.

Measured on this branch, substituting over a run of `a` characters:

| input | unbounded | bounded |
|---|---|---|
| 10,000 | 0.085s | 0.0012s |
| 25,000 | 0.525s | 0.0032s |
| 50,000 | 2.098s | 0.0061s |

The old form quadruples for each doubling of input. The new one scales linearly.

## Why this shape

The issue also offers a defensive `[:100_000]` cap on the input handed to the fallback. That is not in this change. With the bound in place, 200 KB of hostile input redacts in roughly 0.02 seconds, so the cap would buy nothing measurable, and truncating the input carries its own risk: a cut landing mid-URL could leave a credential on the far side of the boundary and defeat the redaction this module exists to perform. Trading a fixed DoS for a possible disclosure is the wrong direction.

## Verification

```
cd backend
uv run pytest tests/test_url_redaction_sec.py -q     # 26 passed
uv run pytest tests/test_layering.py -q              # 36 passed
uv run ruff check .                                  # All checks passed
uv run ruff format --check .                         # 875 files already formatted
```

The timing regression test asserts the returned string as well as the elapsed time, so a regex that matched nothing at all would be fast and would still fail. Its threshold sits about 80x above the measured bounded time, which keeps it decisive on a loaded runner without being tuned to a number that will flake.

Closes #1116

### DIFF
--- a/backend/app/core/url_redaction.py
+++ b/backend/app/core/url_redaction.py
@@ -7,7 +7,14 @@ from urllib.parse import parse_qsl, urlencode, urlsplit, urlunsplit
 
 REDACTED_QUERY_VALUE = "<redacted>"
 REDACTED_USERINFO = "redacted"
-URL_LIKE_RE = re.compile(r"(?:(?:[A-Za-z0-9_+.-]+:)?https?://)[^\s\"'<>]+")
+# fix(#1116): the optional scheme prefix is bounded to 64 characters, which is
+# longer than any registered URI scheme. An unbounded `+` here is ambiguous
+# against the `https?` that follows, so a long run of prefix-class characters
+# that never completes a match made the engine retry every prefix length at
+# every start position — O(n²) on a fallback that is fed GDAL stderr and
+# uploaded-VRT paths. A longer prefix still redacts: the match simply starts
+# later in the string and still spans the whole URL.
+URL_LIKE_RE = re.compile(r"(?:(?:[A-Za-z0-9_+.-]{1,64}:)?https?://)[^\s\"'<>]+")
 
 SENSITIVE_QUERY_PARAMS = frozenset(
     {
--- a/backend/tests/test_url_redaction_sec.py
+++ b/backend/tests/test_url_redaction_sec.py
@@ -1,5 +1,7 @@
 """Security tests for URL query credential rejection/redaction."""
 
+import time
+
 import pytest
 from pydantic import ValidationError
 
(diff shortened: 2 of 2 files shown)


ishiland:
@codex review

ishiland:
@codex review

PR #1120, opened 2026-08-02
test(analysis): finish the backend analysis coverage gaps and rehome the sync tests

### LINKED ISSUE #789: test(analysis): close the backend analysis coverage gaps

### LINKED ISSUE #1085: test(analysis): move the two pure _user_error_message tests out of TestMaterializeWorker
Small tidy-up, deliberately split out of #1030 rather than folded into it.

## What

`test_analysis_materialize.py` has two tests sitting in the wrong class:

```
TestMaterializeWorker::test_timeout_message_names_the_budget_that_fired   (:2571)
TestMaterializeWorker::test_analysis_timeouts_track_their_settings        (:2591)
```

Both are pure unit tests of `_user_error_message` and `MATERIALIZE_TIMEOUT` — sync, `monkeypatch` only, no database, no worker. `TestMaterializeWorker` is otherwise entirely async tests that run `_materialize` against a real table.

The file already has the right home for them: `TestUserErrorMessage`, whose five tests are the same shape and which exists precisely to hold the sync, DB-free ones.

## Why it is filed rather than done

They arrived in #1057 and were the trigger for #1030's fixture bug — a module-level autouse **async** fixture cannot be handed to a sync test, so they errored at setup and poisoned the `client` FixtureDef for the rest of the worker (~950 errors, not the 25 a single-file run shows).

#1030 fixes that at the fixture, which is the right layer: the teardown is synchronous now and no longer constrains what shape any test in the module may be. So this relocation is **no longer required for correctness** — it is only tidiness, and folding a tidy-up of freshly merged code into a bug fix would have made #1030 harder to review for the thing that actually mattered.

## Not a substitute for the fixture fix

Worth recording, because it was considered and rejected as the fix for #1030: relocation alone would have left the async fixture and its class-level override in place, so the next sync test added to any of the other four classes would have hit the same trap. It fixes one instance; #1030 removes the shape.

## Acceptance

- The two tests live in `TestUserErrorMessage`, unchanged in body.
- `TestMaterializeWorker` contains only async tests.
- Full suite green under `-n 4 --dist loadgroup` (this file's failure mode is invisible to a single-file run).

Refs #1030, #1057, #1082

ishiland:
## Problem

Two small test-suite gaps, both in the same file family, so they land together.

**#789** was re-scoped to backend pytest only, and three of its four items already shipped in #1030 (verified: `fix(#789)` docstrings at `test_analysis_materialize.py:149` and `:1084`, and `test_analysis_preview.py:803`). Item 4 was never done, and a fifth gap opened after item 2 landed.

- The enqueue path stamps `job.current_step = "queued"`, and nothing asserted it. That stamp carries weight because #703 put analysis below the default job priority, so a queued analysis job now waits behind uploads by design. The stamp is the only signal that separates "waiting as intended" from "stuck".
- `_SANDBOX_STATUS` gained a `query_at_capacity` category in #1014, after item 2's parametrize was written against the four that existed then. The new category simply went untested and the suite stayed green.

**#1085** had two sync unit tests of `_user_error_message` sitting in `TestMaterializeWorker`, which is otherwise async tests running `_materialize` against a real table.

## Fix

For #789, assert the `queued` stamp at enqueue, add the missing category to the parametrize, and pin the mapping itself. The new `test_every_sandbox_status_category_is_parametrized` asserts the key set of `_SANDBOX_STATUS`, so the next category added fails here until someone gives it a case. That is the part worth keeping: the hand-kept parametrize is what let #1014's category slip through, and one more entry would not have stopped it happening again.

For #1085, move the tests into `TestUserErrorMessage`, bodies unchanged.

Four tests moved, not the two the issue names. Its second acceptance clause is that `TestMaterializeWorker` contains only async tests, and two `work_mem` tests were also sync. Those went to a new `TestMaterializeWorkMem` class rather than into `TestUserErrorMessage`, since they test a different function. `TestMaterializeWorker` is now 0 sync and 39 async, confirmed by walking the AST rather than by reading.

The four moved bodies are byte-identical to their originals, checked by extracting each function from both commits and comparing the strings.

## Why this shape

The `current_step` assertion is load-bearing rather than decorative. `current_step` is `Mapped[str | None]`, nullable with no column default, and `create_ingest_job` never writes it, so the router at `router_analysis.py:688` is the only writer on this path. Remove the stamp and the assertion fails on `None`.

`_SANDBOX_STATUS` has four keys and exactly one consumer, at `router_analysis.py:404`, so there is no second call path where the mapping could be read differently.

## Correction to the #789 issue body

The body says item 4 is "still gated on #695, which is open". That is wrong in both halves. The `current_step` stamp exists today, and the #695 below-default-priority assertions are also already landed at `test_analysis_materialize.py:227-230`. Nothing here was gated on anything.

## Verification

```
cd backend
# full suite, required by #1085 (its failure mode is invisible to a single-file run)
uv run pytest -n 4 --dist loadgroup
#   6829 passed, 83 skipped, 270 warnings in 696.59s

uv run pytest tests/test_analysis_materialize.py tests/test_analysis_preview.py
#   158 passed
uv run pytest tests/test_layering.py -q     # 36 passed
uv run ruff check .                          # All checks passed
uv run ruff format --check .                 # 875 files already formatted
```

The relocated tests were re-run by node ID to confirm the move, not just re-read: all four pass in their new classes. Every symbol they reference was already a module-level import, so class membership never affected resolution.

Closes #789
Closes #1085

### DIFF
--- a/backend/tests/test_analysis_materialize.py
+++ b/backend/tests/test_analysis_materialize.py
@@ -2616,48 +2616,6 @@ class TestMaterializeWorker:
         # Same transaction, ahead of the CTAS it protects.
         assert executed.index(hashaggs[0]) < executed.index(ctas[0])
 
-    def test_timeout_message_names_the_budget_that_fired(self, monkeypatch):
-        """fix(#1013 review): the two budgets are independently configurable
-        now, so a registration timeout quoting the materialize budget would
-        send an operator to tune the setting that did not fire."""
-        from sqlalchemy.exc import OperationalError
-
-        from app.core.config import settings
-        from app.processing.analysis.tasks import _user_error_message
-
-        monkeypatch.setattr(settings, "analysis_materialize_timeout_seconds", 300)
-        monkeypatch.setattr(settings, "analysis_registration_timeout_seconds", 900)
-        exc = OperationalError(
-            "SELECT 1", {}, Exception("canceling statement due to statement timeout")
-        )
-
-        assert "300s" in _user_error_message(exc)
-        assert "900s" not in _user_error_message(exc)
-        assert "900s" in _user_error_message(exc, registered=True)
-        assert "300s" not in _user_error_message(exc, registered=True)
-
-    def test_analysis_timeouts_track_their_settings(self, monkeypatch):
-        """fix(#1013): the budgets are operator settings now, and they are read
-        at call time — a module-level snapshot would freeze whatever the
-        settings object held when this module was first imported."""
-        from app.core.config import settings
-
-        from app.processing.analysis.tasks import registration_timeout
-
-        # Set the baseline rather than asserting the shipped defaults: settings
-        # is a module-level singleton, so an ANALYSIS_*_TIMEOUT_SECONDS in the
-        # environment or the root .env would already have overridden them and
-        # the "before" assertions would fail for a reason unrelated to the
-        # behaviour under test.
-        monkeypatch.setattr(settings, "analysis_materialize_timeout_seconds", 300)
-        monkeypatch.setattr(settings, "analysis_registration_timeout_seconds", 600)
-        assert materialize_timeout() == "300s"
-        assert registration_timeout() == "600s"
-        monkeypatch.setattr(settings, "analysis_materialize_timeout_seconds", 1200)
-        monkeypatch.setattr(settings, "analysis_registration_timeout_seconds", 1800)
-        assert materialize_timeout() == "1200s"
-        assert registration_timeout() == "1800s"
-
     async def test_ctas_scopes_work_mem_to_its_transaction(
         self,
         test_db_session: AsyncSession,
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

Issue #1117, opened 2026-08-02
bug(auth): admin OAuth provider changes never invalidate the login page's provider list


ishiland:
## Problem

The admin auth-settings surfaces invalidate the wrong query key family after OAuth/SAML provider mutations, so the login page's provider buttons go stale for the rest of the session.

- `frontend/src/components/settings/SettingsAuthTab.tsx:145,157,168` and `frontend/src/components/settings/SamlProvidersSection.tsx:135,148,160` invalidate `settingsOAuth.providers` after create/update/delete.
- The login page's `OAuthButtons.tsx:86` reads a DIFFERENT key: `authConfig.oauthProviders`.
- Neither surface invalidates the other's key, so an admin who adds or removes a provider and then logs out sees the old button set until a full reload.

## Fix

Add `queryClient.invalidateQueries({ queryKey: queryKeys.authConfig.oauthProviders })` (exact key name per `frontend/src/lib/query-keys.ts`) alongside the six existing `settingsOAuth.providers` invalidations, or fold both into one invalidator hook so the pairing can't drift again.

## Verification

Vitest: mutate a provider via the settings hooks with a spy on `invalidateQueries`, assert both key families are hit. Manual: add a provider in admin settings, log out, confirm the new button appears without a reload.

Found during the 2026-08-02 cache-invalidator sweep for #1021 (this is the inverse class: a missing invalidation rather than an inert one).

PR #1121, opened 2026-08-02
chore(frontend): document three invalidators whose liveness is not local

### LINKED ISSUE #1021: chore(frontend): sweep for cache invalidators whose surface never populated the cache
Decomposed from #893 (item 13). Verified against `main` at `9a788ba92` on 2026-07-31.

## The pattern

A component calls a shared cache invalidator that looks like it participates in cross-surface invalidation, but the surface it lives on never populated that cache. The call is a no-op. Nothing fails, no test catches it, and the behaviour the call was meant to provide is quietly absent.

The known instance: `useInvalidateTileTokens()` in `ViewerMap` backs the #438 BLD-04 WebGL-context-restore path, and it was a **no-op for four or more releases** because nothing on the viewer route used the query cache. #897's TanStack unification fixed that instance, and the fix is recorded at `frontend/src/components/viewer/hooks/use-viewer-tokens.ts:47`, which now notes the change "makes ViewerMap's existing `useInvalidateTileTokens()` (WebGL context restore, fix(#438) BLD-04) actually reach the viewer's tokens."

What has not happened is the sweep for others.

## Why no test catches it

The failure is a call that does nothing rather than a call that does the wrong thing. A unit test mocking the invalidator sees it called and passes; the existing tests do exactly that (`useInvalidateTileTokens: () => vi.fn()` in `DatasetMap.test.tsx:45` and `BuilderMap.terrain-visibility.test.tsx:45`). An integration test would have to assert that a cache entry actually changed, which is not what any of them assert. So the sweep has to be done by reading, not by running something.

## Scope

Start from the three remaining `useInvalidateTileTokens` call sites and check each one against the cache its surface actually populates:

- `frontend/src/components/dataset/DatasetMap.tsx:169`
- `frontend/src/components/viewer/ViewerMap.tsx:1019` (the fixed instance, as the reference for what "correct" looks like)
- `frontend/src/components/builder/BuilderMap.tsx:607`

Then widen to the general shape: any `queryClient.invalidateQueries` or hook wrapping one, called from a surface that fetches the same data by another route (direct `apiFetch`, local state, props). The question for each is "does this surface have an entry in that cache to invalidate?", and a negative answer is the finding.

## Output

A list of sites with a verdict each, and a fix or a comment at every site that turns out to be inert. A site that is inert and should stay inert wants a comment saying so, since the next reader will otherwise re-derive this from scratch, which is how the original instance survived four releases.

## Verification

```
```

ishiland:
## Problem

A component can call a shared cache invalidator that looks like it participates in cross-surface invalidation while the cache it names can never hold an entry for that caller. The call does nothing, no test fails, and the behaviour it was meant to provide is quietly absent. The known instance survived four or more releases.

The sweep covered 151 `invalidateQueries` sites plus the 5 hooks wrapping one. Nothing high severity, three worth documenting. All three `useInvalidateTileTokens` call sites named in the issue came back live and are unchanged.

## What this changes

No behaviour. Three comments, one test assertion documented, and nothing removed.

That is a different outcome from where this PR started, and the reason is worth recording, because it is the issue's own thesis biting the people auditing for it.

## The site that looks dead and is not

`useImportConfig` invalidates both `admin.aiStatus` and `maps.aiAvailability`. They are a mutually exclusive pair: `useAIAvailability` mounts one when `useAIStatusReader()` is true and the other when it is false, so at most one holds an entry per session and each covers an operator class the other misses.

The second line reads as dead. The argument goes: config-ops gates on `useSettingsAdmin()`, and `validate_permission_matrix` grants `manage_settings` to `admin` alone while refusing to let that role drop `manage_users`, so every config-ops admin is also a status reader and the key can never hold an entry.

**That argument is wrong, and it got the line deleted in an earlier revision of this PR.** It holds only for the stored role matrix, which is not the capability authority. `me_permissions` (`backend/app/modules/auth/router.py`) resolves every capability through `PermissionExtension.check_permission`, and that is what `usePermissions()` reads. An enterprise overlay can grant `manage_settings` and `use_ai_chat` without `manage_users`. That operator reaches config-ops with `canReadStatus` false, mounts `maps.aiAvailability`, and owns the entry the line invalidates.

Community deployments use `DefaultPermissionExtension`, which does read the stored matrix, which is exactly why the line looks dead by default.

The line is restored and the call site now carries the whole argument, including the version that is wrong, so the next reader does not re-derive the mistake. Credit to the codex P2 on this PR for catching the deletion.

## Two live cross-surface calls documented

`use-builder-save.ts` and `CollectionCreateDialog.tsx` both invalidate caches their own surface does not mount, and both are correct. "This surface does not mount the query" is not grounds to call a site inert: `invalidateQueries` marks a cached-but-unmounted entry stale and it refetches on next mount, which is how the Settings tabs and the login page pick up an import. The test is whether an entry can exist in that user's client at all.

## Related

`authConfig.oauthProviders` is invalidated here but missed by the sibling admin surfaces that mutate the same providers (`SettingsAuthTab`, `SamlProvidersSection`), which invalidate only `settingsOAuth.providers`. That is the inverse class, a cache needing invalidation with no invalidator, tracked as #1117. Not fixed here.

## Note for the next sweep

The lesson generalises beyond these three sites. A site's liveness is not readable from the file it lives in, and for anything permission-gated it is not readable from the stored role matrix either, because the extension seam can produce capability combinations the matrix refuses to store. Two reviewers reasoned from the matrix and both concluded wrongly.

## Verification

```
cd frontend
npm run lint          # clean
npm run typecheck     # clean
npm run test:coverage
npx vitest run src/hooks/__tests__/use-config-ops.test.ts   # 3 passed
```

Closes #1021

### DIFF
--- a/frontend/src/components/builder/hooks/use-builder-save.ts
+++ b/frontend/src/components/builder/hooks/use-builder-save.ts
@@ -81,6 +81,16 @@ function doCapture(map: MaplibreMap, mapId: string, queryClient: ReturnType<type
       // 400×250 thumbnail — unchanged behavior
       const thumb = cropResize(srcCanvas, 400, 250);
       uploadThumbnail(mapId, thumb.toDataURL('image/jpeg', 0.7)).then(() => {
+        // chore(#1021): this refetches nothing on the builder route, and that is
+        // expected rather than a bug. maps.all is ['maps'] while the builder mounts
+        // useMap, which is ['map', id], a different root string. The target is the
+        // MapsPage browse list (['maps', params]), whose MapCard renders
+        // thumbnail_url/thumbnail_updated_at: that entry is cached but unmounted
+        // while you are in the builder, so marking it stale here is what gets the new
+        // thumbnail on screen when the user navigates back. Do not "helpfully" add
+        // maps.detail. The builder reads thumbnail_url only as the hasThumbnail
+        // auto-capture gate, which shouldAutoCapture's module-level LRU has already
+        // settled by the time this upload resolves.
         queryClient.invalidateQueries({ queryKey: queryKeys.maps.all });
       }).catch(() => {
         // Silent failure for thumbnails
--- a/frontend/src/components/collections/CollectionCreateDialog.tsx
+++ b/frontend/src/components/collections/CollectionCreateDialog.tsx
@@ -48,6 +48,13 @@ export function CollectionCreateDialog({ open, onOpenChange }: CollectionCreateD
       // Invalidate + refetch all collection queries so the list updates
       // before the dialog closes. Use predicate to match any key starting
       // with 'collections' since prefix matching may not catch parameterized keys.
+      // chore(#1021): the sweep itself duplicates the identical predicate that
+      // useCreateCollection already runs in its own onSuccess, so the marking-stale
+      // half is redundant wherever this dialog is mounted. What this call adds is the
+      // `await`, which holds the dialog open until the refetch lands, and that only
+      // does work on CollectionsPage where a collections query is active. This dialog
+      // is also mounted globally from Navbar, where nothing is active and the await
+      // resolves immediately. Keep it for the CollectionsPage path.
       await qc.invalidateQueries({
         predicate: (query) =>
           Array.isArray(query.queryKey) && query.queryKey[0] === 'collections',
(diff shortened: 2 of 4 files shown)


ishiland:
@codex review

ishiland:
@codex review

PR #1123, opened 2026-08-02
fix(analysis): carry json and xml overlay attributes through intersect

### LINKED ISSUE #1099: intersect: carry json and xml overlay attributes instead of refusing the layer
`render_intersect_pairs` groups by `_src.gid, _mp._mask_gid` plus every carried overlay column. The source's columns need no grouping — `_src.gid` is a real table's primary key, so PostgreSQL licenses the rest by functional dependency — but `_mask_pieces` is a CTE with no key, so each overlay column it carries has to be named in the `GROUP BY`. `json` and `xml` have no equality operator, so grouping on one fails the CTAS with SQLSTATE 42883.

#1097 closed the sharp edge: the failure used to land at materialize, after the whole queue wait, quoting a generated alias the user never wrote, because the preview carries `gid` and `source_gid` only and so succeeded first. It is refused at enqueue now, with a message naming the column and its type. That is the same thing `_NON_GROUPABLE_TYPES` already does for dissolve's `by_field`.

Refusing is not the end state. Nested GeoJSON properties commonly land as `json`, so this makes intersect unavailable for a real class of layer rather than merely awkward.

## What the fix is

Group the aggregate by the two gids alone, and join the overlay's attributes back afterwards:

- drop the overlay columns from `_mask_pieces` and from the inner `GROUP BY`
- carry `_mask_gid` out of the inner aggregate (it is already grouped)
- in the outer `SELECT`, join the overlay table on `gid = _p._mask_gid` and take the attributes from there

The outer query does no aggregation, so no grouping applies and the type no longer matters. Row counts are unchanged: the inner aggregate already produces one row per (source gid, overlay gid) pair, and the join back is 1:1 against a table with a primary key.

Deliberately not done in #1097 — it changes the shape of a new and fairly intricate query, where the guard is small and provably correct. Worth doing on its own, with a materialize-level test that a `json` overlay column survives into the output dataset.

## Where

- `backend/app/platform/analysis_sql.py` — `render_intersect_pairs`
- `backend/app/modules/catalog/datasets/api/router_analysis.py` — `_validate_intersect_columns`, whose overlay branch comes out when this lands
- `backend/tests/test_analysis_intersect.py` — `test_an_ungroupable_overlay_column_is_rejected_at_enqueue` inverts

The sibling test `test_an_ungroupable_SOURCE_column_is_still_allowed` should stay: it pins that the source side never needed the guard.

ishiland:
## Problem

`render_intersect_pairs` grouped the aggregate by `_src.gid`, `_mp._gl_mask_gid`, and every carried overlay column. The source side needs no grouping, because `_src.gid` is a real table's primary key and PostgreSQL licenses the rest by functional dependency. But `_mask_pieces` is a CTE with no key, so each overlay column routed through it had to be named in the `GROUP BY`.

`json` and `xml` have no equality operator, so grouping on one fails with SQLSTATE 42883. #1097 moved that failure from the worker to enqueue, which stopped it landing after the whole queue wait quoting a generated alias the user never wrote. But refusing is not the end state: nested GeoJSON properties land as `json` routinely, so this took a real class of layer out of service as an overlay.

## Fix

Group by the two gids alone and join the overlay's attributes back afterwards. The aggregate carries `_gl_mask_gid` out; the outer query joins the overlay table on its primary key and takes the attributes from there. The outer query does no aggregation, so no type is off-limits.

The join is rendered only when there are overlay columns to fetch, so the preview path, which carries `gid` and `source_gid` only, emits the statement it always did.

**LEFT, not INNER**, which is a deviation from the issue's wording. Every `_gl_mask_gid` was read from that same table in the same statement, so a miss is impossible and INNER would be correct. But that makes the row-count guarantee rest on the snapshot argument being right, and a silent row *loss* is the worst failure available here. LEFT makes the guarantee structural instead.

## Row counts are unchanged, and that is pinned

The aggregate already emits one row per (source gid, overlay gid) pair, and `gid` is a serial primary key on every path that builds a feature table (`ogr.py:661` via `-lco FID=gid`, `parquet.py:450`, `layers/service.py:93`, `service_create.py:93`, plus `ADD PRIMARY KEY (gid)` on analysis outputs), so the join matches at most one row.

`test_the_join_back_does_not_multiply_rows` covers the case a 1:many join would actually expose: one overlay feature matched by several source features, so each overlay gid is looked up more than once. Two half-height bars across three unit-square zones, with a `json` column on the overlay so the count is measured on the shape the join exists for. Six pairs in, six rows out.

## The worker-side guard comes out too

`_reject_ungroupable_overlay_columns` in `processing/analysis/tasks.py` is the live-schema twin of the router branch this removes, and it is not in the issue's file list. Removing only the router half would have made intersect **worse than today**: enqueue accepts the layer, then the job dies minutes later after the queue wait, on a guard for a `GROUP BY` that no longer exists. That is exactly what #1097 was written to eliminate.

Nothing that still groups by a user-chosen column lost its recheck. `_ungroupable_type_name` stays, including its ARRAY/`udt_name` branch, and `_reject_ungroupable_by_field` is still called from the dissolve branch. Three tests of the removed function went with it; the `json[]` case they covered is independently pinned by `test_analysis_materialize.py::test_dissolve_by_a_json_array_field_fails_with_a_clear_message`, which drives it through `_materialize`.

## Both directions pinned

- `test_a_json_overlay_column_survives_into_the_output` drives real `_materialize` and asserts the columns arrive **as their types** (read from `information_schema.columns`) and that the values ride the right rows. Presence alone would pass under a join on the wrong key.
- `test_an_ungroupable_overlay_column_is_admitted_at_enqueue` is the inverted refusal test.
- `test_the_overlay_guards_still_refuse_what_actually_breaks` is new. Admitting `json` must not have admitted everything: it pins a duplicate name across layers, a name in the `_gl_` alias namespace, and dissolve's `by_field` still refusing `json`. That last one fails if someone removes the shared `NON_GROUPABLE_COLUMN_TYPES` machinery and calls it fixing this issue.
- `test_an_ungroupable_SOURCE_column_is_still_allowed` is kept. Its second half asserted the mirror image raises, which is the branch this removes, so that half now asserts the call passes.

## Module line caps

Both moved in this commit, measured exact: `analysis_sql.py` 1227 → 1255, `tasks.py` 1449 → 1413.

## Verification

```
cd backend
uv run pytest -n 4 --dist loadgroup        # 6827 passed, 83 skipped
uv run pytest tests/test_analysis_intersect.py               # 24 passed
uv run pytest tests/test_analysis_materialize.py tests/test_analysis_preview.py  # 156 passed
uv run pytest tests/test_layering.py -q    # 36 passed
uv run ruff check .                        # All checks passed
uv run ruff format --check .               # 875 files already formatted
```

The materialize test was checked against the old renderer rather than assumed to have teeth: restoring the previous `render_intersect_pairs` makes it fail at CTAS with `could not identify an equality operator for type json`, quoting the `GROUP BY` that carried the overlay columns.

Closes #1099

### DIFF
--- a/backend/app/modules/catalog/datasets/api/router_analysis.py
+++ b/backend/app/modules/catalog/datasets/api/router_analysis.py
@@ -290,36 +290,11 @@ def _validate_intersect_columns(source, overlay) -> None:
                 "choose a different layer."
             ),
         )
-    # fix(#1097 review): the OVERLAY's columns, and only the overlay's.
-    #
-    # render_intersect_pairs groups by `_src.gid, _mp._mask_gid` plus every
-    # carried overlay column. The source's columns need no grouping — gid is a
-    # real table's primary key, so PostgreSQL licenses them by functional
-    # dependency — but `_mask_pieces` is a CTE with no key, so each overlay
-    # column it carries has to be named in the GROUP BY explicitly. json and
-    # xml have no equality operator, so grouping on one fails the CTAS with
-    # SQLSTATE 42883, and it fails there rather than in the preview: the
-    # preview carries gid and source_gid only, so it succeeds and the operation
-    # dies after the queue wait with an error naming a generated alias the user
-    # never wrote.
-    #
-    # Refused at enqueue instead, which is what NON_GROUPABLE_COLUMN_TYPES already
-    # does for dissolve's by_field. Refusing rather than silently dropping the
-    # column is the same choice the sibling guards above make: an overlay whose
-    # output columns depend on which of them happened to be groupable is worse
-    # for anyone scripting against the result than one that says no.
-    #
-    # Carrying these through an overlay is a real gap and is worth doing — it
-    # needs the aggregate to group by the two gids alone and join the overlay's
-    # attributes back afterwards, which is a change to this query's shape
-    # rather than a guard. Tracked separately.
     # fix(#1097 review): a carried column may not sit in the alias namespace.
-    # Both layers, because an overlay carries columns from both and the inner
-    # aggregate names them in one select list alongside _gl_src_type; the
-    # overlay's are additionally re-listed inside _mask_pieces beside
-    # _gl_mask_gid and _gl_g. Checked against the PREFIX rather than the three
-    # alias names, so an alias added to that query later is covered without a
-    # second place to update.
+    # Both layers, because an overlay carries columns from both and they share
+    # the statement with _gl_src_type and _gl_mask_gid. Checked against the
+    # PREFIX rather than the alias names, so an alias added to that query later
+    # is covered without a second place to update.
     reserved = sorted(
         name
         for name in (_column_names(source) | _column_names(overlay))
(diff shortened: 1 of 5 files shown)


ishiland:
@codex review

Issue #1112, opened 2026-08-01
fix(maps): dataset_extent_bbox is span-flattened before the #903 guard can read it


ishiland:
Follow-up to #903, which is correctly closed: the frontend guards landed and work. This is the one path where they never get the chance to run, because the backend flattens the bbox before the frontend sees it.

## The chain

1. `backend/app/modules/catalog/maps/_router_helpers.py:117` builds a map layer's `dataset_extent_bbox` with `extent_to_span_bbox(meta.get("extent"))`. `extent_to_span_bbox` widens a seam-crossing extent to the monotonic `[-180, s, 180, n]` — honest but world-wide.
2. `frontend/src/components/builder/BuilderMap.tsx:97` `getVisibleLayerBounds()` reads exactly that field, and carries a `fix(#903)` guard that unwraps a `west > east` pair past 180 so a crossing layer fits the few degrees it occupies.
3. The guard therefore never fires on this path: by the time the frontend reads the value it is a well-formed world bbox, not a crossing pair. There is nothing left to detect.

Net effect: a seam-crossing dataset (Fiji-shaped) added to a map auto-fits and "Zoom to Layer"s to the **whole world** instead of to the data. No error, no visual cue that anything was widened — it just looks like a dataset with a global extent.

This is a behavioural gap, not a correctness bug: the widened box does contain the data, which is why #903's acceptance work did not catch it. #1040 and #1060 flipped the dataset and collection extent endpoints to RFC 7946 spec form for exactly this reason; this producer was not part of that change.

## Suggested fix

Serve the spec form at `_router_helpers.py:117` (`extent_to_bbox`), because its only consumer is now seam-aware. Worth confirming before writing it:

- Whether any other client reads `dataset_extent_bbox` and assumes `west <= east` — it is on the map-layer response, so the SDKs and the MCP server surface it too. If a non-seam-aware consumer exists, the alternative is a second field rather than changing this one.
- `backend/app/processing/tiles/router.py:1278` deliberately keeps the span form for TileJSON `bounds`, and the three DCAT services keep it for external harvesters. Those are correct as-is and should not be swept along — an inverted pair in a MapLibre source `bounds` matches no tiles at all.

## Verification when it is done

Ingest a pre-split crossing MultiPolygon, add it to a map, and confirm auto-fit and Zoom to Layer centre on the data rather than the world. A unit test on the producer is not sufficient on its own — the defect is that a plausible value reaches a guard that then has nothing to do.

PR #1124, opened 2026-08-02
fix(builder): measure small-DEM viewport coverage across the antimeridian

### LINKED ISSUE #1122: builder: small-DEM coverage check clips at ±180, warns on a DEM that fills the screen
## Problem

`demViewportCoverage` in `frontend/src/components/builder/terrain-coverage.ts` measures DEM coverage as `(DEM bounds ∩ viewport) / viewport` on a planar longitude axis:

```ts
ix = Math.max(0, Math.min(ve, demBounds[2]) - Math.max(vw, demBounds[0]));
```

`map.getBounds()` is monotonic, because MapLibre takes min/max over four unwrapped corner longitudes. But it runs **past ±180** when the viewport straddles the antimeridian, returning something like `[179.5, …, 182, …]`.

The DEM's own bounds do not run past 180. So `Math.min(ve, demEast)` clips the DEM at +180 while the viewport keeps its 182, and the intersection comes out a fraction of the truth. A DEM that covers the entire screen scores about a fifth of its real coverage.

## Impact

The `#186 (b)` "this DEM only covers a small slice of the viewport" toast fires for a DEM that visibly fills the screen, whenever the user has the builder panned across the antimeridian. Cosmetic, but it tells the user the opposite of what they can see, which is worse than saying nothing.

The reverse case exists too: a genuinely small DEM near the seam can score higher than it should.

## Where

- `frontend/src/components/builder/terrain-coverage.ts` — `demViewportCoverage`, and `isFiniteBounds4`, which rejects `west > east` outright as malformed.
- Call sites: `BuilderMap.tsx` (`maybeWarnSmallDemCoverage`), which reads `RasterTileToken.bounds`.

## Not related to #1112

Worth stating, because the two look adjacent and are not. This path reads the **tile token's** `bounds`, which `backend/app/processing/tiles/router.py:1278` builds with `extent_to_span_bbox` and documents as deliberately the monotonic span form at `:397`. It does not read the map layer's `dataset_extent_bbox`, so #1112's flip of that field to RFC 7946 spec form does not reach here. This bug predates #1112 and would survive it.

Found while tracing #1112's consumers.

## Suggested fix

Measure the longitude axis on the circle rather than the plane. Turn each rectangle into an unwrapped interval (a crossing pair unwraps past 180) and score the DEM against each turn of the globe it can occupy, summing the overlaps rather than taking the largest, since a viewport straddling the seam is genuinely covered by two copies of a world-spanning DEM at once. Latitude needs none of this.

`isFiniteBounds4` should also stop treating `west > east` as malformed. That is the RFC 7946 §5.2 encoding of a crossing box. Only a zero-width longitude span is degenerate.

Both directions want pinning: a crossing DEM that genuinely covers the viewport must not warn, and one that genuinely does not cover it must still warn.

ishiland:
## Problem

`demViewportCoverage` compared two longitude ranges with planar min/max:

```ts
ix = Math.max(0, Math.min(ve, demBounds[2]) - Math.max(vw, demBounds[0]));
```

`map.getBounds()` is monotonic, because MapLibre takes min/max over four unwrapped corner longitudes. But it runs **past ±180** when the viewport straddles the antimeridian, returning something like `[179.5, …, 182, …]`. The DEM's own bounds do not, so `Math.min(ve, demEast)` clipped the DEM at +180 while the viewport kept its 182.

A DEM covering the entire screen scored 0.2 of that viewport and fired the `#186 (b)` "this DEM only covers a small slice of the viewport" toast. The message told the user the opposite of what they could see.

`isFiniteBounds4` also rejected `west > east` outright as malformed, so a spec-form crossing bbox produced no signal at all rather than a wrong one.

## Fix

Measure longitude on the circle, and treat the DEM as what it is: a pattern repeating every 360 degrees.

`coveredLength(t, width)` measures how much DEM lies between its own west edge and `t` degrees east of it. Whole turns contribute a full `width` each, the partial turn contributes however far into it the DEM reaches. The overlap with a viewport is that quantity measured at the viewport's two edges and subtracted.

Latitude is untouched and needs none of this.

`isFiniteBounds4` now accepts `west > east`. That is the RFC 7946 §5.2 encoding of a crossing box, not a malformed one. Only a zero-width longitude span is degenerate.

## Why not enumerate turns

The first version of this fix scored the DEM against a fixed `[-360, 0, 360]` shift list and summed the overlaps. Codex caught that on review, and it was the same class of bug this function exists to fix, one level out.

`renderWorldCopies` is on by default, so `getBounds()` runs further out the more the user pans, without bound. A Fiji DEM can legitimately be viewed at `[899.5, 902]`, two whole turns east. A fixed list answers "how many turns?" with a guess, and one pan past the last entry reported zero coverage and warned about a DEM filling the screen.

The closed form has no shift list to outgrow. It is exact at any pan distance and costs the same at two turns as at two thousand. It is also continuous across turn boundaries by construction — approaching from below gives `q * width + width`, from above `(q + 1) * width + 0` — so float noise near a boundary cannot jump the result by a period.

## Not related to #1112

Worth stating, because the two look adjacent. This path reads the **tile token's** bounds, which `processing/tiles/router.py:1278` builds with `extent_to_span_bbox` and documents at `:397` as deliberately the monotonic span form. It does not read the map layer's `dataset_extent_bbox`; `BuilderMap.tsx` resolves the token from `useTileTokens()`, not from `rasterTokenFromLayer`. So #1112's flip of that field does not reach here, and this bug predates it.

## Verification

```
cd frontend
npm run lint       # clean
npm run typecheck  # clean
npx vitest run src/components/builder/__tests__/terrain-coverage.test.ts   # 26 passed
npm run test:coverage   # 369 files, 4532 tests passed
```

Both the original fix and the turn-invariance fix were checked against the code they replace, rather than assumed to have teeth.

Restoring the pre-fix module fails 5 tests, in both directions: `expected 0.2 to be 1` (the reported bug), `expected null to be close to 0.8` (the old `isFiniteBounds4` rejecting a crossing pair), `expected true to be false` (the spurious warning), and `expected false to be true` — a genuinely small DEM near the seam that **stops** warning.

Restoring the three-shift version fails 4, including `is turn-invariant for an ordinary DEM at any pan distance`, which pins the general property rather than the single far-pan case that prompted it.

Closes #1122

### DIFF
--- a/frontend/src/components/builder/__tests__/terrain-coverage.test.ts
+++ b/frontend/src/components/builder/__tests__/terrain-coverage.test.ts
@@ -62,6 +62,112 @@ describe('demViewportCoverage', () => {
   });
 });
 
+// fix(#1112 review): the antimeridian cases. `demBounds` here is the TILE
+// TOKEN's bounds (span form, from processing/tiles/router.py), not the map
+// layer's `dataset_extent_bbox`, so both forms are exercised: the span form is
+// what ships today, and the spec form is what a crossing extent looks like if
+// this input is ever sourced seam-aware.
+//
+// The rectangle that actually crosses is the VIEWPORT: MapLibre's getBounds()
+// takes min/max over unwrapped corner longitudes, so a viewport straddling the
+// seam reads e.g. [179.5, 182] rather than inverting.
+describe('demViewportCoverage across the antimeridian', () => {
+  // Fiji-shaped DEM: ~3 degrees wide, sitting on the seam.
+  const FIJI_SPEC = [178.5, -20, -178.5, -15];
+  const FIJI_SPAN = [-180, -20, 180, -15];
+  // Zoomed in over Fiji, straddling the seam. 2.5 x 2 degrees.
+  const SEAM_VIEW = [179.5, -18, 182, -16];
+  // Zoomed way out. 120 x 120 degrees.
+  const WIDE_VIEW = [120, -60, 240, 60];
+
+  it('reports a world-spanning DEM as fully covering a seam-straddling viewport', () => {
+    // The regression this fixes: clipping the DEM at +180 scored this 0.2 and
+    // warned about a "small DEM" that covers the entire screen.
+    expect(demViewportCoverage(FIJI_SPAN, SEAM_VIEW)).toBe(1);
+  });
+
+  it('scores a crossing DEM against the part of the viewport it really covers', () => {
+    // [178.5, 181.5] over a [179.5, 182] viewport → 2 of 2.5 degrees wide,
+    // full height → 0.8. The old code returned null (west > east was rejected).
+    expect(demViewportCoverage(FIJI_SPEC, SEAM_VIEW)).toBeCloseTo(0.8, 6);
+  });
+
+  it('still reports a crossing DEM as a sliver of a zoomed-out viewport', () => {
+    // 3 degrees wide, 5 of 120 degrees tall, in a 120x120 viewport.
+    expect(demViewportCoverage(FIJI_SPEC, WIDE_VIEW)).toBeCloseTo((3 * 5) / (120 * 120), 6);
+  });
+
+  it('does not credit a DEM on the far side of the globe', () => {
+    // Central France against a Fiji viewport: no turn of the globe overlaps.
+    expect(demViewportCoverage([1.5, -18, 2.5, -16], SEAM_VIEW)).toBe(0);
+  });
+
+  // fix(#1124 codex P2): `renderWorldCopies` is on by default, so getBounds()
+  // keeps running further out the more the user pans east or west. There is no
+  // bound on how far, which is why the scoring cannot enumerate a fixed set of
+  // turns. FAR_VIEW is SEAM_VIEW moved two whole turns east (+720) and must
+  // score identically; DEEP_VIEW is ten turns west (-3600) and so must the
+  // rest of the axis.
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
Confirmed and correct, and it is being tracked separately as #1128 rather than fixed here.

The numbers check out. For a Fiji footprint in a viewport `[179.5, -20, 190, -15]`: the pre-#1122 planar math returned 4.8%, this PR returns 100%, and the true answer is 19%. So this PR does change that case from warning to silent.

The reason it is not fixable here is that the correct answer is not derivable from the input. `processing/tiles/router.py:1278` builds the tile token's bounds with `extent_to_span_bbox`, so a crossing raster and a genuinely global one arrive as the same `[-180, s, 180, n]` value. No arithmetic on that can separate them, and the previous 4.8% was a coincidentally-warning answer produced by the clipping bug this PR removes.

The fix belongs one layer up, at the call site rather than in the coverage math: `dataset_extent_bbox` on the map-layer response became the RFC 7946 spec form in #1112, which landed today, so the real footprint is now available to `BuilderMap` where the token's is not. That is a change of data source and wants its own review, so it is #1128.

This PR stays scoped to the arithmetic, where it fixes a warning that fires on a DEM visibly filling the screen. The residual gap in the other direction is a missing hint rather than a false statement, and the limitation is now documented at the branch that causes it.

ishiland:
@codex review

PR #1125, opened 2026-08-02
fix(maps): serve dataset_extent_bbox in the RFC 7946 spec form

### LINKED ISSUE #1112: fix(maps): dataset_extent_bbox is span-flattened before the #903 guard can read it

ishiland:
## Problem

`_router_helpers.py` built a map layer's `dataset_extent_bbox` with `extent_to_span_bbox`, which widens a seam-crossing extent to the monotonic `[-180, s, 180, n]`. Honest, but world-wide.

The two builder fit paths read exactly that field and both carry a `fix(#903)` guard that unwraps a `west > east` pair past 180 so a crossing layer fits the few degrees it occupies. Those guards can never fire on this path: by the time the frontend sees the value it is a well-formed world bbox, with nothing left to detect.

Net effect: a seam-crossing dataset added to a map auto-fits and "Zoom to Layer"s to the whole world. No error and no visual cue, because a flattened Fiji extent is bit-identical to a genuinely global one. #1040 and #1060 flipped the dataset and collection endpoints to spec form for this reason; this producer was not part of that change.

## Fix

Serve the spec form. Each consumer now declares the form it wants, rather than the producer flattening for the most restrictive one.

## Consumer audit

The issue asks whether any consumer assumes `west <= east`, because one would mean a second field rather than changing this one. Every reference, excluding tests:

| Consumer | Verdict |
|---|---|
| `maps/_router_helpers.py` | the producer, changed here |
| `maps/schemas.py` | type only, `list[float] \| None`, unchanged |
| `BuilderMap.getVisibleLayerBounds` | seam-aware, unwraps past 180 (`fix(#903)`) |
| `use-builder-layers.handleZoomToLayer` | seam-aware, `bbox[0] > bbox[2]` is explicitly not a rejection (`fix(#903)`) |
| `map-sync.toSyncInput` | passes through; every downstream MapLibre source normalizes |
| `use-builder-editor-scene` | writes `null`, never reads |
| `frontend/src/types/api.ts`, `api.generated.ts` | type only, unchanged |
| Python SDK `map_layer_response.py` | (de)serialization only, no arithmetic |
| TypeScript SDK `types.gen.ts` | type only |
| MCP server, CLI | no references |

No consumer assumes a monotonic pair, so no second field is needed.

## Why map-sync is not wrapped

The obvious-looking edit is to span the value back at `map-sync.ts` before it reaches a MapLibre source, where an inverted pair matches no tiles. That would be both redundant and harmful.

Redundant, because every path from `SyncLayerInput.bounds` to a source already normalizes: `ensureRasterDemTerrainSource` and the vector source in `map-sync.ts`, and the raster and hillshade adapters. `normalizeRasterBounds` in `layer-adapters/shared.ts` is the keystone, and its docstring already names those four as inheriting the guard from it.

Harmful, because `map-sync` output feeds the fit paths as well. Converting there would discard the seam before those guards could use it, re-breaking the exact defect this PR fixes. The conversion belongs at the MapLibre boundary, which is the one place that cannot express a crossing.

## API impact

No OpenAPI or SDK regeneration: the type is unchanged, `list[float] | None` either way.

**The VALUE changes for SDK and MCP consumers.** A seam-crossing layer now returns `west > east` where it previously returned `[-180, s, 180, n]`. Anything consuming this field arithmetically outside this repo should expect the RFC 7946 §5.2 form.

The span form is deliberately kept where it is still correct: TileJSON `bounds` in `processing/tiles/router.py` and the three DCAT services for external harvesters. Neither is swept along.

## Verification

```
cd backend
uv run pytest tests/test_antimeridian_extent.py -q   # 45 passed
uv run pytest tests/test_layering.py -q              # 36 passed
uv run ruff check . && uv run ruff format --check .  # clean, 875 files

cd frontend
npm run lint && npm run typecheck                    # clean
npm run test:coverage                                # 370 files, 4525 tests passed
```

`test_map_layer_response_bbox_stays_monotonic` is inverted, and the "do NOT flip the sibling" comment in `datasets/domain/helpers.py` is rewritten, since it documented the opposite of what is now true. New `map-sync.seam-bounds.test.ts` pins the crossing bbox travelling unconverted through `toSyncInput` and being spanned by the time it reaches the vector source, plus the non-crossing case being left alone.

Closes #1112

### DIFF
--- a/backend/app/modules/catalog/datasets/domain/helpers.py
+++ b/backend/app/modules/catalog/datasets/domain/helpers.py
@@ -183,10 +183,13 @@ def dataset_to_response(
         # extent to [-180, s, 180, n], which is bit-identical to a genuinely
         # global dataset, so isLargeExtent fired first and no client-side test
         # could tell the two apart. The seam information has to survive the
-        # wire. Do NOT flip the sibling `dataset_extent_bbox` in
-        # maps/_router_helpers.py: it bounds a vector tile source, where an
-        # inverted pair yields no tiles at all. One column, two contracts —
-        # both pinned in tests/test_antimeridian_extent.py.
+        # wire. fix(#1112): the sibling `dataset_extent_bbox` in
+        # maps/_router_helpers.py now serves the same form for the same reason.
+        # It reaches a MapLibre source `bounds`, where an inverted pair yields
+        # no tiles at all, but the span conversion belongs at that boundary
+        # (normalizeRasterBounds in layer-adapters/shared.ts) rather than on the
+        # wire, so the builder's fit paths still get the seam. One column, one
+        # contract — pinned in tests/test_antimeridian_extent.py.
         extent_bbox=extent_to_bbox(record.spatial_extent),
         column_info=dataset.column_info,
         quality_detail=dataset.quality_detail,
--- a/backend/app/modules/catalog/maps/_router_helpers.py
+++ b/backend/app/modules/catalog/maps/_router_helpers.py
@@ -12,7 +12,7 @@ import structlog
 from fastapi import HTTPException, status
 from sqlalchemy.ext.asyncio import AsyncSession
 
-from app.core.geo import extent_to_span_bbox
+from app.core.geo import extent_to_bbox
 from app.core.identity import Identity
 from app.modules.catalog.authorization import get_user_roles
 from app.modules.catalog.maps.models import Map, MapLayer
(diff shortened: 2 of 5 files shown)


ishiland:
@codex review

PR #1126, opened 2026-08-02
feat(extensions): let the permission seam answer who a record's audience is

### LINKED ISSUE #1068: feat(extensions): let the permission seam answer "who can see this?" for audience checks

ishiland:
## Problem

`PermissionExtension` exposed two methods and both take a concrete user: `filter_visible` and `can_access_dataset`. Neither can answer "who is the audience for this map?", because an audience is a set with no representative member to pass in. A guard needing to reason about a whole audience therefore had to reconstruct it from the community ladder directly, which an overlay is free to change.

#931 / #1056 added exactly such a guard, refusing a dataset visibility change that would strand a shared map's audience.

## Why this is more urgent than the issue says

The issue argues the failure direction is safe, on the grounds that an overlay can only narrow a rung, which makes the guard more conservative. **That is out of date.** #1111 landed afterwards and documents the opposite for a widening overlay: the guard returns `[]`, the change applies, and a granted viewer silently loses the layer. Both directions are wrong, one loudly and one silently, and the silent one is why #1111 fell back to a stopgap rather than trusting the argument.

That stopgap refused the operation whenever `type(get_permission_extension()) is not DefaultPermissionExtension`. In practice **every overlay deployment got blanket refusals** on visibility changes touching shared maps.

## Fix

`PermissionExtension.record_audience` answers the set-shaped question: which principals can read this record, at a named visibility, without inventing a user. `EXTENSION_API_VERSION` goes 3 → 4, since this is a required method on an existing Protocol.

The community implementation is the same ladder `filter_visible` walks, read from the user end: admins always, `public`/`internal` any signed-in account, `private` the owner, `restricted` the owner plus grant-holding roles, with the status gate ANDed across every rung.

## The refusal narrows rather than disappearing

An authority that cannot answer still gets a conservative refusal, but the test is now what the overlay actually overrides, not its type:

- No `record_audience` at all. Only an overlay declaring no `EXTENSION_API_VERSION` reaches this, because a declared-but-mismatched version is already a hard boot failure.
- It inherits the community audience while overriding `filter_visible` or `can_access_dataset`. Then the audience it reports and the rows its viewers get are two different policies, and core cannot tell a missing answer from a wrong one.

`check_permission` is deliberately not in that set: capabilities do not decide which records a viewer can read.

This is a real widening over #1111. A subclass that changes nothing about reads, or one that implements `record_audience`, now gets a real answer instead of a refusal.

## Compatibility

Behaviour is unchanged for community deployments, and that is tested rather than asserted. `test_permission_audience.py` computes the reader set through `filter_visible` and compares it account by account against `record_audience`, across every visibility, record status and role, plus the anonymous flag against `filter_visible(user=None)`. A change to one direction that is not mirrored in the other fails there rather than quietly making the shared-map guard disagree with what viewers actually see.

Overlay authors must implement `record_audience` when replacing the `permission` slot, and must implement it whenever they change either read method.

## Verification

```
cd backend
uv run pytest -n 4 --dist loadgroup        # 6846 passed, 83 skipped
uv run pytest tests/test_layering.py -q            # 36 passed
uv run pytest tests/test_rule1_structural.py -q    # 6 passed
uv run pytest tests/test_rule2_structural.py -q    # 76 passed
uv run ruff check . && uv run ruff format --check .  # clean, 876 files
```

Closes #1068

### DIFF
--- a/backend/app/modules/catalog/datasets/domain/service_metadata.py
+++ b/backend/app/modules/catalog/datasets/domain/service_metadata.py
@@ -178,6 +178,7 @@ async def _apply_visibility_change(
         old_visibility=record.visibility,
         new_visibility=new_visibility,
         record_status=record.record_status,
+        record_id=record.id,
         owner_id=record.created_by,
     )
     if broken_maps:
--- a/backend/app/modules/catalog/maps/service_public.py
+++ b/backend/app/modules/catalog/maps/service_public.py
@@ -225,6 +225,19 @@ async def find_maps_broken_by_dataset_visibility(
 
     Empty = safe to apply.
     """
+    # fix(#1126 codex P2): a change that changes nothing strands nobody, and
+    # that is true of every authority, which is why it is settled before asking
+    # one. `_apply_visibility_change` runs for any non-null `visibility` in the
+    # body without comparing it to the stored value, so a client that round
+    # trips the whole record resubmits the current one — and both branches
+    # below got that wrong. The fallback named every shared map using the
+    # dataset. The seam-answering branch was subtler: `before` and `after` are
+    # the same predicate for a no-op, and NULL passes `IS NOT false` and
+    # `IS NOT true` alike, so an overlay that cannot classify an account
+    # reported it as stranded by a move that did not happen. Until #1068 the
+    # rank comparison absorbed both, since `X < X` is false.
+    if old_visibility == new_visibility:
+        return []
     permission = get_permission_extension()
     if _answers_audience_questions(permission):
         query = RecordAudienceQuery(
--- a/backend/app/modules/catalog/maps/service_public.py
+++ b/backend/app/modules/catalog/maps/service_public.py
@@ -3,6 +3,7 @@
 import hashlib
 import secrets
 import uuid
+from dataclasses import replace
 from datetime import datetime, timezone
 from typing import TYPE_CHECKING
 
@@ -29,6 +30,8 @@ from app.modules.embed_tokens.models import EmbedToken
 from app.modules.embed_tokens.service import resolve_embed_scope_for_map
 from app.platform.extensions import (
     DefaultPermissionExtension,
+    RecordAudience,
+    RecordAudienceQuery,
     get_catalog_port,
     get_permission_extension,
 )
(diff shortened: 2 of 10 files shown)


ishiland:
@codex review

ishiland:
@codex review

PR #1127, opened 2026-08-02
test(conftest): restore process-global logging config around every test

### LINKED ISSUE #1066: test: add an autouse conftest guard for global logging config
## Summary

`backend/tests/conftest.py` has seven autouse fixtures that snapshot and restore process-global state. None covers logging config. That seam has now produced three separate defects, and it is the one seam where a leak can make a **security assertion pass vacuously**.

Not urgent: after #1065 there are **zero live unguarded sites**. This is prophylactic, and it is filed so the decision is not lost rather than to schedule work.

## The three recurrences

- `test_tile_signing.py:648-656` — documents that neither `caplog` nor `capture_logs()` reliably sees the tile logger, and works around it per-call-site.
- #1024's EXISTS negative controls — went red on main, fixed in #1063.
- #1064 / #1065 — three SMTP/webhook credential-leak assertions that had never been able to fail. Demonstrated by injecting a real password leak: the old assertion reported `CAPLOG_RECORDS: 0` and **passed**.

The third is the one that matters. A disconnected tripwire and a working one are indistinguishable from the outside, so the cost is unbounded and silent.

## Why a conftest guard rather than per-file fixes

The repo has already made this exact call once. `_restore_process_global_config` (`conftest.py:325`) exists because only one module restored `GEOLENS_TENANCY_MODE` while nine others leaked it, and its docstring records the conclusion: restoring after every test is "the deterministic ROOT fix the per-module xdist grouping could not provide."

Logging is the same shape. `setup_logging()` sets `cache_logger_on_first_use=True` and replaces the root handlers; measured `True` before / `False` after around the one test that now restores it.

## Design constraint, worth ten minutes when someone does this

`setup_logging()` has to run inside the test body, not in fixture setup. pytest swaps `sys.stdout`/`stderr` for the call phase only, so a handler created during setup binds to a stream that is not the captured one and every write fails with `--- Logging error ---`. An autouse snapshot/restore must therefore not fight tests that legitimately configure logging mid-test. This is why the guard is not a copy-paste of the settings one.

Restoring to structlog's **library default** is preferable to faithfully restoring the prior value: the default leaves caching `False`, whereas restoring a prior `True` hands the next test the exact global that makes captures go blind.

## Suggested trigger

Do this when something next touches the logging seam, rather than as standalone work. Roughly ten lines modelled on `_restore_process_global_config`.

## Also noted, no action

- `_reload_settings_multi()` mutates env before its `try:` (`test_dp04_data_plane_structural_gate.py:76`, same shape in `test_embed_tokens.py` and `test_tenant_rls_migration.py`), so a raising `Settings()` skips the restore. The existing autouse guard catches it.
- `structlog.contextvars.bind_contextvars(service="api")` at `conftest.py:1655` is never cleared, adding a field to every later record in the worker. Benign, though `test_ingest_ogr_pure.py` calls `clear_contextvars()` five times, which suggests someone hit it.

ishiland:
## Problem

`backend/tests/conftest.py` has seven autouse fixtures that snapshot and restore process-global state. None covered logging, and that seam has produced three separate defects. It is also the one seam where a leak can make a **security assertion pass vacuously**.

- `test_tile_signing.py` documents that neither `caplog` nor `capture_logs()` reliably sees the tile logger, and works around it per call site.
- #1024's EXISTS negative controls went red on main, fixed in #1063.
- #1064 / #1065: three SMTP and webhook credential-leak assertions that had never been able to fail. Injecting a real password leak made the old assertion report `CAPLOG_RECORDS: 0` and pass.

A disconnected tripwire and a working one are indistinguishable from outside.

## What actually causes the blindness

Worth stating precisely, because the existing notes in the tree describe an incomplete model and the fix follows from the real one. Measured on structlog 25.5.0:

| proxy | processor list | records captured |
|---|---|---|
| frozen | same object | 1 |
| frozen | rebound to a copy | **0** |
| unfrozen | rebound to a copy | 1 |

`capture_logs()` clears and extends the live list *in place*, explicitly "to not break references held by bound loggers". So a frozen `BoundLoggerLazyProxy` alone does **not** blind it. Blindness needs a freeze **and** a processor-list identity change — and `setup_logging()` builds a fresh list on every call, so a second call after a freeze supplies the missing half.

Keeping the proxy unfrozen makes the whole conjunction unreachable.

## Fix

**Primary: `structlog.configure` is clamped for the test session** so `cache_logger_on_first_use` can never arm. Installed at conftest module level, in force before test modules import and therefore before `app/api/main.py`'s import-time `setup_logging()`. Clamping `configure` rather than wrapping `setup_logging` keys the invariant off the resolved state at the single choke point instead of chasing call paths; clamping in conftest keeps `backend/app/` test-unaware.

**The baseline is rebased off `sys.stdout`, once, at conftest import.** Unfreezing the proxies exposed a second latent state: from the second test onward, each test's arrival config was structlog's *unconfigured default* — `PrintLoggerFactory` + `ConsoleRenderer` — measured by dumping the config around every test. `PrintLogger` resolves `sys.stdout` at logger construction, which for an uncached proxy is every call. The freeze had been hiding this by pinning emissions to wherever the first one pointed. With it gone, an app emission during a `CliRunner.invoke()` rendered a console line into the runner's captured stdout, and `test_cli_round_trip.py` failed parsing CLI JSON with `Extra data: char 4` (the timestamp's `2026` parses; the rest does not). "Application logs never reach stdout" is an absolute the whole session needs, like the caching flag, rather than something every stdout-parsing module defends for itself — so the guard's floor is stdlib routing, which is also exactly what `setup_logging()` installs.

**Secondary: an autouse fixture restores everything else `setup_logging()` moves**, *relative* to the snapshot the test arrived with, because the app legitimately owns that state. A blanket `structlog.reset_defaults()` is wrong for the same reason — it drops `_redact_sensitive_fields` and the stdlib routing.

**Handlers are restored by identifying pytest's own** (handler class's root module is `_pytest`), not by saved-handler presence — pytest reattaches its capture handlers for the teardown phase before the finalizer runs, so a leaked `StreamHandler` can ride along while every saved handler is present.

**Nothing is configured during setup, only after `yield`** — pytest swaps `sys.stdout`/`stderr` for the call phase only.

## Which part carries which fix

- The clamp closes the freeze class: disabling it fails `test_mid_test_setup_logging_does_not_blind_capture_logs` with `capture_logs() went blind` / `assert [] == ['sentinel']`.
- The stdout-safe baseline carries the round-trip fix: reverting it alone re-fails `test_cli_round_trip.py` (2 failed), restoring it passes (8).
- The fixture's relative restore covers root level and `uvicorn.access.propagate` surviving into later tests.

None of the three is dead code, and each has the test that fails without it.

## Verification

```
cd backend
uv run pytest -n 4 --dist loadgroup                      # 6844 passed, 84 skipped
uv run pytest tests/test_logging_conftest_guard.py -q    # 9 passed
uv run pytest tests/test_cli_round_trip.py -q            # 8 passed, 2 skipped
uv run pytest tests/test_layering.py -q                  # 36 passed
uv run ruff check . && uv run ruff format --check .      # clean, 876 files
```

The round-trip file needs the CLI deps installed into the backend venv to run at all on the host (it module-skips otherwise), which is why the regression only surfaced in the CLI Tests CI job. Ordering was varied explicitly (no randomizer plugin is installed in this repo, so `-p no:randomly` disables nothing): the logging-adjacent files pass in forward and reversed order, and the round-trip file passes in both positions relative to the guard tests.

Closes #1066

### DIFF
--- a/backend/tests/conftest.py
+++ b/backend/tests/conftest.py
@@ -1,10 +1,11 @@
 import asyncio
+import logging
 import os
 import time
 import uuid
 import tempfile
 import warnings
-from contextlib import asynccontextmanager
+from contextlib import asynccontextmanager, contextmanager
 
 import asyncpg.exceptions
 import pytest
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1129, opened 2026-08-02
fix(builder): measure DEM coverage from the layer extent, not the token span

### LINKED ISSUE #1128: builder: DEM coverage cannot tell a seam-expanded token bbox from a globe-spanning DEM
## Problem

`maybeWarnSmallDemCoverage` in `BuilderMap.tsx` measures DEM coverage from `RasterTileToken.bounds`. For an antimeridian-crossing raster, `backend/app/processing/tiles/router.py:1278` deliberately builds that with `extent_to_span_bbox`, so a Fiji-shaped DEM arrives as `[-180, s, 180, n]` — byte-identical to a genuinely global DEM.

The coverage check therefore cannot tell the two apart, and no arithmetic on that value can. For a Fiji footprint (~2° wide) in a viewport `[179.5, -20, 190, -15]`:

| | longitudinal overlap | coverage | warns? |
|---|---|---|---|
| before #1122 (planar, clipped at +180) | 0.5 | 4.8% | yes, for the wrong reason |
| after #1122 (circle-aware) | full | 100% | no |
| truth | 2.0 | 19% | should warn |

So the warning is a false negative for crossing DEMs. It was previously a *coincidentally* right answer reached by clipping, which #1122 correctly removed as the cause of the opposite bug (a globe-spanning DEM warning while filling the screen).

Found by codex review on PR #1124.

## Why it is filed rather than fixed there

#1122 is an arithmetic fix to `terrain-coverage.ts`. This is a data-availability problem one layer up: the module is given a value that has already lost the footprint. Fixing it inside the coverage math is not possible, and changing where the value comes from is a design change that wants its own review.

## The option that just became available

`dataset_extent_bbox` on the map-layer response is now the RFC 7946 spec form, as of #1112, so a crossing extent arrives as a `west > east` pair with the real footprint intact. `BuilderMap` already has the DEM layer in hand at the call site, so `maybeWarnSmallDemCoverage` could take the layer's extent instead of the token's.

Worth confirming before building:

- Whether the DEM layer's `dataset_extent_bbox` is reliably populated for raster and VRT records, or only for vector datasets.
- Whether the terrain resolver's DEM layer is always a `MapLayerResponse` carrying that field, on both the builder and viewer paths.
- `terrain-coverage.ts` already accepts a crossing pair since #1122, so the consumer needs no change — only the caller.

The alternative is a second, honest field on the tile token. That is a wider contract change and should only be considered if the layer route does not work.

## Not a regression to revert

#1122 is right to land as-is. It fixes a warning that fires on a DEM visibly filling the screen, which tells the user the opposite of what they can see. This issue is the residual false negative in the other direction, which is a missing hint rather than a false statement.

ishiland:
## Problem

The small-DEM viewport warning measured coverage from `RasterTileToken.bounds`. The tile-token producer builds that with `extent_to_span_bbox` (`backend/app/processing/tiles/router.py`), because a MapLibre source cannot be bounded by a `west > east` pair — so a seam-crossing raster is widened to exactly `[-180, s, 180, n]`, byte-identical to what a genuinely global raster produces.

The two are therefore indistinguishable by the time they reach the coverage math, and no arithmetic on that value can separate them. Post-#1122 the circle-aware math reads the widened pair as full coverage and stays silent for a DEM occupying a fifth of the screen. For a Fiji footprint in viewport `[179.5, -20, 190, -15]`:

| | longitudinal overlap | coverage | warns? |
|---|---|---|---|
| before #1122 (planar, clipped at +180) | 0.5 | 4.8% | yes, for the wrong reason |
| after #1122 (circle-aware) | full | 100% | no |
| truth | 2.0 | 19% | should warn |

## Fix

Feed the guard the DEM layer's `dataset_extent_bbox` instead of the token's bounds. That field has been the RFC 7946 §5.2 spec form since #1112, so a crossing extent arrives as a `west > east` pair with its real width intact, and the coverage math needs no edit at all — one line changes at the caller.

## Why this shape

The coverage module was given a value that had already lost the footprint, so the fix has to move up a layer rather than get cleverer about the arithmetic. #1122 is correct and is untouched here; this is the residual false negative in the other direction (a missing hint, not a false statement).

Both preconditions the issue flagged as unconfirmed were checked before building:

- **`dataset_extent_bbox` is populated for raster and VRT records.** It is produced at `maps/_router_helpers.py:122` from `Record.spatial_extent`, selected at `maps/service_shared.py:233` with no record-type filter; raster writes the column at `processing/ingest/tasks_raster.py:152`, VRT at `tasks_vrt.py:124` and `:939`. Decisively, the tile token reads the *same* column at `tiles/router.py:1274-1278` — so the new source is populated in exactly the cases the old one was.
- **The viewer path needs no change.** `use-viewer-terrain.ts` has a `SharedLayerResponse`, which carries no extent field — but it passes `audience: 'viewer'`, and the guard returns at `terrain-coverage.ts:234` before `demBounds` is read (#430 V-06). The viewer emits no coverage toast for any DEM, so there is no under-warning there to fix. That call site keeps the token bounds deliberately, with a comment saying why.

A layer with no extent falls back to the token span. That branch is reachable only if the two converters disagree about the same non-null `records.spatial_extent`, and where it is reached it keeps today's warning rather than silently dropping one — an absent hint is the exact failure this issue is about.

The module docstring and the `lonOverlap` branch comment both asserted the opposite of what is now true (they documented the token span as the only possible source), so they are rewritten rather than left to mislead the next reader.

## Verification

- `npm run lint` — pass
- `npm run typecheck` — pass
- `npm run test:coverage` — **370 files, 4541 tests passed, 0 failed** (baseline +5)
- `npm run test:i18n` — 1 file, 4 tests passed (no `t()` key added; run for completeness)

5 new tests pin **both directions at two levels**: the pure scorers in `terrain-coverage.test.ts` reproduce the issue's numbers (crossing = 19.05% → warns, global = 100% → silent), and `BuilderMap.terrain-visibility.test.tsx` pins the caller, where both cases get the identical token `bounds: [-180,-20,180,-15]` and differ only in layer extent. Reverting the one behavioral line fails the crossing case while the global case still passes, so the test bites on this fix specifically.

Note for the record: that caller harness's fake map had no `getBounds`, so the guard had been throwing into its own try/catch and the caller path had zero real coverage before this. The new default viewport keeps pre-existing tests behaving identically.

Closes #1128

### DIFF
--- a/frontend/src/components/builder/BuilderMap.tsx
+++ b/frontend/src/components/builder/BuilderMap.tsx
@@ -550,9 +550,20 @@ export const BuilderMap = memo(function BuilderMap({
     // of the current viewport — a small high-res DEM zoomed out reads as a
     // pedestal. Keep the active dataset's dedupe key, drop any stale one.
     resetSmallDemWarning(map, currentTerrainConfig.source_dataset_id);
+    // fix(#1128): the LAYER's extent, not the token's bounds. Both come from
+    // `records.spatial_extent`, but the token applies `extent_to_span_bbox`
+    // (processing/tiles/router.py) because a tile source cannot be bounded by a
+    // west > east pair — which flattens a Fiji-shaped DEM to [-180, s, 180, n],
+    // bit-identical to a global one, and the guard then read 100% coverage for a
+    // DEM occupying a fifth of the screen. `dataset_extent_bbox` is the RFC 7946
+    // spec form since #1112, so the crossing survives; the math needs no change.
+    //
+    // The fallback is reachable only if the two conversions disagree about the
+    // same non-null column, and it keeps today's warning rather than silently
+    // deleting it — an absent hint is the exact failure #1128 is about.
     maybeWarnSmallDemCoverage({
       map,
-      demBounds: token.bounds,
+      demBounds: demLayer.dataset_extent_bbox ?? token.bounds,
       dedupeKey: currentTerrainConfig.source_dataset_id,
     });
   }, []);
--- a/frontend/src/components/builder/__tests__/BuilderMap.terrain-visibility.test.tsx
+++ b/frontend/src/components/builder/__tests__/BuilderMap.terrain-visibility.test.tsx
@@ -12,8 +12,10 @@
 // the tokenMap has an entry and the attach path can run.
 
 import type { ReactNode } from 'react';
+import { toast } from 'sonner';
 import { act, render } from '@/test/test-utils';
 import { TERRAIN_SOURCE_ID, applyBasemapConfigToMap, syncLayersToMap } from '../map-sync';
+import { resetSmallDemWarning } from '../terrain-coverage';
 import type { MapBasemapConfig, MapLayerResponse, MapTerrainConfig } from '@/types/api';
 import type { RasterTileToken } from '@/api/tiles';
 import { BuilderMap } from '../BuilderMap';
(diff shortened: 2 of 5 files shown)


ishiland:
@codex review

Issue #1009, opened 2026-07-31
feat(analysis): show the ephemeral preview as a non-persisting row in the layer stack


ishiland:
Split out of #790 (item 4), which was carrying four unrelated pieces of work. Verified 2026-07-31 against `main` at `9a788ba92`.

An analysis preview draws an ephemeral overlay on the map that the layer stack and the legend know nothing about. With several layers on screen there is nothing that says which one is the preview, or how to get rid of it other than finding the badge. A dismissible, non-persisting row in the stack would make it obvious.

No milestone: the row's visual treatment and where it sorts are design questions, and this wants UX input before it is planned.

## What exists today

`frontend/src/components/builder/EphemeralBadge.tsx` is a badge pinned to the map's bottom-left corner, not a stack row. Since #790 was filed it has grown a second action: `onSaveAsDataset` (`:16`, rendered at `:56-64`) is the feat(#675) handoff that opens the Analysis panel prefilled with the operation behind the preview. **A stack row has to carry that action too**, or moving the affordance into the stack quietly removes it. The badge also carries the truncation count (`ephemeralBadge.featureCountTruncated`, `:24-33`) and the dismiss button (`:65-73`).

## The dnd-kit constraint

The stack is `frontend/src/components/builder/UnifiedStackPanel.tsx` (the panel and its `SortableContext`), `StackRow.tsx` (the row itself), and `SortableStackRow.tsx` (the sortable wrapper).

A non-reorderable row must **bypass the sortable wrapper**, not set a flag on it. `SortableStackRow` calls `useSortable({ id: layer.id, disabled: dragDisabled })` and is typed on `MapLayerResponse`; `dragDisabled` stops the row being dragged but the row still registers with dnd-kit and still occupies a position in the sortable collection. The `SortableContext` `items` array is `sortableIds` (`UnifiedStackPanel.tsx:507`, built from `visibleStackLayers` plus the basemap group id), and `handleDragEnd` in `MapBuilderPage` maps those ids back to real layers to write `sort_order`. A pseudo-row in that array is an id with no layer behind it, which is a reorder bug waiting for someone to drag past it.

Render the preview row as a bare `StackRow` (or a purpose-built sibling) outside the `SortableContext`, the way the `DragOverlay` ghost already renders one directly at `UnifiedStackPanel.tsx:947-957`. Keep its id out of `sortableIds` entirely.

## Acceptance criteria

- [ ] The preview appears in the layer stack as a visually distinct, non-persisting row.
[...]

PR #1130, opened 2026-08-02
feat(analysis): chain another operation onto a materialized result

### LINKED ISSUE #790: feat(analysis): two open workflow decisions — ephemeral preview as an input, shared lock budget
Companion enhancements from the 2026-07-27 audit. **Decomposed 2026-07-31** against `main` at `9a788ba92`: this issue was carrying four unrelated pieces of work, two of them executable and two of them blocked on a decision, which is why it sat.

What is left here is the two decisions. Both touch surfaces beyond the Analysis panel, and neither should be settled inside a PR that is really about something else. No milestone until they are answered.

## Split out or shipped

- Item 2, per-map persistence of last analysis parameters: **shipped** in #793. `frontend/src/stores/analysis-form-store.ts` holds `forms: Record<string, SavedAnalysisForm>` keyed per map, rather than a single shared slot, because a shared one let map B's save overwrite map A's draft in the same session.
- Item 3, cross-tab job-store sync: **→ #1008**, with the mirror-versus-claim distinction spelled out.
- Item 4, the preview as a stack row: **→ #1009**, with the dnd-kit constraint and the #675 action it has to carry.

## Decision 1: can an ephemeral preview be an input?

Today a second operation on a result requires materialize → wait → Add to map → reselect. The cheap half is a "Run another operation on this result" affordance in the completion state, which chains on the *materialized* dataset and needs no new concepts; verified absent, there is no chain affordance among the `analysisTools` keys in `frontend/src/i18n/locales/en/builder.json`.

The hard half is allowing the **preview** as an input. That means the preview becomes addressable, and it is ephemeral by design: nothing persists it, which is the property #1009 also has to respect.

The same wall is hit from the backend side by #955 (select by location), which has already recorded that it is not deciding this: "A preview is ephemeral and never persisted, so 'select within the buffer I just drew' requires materializing the buffer first. That is a workflow question about what the builder does with an ephemeral result, tracked as #790 item 1. Deciding it inside this issue would settle it for the wrong surface." So #955 ships without it, and this is the one place the question gets answered for both.

- [ ] Decide whether an ephemeral preview can be an operation input, and record the reasoning. It governs #955 and #1009 as well as this issue.
- [ ] If yes: the completion state offers a second operation on the result without a catalog round-trip, and new keys are added to all four locales (`npm run test:i18n` is the gate).
- [ ] If no: the "chain on result" affordance ships against materialized datasets only, and the constraint is written down where the next person proposing it will find it.

## Decision 2: name the shared advisory-lock budget, or split it

Preview shares its lock namespace with AI chat, so from the Analysis panel a busy chat query surfaces as "another data query is already running," which is baffling.

**Citation correction, still accurate:** the key literal is `f"geolens:ai-sql:{concurrency_key}"` at `backend/app/platform/sandbox/executor.py:163`. `run_analysis_preview` only passes `concurrency_key=str(user_id)` into `execute_safe` (`backend/app/modules/catalog/datasets/domain/service_analysis.py:190-195`).

**Blast radius, which the original framing understates:** the key lives in the shared sandbox executor that also serves AI chat data queries, so this is not an analysis-local change. Splitting the namespace changes the concurrency contract for every `execute_safe` caller. Today that budget is one expensive read per user across both features, which is a deliberate budget, not an accident. Naming the other feature in the message is the cheap half and may be the whole fix.

- [ ] Decide between splitting the key and improving the message, with the reasoning recorded, since they are not equivalent.
- [ ] If split: the per-user total concurrent-read budget is stated, and every `execute_safe` caller is checked against the new contract.
- [ ] If message only: the copy names the other feature and is added to all four locales.

---

ishiland:
## Problem

Two workflow defects from the 2026-07-27 analysis audit, both decided on #790 before this branch existed.

**1. Chaining a second operation is a scavenger hunt.** Running an operation on a result meant materialize → wait → Add to map → scroll back up and hunt the new layer out of the source picker. There was no chain affordance anywhere among the `analysisTools` keys.

**2. A busy AI chat query refuses an analysis preview with no hint why.** The sandbox advisory lock `geolens:ai-sql:{user}` (`platform/sandbox/executor.py:163`) is shared between analysis previews and AI chat data queries — one expensive read per user across both features, deliberately.

The issue describes symptom 2 as surfacing "another data query is already running", which is unattributed. It is actually one notch worse than that: `error-map.ts` had no mapping for the literal, so `client.ts:252` routed the 429 through `fallbackDescriptor(429)` → `errors.rateLimited`, and the panel toasted **"Analysis failed / Too many requests. Wait a moment and try again."** — generic rate-limit advice for something that is not rate limiting.

## Fix

**Chain affordance.** A "Run another operation on this result" button in the completion state, targeting the materialized dataset's layer. The operation input is already a layer id (`selectedLayer?.dataset_id` feeds the request), so the button reuses the source-layer switch the picker uses. The switch is extracted out of the `<Select>`'s inline `onValueChange` into `selectSourceLayer()` so both paths share one reset path — the body is the previous handler verbatim with the parameter renamed, preserving the #680/#1097 resets.

**Shared-lock message.** A new exact mapping for the backend literal → `errors.sharedQueryBusy`, whose copy names *both* features rather than only the other one, because either side can be the holder.

## Why this shape

**No new concepts, by decision.** #790 settled that an ephemeral preview is **not** an operation input: making an explicitly ephemeral result addressable is a new concept every surface holding one would have to honour (#1009's preview stack row, #955's select-by-location, the shared overlay slot chat also writes to), while chaining on the materialized half covers the workflow complaint and needs nothing new. That constraint is written down at the completion-state call site, so the next person proposing preview-as-input finds the reasoning instead of re-deriving it. No store slot, no addressable-result concept, no backend surface — the only structural change is the handler extraction.

**The lock key is unchanged.** The shared per-user budget is deliberate; splitting it would change the concurrency contract for every `execute_safe` caller and oblige a newly stated per-user total. The message was the defect.

Worth knowing about the button's placement: because the input is a layer, it appears after "Add to map", beside a satisfied "Added to map". It removes the hunt-for-the-layer step, not the Add-to-map step. Collapsing that too would need a pending-chain state machine watching `layers`, which is the concept this deliberately stops short of.

Two related things left alone on purpose:
- `chat_constants.py:55` is unchanged, so in the mirror direction (a preview holding the lock, chat refused) chat's own copy is still unattributed. That string is backend prose fed to the model, not a locale key. The new user-facing copy names both sides, so the panel direction reads correctly regardless of holder.
- The backend detail literal is unchanged. `error-map.ts` treats backend English as the stable compatibility key, so editing it would only move the coupling — and CLI/SDK callers keep the string they have today.

## API / i18n impact

No API change. Four locale files, additive only, no plurals and no interpolation:

- `analysisTools.chainOnResult` — `builder.json`, en/es/fr/de
- `errors.sharedQueryBusy` — `common.json`, en/es/fr/de

## Verification

- `npm run lint` — pass
- `npm run typecheck` — pass
- `npm run test:coverage` — **370 files, 4540 tests passed, 0 failed**
- `npm run test:i18n` — pass (all static source-referenced keys exist in the en bundles; 1 file, 4 tests)
- `npm run check:i18n:api-errors` — pass (83 stable keys, 62 exact mappings, 557 backend detail assignments)
- `npm run check:i18n:changed` — pass (builder.json, common.json present in all four locale dirs)
- `npm run build` — pass
- `npx vitest -t "790"` — 3 passed / 112 skipped, confirming the new tests actually run rather than being filtered out

New tests: `AnalysisPanel.test.tsx:2143` (3 tests, chain on the materialized result) and `error-map.test.ts:162`, which asserts the new key and that the rendered copy names AI chat instead of the rate-limit fallback.

Not verified here: the button's rendering is covered in jsdom only, not against a live stack.

Closes #790

### DIFF
--- a/frontend/src/components/builder/AnalysisPanel.tsx
+++ b/frontend/src/components/builder/AnalysisPanel.tsx
@@ -591,6 +591,15 @@ export function AnalysisPanel({
 
   const datasetLayers = layers.filter(isAnalysableLayer);
   const selectedLayer = datasetLayers.find((l) => l.id === layerId);
+  // feat(#790): the layer a finished materialize produced, once it is on the
+  // map. The operation input is a LAYER id (the select below resolves it to
+  // dataset_id), so the result has to be on the stack before it can be
+  // chained — which is why the affordance appears alongside "Add to map"
+  // rather than replacing it.
+  const resultLayer =
+    job?.status === 'complete' && job.dataset_id
+      ? datasetLayers.find((l) => l.dataset_id === job.dataset_id)
+      : undefined;
   // Candidate clip-mask layers. ux(#698): filtered to polygonal layers rather
   // than deferring to the server's 422 — dataset_geometry_type is already here,
   // so offering a point or line layer only buys the user a failed request.
@@ -790,6 +799,32 @@ export function AnalysisPanel({
       setOutputTitle('');
     }
   }, [onClearPreview]);
+
+  // feat(#790): switching the source layer is now reachable from two places —
+  // the Layer select below and the chain affordance in the completion state —
+  // so the resets live here rather than being spelled twice. A second copy
+  // would drift: every one of these lines is a fix for a field that outlived
+  // the layer it belonged to (#680, #1097).
+  const selectSourceLayer = (nextLayerId: string) => {
+    handleInputsChanged();
+    setLayerId(nextLayerId);
+    // A mask layer can't clip itself.
+    if (nextLayerId === maskLayerId) setMaskLayerId(MASK_LAYER_NONE);
+    // fix(#680): a group-by column chosen for one dataset must not carry to
+    // another — it may not exist there (422 from the API) or silently group by
+    // a same-named field.
+    setByField(BY_FIELD_NONE);
+    // fix(#1097 review): the transferred field is the same problem and was
+    // left out of this reset. It belongs to the JOIN layer, so it survives a
+    // source change intact — but whether it is usable depends on the SOURCE,
+    // since join_<name> has to not collide with a source column. Picking
+    // `zone` against a source with no join_zone and then switching to one that
+    // has it left the menu filtering `zone` out while the state still held it,
+    // and the request still went (and earned a 422). A join layer can't join
+    // against itself either, so the layer goes with the field.
+    if (nextLayerId === joinLayerId) setJoinLayerId(MASK_LAYER_NONE);
+    setJoinField(BY_FIELD_NONE);
+  };
 
   // fix(#793 review), mount half: see staleRestoreRef above.
   useEffect(() => {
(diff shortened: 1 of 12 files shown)


ishiland:
@codex review

PR #1133, opened 2026-08-02
test(ai): observe which tool chat routes a message to

### LINKED ISSUE #1007: test(ai): tool-choice eval harness so a chat routing regression fails a run

ishiland:
## Gap

The repo has AI evals, but none can catch a tool-routing regression. `tests/evals/test_sql_generation_evals.py` calls `generate_sql()` directly and asserts on SQL text and executed values. It never runs the tool-calling loop, so nothing anywhere asserts **which tool the model chose** for a given message.

That is where routing bugs live. #549 was one — "show me X" landing on `set_filter` instead of `query_data`. Its fix has already landed (`tools.py` carries `fix(#549)` comments on `set_filter`, `set_style`, `query_data` and `run_analysis`; `chat_service.py:268-305` is now the single owner of verb classification), which strengthens rather than weakens the premise: that prose exists today and nothing protects it.

## What this adds

`backend/tests/evals/test_tool_choice_evals.py` — six cases driving the live chat loop and observing the emitted tool name.

| case | message | expects |
|---|---|---|
| `show_ada_stations` | "Show the ADA accessible stations served by the A line" | `query_data` (the #549 regression itself) |
| `filter_ada_stations` | "Filter to ADA accessible stations on the A line" | `set_filter` (the over-correction guard — other direction of the pair) |
| `centroid_of_parcels` | "Show the centre point of each parcel" | `run_analysis` (transform must outrank the "show" question verb) |
| `count_near_schools` | "How many parcels are within 500 meters of a school?" | `query_data` (reuses buffer phrasing but asks a count, which no preview can produce) |
| `readonly_surface_no_edit_tool` | `can_edit=False`, the `set_filter` message | forbids all of `_EDIT_TOOLS` |
| `mapless_surface_no_run_analysis` | `has_map=False`, the `run_analysis` message | forbids `run_analysis` |

It drives the real path, not a reconstruction: the tool set comes from `select_chat_tools(case.can_edit, case.has_map)` and the prompt from `build_chat_system_prompt` / `build_dataset_chat_system_prompt`, the same builders chat uses. The read-only case forbids the production `_EDIT_TOOLS` set rather than a transcribed literal, so adding an editing tool extends the guard for free.

## Every case ships reported-only, and that is deliberate

**Not one row has ever been executed against a live provider.** No provider key was reachable when this was built, and validating tool-name stability is an observation, not something that can be reasoned out — the issue says as much.

This matters because the nightly job (`ci.yml`) runs `pytest tests/evals/` with `RUN_AI_EVALS=1`, so this file is picked up there automatically. An assertion nobody had ever run could turn a nightly red for a reason no one had seen. So every row carries `stable=False` and warns instead of asserting; verified safe rather than assumed — there is no `filterwarnings=error` in `pyproject.toml`/`pytest.ini`/`setup.cfg`/`conftest.py` and no `-W error` in the job, so `warnings.warn` reports without failing.

The never-run state is loud in the file, not just here: a block comment above the case table opens "THE CASE TABLE. NOT ONE ROW HAS EVER BEEN RUN AGAINST A LIVE PROVIDER", states that `stable=False` is **not** a flakiness judgement, and gives the five-step promotion procedure. Promotion is editing `stable=False` → `stable=True` on one line — proven, not asserted: feeding the same wrong tool through the real provider loop, `stable=False` warned and passed, `stable=True` raised `AssertionError` and failed.

One hard assertion does run — `assert case.expects in offered` — which catches a table row naming a tool its own surface does not offer. It is deterministic, has no live behaviour, and cannot flake.

## Two things found on the way

- **`temperature=0.0` is a no-op on the Anthropic path.** `DefaultAnthropicProvider.complete` does `del temperature` (`defaults_ai_anthropic.py:50` and `:253`) because Claude 4.6+ rejects a non-default value. The flake policy both eval files describe therefore only binds on the OpenAI-compatible path, and production chat runs at 0.3 (`chat_service.py:420`, `streaming.py:495`) rather than 0.0. Documented in the new module docstring instead of repeating the claim as fact.
- The nightly step was named "Run live NL→SQL evals" but globs `tests/evals/`, so it now runs both. Renamed in this PR, since this is the change that made the old name wrong.

## Verification

- `uv run ruff check .` — All checks passed
- `uv run ruff format --check .` — 878 files already formatted
- `uv run pytest tests/evals/ -v` — **16 collected, 16 skipped, 0 failed** (10 pre-existing + 6 new), zero provider calls: the skip is evaluated at collection, so no test body runs
- `uv run pytest tests/test_layering.py -q` — 36 passed (nothing under `backend/app/` touched)

Surface differences were confirmed against the real provider with a stubbed SDK client (zero tokens): `select_chat_tools(True,True)` → 11 tools, `(False,True)` → `['query_data','run_analysis']`, `(False,False)` → `['query_data']`; the read-only prompt carries its "Read-Only Access" block; the dataset prompt never mentions `run_analysis`; `max_rounds=1` bills exactly one call across all three response shapes; and two tool blocks in one turn are both captured, so "also emitted an editing tool" is visible.

**Follow-up required before any row can be trusted:** three live runs of this file, then flip `stable=True` per row that agreed every time.

Closes #1007

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1877,7 +1877,7 @@ jobs:
       - name: Run migrations
         run: uv run alembic upgrade heads
 
-      - name: Run live NL→SQL evals
+      - name: Run live AI evals (NL→SQL and tool choice)
         run: uv run pytest tests/evals/ -v --tb=short
 
   # ---- E2E / A11Y trigger matrix (REL-03) ----
--- a/backend/tests/evals/test_tool_choice_evals.py
+++ b/backend/tests/evals/test_tool_choice_evals.py
@@ -1,0 +1,400 @@
+"""Live-provider AI evals for chat TOOL CHOICE (#1007, follow-up to #549).
+
+``test_sql_generation_evals.py`` asserts what the model WRITES once a tool has
+been chosen. These watch WHICH TOOL it chooses for a given user message, which
+is the layer routing bugs actually live in. #549 was "show me X" landing on
+``set_filter`` instead of ``query_data``; its fix is prompt and tool-description
+prose whose entire risk is that a later edit quietly reverses it, with nothing
+in CI or in ``make ai-evals`` to notice.
+
+Every case drives the production routing inputs rather than a reconstruction:
+
+* the tool set comes from ``select_chat_tools(can_edit, has_map)``, and
+* the system prompt from the same builders the chat surfaces use,
+  ``build_chat_system_prompt`` for map chat (streaming.py:773,
+  chat_service.py:370) and ``build_dataset_chat_system_prompt`` for
+  dataset-scoped chat (router.py:716),
+
+so an edit to either is what the eval sees. Nothing here restates a tool
+description or a verb-class rule. If it did, the eval would keep passing after
+the prose it protects changed, which is the failure mode this file exists to
+prevent.
+
+## What is checked
+
+The emitted tool NAME, and for the surface cases the ABSENCE of tools the
+surface does not offer. Never the response prose: a routing eval that reads
+prose fails on wording changes that broke nothing.
+
+## Cost and gating
+
+Live-provider, real tokens. SKIPPED unless ``RUN_AI_EVALS=1``; run with::
+
+    make ai-evals
+    # or, this file only:
+    cd backend && set -a && source ../.env.test && set +a && \\
+        RUN_AI_EVALS=1 uv run pytest tests/evals/test_tool_choice_evals.py -v
+
+Each case costs exactly ONE provider round-trip: ``_route()`` passes
+``max_rounds=1`` and never executes a tool, so the loop stops as soon as the
+routing decision is visible. No database rows are seeded, because tool choice
+is decided entirely by the prompt and the tool set; the layers are in-memory
+``ChatMapLayer`` values.
+
+## Flake policy
+
+Inherited from the SQL evals, with one caveat worth stating plainly:
+``temperature=0.0`` is passed, but ``DefaultAnthropicProvider.complete`` deletes
+the argument (Claude 4.6+ rejects a non-default temperature), so on the
+Anthropic path determinism comes from the provider default rather than from
+this call. It is honoured on the OpenAI-compatible path.
+
+Checks therefore leave the model freedom everywhere it legitimately has it: a
+case names one tool, or an absence, never a tool ARGUMENT and never the
+accompanying text.
+
+## Read the case table before you trust a result
+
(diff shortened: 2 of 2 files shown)


ishiland:
@codex review

PR #1134, opened 2026-08-02
refactor(analysis): split analysis_sql by operation family

### LINKED ISSUE #1089: refactor(analysis): split analysis_sql by operation family

ishiland:
## Why

`backend/app/platform/analysis_sql.py` reached **1255 lines** — the largest module under `platform/`, gated by a ratchet rather than split. This is the follow-up that entry points at.

## The split

A package with the shared machinery central and one module per operation family:

| module | lines | contents |
|---|---|---|
| `__init__.py` (façade) | 193 | the stable import surface |
| `shared.py` | 365 | machinery more than one family needs |
| `overlay.py` | 397 | clip, intersect, select by location |
| `transform.py` | 307 | buffer, centroid, dissolve |
| `spatial_join.py` | 220 | spatial join |
| `measure.py` | 77 | area, length |

**Not split by caller** — that is the one thing this module exists to prevent. It lives in `platform/` so the catalog preview path and the processing materialize worker render *identical* SQL; before it existed each had its own copy and they drifted, so an approved preview and the dataset it saved could disagree about what the operation meant.

`LATERAL_ALIAS`, `NOT_EMPTY_PREDICATE`, the `OFFSET 0` fence, the `MAX_*` bounds with their benchmark tables, the antimeridian helpers and the injection boundary all stay central — duplicating any of them would recreate a drift surface.

**One deviation from the issue's taxonomy: spatial join gets its own module.** The issue names overlay/measure/transform, written before spatial join landed in the same batch (#953). It fits neither: it reads a second layer like overlay but has no mask, no `ST_Dimension = 2` guard and no subdivide path — every specific thing the issue says overlay members share — and it shares measure's `(select_columns, join_clause)` contract while not being a measurement and being 3× its size. Its 220 lines of identifier-truncation and join-field rules touch nothing else. Easy to fold into `measure` if reviewers prefer.

Three symbols the issue's central list predates (#1097/#1099) needed the most judgement — `INTERNAL_ALIAS_PREFIX`, `NON_GROUPABLE_COLUMN_TYPES`, `MAX_IDENTIFIER_LENGTH`. All went central, with the reasoning beside each.

## Acceptance

1. **Import surface stable.** All seven callers — `service_analysis.py`, `tasks.py`, `router_analysis.py`, `schemas.py`, the sandbox validator, `sql_generator.py`, `defaults_processing_port.py` — are byte-identical to main; the diff touches only the package and its tests. No OpenAPI or SDK impact.
2. **Every rendered statement byte-identical.** 94 cases compared against a fixture generated by running the harness on the *pre-split* module, mutation-tested in both directions. It earned its keep: it caught `ST_MakeValid` being dropped from the clip `WHERE` during the overlay carve — still valid SQL, still a plausible predicate, different row set on any dataset with a self-intersecting ring.
3. **Ratchet updated by measurement.** The `analysis_sql.py: 1255` entry is removed because the file is gone; no other cap value changed. The façade is 193 and no family reaches 400, so nothing meets the 1000-line inclusion threshold. Retargeted mid-branch so every commit stays green. The entry's narrative was rewritten rather than deleted — it carried the record of *why* this must never be split by caller.
4. **`platform/` rule holds** — zero `app.modules.*` imports in the package. A new layering test fails the build if a caller bypasses the façade.

**Symbol surface, measured end to end** (an earlier revision of this body claimed "the symbol diff is empty" — that was drawn from the `__all__` comparison alone, and `__all__` is not what callers bind to; corrected here):

- `__all__` equals the 35 pre-split defined names exactly — no additions, no removals, every one importable.
- Importable public attributes total **42**: those 35, the 5 family submodules (bound on the parent package by the import machinery — unavoidable for a package), plus `annotations` and `Any`.
- Against the pre-split module: the **losses** are 7 of its own stdlib/shapely imports (`GEOSException`, `LON_EPSILON_DEGREES`, `Sequence`, `math`, `re`, `shape`, `shapely`), none of which is imported from `analysis_sql` anywhere in `backend/`, `cli/`, `mcp/` or `sdks/`. The **gains** are the 5 submodules and nothing else.

So no name the single file defined is gone, and no new callable is reachable.

## Not in scope

No SQL changed. This is a move. Seven commits, one per carve, so the move can be read separately from any accident.

## Review rounds

Four rounds, all on the guard rather than the move — the rendered SQL never changed.

**Rounds 1–2, the façade-bypass guard.** It matched literal names where it should resolve a name to what it is *bound* to, at two depths: `node.module` first, then an attribute base. Chasing alias spellings cannot terminate, so the mechanism was replaced with a per-file binding map. Measured against a 22-shape table, the original matcher missed 9; the current one resolves imported aliases, plain rebinds, annotated rebinds, walrus, tuple unpack, function-local imports and instance attributes, with 0 false positives across 12 must-pass shapes — including `from app.platform import cache as sql` → `sql.shared`, which a guard that had merely learned the word `sql` would flag.

Bindings are collected file-wide rather than per-scope, so the guard errs toward flagging and never toward missing. What remains structurally out of reach is written into the helper docstring — `getattr`, non-literal `importlib`, protocol-mediated binding (`for`/`with`), parameter defaults, and unmodelled data flow — so the guard's contract is "catches every statically resolvable path", not "catches everything". None of those shapes appears anywhere under `backend/app/`.

**Round 3, a real surface expansion.** Composing `render_geometry_expr` in the façade imported six family helpers under public names, so `from app.platform.analysis_sql import render_clip_expr` succeeded on this branch and fails on `main` — the six do not exist in the pre-split namespace at all. They are now bound `_`-prefixed, and a new test asserts public attributes equal `__all__` with a named exemption per unavoidable entry. That test also fails on a *stale* exemption, so the list cannot become the next place an expansion hides. It would have caught this without a reviewer.

The byte-identity harness was load-bearing here: round 3 is the first review commit to touch app code, and all 94 rendered statements were unchanged under it.

## Verification

Re-run after rebasing onto `bf3548f68`:

- `uv run ruff check .` — All checks passed (never `--fix`; it strips façade re-exports)
- `uv run ruff format --check .` — 883 files already formatted
- `uv run pytest tests/test_layering.py tests/test_analysis_sql_split_1089.py -q` — **41 passed**
- `uv run pytest tests/ -k "analysis or sandbox or sql" -n 4 --dist loadgroup` — **733 passed, 11 skipped**
- Full backend suite before the rebase — **6870 passed, 83 skipped**

The byte-identity harness (`test_analysis_sql_split_1089.py` + its fixture) has been **retired in the final commit**, having done its job twice — once catching `ST_MakeValid` dropped from the clip `WHERE` during the overlay carve, and again over the review commit that touched app code. It was deliberately kept through the review rounds rather than deleted with the last carve, because a fix pushed during review is exactly when byte-identity can break.

What protects the module from here is not a frozen copy of today's SQL, which would only make the next legitimate statement change look like a regression. It is the façade-bypass guard and the declared-surface test, both of which police a property that outlives the move.

Closes #1089


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1136, opened 2026-08-03
chore(repo): decision-queue hygiene — env-correct docs, CI-matched Python pin, #695 trap preserved

### LINKED ISSUE #695: perf(jobs): dedicated analysis queue when a second worker exists

ishiland:
## Problem

Three small repo hazards, each found the expensive way during the post-v1.7.1 waves:

1. **AGENTS.md's single-test guidance was wrong about `test_layering.py`.** It said the file "needs no database and takes seconds" — but it boots `app.core.config`, so a bare run exits non-zero on missing env vars *before collecting anything*, which reads exactly like a gate failure. That line is what sends agents and contributors into it.
2. **Local Python silently diverged from CI.** CI pins 3.13 (`setup-python`, four jobs), but with no `.python-version` and `requires-python = ">=3.13"`, uv gave a fresh worktree venv CPython 3.14.5. Harmless until a 3.14-only behaviour passes locally and fails CI — and invisible until that day by construction.
3. **#695's deploy-order trap would have died with its issue.** #695 is decided (no second worker service in the default topology; #703's defer-site priority already solved single-worker head-of-line), but the trap it documented outlives it: a job deferred to a queue no worker subscribes to is *not an error* — and (worse than #695's own text believed) it is never auto-failed either: #724 made both stale-sweep paths skip live `todo` rows (`no_live_procrastinate_job`), so it sits pending **indefinitely**, silent, until a worker subscribes.

## Fix

- AGENTS.md recipe now sources `.env.test` like every other backend invocation, and says why. Same paragraph area gains the standing CodeQL policy (decided 2026-08-03): next time the validated-identifier `py/sql-injection` class fires on an ingest-adjacent PR, adopt the alert-suppression query pack instead of another round of manual dismissals.
- `.python-version` = `3.13` in **each uv project** (`backend/`, `cli/`, `mcp/`) — not at the repo root, which was tried first and measurably did not reach `uv run` inside `backend/`.
- The queue trap note sits at the exact line someone would edit — `WORKER_QUEUES` in **both** compose files — and names the decorator site (`processing/analysis/tasks.py`) and `worker_queues` (`core/config.py`) so the whole change-set is findable from one place.

## Verification

- `docker compose config -q` and `docker compose -f docker-compose.prod.yml config -q` — both parse.
- Backend venv recreated under the pin: `uv run python --version` → **3.13.13** (was 3.14.5); cli and mcp resolve 3.13.13 likewise.
- On the rebuilt 3.13 env: `uv run pytest tests/test_layering.py -q` → **39 passed**; `ruff check` clean; `ruff format --check` → 883 files already formatted.
- No app code touched; comments, docs, and interpreter pins only.

## Review rounds

1. **P1 — the 3.13 pin must not reach the Docker build.** The builder is `python:3.14.6-slim`; `COPY backend/ /app` carried the pin in, and uv would have satisfied it with a managed 3.13 *outside* `/app` — which the runtime stage never copies. This failure mode **builds green** and dies at container exec, so it is verified from artifacts, not from a passing build: a context probe shows `.python-version` absent, and the rebuilt builder's `/app/.venv/bin/python` resolves to `/usr/local/bin/python3.14` inside the image. Fixed with one `.dockerignore` entry (`**/.python-version`). The deeper thing this exposed — CI tests on 3.13 while published images run 3.14.6 — is deliberately not bundled here; filed as #1137.
2. **P2 — the layering recipe broke in a fresh clone** (`.env.test` is gitignored and absent there). Now carries the one-time `cp .env.test.example .env.test` step.
3. **P2 — the unserved-queue lifetime was stale.** The note shipped with #695's own "failed an hour later" claim, which #724 has since made false: both auto-fail paths deliberately skip live `todo` rows, so the outage is indefinite. Both compose copies now state the real lifetime — a warning that understates an outage teaches the reader to under-react.

Closes #695


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

Issue #1119, opened 2026-08-02
url_redaction: redact_url_credentials raises ValueError on malformed IPv6 authorities


ishiland:
## Problem

`redact_url_credentials` in `backend/app/core/url_redaction.py` lets `urlsplit`'s `ValueError` escape. Nothing catches it.

```
>>> redact_url_credentials("https://.[::1]")
ValueError: Invalid IPv6 URL
>>> redact_url_credentials("ogrinfo failed: https://.[::1] bad")
ValueError: Invalid IPv6 URL
```

Both forms reproduce on current main. The second matters more: it goes through the scheme-less free-text path, which is the one fed by GDAL/ogr2ogr stderr and source-preview text.

Found while working #1116. It is present identically before and after that change, so #1116 neither caused it nor fixes it. A differential fuzz run over 20,000 randomized inputs hit it 485 times, so the input shape is not especially exotic.

## Impact

A redaction call is supposed to make an error message safe to log. Here it replaces the error with a different, unhandled one.

At `backend/app/modules/catalog/sources/preview.py:187` the redacted string is interpolated into an `IngestionError`. If the stderr being redacted contains a malformed IPv6 authority, the `ValueError` propagates instead and the caller gets an unhandled exception rather than the intended ingestion failure. The same applies at the three `processing/ingest/validation.py` sites.

So a hostile or merely malformed source can convert a clean, user-facing ingest error into a 500.

## Proposed fix

Catch `ValueError` around the `urlsplit` call and fall back to returning the input with the regex redaction applied, or to a fixed placeholder if even that cannot be done safely. The function's contract is "return something safe to log", and raising is the one outcome that breaks it.

Worth pinning with a test that the function never raises for any input, since that is the property callers actually depend on. A parametrized set of malformed authorities plus the free-text wrapper covers it.

## Notes

Out of scope for #1116, which bounds the regex quantifier against quadratic backtracking and deliberately changes no parsing behaviour.

PR #1162, opened 2026-08-03
fix(core): refuse an unparsable authority in has_url_credentials

### LINKED ISSUE #1132: url_redaction: has_url_credentials raises ValueError on a malformed authority, unguarded at the connector-config gate
## Problem

`has_url_credentials` (`backend/app/core/url_redaction.py:66`) calls `urlsplit` unguarded, so it has the same failure mode #1119 fixed in its sibling `redact_url_credentials`:

```
>>> has_url_credentials("https://.[::1]")
ValueError: Invalid IPv6 URL
>>> has_url_credentials("https://[::1")
ValueError: Invalid IPv6 URL
>>> has_url_credentials("https://user:pw@.[::1]/x")
ValueError: Invalid IPv6 URL
```

All three reproduce on current main (`0506a5260`).

## Why most call sites are fine, and one is not

Most callers are pydantic validators, where a `ValueError` becomes a clean 422; `modules/catalog/sources/router.py:226` catches it into a 502. Those are all acceptable outcomes.

The exception is `_metadata_contains_secret` (`sources/router.py:159-170`), which recurses through a connector config and calls `has_url_credentials(value)` on **every string it finds**:

```python
elif isinstance(value, str):
    return has_url_credentials(value)
```

It is reached from `_reject_inline_connector_secrets` (`:173`), which request handlers call at `:282` and `:355`, and nothing between the handler and the raise catches `ValueError`. A connector config carrying a string with a malformed bracketed authority therefore turns the inline-secret gate into an unhandled 500 instead of the intended 400.

Same shape as #1119: a guard whose job is to answer a yes/no question about a string raises on the string instead.

## Suggested fix

Guard the `urlsplit` call the same way #1119 guarded its own, and pick the answer deliberately — for a **detector**, the safe default is the one where a mistake is visible. Returning `True` on an unparsable authority ("might contain credentials, refuse it") fails loudly and is self-correcting; returning `False` makes the gate silently permissive on exactly the malformed input an attacker controls. Prefer the refusing default unless there is a concrete reason not to.

Worth pinning with the same never-raises property test #1119 added, and a case driving the real gate: a connector config containing a malformed-authority string must produce the 400, not a 500.

## Notes

Found while implementing #1119, which deliberately scoped itself to `redact_url_credentials`. The two functions were already noted as mirroring each other (`has_url_credentials`'s own `fix(#430 BA-04)` comment says it mirrors `redact_url_credentials`) — the mirror just did not extend to the error path.

ishiland:
## Problem

`has_url_credentials` called `urlsplit` unguarded, so a malformed bracketed authority raised instead of answering:

```
>>> has_url_credentials("https://.[::1]")
ValueError: Invalid IPv6 URL
```

Same escape #1119 fixed in its sibling `redact_url_credentials` — whose own `fix(#430 BA-04)` comment says the two functions mirror each other. The mirror never reached the error path.

Most callers absorb the raise acceptably: three pydantic validators run `HttpUrl()` before asking and it rejects every one of these strings first, so the detector never sees them; `_validate_connector_job` catches `ValueError` into a 502.

`_metadata_contains_secret` is the exception. It walks a caller-supplied connector config asking this question of **every string it finds**, and both handlers reach it through `_reject_inline_connector_secrets` *before* their `try` block. A bracketed authority nothing can parse therefore escaped as an unhandled 500 instead of the gate's intended 400.

## Fix

Guard the `urlsplit` the way #1119 guarded its own, and answer **True**.

The asymmetry is the whole point. This is a detector, so refusing an authority nobody can parse costs a rejected config that fails loudly and gets fixed; admitting it leaves the inline-secret gate silently permissive on exactly the malformed input an attacker controls. No credential gets reported as absent because the parser gave up.

## Why this shape

Tests reuse #1119's corpus rather than a fresh list, so the two mirrored functions cannot drift apart again without something failing.

The malformed-authority list is partitioned by whether `urlsplit` actually rejects the string: `":notaport"` parses, and this function never reads `SplitResult.port`, so it belongs on the admission side and is asserted `False` there. That partition matters — a refusal assertion alone cannot notice the guard widening into "anything bracket-shaped is a credential". Alongside it: the sibling seeded fuzz, the credentialed-malformed list, and four connector configs driving the real gate to its 400.

## Verification

Measured against the unguarded function: **38 new failures, four of them raw `ValueError`s out of the handlers** — so the tests fail without the fix.

```
uv run ruff check .                                    All checks passed!
uv run ruff format --check .                           883 files already formatted
uv run pytest tests/test_layering.py -q                39 passed
uv run pytest tests/test_url_redaction_sec.py \
              tests/test_connector_routes.py -q       190 passed
```

Closes #1132

### DIFF
--- a/backend/app/core/url_redaction.py
+++ b/backend/app/core/url_redaction.py
@@ -57,14 +57,33 @@ def query_has_credentials(query: str) -> bool:
 
 
 def has_url_credentials(url: str) -> bool:
-    """Return True if a URL carries credential-like userinfo or query params."""
+    """Return True if a URL carries credential-like userinfo or query params.
+
+    Also True when the authority is unparsable: callers use this to admit or
+    refuse a string, so "cannot tell" has to resolve to refusal, not to "no".
+    """
     # fix(#430 BA-04): strip GDAL-style prefixes (ESRIJSON:, WFS:, ...) before
     # inspecting userinfo — otherwise urlsplit sees no netloc and misses
     # `user:pass@` behind the prefix, mirroring redact_url_credentials.
     prefixed = _split_prefixed_url(url)
     if prefixed is not None:
         return has_url_credentials(prefixed[1])
-    parts = urlsplit(url)
+    try:
+        parts = urlsplit(url)
+    except ValueError:
+        # fix(#1132): the same malformed-authority raise #1119 fixed in
+        # redact_url_credentials — the mirror stopped at the happy path. The
+        # caller that makes it a bug is _metadata_contains_secret in
+        # modules/catalog/sources/router.py, which asks this question of every
+        # string in a connector config from OUTSIDE the handler's try block, so
+        # a raise there is an unhandled 500 instead of the gate's 400.
+        #
+        # True, not False, and the asymmetry is the whole point: this is a
+        # detector, so refusing an authority nobody can parse costs a rejected
+        # config that fails loudly and gets fixed, while admitting it makes the
+        # gate silently permissive on exactly the input an attacker controls.
+        # No credential is reported as absent because the parser gave up.
+        return True
     return bool(parts.username or parts.password) or query_has_credentials(parts.query)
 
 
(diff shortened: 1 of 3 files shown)


ishiland:
@codex review

PR #1163, opened 2026-08-03
fix(auth): refresh the login page provider list on admin OAuth/SAML changes

### LINKED ISSUE #1117: bug(auth): admin OAuth provider changes never invalidate the login page's provider list

ishiland:
## Problem

Both admin provider surfaces invalidated only their own table key, `settingsOAuth.providers`. The login page's buttons read a different key, `authConfig.oauthProviders` (`components/auth/OAuthButtons.tsx:86`). Neither surface invalidated the other's key, so an admin who added or removed a provider and then logged out kept the stale button set until a full reload.

Six mutation sites had the gap — create/update/delete in each of two files:

- `components/admin/settings/SettingsAuthTab.tsx:145,157,168`
- `components/admin/saml/SamlProvidersSection.tsx:135,148,160`

## Fix

A shared `useInvalidateAuthProviders()` hook (`hooks/use-auth-providers.ts`) that invalidates both key families together. All six `onSuccess` handlers call it. The issue offered either six extra inline lines or one hook; the hook is what stops the pairing drifting apart again, which is the actual failure mode here.

`SamlProvidersSection` keeps its own `['saml', 'providers']` invalidation — that key is uniquely its own and unrelated to the login page.

## Why SAML is included

Not defensive — verified from source. An enabled SAML row is returned to the login page verbatim:

1. `OAuthButtons.tsx:86` reads `queryKeys.authConfig.oauthProviders` via `getOAuthProviders`
2. `api/auth.ts:100-108` → `GET /auth/oauth/providers/`
3. `backend/app/modules/auth/oauth/router.py:501` → `get_enabled_providers(db)`
4. `service.py:468-470` → `list_providers(db, enabled_only=True)`
5. `service.py:452-465` builds `select(OAuthProvider)` + `.where(enabled.is_(True))` — **no `provider_type` filter anywhere in that query**
6. SAML providers persist in the same `catalog.oauth_providers` table (`SamlProviderConfig extends Omit<OAuthProviderConfig, 'provider_type'>`) via the same `/settings/oauth-providers/` endpoints

So creating or deleting a SAML provider changes the login button set.

## Also corrected

`hooks/use-config-ops.ts` carried a stale comment asserting this gap "is not fixed here". `useImportConfig` already invalidated `authConfig.oauthProviders` (#1021), so it never had the bug and the note is now false. Comment-only change; the call site stays inline because it invalidates six unrelated key families, not just the provider pair.

## Verification

```
npm run typecheck      clean
npm run lint           clean
npm run test:coverage  372 files, 4553 tests passed
npm run test:i18n      4 passed (no locale keys touched; run as insurance)
```

New tests assert **both** key families for create/update/delete in three files, including a new `__tests__/` dir for `SamlProvidersSection` (which had none).

Non-vacuity checked rather than assumed: stubbing the second `invalidateQueries` out of the hook produced 7 failed / 13 passed, and the 7 failures were exactly the 7 new assertions.

## Noted, not fixed

`SamlProvidersSection` reads and invalidates its own list under a raw inline `['saml', 'providers']` array literal instead of an entry in `lib/query-keys.ts` — the same class of drift risk this PR is about, but out of scope and it would touch the query-keys parity test.

Closes #1117

### DIFF
--- a/frontend/src/components/admin/saml/SamlProvidersSection.tsx
+++ b/frontend/src/components/admin/saml/SamlProvidersSection.tsx
@@ -72,6 +72,7 @@ import {
   type SamlProviderUpdateData,
 } from '@/api/saml';
 import { queryKeys } from '@/lib/query-keys';
+import { useInvalidateAuthProviders } from '@/hooks/use-auth-providers';
 import { triggerDownload } from '@/lib/download';
 import { Textarea } from '@/components/ui/textarea';
 
@@ -111,6 +112,10 @@ const EMPTY_FORM: SamlFormData = {
 export function SamlProvidersSection() {
   const { t } = useTranslation('admin');
   const queryClient = useQueryClient();
+  // fix(#1117): SAML rows live in the same table as OAuth and are returned by the
+  // login page's /auth/oauth/providers/, so every mutation here also has to refresh
+  // the login buttons — see hooks/use-auth-providers.ts.
+  const invalidateAuthProviders = useInvalidateAuthProviders();
 
   const {
     data: providers = [],
@@ -132,7 +137,7 @@ export function SamlProvidersSection() {
     mutationFn: (data: SamlProviderCreateData) => createSamlProvider(data),
     onSuccess: () => {
       queryClient.invalidateQueries({ queryKey: ['saml', 'providers'] });
-      queryClient.invalidateQueries({ queryKey: queryKeys.settingsOAuth.providers });
+      invalidateAuthProviders();
       toast.success(t('saml.created'));
     },
     onError: () => {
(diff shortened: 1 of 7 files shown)


ishiland:
@codex review

PR #1165, opened 2026-08-03
feat(builder): show the ephemeral preview as a row in the layer stack

### LINKED ISSUE #1009: feat(analysis): show the ephemeral preview as a non-persisting row in the layer stack

ishiland:
## What

An analysis preview drew an ephemeral overlay the layer stack and legend knew nothing about. With several layers on screen nothing said which one was the preview, or how to remove it short of finding the badge in the map corner. This adds a dismissible, non-persisting row pinned to the top of the stack.

Design settled on the issue: pinned above the sortable region, reusing the ephemeral vocabulary the badge established (dashed outline, existing muted/accent tint, an explicit "Ephemeral — not saved" tag), no new tokens.

## How the row stays out of the sortable collection

Structurally, not by a flag — two independent facts:

**It cannot enter `sortableIds`.** That memo (`UnifiedStackPanel.tsx:514-520`) is built only from `visibleStackLayers.map(l => l.id)` plus the basemap group id. The preview is never in `layers` — it lives in `useEphemeralLayers` state and arrives as its own prop. No filter or exclusion was added because there is nothing to exclude.

**It renders outside the `SortableContext`.** The row renders inside the scroll container but *before* the `isEmpty` ternary, and `<SortableContext>` only opens inside the non-empty branch. It is a sibling preceding the provider, exactly like the `DragOverlay` ghost's bare `StackRow`. It calls no `useSortable`, carries no `data-row-id` (what the Shift+Arrow walker reads), and renders no grip.

That placement also satisfies pinned-at-top for free: it precedes the basemap row even when `basemapPosition === 'top'`.

## The badge is not deleted — it is now the no-stack surface

The recorded decision was "badge removed in the same PR". Taken literally that breaks two places, because **the decision's rationale presumes a layer stack exists and there are two contexts where none does**:

1. **The public viewer.** `ViewerChatPanel.tsx:279` renders the badge, under a `fix(#542)` comment reading *"the viewer applied the query overlay with no legend entry, badge, or dismissal — orange highlights persisted unexplained."* The viewer has no layer stack.
2. **The <1100px builder rail.** `MapBuilderPage.tsx:1856` swaps the entire `UnifiedStackPanel` for `SidebarRail` below `BUILDER_RAIL_BREAKPOINT = 1100`. `SidebarRail` is a 64px icon strip — zero `StackRow`, zero `data-row-id` in its 171 lines.

Deleting the badge would lose the count, the dismiss control and the #675 save-as-dataset hand-off entirely in both — not degraded, absent, and #542 re-opens verbatim.

So the badge is gated on `isRail` in the builder and kept in the viewer. **Row and badge are now mutually exclusive by construction — never both mounted** — which makes "one ephemeral object gets one surface" literally true rather than approximately, and loses nothing at any width.

Net rule, recorded in-source in three places (`EphemeralBadge.tsx` header, the viewer render site, the badge test file): **the row is the surface wherever a layer stack exists; the badge is the surface where none does.**

## Verification

```
npm run typecheck      clean
npm run lint           clean
npm run test:i18n      4 passed
npm run test:coverage  371 files, 4559 tests passed
```

14 new tests in `__tests__/UnifiedStackPanel.preview-row.test.tsx`. The collection invariant is asserted directly: every id the panel hands `SortableContext` resolves to a real layer or the basemap, and `items.length === layers.length + 1` — which is what `handleDragEnd` depends on, since it does `findIndex(l => l.id === active.id)` and writes `sort_order` from the result, so a member with no layer behind it is the whole corruption vector. Also pinned: identical collections with and without a preview, and `compareDocumentPosition` proving the row precedes every `[data-row-id]`.

Non-vacuity checked rather than assumed: prepending a fake id to the `sortableIds` memo produced exactly 2 failed / 12 passed (the two collection tests), then restored to 14/14.

Stated plainly: this pins the id boundary, not a simulated drag. No test here drives dnd-kit end-to-end — `handleDragEnd` is a `useCallback` inside a 2250-line page with no exported seam.

`EphemeralBadge.test.tsx` passes unchanged (6 cases); only an 8-line scope note was added so the file does not read as something the deletion missed.

## Locale keys

Exactly one added, in all four locales: `previewRow.ephemeralTag`. None removed, none relocated — `featureCountTruncated` must stay because the badge still needs it, so both surfaces now read the same `ephemeralBadge.*` keys through a shared `ephemeral-preview.ts` and there is no duplicate to drift. No plural key was touched; verified every `_one` interpolates `{{count}}` and `_many` count is 0 in all four.

## Note for the e2e run

`e2e/dataset-chat.spec.ts` edits are comments only; the assertion is unchanged and still valid because the row carries the same `role="status"` sentence the badge did. Playwright's `Desktop Chrome` is 1280px — above the rail breakpoint — so that spec now exercises the row, not the badge.

Closes #1009

### DIFF
--- a/e2e/dataset-chat.spec.ts
+++ b/e2e/dataset-chat.spec.ts
@@ -215,9 +215,9 @@ test.describe('Dataset AI chat', () => {
     expect(createdMapId).toBeTruthy();
 
     await expect(page.locator('canvas.maplibregl-canvas')).toBeVisible({ timeout: 15_000 });
-    // Two independent assertions on purpose: the badge is React state, the
-    // camera fit is map state. The fix(#542) race kept the badge visible while
-    // the overlay and fitBounds silently never applied. The race itself is
+    // Two independent assertions on purpose: the preview row is React state,
+    // the camera fit is map state. The fix(#542) race kept the preview visible
+    // while the overlay and fitBounds silently never applied. The race is
     // timing-dependent (it needs the style transiently unloaded at pickup, so
     // a fast local run may not reproduce it — the deterministic regression
     // test lives in use-ephemeral-layers.test.ts); what this spec pins down is
@@ -225,10 +225,13 @@ test.describe('Dataset AI chat', () => {
     // fix(#894): #674 renamed this badge's copy "Query result" → "Result" (the
     // badge stopped being query-only once analysis previews reused it) and this
     // assertion was never updated, so the spec has failed every nightly since.
-    // Scope to role="status": EphemeralBadge renders the label twice — the
+    // Scope to role="status": the preview surface renders the label twice — the
     // sr-only live region plus an aria-hidden visible copy — so a bare getByText
     // matches two nodes and trips strict mode. The aria-hidden copy is out of
     // the accessibility tree, so the role lookup resolves to exactly one.
+    // fix(#1009): the surface under test is now the ephemeral stack row
+    // (EphemeralPreviewRow) rather than the corner badge — the assertion is
+    // unchanged because the row carries the same live-region sentence.
     await expect(
       page.getByRole('status').filter({ hasText: 'Result · 2 features' }),
     ).toBeVisible({ timeout: 15_000 });
--- a/frontend/src/components/builder/ChatPanel.tsx
+++ b/frontend/src/components/builder/ChatPanel.tsx
@@ -513,15 +513,20 @@ export function ChatPanel({
       if (minX < -180 || minY < -90 || maxX > 180 || maxY > 90) return;
       // fix(#527 B-054/C-06): inverted bounds also throw in fitBounds.
       if (minX > maxX || minY > maxY) return;
-      // fix(#674 audit): forward the truncation pair so EphemeralBadge can say
-      // "500 of 10,651 features". Both query_data and run_analysis emit them
-      // only when the server-side cap actually bit.
+      // fix(#674 audit): forward the truncation pair so the preview surface can
+      // say "500 of 10,651 features". Both query_data and run_analysis emit
+      // them only when the server-side cap actually bit.
+      // fix(#1009): that surface is the ephemeral stack row in this (builder)
+      // path — EphemeralPreviewRow, via UnifiedStackPanel's `preview` prop.
+      // The badge still consumes the same pair in the viewer and in the
+      // builder's <1100px rail layout.
       //
       // fix(#1076): the flag no longer waits for the total. A clip filters
       // rows, so the server reports no source total for it, and requiring one
       // dropped the disclosure entirely — a capped clip preview then read as
-      // complete. The total rides along when there is one; the badge picks its
-      // wording from whether it arrived.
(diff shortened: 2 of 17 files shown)


ishiland:
@codex review

PR #1167, opened 2026-08-03
fix(frontend): define the SAML providers query key in query-keys.ts

### LINKED ISSUE #1164: SamlProvidersSection uses a raw ['saml','providers'] key literal instead of a query-keys entry
## What

`frontend/src/components/admin/saml/SamlProvidersSection.tsx` reads and invalidates its own provider list under an inline array literal:

```ts
queryClient.invalidateQueries({ queryKey: ['saml', 'providers'] })
```

Every other query family in the app is defined once in `frontend/src/lib/query-keys.ts` and referenced through it. This one has no single definition, so the read key and the invalidation key are two independent literals that agree only by convention.

## Why it matters

This is the same class of defect #1117 just fixed, one step earlier in its lifecycle. #1117 was "two surfaces invalidate different key families and nobody notices"; this is "one surface's key exists in two places with nothing making them agree". A typo in either literal, or a future third call site that writes `['saml','provider']`, produces an invalidation that silently does nothing — and an inert invalidator is exactly the #1021 class, which took a dedicated sweep to find.

`lib/query-keys.ts` also has a parity test (`lib/__tests__/query-keys.test.ts`) asserting key shapes, so a key defined outside it is a key nothing tests.

## Fix

Add a `saml.providers` entry to `queryKeys` and point both the `useQuery` and the `invalidateQueries` at it. Touches the query-keys parity test, which is why it was left out of #1117's scope rather than folded in.

## Provenance

Found while implementing #1117 (PR #1163). Deliberately not fixed there: #1117's diff is about the OAuth/SAML → login-page pairing, and widening it to a key-definition refactor would have mixed two changes with different blast radii in one review.

ishiland:
Closes #1164.

`SamlProvidersSection` read its provider list under an inline `['saml', 'providers']` literal and invalidated it under three more, one per create/update/delete mutation. Four independent literals that agreed only by convention. A typo in any one of them gives an invalidation that silently does nothing, which is the #1021 class of defect that took a dedicated sweep to find last time.

The key is now defined once in `lib/query-keys.ts` and all four sites reference it. The key string itself is unchanged, so cached entries still match across a deploy.

## Why `saml` gets its own root

`listSamlProviders` fetches the same `/settings/oauth-providers/` endpoint as `queryKeys.settingsOAuth.providers` and filters the response down to the SAML rows. Same endpoint, different payload, so the two caches must not prefix-match. Nesting this under `settingsOAuth` would make `invalidateAuthProviders()` sweep the SAML list as a side effect. Fanning a provider mutation across both is what `useInvalidateAuthProviders` (#1117) already does, and this section calls it alongside the new key.

## One test beyond the issue's scope

The parity test in `lib/__tests__/query-keys.test.ts` is extended, and the invalidation spies in `SamlProvidersSection.test.tsx` now cover all three mutation paths.

There is also one test the issue did not ask for. Under the exact defect #1164 describes, a read key and an invalidation key drifting apart, the parity test and all three spy tests still pass. Only a test that lets the real invalidation run and watches for the resulting refetch catches it. That case is covered now.

## Verification

Four sabotage runs, each reverted afterwards:

| sabotage | failures |
|---|---|
| key factory to `['saml','provider']` | 1 (parity) |
| key factory to `['settings','oauth-providers','saml']` | 3 (parity, prefix-match, refetch) |
| delete the create-path `invalidateQueries` | 2 |
| read key only to `['saml','provider']` | 1, the refetch test alone |

Gates from `frontend/`:

- `npx vitest run`: 373 files, 4570 tests, all passing
- `npm run lint`: clean
- `npm run typecheck`: clean

No new `t()` keys, so locale parity is untouched. No API, schema, or config impact.

### DIFF
--- a/frontend/src/components/admin/saml/SamlProvidersSection.tsx
+++ b/frontend/src/components/admin/saml/SamlProvidersSection.tsx
@@ -122,7 +122,7 @@ export function SamlProvidersSection() {
     isLoading,
     isError,
   } = useQuery({
-    queryKey: ['saml', 'providers'] as const,
+    queryKey: queryKeys.saml.providers,
     queryFn: listSamlProviders,
   });
 
@@ -136,7 +136,7 @@ export function SamlProvidersSection() {
   const createMutation = useMutation({
     mutationFn: (data: SamlProviderCreateData) => createSamlProvider(data),
     onSuccess: () => {
-      queryClient.invalidateQueries({ queryKey: ['saml', 'providers'] });
+      queryClient.invalidateQueries({ queryKey: queryKeys.saml.providers });
       invalidateAuthProviders();
       toast.success(t('saml.created'));
     },
@@ -149,7 +149,7 @@ export function SamlProvidersSection() {
     mutationFn: ({ id, data }: { id: string; data: SamlProviderUpdateData }) =>
       updateSamlProvider(id, data),
     onSuccess: () => {
-      queryClient.invalidateQueries({ queryKey: ['saml', 'providers'] });
+      queryClient.invalidateQueries({ queryKey: queryKeys.saml.providers });
       invalidateAuthProviders();
       toast.success(t('saml.updated'));
     },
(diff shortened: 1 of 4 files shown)


ishiland:
@codex review

PR #1168, opened 2026-08-03
chore(ci): converge CI and local dev to Python 3.14

### LINKED ISSUE #1137: ci/docker: tests run on Python 3.13 while published images run 3.14.6
## The skew

Surfaced by codex review on PR #1136 (which added `.python-version` = 3.13 to match CI and had to keep it *out* of the Docker build):

- **CI** pins `python-version: "3.13"` via `setup-python` in four jobs (`ci.yml:327/359/746/777`), and local dev now matches via per-project `.python-version` (#1136).
- **Published images** build and run on `FROM python:3.14.6-slim` (`Dockerfile:9` builder, `:119` runtime base for api and worker).

So the interpreter that serves production is one the test suite never executes on. Everything that has shipped works — this is not a known breakage — but the gap is structural: a 3.14-only behavioural difference (or a 3.13-only one) is invisible to every gate until it fires in a container.

Concretely observed already, in the harmless direction: the `#1127` cli-round-trip investigation and this session's worktrees ran suites on 3.14.5 that CI runs on 3.13.13, with identical results. The skew has been benign so far, not absent.

## The decision to make

Converge, in one of two directions:

1. **Images down to 3.13** — matches CI and dev exactly; touches wheel availability for the GDAL/shapely/pyproj stack in the slim image and needs a full image smoke (ingest worker especially).
2. **CI/dev up to 3.14** — matches what prod already runs (and has been running fine); touches four CI pins, three `.python-version` files, and possibly `uv.lock` behaviour; `requires-python >=3.13` stays honest either way.

Option 2 has the evidence advantage (prod has effectively been the 3.14 canary for months); option 1 is the smaller conceptual change ("test what you ship" by moving the ship). Either is fine — the wrong state is the current split one.

Deliberately NOT bundled into #1136, which is docs/pins hygiene: changing the runtime interpreter of published images (or the tested interpreter of CI) is a real change wanting its own verification pass.

Refs #1136.

ishiland:
Closes #1137.

Published images build and run on `python:3.14.6-slim` while every gate ran on 3.13, so the interpreter serving production was one the test suite never executed. Nothing is known to be broken by that. The problem is structural: a 3.14-only behaviour difference stays invisible until it fires in a container.

Direction is option 2, as decided in the comment on #1137. Prod has been the 3.14 canary for months, so moving CI up tests something already known to work, whereas moving images down to 3.13 would put an untested interpreter into production to fix a testing problem and re-open wheel availability for the GDAL/shapely/pyproj stack.

## What moved

15 `setup-python` pins, not the four the issue body estimated: 12 in `ci.yml` and one each in `publish-cli.yml`, `publish-mcp.yml`, `publish-sdks.yml`. Plus the three per-project `.python-version` files (`backend/`, `cli/`, `mcp/`) added by #1136.

Pins stay minor-only (`3.14`, not `3.14.6`), matching what they replace. `setup-python` resolves `3.14` against whatever patch the runner ships, and a minor-only `.python-version` is still satisfied by the base image's 3.14.6 if the file ever escapes `.dockerignore`. An exact patch pin would not be, and would break on the next Dependabot base-image bump.

## What deliberately did not move

`verify-published.yml` keeps its `python:3.13-slim` containers. That job installs the published wheels the way a downstream consumer would, and it is now the only gate proving they work on an interpreter other than 3.14. Dropping it to 3.14 would delete that signal.

No runtime artifact changes. The `Dockerfile` stays on `python:3.14.6-slim` at both the builder and runtime stages. The `requires-python` floors stay put: `>=3.13` for the backend, `>=3.11` for cli and mcp. Nothing narrows what we claim to support.

`uv.lock` is unchanged in all three projects. Resolution is universal across `requires-python`, which this does not touch.

## The .dockerignore trap is unchanged

`**/.python-version` stays in `.dockerignore` with its `fix(#1136 review)` comment. Inside the image build uv must use the base image's system python. A pin reaching the build makes uv fetch a managed interpreter outside `/app`, and since the runtime stage copies only `/app`, the image builds green and dies at exec with a dangling `.venv/bin/python`. This change does not alter that, and the line was verified present.

## Verification

CI on this PR is the real gate: it runs the full backend suite, lint, and the layering ratchet on 3.14 for the first time. That is the change's own acceptance test, so a green run here is the evidence.

Checked locally against the branch:

- `Dockerfile` still `python:3.14.6-slim` at both stages
- `.dockerignore` still carries `**/.python-version`
- no `python-version:` pin below 3.14 remains anywhere in `.github/workflows/`
- `uv.lock` untouched in backend, cli, and mcp

## Impact

CI and local dev interpreter only. No schema, API, or config change. Contributors on a per-project `.python-version` will have uv fetch 3.14 on their next command.


ishiland:
@codex review

ishiland:
## Image-level verification (the acceptance criterion from the issue)

The issue asked for artifact evidence, not a green build: "context probe plus `readlink` of `/app/.venv/bin/python`". Both are now done.

**Context probe, run against this branch's worktree** (which carries the 3.14 pins), using a throwaway `COPY . /ctx` image:

```
--- .python-version files IN BUILD CONTEXT ---
--- (empty above = correctly excluded) ---
sanity, context is non-empty:
Dockerfile
backend
frontend
scripts
```

Zero `.python-version` files reach the build context, and the context is non-empty, so the emptiness is the `.dockerignore` rule working rather than a broken probe.

**Interpreter resolution inside the image:**

```
readlink -f /app/.venv/bin/python  ->  /usr/local/bin/python3.14
/app/.venv/bin/python -V           ->  Python 3.14.6
system python3                     ->  /usr/local/bin/python3.14  (Python 3.14.6)
/app/.venv/pyvenv.cfg              ->  home = /usr/local/bin
                                       version_info = 3.14.6
```

The venv uses the base image's system interpreter. No managed interpreter was fetched, which is the failure mode `.dockerignore:58` exists to prevent.

One caveat on how to read that second block, since it nearly produced a false alarm here. A `find / -name .python-version` inside the running dev container **does** return `/app/backend/.python-version` and `/app/cli/.python-version`. Those are not baked into the image: `docker inspect` shows `/app/backend` and `/app/cli` are bind mounts from the host checkout, so the `find` is reading host files through the mount. `/app/.venv` is not mounted, so the readlink evidence above is genuinely image-level. Anyone repeating this check on a dev stack should expect the same hit and not treat it as a leak.

## Local gates on 3.14.5

- full backend suite, `pytest -n 4`: `7006 passed, 89 skipped, 269 warnings in 658.10s (0:10:58)`, on `Python 3.14.5, pytest-9.1.1`
- `ruff check .`: All checks passed
- `ruff format --check .`: 883 files already formatted
- `pytest tests/test_layering.py -q`: 39 passed
- `make cli-test`: 282 passed, then 8 passed / 2 skipped, both halves reporting Python 3.14.5
- all 11 workflow files parse; ci.yml has 25 jobs, 12 carrying a setup-python step, and the set of distinct pinned versions across the parsed file is `['3.14']`
- `uv lock` run for real (not just `--check`) in backend, cli, and mcp: `git status` clean afterwards, so no resolution movement

## One 3.14-attributable signal worth recording

Across 7006 tests, exactly one warning is attributable to the interpreter change: a `DeprecationWarning` from third-party **slowapi** (`extension.py:720` and `:877`) that `asyncio.iscoroutinefunction` is deprecated and slated for removal in Python 3.16. Not actionable now, and upstream's call, but it is precisely the class of difference the 3.13/3.14 skew was hiding. Every other warning is interpreter-independent library noise (NumPy 2.5, Starlette, SQLAlchemy).

PR #1169, opened 2026-08-03
docs(ingest): correct ensure_geom_4326_gist_index no-op docstring

### LINKED ISSUE #1020: investigate(ingest): is ensure_geom_4326_gist_index's no-op case reachable, or is the docstring stale?
Decomposed from #893 (item 9b). This one is an **investigation**, not a fix: the answer decides whether the work is a one-line docstring edit or a real bug on four request paths. No milestone until it is answered. Verified against `main` at `9a788ba92` on 2026-07-31.

## The question

`ensure_geom_4326_gist_index` (`backend/app/processing/ingest/metadata.py:1583`) ends its docstring with:

```
No-op when the table has no geom_4326 column: a dataset can carry a
geometry_type while only exposing the native ``geom`` column (the
effective_srid=None staging path skips add_4326_column).
```

The function backs that up: it probes `information_schema.columns` for `geom_4326` and returns early when it is absent (`:1602-1610`).

That case was traced during #898 and appeared to be unreachable. `_resolve_effective_srid` never returns None, `tasks_common._finalize_ingest` passes `effective_srid=None` only for non-spatial datasets, and `service_create.py`'s DDL includes the column. So either the docstring describes a case that cannot happen, or there is a legacy path the #898 trace did not cover.

Both answers are actionable and they are very different sizes.

## If the docstring is stale

Delete the paragraph, or rewrite it to say the branch is defensive rather than describing a reachable state. Ten minutes.

## If a legacy path exists

Then a dataset can carry `geometry_type` set with no `geom_4326` column, and a bbox request against it would 500 on four paths, each of which assumes the column exists whenever `geometry_type` is set:

1. **Parquet export.** `backend/app/processing/export/parquet.py:228` puts `ST_AsBinary(geom_4326)` into the select list unconditionally for geometry datasets.
2. **Export pre-count.** `_count_selected_features` (`backend/app/processing/export/router.py:63`) appends `bbox_where_sql(bbox)` at `:123` when `bbox is not None and has_geometry` (`:101`).
3. **Feature reads.** `backend/app/modules/catalog/features/service.py:156` selects `ST_AsGeoJSON(geom_4326, 6)`, and the bbox predicates at `:175-183` filter on `geom_4326` directly.
4. **The ogr export path.** `bbox_where_sql` (`backend/app/processing/export/ogr.py:48-74`) builds `(geom_4326 && ... AND ST_Intersects(geom_4326, ...))` as a raw SQL fragment.

None of these probe for the column first. A 500 on all four is the failure mode, and it would be reported as "export is broken for this dataset" with nothing pointing at the missing column.

## How to answer it

Query the live dev catalog for the contradiction directly rather than re-reading the ingest code, since the code path trace is what is in doubt:

```sql
SELECT d.id, d.name, d.table_name
FROM datasets d
WHERE d.geometry_type IS NOT NULL
  AND NOT EXISTS (
    SELECT 1 FROM information_schema.columns c
    WHERE c.table_schema = 'data'
      AND c.table_name = d.table_name
      AND c.column_name = 'geom_4326'
  );
```

An empty result on a catalog with real ingest history is decent evidence for the stale-docstring answer. A non-empty result names the datasets and tells you which ingest path produced them.

Also worth checking: the reupload and manifest-apply paths, and anything that creates a `data.*` table outside `service_create.py`'s DDL, since those are where a divergent DDL would live if one does.

## Acceptance

- A recorded answer, either way, with the evidence that settled it.
- If stale: the docstring no longer describes an unreachable state as reachable.
- If reachable: a follow-up issue for the four call sites, and this one links to it.

ishiland:
Closes #1020.

`ensure_geom_4326_gist_index` returns early when the table has no `geom_4326` column, and its docstring described that as a state production reaches: a dataset carrying a `geometry_type` while only exposing the native `geom` column, via an `effective_srid=None` staging path. No such path exists. The investigation and its evidence are recorded [on the issue](https://github.com/geolens-io/geolens/issues/1020#issuecomment-5167600193).

Short version, from both directions:

The live dev catalog produces zero rows for the contradiction, with inverse checks confirming that is not a wrong-schema false negative (17 of 17 spatial datasets have the column; no `data.*` table has `geom` without `geom_4326`). That evidence is weaker than it looks on its own, because the catalog's spatial half is only geojson and one shapefile.

Source closes the rest. `_run_staging_pipeline` does carry the `effective_srid=None` skip, but its only production caller is `reupload_file`, whose `effective_srid` falls back to 4326 and is never None. Both call sites of this function reach it with the column present: `add_4326_column` has just added it, and `_apply_reupload_swap` gates its call on a `geometry_type` that only `extract_metadata` produces, and `extract_metadata` reads `geom_4326` for the extent on both its CTE and fallback paths.

`source_format='created'` was checked specifically, since it has two distinct DDL paths and the catalog has zero `created` rows. `create_empty_dataset` declares `geom_4326` inline in its `CREATE TABLE`; the layers module calls `add_4326_column` before `create_dataset`. Both add it.

The branch is unchanged and stays as a defense. Only the claim about it changes, in the docstring and in the mirroring comment on the test that covers it.

No follow-up issue for the four call sites the investigation flagged as at risk (parquet export, export pre-count, feature reads, ogr `bbox_where_sql`). Their assumption that `geom_4326` exists whenever `geometry_type` is set holds.

## Verification

- `uv run ruff check .`: All checks passed
- `uv run ruff format --check .`: 883 files already formatted
- `uv run pytest tests/test_layering.py -q`: 39 passed
- `uv run pytest tests/test_staging_pipeline.py -k EnsureGeom4326GistIndex`: 4 passed, 14 deselected

Docs and one test comment only. No schema, API, or config impact.

### DIFF
--- a/backend/app/processing/ingest/metadata.py
+++ b/backend/app/processing/ingest/metadata.py
@@ -1744,9 +1744,9 @@ async def ensure_geom_4326_gist_index(
     Called from both add_4326_column (staging load) and _apply_reupload_swap
     (post-swap belt-and-braces), so any re-ingest self-heals a missing index.
 
-    No-op when the table has no geom_4326 column: a dataset can carry a
-    geometry_type while only exposing the native ``geom`` column (the
-    effective_srid=None staging path skips add_4326_column).
+    The no-geom_4326 early return is defensive, not a reachable state (#1020):
+    add_4326_column has just added the column, and the swap call is gated on a
+    geometry_type that extract_metadata cannot report without reading geom_4326.
     """
     has_col = await session.execute(
         text(
--- a/backend/tests/test_staging_pipeline.py
+++ b/backend/tests/test_staging_pipeline.py
@@ -946,7 +946,7 @@ class TestEnsureGeom4326GistIndex:
 
         await ensure_geom_4326_gist_index(session, "my_table")
 
-        # effective_srid=None staging path: geometry_type set, no geom_4326.
+        # Defensive branch (#1020): column probe only, no index DDL.
         assert session.execute.await_count == 1
 
     @pytest.mark.asyncio


ishiland:
@codex review

PR #1171, opened 2026-08-03
test(analysis): pin tenant attribution on materialized analysis outputs

### LINKED ISSUE #1010: test(analysis): verify materialized outputs carry tenant attribution end to end

ishiland:
Closes #1010.

`Dataset.tenant_id` and `Record.tenant_id` are stamped by the `0018_tenant_insert_stamping` trigger, which reads the `app.current_tenant` GUC. The application deliberately never passes `tenant_id`. So the invariant worth testing is not "does the code set the field" but **does the analysis worker's connection carry that GUC at the moment it inserts**.

That survives on a detail which is easy to break by accident. `_materialize` commits the CTAS transaction immediately before registration, and `SET LOCAL` does not survive a commit, which is why the code re-issues its statement timeout by hand on the next line. The tenant GUC comes back only because the engine's `begin` listener re-issues `set_config` on every connection-level transaction. Nothing pinned that. A refactor onto a different engine, or a job that loses `current_tenant_var`, writes a NULL-tenant dataset that then 404s on its own tenant's tiles.

## What the test does

Six cases over the real `materialize_analysis` task entry point, so the `tenant_id` → `tenant_task` → `tenant_job_context` binding is exercised rather than bypassed.

Both catalog rows are asserted as stored column values, read over an autocommit superuser connection with no tenant hooks installed. That bypasses RLS and never sets the GUC, so what comes back is the stored column rather than a row some filter happened to admit. A lookup that merely succeeds would not prove the stamp.

Beyond the issue's list, one case asserts `event.contains(engine.sync_engine, "begin", _on_begin)` directly, with no DB and no fixtures. The issue names "a refactor that registers the hook on a different engine" as the thing nothing pins; the DB tests can only simulate that, and this closes it.

## Fixture decision: B

Recorded in full [on the issue](https://github.com/geolens-io/geolens/issues/1010#issuecomment-5168065851). `multi_tenant_harness.py` is byte-identical to main.

Short version: option A's `writer_session()` could not serve the consumer that motivates it, because `_materialize` never accepts a session and opens its own from `app.core.db.async_session`. What the test needs is a session factory patched into that module, which is a different shape. And a writer session is useless without a substrate, so A would have made the shared harness create and drop cluster-global roles in setup and teardown for all eleven files that reference it, none of which needs them. `conftest.py` already documents why that churn is hazardous across xdist workers.

## Two corrections to the issue's acceptance criteria

**The literal negative case cannot work as written.** `register_existing_table` resolves the tenant schema from `current_tenant_var`, so clearing it before registration makes `tenant_data_schema(None)` raise and the job dies before any insert. That proves the guard, not the stamp. The test instead clears the var only across the `begin` of the registration transaction, which is exactly the state a hook on the wrong engine produces. The literal reading is covered separately: a task with no `tenant_id` kwarg is refused by `tenant_task` with `RuntimeError` and nothing is written.

**"A NULL `tenant_id` and no error anywhere" is not what this path does.** There is a second guard. The dataset does insert with NULL, but the `ingest_jobs` completion update is then refused by a trigger (`Ingest job creator and dataset belong to different tenants`), `_materialize` catches it, and the registration is rolled back, so the dataset never commits. The analysis fails visibly. This is why the negative test reads its values from inside the registration transaction, since an outer query sees nothing.

The invariant is still worth pinning: that visible failure is a second line of defence another call path may not have, and an opaque privilege error is a poor way to discover the begin-hook moved.

## Non-vacuity

Two sabotage runs, both reverted before committing.

Removing the GUC clearing from the negative case (so registration runs with the tenant bound) fails that test with the assertion firing on a real UUID where it expects `None`:

```
AssertionError: expected an unstamped dataset with the GUC absent, got
UUID('f3ae90f3-62c1-4417-b309-7ad41bea0c71') — the positive assertion would be vacuous
1 failed in 3.11s
```

Replacing `install_tenant_session_hook(engine)` with a no-op, which is the refactor the issue names, fails the positive case and the tile case: `2 failed, 2 passed in 9.50s`.

## Verification

Branch is rebased onto `07c43c1c6`, so these ran on the Python 3.14 that #1137 just converged CI onto (the original run was on 3.13.13, before that landed).

- new test file: `6 passed in 20.88s` on Python 3.14.5
- **whole-tree backend suite after the rebase**, `pytest -n 4` on 3.14.5: `7012 passed, 89 skipped, 270 warnings in 634.37s (0:10:34)`. That covers `test_gate01_cross_tenant_isolation.py`, `test_iso_single_tenant_byte_identical.py`, and the nine modules importing the harness. The count also reconciles: #1137's pre-merge run measured 7006 on 3.14 without this file, and 7006 + 6 new tests is 7012, so the interpreter change and this change compose rather than interacting.
- `ruff check .`: All checks passed · `ruff format --check .`: 884 files already formatted
- `pytest tests/test_layering.py -q`: 39 passed. No lines added under `backend/app/`, so no `_MODULE_LOC_CAPS` change.

The pre-rebase run on 3.13.13 gave the same `7012 passed, 89 skipped`, so nothing here is interpreter-sensitive.

`conftest.py` gains three lines: one entry in `_TENANCY_GLOBAL_STATE_MODULES`. The module flips process-global tenancy mode and creates and drops cluster-global per-tenant roles, so it belongs to both documented categories in that set. Without the entry, xdist could schedule it concurrently with `test_analysis_materialize` and hit the `pg_shdepend` contention that block exists to prevent.

## One thing a reviewer will see

Host coverage reports **77.69% against the 80% floor, and this change does not move it**. Measured both ways on the same command: with the new module, `TOTAL 34116 7610 78%`; with `--ignore` on it, byte-identical. `[tool.coverage.run] source = ["app"]` makes the denominator the whole `app/` tree regardless of imports, so a test-only addition cannot move it. The shortfall is the 89 host-skipped tests (ogr2ogr, live_stack, overlay-gated). `make test-cov` in-container is where that gate means anything.

Tests only. No schema, API, or config impact.

### DIFF
--- a/backend/tests/conftest.py
+++ b/backend/tests/conftest.py
@@ -208,6 +208,9 @@ _TENANCY_GLOBAL_STATE_MODULES = {
     "test_tenant_rls_migration",
     "test_tenant_job_recovery",
     "test_tenant_session_guc",
+    # #1010: multi_tenant harness (1) PLUS per-tenant reader/writer role
+    # CREATE/DROP (2) — it belongs to both categories below.
+    "test_tenant_attribution_materialize",
     # (2) cluster-global geolens_reader role GRANT/REVOKE mutators — must not run
     #     concurrently with each other or the group above (pg_shdepend contention).
     "test_analysis_materialize",  # register_existing_table → grant_reader_access
--- a/backend/tests/test_tenant_attribution_materialize.py
+++ b/backend/tests/test_tenant_attribution_materialize.py
@@ -137,6 +137,53 @@ async def _deprovision_substrate(conn, tenant_id: str) -> None:
     await conn.execute(sa.text(f"DROP ROLE IF EXISTS {writer}"))
 
 
+# The tenant boundary, minus ``users`` (the harness deletes its own two seeded
+# rows).  Deleting by tenant is what makes this safe on a shared per-worker DB:
+# the harness mints both tenant UUIDs per test, so no row carrying one of them
+# belongs to anything else.  Listing the whole boundary rather than the three
+# tables this module happens to write today means a case added later that
+# creates a map or a collection is cleaned up too.
+#
+# Order is FK-derived, not alphabetical:
+#   - ``ingest_jobs.dataset_id -> datasets`` is ON DELETE SET NULL, so a job row
+#     SURVIVES a dataset delete with a nulled reference. It has to go first and
+#     explicitly. (Everything else hanging off records/datasets is CASCADE.)
+#   - ``maps``/``collections``/``embed_tokens`` reference datasets, so they
+#     precede it.
+#   - ``records`` last of the record graph: its CASCADEs take record_keywords
+#     (the provenance INSERT..SELECT writes there), record_translations,
+#     record_contacts, record_distributions, record_embeddings and
+#     dataset_relationships with it.
+_TENANT_SCOPED_CATALOG_TABLES = (
+    "ingest_jobs",
+    "embed_tokens",
+    "maps",
+    "collections",
+    "datasets",
+    "records",
+    "audit_logs",
+    "oauth_accounts",
+)
+
+
+async def _delete_tenant_catalog_rows(conn, tenant_ids: list[str]) -> None:
+    """Delete every committed catalog row belonging to *tenant_ids*.
+
+    ``_seed_source_dataset`` and ``_create_job`` commit before the worker even
+    starts, a successful materialize commits a second record/dataset pair plus
+    its provenance, and the missing-context case deliberately leaves a PENDING
+    job. The per-worker test DB is shared, so anything left behind outlives the
+    test: rows naming a table whose schema this fixture is about to drop, and a
+    pending job that later quota or stale-sweep tests would count.
+    """
+    for table in _TENANT_SCOPED_CATALOG_TABLES:
+        # Table name comes from the module-level tuple above, never from data.
+        await conn.execute(
+            sa.text(f"DELETE FROM catalog.{table} WHERE tenant_id = ANY(:tenant_ids)"),
+            {"tenant_ids": tenant_ids},
+        )
+
+
 @asynccontextmanager
 async def _admin_connection(db_url: str):
     """An AUTOCOMMIT superuser connection with NO tenant hooks installed.
@@ -220,6 +267,9 @@ async def tenant_analysis(multi_tenant_rls, monkeypatch):
             _dataset_cache.clear()
         await engine.dispose()
         async with _admin_connection(ctx.db_url) as conn:
+            # Committed rows first: a catalog row naming a table whose schema
+            # the next statement drops is worse than either alone.
+            await _delete_tenant_catalog_rows(conn, [ctx.tenant_a, ctx.tenant_b])
             await _deprovision_substrate(conn, ctx.tenant_a)
             await _deprovision_substrate(conn, ctx.tenant_b)
 
@@ -613,12 +663,14 @@ class TestSingleTenantUnchanged:
 
 
 def test_shared_engine_carries_the_tenant_begin_hook():
-    """The GUC re-issue is registered on the engine the worker actually uses.
+    """``app/core/db/session.py`` registers the GUC re-issue on its engine.
 
-    ``_materialize`` opens its session from ``app.core.db.async_session``,
-    which is built on ``app.core.db.session.engine``.  Installing the hook on
-    any other engine is the silent refactor the tests above simulate; this
-    asserts the production registration is where it has to be.
+    Half of the wiring: the hook exists on the engine that module builds. The
+    other half — that the worker's session factory is bound to THAT engine —
+    is the test below. Both are kept because they fail for different reasons
+    and the distinction is the whole diagnosis: a passing engine assertion
+    beside a failing factory assertion says the hook is installed but on the
+    wrong object.
     """
     from sqlalchemy import event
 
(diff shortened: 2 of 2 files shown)


ishiland:
@codex review

ishiland:
@codex review

PR #1172, opened 2026-08-04
refactor(datasets): delete unused update_auto_metadata

### LINKED ISSUE #1170: hygiene(catalog): update_auto_metadata sets geometry_type with no geom_4326 probe and has zero callers
## What

`update_auto_metadata` (`backend/app/modules/catalog/datasets/domain/service_metadata.py`) assigns `dataset.geometry_type` without probing the underlying `data.*` table for a `geom_4326` column.

It currently has **zero callers** across `backend/`, `cli/`, and `mcp/`. The only references are the façade re-export in `datasets/domain/service.py` and its own `__all__`.

## Why it is worth recording rather than ignoring

Found while settling #1020, which asked whether a dataset can carry a `geometry_type` while its table lacks `geom_4326`. The answer there was no, and the reason is load-bearing: every path that sets `geometry_type` sources it from `extract_metadata`, which reads `ST_Extent(geom_4326)` on both its CTE and fallback paths and therefore raises rather than reporting a `geometry_type` for a table without the column. The two paths that bypass `extract_metadata` (`create_empty_dataset` and the layers module) both add the column before writing the catalog row.

`update_auto_metadata` is the one writer that sits outside that argument. It cannot produce the contradiction today because nothing calls it, but wiring it up to any caller would reopen #1020's failure mode: a dataset with `geometry_type` set and no `geom_4326`, which 500s on four request paths (parquet export, export pre-count, feature reads, and the ogr `bbox_where_sql` fragment), none of which probe for the column.

So the risk is not current behaviour, it is that #1020's conclusion silently stops holding the moment someone gives this function a caller, with nothing in the tree to catch it.

## Options

Roughly in order of laziness:

1. **Delete it.** Dead code with a façade re-export to remove alongside. Smallest diff, and it removes the way to be wrong rather than documenting it.
2. **Keep it and probe.** Have it verify `geom_4326` exists before setting `geometry_type`, so a future caller cannot manufacture the state.
3. **Keep it and comment.** Anchor a note to #1020 explaining the invariant it must not break. Cheapest, and the weakest, since a comment does not fail a build.

Option 1 looks right unless there is a planned caller, which is the part I do not know.

## Acceptance

- A decision recorded here, and either the function removed or the invariant enforced.
- If it is kept, #1020's conclusion is restated somewhere that fails when violated rather than in prose alone.

Refs #1020.

ishiland:
## Problem

`update_auto_metadata` (`service_metadata.py`) sets `geometry_type` with no `geom_4326` probe and has zero callers anywhere — backend, CLI, MCP, SDKs, frontend, or `openapi.json` (evidence on the issue). It is the one `geometry_type` writer that bypasses the impossibility argument recorded in #1020, purely dead code.

## Fix

Delete the function and its `__all__` entry, remove both façade lines in `domain/service.py` (import and `__all__`), and lower the `service_metadata.py` ceiling in `test_layering.py` 490 → 453 (exact new count) with a `fix(#1170)` comment.

Deletion over enforcement: a probe would be machinery guarding a function nobody invokes.

## Verification

- `uv run ruff check .` and `uv run ruff format --check .` clean
- `uv run pytest tests/test_layering.py -q` — 39 passed
- Full suite on the tree: 7013 passed, 89 skipped
- Repo-wide grep for `update_auto_metadata` matches only the cap comment

No API/schema impact (never exposed via any endpoint).

Closes #1170

https://claude.ai/code/session_01PnJqoYvwnHaFT3w4E9zNvE

### DIFF
--- a/backend/app/modules/catalog/datasets/domain/service.py
+++ b/backend/app/modules/catalog/datasets/domain/service.py
@@ -51,7 +51,6 @@ from app.modules.catalog.datasets.domain.service_metadata import (
     list_attributes,
     reset_attribute,
     update_attribute,
-    update_auto_metadata,
     update_user_metadata,
 )
 from app.modules.catalog.datasets.domain.service_query import (
@@ -99,6 +98,5 @@ __all__ = [
     "resolve_source_feature_count",
     "run_analysis_preview",
     "update_attribute",
-    "update_auto_metadata",
     "update_user_metadata",
 ]
--- a/backend/app/modules/catalog/datasets/domain/service_metadata.py
+++ b/backend/app/modules/catalog/datasets/domain/service_metadata.py
@@ -33,7 +33,6 @@ __all__ = [
     "list_attributes",
     "reset_attribute",
     "update_attribute",
-    "update_auto_metadata",
     "update_user_metadata",
 ]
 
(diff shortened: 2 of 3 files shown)


ishiland:
@codex review

ishiland:
@codex review

PR #1176, opened 2026-08-04
docs(ingest): record the registered-table linear-geometry contract

### LINKED ISSUE #1114: decide: linearity of geom_4326 on registered tables written externally after registration
Follow-up from #1113's review (r6). #1104 makes `geom_4326` always linear at every app-controlled write boundary: ingest (`add_4326_column`), the 0034 backfill, and registration (`linearize_existing_4326`). One writer remains outside the app's reach: a registered table's owner writing directly to it afterward. The register UI contract (`frontend/src/components/import/WorkflowRail.tsx`) is explicit that registration copies no data and serves tiles from the live table, so post-registration external writes are a supported shape, and a curved row written that way reaches the readers unguarded.

## Why this was split out rather than fixed in #1113

- **The exposure predates #1104.** Before it, a curved row written directly to a registered table already broke vector tiles and feature reads (two of the four #1104 surfaces) — the per-read `ST_CurveToLine` wraps only covered the analysis SQL. The post-#1104 delta is analysis-only, on a row shape that already broke the primary surfaces.
- **Closing it properly is new machinery, not a rider**: a linearity trigger installed at registration, backfilled onto already-registered tables by migration, and removed at unregistration — or an explicit operator contract that direct writes must keep `geom_4326` linear.

## Options

1. **Trigger at registration** (structural): `CREATE TRIGGER ... BEFORE INSERT OR UPDATE OF geom_4326 ... EXECUTE linearize_row()`. Complete, but adds trigger lifecycle management (register/unregister/migration) and a per-row write cost on tables GeoLens does not own.
2. **Documented operator contract** (cheap): the register flow's docs state that `geom_4326` must stay linear; a curved row degrades tiles/reads/analysis for that dataset only. Matches how other invariants on registered tables are handled (e.g. nobody re-validates SRID on every external write).
3. **Both, edition-split**: contract in OSS, trigger in deployments that want it.

Option 2 is probably proportionate — the same external writer can already break the table in a dozen other ways (drop the column, change the SRID), and GeoLens does not currently police any of them post-registration. Deciding that is what this issue is for.

## Acceptance (whichever option)

- A curved row written directly to a registered table after registration either cannot persist (option 1) or is a documented, dataset-scoped degradation (option 2).
- `test_analysis_curved_sources.py`'s register tests keep passing unmodified.

ishiland:
## Problem

#1114's decision (2026-08-02, option 2: documented contract) left one in-repo residual: nothing in the source states the contract for externally-written curved geometries on registered tables. Operators who violate it do so from psql, not the import wizard, so the contract note belongs on `register_existing_table`, not in UI copy.

## Fix

A 12-line `fix(#1114)` paragraph in the `register_existing_table` docstring (`backend/app/processing/ingest/service.py`): `geom_4326` must stay linear; GeoLens linearizes once at registration (`linearize_existing_4326`) and does not police the table afterward; a curved row written directly degrades tiles, feature reads, and analysis for that dataset only; STORED GENERATED columns fall under the same contract.

`_MODULE_LOC_CAPS` bumped 1032 → 1044 in the same commit (the cap is exact both directions).

Docs-site half filed separately as geolens-io/getgeolens.com#84.

## Verification

- `uv run ruff check .` / `uv run ruff format --check .` clean
- `uv run pytest tests/test_layering.py -q` — 39 passed (re-run after rebase onto post-#1170 main)
- The five register tests in `test_analysis_curved_sources.py` untouched

Docs-only; no API/schema impact.

Closes #1114

### DIFF
--- a/backend/app/processing/ingest/service.py
+++ b/backend/app/processing/ingest/service.py
@@ -526,6 +526,18 @@ async def register_existing_table(
     Verifies the table exists, checks for duplicate registration,
     ensures geom_4326 column and reader access, extracts metadata,
     and creates a Dataset record.
+
+    fix(#1114): registered-table linear-geometry contract. ``geom_4326``
+    on a registered table must stay linear -- curved geometry types
+    (CIRCULARSTRING, COMPOUNDCURVE, CURVEPOLYGON, MULTICURVE,
+    MULTISURFACE) are not supported. GeoLens linearizes the column once
+    at registration (``linearize_existing_4326`` below) and does not
+    police the table afterward: registration copies no data and serves
+    from the live table, so the owner keeps writing to it directly. A
+    curved row written that way degrades vector tiles, feature reads,
+    and analysis for that dataset only; other datasets are unaffected.
+    STORED GENERATED ``geom_4326`` columns fall under the same
+    contract -- their generation expression must produce linear output.
     """
     table_name = request.table_name
 
--- a/backend/tests/test_layering.py
+++ b/backend/tests/test_layering.py
@@ -1892,7 +1892,10 @@ _MODULE_LOC_CAPS: dict[str, int] = {
     # pre-existing geom_4326 (savepoint + error contract mirroring the
     # add_4326_column branch beside it); see linearize_existing_4326.
     # Cap 1017 -> 1032, still exact.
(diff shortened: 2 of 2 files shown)


ishiland:
@codex review

Issue #651, opened 2026-07-23
Metrics stack reports per-worker registries only — adopt prometheus multiprocess mode


ishiland:
**Priority: P3** — surfaced by codex on #650; a pre-existing, repo-wide property of the metrics stack, split out so it can be decided deliberately.

## Problem

The api runs multiple uvicorn workers (`UVICORN_WORKERS=2` on the demo), and every metric — the HTTP histograms from `prometheus_fastapi_instrumentator`, the pool gauges in `observability/metrics/pool.py`, the jobs metrics, and the per-worker RSS gauge from #650 — lives in each worker's own default `prometheus_client` registry. `/metrics` is answered by whichever worker accepts the scrape, so a scrape sees exactly one worker's values; counters and histograms undercount roughly by the worker count, and gauges flip-flop between workers.

#650 mitigates for the RSS gauge specifically by labeling it with `pid` (`backend/app/observability/metrics/memory.py:100`, whose module docstring cites this issue), so successive scrapes accumulate distinct per-worker series rather than silently overwriting one, and by making the structured-log heartbeat/watermark the authoritative signal. That is a mitigation, not a fix.

**Scope of the symptom.** Nothing in the repo sets `PROMETHEUS_MULTIPROC_DIR` today; a repo-wide grep returns zero matches, so every deployment runs single-registry-per-worker. The undercount only manifests at `UVICORN_WORKERS > 1`, which is the prod default of 2 (`docker-compose.prod.yml:272`) and not the dev default of 1 (`docker-compose.yml:341`). A developer reading `/metrics` locally sees correct numbers and has no way to notice the problem, which is part of why it went unnoticed.

## Proper fix

Adopt `prometheus_client` **multiprocess mode**: set `PROMETHEUS_MULTIPROC_DIR` in the container env, serve `/metrics` from a `MultiProcessCollector` registry, declare gauges with an explicit `multiprocess_mode` (`'all'`/`'livesum'` as appropriate per metric), and clean stale pid files on boot. This touches the semantics of every existing metric, so it should ship as its own change.

Two implementation details that decide whether it works in this stack:

[...]

PR #1177, opened 2026-08-04
docs(runbook): add API worker memory recycling guidance

ishiland:
## Problem

RUNBOOK.md had zero mentions of uvicorn, max-requests, or recycling, even though `UVICORN_MAX_REQUESTS` is the shipped backstop for the #643 signature (worker RSS growth toward the container cap ending in a silent OOM kill).

## Fix

New "API worker memory & recycling" section in RUNBOOK §4 (Monitoring), between the alert-rules table and the temp-file ceiling:

- The `geolens_worker_rss_bytes` gauge and the two structured-log markers from #650 ("API worker memory" baseline/heartbeat, "API worker memory above watermark" at 60% of cgroup limit) as the per-worker signal — verified against `backend/app/observability/metrics/memory.py`. Notes that one `/metrics` scrape reports one worker (per-process registries, aggregation tracked in #651), so the logs stay authoritative.
- What `UVICORN_MAX_REQUESTS` does and when to reach for it, qualified per review: production-only (the dev compose overrides the CMD with `--reload` and no `--limit-max-requests`), in-place respawn only under `workers>1` (single-worker deployments rely on the container restart policy), and disabling under production Compose requires editing the compose value because `${UVICORN_MAX_REQUESTS:-10000}` maps an empty value to 10000.
- That last fact also corrects `.env.example`'s "leave empty to disable" rule at its source — blanking disables recycling only for direct image runs.

This is the bounded documentation slice; the root-cause investigation stays open. No `scripts/*.sh` paths named (env-doc-check rule 6).

## Verification

- `make env-doc-check` — "env-doc-check OK: 135 documented key(s) … aligned" (re-run after every revision)

Docs-only; no code, no API/schema impact.

Refs #643

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -257,7 +257,10 @@ GEOLENS_ADMIN_PASSWORD=
 
 # fix(#643): recycle each API worker after this many requests — a graceful exit +
 # supervisor respawn, so slow memory growth can't ride a worker into the container
-# memory cap. Leave empty to disable; must be a positive integer (0 is invalid).
+# memory cap. Must be a positive integer (0 is invalid). Leaving it empty
+# disables recycling only for direct image runs; docker-compose.prod.yml maps
+# an unset or empty value to its own default of 10000, so disabling it there
+# means editing the compose value.
 # Type: integer (minimum 1) or empty | Default: 10000 (prod compose)
 # UVICORN_MAX_REQUESTS=10000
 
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

Issue #1070, opened 2026-07-31
feat(catalog): tell an owner which keywords a derived dataset inherited before they publish it


ishiland:
## Summary

When an analysis output is materialized, #1045 copies the source record's keyword rows onto the new record (`backend/app/processing/analysis/provenance.py`, `INSERT INTO catalog.record_keywords ... SELECT`). That is the intended catalog behaviour: a derived product is still about its source's subject, and STAC/DCAT both treat subject terms as inheritable.

The output registers private (`register_existing_table(..., visibility="private")`, `backend/app/processing/analysis/tasks.py:770`), so nothing reaches a third party until the owner publishes. The owner already had read access to the source, and keywords are editable record metadata they can strip.

What is missing is that the owner is never told the inheritance happened.

## The gap

An owner who buffers a private dataset and later publishes the result has no point in the flow where the UI says the output carries keywords it did not author. If a source keyword is a project codename, a client name, or a site identifier, publishing discloses it to everyone with no prompt and no diff.

The reference is already gated for this exact case: `visible_derived_from` omits `derived_from` for a requester who cannot open the source, and #1045 tests it with a private source and a published output. Keywords take the opposite path by design, because they are values rather than an actionable id. The design is defensible. The silence is not.

## Shape of a fix

Surface inherited metadata at the moment it becomes consequential rather than at the moment it is copied:

- Mark inherited keyword rows so the origin is recoverable, or derive it from `derived_from` at read time.
- In the publish/visibility-change flow, show which keywords came from a source the audience cannot open, and let the owner drop them in place.

Marking rows means a migration, so deriving from `derived_from` is worth costing first.

## Not a security fix

Filing this as product work, not a vulnerability. Exposure requires the owner's deliberate publish, the owner can already edit the keywords, and no requester gains an id they can act on. The improvement is informed consent at publish time.

Raised during review of #1045. Related: #1068 (the other deliberately-deferred finding from this wave), #765.

ishiland:
Route decision for the implementation: derive the inherited keyword set at read time, no migration.

- `Record.derived_from` already carries the source dataset id (`build_derived_from`, `backend/app/processing/analysis/provenance.py`), and `apply_analysis_provenance` copies source keywords at materialize time — so the inherited set is recoverable by resolving the source record and intersecting keyword sets. One extra query on the keywords read path, no schema change.
[...]

PR #1184, opened 2026-08-04
fix(records): honor the keyword counterfactual only for the record owner

ishiland:
Found by the 2026-08-04 security audit of the #1070 surface (`docs-internal/audits/sec-audit-20260804.md`, finding S01, Low / CVSS 3.1).

## Problem

`list_keywords_endpoint` honored `audience_visibility` and `audience_record_status` from **any** requester who cleared the read gate — which admits anonymous. Those params answer one question, the counterfactual an owner asks before widening access, and their only caller is `AccessTab`, owner-only UI.

With `?audience_visibility=private&audience_record_status=published`, `record_audience` collapses the derived audience to "admin OR the output's owner". `audience_exceeds_source` then looks for a user in the derived audience but not the source's: admins satisfy both sides and cancel, every non-owner fails the first clause. The returned `inherited_audience_gap` therefore reduces to **"this output's owner does not have access to the source dataset"** — a fact about a named third party, since `created_by` is already on the dataset response. Grant lists are not otherwise readable by non-owners.

Live case is narrow: a **restricted + published** source, where a grant holder reading someone else's derived output learns whether that output's owner also holds a grant. A private or unpublished source means only the owner/admin clears the outer `visible_derived_from` gate; a public or internal source makes the gap always false.

## Fix

The overrides reach `audience_exceeds_source` only when `_is_record_owner_or_admin(db, record, user)` holds; otherwise the gap is computed at the record's **stored state**. No 403, response shape unchanged, `AccessTab` unaffected — a non-owner still gets a correct answer, just not a hypothetical one.

`_is_record_owner_or_admin` is a new non-raising helper mirroring the two places the codebase already writes that predicate — `_check_record_ownership` (same file) and `check_dataset_write_access` (`catalog/authorization.py`): `created_by is not None and created_by == user.id`, else the `admin` role. Both existing helpers raise, which a read endpoint cannot do.

Also documents on `audience_record_status` why it is deliberately **not** pinned to an enum: the lifecycle set comes from the workflow extension's `status_order()`, so a literal pattern would 422 a valid status in an overlay deployment. (An earlier revision of this branch pinned one and would have rejected `internal`.)

## Tests

`test_counterfactual_ignored_for_non_owner_with_source_access` adds the actor the file did not cover — a non-owner **with** source access — building its own role + grant rather than borrowing a fixture:

- attacker, with and without the params: **identical** `inherited_audience_gap`, so the response carries no information about the owner's grant (with a precondition asserting the inherited flags really resolve, so a redaction cannot masquerade as the gate)
- owner control: the same params still flip the answer — the feature is intact, only the gate changed
- admission control: owner with `audience_record_status=internal` is honored, not merely accepted

Negative control run twice: forcing the gate open makes the new test fail.

## Verification

`ruff check` / `format --check` clean · full backend suite 7032 passed, 89 skipped · targeted re-run at the committed tree 73 passed (`test_inherited_keywords` + `test_layering` + `test_workflow_extension` + `test_rule1_structural`) · `make openapi-check`, `make sdks-check`, frontend `npm run types:check`, `make cli-test` all pass.

Regen touched four artifacts (openapi.json, both SDKs, `frontend/src/types/api.generated.ts`) — description text only; `git diff backend/openapi.json | grep -E '^[-+].*(maxLength|pattern)'` is empty, so no constraint or field changed.

Refs #1070

### DIFF
--- a/backend/app/modules/catalog/records/router.py
+++ b/backend/app/modules/catalog/records/router.py
@@ -199,6 +199,23 @@ async def _check_record_read_access(
     )
 
 
+async def _is_record_owner_or_admin(
+    db: AsyncSession, record: Record, user: Identity | None
+) -> bool:
+    """Return whether ``user`` owns ``record`` or is a global admin.
+
+    The non-raising twin of the predicate `_check_record_ownership` enforces
+    (and the same notion `check_dataset_write_access` uses on the dataset side):
+    creator match on `created_by`, else the `admin` role. Read endpoints that
+    need to branch on ownership without rejecting the request use this.
+    """
+    if user is None:
+        return False
+    if record.created_by is not None and record.created_by == user.id:
+        return True
+    return "admin" in await get_user_roles(db, user)
+
+
 async def _check_record_ownership(
     db: AsyncSession, record_id: uuid.UUID, user: Identity
 ) -> Record:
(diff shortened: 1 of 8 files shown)


ishiland:
@codex review

PR #1194, opened 2026-08-04
docs(schemas): describe record_status as extension-defined, not a closed set

### LINKED ISSUE #1183: docs(schemas): record_status Field descriptions omit 'internal' and already caused a wrong implementation
Found 2026-08-04 while reviewing the #1070 security fix.

## The defect

Two `Field(description=...)` strings describe the record lifecycle as a **three**-value set:

- `backend/app/modules/catalog/datasets/domain/schemas.py:338` — `description="Lifecycle status: draft, ready, published"`
- `backend/app/modules/catalog/datasets/domain/schemas.py:520` — same string

The authoritative set has **four** values:

```
backend/app/platform/extensions/defaults_extensions.py:245
DEFAULT_STATUS_ORDER = ("draft", "ready", "internal", "published")
```

`internal` is missing from both descriptions.

## Why it matters more than a typo

1. **These descriptions ship.** They flow into `backend/openapi.json`, both generated SDKs, and `frontend/src/types/api.generated.ts`. Every API consumer — including anyone reading the published docs or SDK docstrings — is told the lifecycle has three states.
2. **It already caused a wrong implementation today.** While fixing a security finding on `GET /records/{id}/keywords/`, an agent needed the valid `record_status` values, read these descriptions rather than `DEFAULT_STATUS_ORDER`, and pinned a request-validation pattern to `^(draft|ready|published)$`. That would have made `?audience_record_status=internal` return 422 for a legitimate transition the feature exists to warn about. Caught in review; the pattern was reverted.

## Fix

Update both descriptions to name all four states. Prefer wording that does not re-hardcode the list as a contract — the set comes from the workflow extension's `status_order()` and an overlay may define its own, so something like "Lifecycle status (community default order: draft, ready, internal, published; a workflow extension may define its own)" is more honest than a bare enumeration.

Then regenerate the four artifacts (`make openapi && make sdks`, `cd frontend && npm run types:generate`) — description-only edits still produce diffs in all of them.

## Sweep

Worth a mechanical grep for other places the status set is written out by hand rather than derived, in both source and docs. Any hardcoded enumeration of an extension-defined set is the same latent trap.

ishiland:
## Problem

Two `Field(description=...)` strings described the record lifecycle as a three-value set — `"Lifecycle status: draft, ready, published"` — omitting `internal`. The authoritative order is `DEFAULT_STATUS_ORDER = ("draft", "ready", "internal", "published")`.

These descriptions ship. They flow into `backend/openapi.json`, both generated SDKs, and `frontend/src/types/api.generated.ts`, so every API consumer reading the published docs or SDK docstrings was told the lifecycle has three states.

It already caused a wrong implementation: while fixing a security finding on `GET /records/{id}/keywords/`, the valid `record_status` values were read off these descriptions rather than `DEFAULT_STATUS_ORDER`, and a request validator was pinned to `^(draft|ready|published)$`. That would have 422'd `?audience_record_status=internal` — a legitimate transition the feature exists to warn about. Caught in review on #1184 and reverted.

## Fix

The set is not closed, so the wording no longer presents a list as the contract. Both descriptions now name the seam first and the default second:

> Lifecycle status. Deliberately not pinned to an enum: the values come from the workflow extension's status_order(), so an overlay may define its own. Community default order: draft, ready, internal, published.

This matches the wording precedent already in `records/router.py` from #1070, so the two shipped descriptions of this set now read the same way.

### Why "extension-defined" is the accurate framing

- `status_order()` is a method on the `WorkflowExtension` Protocol; the `"workflow"` registry slot is replaceable by an enterprise overlay, and `DefaultWorkflowExtension` is only the community implementation.
- The decisive evidence is a schema fact: `visibility` is backed by the `chk_records_visibility` CHECK constraint, and **`record_status` has no CHECK constraint at all** — it is a bare `String(20)`.
- `backend/tests/test_workflow_extension.py` already registers an overlay returning `("draft", "review", "published")` and asserts it wins, so the open set is a tested property rather than a theoretical one.

## Sweep

The issue asked for a mechanical sweep of other places the set is written out by hand rather than derived. Three additional instances were in scope and are fixed:

- `router_data.py` — two FastAPI docstrings that ship as OpenAPI **operation** descriptions. `PATCH /datasets/{id}/status/` stated "Allowed transitions: draft -> ready -> internal -> published" as the rule, while the code one screen down calls `workflow.allowed_transitions(context)`.
- `frontend/src/types/api.ts` — `RecordStatus` was typed `'draft' | 'published'`. **Two** values, worse than the three-value description this issue was filed about, and a type rather than a doc: it made `ready` and `internal` a compile error to handle. Now `string`, with a comment recording why.

`DatasetVisibility` deliberately stays a union in the same file, because it does have a CHECK constraint behind it; the mirror comment now names it.

Behaviour is unchanged. `getRecordStatusLabel` already accepted `string` and falls back to `humanizeToken` for unrecognised values, so the runtime was open-set-tolerant already and only the type disagreed. The sole literal comparison, `record_status === 'published'`, is identical under `string`.

Deliberately left: `protocols.py` and `defaults_extensions.py` (already honest / the authority itself), comments describing one concrete path rather than enumerating the set, and test assertions naming all four values (correct — they pin the community default).

Out of scope, worth a follow-up: `processing/ingest/manifest_schemas.py` and `manifest_sources.py` hold a closed `Literal[...]` over this open set, mirrored by the CLI's `geolens-manifest-v1.schema.json` `enum` and pinned by `cli/tests/test_manifest_examples.py`. Widening it is a manifest-schema-version decision, not a docs fix. Consequence today: an overlay defining `review` cannot be expressed in a manifest.

## Regenerated artifacts

A description-only edit is still a full regen. `backend/openapi.json`, both SDKs, and `frontend/src/types/api.generated.ts` are regenerated and committed here; their diffs are pure description propagation.

`backend/tests/test_layering.py` raises the `schemas.py` cap 1156 -> 1167, with a comment recording what the lines bought.

## Verification

- `uv run ruff check .` — All checks passed. `ruff format --check .` — 887 files already formatted.
- `pytest tests/test_layering.py` — 40 passed.
- `pytest test_workflow_extension test_publication_lifecycle test_workflow_overlay test_layering` — 67 passed, 8 skipped (all `geolens_enterprise package is not installed`, pre-existing).
- `pytest test_datasets.py test_inherited_keywords.py test_rule1_structural.py` — 90 passed.
- `npm run lint`, `npm run typecheck` — clean.
- `npm run test:coverage` — 374 test files passed, 4582 tests passed. Full suite rather than targeted, since this is a type-shape change.
- `make openapi-check` and `make sdks-check` — exit 0, run after the commit (both false-fail on uncommitted regen).

Two gates were RED-checked rather than trusted:

1. The LOC ratchet asserts `actual != cap` fails in **both** directions, so the raise is exact at 1167, not slack.
2. `npm run typecheck` genuinely covers the hand-typed `api.ts`: temporarily setting `RecordStatus = number` failed with 11 errors naming every consumer, which both proves the clean pass was real and enumerated the consumer set.

Closes #1183

### DIFF
--- a/backend/app/modules/catalog/datasets/api/router_data.py
+++ b/backend/app/modules/catalog/datasets/api/router_data.py
@@ -281,8 +281,9 @@ async def update_publication_status(
 ) -> StatusUpdateResponse:
     """Transition a dataset's publication status following allowed paths.
 
-    Allowed transitions:
-      draft -> ready -> internal -> published (and back one step).
+    The allowed set comes from the workflow extension's
+    allowed_transitions(), so an overlay may define its own. The community
+    default is draft -> ready -> internal -> published (and back one step).
     """
     dataset = await db.execute(
         select(DatasetModel)
@@ -346,8 +347,10 @@ async def set_target_status(
 ) -> StatusUpdateResponse:
     """Walk the publication chain from current status to target in one request.
 
-    Executes each intermediate transition so the full chain
-    (e.g. draft -> ready -> internal -> published) completes server-side.
+    Executes each intermediate transition so the full chain completes
+    server-side. The order comes from the workflow extension's
+    status_order(), so an overlay may define its own; the community default
+    is draft -> ready -> internal -> published.
     """
     dataset = await db.execute(
         select(DatasetModel)
--- a/backend/app/modules/catalog/datasets/domain/schemas.py
+++ b/backend/app/modules/catalog/datasets/domain/schemas.py
@@ -335,7 +335,13 @@ class DatasetResponse(BaseModel):
     collections: list["CollectionRef"] | None = None
     # ISO governance fields
     record_status: str = Field(
-        default="draft", description="Lifecycle status: draft, ready, published"
+        default="draft",
+        description=(
+            "Lifecycle status. Deliberately not pinned to an enum: the values "
+            "come from the workflow extension's status_order(), so an overlay "
+            "may define its own. Community default order: draft, ready, "
+            "internal, published."
+        ),
     )
     lineage_summary: str | None = Field(
         default=None, description="Free-text provenance / lineage statement"
@@ -517,7 +523,12 @@ class DatasetMeta(BaseModel):
     record_status: str | None = Field(
         default=None,
         max_length=20,
-        description="Lifecycle status: draft, ready, published",
+        description=(
+            "Lifecycle status. Deliberately not pinned to an enum: the values "
+            "come from the workflow extension's status_order(), so an overlay "
+            "may define its own. Community default order: draft, ready, "
+            "internal, published."
+        ),
     )
     owner_org: str | None = Field(
         default=None, max_length=1000, description="Owning organization name"
--- a/backend/openapi.json
+++ b/backend/openapi.json
@@ -4819,7 +4819,7 @@
                 "type": "null"
               }
             ],
-            "description": "Lifecycle status: draft, ready, published",
+            "description": "Lifecycle status. Deliberately not pinned to an enum: the values come from the workflow extension's status_order(), so an overlay may define its own. Community default order: draft, ready, internal, published.",
             "title": "Record Status"
           },
           "sensitivity_classification": {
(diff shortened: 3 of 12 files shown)


ishiland:
@codex review

PR #1195, opened 2026-08-04
fix(admin): make the Jobs badge read as a failed-count alert, not a total

### LINKED ISSUE #1185: Admin sidebar Jobs badge is unintuitive: failed-count reads like a total
## Problem

The admin sidebar badge next to **Jobs** shows the failed-job count (`useFailedJobCount`, `frontend/src/hooks/use-admin.ts:105`), but it renders identically to the Users / Published Maps / Audit Log badges, which show totals. Clicking Jobs opens the unfiltered list, so the badge says e.g. 3 while the list total says 200. It looks like a count mismatch bug.

## Proposal

Keep the failed count (it's the useful signal), but make it read as an alert:

1. Style the Jobs badge with destructive/red styling, unlike the neutral total badges (`frontend/src/components/admin/AdminSidebar.tsx:62`).
2. Add a tooltip/aria-label like "3 failed jobs".
3. When the badge is shown, clicking the Jobs nav item should land on `/admin/jobs` pre-filtered to `status=failed`, so the clicked number matches the visible list.

No backend change needed — count and list already share identical filters (`backend/app/modules/admin/service.py:621`).

ishiland:
## Problem

The admin sidebar badge next to **Jobs** shows the failed-job count, but it renders identically to the Users / Published Maps / Audit Log badges, which are totals. Clicking Jobs opened the unfiltered list, so the badge could read 3 while the list total read 200 — it looked like a count-mismatch bug rather than an alert.

## Treatment

**Styling.** The Jobs badge takes `semanticBadgeColors.destructive` from `lib/status-colors.ts` — the same recipe the `failed` pill in the jobs table already uses, so the sidebar badge and the row it leads to are the same red by construction. No hex, no new token. The other three badges are untouched and stay neutral.

One non-obvious detail is worth calling out for reviewers: `SidebarMenuBadge` pins its color with `peer-hover/menu-button:text-sidebar-accent-foreground` and `peer-data-[active=true]/menu-button:text-sidebar-accent-foreground`. A bare `text-destructive` loses to both, so the badge would revert to the accent color on hover and stay accent-colored whenever Jobs is the active route — i.e. it would stop being red exactly when you are looking at it. Both variant prefixes are re-declared. Verified two ways: `cn()`/twMerge drops the base colors from the class attribute (asserted in a test), and both variant rules are present in the built CSS after `npm run build`.

**Accessible name.** The badge renders `<span aria-hidden="true">3</span><span class="sr-only">3 failed jobs</span>`. An `aria-label` on the badge `<div>` would be unreliable — it is a generic element with no role — so the label is real text. The digits are `aria-hidden` so it is not announced as "3, 3 failed jobs". The sr-only text uses the true count while the visible digits keep the existing 999+ cap, so at 1500 the badge shows `999+` and speaks "1500 failed jobs".

The nav row also carries `title` as a hover affordance. It could not go on the badge itself: `SidebarMenuBadge` is `pointer-events-none`, and making it `pointer-events-auto` would carve a dead zone out of the row's click target. A test asserts the row's computed accessible name is still `Jobs` and that no link resolves by the label text — the name, not the JSX, per the `<label for>` trap.

**Click target.** While the badge is showing, the row targets `/admin/jobs?status=failed`, so the number clicked equals the number the list shows. At count 0 there is no badge and the row is a plain link to the unfiltered list. `isActive` still keys off the base path, so the row highlights on `/admin/jobs` either way.

That link only means something if the page honours it, so `JobList` now owns `status` in the URL via `useSearchParams` rather than `useState`: it reads on mount, writes back on dropdown change with `{ replace: true }`, and `clearFilters` deletes the param. A status not in `STATUS_OPTIONS` falls back to "all statuses" instead of being forwarded to the API. Page resets to 0 on filter change, as before. Side benefit: a filtered jobs view is now bookmarkable. `user_id`, `search` and `page` remain local state — scope was not widened.

## i18n

New plural pair `adminNav.failedJobs_one` / `_other` in all four locales:

| | `_one` | `_other` |
|---|---|---|
| en | `{{count}} failed job` | `{{count}} failed jobs` |
| es | `{{count}} trabajo fallido` | `{{count}} trabajos fallidos` |
| fr | `{{count}} tâche échouée` | `{{count}} tâches échouées` |
| de | `{{count}} fehlgeschlagener Auftrag` | `{{count}} fehlgeschlagene Aufträge` |

Every `_one` interpolates `{{count}}` rather than hardcoding "1", because French resolves count 0 to `_one`. Only `_one`/`_other` — matching every existing plural key in these bundles; no `_many`. Wording follows each locale's existing `jobs.filters.failed` so the badge and the filter agree (fr "Échouée" takes feminine agreement with "tâche"; de "Aufträge" matches `adminNav.jobs`).

## Judgment calls worth a reviewer's attention

- **URL-owning `JobList` is the one place this goes past the issue text.** The issue says no backend change is needed; it does not say the frontend needed one. But `JobList` held `status` in `useState`, so `?status=failed` would have been silently ignored and the link would have shipped looking correct while doing nothing. The cheaper alternative — seed state from the URL once, never write back — is ~5 fewer lines but leaves the URL lying the moment the user touches the dropdown.
- **At count 0 the row reverts to unfiltered.** Landing on an empty filtered list when nothing is wrong seemed worse than landing on the full list, at the cost of the row's destination changing as jobs fail and get retried.
- **Soft fill (`bg-destructive/10`) rather than a solid red chip**, chosen for consistency with every other status surface over volume.

## Verification

- `npm run lint` — clean. `npm run typecheck` — exit 0 (the real gate; `npx tsc --noEmit` is a no-op here).
- `npm run test:coverage` — **375 test files passed, 4600 tests passed, 0 failed**. Full suite, not targeted.
- `npm run test:i18n` — source-key guard passed, then 1 file / 4 tests passed.
- `npm run build` — clean; also how the peer-variant CSS emission was confirmed.
- Targeted: `AdminSidebar.test.tsx` 29 passed; the three admin test files together 37 passed.

Four RED checks, each disabling exactly one thing, each failing on the assertion under test rather than a fixture error:

1. alert className → `undefined`: 1 failed / 28 passed, on the real base class string.
2. sr-only span + `aria-hidden` removed: 3 failed / 26 passed, "Unable to find an element with the text: 3 failed jobs".
3. `href` forced to the base path: 1 failed / 28 passed, `Expected href="/admin/jobs?status=failed"`.
4. `JobList` status forced to `''`: 3 failed / 49 passed, on the query the hook receives.

One test-quality fix made along the way: the first badge locator found the element by text containing "failed job" — the sr-only label — which made the styling assertion silently depend on the copy, so RED check 1 failed two tests instead of one. It now locates by position within the Jobs menu item, and checks 1 and 2 are disjoint.

Closes #1185

### DIFF
--- a/frontend/src/components/admin/AdminSidebar.tsx
+++ b/frontend/src/components/admin/AdminSidebar.tsx
@@ -42,24 +42,59 @@ import { useEdition } from '@/hooks/use-edition';
 import { useEnterpriseOnlyTabs } from '@/hooks/use-settings';
 import { usePermissions } from '@/hooks/use-permissions';
 import { useSettingsAdmin } from '@/hooks/use-settings-admin';
+import { semanticBadgeColors } from '@/lib/status-colors';
 import type { Capability } from '@/lib/capabilities';
+
+// fix(#1185): the destructive recipe an alert badge needs, matching the
+// `failed` pill in the jobs table (both come from `semanticBadgeColors`).
+// SidebarMenuBadge pins its color with `peer-hover/menu-button:` and
+// `peer-data-[active=true]/menu-button:` variants, so those prefixes have to
+// be re-declared here — a bare `text-destructive` reverts to the accent color
+// the moment the row is hovered or becomes the active route.
+const ALERT_BADGE_CLASSES = [
+  'border',
+  semanticBadgeColors.destructive,
+  'peer-hover/menu-button:text-destructive',
+  'peer-data-[active=true]/menu-button:text-destructive',
+].join(' ');
 
 const overviewItems = [
   { labelKey: 'adminNav.overview', to: '/admin/overview', icon: LayoutDashboard, capability: 'manage_users' },
 ] as const;
+
+// fix(#1185): a badge that counts a PROBLEM rather than a total. Every other
+// Operations badge is a row count for the list behind it, so a bare failed-job
+// count read as a broken total (badge 3, list 200). An alert badge renders in
+// destructive colors, is spoken as "N failed jobs" instead of a naked number,
+// and rewrites the nav target to the matching filtered list so the number the
+// user clicked equals the number the list shows.
+type AlertBadge = {
+  /** Query string appended to `to` while the badge is visible. */
+  filter: string;
+  /** Plural i18n key rendered with `{ count }` for the spoken/hover label. */
+  labelKey: string;
+};
 
 type OperationItem = {
   labelKey: string;
   to: string;
   icon: LucideIcon;
   badgeKey?: 'users' | 'failed' | 'audit' | 'sharedMaps';
+  alertBadge?: AlertBadge;
   enterpriseOnly?: boolean;
   capability: Capability;
 };
 
 const operationsItems: readonly OperationItem[] = [
   { labelKey: 'adminNav.users', to: '/admin/users', icon: Users, badgeKey: 'users', capability: 'manage_users' },
-  { labelKey: 'adminNav.jobs', to: '/admin/jobs', icon: Briefcase, badgeKey: 'failed', capability: 'manage_users' },
+  {
+    labelKey: 'adminNav.jobs',
+    to: '/admin/jobs',
+    icon: Briefcase,
+    badgeKey: 'failed',
+    alertBadge: { filter: 'status=failed', labelKey: 'adminNav.failedJobs' },
+    capability: 'manage_users',
+  },
   { labelKey: 'adminNav.auditLog', to: '/admin/audit', icon: ScrollText, badgeKey: 'audit', capability: 'manage_settings' },
   { labelKey: 'adminNav.sharedMaps', to: '/admin/shared-maps', icon: Link2, badgeKey: 'sharedMaps', capability: 'manage_users' },
   { labelKey: 'adminNav.saml', to: '/admin/saml', icon: Lock, enterpriseOnly: true, capability: 'manage_settings' },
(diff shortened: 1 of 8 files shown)


ishiland:
@codex review

ishiland:
Fixed in 5877e905c.

You're right, and it is the same defect this PR exists to remove arriving through a second door: the badge advertises a count of *all* failed jobs, so a leftover `user=alice` or `skip=75` makes the user click 3 and see 0.

`status` was the only filter in the URL, so on a route where `JobList` is already mounted the other three survived the navigation. The fix resets `userId`, `searchQuery` and `page` when the status param changes from outside the component, and leaves an in-component dropdown change alone — combining status with an existing filter is the point of the dropdown. The two are distinguished by recording our own writes in a ref, which also gives a bookmark or back-button landing the same clean reset.

Both directions are pinned, because a refusal test alone would let the false-positive half regress silently: nothing in "external nav clears the search" would notice the dropdown starting to clear it too.

RED-checked by disabling only the reset body: 1 failed / 6 passed with `AssertionError: expected 'tiles' to be undefined` — the stale filter surviving the nav, which is exactly this finding. The dropdown-direction test stayed green, so the two assertions are disjoint rather than one masking the other.

Full frontend suite: 375 files / 4602 tests passed. lint and typecheck exit 0.

@codex review

ishiland:
Fixed in 3f972fa4f — and you caught the more likely half of the two.

This is the same class one position over, so rather than patch the reported instance I changed the trigger. The reset was keyed on the status *value*, which is silent exactly when an admin already sitting on `?status=failed` applies a filter and clicks the badge again: same URL, same value, no effect run.

It is now keyed on `location.key`, which covers link, bookmark, back-button and same-URL re-click uniformly instead of enumerating URL shapes.

Two facts measured against the running app rather than assumed, since the whole design turns on them:

- A same-URL link click **does** change the key: `6g37p9up` → `25bb8de4` (idx unchanged, so React Router replaces). That is what makes the re-click fire.
- The dropdown's own `replace: true` write **also** produces a new key: `25bb8de4` → `ml9ngm6i`. That is what makes the self-write guard safe — the effect is guaranteed to run after our own write, so the flag is always cleared. Had `replace` reused the key, the flag would have gone stale and swallowed the *next* external navigation, which would have been a worse bug than the one being fixed.

New test for the same-URL case. RED-checked by reverting only the dependency array to the old value-keyed trigger: **1 failed / 7 passed**, and the failure is exactly `clears filters when the alert is re-clicked from the same URL` with `expected 'tiles' to be undefined`. The other seven stayed green, including the original external-nav test, so the new assertion isolates this gap rather than overlapping the previous fix.

Full suite: 375 files / 4603 tests passed. lint and typecheck exit 0.

Also verified end to end in the real app against a live backend: with search `swissalti` set and paged forward, following the alert clears the search, resets the page, and renders exactly the 3 rows the badge advertised.

@codex review

PR #1196, opened 2026-08-04
fix(ingest): stamp raster file_type on the presigned upload path

### LINKED ISSUE #1186: Presigned upload path skips raster metadata stamping → GeoTIFF preview returns 422
## Bug description

Uploading a `.tif` / `.tiff` file on an S3-backed deployment always fails at the preview step with a **422 Unprocessable Entity**:

```
Unable to preview file. The file may be malformed or unsupported.
```

The file itself is a valid GeoTIFF and is successfully uploaded to S3. Only the preview fails.

## Root cause

`_stamp_raster_metadata` — the function that sets `job.user_metadata["file_type"] = "raster"` — is only called inside the non-presigned `upload_file` endpoint (`router.py` line ~560). It is **never called** in `complete_presigned_upload`.

On any S3 deployment `presigned_uploads` is `True` (derived from `storage_provider == "s3"` with no override), so the frontend always uses the presigned path for uploads. This means `file_type` is never stamped for any upload on S3.

When `POST /api/ingest/preview/{job_id}` is called, the branch:

```python
if um.get("file_type") == "raster":   # always False on S3
    ...  # rasterio path — never reached
```

falls through to the vector path, which runs `ogrinfo` on the GeoTIFF. `ogrinfo` fails → 422.

## Why vector files are unaffected

Vector files don't need `file_type` to be set — the code defaults to the ogrinfo path, which handles them correctly regardless.

## Steps to reproduce

1. Deploy GeoLens with `STORAGE_PROVIDER=s3`
2. Upload any valid `.tif` file through the UI
3. Click **Preview** — observe 422 "Unable to preview file. The file may be malformed or unsupported."

## Expected behaviour

The preview endpoint opens the GeoTIFF with rasterio and returns a `RasterPreviewResponse` with band count, CRS, resolution, and COG compliance status.

## Proposed fix

In `complete_presigned_upload`, add a call to `_stamp_raster_metadata` after the file is confirmed present in S3, before the final `db.commit()`:

```python
    job.file_path = s3_key
    await _stamp_raster_metadata(job, s3_key, job.source_filename)  # ← add this
    await db.commit()
```

## Version

Reproduced on **1.6.1** and confirmed still present in **1.7.1** (checked against the published container image).

ishiland:
## Problem

`_stamp_raster_metadata` — which sets `job.user_metadata["file_type"] = "raster"` — was only called from the non-presigned `upload_file` endpoint, never from `complete_presigned_upload`. On any S3 deployment `presigned_uploads` is `True`, so the frontend always uploads through the presigned path and `file_type` was never stamped.

`file_type == "raster"` gates three consumers, and the presigned path missed all of them:

- the preview branch — the reported 422 (`ogrinfo` run against a GeoTIFF)
- `_pick_commit_subclass` — a vector commit body parsed for a raster job
- the ingest dispatch in `queue_ingest_job` — `ingest_file` (ogr2ogr) instead of `ingest_raster`

So the presigned raster pipeline was broken end to end, not just at preview. Reported against 1.6.1 and confirmed on 1.7.1.

Thanks to the reporter — the root cause in the issue was exactly right and made verification fast.

## The design decision

The triage laid out two options. This PR takes **(b): stamp `file_type` by extension unconditionally and stop validating CRS at upload time.**

`_stamp_raster_metadata` is now a synchronous extension check that sets `file_type` only, called from both upload endpoints. Its download-and-validate half is gone. Two things ruled out option (a) — calling the helper as-is from the presigned endpoint:

1. **Cost.** Preview already downloads the same object seconds later via `resolve_file_path`. Option (a) means two full downloads per raster upload, and it puts the second one on the completion request — the one that currently returns instantly. nginx allows 600s on `/api/`, so it would not have timed out; it would just hang for minutes on exactly the file class presigned uploads exist for.
2. **A failure mode the completion endpoint cannot absorb** (this one was not in the issue or the triage). The helper caught only `ValueError` around `validate_raster_crs`. A truncated GeoTIFF raises `RasterioIOError`, which `upload_file` swallows in its outer `except Exception` (500 + staged-file cleanup) and which `complete_presigned_upload` has no equivalent for. Under (a) a corrupt `.tif` would 500 at completion with the S3 object left behind, instead of getting preview's clean 422.

## Why `tasks_raster.py` is in this diff

Deferring CRS validation moves the `crs_missing` flag, so its consumers had to be read rather than assumed. There is exactly one reader: `ingest_raster`. Writes at `router.py`, reads at `tasks_raster.py` — nothing else, and no test referenced it.

- One read only picks between two error messages. Both raise, both fail the job. Cosmetic.
- The load-bearing one is `assign_crs=assign_crs if assign_crs and crs_missing else None`. **An unset flag silently drops a user-supplied `srid_override`**, so `check_and_prepare_cog` runs on a CRS-less source and the COG comes out with no CRS.

That is not pre-existing-and-separable: before this fix presigned rasters never reached `ingest_raster` at all (they dispatched to `ingest_file`), so the router fix is what first routes them there and exposes the gap. `ingest_raster` now derives `crs_missing = not meta.get("crs_wkt")` from the raster it has already opened — the authority the upload-time flag was standing in for. `validate_raster_crs` raises iff `src.crs is None`, and `crs_wkt` is `crs.to_wkt() if crs else None`, so the two conditions are equivalent; the raise condition is byte-identical for the non-presigned path. The now-unreachable short error message is deleted so the actionable one always shows.

Deferring to preview instead would not close this: `commit_import` requires only `status == "pending"`, so a client can skip preview entirely.

## Tests

New `backend/tests/test_presigned_raster_stamp_1186.py`. Every pre-existing raster test drives the non-presigned endpoint, so the presigned path had no coverage at all. The new file runs a GeoTIFF through request-presigned → complete and then asserts each of the three consumers: the stamp itself, preview returning a raster response, and commit dispatching to `ingest_raster`. A vector presigned upload is included as a control, plus direct coverage of the derived `crs_missing`.

Two RED checks, disabled separately:

1. Removing only the `_stamp_raster_metadata` call from `complete_presigned_upload` — 3 failed / 1 passed: `KeyError: 'file_type'`; the reported bug verbatim (`422 ... ogrinfo failed (exit 1): ERROR 4: ... not recognized as being in a supported file format`); and the wrong deferred task. **The vector control still passed**, which is what rules out fixture breakage.
2. Restoring the old flag-based `crs_missing` — 1 failed / 4 passed, on the error-message regex.

Only the worker test moved for edit 2 and only the router tests for edit 1, so the two edits are independently load-bearing and neither masks the other. Both files were restored by position, not by pattern.

## Verification

- `uv run ruff check .` — All checks passed. `uv run ruff format --check .` — 888 files already formatted.
- New file alone — 5 passed.
- `test_presigned_raster_stamp_1186 + test_layering + test_ingest + test_upload_size_limit + test_raster_ingest` — 116 passed.
- `test_rule1_structural + test_rule2_structural + test_api_contract_openapi + test_strict_cog_enforcement + test_raster_ingest_orphan_cleanup_gap017 + test_ingest_job_attempt_fencing` — 111 passed.
- `test_manifest_apply_roundtrip + test_manifest_apply_service + test_reupload + test_reupload_idor + test_reupload_record_type_guard + test_preview_temp_cleanup + test_ingest_fan_out + test_ingest_parquet` — 148 passed.
- `test_layering.py` — 40 passed. The `router.py` cap moves 1493 → 1473 in the same commit (caps are exact, so a shrink fails the gate too), with a comment recording what the removed lines were.

No OpenAPI regeneration: no schema, route, or endpoint-docstring change, and `test_api_contract_openapi.py` passes.

## Known limits, stated rather than glossed

- The worker test observes the derived `crs_missing` through the raise path. The `assign_crs` pass-through is the same variable but is not directly asserted — reaching it means a full COG conversion, and no existing test drives `ingest_raster`'s body.
- `complete_presigned_upload` still never calls `validate_file_content`, which the non-presigned path does. That asymmetry is real and out of scope here; worth its own issue.
- `datasets/api/router_reupload.py` has its own presigned completion with the same missing stamp, but it is unaffected: `_assert_compatible_record_type` rejects raster inputs at every reupload entry point. Checked rather than assumed.

Closes #1186

### DIFF
--- a/backend/app/processing/ingest/router.py
+++ b/backend/app/processing/ingest/router.py
@@ -419,6 +419,11 @@ async def complete_presigned_upload(
     )
 
     job.file_path = s3_key
+    # fix(#1186): the presigned path never stamped file_type, and on an S3
+    # deployment the frontend always uploads through it — so every GeoTIFF
+    # fell through to the vector branch: 422 at preview, a vector commit body
+    # after that, and an ogr2ogr dispatch after that.
+    _stamp_raster_metadata(job, job.source_filename)
     await db.commit()
 
     return UploadResponse(
@@ -454,49 +459,25 @@ async def _cleanup_saved_upload(
         )
 
 
-async def _stamp_raster_metadata(
-    job: "IngestJob",
-    saved_path: Path | str,
-    filename: str | None,
-) -> None:
-    """Perform raster CRS validation and stamp job.user_metadata accordingly.
+def _stamp_raster_metadata(job: "IngestJob", filename: str | None) -> None:
+    """Stamp ``user_metadata["file_type"] = "raster"`` from the filename.
 
-    Non-fatal: missing CRS is acceptable at upload time (the user can supply
-    srid_override at commit time). This helper isolates the raster-specific
-    branch from the main upload flow (KISS-N9).
+    ``file_type`` is the raster discriminator for three consumers — the
+    preview branch below, ``_pick_commit_subclass``, and the ingest dispatch
+    in ``queue_ingest_job`` — so every upload endpoint has to set it before
+    the job is previewable or committable.
+
+    fix(#1186): this used to download the whole object and run
+    ``validate_raster_crs``, which is exactly why ``complete_presigned_upload``
+    never called it — presigned uploads exist for the multi-GB case, and the
+    completion request cannot afford a full-object download (preview downloads
+    the same object seconds later anyway). The only reader of the resulting
+    ``crs_missing`` flag is ``ingest_raster``, which has the raster's metadata
+    in hand and now derives the answer there. So the stamp costs no I/O and
+    both upload endpoints can afford it.
     """
-    lower_filename = (filename or "").lower()
-    if not lower_filename.endswith((".tif", ".tiff", ".vrt")):
+    if not (filename or "").lower().endswith((".tif", ".tiff", ".vrt")):
         return
-
-    from app.processing.raster.cog import validate_raster_crs
-
-    raster_check_path: str | None = None
-    downloaded: Path | None = None
-    if isinstance(saved_path, Path):
-        raster_check_path = str(saved_path)
-    else:
-        try:
-            raster_check_path = await resolve_file_path(str(saved_path), str(job.id))
-            downloaded = Path(raster_check_path)
-        except (
-            Exception
-        ):  # broad: S3 download can fail for network/auth/key-not-found reasons
-            raster_check_path = None
-
-    if raster_check_path:
-        try:
-            await asyncio.to_thread(validate_raster_crs, raster_check_path)
-        except ValueError:
-            # Allow CRS-missing rasters through; user can provide
-            # srid_override at commit time. Store flag for ingest_raster.
-            job.user_metadata = {
-                **(job.user_metadata or {}),
-                "crs_missing": True,
-            }
-        finally:
-            if downloaded is not None:
-                downloaded.unlink(missing_ok=True)
 
     job.user_metadata = {**(job.user_metadata or {}), "file_type": "raster"}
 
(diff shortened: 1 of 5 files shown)


ishiland:
@codex review

Issue #1190, opened 2026-08-04
chore(titiler): bump sidecar image 2.0.5 → 2.2.1 across all three topologies


ishiland:
## Context

The 2026-08-04 TiTiler reassessment recommends keeping the sidecar and holding version drift near zero. We pin `ghcr.io/developmentseed/titiler:2.0.5` in three places; upstream is at **2.2.1** (2026-07-29).

## Why this needs a deliberate bump

Upstream ships security fixes as ordinary bugfix releases with no advisory. Both published exploit chains against this stack (XBOW's expression RCE, Enclave's "MapRoot") were fixed without a GHSA, and OSV plus the GitHub advisory database return nothing for `rio-tiler` or `titiler.core`. Dependabot cannot see this drift, so a deliberate bump is the only control.

This is hygiene, not an emergency. The running 2.0.5 sidecar resolves `rio-tiler 9.2.1`, which already carries the AST expression validator introduced in 9.1.0, so we are not currently exposed on that path.

## Scope

Bump 2.0.5 to 2.2.1 in all three topologies:

- `docker-compose.yml`, titiler service
- `docker-compose.prod.yml`, titiler service
- `geolens-deployments` chart, `values.yaml` -> `titiler.image.tag`

Intermediate releases are 2.1.0, 2.1.1, 2.2.0, 2.2.1. The changes are additive relative to our surface (min/max zoom params, a WMTS `zooms` param, httpx to httpx2). We consume four endpoints (`/cog/tiles`, `/cog/statistics`, `/cog/info`, `/healthz`) and six query params; nothing in that range removes any of them.

## What this does not fix

The titiler image installs rasterio from PyPI wheels rather than a GDAL base image, and 2.0.5 ships GDAL 3.12.1. If 2.2.1 resolves the same rasterio generation, the bump will **not** bring GDAL 3.13.2's `CPL_VSIL_CURL_HEADER_FILE_KVP_ENABLED`, which is the direct fix for MapRoot's `header_file` arbitrary read. Confirm the GDAL version in the new image as part of this work and track that gap separately if it persists.

## Verification

- `backend/tests/test_vrt_titiler.py` (live integration; builds a COG and VRT on the shared staging volume)
- `test_raster_tiles.py`, `test_raster_colormap_proxy.py`, `test_titiler_redirect_env.py`
- `test_phase_272_compose.py` and `test_phase_275_compose_alignment.py` pin the compose shape
- In-container check after the bump:
[...]

PR #1197, opened 2026-08-05
sec(titiler): pin the GDAL rawband env and add a TITILER_WORKERS knob

### LINKED ISSUE #1192: sec(titiler): pin the GDAL security env explicitly instead of inheriting version-dependent defaults
From the 2026-08-04 TiTiler reassessment.

## Problem

The titiler sidecar inherits GDAL's security-relevant defaults rather than pinning them. Defaults are version-dependent, so a base image bump can silently relax them, and the current image is already behind on one of them.

Verified in the running 2.0.5 sidecar: `rio-tiler 9.2.1`, `rasterio 1.5.0`, `GDAL 3.12.1`. The image installs rasterio from PyPI wheels rather than a GDAL base image, so the GDAL version tracks the rasterio wheel.

## Change

Pin these explicitly on the titiler service in all three topologies (compose, prod compose, chart):

- `GDAL_VRT_ENABLE_RAWRASTERBAND` (GDAL >= 3.12, default `YES`). Cannot be set to `NO` here since VRT is a product feature, but pinning it documents the decision.
- `GDAL_VRT_RAWRASTERBAND_ALLOWED_SOURCE` (GDAL >= 3.12, default `SIBLING_OR_CHILD_OF_VRT_PATH`). Already set to `NETWORK_OR_LOCAL`; keep it explicit and re-confirm that value is still the narrowest that works for S3-backed VRTs.
- `CPL_VSIL_CURL_HEADER_FILE_KVP_ENABLED` (GDAL >= 3.13.2). Restricts the `/vsicurl/` `header_file` option, the direct fix for MapRoot's arbitrary read. **Not yet available**: 3.12.1 predates it. Add when the image's GDAL reaches 3.13.2, which is gated on the rasterio wheel, not on the titiler version.

Defaults verified against https://gdal.org/en/stable/drivers/raster/vrt.html and https://gdal.org/en/stable/user/virtual_file_systems.html

## Related

Track alongside #1190 (the 2.0.5 to 2.2.1 bump), which should re-check `rasterio.__gdal_version__` in the new image.

### LINKED ISSUE #1193: chore(titiler): make the worker count env-tunable (TITILER_WORKERS), mirroring UVICORN_WORKERS
Carried over from the 2026-07-02 TiTiler assessment and still open after the 2026-08-04 reassessment.

## Problem

The titiler worker count is hardcoded to `--workers 1` in all three topologies (`docker-compose.yml`, `docker-compose.prod.yml`, and the chart's Deployment command). Tile rendering is CPU-bound under the GIL, so one worker means one render at a time and concurrent cache misses queue.

This is the first knob to reach for when raster feels slow, and it currently requires editing the compose file or the chart template.

## Change

Mirror the existing `UVICORN_WORKERS` pattern with a `TITILER_WORKERS` input, defaulting to 1 so nothing changes for existing installs. Scale `TITILER_MEM_LIMIT` with it: `GDAL_CACHEMAX` is per process, so N workers multiply the cache ceiling.

On Kubernetes `titiler.replicas` already provides horizontal scale-out, so this is mainly a compose and single-host concern. The chart should still gain the knob for parity.

## Caution

Interacts with #678. TiTiler's RSS ratchets under concurrent tile reads, and `MALLOC_ARENA_MAX=2` slows it without stopping it. More workers means more processes each ratcheting, so raising the worker count without raising the memory limit will reach an OOMKill sooner. Document the pairing.

ishiland:
## What this does

Two changes to the titiler sidecar in both compose topologies:

**#1192 — pin the VRT rawband controls** instead of inheriting GDAL's version-dependent defaults, so a base-image bump cannot silently relax them.

**#1193 — make the worker count env-tunable** as `TITILER_WORKERS`, defaulting to 1 so existing installs are byte-identical.

## The #1192 finding: the value we were setting was never valid

`GDAL_VRT_RAWRASTERBAND_ALLOWED_SOURCE` was set to `NETWORK_OR_LOCAL`, commented "Allow VRT bands to reference remote (S3) sources". That is not one of the values GDAL accepts — the set is `SIBLING_OR_CHILD_OF_VRT_PATH | ONLY_REMOTE | ALL | <absolute path>` — so GDAL parsed it as an absolute path and refused **every** VRTRawRasterBand source. Measured in both the 2.0.5 and 2.2.1 images:

```
GDAL_VRT_RAWRASTERBAND_ALLOWED_SOURCE=NETWORK_OR_LOCAL
  -> REFUSED: Invalid value for GDAL_VRT_RAWRASTERBAND_ALLOWED_SOURCE.
     'NETWORK_OR_LOCAL' is not an absolute path
```

So the effective policy has always been deny-all, and the comment claiming otherwise was the only thing that read as permissive. This is the #937 shape again: a line that reads as a decision and is not one.

### Why this pins `ENABLE_RAWRASTERBAND=NO` rather than just correcting the token

The issue text says `NO` is unavailable "since VRT is a product feature". That premise conflates VRT with the `VRTRawRasterBand` **subclass**, and measurement disagrees:

| setting | result |
|---|---|
| `ALLOWED_SOURCE=NETWORK_OR_LOCAL` (today) | refused — invalid value |
| `ALLOWED_SOURCE=SIBLING_OR_CHILD_OF_VRT_PATH` | **raw-band VRT opens and reads** |
| `ENABLE_RAWRASTERBAND=NO` | refused: "VRTRawRasterBand support has been disabled at run-time" |
| ordinary SimpleSource VRT under `NO` | opens and reads, unchanged |

The second row is the decisive one. Correcting the token alone would *start admitting* sibling raw sources that are refused today — a real loosening dressed up as a cleanup. `NO` states today's actual deny-all behaviour in a value GDAL accepts, so it is behaviour-preserving, and it can only ever be equal-or-stricter than any `ALLOWED_SOURCE` value.

It costs nothing here: `VRTRawRasterBand` appears nowhere in this repo — the builder in `processing/raster/vrt.py` emits `SimpleSource` / `ComplexSource` / `AveragedSource`, as `gdalbuildvrt` does. VRT mosaics as a product feature are untouched; only the raw-binary band subclass is off. `ALLOWED_SOURCE` stays pinned at GDAL's documented narrowest legitimate value so that flipping `ENABLE` back to `YES` lands on a sane clamp rather than whatever the base image defaults to that year.

**`CPL_VSIL_CURL_HEADER_FILE_KVP_ENABLED` is deliberately NOT set.** It needs GDAL >= 3.13.2 and both images ship 3.12.1. The image installs rasterio from PyPI wheels, so the **rasterio wheel** gates the GDAL version, not the titiler tag — a future titiler bump will not deliver it either. A comment marks the spot and names what gates it, and a test fails if someone adds it prematurely.

## #1193 — TITILER_WORKERS

Mirrors the existing `UVICORN_WORKERS` pattern. Tile rendering is CPU-bound under the GIL, so one worker renders one tile at a time and concurrent cache misses queue behind it.

The memory pairing is documented at the setting rather than left implicit: `GDAL_CACHEMAX` is per process, so N workers multiply the block-cache ceiling on top of N interpreters, and the #678 RSS ratchet runs independently in each one. Raising workers without raising `TITILER_MEM_LIMIT` reaches OOMKill sooner, not later.

## Verification

Compose-shape suites, from the branch: `test_titiler_gdal_security_env.py` (new, 138 lines) + `test_titiler_redirect_env.py` + `test_phase_272_compose.py` + `test_phase_275_compose_alignment.py` — **64 passed**. Widened to 12 compose/config files — **132 passed**.

`docker compose config -q` exits 0 on both files. All references are parse-safe `${VAR:-default}`; no `:?` was added (INST-01).

Two RED checks, each failing exactly one test for the right reason: reverting `ALLOWED_SOURCE` to `NETWORK_OR_LOCAL` fails `test_allowed_source_is_a_value_gdal_accepts` with the invalid-value message; restoring hardcoded `--workers 1` fails `test_titiler_worker_count_is_env_tunable`.

### Live verification on a running stack

Compose-shape tests cannot prove the container behaves, so this was also exercised end to end against a real stack with both this branch and #1190 applied:

- **Env reaches the container**: `GDAL_VRT_ENABLE_RAWRASTERBAND=NO`, `GDAL_VRT_RAWRASTERBAND_ALLOWED_SOURCE=SIBLING_OR_CHILD_OF_VRT_PATH`.
- **VRT capability, inside the running container**: an ordinary SimpleSource VRT of our writer's shape opened and read (sum 2016); a raw-band VRT was refused with "VRTRawRasterBand support has been disabled at run-time". So mosaics work and only the disabled subclass is blocked.
- **The worker knob actually interpolates**: at default the argv is `--workers 1`; with `TITILER_WORKERS=3` it is `--workers 3` with three forked worker children. The default is byte-identical to the previous hardcoded value.
- **A real tile through the API proxy** was byte-identical to the pre-change baseline (HTTP 200, 3098 bytes, same sha256).

One coverage gap stated plainly: the dev database has zero VRT generations, so no live *mosaic dataset* was rendered through the proxy. The VRT capability was verified directly in the container instead, which is the closest available substitute and is not the same thing.

Closes #1192
Closes #1193

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -338,6 +338,16 @@ FRONTEND_PORT=8080
 # Type: Docker byte value | Default: 1g
 # TITILER_MEM_LIMIT=1g
 #
+# Number of Titiler uvicorn worker processes. Tile rendering is CPU-bound under
+# the GIL, so one worker renders one tile at a time and concurrent cache misses
+# queue — this is the first knob to reach for when raster feels slow.
+# Raise TITILER_MEM_LIMIT above with it. GDAL_CACHEMAX is per process, so N
+# workers multiply the block-cache ceiling on top of N interpreters, and
+# TiTiler's RSS ratchet runs in each one; more workers at the same 1g cap
+# reach OOMKill sooner, not later.
+# Type: integer | Default: 1
+# TITILER_WORKERS=1
+#
 # glibc malloc arena cap inside the Titiler container. Slows (does not stop)
 # TiTiler's RSS ratchet under concurrent tile reads; raise only if malloc
 # contention shows up on many-core hosts. See the docker-compose.yml note.
--- a/backend/tests/test_titiler_gdal_security_env.py
+++ b/backend/tests/test_titiler_gdal_security_env.py
@@ -1,0 +1,138 @@
+"""fix(#1192, #1193): pin the titiler service's GDAL security env and worker knob.
+
+Two invariants, both learned from settings that read as decisions and were not.
+
+#1192 — the VRT rawband controls must be set EXPLICITLY, because their defaults
+are version-dependent and a base-image bump would otherwise move them silently.
+The stronger half is ``test_allowed_source_is_a_value_gdal_accepts``: the value
+that shipped here for months, ``NETWORK_OR_LOCAL``, is not one of GDAL's four
+accepted forms. GDAL fell through to its absolute-path branch and refused every
+VRTRawRasterBand source, so the effective policy was deny-all while the inline
+comment claimed "allow VRT bands to reference remote (S3) sources". Measured
+2026-08-04 in ghcr.io/developmentseed/titiler 2.0.5 and 2.2.1, both
+rasterio 1.5.0 / GDAL 3.12.1::
+
+    Invalid value for GDAL_VRT_RAWRASTERBAND_ALLOWED_SOURCE.
+    'NETWORK_OR_LOCAL' is not an absolute path
+
+The membership check below defaults RESTRICTIVE on purpose — an unrecognised
+token fails rather than passing — because that is the direction in which this
+class of mistake is loud. A test that only asserted "the key is present" would
+have passed against the broken value.
+
+#1193 — the worker count is env-tunable, and the memory cap it must be raised
+with stays env-tunable too. ``GDAL_CACHEMAX`` is per process, so N workers
+multiply the block-cache ceiling; raising workers against a pinned memory limit
+reaches OOMKill sooner (#678).
+
+Static analysis only — no docker daemon required.
+"""
+
+import pytest
+import yaml
+
+from tests.repo_paths import repo_root
+
+_REPO_ROOT = repo_root(__file__)
+_COMPOSE_FILES = ("docker-compose.yml", "docker-compose.prod.yml")
+
+# https://gdal.org/en/stable/drivers/raster/vrt.html (GDAL >= 3.12). The fourth
+# accepted form is an absolute path, handled separately below.
+_ALLOWED_SOURCE_TOKENS = frozenset(
+    {"SIBLING_OR_CHILD_OF_VRT_PATH", "ONLY_REMOTE", "ALL"}
+)
+
+
+def _titiler_service(filename: str) -> dict:
+    body = yaml.safe_load((_REPO_ROOT / filename).read_text(encoding="utf-8"))
+    return body["services"]["titiler"]
+
+
+def _titiler_env(filename: str) -> dict:
+    return _titiler_service(filename)["environment"]
+
+
+def _titiler_command(filename: str) -> str:
+    command = _titiler_service(filename)["command"]
+    return command if isinstance(command, str) else "\n".join(command)
+
+
+@pytest.mark.parametrize("filename", _COMPOSE_FILES)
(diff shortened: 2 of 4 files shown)


ishiland:
@codex review

ishiland:
@codex review

PR #1206, opened 2026-08-05
fix(ingest): validate manifest intent against the live status order

### LINKED ISSUE #1201: fix(ingest): manifest schemas pin a closed record_status Literal over an extension-defined open set
#1183 (PR #1194) documented `record_status` as extension-defined — the values come from the workflow extension's `status_order()`, so an overlay may define its own — and widened the frontend type accordingly. One layer down, the same closed-set assumption survives:

- `backend/app/processing/ingest/manifest_schemas.py:177` — `intent: Literal["draft", "ready", "internal", "published"]`
- `manifest_sources.py` consumes it under the same assumption
- Mirrored in the CLI v1 manifest JSON schema (`cli/geolens_cli/manifest/schemas/geolens-manifest-v1.schema.json`)

An overlay that defines its own status order cannot express that status in a manifest: the API accepts it, the manifest layer 422s it.

Per the convention that a correct convention crossing a boundary where its precondition stops holding should be fixed at the generator: the fix is to validate manifest `intent` against the live extension's `status_order()` at apply time rather than a frozen Literal, and regenerate the CLI schema from that. The community default stays the documented example set.

Refs #1183, #1194.

ishiland:
## Problem

#1183 (PR #1194) documented `record_status` as extension-defined: the values come from the workflow extension's `status_order()`, so an overlay may define its own. The manifest layer still froze them — `ManifestPublication.intent` was a `Literal["draft", "ready", "internal", "published"]`, mirrored in the CLI v1 manifest JSON schema. A deployment whose workflow extension defines its own lifecycle could not express that status in a manifest: the catalog status endpoints accepted it, the manifest apply request 422'd it.

## Fix

`intent` becomes an open string bounded to the shape of the column it lands in (1–20 chars; `record_status` is `String(20)`), and the live set is enforced at apply time by `validate_publication_intent`, which reads `get_workflow_extension().status_order()` — the same authority `set_target_status` uses. Resolved per call, never cached, since an overlay registers after import.

The check runs at the top of `_classify_dataset`, before any staging, download, or queue work and before the `dry_run` early return — validating only where the intent is consumed would have left dry-run, the branch operators use to validate a manifest, as the one branch validating nothing.

The CLI schema drops the enum and keeps `minLength`/`maxLength`, listing the community four under `examples`: the CLI cannot know a server's overlay status order, so an unrecognized intent has to reach the apply endpoint to be judged.

## API impact

- **Error shape changes for a bogus intent**: previously a pydantic 422 at request parse (whole batch rejected); now a per-entry `action="error"` with `accepted: false`, same as every other manifest source error (bad extension, bad CRS, SSRF-refused URL). The message names the allowed set. Unavoidable once the schema cannot know the set; per-entry is also the better behavior for multi-dataset manifests.
- **Visibility fail-closed**: an extension-defined status with no community visibility mapping ingests as `private`. A status this backend cannot interpret must never be the thing that widens who can see the data. The community mapping is byte-identical to before.
- Full regen committed: `backend/openapi.json`, both SDKs (the generated `ManifestPublicationIntent` Literal module goes away), and `frontend/src/types/api.generated.ts`.

## Verification

- Manifest + workflow + lifecycle suites: `87 passed, 8 skipped` (skips are pre-existing enterprise-gated tests); re-run in a different file order to check for state leaks from the overlay helper: `64 passed, 8 skipped`
- Both directions pinned at both layers: a community value passes; a bogus value is refused with the allowed set named; under a simulated `("draft", "review", "published")` overlay, `review` passes while `internal` is refused — which proves the check reads the live `status_order()` rather than a union
- Dry-run refusal RED-checked: with the `_classify_dataset` call replaced by `pass`, the dry-run refusal test fails on `assert response.accepted is False` while the overlay acceptance test still passes
- `pytest tests/test_layering.py -q` — `40 passed`; broad set (manifest + ingest + layering + rule1/rule2 structural) — `232 passed`
- `cd cli && uv run --extra dev python -m pytest -q` — `283 passed`, re-run after SDK regen — `283 passed`
- `make openapi-check`, `make sdks-check` — clean after commit; `cd frontend && npm run types:check` — no drift

Closes #1201


ishiland:
@codex review

PR #1208, opened 2026-08-05
feat(admin): adopt sortable headers on Jobs, Audit Log, and Published Maps

### LINKED ISSUE #1204: feat(admin): adopt sortable column headers on Jobs, Audit Log, and Published Maps
PR #1200 adds server-side sorting to the Users list and ships `SortableColumnHeader` as a shared component. The other three admin tables render the same paginated-table shape but each hardcodes a single `ORDER BY`, so each needs its own backend ordering param before the header can be wired up (client-side sorting over a server-paginated list sorts only the visible page — looks correct, is not).

Adoption analysis (from the #1200 scoping):

| surface | backend site | sortable | NOT sortable (computed post-query) | est. |
|---|---|---|---|---|
| JobList | `AdminService.list_jobs`, `.order_by(IngestJob.created_at.desc())` | created_at, source_filename, status, joined `User.username` | duration (check whether derivable as a column expression) | ~25 be + ~30 fe lines |
| AuditLogViewer | `audit/service.py:197` | created_at, action, resource_type, ip_address, joined username | `resource_name` (resolve_resource_names() runs after the query) | ~25 be + ~25 fe |
| AdminSharedMapsPage | `list_share_tokens` (DISTINCT ON + 3-table join + embed-count aggregate) | Map.name, Map.created_at, User.username, share.expires_at, embed count (NULLs!) | `linkStatus` (derived in Python from is_active + expiry; would need a CASE) | ~40-50 be + ~25 fe |

Shared constraints, same as #1200: closed-enum params (422 at boundary), dict-lookup column allowlist (CodeQL `py/sql-injection` policy), `User.id`-style unique tiebreak on every ordering, `NULLS LAST` pinned on nullable sort columns, sort params in the TanStack query key, URL-owned state. The Audit CSV export shares the list filters — leave export ordering alone (or decide explicitly).

Frontend component needs zero changes; ~15 lines of URL state + header swaps per table.

Refs #1200.

ishiland:
## Problem

#1200 made the Users list sortable and shipped `SortableColumnHeader` as a shared component. The three sibling admin tables render the same paginated-table shape but each hardcoded a single ORDER BY. These lists are server-paginated, so client-side sorting would order only the visible page — which looks correct and is not. Each surface needed its own backend ordering param before the header could be wired up (#1204's scoping table).

## What each surface gained

- **Jobs** (`GET /admin/jobs/`): `created_at`, `source_filename`, `status`, joined `username`, and `duration` — derivable as `completed_at - started_at`, which is NULL for exactly the rows whose Duration cell renders "-", so ordering and display agree. The retry affordance stays unsortable (computed per page after the query).
- **Audit Log** (`GET /admin/audit-logs/`): `created_at`, `action`, `resource_type`, `ip_address`, joined `username` through a dedicated alias so the ORDER BY join cannot interact with the username sub-select the search filter builds (both compile identically today — measured; the alias is insurance against a future filter rewrite). `resource_name` stays unsortable: `resolve_resource_names()` runs after the query.
- **Published Maps** (`GET /admin/share-tokens/`): `map_name`, `created_at`, `creator`, `expires_at`, and the embed count, ordered by the same COALESCE'd expression the cell renders — the raw aggregate is NULL (not 0) for a map with no active embed token, and ordering by that would strand those rows in a band of their own. Link status stays unsortable (derived in Python from `is_active` + expiry).

Shared shape, copied from #1200 on every surface: closed `Literal` enums so bad input 422s at the boundary, a dict-lookup column allowlist that raises on an unmapped field rather than degrading to the default order, `NULLS LAST` pinned in both directions on every nullable sort column, a unique-id tiebreak so OFFSET paging cannot return a row twice, and unchanged default orderings.

**The audit CSV/JSON export deliberately takes no sort/order.** It streams through `stream_audit_logs`, a separate cursor-based generator with its own `created_at DESC`; an export is an archive the recipient reorders in their own tool, and adding a sort would change the row order of every existing consumer's file. Recorded at the handler and pinned by a test so the omission does not read as an oversight.

Frontend: `SortableColumnHeader` reused unchanged. Sort state lives in the URL on all three pages, sort clicks replace rather than push the history entry (per the #1200 review), `aria-sort` marks exactly one header, and an unrecognised `?sort=` falls back to the default silently. JobList's sort write takes the same self-write exemption as its status dropdown — without it the #1185 location-key effect would read the write as an external navigation and clear the user's search on every header click.

No new translation keys: the `sortable.*` strings landed with #1200 and every column label already existed.

## API impact

Three admin list endpoints gain optional `sort`/`order` query params (defaults preserve historical behavior). Full regen committed: `backend/openapi.json`, both SDKs, `frontend/src/types/api.generated.ts`.

## Verification

- New backend suites: `test_admin_job_sort.py` (21 passed), `test_admin_audit_sort.py` (23 passed), `test_admin_share_token_sort.py` (20 passed) — covering 422 on bogus params, NULLS LAST both directions, id tiebreak, unchanged defaults, and that the export takes no sort parameters
- Whole-tree after rebase onto the #1206 merge: backend `7136 passed, 88 skipped`; frontend `npm run lint` + `npm run typecheck` clean, `npm run test:coverage` — 379 files, 4656 tests passed; `npm run test:i18n` — 4 passed
- Regen verified by regenerating from the rebased source: zero drift (`make openapi` + `make sdks` + the frontend types generator all exit 0 with a clean tree)
- `pytest tests/test_layering.py -q` — 40 passed (two caps raised in the same commit with comments: `admin/router_operations.py` 296→316, `catalog/maps/service_public.py` 914→983)
- NULLS LAST on duration RED-checked (fired on DESC with the right assertion); the JobList self-write exemption RED-checked (removing it fails exactly the search-preservation test)

Closes #1204

### DIFF
--- a/backend/app/modules/admin/router.py
+++ b/backend/app/modules/admin/router.py
@@ -25,6 +25,7 @@ from app.modules.admin.schemas import (
     BackfillResponse,
     CatalogStatsResponse,
     EmbeddingStatsResponse,
+    JobSortField,
     SamlToLocalConversion,
     SortDirection,
     UserListResponse,
@@ -713,12 +714,30 @@ async def list_admin_jobs(
     search: str | None = Query(None),
     skip: int = Query(0, ge=0),
     limit: int = Query(50, ge=1, le=200),
+    sort: JobSortField = Query(
+        "created_at",
+        description=(
+            "Column to order by. Duration orders by the completed_at - "
+            "started_at interval, so unfinished jobs sort last either way."
+        ),
+    ),
+    order: SortDirection = Query("desc", description="Sort direction."),
     db: AsyncSession = Depends(get_db),
 ) -> AdminJobListResponse:
-    """List all ingestion jobs with optional status/user/search filters (admin only)."""
+    """List all ingestion jobs with optional status/user/search/sort filters (admin only).
+
+    `sort` and `order` are closed enums, so an unrecognised value is refused
+    with a 422 and never reaches the query.
+    """
     service = AdminService(db)
     rows, total = await service.list_jobs(
-        status=status, user_id=user_id, search=search, skip=skip, limit=limit
+        status=status,
+        user_id=user_id,
+        search=search,
+        skip=skip,
+        limit=limit,
+        sort=sort,
+        order=order,
     )
     retry_capabilities = await asyncio.gather(
         *(get_retry_capability(job) for job, _username in rows)
--- a/backend/app/modules/admin/router_operations.py
+++ b/backend/app/modules/admin/router_operations.py
@@ -19,6 +19,8 @@ from app.modules.admin.schemas import (
     AdminShareTokenResponse,
     InfrastructureConfig,
     InfrastructureResponse,
+    ShareTokenSortField,
+    SortDirection,
 )
 from app.modules.admin.service import AdminService
 from app.modules.audit.service import AuditEvent, audit_emit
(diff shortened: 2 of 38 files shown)


ishiland:
@codex review

ishiland:
@codex review

Issue #1207, opened 2026-08-05
sec(reupload): bring complete_presigned_reupload up to the upload door's completion contract (third door of #1202)


ishiland:
#1202 (PR #1209) fixed the presigned **upload** completion door. The presigned **reupload** completion door — `complete_presigned_reupload`, `backend/app/modules/catalog/datasets/api/router_reupload.py:824-922` — was found during that fix with the same content-validation gap, and the five review rounds on #1209 widened the delta: the reupload door now lacks **all nine** protections the upload door gained, not one.

This is pre-existing (the door shipped several releases ago), not a v1.8.0 regression.

## What the door is missing

Measured against PR #1209's failure-point enumeration (its PR body; that enumeration is the spec for this issue):

1. Content validation — the direct reupload door validates at `router_reupload.py:~255` (422, `detail=str(exc)`); the completion door does not.
2. Freeze-first — it judges nothing and binds the client-writable staging key directly (`job.file_path = s3_key` at ~:915), so it has the same TOCTOU the upload door had.
3. One-shot completion, 4. per-job row lock, 5. pre-copy size gate, 6. drained freeze copy, 7. multipart-retry skip (the exists-iff), 8. delete-after-commit ordering — none present.
9. **No staging reaper at all — the biggest delta, and worse than the upload door ever was.** `tasks_reupload.reupload_file` (~:457) unlinks the local file only; it never deletes the S3 staging object. And a successful reupload job is the per-dataset latest-complete row, which the stale purge exempts from deletion forever (`jobs/router.py` `latest_complete_ids`). So every reupload staging object — recreatable through the client's still-valid PUT URL — lives forever. A fix copied from the upload door would silently not cover this, because the upload door's row 15 relies on reapers this surface does not have.

## Fix shape — extraction, not hoisting

**Do NOT hoist `_validate_presigned_content` into `verify_completed_presigned_upload`.** That was this issue's original guidance and it is now wrong: `verify_completed_presigned_upload` runs on the frozen key as one middle step of six and neither creates nor owns the frozen object. Hoisting only the content check would give this door validation with no freeze, no one-shot, no lock, no size gate, no drain, no retry-skip, no sweep — it would read as fixed and still be racy.

Instead extract three helpers into `processing/ingest/presigned.py`, shared by both doors:

[...]

PR #1209, opened 2026-08-05
sec(ingest): validate content at presigned upload completion

### LINKED ISSUE #1202: sec(ingest): complete_presigned_upload skips validate_file_content that the direct path enforces
The direct upload path calls `validate_file_content(validation_path, file.filename)` (`backend/app/processing/ingest/router.py:534`). The presigned completion path (`complete_presigned_upload`, router.py:345-435) never does — it stamps metadata and commits.

So the two ingest doors enforce different contracts: content sniffing/validation happens only when the bytes pass through the API. A client that presigns can land a file the direct path would have rejected, and the discrepancy is invisible until the worker chokes on it (or doesn't).

Options, in preference order:
1. Validate at completion time from storage (ranged read of the header bytes is likely enough for the current checks) — one boundary, keyed off resolved state.
2. If completion-time validation is too expensive, document the asymmetry explicitly and make the worker's first touch fail closed with the same error taxonomy the API uses.

Noticed while reworking `_stamp_raster_metadata` for #1186 (PR #1196).

ishiland:
## Problem

The direct upload path runs `validate_file_content` on the bytes it stages (`router.py`, the `upload_file` handler). `complete_presigned_upload` did not: it stamped metadata and committed. So the two ingest doors enforced different contracts, and a client that presigned could land a file the direct door refuses. Preview then hands that file to GDAL before the worker's own content check ever runs. #1196 (this release) made the presigned path the default browser path on S3 deployments, which is why this rides in v1.8.0.

## Fix

Completion freezes the object, then judges the frozen copy:

1. **Freeze first.** The object is server-side copied to `staging/{job_id}/frozen/{filename}`, a key no presign endpoint has ever minted a URL for. The client's presigned PUT URL stays valid until expiry (an hour by default) and a PUT to an existing key replaces the object, so anything checked at the staging key can be swapped after the check: upload clean bytes, complete, re-PUT garbage. The order is the fix. Verifying or validating before the copy re-opens the same race one step earlier.
2. **Verify and validate the frozen bytes.** `verify_completed_presigned_upload` (size/quota, which must judge the immutable bytes too, or a 1 KB declaration can swell past quota before the snapshot) and then the same `validate_file_content` the direct door runs, returning the identical 422 with the identical detail string. The check reads only the bytes the validator inspects instead of downloading the object: the first `HEADER_READ_SIZE` (8192) bytes, plus the trailing `PAR1` magic for `.parquet` (appended only when the object exceeds the header window). The `.vrt` branch (whole-body read) is unreachable: `_reject_standalone_vrt` refuses the filename at presign-request time.
3. **Bind and clean up.** `job.file_path` points at the frozen key. Every staging reaper keys off `job.file_path`, so post-ingest deletion and stale-job purges reach it unchanged. The staging object is dropped only after the commit succeeds.

Review hardening on top of the freeze, all the same class (client-writable state re-entering after acceptance, or a failure path leaving state the retry cannot proceed from):

- **One-shot completion.** A job with `file_path` set refuses further completions with a 400. Without it, a second completion re-judged whatever sat at the staging key, and the still-valid PUT URL controls that.
- **Size gate before the copy.** The declared-vs-actual size check runs against the staging object before the freeze, so an oversize upload is refused without paying for a server-side copy of it.
- **Drained freeze.** The copy runs through `_await_provider_call_draining`; cancellation deletes the partial snapshot and re-raises instead of leaving a half-frozen object.
- **Multipart retry works.** Assembly is skipped when the staging object already exists. For S3 that is an iff: uploaded parts are invisible as an object until `CompleteMultipartUpload` succeeds, so the object's presence is a sound record that assembly is done. Without this, a completion that got past assembly and then failed at the freeze left the upload id spent, and the retry the 502 advertises could never succeed.
- **Staging outlives a failed commit.** The staging delete moved after `db.commit()`. Deleting first meant a rolled-back commit stranded the retry: `file_path` unset, bytes gone, frozen copy orphaned.
- **Completion is serialized per job.** The one-shot guard was an unlocked read: two overlapping completions both saw an empty `file_path`, raced over the same deterministic frozen key, and a concurrent refusal could delete the object a concurrent accept then committed a binding to. The handler now re-fetches the job (`lock_presigned_job`) `with_for_update` AND `populate_existing` — the lock alone serializes without informing: SQLAlchemy keeps the already-loaded attributes on the identity-map instance, so the guard would read the pre-commit value and pass on an already-completed job. Both halves are pinned by tests: a driver-level assertion that a `FOR UPDATE` statement is emitted, and a two-session test that the re-fetch reflects a commit made in between. One row locked; completions of different jobs are unaffected.
- **The staging key is reaped at job end.** After completion, `user_metadata["s3_key"]` is the only reference to the client-writable staging key, and the still-valid PUT URL can recreate the object there — S3 cannot revoke an individual presigned URL. The vector and raster task tails both sweep it at terminal status through a shared `reap_presigned_staging_object` helper (extracted so the two cannot drift; raster previously had NO storage sweep at all), and the stale-job purge sweeps it when it deletes a job row. The raster capture happens before any exit from the job-phase block — the sweep's coverage is exactly the exits the capture dominates, which is not visible from the call site. The two event sweeps are backed by a once-per-job post-expiry pass in the periodic purge, keyed off the presign TTL (`created_at` anchors it soundly: the row exists before any URL is minted): a re-PUT after both event sweeps have fired is reaped once the URL is dead, and a `user_metadata` marker takes the job out of the query so permanent latest-complete rows never cost a storage call per pass. That backstop is what makes "nothing can recreate it afterwards" true rather than "we delete it when we notice". The purge is a backstop, not a guarantee: it exempts the newest complete job per dataset — exactly what a successful ingest leaves — which is why the task-level sweeps have to exist.
- **Sweep ownership is decided by the key's own `staging/{job_id}/` prefix, NOT by "differs from `file_path`".** This is the rule a future change will trip over: fan-out children clone the parent's `user_metadata` wholesale, so every child carries the parent's `s3_key`, and a difference-based sweep would delete the shared staging original out from under siblings. The prefix settles ownership outright (`owned_presigned_staging_key`).

`StorageProvider` gains `get_range(key, start, length)` and `copy(src, dst)` with local/S3/Azure implementations on the existing clients (S3: `Range` reads with 416 normalized to empty to match POSIX seek; managed `client.copy` for the >5 GB case `copy_object` caps out on; missing keys normalize to `FileNotFoundError`).

## Failure points, and what a retry sees

Every exit from `complete_presigned_upload`, top to bottom. State is (`file_path`, staging object, frozen object).

| # | Exit | State left behind | The retry |
|---|------|-------------------|-----------|
| 1 | job not found, 404 | untouched | same 404, terminal |
| 2 | not a presigned job, 400 | untouched | terminal |
| 3 | one-shot 400, already completed | untouched | terminal by design: `file_path` set means the bytes were accepted |
| 4 | multipart with no parts, 400 | upload id aborted | must re-presign (pre-existing abort semantics, unchanged) |
| 5 | multipart complete fails, 502 | upload id aborted | must re-presign (pre-existing) |
| 6 | multipart complete cancelled | object deleted, cancel re-raised | must re-presign |
| 7 | staging object missing, 400 | unset / absent / absent | client re-PUTs or re-presigns |
| 8 | pre-copy size check, 422 | staging deleted | hits case 7; the object was refused |
| 9 | freeze copy fails, 502 | unset / present / absent | re-enters, skips assembly, re-copies, proceeds (the multipart-retry fix) |
| 10 | freeze copy cancelled | unset / present / absent | same as 9 |
| 11 | verify 422 (size or quota) | both objects deleted | hits case 7; refused |
| 12 | content validation 422 | both objects deleted | hits case 7; refused |
| 13 | verify/validate transient error | unset / present / absent | same as 9 |
| 14 | `db.commit()` fails | unset / present / present | re-enters, skips assembly, re-copies over the frozen key, re-verifies, binds, commits (the delete-ordering fix) |
| 15 | staging delete fails after commit | bound / present / present | 200 already returned; job bound to frozen. The orphan is swept at job end by whichever terminal reaper owns the job (vector and raster tails) with the stale purge as backstop, so the residual window is completion-to-reap rather than forever |

The invariant that falls out: **staging is deleted only when the upload was refused (8, 11, 12), when the job is durably bound to the frozen copy (15), or when the job reaches its end and the reapers sweep it. Every transient failure keeps it, and per-job row locking means a concurrent refusal can never delete what a concurrent accept has bound.** That is the property that makes "please try again" true.

## Scope note

`complete_presigned_reupload` (the dataset reupload surface) has every gap this PR closes: no content validation, no freeze, no one-shot guard (and so no row lock on it), no size gate, no drained copy, no multipart-retry guard — **and no staging reaper of any kind**: `reupload_file` unlinks the local file only, and the stale purge exempts the newest complete job per dataset, which is exactly what a successful reupload produces, so its staging objects are currently never swept. It is deliberately not fixed here: it sits behind `get_catalog_port()` and needs the completion sequence extracted into shared helpers (`lock_presigned_job`, `should_assemble_multipart`, `finalize_presigned_object` with a postcondition contract), not a single check hoisted into `verify_completed_presigned_upload` — that boundary now runs on the frozen object and owns only one of the six completion steps. Tracked as #1207, which depends on this PR's helpers and uses the enumeration above as its spec.

## Verification

- `backend/tests/test_presigned_content_validation_1202.py` (14 tests): both doors driven with the same payloads, details asserted EQUAL both directions; the rejected completion removes both objects; a read-width pin so a future whole-object download fails; the job binds to a frozen key under `staging/` with the staging object gone; a TOCTOU regression (a re-PUT to the staging key after successful completion cannot change the bytes the job points at); a one-shot replay refusal with a same-length garbage payload so the size check cannot mask the guard; a multipart retry after a failed freeze that succeeds with `complete_multipart_upload` called exactly once; a failed commit that leaves both objects for a successful retry; and a row-lock pin that listens at the driver level and requires a `FOR UPDATE` statement against `ingest_jobs` (a call-level spy would pass on an identity-map hit that never locked anything)
- `backend/tests/test_presigned_staging_reap_1202.py` (13 cases): the ownership predicate including a fan-out child declining the parent's inherited key; both task tails and the stale purge sweeping through the shared helper; sweep failures counted not raised; and a real `ingest_raster` run to its terminal `finally` asserting the staged object is gone
- RED checks, each read for cause: guard disabled → the three refusal tests fail on `assert 200 == 422`; freeze reverted → the TOCTOU test fails with the job's content equal to the attacker's payload; multipart guard reverted → the retry is refused with the parts-required 400 before it reaches the spent-id path; delete moved back before the commit → the bucket holds only the frozen key with `file_path` rolled back, the stranded retry exactly; row lock dropped → the pin fails and dumps the four `ingest_jobs` statements it saw, none carrying `FOR UPDATE`; sweep disabled → the staging key survives in the bucket
- Whole-tree backend suite run on every review round's tree (latest full-tree pass `7141 passed, 89 skipped`); ingest/storage/jobs, reupload/raster/structural/contract, and every purge/reaper consumer (enumerated by grep, not by memory) green on the branch tip; order-varied reruns green
- `tests/test_layering.py` caps updated in the same commits with comments (router.py 1703, tasks_common.py 1703, tasks_vector.py 1075 — the last ratcheted down when the sweep was extracted; all exact)
- `make openapi-check` no drift; `uv run ruff check .` / `format --check` clean

Closes #1202

### DIFF
--- a/backend/app/platform/jobs/models.py
+++ b/backend/app/platform/jobs/models.py
@@ -20,6 +20,34 @@ from sqlalchemy.orm import Mapped, mapped_column
 from app.core.db import Base
 
 
+def owned_presigned_staging_key(
+    job_id: uuid.UUID | str,
+    user_metadata: dict[str, Any] | None,
+    file_path: str | None,
+) -> str | None:
+    """Return the presigned staging key this job alone is responsible for.
+
+    fix(#1202 review r5): a completed presigned upload points ``file_path`` at
+    a frozen copy, which leaves ``user_metadata["s3_key"]`` as the only
+    reference to the client-writable staging key. That URL stays valid until
+    expiry, so the client can recreate the object after completion. Reapers
+    use this to sweep it alongside ``file_path``.
+
+    Ownership is decided by the key's OWN prefix, not by "differs from
+    file_path". ``create_fan_out_jobs`` clones the parent's ``user_metadata``
+    wholesale, so every fan-out child carries the PARENT's ``s3_key``:
+    sweeping on difference alone would delete the shared original out from
+    under siblings that still need it — the same breakage the
+    ``is_fan_out_child`` default-true guard exists to prevent. A staging key
+    is namespaced by the job that presigned it, so the prefix settles
+    ownership outright and needs no survivor query.
+    """
+    key = (user_metadata or {}).get("s3_key")
+    if not isinstance(key, str) or not key or key == file_path:
+        return None
+    return key if key.startswith(f"staging/{job_id}/") else None
+
+
 class IngestJob(Base):
     __tablename__ = "ingest_jobs"
     __table_args__ = (
--- a/backend/app/platform/jobs/router.py
+++ b/backend/app/platform/jobs/router.py
@@ -136,6 +136,108 @@ class StaleCleanupOutcome:
         }
 
 
+# fix(#1202 review r8): how long a presigned PUT URL can still recreate the
+# staging object after the event-triggered sweeps have already run.
+#
+# Grounded in the provider default rather than assumed: S3StorageProvider
+# .generate_presigned_put_url takes `expiration: int = 3600`, and neither call
+# site — the ingest and reupload presign requests — passes one, so 3600 is
+# always the value. No setting exposes it; if one is ever added, derive this
+# from that instead of the literal.
+#
+# Part URLs default to 7200 but cannot recreate an OBJECT: they need a live
+# upload id, and CompleteMultipartUpload consumes it (an abort kills it). Only
+# the single-part object-key URL is a recreation vector, so 3600 bounds it.
+#
+# `created_at` is a sound conservative anchor because the row is inserted
+# BEFORE any URL for it is minted, so no URL can outlive created_at + TTL.
+_PRESIGNED_PUT_URL_TTL_SECONDS = 3600
+_POST_EXPIRY_SWEEP_MARGIN_SECONDS = 900
+_POST_EXPIRY_SWEEP_AFTER_SECONDS = (
+    _PRESIGNED_PUT_URL_TTL_SECONDS + _POST_EXPIRY_SWEEP_MARGIN_SECONDS
+)
+# Set once the post-expiry sweep has run for a job, so it never runs twice.
+_STAGING_REAPED_MARKER = "s3_key_reaped"
+
+
+async def _sweep_expired_presigned_staging(
+    db: AsyncSession, outcome: StaleCleanupOutcome, *, now: datetime | None = None
+) -> StaleCleanupOutcome:
+    """Sweep staging objects a now-dead PUT URL may have recreated.
+
+    The other two sweeps are EVENT-triggered — one at completion, one at job
+    end — and a fast ingest fires both while the client's URL is still valid.
+    A re-PUT after them recreates an object nothing later reaps, because a
+    successful job is the per-dataset latest-complete that the retention purge
+    exempts forever. Those sweeps close the URL's past; this closes its future.
+
+    Runs exactly once per job, ever: the marker takes the row out of the query
+    below, so the exempt rows do not cost a storage call on every pass forever.
+
+    Delete first, mark second — the opposite of ``_reap_committed_staged_paths``
+    and for a different reason. That one must not destroy an artifact a
+    rolled-back row DELETE might still need. Here the row survives either way,
+    so the only asymmetric outcome is marking a FAILED delete as done, which
+    leaks the object permanently. A crash between the two costs one redundant
+    delete on the next pass instead, which is a no-op.
+    """
+    cutoff = (now or datetime.now(timezone.utc)) - timedelta(
+        seconds=_POST_EXPIRY_SWEEP_AFTER_SECONDS
+    )
+    candidates = (
+        await db.execute(
+            select(IngestJob.id, IngestJob.file_path, IngestJob.user_metadata).where(
+                IngestJob.status.not_in(("pending", "running")),
+                IngestJob.created_at < cutoff,
+                IngestJob.user_metadata["s3_key"].astext.is_not(None),
+                IngestJob.user_metadata[_STAGING_REAPED_MARKER].astext.is_(None),
+            )
+        )
+    ).all()
+
+    reaped = 0
+    failures = 0
+    for job_row_id, file_path, metadata in candidates:
+        staging_key = owned_presigned_staging_key(job_row_id, metadata, file_path)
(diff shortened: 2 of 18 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1213, opened 2026-08-05
sec(reupload): bring the presigned completion door up to the upload contract

### LINKED ISSUE #1207: sec(reupload): bring complete_presigned_reupload up to the upload door's completion contract (third door of #1202)

ishiland:
Closes #1207.

#1209 brought the presigned **upload** completion door up to a strict contract: per-job row lock, one-shot completion, freeze-first staging copy, verification and content validation of the frozen bytes, and staging cleanup on every exit. The dataset **reupload** completion door (`POST /api/datasets/{id}/reupload/{job}/commit` path) had the same gaps and was explicitly carved out of v1.8.0 (see the CHANGELOG note). This PR closes it by extracting the contract into shared helpers and applying them to both doors.

## What changed

**`backend/app/processing/ingest/presigned.py`** gains the shared completion contract (+271 lines):

- `lock_presigned_job(db, job_id)`: the #1209 lock helper moved here verbatim, including the two-part docstring property (the SELECT must lock via `with_for_update`, and the read must be fresh via `populate_existing=True`; the lock alone serializes without informing, because the identity map keeps stale attributes).
- `should_assemble_multipart(...)`: the exists-iff multipart retry-skip check.
- `finalize_presigned_object(...)`: freeze to `staging/{job_id}/frozen/{filename}`, verify the frozen bytes, run content validation, and return the frozen key. Owns the failure-point enumeration rows 7-13 from the #1209 PR table and the invariant: it never deletes staging except as a deliberate refusal.

`backend/app/processing/ingest/router.py` (the upload door) now calls the shared helpers instead of its private copies (net -194 lines there).

**`router_reupload.py`** (the reupload door) adopts the full contract:

- Fetch stays the stricter unlocked 404-on-missing lookup, then `lock_presigned_job` before any state read.
- One-shot completion keyed on `job.file_path` presence, same 400 as the upload door.
- Multipart assembly goes through `should_assemble_multipart`.
- `finalize_presigned_object` failures map the same way as upload: a 422 validation refusal stamps the job failed (audit trail) before re-raising; a 502 verification error leaves the job retryable.
- On success: bind the frozen key, commit, then `_cleanup_uncommitted_reupload_source` removes the client-writable source object.

**`tasks_reupload.py`** captures the owned staging key first thing after the job is in hand and reaps it in the task tail, mirroring the #1209 vector/raster tails.

**Port seam**: the three helpers are exposed through `CatalogPort` protocol stubs with `DefaultCatalogPort` delegations, keeping the datasets domain off a direct `processing/` import (`test_layering.py` enforces the boundary; caps adjusted in the same commit with comments).

No OpenAPI change: request/response shapes are untouched, so there is no snapshot or SDK regen in this PR.

## Tests

`test_reupload_completion_contract_1207.py` (8 tests) covers: one-shot rejection, lock freshness (two-session test through the real helper), multipart skip-iff, validation refusal stamping the job failed, verification error leaving it retryable, frozen-key binding, source cleanup after commit, and the task-tail reap. The suite was written RED against the old door (5 failed before the fix) so it demonstrably exercises the gaps. The #1202 content-validation suite passes unchanged against the extracted helpers, pinning the upload door's behavior across the refactor. `test_presigned_staging_reap_1202.py` is adapted for the reaper's new `final_status` keyword (the terminal-status guard folded into the helper from three identical copies in the task tails) and gains a non-terminal-status-never-sweeps case; `test_tenant_staging_storage_keys.py` covers the new `frozen/` key shape.

## Verification

- Whole-tree backend suite at the rebased tip in a detached worktree (results in the PR checks and below).
- `ruff check` and `ruff format --check` clean.
- `test_layering.py` passes with the adjusted caps.

### DIFF
--- a/backend/app/core/catalog_port.py
+++ b/backend/app/core/catalog_port.py
@@ -74,6 +74,10 @@ class CatalogPort(Protocol):
         self, storage: Any, um: dict, physical_key: str
     ) -> bool: ...
 
+    def require_completable_presigned_job(
+        self, job: Any, *, restart_hint: str
+    ) -> None: ...
+
     async def finalize_presigned_object(
         self,
         *,
--- a/backend/app/core/catalog_port.py
+++ b/backend/app/core/catalog_port.py
@@ -65,6 +65,28 @@ class CatalogPort(Protocol):
         job_id: uuid.UUID,
     ) -> int: ...
 
+    # fix(#1207): the presigned completion contract, shared by both doors.
+    # `finalize_presigned_object` owns freeze/verify/validate and every
+    # cleanup decision; see its docstring for the failure postconditions.
+    async def lock_presigned_job(self, db: AsyncSession, job_id: uuid.UUID) -> Any: ...
+
+    async def should_assemble_multipart(
+        self, storage: Any, um: dict, physical_key: str
+    ) -> bool: ...
+
+    async def finalize_presigned_object(
+        self,
+        *,
+        db: AsyncSession,
+        storage: Any,
+        job_id: uuid.UUID,
+        logical_key: str,
+        expected_size: Any,
+        filename: str,
+        user_id: uuid.UUID,
+        request: Any,
+    ) -> str: ...
+
     def visibility_default(self) -> str: ...
 
     async def compute_quality_score(
--- a/backend/app/modules/catalog/datasets/api/router_reupload.py
+++ b/backend/app/modules/catalog/datasets/api/router_reupload.py
@@ -858,16 +858,13 @@ async def complete_presigned_reupload(
             detail="Job is not a presigned upload",
         )
 
-    # fix(#1207): completion is ONE-SHOT, keyed off resolved state. `file_path`
-    # is created empty by the presign request and set only by a completion that
-    # ran to the end, so its presence is the fact "these bytes were accepted".
-    # Without it a second call re-freezes whatever now sits at the staging key,
-    # which the client's unexpired PUT URL controls.
-    if job.file_path:
-        raise HTTPException(
-            status_code=status.HTTP_400_BAD_REQUEST,
-            detail="Upload already completed for this job",
-        )
+    # fix(#1213 review r3): both one-shot facts, shared with the upload door.
+    # This door stamps `failed` itself before a content 422 (below), so without
+    # the status half a client could re-PUT and complete again: a 200 that
+    # binds a frozen object to a row preview and commit will refuse.
+    get_catalog_port().require_completable_presigned_job(
+        job, restart_hint="Start the reupload again."
+    )
 
     storage = get_storage()
     s3_key = um["s3_key"]
--- a/backend/app/modules/catalog/datasets/api/router_reupload.py
+++ b/backend/app/modules/catalog/datasets/api/router_reupload.py
@@ -845,6 +845,11 @@ async def complete_presigned_reupload(
         dataset_id=dataset_id,
         user_id=user.id,
     )
+    # fix(#1207): re-fetch under a row lock with attributes reloaded, then read
+    # the one-shot fact. `_get_bound_reupload_job_or_404` above stays unlocked —
+    # it carries the stricter binding checks (dataset, owner, reupload marker)
+    # and its 404 semantics, which the lock helper must not replace.
+    job = await get_catalog_port().lock_presigned_job(db, job.id)
     um = job.user_metadata or {}
 
     if not um.get("presigned"):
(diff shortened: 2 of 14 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1232, opened 2026-08-06
fix(ingest): refuse manifest source keys under the staging/ prefix

### LINKED ISSUE #1216: ingest: an operator-declared manifest source under a staging/-shaped key would be reaped at terminal status
Found during the #1213 review rounds, pre-existing and orthogonal to that PR.

`reap_downloaded_staging_source` (backend/app/processing/ingest/tasks_common.py) deletes the storage object at `original_file_path` after terminal job status when the path matches the `staging/` prefix — the discriminator for presigned frozen snapshots. `resolve_file_path`'s own comment notes manifest sources are operator-owned physical keys. If an operator ever declared a manifest source under a `staging/`-shaped key, the ingest tail would delete the operator's source object at terminal status.

This predates #1213 (the old path-rewrite guard also fired for downloaded manifest sources) and is unreachable under the documented convention that `staging/` is transient and system-owned. Options, in preference order: (1) validate at manifest-declaration time that source keys do not use the `staging/` prefix — turns the convention into a contract; (2) thread an is-manifest-source flag into the tail to skip the reap; (3) document-only. Option 1 removes a way to be wrong.

ishiland:
Closes #1216.

After #1213, the ingest tail reaps any `staging/`-prefixed storage key at terminal job status — the prefix is the discriminator for system-owned transient presigned-upload objects. An operator who declares an `s3://` manifest source under a `staging/`-shaped key therefore loses the object after its first ingest, and `resolve_file_path` also mis-routes such keys through the tenant storage-key resolver.

This refuses the key at declaration time instead. The check lives in `_storage_uri_to_key`, the single funnel both the apply-create and apply-update/reupload paths pass through (`classify_manifest_source`), beside the existing scheme/bucket/empty-key contract checks. Local schemeless sources are unaffected: they resolve to absolute paths under `upload_staging_dir` and can never match the prefix.

No API or schema impact. The refusal raises `ManifestSourceError`, which surfaces as the existing per-entry `action="error"` in the apply response — the same shape as every neighboring source-contract violation. No Pydantic model changed, so no OpenAPI regeneration.

Tests were RED-verified against the pre-fix code: the rejection case failed with `DID NOT RAISE ManifestSourceError` before the check existed, with a control test proving an ordinary operator key (`operator/roads.geojson`) still classifies.

Verification:

```bash
uv run pytest tests/test_manifest_apply_service.py tests/test_manifest_apply_roundtrip.py -q   # 46 passed
```

Whole-tree backend suite run green on the rebased head before this PR was opened.

### DIFF
--- a/backend/app/processing/ingest/manifest_sources.py
+++ b/backend/app/processing/ingest/manifest_sources.py
@@ -142,6 +142,19 @@ def _storage_uri_to_key(uri: str) -> str:
     key = parsed.path.lstrip("/")
     if not key:
         raise ManifestSourceError("s3:// manifest source must include an object key")
+    # fix(#1216): `staging/` is the discriminator for system-owned transient
+    # presigned-upload objects — the ingest tail reaps `staging/`-prefixed
+    # storage keys at terminal job status, and `resolve_file_path` routes
+    # them through the tenant resolver. An operator-declared manifest source
+    # under that prefix would be deleted at terminal status, so refuse it at
+    # declaration time instead of relying on the naming convention.
+    if key.startswith("staging/"):
+        raise ManifestSourceError(
+            "s3:// manifest source keys must not use the 'staging/' prefix: "
+            "staging/ is system-owned transient storage that is deleted when "
+            "ingest jobs reach terminal status. Move the object to a "
+            "different prefix."
+        )
     return key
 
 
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

PR #1235, opened 2026-08-06
fix(ingest): pre-tag lifecycle fixes — assembly-cancel retry record, sweep race, part-URL TTL

### LINKED ISSUE #1233: fix(ingest): cancel during multipart assembly deletes the presence record after the upload id is spent
Found by the v1.9.0 pre-tag adversarial lifecycle pass (postmortem class: cancellation-during-signal-producing-step x retry-path). Applies to BOTH completion doors: backend/app/processing/ingest/router.py (~:399-407) and backend/app/modules/catalog/datasets/api/router_reupload.py (~:886-894).

should_assemble_multipart (#1202 r3) makes the staging object's presence the one and only record that assembly succeeded, so a retry after a post-assembly failure skips re-assembly. The cancel branch violates that contract:

1. Client uploads N parts (multi-GB), POSTs /complete.
2. CompleteMultipartUpload runs long; client times out / closes the tab; uvicorn cancels the handler.
3. run_in_thread_draining_capture_cancel drains — assembly SUCCEEDS, the upload id is now consumed.
4. The cancel branch deletes the assembled object and re-raises. State: no object, no live upload id, parts consumed.
5. Client retries /complete. Presence check finds no object -> re-assembles with the consumed id -> SDK error -> 502 "the upload session may have expired. Please try again."
6. Every retry 502s until the stale-pending reaper fails the job at 1h; then 400 "Start a new upload." A full re-upload is forced.

Fix: remove the delete-on-cancel in both doors. Without it the natural retry works (object present -> skip assembly -> finalize). An abandoned assembled object is already owned by the pending reaper + post-expiry sweep, so the delete buys nothing except destroying retryability. The line predates the presence-as-record mechanism (#476) and was copied into the reupload door by #1213.

Pre-tag for v1.9.0.

### LINKED ISSUE #1234: fix(ingest): stale-pending sweep can fail a completed-uncommitted presigned job; part-URL TTL outlives the pending window
Found by the v1.9.0 pre-tag adversarial lifecycle pass (postmortem class: timer-window overlap). Two coupled defects around PENDING_TIMEOUT_SECONDS = 3600 (backend/app/platform/jobs/router.py:49):

**(a) The sweeper races completion and wins wrongly.** Presigned completion stamps file_path and commits but leaves status='pending' (processing/ingest/router.py ~:441-447). The pending sweep's UPDATE predicate is only status='pending' AND created_at < cutoff AND no_live_procrastinate_job() (~:464-478). A completion that finishes just past the 1h mark holds the row lock through its S3 copy; the sweeper's UPDATE blocks on that lock, resumes after the commit, still matches (status unchanged), and fails a job that returned 200 "ready for preview" milliseconds earlier. Preview then 400s.

**(b) The server sells a 2h capability on a 1h lifetime.** Multipart part URLs default 7200s (platform/storage/provider.py:113) but the job is guaranteed-failed at 3600s pending. Any multipart upload whose part phase runs past ~1h is dead while its URLs remain valid for another hour.

Scoped fix (the 1h pre-completion abandonment policy itself is deliberate and stays):
1. Add file_path IS NULL to the 1h sweep clause, and add a separate clause failing completed-uncommitted rows (file_path IS NOT NULL, still pending) on a longer window (24h) so they cannot linger forever — without it they would never be swept or purged (the retention purge skips pending rows).
2. Cap the multipart part-URL TTL at PENDING_TIMEOUT_SECONDS so the sold capability matches the job lifetime.

RED-first on (a): a test that completes a job whose created_at is past the cutoff, then runs the sweep, must observe the job survive. Watch the lock semantics in the test (completion holds the row lock to commit).

Pre-tag for v1.9.0.

ishiland:
Closes #1233. Closes #1234.

Two pre-tag fixes for v1.9.0, both found by an adversarial state-interaction review of the presigned lifecycle (the postmortem practice adopted after v1.8.0). Two commits, separable.

**#1233 — keep the assembled object when completion is cancelled.** Both presigned completion doors deleted the assembled object on the cancel branch, after `CompleteMultipartUpload` had already consumed the upload id. That object's presence is the only record assembly succeeded (`should_assemble_multipart`, #1202 r3) — deleting it sent the client's natural retry back into assembly with a spent id, a 502 loop with no way out short of the reaper failing the job and a full re-upload. The branch now drains and re-raises only. The broad `except Exception` below is unchanged: it still maps SDK errors to 502 and aborts the multipart, and cannot catch the re-raised cancellation.

**#1234 — two coupled timer fixes.** The 1h pre-completion abandonment policy stays; what changes:

- The stale-pending sweep no longer races completions. Completion stamps `file_path` then commits while status stays `pending`; the sweep's blocked UPDATE would re-fire after the commit and fail a job that had just returned 200. The 1h clause now applies only to rows that never bound bytes. The guard is a falsy check, not `IS NULL` — the column is nullable but every creator writes `""`, so an `IS NULL` guard would match nothing while reading as fixed.
- Bound-but-uncommitted rows get their own 24h backstop (the retention purge only considers terminal rows, so exempting them from the 1h clause without it would make them immortal), with an error message that no longer claims the upload "never queued".
- Part URLs were sold at 7200s against a 3600s job lifetime. Both now read one setting (`pending_job_timeout_seconds`, default 3600, in `core/config` to avoid a platform/jobs↔platform/storage import cycle) and the S3 provider clamps to it; the protocol docstring states the clamp as a requirement on implementations.

Config impact: new optional setting `pending_job_timeout_seconds` (default preserves current behavior). No API or schema impact; `make openapi-check` clean.

All fixes were RED-verified (each disabled alone reproduces its failure), and the four tests that encode the sweep's positional `db.execute` sequence were updated with comments.

Verification:

```bash
uv run pytest tests/test_stale_pending_reaper.py tests/test_presigned_content_validation_1202.py \
  tests/test_vrt_stale_sweep_gap002.py tests/test_tenant_staging_storage_keys.py tests/test_layering.py -q
make openapi-check
```

Whole-tree backend suite green at the rebased head (7196 passed).


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1245, opened 2026-08-06
fix(observability): adopt prometheus multiprocess mode for multi-worker /metrics

### LINKED ISSUE #651: Metrics stack reports per-worker registries only — adopt prometheus multiprocess mode

### LINKED ISSUE #1240: API /metrics is per-process: multi-worker uvicorn makes rate() and latency alerts fabricate data
## Problem

The prod compose default is \`UVICORN_WORKERS=2\`, but the API exposes plain per-process \`prometheus_client\` metrics on \`/metrics\`. Each scrape lands on one of the N worker processes, so scraped counters sawtooth between per-process values. Prometheus reads every downward flip as a counter reset, with two consequences:

- \`rate()\` manufactures phantom traffic (observed on the demo: ~1 req/s reported vs ~0.03 req/s real).
- \`rate(sum)/rate(count)\` collapses to the processes' **lifetime** mean latency, not recent latency.

## Observed impact

On the demo (2026-08-06, right after the v1.9.0 deploy), \`GeoLensApiTileLatencyMean\` fired and stayed firing on a healthy, idle box (\`/api/health\` 10ms, load 0.44): the fresh container's only traffic had been the heavy cold-cache smoke renders (~2.4s lifetime mean > 2s threshold), and at demo traffic levels nothing dilutes that for days. The prom-guard dead-man check then withheld its Healthchecks pings, producing a false "site DOWN" page. The v1.8.0 post-deploy firing of the same alert retroactively matches this signature.

Diagnostic tell, for anyone reproducing:

```
sum(http_request_duration_seconds_count{job="geolens-api",handler=~"/tiles/.*"})
```

as a \`query_range\` at \`step=15\` — any downward step in a counter-sum means scrapes are alternating between processes.

## Affected surface

Every deployment running the prod compose defaults (2 workers), i.e. all self-hosted installs plus the demo — any alert or dashboard built on \`http_request_duration_seconds\` (and any other per-process counter/histogram) is unreliable there. Dev compose (1 worker) is unaffected, which is why this never shows locally.

## Fix direction

Adopt prometheus_client multiprocess mode:

- Set \`PROMETHEUS_MULTIPROC_DIR\` (tmpfs path) in the api service env.
- Serve \`/metrics\` from a \`CollectorRegistry\` with \`multiprocess.MultiProcessCollector\`.
- Clear the dir on container start; handle worker death mark-files per the prometheus_client docs.

Alternative considered and rejected: pinning \`UVICORN_WORKERS=1\` everywhere (that is the temporary demo mitigation, applied 2026-08-06, and should be reverted when this lands).

## Acceptance

- With \`UVICORN_WORKERS>=2\`, successive \`/metrics\` scrapes are monotonic for all counters.
- A load burst followed by idle shows latency returning to baseline in the \`GeoLensApiTileLatencyMean\` expression within one rate window.
- Demo \`.env\` override \`UVICORN_WORKERS=1\` removed.

ishiland:
## Summary

With `UVICORN_WORKERS>=2` (the prod compose default is 2), every `prometheus_client` metric lived in each worker's own local registry, and `/metrics` was answered by whichever worker took the scrape. Successive scrapes sawtoothed between per-process values. Prometheus reads that as a counter reset, which fabricates `rate()` traffic and collapses `rate(sum)/rate(count)` latency down to a single worker's lifetime mean. That is what caused the `GeoLensApiTileLatencyMean` false "site DOWN" firings after the v1.8.0 and v1.9.0 demo deploys (#1240): the alert pinned on the cold-cache smoke-test render's lifetime mean, and nothing at low demo traffic ever diluted it.

This adopts `prometheus_client` multiprocess mode for the api service (#651):

- `PROMETHEUS_MULTIPROC_DIR` is now set for the api service in both compose files, pointing at a subdirectory of the existing tmpfs `/tmp` mount so it's writable under the read-only rootfs and never survives a container restart.
- `backend/scripts/api-entrypoint.sh` clears stale `.db` files from that directory before uvicorn forks its workers, so a leftover file from a previous container generation (say, a different `UVICORN_WORKERS` topology) can't pollute the new generation's sums.
- `shutdown_worker_metrics()` (new, in `backend/app/observability/metrics/__init__.py`) calls `multiprocess.mark_process_dead()` in the API lifespan's shutdown path, so a worker recycled under `UVICORN_MAX_REQUESTS` (#643) drops its own files instead of leaving a stale series for future scrapes to keep summing.
- `pool.py`'s four connection-pool gauges now declare `multiprocess_mode="livesum"` explicitly. Left at the default (`"all"`), multiprocess mode auto-labels every sample by pid, turning one series into N and silently breaking `GeoLensDbPoolSaturated` and any other alert or dashboard expression that reads the bare metric name. `livesum` keeps the pre-fix single-series shape, now correctly summed across live workers instead of showing one arbitrary worker's pool.
- `geolens_worker_rss_bytes` (#650) needed no change. It already declares its own `pid` label and defaults to multiprocess mode `"all"`, which is exactly the "every live worker in one scrape" behavior #651 wants.
- `prometheus_fastapi_instrumentator` (already pinned at 8.1.0) turns out to already serve `/metrics` from a fresh `CollectorRegistry` plus `MultiProcessCollector` whenever `PROMETHEUS_MULTIPROC_DIR` is set, and falls back to the plain default registry otherwise. So `init_metrics()` itself needed no code change, just a docstring recording why.
- The worker service's separate `/metrics` (port 8001, `backend/app/observability/health/worker.py`) is untouched on purpose. It's a single process, so multiprocess mode would only add pid-file cleanup risk for no benefit, per #651's scoping note. `PROMETHEUS_MULTIPROC_DIR` is set only in the api service's env block, never the worker's, so `jobs.py`'s queue-depth/completed/failed metrics (worker-only) are unaffected.
- Updated `.env.example` and `RUNBOOK.md` to document the new env var and drop the stale "one scrape sees one worker" caveat.

## Ops / config impact

- **New env var**: `PROMETHEUS_MULTIPROC_DIR`, default `/tmp/prometheus-multiproc`, set in both `docker-compose.yml` and `docker-compose.prod.yml` for the `api` service only. Operators running the api image outside these compose files with `UVICORN_WORKERS>1` need to set it themselves, to a writable, container-local, ephemeral directory (documented in `.env.example` and `RUNBOOK.md`).
- No schema/migration/API changes.
- The demo's temporary mitigation (`UVICORN_WORKERS=1` override on the demo host, applied 2026-08-06) should be reverted once this ships. That's an ops action on the demo deployment, not a change in this repo.

## Verification

- `cd backend && uv run ruff check .`: clean
- `cd backend && uv run ruff format --check .`: clean
- `cd backend && set -a && source ../.env.test && set +a && uv run pytest`: full backend suite passes
- `cd backend && set -a && source ../.env.test && set +a && uv run pytest tests/test_layering.py -q`: passes (`backend/app/api/main.py` cap raised 1292 to 1296 in the same commit, with a carve-out comment)
- New regression tests in `backend/tests/test_metrics.py` spawn real subprocesses (the only way to exercise `prometheus_client`'s multiprocess write path, since the value-storage backend is picked at first import) simulating two uvicorn workers, and assert against a real `CollectorRegistry` plus `MultiProcessCollector`:
  - `test_multiprocess_histogram_sums_across_simulated_workers`: the exact `http_request_duration_seconds` metric behind the #1240 alert sums to the true total (10, not 7 or 3) and stays stable across repeated reads.
  - `test_multiprocess_rss_gauge_shows_every_live_worker`: `geolens_worker_rss_bytes` shows both simulated worker pids in one scrape.
  - `test_multiprocess_pool_gauge_uses_livesum_not_per_pid`: `geolens_db_pool_checkedout` (the real production Gauge) stays a single series summed across workers instead of exploding into per-pid series.
  - `test_shutdown_worker_metrics_marks_process_dead_when_multiprocess_active` / `..._noop_without_multiprocess_dir`: the new lifespan-shutdown hook.

Closes #1240, Closes #651

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -264,6 +264,18 @@ GEOLENS_ADMIN_PASSWORD=
 # Type: integer (minimum 1) or empty | Default: 10000 (prod compose)
 # UVICORN_MAX_REQUESTS=10000
 
+# fix(#1240, #651): prometheus_client multiprocess mode for the api service.
+# With UVICORN_WORKERS>=2 each worker used to answer /metrics from its own
+# local registry, so successive scrapes alternated between per-process values
+# and Prometheus read every downward step as a counter reset. Adopting
+# multiprocess mode fixes that. Must be a directory writable by the api
+# container's runtime user and shared by every worker of that ONE container
+# (a subdirectory of the api service's tmpfs /tmp mount, per the compose
+# files, satisfies both). Applies to the api service only; the worker
+# service's separate /metrics endpoint (port 8001) stays single-process.
+# Type: string (directory path) | Default: /tmp/prometheus-multiproc
+# PROMETHEUS_MULTIPROC_DIR=/tmp/prometheus-multiproc
+
 # --- Configuration Lockdown ---
 # [OPTIONAL] When true, all admin-overridable settings are locked to their
 # environment values. The PersistentConfig DB layer is bypassed for reads and
--- a/RUNBOOK.md
+++ b/RUNBOOK.md
@@ -653,7 +653,13 @@ happily (and wrongly) accept.
 ### Metrics & alerting (Prometheus / Grafana)
 
 Metrics are exposed on **two separate endpoints** — the API and the worker each
-run their own:
+run their own. The api service (only) runs multiple uvicorn workers in
+production and is scraped correctly under that fan-out because
+`docker-compose.yml`/`docker-compose.prod.yml` set `PROMETHEUS_MULTIPROC_DIR`
+for it by default (fix #1240 / #651); a bespoke deployment that runs the api
+image outside these compose files with `UVICORN_WORKERS>1` needs to set that
+same env var to a writable, container-local directory or every scrape reverts
+to seeing one arbitrary worker.
 
 | Source | On the Compose network | Host mapping | Exports |
 |---|---|---|---|
@@ -700,14 +706,16 @@ targets just need network reach to `api:8000` and `worker:8001`.
 
 Each API worker samples its own RSS from `/proc/self/status` every 60 seconds
 and exports it as the Prometheus gauge `geolens_worker_rss_bytes` (labelled by
-`pid`) on the API `/metrics` endpoint. One caveat when running more than one
-worker: each process has its own metrics registry and a scrape is answered by
-whichever worker takes it, so a single scrape reports one worker's gauge and
-successive scrapes may alternate pid series — or keep reaching the same worker
-and never observe the one that is growing (multiprocess aggregation is tracked
-in #651). The structured log lines the same sampler writes are therefore the
-authoritative per-worker signal, and a growth curve exists in
-`docker compose logs api` even if `/metrics` is never scraped:
+`pid`) on the API `/metrics` endpoint. The api service runs in
+`prometheus_client` multiprocess mode (`PROMETHEUS_MULTIPROC_DIR`, fix #1240 /
+#651): every live worker's gauge is present in a single scrape, and HTTP
+request counters/histograms sum across workers instead of alternating between
+per-process values. Before this fix, a scrape only ever saw one worker, so
+successive scrapes could sawtooth between unrelated running totals and
+Prometheus read the downward steps as counter resets. The structured log
+lines the RSS sampler writes remain useful as a secondary signal independent
+of scraping — a growth curve exists in `docker compose logs api` even if
+`/metrics` is never scraped:
 
 | Log message | Level | Meaning |
 |---|---|---|
--- a/backend/app/api/main.py
+++ b/backend/app/api/main.py
@@ -23,7 +23,11 @@ from starlette.middleware.gzip import GZipMiddleware
 from starlette.middleware.sessions import SessionMiddleware
 
 from app.api.router import api_router
-from app.observability.metrics import init_metrics, shutdown_worker_metrics
+from app.observability.metrics import (
+    init_metrics,
+    shutdown_worker_metrics,
+    sweep_dead_worker_metrics,
+)
 from app.platform.cache.provider import init_tile_cache
 
 # settings already imported above for the tempdir override — do NOT reimport
@@ -332,6 +336,10 @@ async def lifespan(app: FastAPI):
     # fix(#643): per-worker RSS gauge + log watermark so an OOM-bound worker
     # is visible in normal logs before the kernel kills it.
     memory_metrics_task = asyncio.create_task(update_memory_metrics())
+    # fix(#1240, #651 review round 2): reap gauge_live*.db files left by a
+    # sibling worker that was OOM-killed/SIGKILLed rather than shut down
+    # gracefully -- shutdown_worker_metrics() below never runs for that case.
+    metrics_sweep_task = asyncio.create_task(sweep_dead_worker_metrics())
 
     async def _stale_jobs_sweeper() -> None:
         """Periodically fail jobs whose worker crashed mid-run.
@@ -410,6 +418,7 @@ async def lifespan(app: FastAPI):
 
     pool_metrics_task.cancel()
     memory_metrics_task.cancel()
+    metrics_sweep_task.cancel()
     stale_jobs_task.cancel()
     rate_limit_warmer_task.cancel()
     await task_app.close_async()
--- a/backend/app/api/main.py
+++ b/backend/app/api/main.py
@@ -23,7 +23,7 @@ from starlette.middleware.gzip import GZipMiddleware
 from starlette.middleware.sessions import SessionMiddleware
 
 from app.api.router import api_router
-from app.observability.metrics import init_metrics
+from app.observability.metrics import init_metrics, shutdown_worker_metrics
 from app.platform.cache.provider import init_tile_cache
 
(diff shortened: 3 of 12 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1246, opened 2026-08-06
chore(catalog): drop unused quality_score_numeric column

### LINKED ISSUE #1231: chore(catalog): quality_score_numeric is written nowhere; wire it or drop it
`datasets.quality_score_numeric` (`backend/app/modules/catalog/datasets/domain/models.py:442`, CHECK constraint at :369-371) is written nowhere, read nowhere, and absent from `openapi.json` (verified 2026-08-05). The live quality surfaces are `quality_detail` (JSONB, written at ingest in `backend/app/processing/ingest/tasks_common.py`) and the free-text `quality_statement`.

Either wire it (populate from `quality_detail` at ingest and on validation refresh, then expose it) or drop it (migration removing the column and its CHECK). The decision is most of the work; the diff is small either way.

Leaning drop: a numeric score can be derived from `quality_detail` at read time if it is ever needed, and dead schema invites wrong assumptions, like sorting by a column that is always NULL.

## Where to look

- The column and CHECK: `domain/models.py:442` and :369-371.
- Confirm it is still dead before touching it: `git grep -n quality_score_numeric` across `backend/`, `frontend/src/`, `cli/`, `sdks/`, and `backend/openapi.json` should hit only the model and old migrations.
- If dropping: new Alembic migration with `op.drop_constraint` for the CHECK then `op.drop_column`; check whether the original migration that added it names the constraint explicitly and reuse that name.
- If wiring: the quality score computation already exists as `quality_detail` at ingest (`tasks_common.py`, around the quality sections) and the lazy backfill on first read in `datasets/api/router_data.py:199`; the numeric would be a derived summary of that.

## Suggested path

1. Post the wire-vs-drop recommendation here with the grep output; wait for a maintainer thumbs-up (small decision, quick turnaround).
2. Implement in one small PR. If dropping, remember the ORM model, the CHECK, and the migration all change together.

## Acceptance

- [ ] Decision recorded here (wire vs drop)
- [ ] Migration merged; `make alembic-check` clean
- [ ] If wired: exposed on the API and covered by tests, regen trio committed; if dropped: zero references remain

## Verify

```bash
make migrate && make alembic-check
cd backend && set -a && source ../.env.test && set +a && uv run pytest tests/ -k "quality" -v
```

ishiland:
## Summary

**Schema change — this migration drops a column and its CHECK constraint. Destructive; no downtime concern since the column carries no data (confirmed dead below), but flagging per convention.**

`datasets.quality_score_numeric` was written nowhere, read nowhere, and absent from `openapi.json`. Per ADR-002 Decision 8 (`docs-internal/decisions/adr-002-sources-refresh-v1.md`), the decision is to drop it rather than wire it up — a numeric quality roll-up is a pure function of `quality_detail` (JSONB, written at ingest) and belongs at read time if ever needed; persisting a derived value separately is a second source of truth that goes stale. This decision is explicitly called out in the ADR as **not gated** on the rest of ADR-002 / Milestone 5, so it ships as a standalone change.

Changes:
- `backend/app/modules/catalog/datasets/domain/models.py`: remove the `quality_score_numeric` column and the `chk_quality_score_range` CHECK constraint from the `Dataset` model.
- `backend/alembic/versions/0035_drop_quality_score_numeric.py`: new migration (`0034_linearize_geom_4326` → `0035_drop_quality_score_numeric`) drops the CHECK constraint then the column. `downgrade()` re-adds the nullable column and CHECK — no data restoration since there was none to lose.

Re-confirmed dead via `git grep -n "quality_score_numeric" -- backend/ frontend/ cli/ sdks/ mcp/`: only hits were `domain/models.py` (now removed) and the immutable baseline migration `0001_baseline.py` (expected — migration history doesn't change). `backend/openapi.json` has 0 references, so no SDK/CLI/openapi regen is needed.

Closes #1231

## Verification

- `uv run pytest` (full suite, host, against Postgres :5434) — 7229 passed, 89 skipped, 14 deselected
- `alembic upgrade heads` + `alembic check` against a dedicated scratch database built from this worktree's models/migrations — `No new upgrade operations detected.`
- Pre-commit hooks (ruff, SSRF/visibility structural gates, secret detection) — all passed on commit

### DIFF
--- a/backend/alembic/versions/0035_drop_quality_score_numeric.py
+++ b/backend/alembic/versions/0035_drop_quality_score_numeric.py
@@ -1,0 +1,46 @@
+"""Drop the unused datasets.quality_score_numeric column.
+
+fix(#1231): quality_score_numeric was written nowhere, read nowhere, and
+absent from openapi.json — a scalar that would only ever be a derived
+roll-up of quality_detail (JSONB, written at ingest in tasks_common.py).
+Persisting a derived value separately is a second source of truth that
+goes stale, so ADR-002 Decision 8 calls for dropping it rather than wiring
+it up. This decision is not gated on the rest of ADR-002 / Milestone 5.
+
+Revision ID: 0035_drop_quality_score_numeric
+Revises: 0034_linearize_geom_4326
+Create Date: 2026-08-06
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "0035_drop_quality_score_numeric"
+down_revision: Union[str, None] = "0034_linearize_geom_4326"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.drop_constraint(
+        "chk_quality_score_range", "datasets", schema="catalog", type_="check"
+    )
+    op.drop_column("datasets", "quality_score_numeric", schema="catalog")
+
+
+def downgrade() -> None:
+    # Column had no writers in production, so there is no data to restore.
+    op.add_column(
+        "datasets",
+        sa.Column("quality_score_numeric", sa.Float(), nullable=True),
+        schema="catalog",
+    )
+    op.create_check_constraint(
+        "chk_quality_score_range",
+        "datasets",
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

PR #1247, opened 2026-08-06
fix(analysis): show honest partial-result state for capped previews

### LINKED ISSUE #727: ux(analysis): a capped preview draws the first 500 rows, which reads as a failed operation
Found while filming the 1.5 analysis demo. Not a correctness bug: the cap is labelled honestly in two places. The problem is that the picture it draws is misleading, and it is the common case rather than an edge case.

## What happens

`run_analysis_preview` builds `... ORDER BY gid` and hands it to `execute_safe` with `row_limit=PREVIEW_FEATURE_CAP` (500). So a capped preview returns the 500 lowest gids, which is ingest order, which for a file-sourced dataset is usually the source file's order, which is usually spatially clustered.

Centroids of the 22,324-feature Manhattan Building Heights layer put every dot in two clumps in lower Manhattan and left the rest of the island empty. At a glance that reads as an operation that failed everywhere except two blocks. The toast says `Showing the first 500 of 22,324 source features` and the map chip says `Result · 500 of 22,324 features`, so a user who reads either one is not misled. A user who looks at the map is.

This bites any preview over 500 features, which is most real datasets.

## The ordering is load-bearing, so do not change it

Worth stating explicitly because it is the obvious first idea and it is wrong. `ORDER BY gid` is what lets the sandbox row cap stop the scan early, and the `CROSS JOIN LATERAL (... OFFSET 0)` shape exists specifically to keep it able to ride the pkey index (see the `fix(#700 review)` comment in `service_analysis.py`).

Ordering by `random()`, or by any spatial key, forces the operation to be evaluated over every row and the whole result set sorted before the limit applies. On a 22k-polygon buffer that is the exact cost the current design was built to avoid, and it would be paid on every preview.

Current locations, verified against `main` at 9a788ba92: `ORDER BY gid` at `backend/app/modules/catalog/datasets/domain/service_analysis.py:119`, `row_limit=PREVIEW_FEATURE_CAP` at `:193` (the cap itself at `:39`), and the lateral shape with its `fix(#700 review)` comment at `:102-118`.

## Direction: scope the preview to the viewport

Send the map's current bounds with the preview request and filter on them before the LATERAL:

```sql
WHERE geom_4326 && ST_MakeEnvelope(:minx, :miny, :maxx, :maxy, 4326)
```

Why this fits:

- Every dataset table already carries a GIST index on `geom_4326` (verified: `manhattan_building_heights_geom_4326_idx`, and analysis outputs get one too), so the filter is cheap.
- `ORDER BY gid` survives untouched, so the row-cap early stop is preserved.
- The 500 rows stop being an arbitrary sample and become "the operation applied to what is on screen," which is what the word preview should mean.
- It removes a camera jump as a side effect. Today the preview auto-fits to the result bbox, which yanks the view somewhere the user was not looking. A viewport-scoped result makes that fit a near no-op. (The auto-fit is `onPreviewResult?.(result.geojson, bbox, ...)` at `frontend/src/components/builder/AnalysisPanel.tsx:741` and `:747`, fed by `_extend_bbox` at `service_analysis.py:123-138`.)

The bbox has to be optional. The chat `run_analysis` tool calls the same service with no viewport (`backend/app/processing/ai/chat_analysis.py:110-116`), and the current behaviour is the right fallback there.

### Where the frontend gets the bbox

The panel already holds the live map. `AnalysisPanel` takes `mapInstanceRef?: React.RefObject<MaplibreMap | null>` (`AnalysisPanel.tsx:103`, destructured at `:139`) and already dereferences it for the draw adapter at `:428`, `:454`, and `:625`; `BuilderRail` passes it down at `BuilderRail.tsx:398`. So the bbox is `mapInstanceRef.current?.getBounds()` read at submit time, and no new prop, store, or context is needed. Guard for `null`, exactly as the existing mask-draw button does at `:1158`, and send no bbox when the ref is empty.

### `bbox` must NOT get an `_ANALYSIS_PARAM_OWNERS` entry

`_ANALYSIS_PARAM_OWNERS` at `backend/app/modules/catalog/datasets/domain/schemas.py:863-878` maps operation-scoped request fields to the single operation that reads them, and `_drop_params_for_other_operations` silently deletes any key whose owner does not match the request's `operation`. `bbox` applies to every operation, so it belongs to none of them. Adding it to that dict would make it work for one operation and vanish for the other three, with no error. Leave it out; unlisted keys pass through untouched.

### The denominator: decided

The response already carries a denominator, so this is an adjustment to an existing field rather than a new concept. `AnalysisPreviewResponse.source_feature_count` is `source_feature_count if request.operation != "clip" else None` (`service_analysis.py:217-219`), and for the other operations the value is `dataset.feature_count`, the **cached snapshot**, read at `:187`.

The three cases, decided:

1. **1:1 operation, no bbox.** Unchanged. Cached `dataset.feature_count`, exactly as today.
2. **1:1 operation, with a bbox.** Run a second query counting source rows intersecting the envelope, and return that as `source_feature_count`. The cached total is the wrong denominator here: "500 of 22,324" alongside a viewport-scoped result asserts something the result does not support, which is the same class of dishonesty this issue is about. Bound the count the way `resolve_source_feature_count` bounds its probe (`service_analysis.py:43-67`, `LIMIT cap + 1` inside a subquery) so the denominator query cannot become the expensive half of a preview.
3. **clip, either way.** Still `null`. Clip filters rows and its total is unknowable without evaluating the clip itself. Unchanged.

`truncated` and the returned count keep deriving from the capped row fetch, not from the denominator: `truncated` is "the row cap was hit," and the returned count is `len(features)` after the NULL/EMPTY skip. Do not try to compute either from the new count.

**Ordering trap.** The viewport count has to be issued **before** `await db.rollback()` at `service_analysis.py:188-189`. That rollback is the `release_session` path, and it expires every ORM instance, so a later read of `dataset.table_name` to build the count query raises `MissingGreenlet`. Read what you need and run the count first, the way `source_feature_count` is already read at `:187`.

### The overlay stays a snapshot

Panning after a preview leaves the overlay showing the extent it was computed for. **Re-running the preview on pan is out of scope**: it would fire an advisory-lock-taking sandbox query on every map move, and the lock is shared with AI chat (see #790). The overlay keeps the bbox it was computed against, and the truncation copy names it, so "500 features in the previewed extent" stays true after the user pans away. Live-following the viewport is a separate proposal if anyone wants it.

## Lesser alternative

If viewport scoping is too much for now, give the capped overlay a treatment on the map instead of only in text: draw the extent the 500 rows actually cover, or mark the overlay as partial. Cheaper, and it at least stops the map from asserting something the data does not support.

**Rejected by the 2026-07-27 decision**, recorded here so it is not picked up as a shortcut. The badge already says "N of M features" when truncated (`EphemeralBadge`, `featureCountTruncated`); the failed-operation misread comes from the *map* drawing the first 500 rows in insertion order, which no copy or badge tweak fixes.

## Regeneration checklist

Adding `bbox` to the preview request is an API contract change, so it is not done when the tests pass:

- `make openapi`: regenerates `backend/openapi.json`.
- `make sdks`: regenerates `sdks/` from that snapshot. Never hand-edit generated files.
- Commit both. `make openapi-check` and `make sdks-check` are the CI gates that fail if either is stale.
- The frontend does **not** pick this up from the generated SDK. `AnalysisPreviewRequest` is hand-typed at `frontend/src/types/api.ts:1889-1897` (the response at `:1899-1906`), so the new field has to be added there by hand or the panel will not be able to send it.

## Do not run this concurrently with #953-#956

The four new-operation issues (#953 spatial join, #954 measurement, #955 select by location, #956 intersect/overlay) each widen the same `operation` `Literal` in `schemas.py` and each regenerate `backend/openapi.json`. This issue touches the same request model and the same snapshot. Two of them in flight at once produces a snapshot conflict that resolves as a silent merge, not a test failure. Serialize: land this one, or land those, but not both at once.

No dependency on #1001. An analysis preview never enters `validate_and_execute`; `execute_safe` takes pre-validated SQL (`backend/app/platform/sandbox/executor.py:119-126`).

## Acceptance criteria

- [ ] `bbox` is optional on the preview request; omitting it reproduces today's behaviour byte for byte, pinned by a test.
- [ ] With a bbox, only features intersecting the envelope are returned, and `ORDER BY gid` is still in the emitted SQL. This wants a test asserting the SQL shape, since the performance property is invisible to a result-only assertion.
- [ ] The three denominator cases above are implemented as decided, each pinned by a test.
- [ ] The viewport count query is bounded, and runs before the `release_session` rollback.
- [ ] `bbox` is absent from `_ANALYSIS_PARAM_OWNERS`, with a test that a bbox survives on a non-buffer operation.
- [ ] The chat path still works unchanged (it passes no bbox).
- [ ] A preview over a >500-feature dataset scoped to a viewport draws features across the visible extent rather than clustered in one corner. That is the actual bug, and it wants verifying live, not only in unit tests.
- [ ] Truncation copy reflects the chosen denominator and names the previewed extent; all four locales updated (`npm run test:i18n` is the gate).
- [ ] `make openapi` / `make sdks` run and both snapshots committed.
- [ ] Backend `ruff check` / `ruff format --check` clean; frontend `npm run lint` / `npm run typecheck` clean.

## Not in scope

Unrelated to #696, which is about bounding preview *concurrency* (a global semaphore around `execute_safe`) rather than which rows a single preview returns.

ishiland:
## Summary

Closes #727.

A capped analysis preview drew the first `PREVIEW_FEATURE_CAP` (500) rows in gid (ingest) order. For a file-sourced dataset that order is usually spatially clustered, so a capped preview over a large dataset (the issue's example: 22,324 Manhattan buildings) put every dot in one or two blocks and read as a failed operation, even though the toast and map badge both stated the truncation honestly.

### Backend

- `AnalysisPreviewRequest` gains an optional `bbox: [minx, miny, maxx, maxy]` (EPSG:4326). When present, the preview's row source is filtered to `geom_4326 && ST_MakeEnvelope(...)` **before** the row cap and `ORDER BY gid` apply, so a capped result reflects what's on screen instead of an arbitrary ingest-order sample. `ORDER BY gid` and the pkey-index early-stop are untouched.
- `bbox` is deliberately **not** added to `_ANALYSIS_PARAM_OWNERS` — it applies to every operation, not one.
- `intersect` is explicitly excluded from bbox filtering: its preview runs through a separate JOIN+GROUP BY pipeline (`render_intersect_preview`) that doesn't share the WHERE-clause composition every other operation goes through, and threading bbox through it safely is out of scope here.
- For 1:1 operations (buffer, centroid, spatial_join, measure), `source_feature_count` switches from the cached whole-table `dataset.feature_count` snapshot to a live, bbox-scoped count when a bbox is present — pairing a whole-table total with a viewport-scoped result would be the same class of dishonesty this issue exists to fix. The bbox-scoped count runs *before* the request session's connection-releasing rollback (it reads `dataset.table_name` off the ORM object). Row-filtering operations (clip, select_by_location, intersect) are unaffected — their denominator was already `None`/`match_count`.
- The chat/AI analysis path is unaffected: it never sends a bbox, so its behavior is byte-for-byte unchanged.

### Frontend

- `AnalysisPanel` reads `mapInstanceRef.current.getBounds()` at preview-submit time and sends it as `bbox` (omitted, not undefined, when the ref isn't ready).
- A `viewportScoped` flag threads through both preview surfaces (`EphemeralBadge` for the public viewer / narrow builder rail, `EphemeralPreviewRow` for the builder's stack row) so the "N of TOTAL" copy names the previewed extent instead of implying TOTAL describes the whole dataset.
- New locale strings (`truncatedNoticeTotalScoped`, `featureCountTruncatedScoped`) added to all four locales (en/es/fr/de) with `_one`/`_other` plural forms.

### API contract impact

`AnalysisPreviewRequest` gains one optional field (`bbox`). `backend/openapi.json` and both SDKs (`sdks/python`, `sdks/typescript`) regenerated via `make openapi` / `make sdks` — diffs are additive only, no other endpoint or field changed shape. Worth flagging for release notes as a minor API addition.

## Verification

- `cd backend && uv run pytest` (full suite, host Postgres) — **7249 passed, 89 skipped, 14 deselected**, 0 failures
- `cd backend && uv run pytest tests/test_layering.py -q` — 40 passed (two LOC ratchets raised with justification comments: `service_analysis.py` 505→594, `schemas.py` 1167→1205)
- `cd frontend && npm run build && npm run lint && npm run typecheck && npm run test:coverage` — clean; 379 test files / 4659 tests passed
- `cd frontend && npm run test:i18n` — clean
- `make openapi-check` / `make cli-test` — clean; `make sdks-check` shows the expected pre-commit diff, resolved by this commit
- New tests: pure SQL-builder tests for the bbox predicate (byte-for-byte no-op when omitted, `ORDER BY gid` preserved, intersect explicitly excluded), request-validation tests (length/finiteness/ordering, survives `_ANALYSIS_PARAM_OWNERS` stripping), live-DB integration tests (bbox actually restricts which rows the cap sees, both denominator cases, the pre-rollback ordering requirement), and component tests for the honest-partial-result badge state

## Not run from this worktree

Per `AGENTS.md`, `e2e/analysis.spec.ts` was not run here (the dev stack bind-mounts `main`, so a worktree e2e run would validate the wrong code). Confirmed the spec has no hardcoded copy or bbox assertions that this PR's copy changes would break. It should be exercised post-merge during the smoke phase.


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1244, opened 2026-08-06
fix(audit): close coverage gaps for password logins, dataset.create, and action-registry drift

ishiland:
## Summary

Closes #1230. A 2026-08-05 audit sweep found four gaps in the audit trail:

- **Password login/logout not audited.** Only the OAuth path emitted login events (`oauth.login.success/failure`). The password form (`POST /auth/login`) and `POST /auth/logout` emitted nothing.
- **No `dataset.create` audit action existed anywhere**, despite the frontend (`ChangeHistory.tsx`) already carrying a display label for it.
- **Frontend action registry had drifted from reality.** `CURRENT_AUDIT_ACTIONS` in `AuditLogViewer.tsx` listed `layer.add`, `layer.remove`, `layer.reorder`, `layer.bulk_remove` — never written to `audit_logs` (builder layer edits go to map edit history instead) — and was missing `connector.discover` / `connector.ingest_dispatch`, which are real emit sites. There was no backend registry to check emit sites against at all.
- **No retention guidance** for `audit_logs`, which the router's own docstring already flags as reaching millions of rows.

## Changes

- `backend/app/modules/auth/router.py`: `POST /auth/login` now emits `user.login.success` on success and `user.login.failure` on bad credentials (no `user_id`, since `LocalAuthProvider` deliberately doesn't disclose whether a username exists; never logs the password). `POST /auth/logout` emits `user.logout`.
- `backend/app/modules/catalog/datasets/domain/service_create.py`: `create_dataset()` — the single function every dataset-creation path (ingest registration via `register_existing_table`, file-upload ingest via `tasks_common.py`, layer/table creation, the empty-dataset endpoint) funnels through — now emits `dataset.create`. One emit site instead of duplicating (and risking missing) the call at each router.
- `backend/app/modules/audit/actions.py` (new): canonical `AUDIT_ACTIONS` registry of every `AuditEvent.action` string literal actually emitted.
- `backend/tests/test_audit_action_registry.py` (new): AST-walks `backend/app` and fails if any `AuditEvent(action=...)` literal isn't registered, or if a registry entry is never emitted (both directions of drift).
- `frontend/src/components/admin/AuditLogViewer.tsx`: `CURRENT_AUDIT_ACTIONS` resynced by hand against the backend registry — the four `layer.*` ghosts removed, `connector.discover`, `connector.ingest_dispatch`, `dataset.create`, `user.login.success`, `user.login.failure`, `user.logout` added.
[...]

PR #1260, opened 2026-08-07
ops(audit): convert RUNBOOK retention procedure to a tested script

### LINKED ISSUE #1248: ops(audit): convert RUNBOOK retention procedure to a tested script
## Context

PR #1244 (#1230) added an audit-log retention procedure to `RUNBOOK.md` as prose bash+psql instructions. Across 5 Codex review rounds, real bugs were found in this section every single round:

1. Round 1: retention SQL not verifying the full archive window before deleting; retention delete not tenant-scoped; export/delete cutoff drift.
2. Round 2: empty `$TENANT_ID` cast-to-uuid error before the `OR` could short-circuit; `:cutoff`/`:tenant_id` psql substitutions not reaching inside a `DO $$...$$` dollar-quoted block.
3. Round 3: archive filenames colliding across tenants/runs on the same day; CSV line-count verification unsafe against embedded newlines in `resource_name`.
4. Round 4: silent HTTP-error archives (curl without `-f`/`--fail` saving a JSON error object as if it were a valid export, then deleting the real data); export/delete predicate boundary mismatch (`<=` vs `<`) that could delete an un-archived boundary row.
5. Round 5: no external-Postgres connection path, so operators on managed/external Postgres can't safely run the procedure at all.

An independent second-opinion review (run in parallel with Codex during this same PR) predicted round 4's findings before they landed, and gave this assessment: *"it's a design smell... every single one was a 'looks right, is wrong' class of bug — dollar-quote substitution scoping, UUID cast-of-empty-string, moving-cutoff drift, filename collision, CSV-vs-newline counting. That's not coincidence, it's the failure mode this format produces: none of it runs in CI, none of it is unit-testable, and a human has to manually simulate PL/pgSQL variable scoping and shell quoting rules while reading markdown."*

## Proposal

Convert the retention procedure from RUNBOOK prose into one of:
- A `scripts/audit_retention.sh` with `bats` tests, or
- A `geolens_cli` admin subcommand backed by the same tested export/delete service functions the HTTP export endpoint already uses.

Either form gets the logic under CI, closing this bug class structurally instead of continuing to patch individual manifestations of it.

## Scope note

Not blocking for PR #1244 — that PR fixes all findings found so far as prose and ships as-is. This is deliberately scoped as its own follow-up so the conversion gets a proper design pass rather than being rushed mid-review-cycle.

ishiland:
## Problem

RUNBOOK.md §8 carried the audit-log retention procedure as prose bash+psql. Five codex review rounds on #1244 found a real defect in that section every round (tenant scoping, psql substitution inside `DO $$…$$`, filename collisions, curl without `--fail`, boundary-predicate mismatch), and prose runs in no CI, so the sixth defect would have shipped silently.

## Fix

Replace the prose with `scripts/audit_retention.sh` plus a test that runs the script for real, closing each bug class structurally:

- **One frozen cutoff**, evaluated once in the database and reused by the count, the export, and the delete; both window bounds inclusive everywhere, matching the export endpoint's own filters. One `window_predicate()` string serves count and delete, so an unscoped delete under a scoped count is not expressible.
- **Explicit scope modes with no fallback**: `--tenant-slug` resolves and UUID-shape-validates before any query touches the table (a bad slug aborts the run); the unscoped mode requires a typed operator confirmation phrase, never a config lookup.
- **Collision-proof archives**: filenames carry scope + UTC timestamp + pid, and the script refuses to overwrite an existing file.
- **Verified exports before any delete**: `curl --fail`, archive must parse as a complete JSON array, hold exactly the row count the DB reports for the identical window+scope, and contain no row outside the window.
- **Parameters reach SQL via psql `:'name'` bindings on stdin** (one `psql_value()` helper); the batch-delete loop lives in the shell, so there is no dollar-quoted body for substitution to be silently skipped inside.
- **External-Postgres path**: `--db-url` (SQLAlchemy driver suffix stripped for libpq), then `PG*` env vars, then the bundled db container — managed/external Postgres operators finally have a supported path.

RUNBOOK §8 now documents invoking the script (flags, env, external-Postgres usage, what is verified before deleting) and keeps every operational caveat from the prose, including the one the script cannot satisfy (per-action retention windows still need a manual pass). `.github/workflows/ci.yml` adds the script to the backend paths filter so editing it re-runs its gate.

## Verification

- `backend/tests/test_audit_retention_script.py` runs the real script against a throwaway database with a curl stub serving the export from that same DB: 17 passed. Every failure-mode test asserts the row count is unchanged.
- Each guard was disabled in turn to prove the tests can fail for the right reason (cross-tenant delete, stranded boundary row, missing `--fail`, dropped row-count check).
- Whole-tree backend suite at this sha: 7303 passed, 89 skipped. `ruff check` and `ruff format --check` clean. `test_layering.py` 40 passed.

Closes #1248

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,6 +135,10 @@ jobs:
               - '.env.example'
               - '.dockerignore'
               - 'db/**'
+              # test_audit_retention_script.py runs this script for real against
+              # a throwaway database. Editing the script without re-running its
+              # only gate is exactly the regression it exists to catch.
+              - 'scripts/audit_retention.sh'
               # Frontend files that backend parity/drift tests read by literal
               # path (fe-sdk type drift, basemap/capability/paint-key drift,
               # nginx/CSP/Dockerfile regressions). Specific files only — NOT
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -137,8 +137,11 @@ jobs:
               - 'db/**'
               # test_audit_retention_script.py runs this script for real against
               # a throwaway database. Editing the script without re-running its
-              # only gate is exactly the regression it exists to catch.
+              # only gate is exactly the regression it exists to catch. The same
+              # test asserts audit-archives/ stays gitignored, so a
+              # .gitignore-only PR must run Backend Tests too.
               - 'scripts/audit_retention.sh'
+              - '.gitignore'
               # Frontend files that backend parity/drift tests read by literal
               # path (fe-sdk type drift, basemap/capability/paint-key drift,
               # nginx/CSP/Dockerfile regressions). Specific files only — NOT
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,13 @@ backups/
 restore/
 *.dump
 
+# Audit-log retention archives. RUNBOOK section 8 tells operators to point
+# `scripts/audit_retention.sh --archive-dir` outside the checkout, and this is
+# the second line of defence for anyone who takes the script's default of
+# ./audit-archives while sitting in the repo root. An archive is a verbatim
+# export of usernames, IP addresses and activity.
+audit-archives/
+
 # Playwright
 playwright/.auth/
 playwright-report/
--- a/RUNBOOK.md
+++ b/RUNBOOK.md
@@ -1456,14 +1456,34 @@ the other scripts here. On a managed or external Postgres (§3) there is no `db`
 container, so use either:
 
 ```bash
-# A libpq connection URL:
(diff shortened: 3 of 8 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1259, opened 2026-08-07
chore(deps): bump js-yaml lockfile pins to 4.3.1 for GHSA advisory

ishiland:
## Problem

A newly published js-yaml advisory (affects >=4.0.0 <4.3.1) turns the `JS audit (frontend)` and `JS audit (sdks/typescript)` gates red on every open PR. Dependabot's security update for /sdks/typescript failed outright (run 31175599595 — the advisory ships no patched-version metadata), and its /frontend PR #1258 red-crossed on the sdks/typescript gate, so neither directory could land alone.

## Fix

Lockfile-only bumps of js-yaml 4.3.0 -> 4.3.1 in `frontend/package-lock.json` and `sdks/typescript/package-lock.json`. Both directories pin ranges that admit 4.3.1 (frontend directly; sdks/typescript via the direct pin and the transitive `@hey-api/json-schema-ref-parser` requirement), so no manifest changes are needed.

## Verification

- `cd frontend && npm audit` → 0 vulnerabilities
- `cd sdks/typescript && npm audit` → 0 vulnerabilities
- `cd sdks/typescript && npm run build` → clean (tsc + fix-esm-extensions)

Supersedes #1258 (frontend half of the same advisory).

PR #1262, opened 2026-08-07
chore(deps): bump nanoid lockfile pin to 3.3.17 for GHSA-2v37-7h3g-55p8

ishiland:
## Problem

A newly published nanoid advisory (custom generators can loop indefinitely when size is zero, GHSA-2v37-7h3g-55p8, affects <3.3.17) turns the `JS audit (frontend)` gate red on every open PR (first seen on #1261's checks).

## Fix

Lockfile-only bump of nanoid 3.3.16 -> 3.3.17 in `frontend/package-lock.json`. The transitive requirement via vite -> postcss admits 3.3.17 in-range, so no manifest change is needed.

## Verification

- `cd frontend && npm audit` → 0 vulnerabilities

Same shape as #1259 (js-yaml, this morning).

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -7209,9 +7209,9 @@
       "license": "MIT"
     },
     "node_modules/nanoid": {
-      "version": "3.3.16",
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review

Issue #1219, opened 2026-08-05
feat(catalog): durable refresh history that survives job retention


ishiland:
Blocked by #1217 (history-storage decision: extend `DatasetVersion` vs a sibling table). Line refs as of 2026-08-05.

Terminal `ingest_jobs` rows are purged after a retention window (retention setting in `backend/app/core/config.py`), so the jobs table cannot be the record of refreshes. This is the same reason analysis provenance moved onto `records.derived_from` instead of living in job metadata. `DatasetVersion` (`backend/app/modules/catalog/collections/models.py:92-124`) is the only per-import history today and is file-upload-shaped: `version_number`, `source_filename`, `file_hash`, `uploaded_by`, `uploaded_at`.

## Scope

One durable history row per committed refresh across origins (file re-upload, service re-pull, raster replace): outcome (success or failure), actor, timestamps, feature-count delta, and a slot for the schema-drift snapshot (see the drift issue in this milestone). Failed refreshes are recorded too; a refresh that silently vanishes from history is worse than one that visibly failed.

## Where to look

- Existing model: `collections/models.py:92-124`. Find its current write sites with `git grep -n "DatasetVersion("` under `backend/app/`; the committed-reupload path is the pattern to extend.
- Refresh commit paths that must write rows: `backend/app/processing/ingest/tasks_reupload.py` (`reupload_file` at :110, `reupload_service` at :461) plus the future raster path.
- API: add a dataset-scoped list endpoint next to the existing history/versions routes in `backend/app/modules/catalog/datasets/api/router.py` (`GET /datasets/{id}/history` lives around :538 as a reference).
- Access control: `check_dataset_access_or_anonymous` from `backend/app/modules/catalog/authorization.py` (Rule 1; the pre-commit hook and `tests/test_rule1_structural.py` both check this).

## Suggested path

1. Wait for #1217's extend-vs-sibling decision. If extending `DatasetVersion`: add nullable columns (`origin`, `outcome`, `error_message`, `feature_count_delta`, `schema_diff` JSONB or a ref) so old rows stay valid.
2. Write rows from both reupload task success AND failure branches (the failure branch is the one people forget; grep for where the tasks set job status to `failed`).
3. Add the list endpoint with the Rule 1 check and pagination consistent with the existing history endpoint.
4. Regen trio if response shapes change.

## Acceptance

- [ ] History rows written on every committed refresh path, including failures
- [ ] Queryable via a dataset-scoped API with a Rule 1 access check
[...]

PR #1263, opened 2026-08-07
fix(sdks): preserve schema property order for stable SDK constructors

### LINKED ISSUE #1257: chore(sdk): dump_openapi.py's sort_keys=True can silently reorder generated SDK constructor args
## Context

Found by Codex review on PR #1256 (v1.10.0 release). `backend/scripts/dump_openapi.py:31` writes `backend/openapi.json` with `json.dumps(spec, indent=2, sort_keys=True)`. `sort_keys=True` alphabetizes every dict in the spec, including each schema's `properties`, which is what gives `openapi-check`/`sdks-check` deterministic, reviewable diffs.

The side effect: `openapi-python-client` (the Python SDK generator) preserves that alphabetical property order for a generated model's `attrs`-based `__init__` (required fields are pulled to the front, since attrs requires non-default args before default ones; the rest keep spec order). So adding a new optional field whose name sorts alphabetically before an existing optional field silently reorders the generated constructor's positional arguments for every SDK consumer, on every release, not just this one.

## Concrete instance (this release)

v1.9.0's `AnalysisPreviewRequest.__init__(operation, distance_meters=..., ...)` had `distance_meters` as the second positional argument. #727 added a new `bbox` field to the backend Pydantic model (declared *last* in the class body, deliberately, per its own docstring). Because `"bbox"` sorts before `"distance_meters"` alphabetically, the regenerated SDK model now has `bbox` as the second positional argument instead. A caller using `AnalysisPreviewRequest(operation, 500)` positionally now binds `500` to `bbox` instead of `distance_meters`, which fails validation and returns a 422 on the next call, rather than succeeding with the old semantics.

Severity for v1.10.0 itself: bounded. The failure mode is a loud, immediate 422, not silent data corruption, and only affects positional-argument callers of this one model. Shipped as-is with a CHANGELOG/PR caveat rather than blocking the release; see #1256.

## Fix direction

`openapi-python-client` (checked v0.28.3, no `kw_only` config option exists) can't be configured to generate keyword-only constructors, so the fix has to be on our side: preserve Pydantic field declaration order for `properties` specifically while keeping the rest of the spec's key ordering deterministic for diffing. Options:

1. Write a custom JSON serializer / key-order function that sorts everything `sort_keys=True` currently sorts, except leaves each schema's `properties` dict in insertion order (which already matches Pydantic model field declaration order).
2. Or: verify whether Pydantic v2's `model_json_schema()` already preserves declaration order before `sort_keys=True` flattens it (this repo's evidence — `MapCreate`'s properties are also alphabetical in `openapi.json` — suggests it does preserve order upstream and `sort_keys=True` is what destroys it), then confirm the fix and add a regression test that a model's `openapi.json` property order matches its Pydantic declaration order.

Whatever the fix, it should ship with a test that would have caught this instance (add a field between two existing optional fields on any model with 2+ optional fields, assert the generated SDK's constructor argument order matches Pydantic declaration order, not alphabetical).

Bounded blast radius (loud failure, not silent corruption; SDK generators broadly recommend keyword-only usage for exactly this class of instability). Worth fixing properly rather than rushed into a release PR.

ishiland:
Closes #1257.

## Problem

`backend/scripts/dump_openapi.py` and `scripts/flatten_openapi_defs.py` both wrote their JSON with `sort_keys=True`, which alphabetizes every schema's `properties` map. `openapi-python-client` derives generated constructor argument order from that map, so adding an optional field whose name sorts early silently reordered positional arguments for every SDK consumer on every release. Concrete instance in v1.10.0: #727's `bbox` field displaced `distance_meters` as `AnalysisPreviewRequest.__init__`'s second positional argument.

## Fix

Both snapshot writers now sort every dict key EXCEPT each schema's `properties` map, which keeps insertion order — and insertion order is Pydantic field declaration order. Diff determinism is preserved for everything else. The flattener's content-identity serializer (`_serialize`, feeding `InlineDef_*` sha1 names) deliberately stays fully sorted so synthetic class names do not churn on a property reorder.

## Impact

- **One-time SDK churn (intentional):** every generated constructor moves from alphabetical to declaration order in this PR. After this, adding a field only appends/inserts where the model declares it. SDK consumers using positional args see one more reorder — the same class of break this PR eliminates going forward; keyword args are unaffected.
- No runtime backend/frontend behavior change; `frontend/src/types/api.generated.ts` reorders declarations only (TypeScript types are structural).
- Timed to land **before** the milestone-5 source-API regeneration wave (#1219/#1222/#1224 PRs), per the 2026-08-07 milestone audit, so those regens are stable.

## Regression coverage

`backend/tests/test_openapi_property_order.py`:
- serializer unit test: `properties` maps (incl. nested) preserved, all sibling keys sorted;
- flattener serializer proven identical to the dump serializer;
- live-app spec assertion: `AnalysisPreviewRequest` property order equals its Pydantic `model_fields` order — this test fails on the exact v1.10.0 instance.

## Verification

- `uv run pytest tests/test_openapi_property_order.py tests/test_api_contract_openapi.py` — 19 passed
- `make openapi && make sdks && npm run types:generate` — regenerated artifacts committed, re-run is clean
- `cd frontend && npm run typecheck` — clean

### DIFF
--- a/backend/openapi.json
+++ b/backend/openapi.json
@@ -23,18 +23,6 @@
             "title": "Configured",
             "type": "boolean"
           },
-          "error": {
-            "anyOf": [
-              {
-                "type": "string"
-              },
-              {
-                "type": "null"
-              }
-            ],
-            "description": "Short sanitized failure reason. Never contains the key or raw provider error bodies.",
-            "title": "Error"
-          },
           "ok": {
             "anyOf": [
               {
@@ -58,6 +46,18 @@
             ],
             "description": "HTTP status returned by the provider on failure, when available.",
             "title": "Status"
+          },
+          "error": {
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "description": "Short sanitized failure reason. Never contains the key or raw provider error bodies.",
+            "title": "Error"
           }
         },
         "required": [
@@ -85,21 +85,17 @@
       },
       "AIStatusResponse": {
         "properties": {
-          "configured": {
-            "description": "Whether an API key is configured. AI features require both 'enabled' and 'configured'.",
-            "title": "Configured",
-            "type": "boolean"
-          },
-          "enabled": {
-            "description": "Whether AI features are enabled for this instance.",
-            "title": "Enabled",
-            "type": "boolean"
-          },
-          "has_embeddings": {
-            "default": false,
-            "description": "Whether at least one record has embeddings stored.",
-            "title": "Has Embeddings",
-            "type": "boolean"
+          "provider": {
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "description": "Active AI provider name (e.g. 'anthropic', 'openai').",
+            "title": "Provider"
           },
           "model": {
             "anyOf": [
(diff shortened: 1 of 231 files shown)


ishiland:
@codex review

Issue #1218, opened 2026-08-05
feat(catalog): add source-state fields to Dataset (origin pointer, refresh/check timestamps, health, drift)


ishiland:
Blocked by #1217 (origin-pointer shape and per-origin semantics). Line refs are as of 2026-08-05 and may drift.

`Dataset` today carries only descriptive source info: `source_format`, `source_filename`, `original_srid`, `source_url` (`backend/app/modules/catalog/datasets/domain/models.py:449-453`), and `source_url` is user-editable via PATCH because it sits in `_DATASET_FIELD_MAP` (`domain/service_metadata.py:115-118`). No `last_refreshed_at`, `last_checked_at`, health, or drift state exists anywhere in the schema (grep-verified).

## Scope

- New columns per the ADR: system-managed origin pointer, `last_refreshed_at`, `last_checked_at`, `source_health`, drift status.
- Alembic migration with backfill.
- Read-only exposure on dataset API responses. The origin pointer must NOT be writable through the metadata PATCH map.

## Where to look

- Model: `backend/app/modules/catalog/datasets/domain/models.py` (Dataset class starts ~:365; the existing "Source info" block at :449-453 is where the new columns belong; copy the CHECK-constraint style used by `chk_datasets_source_format` at :386-392 if the ADR gives `source_health` a closed value set).
- Migration: `backend/alembic/versions/` (copy the newest migration's structure; backfill is a plain `UPDATE` in `upgrade()`).
- Response schema: `DatasetResponse` in `domain/schemas.py`; PATCH allowlist: `_DATASET_FIELD_MAP` in `domain/service_metadata.py:115` (add nothing there for the origin pointer, and add a test proving PATCH cannot touch it).
- Frontend type mirror: `frontend/src/types/api.ts` (hand-maintained; add the new fields near the existing `source_format`/`source_url` entries at ~:218 and :245).

## Suggested path

1. Wait for #1217 to fix the field names and the per-origin backfill rule (likely: existing datasets get their import time from `records.created_at`; `created` datasets get an explicit n/a state).
2. Add columns to the model, generate the migration, run `make migrate`, then `make alembic-check`.
3. Write the backfill in the same migration and cover it with a test that seeds a pre-migration-shaped row.
4. Add the fields to `DatasetResponse`. Do not add them to `DatasetMeta` (the PATCH schema).
5. Add a test that PATCHes the origin-pointer field name and asserts it is rejected or ignored.
[...]

PR #1264, opened 2026-08-07
feat(catalog): compute dataset source freshness from update_frequency

### LINKED ISSUE #1224: feat(catalog): computed staleness from update_frequency
Blocked by the source-state fields issue in this milestone (needs `last_refreshed_at`); vocabulary per #1217. Line refs as of 2026-08-05.

`records.update_frequency` is stored with the ISO 19115 vocabulary (`continual, daily, weekly, monthly, quarterly, biannually, annually, asNeeded, irregular, notPlanned, unknown`; CHECK at `backend/app/modules/catalog/datasets/domain/models.py:38-43`), exported to DCAT (`backend/app/standards/dcat/service.py`), and recommended by the validation service. Nothing compares it against anything. Once `last_refreshed_at` exists, staleness is a pure read-side computation.

## Scope

- Computed freshness state (`fresh` / `due` / `overdue` / `unknown`) from `last_refreshed_at` against the declared `update_frequency`. No scheduler, no background jobs, no writes.
- Exposed on dataset responses and search results; feeds the chips issue.
- Staleness never blocks any operation.

## Where to look

- Put the pure function in the datasets domain (a small module next to `domain/service_metadata.py`; check `test_layering.py` caps before growing an existing file).
- Dataset response assembly: grep for where `DatasetResponse` is constructed in `domain/` and attach the computed field there.
- Search results: the search domain under `backend/app/modules/catalog/search/`; find where dataset properties are serialized into search features and add the same computed field.
- Frontend type mirror: `frontend/src/types/api.ts`.

## Suggested path

1. Implement the mapping as a pure function of `(last_refreshed_at, update_frequency, now)`. Starting thresholds (tune in review, keep them in one constants table): a dataset is `due` past 1x its period and `overdue` past 2x. Periods: continual/daily 1 day, weekly 7, monthly 31, quarterly 92, biannually 183, annually 366 days. `asNeeded`, `irregular`, `notPlanned`, missing frequency, or missing `last_refreshed_at` all yield `unknown`.
2. Unit-test every vocabulary value plus the two missing-input cases; pass `now` in explicitly so tests are deterministic.
3. Attach to dataset and search responses; regen trio.

## Acceptance

- [ ] Deterministic mapping tested for every frequency vocabulary value and both missing-input cases
- [ ] Present in search and dataset APIs; OpenAPI regen trio committed
- [ ] Zero writes and zero scheduling introduced

## Verify

```bash
cd backend && set -a && source ../.env.test && set +a && uv run pytest tests/ -k "staleness or freshness" -v
cd .. && make openapi && make sdks
```

## Traps

- French locale plural rules and copy come later (chips issue); this issue is API-only, keep it that way.
- Do not persist the computed state; it derives from two stored fields and persisting it creates a second source of truth that goes stale by definition.

ishiland:
Closes #1224. Implements ADR-002 Amendment A6.

## What

`records.update_frequency` has carried the ISO 19115 maintenance-frequency vocabulary since the catalog shipped, and nothing ever compared it to anything. With `datasets.last_refreshed_at` in place (#1218), source freshness becomes a pure read-side computation over stored columns — no new writes, no persistence, computed at response time.

`domain/source_freshness.py` maps `(last_refreshed_at, update_frequency, now, origin)` to `fresh` / `due` / `overdue` / `unknown`: due past one declared period, overdue past two (periods: continual/daily 1 day, weekly 7, monthly 31, quarterly 92, biannually 183, annually 366). `asNeeded`, `irregular`, `notPlanned`, missing frequency, or missing refresh timestamp yield `unknown`.

## Origin gates the answer (Amendment A6)

`created` datasets are stamped with `last_refreshed_at` at creation but answer 409 `refresh_not_applicable` — without the gate, an eighteen-month-old sketch layer declaring monthly cadence would report "overdue" and name an action that does not exist for it. Refreshable origins are an **allowlist pinned against `ORIGIN_KINDS`**, so adding a future origin kind fails loudly instead of silently inheriting a deadline. NULL-origin (VRT) datasets keep their deadline because Decision 5a projects VRT generation timestamps into `last_refreshed_at`.

Named **source freshness** everywhere — deliberately distinct from the existing quality-freshness concept in `frontend/src/lib/quality-freshness.ts`, which has different states and thresholds.

## Surface

- `source_freshness` on `DatasetResponse` and search record responses; `now` read at the response boundary only.
- Zero writes, zero migrations.
- Regen trio committed (post-#1263 declaration-order scripts).
- `test_layering.py`: `schemas.py` cap 1271 → 1286 for the field + description.

## Verification

- Whole-tree backend at the rebased sha: **7507 passed, 89 skipped**; ruff check + format clean.
- `backend/tests/test_dataset_source_freshness.py`: 499 lines — every threshold boundary with explicit `now`, the created-origin gate, the allowlist-pinning test.
- Frontend: typecheck, lint, **4669 vitest passed**.

### DIFF
--- a/backend/app/modules/catalog/datasets/domain/helpers.py
+++ b/backend/app/modules/catalog/datasets/domain/helpers.py
@@ -2,6 +2,7 @@
 
 import uuid
 from collections.abc import Iterable, Mapping
+from datetime import datetime, timezone
 
 from sqlalchemy import select
 from sqlalchemy.ext.asyncio import AsyncSession
@@ -9,6 +10,9 @@ from sqlalchemy.ext.asyncio import AsyncSession
 from app.core.identity import Identity
 from app.core.record_types import RASTER_FAMILY_RECORD_TYPES
 from app.modules.auth.models import User
+from app.modules.catalog.datasets.domain.source_freshness import (
+    compute_source_freshness,
+)
 from app.modules.catalog.datasets.domain.schemas import (
     DatasetResponse,
     RasterBandInfo,
@@ -164,6 +168,11 @@ def dataset_to_response(
             base_url=base_url,
         )
 
+    # feat(#1218/#1224): resolved once. It is both the `origin` response field
+    # and the gate on source freshness below, and computing it twice is how the
+    # badge and the freshness chip would come to disagree.
+    origin = classify_origin(dataset.source_format, record_type)
+
     return DatasetResponse(
         id=dataset.id,
         record_id=dataset.record_id,
@@ -209,7 +218,7 @@ def dataset_to_response(
         # value could only ever disagree with the two it derives from. Serving
         # it retires the frontend's duplicate rule in OriginBadge.tsx and gives
         # the CLI, MCP server, and SDKs the same answer.
-        origin=classify_origin(dataset.source_format, record_type),
+        origin=origin,
         origin_uri=dataset.origin_uri,
         origin_ref=dataset.origin_ref,
         last_refreshed_at=dataset.last_refreshed_at,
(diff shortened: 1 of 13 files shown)


ishiland:
@codex review

PR #1270, opened 2026-08-08
test(e2e): zoom analysis preview to seeded layer

### LINKED ISSUE #1253: test(e2e): analysis.spec.ts previews against the wrong viewport bbox, and isn't in any smoke subset
## Context

Wave A release-gate verification (Playwright, post-merge on main@a6385000d) ran `e2e/analysis.spec.ts` explicitly per the release runbook (it's excluded from `e2e:smoke:core`) and found 1 failure.

## Root cause

The #727/#1247 capped-preview-honesty fix makes the analysis preview send the current map viewport as the request bbox. The spec clicks "Preview" without first zooming to the seeded dataset (which is in New York); the viewport bbox sent is `[-50.6,-32.5,50.6,59.9]` (mid-Atlantic/Africa), so the preview correctly returns zero features and the UI correctly shows the new honest empty-result toast. This is the intended #727/#1247 behavior working correctly — the spec is out of date, not the product.

## Separate finding: CI gap

`analysis.spec.ts` is not in `e2e:smoke:core` (confirmed in `project_analysis_spec_not_in_smoke_subsets` — this was already a known exclusion, not new), so it never ran on any of the 5 Wave A PRs pre-merge. The #727/#1247 behavior change went out without this spec catching the interaction, and only surfaced because the Wave A release runbook explicitly calls for running it post-merge.

## Fix direction

1. Update the spec to zoom/navigate to the seeded NY dataset before clicking Preview (matching the existing e2e convention: open panel THEN zoom-to-layer).
2. Consider whether `analysis.spec.ts` should move into a smoke subset now that analysis preview behavior is bbox-viewport-sensitive, or whether the explicit-run-when-analysis-code-changes convention is sufficient going forward.

Not a product defect — the honest-empty-result behavior is exactly what #727/#1247 shipped to fix. Test-debt + process gap only.

ishiland:
## Summary

- zoom the map to the seeded New York layer before requesting an analysis preview
- preserve the viewport-bounded preview behavior while restoring a trustworthy E2E signal

## Smoke-suite policy

This PR deliberately leaves analysis.spec.ts out of the e2e:smoke scripts. The full stateful spec performs ingest, materialization, polling, and cleanup, so adding it would materially change the PR smoke policy. Until a dedicated analysis gate or lightweight preview-only smoke is approved, changes affecting analysis or viewport-bbox behavior should run the focused analysis spec explicitly; the nightly catch-all remains the backstop.

## Verification

- npx playwright test e2e/analysis.spec.ts --project=chromium (5 passed)

Closes #1253

### DIFF
--- a/e2e/analysis.spec.ts
+++ b/e2e/analysis.spec.ts
@@ -67,6 +67,12 @@ test.describe('builder analysis tools', () => {
     await page.getByRole('button', { name: 'Analysis', exact: true }).click();
     await expect(page.getByTestId('analysis-panel')).toBeVisible();
 
+    // fix(#1253): preview is clipped to the current viewport, so fit the map
+    // to the seeded New York layer after the panel opens and finalizes layout.
+    await page.getByRole('button', { name: `Layer options for ${seed.title}` }).click();
+    await page.getByTestId('kebab-zoom-to-layer').click();
+    await page.waitForTimeout(1500); // the fly-to animation
+
     // Default operation is buffer @ 500 m; the only dataset layer auto-selects.
     await page.getByRole('button', { name: 'Preview', exact: true }).click();
     await expect(page.getByRole('button', { name: 'Clear preview' })).toBeVisible({


ishiland:
@codex review

PR #1272, opened 2026-08-08
test(e2e): match admin jobs navigation pathname

### LINKED ISSUE #1252: test(e2e): admin.spec.ts:274 exact-URL selector breaks when the failed-jobs badge is non-zero
## Context

Wave A release-gate verification (Playwright smoke, post-merge on main@a6385000d) found `e2e/admin.spec.ts:274` failing: `waitForURL('/admin/jobs')` never resolves.

## Root cause

`AdminSidebar.tsx:251` appends a query string (`?status=failed`) to the jobs link only when the failed-job badge count is non-zero. The dev DB used for this run had 2 failed jobs, so the actual navigation target was `/admin/jobs?status=failed`, which the spec's exact-match `waitForURL('/admin/jobs')` doesn't account for. The page navigates correctly; only the test assertion is too strict.

Predates Wave A — the conditional query-string behavior traces to #1195/#1208, not any Wave A PR.

## Fix direction

Use a prefix/pattern match (`waitForURL(url => url.pathname === '/admin/jobs')` or a regex) instead of an exact string, so the assertion is correct regardless of the failed-job count in whatever DB state the test runs against.

Test-only flake tied to DB seed state, not a product defect — the admin sidebar navigation itself works as intended.

ishiland:
## Summary

- match admin jobs navigation by pathname instead of the full URL
- accept both the badge-zero /admin/jobs target and the badge-nonzero /admin/jobs?status=failed target
- preserve the existing conditional sidebar link behavior, which already has unit coverage for both hrefs

## Verification

- cd frontend && npm run typecheck
- cd frontend && npm run lint
- cd frontend && npx vitest run (380 files, 4,669 tests passed)
- npx playwright test e2e/admin.spec.ts --project=chromium (15 passed)

Closes #1252

### DIFF
--- a/e2e/admin.spec.ts
+++ b/e2e/admin.spec.ts
@@ -285,7 +285,7 @@ test.describe('Admin Panel', () => {
     ).toBeVisible();
 
     await page.getByRole('link', { name: 'Jobs' }).click();
-    await page.waitForURL('/admin/jobs');
+    await page.waitForURL((url) => url.pathname === '/admin/jobs');
     await expect(
       page.getByRole('heading', { level: 1, name: 'Jobs' }),
     ).toBeVisible();


ishiland:
@codex review

PR #1275, opened 2026-08-08
test(auth): isolate verification config state

### LINKED ISSUE #1251: test(auth): register-verification-flow test leaks across xdist workers via shared test DB
## Context

Wave A release-gate verification (post-merge, main@a6385000d) found `tests/test_register_verification_flow.py` fails 2/2 under `-n 4` (full suite), both `403 \"Registration is disabled\"` where `201` was expected. The same file passes 19/19 when run in isolation.

## Root cause

`PersistentConfig.get()` resolves its value from the DB *before* falling back to the env default. The test helper patches the env default for `registration_enabled`, but under `pytest -n 4` the test DB is shared across xdist workers — a sibling worker's `PersistentConfig` write (setting registration disabled for its own test) is visible to this worker's DB-first read, overriding the env patch this test relies on.

Confirmed this is not a Wave A regression: PR #1244's changes are confined to the login path and never touch the registration gate or `PersistentConfig`. All five Wave A PRs passed CI (which also runs under `-n 4`) pre-merge — this is either a pre-existing intermittent flake or a fixture-ordering coincidence that happened to surface during this pass.

## Fix direction

The test needs to set `registration_enabled` via the same DB-backed `PersistentConfig` path the app reads from (not just patch the env default), or the fixture needs per-worker isolation for this config key specifically.

Test-only, no product code path affected. Wave A's own gates and CI already passed; this surfaced only under Wave A's extra whole-tree verification pass.

ishiland:
## Summary

- write registration test overrides through the same DB-backed `PersistentConfig` path used by the application
- clear every touched setting during fixture teardown so committed rows cannot leak to later xdist work
- keep the fix local to the affected test module rather than adding global teardown

## Root cause

The tests patched environment defaults even though `PersistentConfig.get()` prefers a committed database value. Because xdist workers share the test database, a setting committed by another test could override that patch and survive into later work on the same worker.

## Verification

- full backend suite after rebasing over #1274: 7,676 passed, 89 skipped
- full register-verification module under `-n 4`: 19 passed, repeated three times
- seeded leakage scenario under xdist: 4 passed, repeated three times
- sequential register-verification module: 19 passed
- `cd backend && uv run ruff check .`
- `cd backend && uv run ruff format --check .`

Closes #1251

### DIFF
--- a/backend/tests/test_register_verification_flow.py
+++ b/backend/tests/test_register_verification_flow.py
@@ -142,35 +142,25 @@ def verification_required_off(monkeypatch):
 
 
 # ---------------------------------------------------------------------------
-# Helper: force PersistentConfig to bypass its DB cache and return a value
+# Helper: set PersistentConfig through the same DB-backed path as the app
 # ---------------------------------------------------------------------------
 
 
-def _patch_persistent_config(monkeypatch, config_obj, value: bool):
-    """Monkeypatch a PersistentConfig[bool] to always return *value* from .get().
+@pytest.fixture
+async def set_persistent_config(test_db_session):
+    """Commit config overrides for a test and remove those rows afterward."""
+    touched_configs = []
 
-    PersistentConfig.get() resolves: env_only → cache → DB → env_default.
-    In integration tests the DB starts without an AppSetting row, and the
-    in-memory cache is cleared per test, so get() falls through to env_default.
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review

ishiland:
@codex review

Issue #1225, opened 2026-08-05
feat(frontend): Source panel on every dataset detail page


ishiland:
Blocked by #1217 and the backend issues in this milestone (source-state fields, refresh endpoint, health, history). Line refs as of 2026-08-05.

The Sources tab exists only for VRTs: `frontend/src/components/dataset/panels/DetailPanel.tsx` gates `showSources = isVrt` (~:61). For every other dataset, source information scatters: an origin badge in the header, an Overview details card that omits `source_url` entirely (`tabs/OverviewTab.tsx:207-226`), and version history on the Metadata tab.

## Scope

A Source panel on every dataset detail page:

- Origin and pointer, with storage-mode copy per the #1217 vocabulary (owned copy vs registered in place vs remote reference).
- Last refreshed, last checked, health, drift status, feature count.
- Refresh action (gated on write access) driving the new endpoint through the existing preview-then-commit dialog flow.
- Refresh history list.
- The VRT member table becomes this panel's VRT-specific section instead of a separate one-off tab.

## Where to look

- Tab gating and layout: `DetailPanel.tsx:53-96` (per-type booleans like `isVrt` at the top; add the new tab or panel here).
- The VRT table to absorb: `tabs/SourcesTab.tsx` (member rows with health at ~:305; "Regenerate" at ~:207).
- Existing pieces to reuse rather than rebuild: `OriginBadge.tsx` (origin chip), `SchemaDiffView.tsx` (drift rendering), `ReuploadDialog.tsx` (the preview-then-commit dialog; the refresh button should open the same flow with the origin pre-resolved), `ProvenanceTimeline` in `tabs/OverviewTab.tsx:95-138` (history row rendering).
- API calls: follow the existing pattern in `frontend/src/api/` (everything goes through `apiFetch()` from `api/client.ts`); add the new endpoints beside `api/vrt.ts` and `api/datasets.ts`.
- Types: `frontend/src/types/api.ts` (hand-maintained mirror; also run `npm run types:check`, the generated-types drift gate).
- Copy: `frontend/src/i18n/locales/{en,es,fr,de}/dataset.json`; every new key goes to all four locales or `npm run test:i18n` fails. A `defaultValue` alone fails the gate.
- UI primitives: use `frontend/src/components/ui/` components, not hand-rolled ones.

## Suggested path

1. Build the panel read-only first (fields, health, history) behind the existing tab structure; get the four-locale copy in from the start.
[...]

Issue #1221, opened 2026-08-05
feat(ingest): raster replace/refresh path (currently rejected at the reupload gate)


ishiland:
Blocked by #1217, specifically its raster source-object lifecycle decision (which absorbs the question parked in #1210). Line refs as of 2026-08-05.

Raster datasets reject re-upload outright: the eligibility gate at `backend/app/modules/catalog/datasets/api/router_reupload.py:176-183` answers "Import a replacement raster dataset instead", and VRTs are excluded separately at :167-174 (edit membership instead). So refreshing a raster today means delete plus re-import, which loses the dataset id, metadata, grants, and every map layer referencing it.

## Scope

A raster replace path: new source file, COG conversion, asset swap on the existing `RasterAsset`, tile-cache invalidation, refresh-history row, `last_refreshed_at`. VRT parents keep their existing regenerate flow; a replaced member should surface on the parent's health.

Hard requirement: if this adds a presigned door, it must enforce the completion-time content-validation contract established by #1202 and extended to reupload by #1207, with the same error taxonomy as the direct path. Two doors already had to be retrofitted; do not add a third that skips it.

## Where to look

- Eligibility gate to relax: `router_reupload.py:150-183`.
- How rasters ingest today: `backend/app/processing/ingest/tasks_raster.py` (task `ingest_raster` at :312; COG conversion around :178). The replace task should reuse these internals.
- Asset model and swap target: `backend/app/processing/raster/models.py` (`RasterAsset`: `asset_uri` :54, `sha256` :55, `cog_status` :68, `status` :90; note `last_regenerated_at` :96, the VRT-rebuild precedent for a lifecycle timestamp).
- Tile-cache invalidation: `bump_tile_cache_version()` on Dataset (`datasets/domain/models.py:499`); grep existing callers to see when it is called.
- Presigned completion contract to copy: `complete_presigned_upload` in `backend/app/processing/ingest/router.py` (validated at completion since commit `ed0ee95fa`); PR #1213 shows the reupload-door version.
- VRT health surface: `router_vrt.py` status endpoint (the parent should show a regenerating/changed member sanely).

## Suggested path

1. Wait for #1217's decision on source-object retention (#1210: keep the original upload for re-processing vs delete after COG conversion). It changes what "replace" deletes.
[...]

Issue #998, opened 2026-07-30
fix(ops): make 0019 tenant-ownership adoption reachable without a destructive downgrade


ishiland:
Surfaced while correcting the multi-tenant DR runbook in #950 / #965. That PR documents the current procedure honestly, including its limits. This issue is about the underlying gap it exposes, which documentation cannot fix.

> **Split 2026-07-31.** The security-relevant piece, 0029's unguarded downgrade, is now **#1016** and should ship on its own. It is a scoped guard on one migration and is worth doing whichever direction this issue eventually takes, so it does not need any of the decisions below.

## The gap

The fresh-cluster multi-tenant DR path has exactly one mechanism for reconstructing ownership: `alembic downgrade 0016`, then adopt 0019. But five migrations between the current head (0030) and 0019 either refuse to run or destroy data the current schema legitimately holds, so on any cluster carrying real data that downgrade is not a recovery procedure, it is a data-loss event.

What sits in the way, walking back from head:

- **0030** — antimeridian extents are stored as MULTIPOLYGON; the downgrade refuses rather than coercing them (deliberately, per #946).
- **0029** — drops `api_keys.expires_at` and both `key_epoch` columns. This one is the worst of the set because it fails *silently and in the unsafe direction*: time-limited keys come back permanent, and revoked keys come back live. Now tracked separately as **#1016**.
- **0027** — drops the parquet / json / xlsx / xls source formats.
- **0022** — drops tenant attribution on `audit_logs` and `ingest_jobs`.
- **0021 + 0020** — reintroduce per-tenant name collisions.

So an operator following the documented recipe on a populated cluster either hits a refusal at 0030 and stops, or, if they force past it, silently resurrects revoked API keys.

## Two facts from RUNBOOK that constrain any solution

**The obvious workaround does not work.** Skipping the downgrade by stamping the schema and replaying 0019 forward fails: 0019 installs its functions with plain `CREATE FUNCTION`, so `alembic stamp 0018 && alembic upgrade 0019` collides with the functions the restored dump already carries (`RUNBOOK.md:244-246`). Any forward-only adoption path has to solve that collision, which means either `CREATE OR REPLACE` semantics or an adoption step that is not 0019 replayed.

[...]

PR #1277, opened 2026-08-08
feat(ingest): one-request service refresh from the stored origin

### LINKED ISSUE #1220: feat(ingest): one-request service refresh from the stored origin
Blocked by #1217 (origin pointer, `duplicate_source` interplay) and the source-state fields issue in this milestone. Line refs as of 2026-08-05.

Re-pulling a service dataset today runs through the re-upload dialog and fights the product the whole way:

- The service preview endpoint requires the URL in the request body (`backend/app/modules/catalog/datasets/api/router_reupload.py:284-384`) instead of reading the stored origin server-side; the dialog merely prefills it client-side (`frontend/src/components/dataset/ReuploadDialog.tsx:123`).
- The user re-picks the layer even though the origin was stored with its layer suffix at ingest (`backend/app/processing/ingest/tasks_vector.py:1005-1007`).
- Tokens are deliberately never persisted, so protected sources need a fresh token each time. That part is correct and stays.
- Importing the same URL through the import flow 409s with code `duplicate_source` (`backend/app/modules/catalog/sources/router.py:820` region) and tells the user to delete the existing dataset first.

## Scope

- A refresh endpoint (name per the ADR): server-side re-pull from the stored origin pointer through the existing preview and commit contract, with no client-supplied URL and no layer re-selection.
- Transient token accepted per call, used, never stored.
- `duplicate_source` updated so a dataset's own origin never counts as a duplicate of itself; true duplicates still 409.
- On commit: refresh-history row, persisted schema diff, `last_refreshed_at` set.

## Where to look

- Existing service re-pull flow end to end: `router_reupload.py:284-384` (preview) then the shared preview/commit handlers below it, then worker task `reupload_service` in `backend/app/processing/ingest/tasks_reupload.py:461` (it re-enriches the URL near :693).
- Duplicate guard: the query at `sources/router.py` matching `(Dataset.source_url, Dataset.source_format, Record.created_by)`; the fix is excluding the refreshing dataset's own id.
- Access control: `check_dataset_write_access` in `backend/app/modules/catalog/authorization.py` (this is a mutation endpoint).
- Frontend caller to adapt afterward: `ReuploadDialog.tsx` service branch (a follow-up inside the Source panel issue wires the button; this issue only needs the API correct).

## Suggested path

1. Add the refresh endpoint beside the reupload routes. Resolve origin + layer from the dataset row, then reuse the existing service-preview internals rather than duplicating them (extract a helper if the handler is monolithic).
2. Thread an optional `token` through exactly like the current preview does; add a distinct error for "source requires a token" so the UI can prompt.
3. Fix the duplicate guard with a `Dataset.id != <refreshing id>` exclusion and keep a regression test for the true-duplicate case.
4. Tests: refresh happy path (public source), token-required path, self-refresh not blocked, cross-dataset duplicate still blocked, error taxonomy matches the existing reupload contract.
5. Regen trio.

## Acceptance

- [ ] Public source refreshes with zero re-entry (no URL, no layer picking)
- [ ] Protected source prompts for a transient token; nothing persisted (assert no new credential columns/rows)
- [ ] Self-refresh no longer trips `duplicate_source`; cross-dataset duplicates still 409
- [ ] Rule 1 write-access check present; `tests/test_rule1_structural.py` green
- [ ] OpenAPI regen trio committed

## Verify

```bash
cd backend && set -a && source ../.env.test && set +a && uv run pytest tests/ -k "refresh or reupload_service" -v
uv run pytest tests/test_rule1_structural.py tests/test_layering.py -q
cd .. && make openapi && make sdks && make openapi-check && make sdks-check
```

Live check against the dev stack (`make dev`): import a public ArcGIS FeatureServer layer via Import → Service URL, then hit the new endpoint and confirm feature counts update and a history row appears.

## Traps

- The "tokens are never persisted" promise is not literally true across the API/worker boundary today: the existing service ingest and reupload paths pass tokens inside durable Procrastinate task arguments, and failed job rows can retain them in PostgreSQL. The refresh path must not add another durable copy of the token, and scrubbing token fields from task arguments or failed-job payloads on the paths this issue touches is in scope.
- Service reupload duplicates parts of the staging flow rather than sharing one executor; prefer extracting the common helper over copying the handler a third time.

## Non-goals

Scheduling, stored credentials, incremental sync. Those are Enterprise connector-overlay capabilities per EDITIONS.md and the locked boundary in #1217.

### LINKED ISSUE #1268: feat(refresh): audit events and operational metrics for refresh runs
ADR-002 Amendment A10 (audit blockers 9 and 12). Refresh-run rows are mutable and cascade with datasets, so they are not an audit ledger; and no metrics exist for the refresh lifecycle.

## Scope

**Audit events** (same transaction as the run transition, action-registry entries + tests):
- `refresh.dispatch` on run-row insert, `refresh.succeeded` / `refresh.failed` on terminal transitions, `refresh.abandoned` from the stale-run sweep.
- Payloads never include URLs, origin_ref contents, tokens, provider text, or schema diffs — ids, origin kind, status, error_code only.

**Metrics** (Prometheus, multiprocess-safe per #1245):
- queue wait (`claimed_at - started_at`), run duration by strategy and outcome, active-run count, probe duration/outcome (#1222's endpoint), sweep reconciliation count.

## Disposition note

Amendment A10 allows folding this into #1220 if its codex-round budget holds; in that case close this issue with #1220's PR. Otherwise implement standalone after #1220 merges.

## Acceptance

- [ ] action-registry drift test covers the new actions (the #1244 pattern)
- [ ] sentinel test: a refresh of a token-bearing service origin leaves the token absent from every audit payload
- [ ] /metrics shows the new series under multi-worker mode

ishiland:
Adds `POST /datasets/{dataset_id}/refresh`: a single request that re-ingests a service dataset from its stored origin binding. No URL, no service type, no layer name in the request; the server reads all of it from `origin_ref`. The request model has exactly one optional field (`token`), and a test pins `model_fields == {"token"}` so a client-supplied address can never grow back.

Closes #1220. Closes #1268 (second commit, independently revertable).

## How it works

- **Admission** goes through `create_pending_run(trigger="api")`, the same function and the same `uq_refresh_runs_one_active` partial index `reupload_commit` uses. There is no second admission path. The AST test that pins run-creating handlers now names both callers and states the property: no handler that merely looks at an origin reserves the dataset.
- **Ordering**: the binding that gets dispatched is re-read *after* the reservation is taken. The worker commits swap + rebind + terminal run in one transaction, so once this request wins the reservation there are only two cases: the other run still holds it (this request gets 409 `dataset_busy`) or its rebind is committed and the re-read sees it. A binding that changed in the window is refused with 409 `origin_changed` rather than silently dispatching either version; the race regression tests were verified to fail against the old snapshot-first ordering.
- **Execution** dispatches the existing reupload worker, so attempt-scoped staging and swap-on-success apply unchanged. A failed refresh never touches the live table.
- **Refusals**: 409 `refresh_not_applicable` (no remote service origin; carries `origin_kind`), 409 `origin_unavailable` (binding too incomplete to re-address the layer), 409 `dataset_busy` (active run exists), 400 (stored URL fails SSRF validation), 503 `credential_store_unavailable` (see below).
- STAC and raster origins are refused with `refresh_not_applicable` carrying their `origin_kind`, so the raster path (#1221) becomes a branch behind this same door rather than a new endpoint.
- Note for #1225: `origin_changed` joins `dataset_busy` in the ProblemDetail detail codes the Source panel will want to render. No OpenAPI change.

## Credential handoff

A supplied token is stashed in Valkey under an unguessable single-use reference (`SET ... EX <ttl> NX`); only the reference travels in the task args. The lifetime is bounded by renewal rather than a constant: renewal re-arms the short TTL (via `EXPIRE`, which cannot resurrect a claimed key) exactly while the task is still claimant-reachable (`todo` or `doing`) and the run is still active — the abandonment sweep's own liveness definition in both dimensions, pinned to it by a consistency test so the two predicates cannot drift apart. `doing` is included because the status flip precedes the claim: the worker revalidates its URL for SSRF (an unbounded DNS resolution) before it redeems the credential, and renewal must survive that window. It is safe to include because `EXPIRE` cannot resurrect: once the claim's `GETDEL` removes the key, every later renewal cycle is a no-op, so renewal self-terminates at the true claim event rather than at a status flip that merely precedes it — a property asserted by a test, not just stated here. There is deliberately no age constant: any constant reproduces the "healthy but longer than the number" failure one level up, so the credential lives precisely as long as a claimant-reachable task exists and dies the instant the task or run leaves that state. No constant can bound a healthy long import (the heartbeat keeps it alive past any timeout), which is why the earlier derivation was replaced. Renewal is dual-hosted — the API's background sweeper and a worker-side loop both call the same helper (`EXPIRE` is idempotent), so the handoff survives either role being briefly down; the worker is the process whose liveness gates the claim, and both roles down past the TTL is the stated accepted floor, since no claimant existed during the credential's entire life. The renewal query filters explicitly on `ingest_jobs.tenant_id` (kept equal to the parent dataset's by `trg_validate_ingest_job_parent_tenant`) rather than relying on future RLS, and runs inside each tenant's context, mirroring the stale-job sweep's branching: without the scoping it would read across tenants today (no `catalog` table has RLS enabled yet; enablement is #998), and would silently stop renewing the moment #998 lands. The multi-tenant test enables FORCE RLS on the real boundary and drops to a non-BYPASSRLS role, so it guards the post-#998 world, and a renewal-query failure now logs `error_type` at warning instead of vanishing into the never-raises contract. The worker redeems it with `GETDEL`, so a second claimant loses atomically. A retry that runs after the credential was consumed fails as `credential_expired`. The token never appears in task args, `user_metadata`, log output, or any committed row; sentinel tests search all of them. A contract test pins the literal `SET NX`/`GETDEL` calls so a refactor to get-then-delete, which reopens the two-claimant window, fails a test rather than a review.

Claim-side failures are classified by evidence: only the store answering "no such key" reports `credential_expired`; a store that cannot be reached (or any unexpected backend failure) reports `credential_store_unavailable`, since an unreachable store is evidence of nothing about the credential.

One product decision to be aware of: Valkey is opt-in in a default install (`cloud-dev` compose profile), so a default install has no shared credential store. A token-bearing refresh is refused up front with 503 `credential_store_unavailable` rather than dispatched into a run that could never authenticate. Public-source refresh, the issue's primary acceptance criterion, needs no store and is unaffected.

The API door and the worker now judge tokens by one policy (`app/core/service_tokens.py`). Previously the request model accepted any printable whitespace-free token while the worker pinned header-auth tokens (WFS, OGC API Features) to a base64url charset with an 8-character floor, so a token containing `+` or `/` was accepted with 202, burned its single-use credential, and failed deterministically in the background. The endpoint now applies the policy the worker will apply, selected by the service type of the post-reservation binding (ArcGIS keeps its wider vocabulary: its token travels as a urlencoded query parameter, never an Authorization header line, so it carries none of the header-smuggling risk the charset exists to prevent), and rejects with 422 `invalid_service_token` before the credential is stashed. The 422 body describes the policy and never echoes any part of the input. The worker keeps enforcing the shared definition at the point of use, since the guarantee is about what reaches libcurl and must not rest on a validator two processes away. Its ValueError still names the first offending character, which is the right diagnostic for whoever debugs a failed job; that puts one character of the submitted string into `error_message` (a control or punctuation character, not credential material, and the exact-value scrub still removes the token itself). The store-availability gate runs before token validation, so a store-less deployment answers 503 for any token-bearing refresh, well-formed or not; that deployment cannot do credentialed refresh at all, so store availability is legitimately the first question.

Failure text can never carry a token: the worker scrubs the claimed secret's exact value (including percent-encoded spellings) from any exception before it reaches a sink, and `redact_url_credentials` is applied both where `run_ogr2ogr_service` composes its error and at the single boundary feeding `error_message`, logs, and notifications. Both layers are necessary (the log sink renders the live traceback, which only value-scrubbing reaches; the boundary scrubber covers tokens this process never held) and the sink tests were verified to fail with the token visible before the fix. This also closes the same pre-existing leak in the re-upload path's failure trail.

One rolling-deploy window is accepted and documented at the dispatch site (same precedent as #1274's commit-to-defer gap): an old-generation worker claiming a credentialed refresh discards `credential_ref` and fetches unauthenticated, yielding a prompt auth-failure on the run with the dataset released immediately, and the stranded credential expires as soon as renewal stops. The alternative (a version-scoped task name old workers can't claim) fails cleanly on the queue but leaves the job and run pending, holding the dataset against admission for up to an hour of apparent hang — a worse failure than an immediate, actionable error. Single-node compose never overlaps generations.

Deliberately out of scope: `reupload_commit` still passes its token as a durable task argument (its remaining exposure is now the queue row only, no longer the failure trail). Converting it would break protected re-upload on store-less default installs, trading a latent problem for a live regression. The new endpoint makes no durable-token promise; the old one keeps its shipped contract until the credential store is a hard dependency.

## Corrections that ride along (both touch shipped behavior)

- `record_refresh_failure(contacted_origin=True)` now requires the origin binding and stamps the dataset with a binding-matched UPDATE (`origin_uri` + `origin_ref` + `source_format`, `IS NOT DISTINCT FROM`). The unguarded ID-only shape flagged in #1274's description is now unreachable; passing `contacted_origin=True` without a binding raises `ValueError`.
- Both duplicate-source guards (service preview, STAC import) re-key to `origin_uri`, with `source_url` only as the un-backfilled fallback. `source_url` is PATCHable, so an owner could previously edit past the guard and register the same layer twice. Regression tests cover both directions.
- A refresh preserves the stored pointer spelling. WFS/OGC imports store `origin_uri`/`source_url` as `base_url/typename`; the refresh dispatch now carries the same layer identity the import did, so those columns stay byte-identical across a refresh (the fetch itself is unaffected — GDAL source composition reads the layer name for these services). Without this, the first refresh rewrote the pointer to the bare base URL and the duplicate guard stopped matching the layer. The dispatched pointer is asserted against the import's own composition expression, not a literal, so the two paths cannot drift silently. A follow-up remains worth doing: the guards are still keyed on a string spelling, and re-keying them onto the canonical `origin_ref` URL/layer pair would make them robust to any future writer; that change wants its own failing-first coverage (never-refreshed and un-backfilled paths) and is deliberately not folded in here.

## Python SDK: optional-body fix (repo-wide)

Wiring the new endpoint into the generated Python SDK exposed a generator defect that predates this PR: every optional-body method assigned the `UNSET` sentinel to `json` when the caller omitted the body, so httpx raised `TypeError` before the request was sent. Five endpoints carried it, four of them shipped (map share, reupload preview, OGC collection items, dataset search). The fix is in the `make sdks` pipeline, not the generated files: a post-generation step rewrites the unguarded assignment and exits nonzero if a future generator bump changes the emission shape so the guard no longer matches. Explicit `None` still serializes to `null`; the TypeScript SDK is structurally unaffected. Regression tests scan every optional-body endpoint (with a vacuity guard) and exercise the previously-raising request build. A path-filter gap flagged by `test_ci_alembic_filter_paths` is also fixed so a PR editing only the guard script still runs Backend Tests.

## Observability (#1268)

Audit events for the refresh lifecycle (dispatch, success, failure, cancel), registered in the action registry the #1244 coverage test checks.

The Prometheus surface is mostly derived gauges rather than per-event counters, on purpose. Terminal transitions happen in the worker, which serves no `/metrics` and shares no `PROMETHEUS_MULTIPROC_DIR` with the API; and every uvicorn worker runs the poll loop while `MultiProcessCollector` sums counters, so poll-and-increment would report N times the truth (the fabricated-number class #1240 removed). Queue wait and duration are SQL `percentile_cont` over a trailing hour published `livemostrecent`; true counters exist only where one request on one worker observes the event (sweep, probe). A test pins the metric declarations.

## API/schema impact

New endpoint plus `DatasetRefreshRequest`/`DatasetRefreshResponse`. `backend/openapi.json`, both SDKs, and `frontend/src/types/api.generated.ts` regenerated, nothing hand-edited. No migration. Frontend is not wired to the endpoint; #1225's Source panel is the consumer and the SDK/TS types are ready for it.

## Verification

- `uv run pytest -q -n 4`: 7776 passed, 88 skipped (whole-tree at the current head, rebased onto main)
- `uv run ruff check .` / `uv run ruff format --check .`: clean
- `tests/test_layering.py`: 40 passed (four cap raises, each with a justification comment)
- `make openapi-check`, `make sdks-check`, `make cli-test`: exit 0
- frontend `npm run build` / `lint` / `typecheck` / `test:i18n`: clean; admin component tests 93 passed
- `test_rule1_structural.py`, `test_rule2_structural.py`, audit action registry: green


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1278, opened 2026-08-08
feat(mcp): expose dataset source trust metadata

### LINKED ISSUE #1228: feat(mcp): expose origin, freshness, and health in MCP and SDK reads
Blocked by the source-state fields and staleness issues in this milestone.

An agent deciding whether to trust a dataset needs exactly the fields this milestone adds: origin, freshness, health, last refreshed. `mcp/geolens_mcp` is hand-maintained (like the CLI) and read-only; the SDKs regenerate from `openapi.json` and pick the fields up mechanically. No competing catalog product ships an MCP server at all, so keeping ours current with the trust metadata is cheap differentiation.

## Scope

- MCP: include origin, freshness state, health, and `last_refreshed_at` in `search_datasets` and dataset-detail outputs, with docstrings that tell a model what the fields mean and when to distrust a dataset (e.g. "overdue means the data has not been refreshed within its declared update cadence").
- SDKs: verify the regen carries the fields and the hand-maintained wrappers need no change.

## Where to look

- The MCP package: `mcp/geolens_mcp/` (small; find the tool definitions for `search_datasets`, `get_dataset_schema`, `get_map`, `list_maps`, `get_features` and see how each shapes its output from SDK responses).
- It depends on the `geolens` Python SDK (`sdks/python/`), which is generated; after the backend fields merge and `make sdks` runs, the SDK models already carry the new fields. The MCP change is choosing to surface them and writing the docstrings.
- Only `auth.*`, `__init__`, and `index` wrappers are hand-maintained in the SDKs; everything else is generated. Do not hand-edit generated files.

## Suggested path

1. Wait for the backend fields + staleness issues to merge with their regen.
2. Bump the MCP package's SDK dependency floor if it pins one, add the fields to tool outputs, and write the docstrings (they are the agent-facing documentation; be concrete).
3. Add or extend the package's tests for the new output shape.
4. Live check against the dev stack: run the MCP server, call `search_datasets`, confirm the fields appear.

## Acceptance

- [ ] MCP tool outputs include origin, freshness, health, `last_refreshed_at`, with meaningful docstrings
- [ ] Package tests cover the new output shape
- [ ] SDK regen committed under the trio rule (in the backend PR, not this one, unless this PR touches the API)
- [ ] Live call against the dev stack shows the fields end to end

## Verify

```bash
# package tests: see mcp/ for its test runner (pytest)
cd mcp && uv run pytest -v
```

Live: configure the MCP server against `http://localhost:8080/api` with a dev token and call `search_datasets` from any MCP client; confirm the new fields and sensible values for an overdue dataset.

## Traps

- MCP tool results echo verbatim into a model's context; keep field names and docstrings unambiguous, and never include raw provider URLs or tokens in outputs.
- Long-lived MCP sessions outlive short JWT expiries; nothing new here, but do not design the new fields to require extra authenticated round trips per row.

ishiland:
## Summary

- add stable agent-facing `source_origin`, freshness, health, checked, and refreshed fields to MCP dataset search and detail reads
- derive a coarse provider-safe origin from search summary fields without an authenticated detail request per row
- return explicit nulls when search summaries do not carry health or timestamps, while dataset detail returns the populated backend state
- document freshness and health semantics in MCP tool docstrings so agents can judge dataset trust correctly
- strip raw source URLs and origin pointers from the shaped MCP results

## API and SDK impact

- MCP-only change; no backend API or schema changes
- generated Python SDK detail models already contain the source-state fields
- current catalog search summaries expose source format and computed freshness but not health/check/refresh values, so those search keys are stable nulls until the summary contract grows
- no OpenAPI regeneration, generated SDK edits, or hand-maintained SDK wrapper changes

## Verification

- `make mcp-test` — 27 passed, 6 live-only skipped
- `make mcp-live-test` — 6 passed against the running API
- `make mcp-build` — wheel and sdist built successfully
- Ruff check and format check over `mcp/geolens_mcp` and `mcp/tests`
- pre-commit hooks on all changed files

Closes #1228

### DIFF
--- a/mcp/README.md
+++ b/mcp/README.md
@@ -51,8 +51,8 @@ Cursor / Codex / any client that reads an `mcpServers` block:
 
 | Tool | What it does |
 |---|---|
-| `search_datasets` | Catalog search by free text (semantic ranking where the instance enables it). Returns dataset records as GeoJSON features. |
-| `get_dataset_schema` | A dataset's columns, geometry type, CRS/SRID, feature count, and extent. |
+| `search_datasets` | Catalog search by free text (semantic ranking where the instance enables it). Returns dataset records as GeoJSON features with safe origin and freshness state; health/check/refresh keys are null when unavailable in the search summary. |
+| `get_dataset_schema` | A dataset's columns, geometry type, CRS/SRID, feature count, extent, and source trust metadata. |
 | `get_features` | Bounded GeoJSON features for a dataset (OGC API — Features), with optional bbox. |
 | `list_maps` | Saved maps (id, name, visibility, layer count). |
 | `get_map` | One saved map's full metadata, including layers and view state. |
--- a/mcp/geolens_mcp/client.py
+++ b/mcp/geolens_mcp/client.py
@@ -28,6 +28,21 @@ import httpx
 
 DEFAULT_TIMEOUT = 30.0
 
+# Kept local because catalog search currently exposes ``source_format`` but not
+# the backend's computed ``origin`` field. These values mirror the public source
+# formats accepted by GeoLens; the MCP response exposes only the coarse origin
+# kind, never the provider pointer used to refresh it.
+_SERVICE_SOURCE_FORMATS = frozenset({"wfs", "arcgis_featureserver", "ogcapi_features"})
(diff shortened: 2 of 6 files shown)


ishiland:
@codex review

Issue #1191, opened 2026-08-04
sec(titiler): give the sidecar a read-only, prefix-scoped object-store credential


ishiland:
Carried over from the 2026-07-02 TiTiler assessment and still open after the 2026-08-04 reassessment.

## Problem

The titiler container holds the same object-store credential as the api and worker: `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` (plus the Azure trio), granting full read and write on the bucket. It performs no authorization of its own and never needs to write.

That matters because GDAL-level local file read is the live exploit class for this stack. Enclave's "MapRoot" chain escalated exactly this way, and a read inside the container reaches `/proc/self/environ`. Today that would hand an attacker a credential that can also write.

## Change

Give titiler its own read-only credential, scoped to the raster prefix, distinct from the api and worker credential:

- `docker-compose.yml` and `docker-compose.prod.yml`: separate `TITILER_S3_*` inputs rather than reusing `S3_*`
- `geolens-deployments` chart: a distinct `secretKeyRef`, defaulting to the shared one so existing installs keep working
- Document the MinIO policy in the deploy guide; self-hosters can express this with a MinIO policy

## Notes

The `#579` sh-wrapper derives GDAL's `AWS_*` trio from the app's `S3_*` settings, so the derivation inputs change with the credential. The chart does the same at template time.

Blast radius is contained today by `read_only`, `cap_drop: ALL`, and no published port. This narrows what a successful read yields rather than preventing the read.

ishiland:
Reopening until the committed geolens-deployments Helm companion lands. Core support merged in #1284; the external chart still needs its serial review and merge gate.

PR #1284, opened 2026-08-08
sec(titiler): add scoped object-store credentials

ishiland:
## Summary

- let TiTiler use a dedicated operator-supplied S3 access key and secret
- preserve the shared `S3_*` credential as an upgrade-compatibility fallback
- ensure only the selected credential enters the sidecar as `AWS_*`
- fail closed before Uvicorn when only half of the dedicated pair is configured
- point operators to the companion deployment guide for the prefix-scoped policy
- add fail-closed Compose and environment-documentation contracts

## Security impact

TiTiler only reads managed raster objects. Operators can now give it a distinct principal with `GetObject` access to the managed raster prefixes instead of exposing the API/worker writer credential. No real or known-public credential is included; every value remains operator supplied.

Existing installations continue to render tiles when the dedicated pair is unset. Once both `TITILER_S3_ACCESS_KEY_ID` and `TITILER_S3_SECRET_ACCESS_KEY` are set, the shared writer credential is not also present in the TiTiler environment. Two non-secret presence flags make a partial rollout fail before Uvicorn starts, preventing mixed-principal AWS signing. The existing Azure mapping is unchanged.

## Compose impact

Both development and production Compose files parse with every profile against a verbatim `.env.example`. Default and shared-fallback AWS credential values remain compatible; the only default rendered additions are two empty, non-secret presence flags on TiTiler. Supplying the dedicated pair changes only the TiTiler service. The MinIO entrypoint guard, parse-safe `${VAR:-}` forms, and every non-TiTiler service remain unchanged.

## Helm companion

A backward-compatible `titiler.s3Credentials` Secret selector and the prefix-scoped MinIO policy are committed in `geolens-deployments` and will land immediately after this core PR. This PR links #1191 without completing the issue.

## Verification

- development and production Compose config with all profiles against verbatim `.env.example`
- normalized default, dedicated-pair, shared-fallback, and partial-pair rendered-config comparisons
- failing-first partial-pair startup regression; corrected suite: 32 passed, 4 skipped
- `make env-doc-check`
- `make public-surface-check`
- full backend Ruff and format checks
- pre-commit hooks on all changed files
- `git diff --check`

Refs #1191

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -535,6 +535,21 @@ FRONTEND_PORT=8080
 # Type: string | No default
 # S3_SECRET_ACCESS_KEY=
 
+# [RECOMMENDED] Dedicated TiTiler access key ID. Grant this principal read-only
+# access to rasters/* and tenants/*/rasters/* in S3_BUCKET; see the deployment
+# guide for the policy:
+# https://github.com/geolens-io/geolens-deployments#titiler-read-only-s3-credentials
+# When unset, Compose falls back to S3_ACCESS_KEY_ID so existing installs
+# continue to render tiles while the dedicated credential is provisioned.
+# Type: string | Compatibility fallback: S3_ACCESS_KEY_ID
+# TITILER_S3_ACCESS_KEY_ID=<read-only-key-id>
+
+# [RECOMMENDED] Secret for the dedicated TiTiler read-only principal. Keep it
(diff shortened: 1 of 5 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1306, opened 2026-08-09
docs(contributing): make the documented backend test commands runnable

ishiland:
Fixes the three contributor-doc drifts from the 2026-08-07 doc audit (DOC-SETUP-01):

- `.github/CONTRIBUTING.md` prescribed a bare `docker compose exec api pytest`, which fails against the container's read-only default uv cache. The full-suite line now points at `make test`, and the filtered/single-file examples use the writable staging uv environment the Makefile uses.
- The single-file example used the host path `backend/tests/test_auth.py`; tests are mounted at `/app/tests` in the container, so the in-container path is `tests/test_auth.py`.
- The component tree named the removed `map-widgets/` directory; it is `map-plugins/` (`frontend/src/components/map-plugins/` exists, `map-widgets/` does not).
- `.github/ARCHITECTURE.md`'s testing bullet repeated the broken bare-pytest command; it now points at `make test` and cross-references the CONTRIBUTING recipe.

Docs-only; no code or config changes.

Verification: the corrected in-container recipe is the same one documented in AGENTS.md ("Running a single test") and used by `make test` (Makefile:78-85); `ls frontend/src/components` confirms `map-plugins/`.


ishiland:
@codex review

PR #1307, opened 2026-08-09
fix(storage): contain local list prefixes

### LINKED ISSUE #1300: fix(storage): LocalStorageProvider.list() bypasses the SEC-026 containment gate
## Problem

`LocalStorageProvider._resolve_contained()` in `backend/app/platform/storage/local.py` documents the SEC-026 contract: "called at the top of every IO method so none is a bypass". Every keyed method honors it except `list()`, which joins the caller-supplied prefix straight onto `base_dir` and then calls `exists()`, `rglob()`, or `glob()`:

```python
search_dir = base_dir / prefix_str   # no containment check
```

So `provider.list("../something/")` reaches the filesystem outside the staging root. The 2026-08-07 internal backend audit reproduced it: with an empty sibling directory present, `list("../outside_empty/")` returned `[]` instead of raising, and a non-empty sibling gets enumerated until `relative_to(base_dir)` raises partway through the walk. Re-verified on `main` at 035b72888.

Current production callers derive prefixes from stored dataset IDs, so there is no known untrusted HTTP path to this today. This is a contract fix so that a future caller, a corrupted row, or a plugin cannot turn it into one.

## Fix

In `list()`, resolve the prefix through `self._resolve_contained(prefix_str)` before any filesystem call, and derive both the search root and the returned keys from the resolved path:

- directory branch (`prefix.endswith("/")`): the resolved path is the search root;
- file-prefix branch: take `parent` and `name` from the resolved path, then glob;
- compute returned keys with `relative_to` against the resolved base, matching what `_resolve_contained` resolved.

`ValueError` propagates exactly as it does from the other methods. The empty prefix (`""`) must keep working; `_resolve_contained` already allows `candidate == resolved_base`.

## Acceptance criteria

- [ ] `list()` raises `ValueError` for: an absolute prefix, a `../` prefix to an absent sibling, an empty sibling, a non-empty sibling, and a prefix containing a null byte. For the non-empty-sibling case, assert the outside file names appear nowhere, including in the raised exception text.
- [ ] Legitimate prefixes still work: `""`, a directory prefix with trailing slash, a file prefix, nested directories, and a no-match prefix returning `[]`.
- [ ] `backend/tests/test_local_storage_containment.py` becomes a parametrized table over every public keyed method of the provider, so the next added method fails the suite unless it is covered. The module docstring claims it covers all I/O methods; make that claim true.

## Verification

```bash
uv run pytest tests/test_local_storage_containment.py -v
```

No database needed, but the env vars are: importing the app boots `app.core.config`, which dies on missing env with an error that looks like a test failure and is not one. `.env.test` is gitignored; create it once with `cp .env.test.example .env.test` from the repo root.

prashit-vora:
## Summary

- Validate local storage list prefixes with `_resolve_contained()` before filesystem access.
- Preserve empty, directory-prefix, and file-prefix listing behavior.
- Add regression coverage for traversal, absolute paths, null bytes, resolved-root prefixes, and all keyed local-storage methods.

Fixes #1300

## Testing

- `uv run pytest tests/test_local_storage_containment.py -q` — 87 passed
- `uv run ruff check app tests` — passed
- `uv run ruff format --check app tests` — passed

### DIFF
--- a/backend/app/platform/storage/local.py
+++ b/backend/app/platform/storage/local.py
@@ -162,28 +162,30 @@ class LocalStorageProvider:
 
     async def list(self, prefix: str) -> list[str]:
         """List keys matching a prefix, relative to base_dir."""
-        base_dir = self.base_dir
-        prefix_str = prefix
+        # SEC-026: resolve the caller-supplied prefix before touching the
+        # filesystem.  Keeping this check outside the worker also ensures a
+        # rejected key never reaches exists(), rglob(), or glob().
+        resolved_prefix = self._resolve_contained(prefix)
+        resolved_base = self.base_dir.resolve()
 
         def _list() -> list[str]:
-            if prefix_str.endswith("/"):
+            if not prefix or prefix.endswith("/") or resolved_prefix == resolved_base:
                 # Directory prefix: list all files recursively under it
-                search_dir = base_dir / prefix_str
+                search_dir = resolved_prefix
                 if not search_dir.exists():
                     return []
                 return [
-                    str(p.relative_to(base_dir))
+                    str(p.relative_to(resolved_base))
                     for p in search_dir.rglob("*")
                     if p.is_file()
                 ]
             # File prefix: glob in the parent directory
-            prefix_path = base_dir / prefix_str
-            parent = prefix_path.parent
+            parent = resolved_prefix.parent
             if not parent.exists():
                 return []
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

ishiland (approved):
LGTM, thanks for the contrib!

Issue #1303, opened 2026-08-09
chore(extensions): remove the deprecated AuditExtension seam at the next EXTENSION_API_VERSION bump


ishiland:
## Problem

The `AuditExtension` seam was deprecated in #836/#873 with removal promised "at the next `EXTENSION_API_VERSION` bump". That promise was written while the version was 2. The version is now 4 (`backend/app/platform/extensions/version.py`), so the seam has survived two bumps, and the compatibility tests still say the symbols must stay "while EXTENSION_API_VERSION == 2", which is now self-contradictory. The registry exposes a dead `audit` slot alongside `AuditSink`, the live audit integration contract, and it is not obvious to an overlay author which one is authoritative.

Surviving pieces, verified on `main` at 035b72888:

- protocol: `AuditExtension` in `backend/app/platform/extensions/protocols.py`
- default: `DefaultAuditExtension` in `defaults_extensions.py`, exported from `defaults.py`
- accessor: `get_audit_extension()` and the package-root aliases in `backend/app/platform/extensions/__init__.py`
- tests: `test_audit_alias_stays_importable_until_version_bump` and the audit-seam rows of `test_pre_pr_import_surface_is_intact` in `backend/tests/test_extensions.py`

## Plan

Do not remove anything while the API version is 4. An overlay built against v2 through v4 may import these names, and removing them mid-version turns into a silent `load_extensions()` skip.

Now, small and unblocked:

- [ ] Re-anchor the stale comments and test docstrings: replace "while EXTENSION_API_VERSION == 2" with "until the next EXTENSION_API_VERSION bump, tracked in this issue", so the tests stop promising a removal point that already passed.

At the next coordinated `EXTENSION_API_VERSION` bump (blocked until one is scheduled):

- [ ] Search the shipped Enterprise/Cloud overlay repositories for `AuditExtension` and `get_audit_extension` imports; removal requires zero hits or coordinated overlay releases.
- [ ] In the same commit as the version bump, delete the protocol, the default implementation and its exports, the accessor, the registry slot handling, the package-root aliases, and both compatibility tests' audit rows.
- [ ] `uv run pytest tests/test_extensions.py -v` passes and `grep -r AuditExtension backend/` returns nothing.

If this slips past version 5, the same deferral has happened a third time; whoever bumps the version owns closing this.

ishiland:
[...]

Issue #836, opened 2026-07-28
chore(backend): maintainability batch — LOC gate globs, defaults.py split, duplicated prologues, unused dep


ishiland:
From the 1.6.0 pre-tag audit. Grouped backend maintainability items:

- The LOC-budget gate globs filenames, so the five largest modules are ungated; `ingest/router.py` sits 22 lines from the default cap cliff and would trip on its next feature. Gate by path, not glob.
- `platform/extensions/defaults.py` is 1815 LOC doing 4 jobs, with 63 platform→processing import edges including a router import — an axis the layering guard doesn't scan. Split by job; extend `test_layering.py` to cover the platform→processing axis.
- The 35-line job claim/heartbeat prologue is pasted at 7 task sites and already forced one cross-module private import. Extract a helper/decorator.
- The raster-family check appears 10× with 2 divergent copies; the email-domain policy 5×. Centralize each.
- `azure-identity` is a production dependency with zero imports — remove. The audit-extension seam has zero consumers; either document the intended consumer or delete. 5 package.json gates (knip + 4 lint:sec pins) exist that nothing runs — wire up or remove.
- Nit: `analysis/tasks.py:502` inline-quotes `by_field` instead of using `_sql_quote_ident` (guarded by `_SAFE_IDENT`, so cosmetic — but the one divergent quoting site is where the next bug hides).
- Consider mypy on the port seam: it carries 113 `type: ignore`s with no checker actually running.

PR #1308, opened 2026-08-09
chore(extensions): re-anchor the audit-seam removal promise to #1303

ishiland:
The deprecated AuditExtension seam (#836/#873) promised removal "at the next EXTENSION_API_VERSION bump" while the version was 2. The version is now 4, so the docstrings in the protocol, the accessor, and the compatibility test promise a removal point that already passed — twice — which reads as self-contradictory to anyone auditing the seam.

This re-anchors those three docstrings to the coordinated removal tracked in #1303 instead of a hardcoded version number. Docstring/comment-only: no behavior, no API surface, no test semantics change. The seam itself stays intact per the compatibility contract (removing it mid-version would ImportError overlays into a silent `load_extensions()` skip).

Refs #1303 (stays open — it tracks the actual removal at the bump).

Verification:
- `uv run ruff check` / `ruff format --check` on the touched files: clean
- `uv run pytest tests/test_extensions.py -q`: 27 passed

### DIFF
--- a/backend/app/platform/extensions/__init__.py
+++ b/backend/app/platform/extensions/__init__.py
@@ -331,10 +331,10 @@ def get_audit_extension() -> AuditExtension:
     """DEPRECATED — scheduled for removal at the next EXTENSION_API_VERSION bump.
 
     fix(#873 review r1+r3): the whole seam is deprecated (no core caller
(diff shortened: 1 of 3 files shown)


ishiland:
@codex review

PR #1309, opened 2026-08-09
fix(datasets): stop returning raw exception text from bulk delete

### LINKED ISSUE #1297: fix(datasets): bulk delete returns raw unexpected-exception text to the client
## Problem

The bulk-delete endpoint returns the raw text of unexpected exceptions to the client. In `backend/app/modules/catalog/datasets/api/router.py`, the per-item loop's broad `except Exception as exc` (currently around line 460) records every failure as:

```python
BulkDeleteResultItem(
    dataset_id=item.dataset_id,
    status="error",
    detail=str(exc),
)
```

That is correct for the two expected cases (`DependentVrtError` and the title-confirmation `ValueError`), whose messages are written to be public. It is wrong for everything else: an asyncpg/SQLAlchemy error, a storage-provider SDK error, or an audit-sink failure ships its full message to any caller allowed to use bulk delete. Because the envelope is a 200 `BulkDeleteResponse`, the central exception sanitization never sees it.

The 2026-08-07 internal backend audit reproduced this by injecting `RuntimeError("relation catalog.secret_table failed; password=fake-sentinel")` into the delete path; the response item contained the complete string, relation name and credential marker included. Re-verified on `main` at 035b72888.

## Fix

Branch on type inside the existing `except` block:

- `DependentVrtError` and `ValueError`: keep `detail=str(exc)` (current public contract).
- Every other exception: keep the existing `logger.exception(...)` server-side log, and set a fixed public string, `detail="Dataset deletion failed unexpectedly"`.

Branch on `isinstance` only; never inspect message contents. A typed per-item error-code envelope is a separate API-contract discussion and out of scope here.

## Acceptance criteria

- [ ] An unexpected exception during a bulk-delete item produces `status="error"` with the fixed detail string; no text from the exception appears anywhere in the serialized response.
- [ ] `DependentVrtError` and title-mismatch `ValueError` messages are unchanged. Existing `TestBulkDeleteDatasets` in `backend/tests/test_datasets.py` passes unmodified.
- [ ] New regression test in `TestBulkDeleteDatasets`: patch a dependency of the delete path to raise a `RuntimeError` whose message contains a sentinel relation name, an object path, and a fake credential marker; assert none of the three appears in the response body and the item detail equals the stable string.
- [ ] The full exception still reaches the server log.

## Verification

```bash
uv run pytest tests/test_datasets.py -k BulkDelete -v
```

## Traps

- `.env.test` is gitignored; create it once with `cp .env.test.example .env.test` from the repo root.
- `caplog` captures zero structlog records in this codebase. If you assert on the server-side log, use the same log-capture pattern as neighboring tests that check `logger.exception` output; otherwise drop the log assertion rather than fight it.

ishiland:
## Summary

The bulk-delete endpoint's per-item exception handler in `router.py` already branched on exception type to decide whether to log server-side, but always serialized `str(exc)` into the response `detail` regardless of that branch — so an unexpected exception (asyncpg/SQLAlchemy error, storage SDK error, audit-sink failure) shipped its raw message, including relation names, paths, and potentially credential fragments, to any caller with bulk-delete access. Since the response is a 200 envelope, the central exception sanitizer never sees it.

Fix: branch the `detail` value the same way the logging branch already does. `DependentVrtError` and title-mismatch `ValueError` keep their public messages unchanged; every other exception now returns a fixed `"Dataset deletion failed unexpectedly"` string while the full exception still reaches the server log via the existing `logger.exception(...)` call.

Checked sibling `except Exception` blocks in the same file (quicklook storage fallback) and in `router_export.py` / `router_reupload.py` / `router_vrt.py` — those already use fixed detail strings with server-side logging, so this was the only instance of the leak.

## Test plan

- [x] New regression test `test_bulk_delete_unexpected_exception_hides_detail` in `TestBulkDeleteDatasets`: monkeypatches `delete_dataset` to raise a `RuntimeError` containing a sentinel relation name, object path, and fake credential marker; asserts none of the three appear anywhere in the response body, the item detail equals the fixed string, and the full exception still reaches the server log via `structlog.testing.capture_logs()`.
- [x] Existing `TestBulkDeleteDatasets` tests pass unmodified (`DependentVrtError`/`ValueError` messages unchanged).

```
cd backend && uv run ruff check app tests && uv run ruff format --check app tests
cd backend && set -a && source ../.env.test && set +a && uv run pytest tests/test_datasets.py -k BulkDelete -v
cd backend && set -a && source ../.env.test && set +a && uv run pytest tests/test_datasets.py -q
```

Closes #1297

### DIFF
--- a/backend/app/modules/catalog/datasets/api/router.py
+++ b/backend/app/modules/catalog/datasets/api/router.py
@@ -52,6 +52,7 @@ from app.platform.cache.provider import get_tile_cache
 from app.platform.cache.tiles import invalidate_catalog_cache
 from app.modules.catalog.collections.service import get_dataset_collections
 from app.modules.catalog.datasets.domain.service import (
+    DatasetTitleMismatchError,
     DependentVrtError,
     create_empty_dataset,
     delete_dataset,
@@ -459,7 +460,7 @@ async def bulk_delete_datasets_endpoint(
             deleted += 1
         except Exception as exc:  # broad: per-item bulk-delete is isolated — any failure is recorded per-item without aborting the batch
             await db.rollback()
-            if isinstance(exc, (DependentVrtError, ValueError)):
+            if isinstance(exc, (DependentVrtError, DatasetTitleMismatchError)):
                 detail = str(exc)
             else:
                 logger.exception(
--- a/backend/app/modules/catalog/datasets/api/router.py
+++ b/backend/app/modules/catalog/datasets/api/router.py
@@ -459,16 +459,19 @@ async def bulk_delete_datasets_endpoint(
             deleted += 1
         except Exception as exc:  # broad: per-item bulk-delete is isolated — any failure is recorded per-item without aborting the batch
             await db.rollback()
-            if not isinstance(exc, (DependentVrtError, ValueError)):
+            if isinstance(exc, (DependentVrtError, ValueError)):
+                detail = str(exc)
+            else:
                 logger.exception(
                     "Unexpected error during bulk delete",
                     dataset_id=str(item.dataset_id),
                 )
+                detail = "Dataset deletion failed unexpectedly"
             results.append(
                 BulkDeleteResultItem(
                     dataset_id=item.dataset_id,
                     status="error",
-                    detail=str(exc),
+                    detail=detail,
                 )
             )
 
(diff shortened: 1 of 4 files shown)


ishiland:
@codex review

ishiland:
@codex review

PR #1310, opened 2026-08-09
fix(frontend): render stroke-only polygon swatches in legend and layer panel

### LINKED ISSUE #1288: fix(frontend): stroke-only polygon layers render an invisible swatch in legend and layer panel
A polygon layer styled stroke-only (fill disabled, colored outline) renders correctly on the map, but its swatch is invisible in both the legend widget and the Layers panel row. The layer looks like it has no symbology at all in either place.

Live example (demo, v1.10.0): https://demo.getgeolens.com/maps/cc97db69-eb0d-46d7-94d2-fe47710d1c73 — one MULTIPOLYGON layer (MNMAP_PLUTO) whose stored style is:

```json
"style_config": { "builder": { "fill_disabled": true, "outline_color": "#ec4b7f", "outline_width": 1, "stroke_disabled": false } },
"paint": { "fill-color": "#3b82f6", "fill-opacity": 0, "fill-outline-color": "#0058ac" }
```

The map draws the pink outlines fine. The legend row renders as `<span style="opacity: 0"><svg>...</svg></span>` — the swatch exists in the DOM but is fully transparent. Same `opacity: 0` span in the Layers panel row.

## Root cause

`frontend/src/components/map/layer-icons.tsx:71` extracts `paint['fill-opacity']` into the swatch style's `fillOpacity`, and `frontend/src/components/map/LegendEntries.tsx:27-30` multiplies it into a compound opacity applied as a CSS `opacity` on the entire swatch SVG:

```ts
const compoundOpacity = (s?.opacity ?? 1) * (s?.fillOpacity ?? 1);
return compoundOpacity < 1 ? { opacity: compoundOpacity } : undefined;
```

For a stroke-only polygon, `fill-opacity` is 0 by design, so the whole element — including the outline the layer is actually drawn with — disappears. The fill's opacity belongs on the SVG fill, not on the container: an element-level opacity can only be correct when fill and stroke share it.

## Fix shape

When `fill-opacity` is 0 (or `builder.fill_disabled` is set) and the layer has a visible outline, render an outline-only swatch: transparent SVG fill, stroke in the outline color at the configured width. More generally, apply fill opacity to the SVG `fill-opacity` attribute and stroke opacity to `stroke-opacity`, keeping element-level `opacity` for the layer-level opacity only.

Check every consumer of `swatchOpacityStyle` / the `fillOpacity` hint: `LegendEntries.tsx` (legend widget and Layers panel both go through it) and `components/viewer/LayerLegend.tsx`.

## Secondary observation, verify while in there

The layer's flat `paint` mirror is stale relative to the builder state: `fill-outline-color` is `#0058ac` while `builder.outline_color` is `#ec4b7f`, and the map renders the builder's pink. The renderer clearly derives from `style_config.builder`, so the swatch should too — deriving it from the stale `paint` mirror would show the wrong color even once the opacity bug is fixed. Whether the stale mirror itself is a save-path bug is worth a look but may be a separate issue.

## Reproduce locally

Import any polygon dataset, style it in the builder with fill disabled and a visible stroke, save. Both the legend widget and the Layers panel row show a blank swatch while the map draws the outlines.

ishiland:
## Summary

A polygon layer styled stroke-only (fill disabled, colored outline) draws correctly on the map, but its swatch was invisible in the legend widget and the Layers panel row.

Root cause: `paint['fill-opacity']` was extracted as a `fillOpacity` hint and multiplied into a single CSS `opacity` applied to the whole swatch element. For a stroke-only polygon, `fill-opacity` is 0 by design, so the entire swatch — including the outline it's actually drawn with — disappeared.

## Fix

- `fillOpacity` now lands on the specific SVG attribute (`fill-opacity` for shape fills, `stroke-opacity` for line-only elements), or is alpha-blended into the CSS background for the div-based polygon chip. Element-level `opacity` is reserved for layer-level opacity only.
- Applied consistently across every consumer of the shared swatch helpers: `layer-icons.tsx` (`ColorizedGeometryIcon` / `LayerTypeIcon`, used by the Layers panel, `LegendPlugin`, and the viewer `LayerLegend`) and `LegendEntries.tsx` (`GeometrySwatch`, `GraduatedRadiusLegend`, `GraduatedWidthLegend`, used by categorical/graduated legend rows).
- While in there: swatch outline color now prefers `style_config.builder.outlineColor` over the flat `paint['_outline-color']` mirror, matching how the map itself renders outlines (the mirror can go stale relative to builder state — same pattern the fill adapter already uses).

## Verification

- `cd frontend && npm run lint`
- `cd frontend && npm run typecheck`
- `cd frontend && npx vitest run src/components/map/__tests__/layer-icons.test.tsx src/components/map/__tests__/LegendEntries.test.tsx src/components/viewer/__tests__/LayerLegend.test.tsx src/components/builder/__tests__/LayerStyleEditor.test.tsx src/components/map-plugins/builtin/__tests__/LegendPlugin.test.ts src/components/map-plugins/builtin/__tests__/LegendPlugin.terrain.test.tsx src/components/builder/__tests__/SidebarRail.test.tsx src/components/builder/__tests__/StackRow.test.tsx src/components/builder/__tests__/LayerEditorPanel.test.tsx`
- `cd frontend && npm run test:coverage` (full suite, 383 files / 4700 tests passing)

Closes #1288

### DIFF
--- a/frontend/src/components/builder/LayerStyleEditor/utils.ts
+++ b/frontend/src/components/builder/LayerStyleEditor/utils.ts
@@ -124,7 +124,10 @@ export function stylePreviewStyle(layer: MapLayerResponse) {
   const gt = (layer.dataset_geometry_type ?? '').toUpperCase();
   if (gt.includes('POLYGON')) {
     return {
-      outlineColor: typeof paint['_outline-color'] === 'string' ? paint['_outline-color'] as string : undefined,
+      // fix(#1288): builder.outlineColor wins over the flat paint mirror, which
+      // can go stale — the map itself renders from style_config.builder.
+      outlineColor: layer.style_config?.builder?.outlineColor
+        ?? (typeof paint['_outline-color'] === 'string' ? paint['_outline-color'] as string : undefined),
       strokeDisabled: Boolean(layer.style_config?.builder?.strokeDisabled ?? paint['_stroke-disabled']),
       opacity: layer.opacity,
       fillOpacity: typeof paint['fill-opacity'] === 'number' ? paint['fill-opacity'] as number : undefined,
--- a/frontend/src/components/map-plugins/builtin/LegendPlugin.tsx
+++ b/frontend/src/components/map-plugins/builtin/LegendPlugin.tsx
@@ -30,12 +30,16 @@ function getSwatchStyleFromPaint(
   geometryType: string | null | undefined,
   masterOpacity: number,
   // fix(#914): the builder block carries the fill-pattern tint stash.
-  builder?: { fillColorSaved?: string },
+  // fix(#1288): also carries outlineColor — the map renders from builder state,
+  // so the legend must prefer it over the paint mirror below, which can go stale.
+  builder?: { fillColorSaved?: string; outlineColor?: string },
 ): SwatchStyle {
   const gt = (inferGeometryType(paint, geometryType) ?? '').toUpperCase();
   const strokeDisabled = !!paint?.['_stroke-disabled'];
 
-  const rawOutline = gt.includes('POINT') ? paint?.['circle-stroke-color'] : paint?.['_outline-color'];
+  const rawOutline = gt.includes('POINT')
+    ? paint?.['circle-stroke-color']
+    : (builder?.outlineColor ?? paint?.['_outline-color']);
   const outlineColor = typeof rawOutline === 'string' ? rawOutline : undefined;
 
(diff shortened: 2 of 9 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1311, opened 2026-08-09
perf(catalog): batch dataset loading for VRT creation and collection linking

### LINKED ISSUE #1298: perf(catalog): batch dataset loading and authorization for VRT creation and collection linking
## Problem

Two write paths authorize linked datasets one ID at a time, one round trip per ID:

- VRT creation: `create_vrt_job()` in `backend/app/processing/ingest/service.py` loops over the requested source IDs and calls `get_dataset(db, sid)` then `check_dataset_access(...)` per source (currently around lines 755-767). `VrtCreateRequest` allows up to 500 sources, so one valid request can issue roughly 500 sequential dataset SELECTs, plus up to one grant query each for restricted datasets, before the job is even queued. The request and its DB session stay open across all of it.
- Collection membership: the linking route in `backend/app/modules/catalog/collections/router.py` (around lines 327-339) does the same per-ID `get_dataset` plus `check_dataset_access` at a cap of 100.

The read-side VRT endpoints already batch through `_load_source_datasets()` in `backend/app/modules/catalog/datasets/api/router_vrt.py` and have a query-count regression test. The write side has neither. The 2026-08-07 internal backend audit measured exactly 500 dataset loads plus 500 access checks in a 500-source mocked reproduction; re-verified on `main` at 035b72888.

## Fix

Add one batch helper and use it at both call sites.

1. In `backend/app/modules/catalog/authorization.py`, next to the scalar checks, add:

   ```python
   async def check_datasets_access_bulk(db, dataset_ids, user, user_roles) -> dict[UUID, Dataset]
   ```

   It loads all requested datasets in a single SELECT (eager-load the joined `Record`, the way `_load_source_datasets()` does), computes the accessible set with the same extension-backed visibility policy that `apply_visibility_filter` uses, and raises the same fail-closed 404 as `check_dataset_access` for the first requested ID (in request order) that is missing or denied. It must go through the registered permission-extension seam, not around it: overlays that wrap the extension have to see the batch path.
2. Replace the per-ID loop in `create_vrt_job()` with one call; keep the compatibility validation that follows it unchanged.
3. Replace the per-ID loop in the collections linking route with the same call.

Do not replace the loop with `asyncio.gather` over the scalar checks; that is still N queries.

## Acceptance criteria

- [ ] Parity test: for a mixed set (public dataset, requester-owned private, other-user private, nonexistent ID), the batch helper allows exactly the IDs for which scalar `check_dataset_access` succeeds, and raises 404 in the same cases.
- [ ] Query-count regression test for VRT creation at 5 sources and at the 500-source cap: dataset loading plus authorization is a small constant number of queries, independent of source count. Mirror the mechanics of the existing read-side query-count test.
- [ ] Same for collection linking at the 100-ID cap.
- [ ] Fail-closed behavior unchanged: a denied or missing ID still produces the same 404 status and message shape as today; existing authz tests pass unmodified.
- [ ] `tests/test_layering.py` passes. If `authorization.py` or a touched service module crosses its LOC cap, raise the cap in `_MODULE_LOC_CAPS` in the same commit with a one-line comment saying what the lines bought.

## Verification

```bash
uv run pytest tests/test_rule1_structural.py tests/test_layering.py -q
uv run pytest tests/ -k "vrt" -q
```

## Traps

- Import dataset service logic only through the `domain/service.py` facade; `test_layering.py` fails direct sub-module imports.
- `tests/test_rule1_structural.py` walks the FastAPI route table and recognizes specific access-check calls. If it stops recognizing these handlers after the swap, add the new helper to its accepted-check list in the same commit.
- Keep the 404 message stable; tests and clients may match on it.

ishiland:
## Summary

VRT creation (`create_vrt_job()`) and the collection-linking route each authorized their linked datasets one id at a time — `get_dataset()` then `check_dataset_access()` per id — so a single request could cost hundreds of sequential round trips before the write even happened (up to 500 for VRT creation, up to 100 for collection linking).

Adds `check_datasets_access_bulk()` in `app.modules.catalog.authorization`, next to the scalar checks. It loads every requested dataset (with `Record` eager-loaded) in one SELECT, resolves the accessible subset through the same `apply_visibility_filter`/`filter_visible` seam the scalar check and other batch helpers use (so a permission-extension overlay still sees every request), and raises the same fail-closed 404 as `check_dataset_access` for the first missing or denied id in request order. Both call sites swap the per-id loop for a single call to this helper; no API-shape or authorization-semantics changes.

## Changes

- `app/modules/catalog/authorization.py`: add `check_datasets_access_bulk()`.
- `app/processing/ingest/service.py`: `create_vrt_job()` now authorizes its sources with one batch call.
- `app/modules/catalog/collections/router.py`: `add_datasets_endpoint()` now authorizes link targets with one batch call.
- `tests/test_batch_dataset_authz_1298.py` (new): parity test against the scalar check, query-count regression at both call sites (5 vs. 500 sources; 5 vs. 100 collection ids), and positional fail-closed coverage (a denied dataset in the middle of an otherwise-valid batch).
- `tests/test_rule1_structural.py`: register the new helper as a sanctioned guard so the structural access-control audit recognizes it.
- `tests/test_vrt_creation_173.py`: update the pure-mock authz stub fixture to match the new call shape (no behavior change — same DB-execute-count savings the original stub existed for).
- `tests/test_layering.py`: drop a now-stale burn-down entry (`ingest/service.py` no longer imports the datasets domain service directly).

## Verification

```bash
uv run pytest tests/test_rule1_structural.py tests/test_layering.py -q
uv run pytest tests/ -k "vrt" -q
uv run pytest tests/test_collections.py tests/test_batch_dataset_authz_1298.py -q
```

All green. The 500-source query-count regression test and full VRT/collections suites pass.

Closes #1298

### DIFF
--- a/backend/app/modules/catalog/authorization.py
+++ b/backend/app/modules/catalog/authorization.py
@@ -18,6 +18,7 @@ from typing import Any
 from fastapi import HTTPException, status
 from sqlalchemy import Select, select
 from sqlalchemy.ext.asyncio import AsyncSession
+from sqlalchemy.orm import joinedload
 
 from app.core.identity import Identity
 from app.modules.auth.models import Role, UserRole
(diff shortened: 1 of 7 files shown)


ishiland:
@codex review

PR #1312, opened 2026-08-09
i18n(frontend): externalize remaining hardcoded UI strings

### LINKED ISSUE #1301: i18n(frontend): externalize the remaining English UI strings (settings labels, cluster reasons, generated labels, composed sentences)
## Problem

The 2026-08-07 internal i18n audit found about 44 remaining user-visible English strings against 3,883 externalized keys per locale. They cluster into four boundaries, re-verified on `main` at 035b72888. Spanish, French, and German users currently get English text in the admin settings tabs, cluster fallback warnings, several generated builder labels, and a set of composed accessible names.

File references below are from the audit; line numbers will have drifted, so anchor on the symbols and grep.

## Workstream A: admin settings labels (12 sites)

`SettingsAITab.tsx` (9 sites) and `SettingsNetworkTab.tsx` (3 sites) render:

```tsx
{findSetting(settings, 'ai_enabled')?.label ?? t('ai.labels.aiEnabled')}
```

Backend `SettingItem.label` is a required field populated from English literals in `backend/app/core/persistent_config.py`, so the translated fallback is dead code and admins always see English.

Decision: these are static, known settings. Drop the `findSetting(...)?.label` half and call `t(...)` directly at all 12 sites. Do not change the API; `label` stays for CLI and script consumers. Find every site with:

```bash
grep -n "findSetting" frontend/src/components/admin/settings/Settings*.tsx
```

## Workstream B: cluster fallback reasons (5 strings)

`clusterFallbackMessage()` in `frontend/src/components/builder/cluster-source.ts` returns five English sentences, which `BuilderMap.tsx` interpolates into the translated `builderMap.clusterFallback` toast, so users see a half-translated warning.

Decision: keep the helper pure. Return the `ClusterSourceStatus` code (the switch already has one case per reason) instead of a sentence, and translate at the call site with keys like `builderMap.clusterReason.missingCount`. Add the five keys to `builder.json` in all four locales. Delete `clusterFallbackMessage` if nothing else uses it.

## Workstream C: synthetic and direct labels

- `use-builder-editor-scene.ts`: the `Settings`, `Sublayer`, `Basemap · ...`, and `Untitled` synthetic editor labels.
- `builder-layer-mutations.ts`: the `Layer` fallback and the `${baseName} rendering` duplicate name.
- `MapBuilderPage.tsx`: title-cased technical basemap IDs and the `Basemap` fallback.
- `SamlProvidersSection.tsx`: the `(leave blank to keep existing certificate)` textarea placeholder.
- `SpatialFilterPanel.tsx`: `1 polygon selected`. Use a plural key, and remember `_one` must interpolate `{{count}}` (French resolves count 0 to `_one`), never hardcode "1".
- `admin:saml.groupClaim` is still the English "Group Claim" in es/fr/de; translate it.

Exemption: the three `AppErrorBoundary` fallbacks stay English on purpose; they render when translation init itself failed.

## Workstream D: composed sentences and accessible names (about 16 sites)

Translated fragments joined with names, punctuation, or raw values, so translators cannot control word order: `DatasetSearchPanel` and `ChatPanel` a11y action labels, `ClusterEditor` field labels, `AttributeTable` filter label, `MapTitleBar` save/status join, `FilterPanel` and `FilterSheet` temporal-extent labels, `UserList` and `AuditLogViewer` search placeholders, `ChangeHistory` changed-fields sentence, `ephemeral-preview.ts` live-region label, `SearchResultCard` middle-dot readout, `SpatialExtentCard` SRID qualifier.

Replace each with a full interpolation key, for example `t('builder.addToMapNamed', { name })`, where the locale value holds the complete sentence.

## Acceptance criteria

- [ ] All four workstreams done; a re-grep of the per-file inventory above comes back clean.
- [ ] Every new key exists in all four locales (en/es/fr/de) with real translations; a `defaultValue` alone fails the parity gate.
- [ ] Plural keys follow the all-four-or-none rule; no `_one` value hardcodes "1".
- [ ] Pure helpers stay pure: no `t()` inside `cluster-source.ts` or the pure sections of the builder hooks; they return keys or codes.

## Verification

```bash
cd frontend
npm run test:i18n
npm run check:i18n:changed
npm run check:i18n:toast-strings
npm run check:i18n:api-errors
npm run lint && npm run typecheck
```

Plus targeted `npx vitest run` for the components you touch. Estimated at 2 to 3 working days including locale updates and tests; the workstreams are independent and can land as separate PRs in the order A, B, C, D.

ishiland:
## Summary

Closes the four string clusters surfaced by the 2026-08-07 internal i18n audit (44 remaining English strings vs. 3,883 externalized keys per locale):

- **Workstream A** — `SettingsAITab.tsx` / `SettingsNetworkTab.tsx` dropped the dead `findSetting(...)?.label` half at all 12 sites so the `t(...)` fallback actually renders (backend `SettingItem.label` is English-only and `label` stays for CLI/script consumers).
- **Workstream B** — `clusterFallbackMessage` in `cluster-source.ts` now returns the `ClusterSourceStatus` code instead of an English sentence; `BuilderMap.tsx` translates at the call site via new `builderMap.clusterReason.*` keys. The helper itself stays pure (no `t()`).
- **Workstream C** — `use-builder-editor-scene.ts` and `builder-layer-mutations.ts` accept translated label objects (defaulting to the current English copy, so existing pure-function tests pass unmodified) instead of hardcoding `'Settings'`/`'Sublayer'`/`'Untitled'`/`'Layer'`/`'rendering'`. `MapBuilderPage.tsx`'s basemap-ID fallback and `SamlProvidersSection.tsx`'s certificate placeholder now reuse existing translated keys. `SpatialFilterPanel.tsx`'s `1 polygon selected` moved to a proper `_one`/`_other` plural key. `admin:saml.groupClaim` is now translated in es/fr/de (it matched the OAuth `groupClaim` sibling key, which was already translated).
- **Workstream D** — replaced string concatenation with full interpolation keys across `DatasetSearchPanel`, `ChatPanel`, `ClusterEditor`, `AttributeTable`, `MapTitleBar`, `FilterPanel`/`FilterSheet`, `UserList`/`AuditLogViewer`, `ChangeHistory`, `ephemeral-preview.ts`, and `SearchResultCard`/`SpatialExtentCard`, so translators control word order instead of JS-level string glue.

Every new/changed key ships with real translations in all four locales (en/es/fr/de) — no `defaultValue`-only keys.

## Test plan

- [x] `cd frontend && npm ci`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm run test:i18n`
- [x] `npm run check:i18n:changed`
- [x] `npm run check:i18n:toast-strings`
- [x] `npm run check:i18n:api-errors`
- [x] `npm run test:coverage` (full suite: 383 files / 4693 tests passed)
- [x] Targeted `npx vitest run` for every touched component/hook

Closes #1301

### DIFF
--- a/frontend/src/components/admin/AuditLogViewer.tsx
+++ b/frontend/src/components/admin/AuditLogViewer.tsx
@@ -302,7 +302,7 @@ export function AuditLogViewer() {
           <DataTableSearch
             value={searchQuery}
             onChange={(v) => { setSearchQuery(v); setPage(0); }}
-            placeholder={t('audit.table.user') + ' / ' + t('audit.table.action')}
+            placeholder={t('audit.searchPlaceholder')}
           />
         </CardAction>
       </CardHeader>
--- a/frontend/src/components/admin/UserList.tsx
+++ b/frontend/src/components/admin/UserList.tsx
@@ -234,7 +234,7 @@ export function UserList() {
             <DataTableSearch
               value={searchQuery}
               onChange={(v) => { setSearchQuery(v); setPage(0); }}
-              placeholder={t('users.table.username') + ' / ' + t('users.table.email')}
+              placeholder={t('users.searchPlaceholder')}
             />
             <FilterSelect
               label=""
--- a/frontend/src/components/admin/saml/SamlProvidersSection.tsx
+++ b/frontend/src/components/admin/saml/SamlProvidersSection.tsx
@@ -463,7 +463,7 @@ export function SamlProvidersSection() {
                 }
                 placeholder={
                   editingProvider
-                    ? '(leave blank to keep existing certificate)'
+                    ? `(${t('saml.idpCertificateKeep')})`
                     : '-----BEGIN CERTIFICATE-----\nMIIC...\n-----END CERTIFICATE-----'
                 }
               />
--- a/frontend/src/components/admin/settings/SettingsAITab.tsx
+++ b/frontend/src/components/admin/settings/SettingsAITab.tsx
@@ -178,7 +178,7 @@ export function SettingsAITab({ settings, envOnly, onSave, onReset, isSaving, sa
         <div className="flex items-center justify-between max-w-md">
           <div className="space-y-0.5">
             <div className="flex items-center gap-2">
-              <Label htmlFor="ai-toggle">{findSetting(settings, 'ai_enabled')?.label ?? t('ai.labels.aiEnabled')}</Label>
+              <Label htmlFor="ai-toggle">{t('ai.labels.aiEnabled')}</Label>
               <SettingSourceBadge source={findSetting(settings, 'ai_enabled')?.source ?? 'default'} settingKey="ai_enabled" onReset={onReset} />
             </div>
             <p className="text-sm text-muted-foreground">{t('settings.general.aiFeaturesDescription')}</p>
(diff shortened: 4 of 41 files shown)


ishiland:
@codex review

Issue #1223, opened 2026-08-05
feat(catalog): persist schema drift from refresh previews


ishiland:
Blocked by #1217; pairs with the refresh-history issue in this milestone. Line refs as of 2026-08-05.

`compute_schema_diff` (`backend/app/modules/catalog/datasets/domain/service_metadata.py:54-93`) already produces `columns_added`, `columns_removed`, `type_changes`, and row-count deltas, but only at re-upload preview time, and the result is discarded once the user commits. Drift is computed today and then forgotten.

## Scope

- Persist the diff produced by the refresh/re-upload preview onto the committed refresh-history row.
- Derive a dataset-level drift status (none / drifted), updated on each committed refresh.
- Expose via the history API. `frontend/src/components/dataset/SchemaDiffView.tsx` already renders this exact shape, so the frontend cost is wiring, not new UI.

## Where to look

- The pure function: `service_metadata.py:54-93` (its return shape is `SchemaDiff` in `domain/schemas.py:618`).
- Its two call sites, both preview-time: `backend/app/modules/catalog/datasets/api/router_reupload.py:344` (service preview) and :478 (file preview).
- The commit paths that must carry the diff forward: the commit handler in `router_reupload.py` and the worker tasks in `backend/app/processing/ingest/tasks_reupload.py`.
- The storage slot: the refresh-history row from the companion issue (JSONB column or ref, per #1217).

## Suggested path

1. At preview time, stash the computed diff in the job's `user_metadata` (it already carries reupload state; see `_get_bound_reupload_job_or_404` in `router_reupload.py:82` for how reupload jobs are bound).
2. At commit time, copy the diff out of job metadata onto the durable history row and set the dataset drift status. Copying at commit is what makes this safe against job purge; the jobs table is never the record.
3. If the preview is stale relative to the actual committed pull (service data changed between preview and commit), prefer recomputing at commit while the staging table still exists; the function is cheap and pure.
4. Tests: diff persisted on commit, nothing persisted on abandoned preview, drift status flips on a column rename and clears on a matching re-refresh.

## Acceptance

- [ ] A committed refresh stores its schema diff; a preview that is never committed stores nothing
- [ ] Dataset drift status visible via API and updated per commit
- [ ] Existing preview UX unchanged
- [ ] OpenAPI regen trio committed if response shapes changed

## Verify

```bash
docker compose up -d --wait db
[...]

Issue #1222, opened 2026-08-05
feat(catalog): health checks for standalone STAC and service origins


ishiland:
Blocked by #1217 and the source-state fields issue in this milestone. Line refs as of 2026-08-05.

VRT members already get health probes: healthy / missing / inaccessible, computed on demand with a bounded ranged read through `make_safe_client` (`backend/app/modules/catalog/datasets/api/router_vrt.py:243-324`, helper `_remote_asset_exists` at :76-93). Standalone STAC datasets, the only genuinely live-referenced kind (`backend/app/modules/catalog/sources/stac_router.py:590` sets `storage_backend="remote"`), have no health check at all: a deleted upstream COG is discovered only when tiles fail. Service origins have no reachability check either.

## Scope

- Health checks covering standalone STAC assets and service origins (endpoint shape per the ADR); persist `last_checked_at` and `source_health` from the source-state fields issue.
- On-demand only. No scheduler, per the locked boundary in #1217.
- Results feed the Source panel and the search/header chips (companion issues).

## Where to look

- Reference implementation to copy: the whole VRT status flow in `router_vrt.py:243-324`. `_remote_asset_exists` (:76-93) is the safe remote probe: `Range: bytes=0-0` request through `make_safe_client` with a timeout. The local branch uses `storage.exists(...)`.
- SSRF factory: `make_safe_client()` in `backend/app/modules/catalog/sources/security.py`. Never construct `httpx.AsyncClient(follow_redirects=True, ...)` directly; a pre-commit grep hook and `tests/test_rule2_structural.py` both fail the PR if you do.
- What to probe per origin: STAC datasets have their remote COG at `RasterAsset.asset_uri` (`backend/app/processing/raster/models.py:54`, rows with `storage_backend="remote"`); service datasets get a reachability check of the stored origin pointer (a HEAD/ranged GET of the service endpoint, not a full pull).
- Access control: `check_dataset_access_or_anonymous` from `backend/app/modules/catalog/authorization.py` (read-side endpoint, Rule 1).

## Suggested path

1. Extract or mirror `_remote_asset_exists` into a shared helper so VRT and dataset-level checks use one probe.
2. Add the health endpoint (or fold into an existing dataset read per the ADR). Compute status, write `last_checked_at` + `source_health`, return both.
3. Frontend consumption happens in the panel/chips issues; this issue is done at the API level.
[...]

PR #1313, opened 2026-08-09
feat(refresh): registered-PostGIS refresh strategy

### LINKED ISSUE #1265: feat(refresh): registered-PostGIS refresh strategy (recount, extent, schema, statistics)
ADR-002 Amendment A10 (ownership gap found by the 2026-08-07 milestone audit). Decision 5a defines postgis refresh as "recount features and refresh extent/statistics; no data movement" and names #1218/#1224 — but those only compute over stored columns. Nothing implements the recount.

## Scope

A `postgis` execution strategy behind #1220's shared refresh orchestration (invariant 11: strategy behind the executor, never a fork of it):

- recount `feature_count`, recompute spatial extent, refresh column schema snapshot and statistics for the registered table named by `origin_ref.table_name`;
- no data movement, no staging table (the source IS the live table) — the strategy declares that and the executor skips swap;
- run record + `last_refreshed_at`/`last_checked_at` semantics identical to every other strategy (success and failure rows);
- `source_health`: `missing` when the physical relation no longer exists, `inaccessible` on permission/connection errors;
- schema drift vs the stored snapshot persisted per #1223's mandatory commit-time rule (Amendment A5), noting there is no staging copy — the diff is live-vs-recorded.

## Acceptance

- [ ] refresh endpoint accepts a postgis-origin dataset (no more 409 for this origin)
- [ ] dropped/renamed physical table → failed run, `source_health='missing'`, old metadata retained
- [ ] concurrent dispatch → second gets `dataset_busy` via the one-active-run index
- [ ] Rule 1 write check; layering + whole-tree gates

Blocked by #1220 (executor extraction).

ishiland:
Closes #1265.

A dataset registered from an existing PostGIS table was the one origin kind GeoLens never re-read. Registration copies no data — the catalog points at the live relation and serves straight from it — so everything stored about that table is a measurement taken once, on the day it was registered, of a table its owner keeps writing to. Rows arrive, columns are added, the extent moves, and the refresh door answered 409 `refresh_not_applicable` forever after. ADR-002 Decision 5a defines a postgis refresh as "recount features and refresh extent/statistics; no data movement"; nothing implemented it.

## The strategy contract

**Behind the executor, not beside it (invariant 11).** One endpoint, one Rule 1 write gate, one `create_pending_run`, one run ledger, one `defer_with_orphan_guard` / `make_refresh_run_failed_rollback` pair. What varies per origin kind is the binding the door unpacks and the task it defers. The postgis strategy resolves no URL, runs no SSRF check and takes no credential, because its origin is a relation in the database the request is already talking to.

`test_dataset_refresh_runs.py::test_only_the_commit_handler_creates_a_run` is the structural guard here: it names `_dispatch_postgis_refresh` explicitly rather than wildcarding the module, so a future strategy that grows its own admission path still fails it.

**No staging, declared structurally.** There is no staging table and no swap anywhere in `tasks_postgis_refresh.py` — the source IS the destination. "The executor skips the swap step" is therefore discharged by there being nothing to skip, rather than by a flag the executor has to be told to honour.

**One measurement, one snapshot.** The measure phase runs at `REPEATABLE READ, READ ONLY`. Under the default `READ COMMITTED` every statement takes its own snapshot, so the count, the extent, the samples and the validity score could each describe a different instant of a table somebody else is writing to, and the catalog would store a combination that never existed. The writes live in a separate transaction at the ordinary isolation level, and that split is what makes the snapshot safe to take: the heartbeat renews this job's row from its own session throughout, so finalizing the job inside the snapshot transaction would collide with it and abort the run with a serialization failure. `READ ONLY` makes a write added to the measure phase fail loudly instead of silently holding a snapshot it should not.

**Health mapping.** This strategy is the only observer a registered table ever gets: the source-health probe (#1222) refuses postgis origins outright, since probing one would mean issuing an HTTP request to a relation. So unlike every other strategy — which leaves `source_health` to the probe's classifier rather than adding a second, weaker one — this one owns the verdict for its origin kind, and the mapping is a SQLSTATE lookup:

| Outcome | SQLSTATE | `source_health` | detail | run `error_code` |
| --- | --- | --- | --- | --- |
| relation dropped or renamed | `42P01`, or the up-front `to_regclass` check | `missing` | `not_found` | `source_missing` |
| GRANT revoked | `42501` | `inaccessible` | `unauthorized` | `source_inaccessible` |
| connection lost | class `08` | `inaccessible` | `network_error` | `source_inaccessible` |
| anything else (statement timeout, deadlock) | — | untouched | — | `postgis_refresh_failed` |

The last row is the one that matters most. The verdict is sticky for this origin kind, so writing one from a failure that established nothing would sit on the dataset until somebody happened to refresh again — reporting a healthy table as missing because one `COUNT(*)` was slow is worse than reporting nothing. For the same reason a successful measurement writes `healthy`: without it, a table marked missing and then restored would carry that verdict forever.

Classification walks the whole exception chain rather than reading the outermost SQLSTATE. `extract_metadata`'s spatial fast path catches every exception and immediately retries its per-helper queries inside the transaction the first failure already aborted, so a table dropped or revoked between two statements of the measurement arrives as `25P02` with the real code in `__context__`. Reading only the outer code reported the exact mid-flight race the classifier exists to cover as inconclusive.

The health words are mirrored from `sources/origin_probe.py` rather than imported, because processing/ may not import `app.modules.catalog`. A test pins the mirror against the probe's own `DETAIL_CODES` and `SOURCE_HEALTH_VALUES`, so a divergence fails a test instead of persisting a value the API schema does not enumerate.

**Drift recording.** Live-vs-recorded: there is no staging copy to diff against, so the comparison is the table as it is now against what the catalog last wrote down. Measured before the writes overwrite the stored values (the same ordering rule the swap paths follow) and recorded, never refused, per #1223 / Amendment A5.

**Invariant 10.** Nothing outside the success block writes `last_refreshed_at`, so a failed refresh leaves the dataset describing the last measurement that actually happened — which for this strategy is the only record of what the table held, at the exact moment the table is gone. The dataset-side health stamp is guarded on the `(origin_uri, origin_ref, source_format)` triple the attempt read, so a refresh that failed against a table the dataset is no longer bound to cannot mark the new binding missing.

## Smaller decisions worth flagging

- **The binding names the table; it does not steer the read.** `origin_ref` is JSONB, and a name taken from it and dropped into a query is one bad row away from measuring somebody else's relation and writing the result onto this dataset. The worker requires the stored pointer to spell exactly `(active tenant data schema, datasets.table_name)` and fails the run otherwise, rather than picking a winner.
- **`spatial_extent` is cleared when the table has no extent**, which is where this parts from `_apply_reupload_swap` (it only ever writes a non-NULL extent). That path installs bytes it fetched and a stale polygon is at worst a missed update; this path exists solely to make the stored metadata agree with the live table, and an emptied table still claiming its old footprint in spatial search is the precise lie the operation was asked to correct.
- **The tile URL version is bumped on every successful refresh**, not only when the count moved. Editing geometry, rewriting attributes, or deleting and reinserting the same number of rows changes every tile while leaving the total identical. The `_v=` bump is the part that demonstrably works — it is a catalog column the frontend reads, so browser and CDN caches are busted. The server-side Valkey purge beside it is best-effort and, per round 8 of review, currently inert from the worker process because `init_tile_cache()` only runs in the API lifespan; that is pre-existing and shared with both re-upload strategies, and is filed as #1315 rather than fixed here.
- **The job is marked `refresh`, deliberately not `reupload`.** Two pieces of shared SQL key off that marker — the legacy-live admission probe and the abandoned-run sweep's other-live-task clause — and both reason about swaps this task never performs. `_retry_capability` gains a matching branch so a failed refresh job stops telling the user their import source is gone.
- **A token is refused (422 `credential_not_applicable`) rather than ignored.** 202 plus a silently discarded secret is the worst available answer. The CLI (#1305) has no mapping for the new code and falls through to echoing the server message, which reads correctly; no CLI change needed.

## Verification

```
cd backend && uv run ruff check . && uv run ruff format --check .          # pass
cd backend && uv run pytest tests/test_postgis_refresh_1265.py -v          # 21 passed
cd backend && uv run pytest tests/test_layering.py -q                      # 40 passed
cd backend && uv run pytest tests/test_rule1_structural.py \
    tests/test_rule2_structural.py -q                                      # 82 passed
cd backend && uv run pytest -q                                             # 7965 passed, 89 skipped
cd frontend && npm run types:check && npm run typecheck && npm run lint \
    && npm run test:i18n                                                   # pass
make openapi-check                                                         # clean after regen
make sdks-check                                                            # clean after regen
make cli-test                                                              # 8 passed, 2 skipped
```

The suite covers the acceptance list end to end: the endpoint accepting a postgis origin, a dropped table and a renamed table both producing a failed run with `source_health='missing'` and retained metadata, concurrent dispatch getting 409 `dataset_busy` through the one-active-run index, drift recorded in both directions, an emptied table clearing the extent, a finished run releasing the dataset for the next one, the measure phase's isolation and read-only state asserted from inside it, and a tile purge on a refresh whose count did not move. Execution tests drive the real Procrastinate callable against a real physical table, so the run ledger, the heartbeat and the failure handler all run for real.

**API surface.** `backend/openapi.json`, the two SDKs and `frontend/src/types/api.generated.ts` carry a description change only — no response shape moved (`origin_kind` was already a plain string). No migration; nothing here needs a schema change.

## Follow-up (not in this PR)

`extract_metadata`'s broad `except Exception` fallback re-runs its per-helper queries inside a transaction the first failure already aborted, so those retries always raise `25P02` and the fallback can never actually succeed. This PR works around it for its own classification by walking the exception chain, but the fallback itself remains dead code for every caller, and `metadata.py` sits at an exact size cap — worth its own change rather than a drive-by here.


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
Round 6, both findings addressed — one fixed, one refuted with the reasoning recorded at the guard itself.

**Keep the record type synchronized with spatiality — fixed** (5d5be10fb). Confirmed against `service_create.py:213`, which sets `record_type = "table" if geometry_type is None else "vector_dataset"`. This task is the only thing that can change the answer for a registered table afterwards, so the derivation is now re-applied on every measurement, guarded to the two record types that rule governs (raster and VRT carry their own modality). Two tests, both verified to fail without the change: an empty-registered spatial table that gains rows becomes a `vector_dataset`, and a dataset whose geom columns are dropped becomes a `table`.

**Detect direct table writes before committing the snapshot — refuted, deliberately.** The premise is right: `tile_cache_version` only moves for GeoLens-side mutations, so an owner writing SQL directly between the measure phase and the write phase is invisible to the guard. That is not a gap the guard could close, and closing it is not what this feature can promise.

A measurement of a relation GeoLens does not own is true as of a point in time. The owner’s next commit makes it stale — whether that commit lands during the write window, one millisecond after the transaction commits, or an hour later. Narrowing the window changes nothing about that, because there is no window in which the answer stays true. The only way to be atomic with an external writer is to hold a lock on their table for the duration of the measurement, and "no data movement, serves from the live table" (Decision 5a) is exactly the promise that forbids it: a `COUNT(*)` on a large table would then block the owner’s writes to their own relation.

Nor is there a token to use instead. `xmin`/`pg_stat_*` are per-row, approximate, or reset on restart; none is a monotonic content version for a relation, which is why the catalog has one and the data table does not.

The two cases are also different in kind, which is why one warranted a fix and this one does not:

- a GeoLens feature edit writes a **newer measurement** of the same table, so applying an older snapshot over it **rolls stored state backwards** and leaves it wrong until the next write. That is a regression the catalog cannot recover from on its own.
- a direct owner write leaves the catalog **stale**, which is the ordinary condition of every registered table between refreshes and the exact thing this endpoint exists to correct on demand. The next refresh fixes it.

The comment at the guard now states this outright, so the next reader does not mistake it for a general "the measurement is current" promise.

ishiland:
@codex review

ishiland:
Round 7: two fixed, one deferred with reasoning.

**Compute quality with the measured modality — fixed.** Correct, and the mismatch was persisted rather than transient. The scoring stand-in now carries the record type the measurement implies. While fixing it I collapsed the geometry-type precedence into one function resolved once in the measure phase and consumed by both the write and the score, so the stored type and the type the score assumed can no longer be two derivations that disagree — which also fixed a smaller case of the same bug: an emptied spatial table was scored off the sampled `None` and silently lost its geometry and CRS dimensions.

**Deactivate geometry metadata when geometry disappears — fixed.** Confirmed: `refresh_attribute_metadata` touches the synthetic `geom` row only when `geometry_type` is non-null and excludes it from the removed-column sweep by name. Reasonable for every caller that replaces a table’s contents; wrong for the one caller whose relation can lose its geometry column while keeping its identity. Retired explicitly through the port’s ORM accessor. (The test runs a first refresh while the table is still spatial — without an existing `geom` row the assertion passes against a dataset that never had one, which is how the first version of it was vacuous.)

**Refresh auto-generated distributions after modality changes — deferred, deliberately, and filed rather than fixed.**

The observation is right, but the framing "fresh evidence in this revision" does not hold: in the demote case (geom columns dropped) the persisted distributions were *already* advertising vector tiles, GeoPackage and GeoJSON before this PR, because nothing ever re-derived them and `record_type` never changed either. In the promote case they were already missing the spatial rows. This PR does not create that staleness — it narrows it, because `build_assets` computes from `record_type` live and now gets the right answer. The persisted `record_distributions` rows are exactly as stale as they were on `main`.

Fixing them properly is a distinct piece of work, not a line in this task:

- `generate_distributions()` has merge semantics — it adds missing rows and never removes any. Calling it again handles the promote direction, but the demote direction needs a *reconcile* that deletes auto-generated rows the new modality excludes, while preserving `auto_generated=False` rows a user authored and any user edits to auto-generated ones. That policy does not exist yet and deciding it inside a metadata-refresh task is the wrong place for it.
- It also needs a new `ProcessingPort` seam: `generate_distributions` lives in `modules/catalog/records/service.py`, and `processing/` reaches catalog only through the port.

Both belong in a change whose subject is distribution reconciliation, with its own tests for the preservation rules. I will file it as a follow-up on the milestone rather than grow this PR a fourth consecutive round into territory #1265 does not scope.

This is my last review round on this PR (the loop budget for it was six; this is the seventh, and rounds five through seven have each been consequences of the previous round’s fix in the modality area). Everything above is either fixed with a discriminating test or refuted in writing.

ishiland:
Filed the distribution-reconciliation follow-up as #1314.

ishiland:
@codex review

ishiland:
Round 8 (verification pass on the round-7 fixes). Both new findings check out, and neither is in the code this round was verifying — both are pre-existing properties this PR inherits from the two re-upload strategies rather than introduces. I am not fixing them here; both are filed.

**Initialize the tile cache before invalidating it — real, repo-wide, filed as #1315.** Confirmed: `init_tile_cache()` has exactly one call site, the API lifespan (`api/main.py:328`). The worker is a separate process and never calls it, so `get_tile_cache()` returns `None` and `invalidate_tile_cache_for_table()` is a no-op there. That applies identically to `reupload_file` (`tasks_reupload.py:489`) and `reupload_service` (`tasks_reupload.py:1051`), which have called it since fix(#394) B-019/VT-01 — so if the reading is right, worker-side tile eviction has never worked, and fixing it changes behaviour for all three strategies at once. That is a repo-wide change with its own blast radius, not a line in this PR.

I have corrected the PR description accordingly: the `_v=` bump is the part that demonstrably works (it is a catalog column the frontend reads, so browser and CDN busting is real), and the Valkey purge is best-effort and currently inert from the worker. Better to state that than to keep a claim the code does not support.

**Finalize the job when its dataset has been deleted — real, and deliberately identical to its siblings.** The `dataset is None` early return leaves the ingest job `pending` until the stale sweep rewrites it. True — and byte-identical to `reupload_file` (`tasks_reupload.py:212`) and `reupload_service` (`tasks_reupload.py:798`), which take the same path for the same reason. Diverging here would give three strategies two different answers to "your dataset vanished mid-flight", which is the failure mode handoff invariant 11 exists to prevent. It should be fixed for all three together, in the shared prologue, and I have noted it on #1315 rather than opening a third issue for a one-line divergence.

Stopping the review loop here. The budget for this PR was six rounds; this is the eighth, rounds five through seven were each consequences of the previous round’s fix, and round eight surfaced only pre-existing repo-wide behaviour. Final state: every finding across all eight rounds is either fixed with a test verified to fail against the unfixed code, or refuted/deferred in writing with a filed follow-up (#1314, #1315).

PR #1321, opened 2026-08-09
sec(catalog): apply owner-or-admin provenance projection to dataset reads and versions

### LINKED ISSUE #1316: sec(catalog): decide one provenance projection — dataset reads, /versions/, and MCP disagree
Three surfaces currently project dataset provenance three different ways, and the strictest one contradicts the loosest one. A decision is needed before #1285 builds Source-panel UI on top of the loosest surface — changing a projection after UI ships on it is a breaking change; deciding now is cheap.

## The three disagreeing surfaces (verified on `main` at 403027cba)

1. **`GET /datasets/{id}` and list reads** — `DatasetResponse` (`backend/app/modules/catalog/datasets/domain/schemas.py:264`, fields at :342-368) returns `origin`, `origin_uri`, and the full `origin_ref` dict to ANY reader who passes the visibility check, including anonymous readers of public datasets. For a registered-PostGIS dataset that is the live schema-qualified table name; for a service dataset it is the upstream service URL and layer reference.
2. **`GET /datasets/{id}/versions/`** (`backend/app/modules/catalog/datasets/api/router_metadata.py:60-87`) — anonymous-readable on public datasets and returns `file_hash` and `uploaded_by` unredacted (`DatasetVersionResponse`, `schemas.py:809-822`). A public dataset's version history therefore enumerates its editors (the exact leak ADR-002 Decision 4e closed on refresh-runs) and exposes content hashes.
3. **`GET /datasets/{id}/refresh-runs`** (`backend/app/modules/catalog/datasets/api/router.py:598`) — the strict model: ADR-002 Decision 4e redacts `triggered_by`, failure text, and the schema diff for any caller who is neither owner nor admin, tested against both a named third party and an anonymous reader.
4. **MCP server** (`mcp/geolens_mcp/client.py:37`) — the strictest: `_RAW_SOURCE_POINTER_FIELDS = {source_url, origin_uri, origin_ref}` are stripped unconditionally, even for owners.

So an anonymous browser request can read the exact fields the MCP server refuses to show the dataset's own owner, while the sibling refresh-runs endpoint redacts a strict superset of what /versions/ freely returns.

## Decision to make

Adopt one policy. The recommended shape is the refresh-runs model, extended:

- **Owner-or-admin sees everything** (they already can via refresh-runs).
- **Other readers of an accessible dataset** get capability-relevant summary fields (`origin` kind, `source_freshness`, `source_health`, `last_refreshed_at`) but NOT raw pointers (`origin_uri`, `origin_ref`), NOT `uploaded_by`, NOT `file_hash`.
- **MCP** then stops needing its own unconditional strip — it inherits the server-side projection (its current behavior over-redacts for owners).

Whatever is decided, record it as an ADR-002 amendment so the next surface (e.g. #1289's reupload dialog) doesn't re-litigate it.

## Implementation once decided

- [ ] Owner/admin-conditional projection for `origin_uri`/`origin_ref` in dataset reads (single and list; check search projections from #1304 while there)
- [ ] Same conditioning for `uploaded_by` + `file_hash` in `DatasetVersionResponse`
- [ ] Tests against a NAMED signed-in third party AND an anonymous reader (the refresh-runs docstring explains why anonymous-only reads as complete and is not)
- [ ] Re-evaluate the MCP `_RAW_SOURCE_POINTER_FIELDS` strip (likely removable)
- [ ] OpenAPI regen (3 artifacts) if field nullability changes

## Why this blocks #1285

The Source panel will render version history and origin detail from these endpoints. If it ships against today's projection, the redaction later becomes a visible regression ("the panel used to show the hash") instead of a policy the UI was built against.

ishiland:
## Summary

Adopts the refresh-runs redaction model (ADR-002 Decision 4e) as the single provenance projection across dataset reads and `/versions/`, per the decision recorded on #1316 (2026-08-09).

Before this change, three surfaces disagreed:
- `GET /datasets/{id}` and list reads served `origin_uri`/`origin_ref` to any accessible-dataset reader, including anonymous readers of public datasets.
- `GET /datasets/{id}/versions/` served `file_hash`/`uploaded_by` unredacted to the same audience — a public dataset's version history enumerated its editors.
- `GET /datasets/{id}/refresh-runs` redacted a strict superset of that information for the same audience (owner-or-admin only).

An anonymous browser request could read fields the MCP server refuses to show the dataset's own owner.

## Changes

- Extracts the owner-or-admin predicate refresh-runs already used into `can_view_dataset_provenance()` in `backend/app/modules/catalog/authorization.py`, and points the refresh-runs endpoint at it instead of its own inline copy.
- Applies the predicate at every `DatasetResponse` construction site: single dataset read, dataset list, and collection dataset listing (`dataset_to_response()` gains a `can_view_provenance` parameter, defaulting to `False` so a call site that forgets to decide redacts rather than leaks).
- Applies the same predicate to `GET /datasets/{id}/versions/`, nulling `file_hash`/`uploaded_by` for non-owner/non-admin readers.
- The capability summary (`origin`, `source_health`, `source_freshness`, `last_refreshed_at`, `last_checked_at`) is never gated — it survives redaction for everyone, matching the decision.
- MCP's unconditional strip is untouched, per the decision (re-evaluating it is a follow-up).
- `origin_uri`/`origin_ref`/`DatasetVersionResponse` field descriptions now state the redaction inline, which is the only reason `openapi.json` and the generated SDKs shifted — no shape change, every affected field was already `X | None`.

## Verification

- `cd backend && set -a && source ../.env.test && set +a && uv run pytest tests/test_provenance_projection_1316.py tests/test_datasets.py tests/test_search.py tests/test_dataset_refresh_runs.py tests/test_reupload.py tests/test_collections.py tests/test_provenance_responses.py tests/test_dataset_source_state.py tests/test_vrt_catalog_175.py tests/test_antimeridian_extent.py -q` — 460 passed
- `cd backend && set -a && source ../.env.test && set +a && uv run pytest tests/test_layering.py tests/test_rule1_structural.py -q` — 57 passed
- `make openapi-check` and `make sdks-check` (both pass once this branch's regen artifacts are committed; TypeScript SDK `npm run build && npm test` also verified separately)

New test coverage (`tests/test_provenance_projection_1316.py`) asserts redaction against both a named signed-in third party and an anonymous reader for dataset reads and `/versions/`, confirms the owner and a non-owner admin still see full fields, confirms the capability summary survives redaction, and confirms per-row (not per-page) redaction on collection dataset listings.

Closes #1316

### DIFF
--- a/backend/app/modules/catalog/authorization.py
+++ b/backend/app/modules/catalog/authorization.py
@@ -456,6 +456,27 @@ async def visible_lineage_summary(
     return (await visible_lineage_summaries(db, [record], user, user_roles))[record.id]
 
 
+def can_view_dataset_provenance(
+    record: Any, user: Identity | None, user_roles: set[str]
+) -> bool:
+    """Owner-or-admin predicate for the provenance projection (#1316).
+
+    Decided 2026-08-09 (ADR-002 amendment): the refresh-runs redaction model
+    (Decision 4e) is the single provenance projection for every surface that
+    carries it — dataset reads (single, list, collection), ``/versions/``,
+    and refresh-runs itself. The dataset's owner and any admin see raw
+    provenance in full (``origin_uri``, ``origin_ref``, ``uploaded_by``,
+    ``file_hash``); every other reader of an accessible dataset, named or
+    anonymous, gets those fields nulled and keeps only the capability summary
+    (origin kind, ``source_freshness``, ``source_health``,
+    ``last_refreshed_at``, ``last_checked_at``).
+
+    Extracted from the inline check refresh-runs used before this decision so
+    every surface applies the same rule rather than a hand-copied one.
+    """
+    return bool(user and (record.created_by == user.id or "admin" in user_roles))
+
+
 async def check_dataset_write_access(
     db: AsyncSession,
     dataset: Any,
--- a/backend/app/modules/catalog/collections/router.py
+++ b/backend/app/modules/catalog/collections/router.py
@@ -15,6 +15,7 @@ from app.core.identity import Identity
 from app.modules.auth.dependencies import get_optional_user, require_permission
 from app.modules.auth.models import User
 from app.modules.catalog.authorization import (
+    can_view_dataset_provenance,
     check_datasets_access_bulk,
     get_user_roles,
     visible_lineage_summaries,
@@ -443,7 +444,13 @@ async def get_collection_datasets_endpoint(
     return DatasetListResponse(
         datasets=[
             dataset_to_response(
-                d, actors_by_id=actor_map, lineage_summary=lineage[d.record_id]
+                d,
+                actors_by_id=actor_map,
+                lineage_summary=lineage[d.record_id],
+                # feat(#1316): per-row, same reasoning as the plain dataset list.
+                can_view_provenance=can_view_dataset_provenance(
+                    d.record, user, user_roles
+                ),
             )
             for d in datasets
         ],
--- a/backend/app/modules/catalog/datasets/api/router.py
+++ b/backend/app/modules/catalog/datasets/api/router.py
@@ -24,6 +24,7 @@ from app.modules.auth.dependencies import (
     require_permission,
 )
 from app.modules.catalog.authorization import (
+    can_view_dataset_provenance,
     check_dataset_access_or_anonymous,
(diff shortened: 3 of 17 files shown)


ishiland:
@codex review

ishiland:
@codex review

Issue #1226, opened 2026-08-05
feat(frontend): freshness and health chips in search and the dataset header


ishiland:
Blocked by the staleness and health issues in this milestone. Line refs as of 2026-08-05.

Search results and the dataset header currently say nothing about whether data is fresh or whether its source is still alive. Separately, `frontend/src/components/dataset/OriginBadge.tsx:9` still calls a service dataset "a mirrored remote service" in its doc comment; the code copies once at import and the visible string says "Registered from a remote service URL" (`i18n/locales/en/common.json:33`), so the comment is wrong on both counts and predates the #1217 vocabulary.

## Scope

- Freshness and health chips on search result cards and the dataset header, driven by the computed staleness state and persisted `source_health` from the API.
- Overdue and inaccessible states get a visible, accessible treatment: icon plus text, never color alone.
- No chip for the unknown state; silence beats noise on datasets with no declared cadence.
- Copy pass over the origin badge labels and tooltips against the #1217 vocabulary; fix the stale comment.

## Where to look

- Render sites: `frontend/src/components/search/SearchResultCard.tsx` (OriginBadge already renders at ~:145 and ~:244; put the chips beside it) and `frontend/src/components/dataset/DatasetDetailHeader.tsx` (the header already shows origin, record type, catalog status, and access badges; follow that pattern).
- The badge component to imitate: `OriginBadge.tsx` (small, icon plus label plus tooltip, driven by a discriminated union). Build the freshness/health chip the same way.
- Copy: the `origin.*` block in `i18n/locales/en/common.json:22-41` is the tone to match; new keys go in all four locales (`en`, `es`, `fr`, `de`).
- UI primitives: badge/tooltip components in `frontend/src/components/ui/`.

## Suggested path

1. Add a `FreshnessChip` (and reuse it for health) as a pure presentational component with a unit test per state, including the render-nothing unknown case.
2. Wire into the card and header from fields already present on the API responses; no new fetching.
3. Do the origin copy pass: align tooltips with the #1217 vocabulary, fix the `OriginBadge.tsx:9` comment.
4. Accessibility: chip text readable by screen readers (no title-attribute-only tooltips), visible focus, dark mode checked. The axe e2e suite is gating on touched paths.

## Acceptance

- [ ] Chips on cards and header for due/overdue and missing/inaccessible states; unknown renders nothing
- [ ] Icon plus text, not color alone; dark mode checked
- [ ] Origin copy aligned with #1217; stale comment fixed
[...]

PR #1322, opened 2026-08-09
fix(vrt): project generation timestamps into last_refreshed_at, reconcile abandoned regenerations

### LINKED ISSUE #1267: fix(vrt): project generation timestamps into last_refreshed_at and reconcile abandoned regenerations
ADR-002 Amendment A10 (ownership gap found by the 2026-08-07 milestone audit). Two VRT gaps with no owner:

1. **Freshness projection (Decision 5a):** the ADR promises VRT's latest generation timestamp projects into `last_refreshed_at` so #1226's chips render uniformly — nothing implements the projection. #1224's freshness function already handles the NULL-origin VRT case on the read side; this issue owns the write.
2. **Abandoned-generation reconciliation:** a VRT stuck in `regenerating` after process death stays there forever — no reconciliation exists. Audit blocker 11.

## Scope

- VRT regeneration success path stamps the owning dataset's `last_refreshed_at` in the same transaction as the generation swap.
- A reconciliation pass (sweep-style, mirroring #1219's stale-run sweep pattern): a `regenerating` generation whose job is proven gone → restore prior `active` generation state, mark the attempt failed, clean staging artifacts. Proven-dead check first — never reconcile a live job.
- Attempt fencing via the full-UUID helper (`attempt_scoped_staging_table`, Amendment A3).

## Acceptance

- [ ] kill the worker mid-regeneration → sweep restores `active`, VRT tiles keep serving
- [ ] successful regeneration updates `last_refreshed_at`; freshness chip reflects it
- [ ] sweep never touches a generation with a live job
- [ ] gates: whole-tree backend, layering

Related: ADR-001 D8 (VRT lifecycle stays its own — this projects state out and repairs it, it does not rewire refresh into VRT).

ishiland:
ADR-002 Amendment A10 (ownership gap found by the 2026-08-07 milestone audit). Two VRT gaps had no owner:

1. **Freshness projection (Decision 5a).** The ADR promises a VRT's latest generation timestamp projects into `last_refreshed_at` so #1226's freshness chips render uniformly across record types. #1224's read-side freshness function already treats a NULL origin (a VRT) as this dataset's refreshable kind — nothing implemented the write side, so a regenerated VRT kept reporting its creation-time floor forever.
2. **Abandoned-generation reconciliation (Audit blocker 11).** A `regenerating` VRT asset whose worker died mid-attempt had no path back to serving. GAP-002's existing sweep already proves the attempt dead via the generation's own heartbeat timing out, but it recovered the asset to `status='failed'` unconditionally — the wrong outcome for most attempts, since `regenerate_vrt` only overwrites the published pointer (`asset_uri`, `sha256`, ...) in the same transaction that clears `current_generation_id` on success, so a dead attempt never reaches that transaction and those fields still describe exactly what the VRT was serving before the attempt started.

## What changed

- `regenerate_vrt`'s success path now stamps the owning dataset's `last_refreshed_at` at the generation's own `completed_at` instant, in the same transaction as the generation swap — one timestamp, no clock skew between the two records of it.
- `sweep_stale_vrt_assets` now restores `status='ready'` for a dead **composition-preserving** attempt (a pure regenerate) instead of unconditionally `status='failed'`. Only the dead attempt's own `VrtGeneration` row is marked `failed`; the VRT keeps serving what it served before.
- **A dead composition-changing attempt (source add/remove) is kept `'failed'`, not restored to `'ready'`.** `add_vrt_source`/`remove_vrt_source` commit their `vrt_source_links` mutation in the same transaction that flips the asset to `'regenerating'`, before the regeneration itself ever runs — so if that attempt dies, the catalog's declared composition is already ahead of the served bytes, and restoring `'ready'` would erase the only visible signal of that drift (the status endpoint reads the link set, not the served VRT). There is no stored "attempt type" to key off (`VrtGeneration.triggered_by` holds a user id on every call site, not a kind), so the sweep asks the question the drift is actually about, from stored state: does the published composition (`built_from`'s key set — what the VRT was actually assembled from) still equal the catalog's current composition (`vrt_source_links`)? A count-match plus a one-directional subset check proves set equality; a legacy VRT with no recorded `built_from` falls to the same conservative branch as a proven mismatch. Covered by 4 new database-backed tests (composition preserved / source removed / source added / `built_from` absent).
- The sweep resolves (never deletes) the dead attempt's own generation-scoped storage keys (`rasters/{dataset}/generations/{generation}/source.vrt` + 2 quicklooks) and returns them to its caller. **Deletion is deferred until strictly after the caller's own commit lands** — threaded through `StaleCleanupOutcome` exactly like the existing `_staged_paths` field for `fail_stale_jobs`, and reaped directly after `session.commit()` in the worker startup recovery path. Deleting inside the sweep, before the reconciling transaction was durable, could leave a rolled-back commit's "still-regenerating" asset pointing at a generation whose bytes were already gone.
- **Composition-preserving alone is still not sufficient to restore `'ready'`.** `regenerate_vrt_endpoint`'s guard rejects only `'regenerating'`, so a caller may retry an asset that is already `'failed'`. If that retry is also abandoned and its membership still matches `built_from`, the composition check alone would call it composition-preserving and restore `'ready'` — erasing a real failure a worker crash had nothing to do with, since no successful artifact was ever produced by the retry and whatever caused the *original* failure is still unaddressed. There is no column recording "the asset's status the instant this attempt started" (`previous_status` in the router is a local variable, never persisted), so the sweep reconstructs the nearest available fact: at most one generation is ever in-flight per dataset (the same guard + advisory lock), so `vrt_generations` rows for one dataset form a strict, gap-free timeline, and the row immediately preceding the one being reconciled records what happened right before this attempt was allowed to start. `'completed'` (or no such row — the dataset's first-ever attempt) means the asset was `'ready'`; `'failed'` means it was not, and the sweep now keeps the asset `'failed'` in that case too. **Accepted conservative cost, stated plainly:** a generation this sweep itself restores to `'ready'` still leaves its own `vrt_generations` row `'failed'` (only the asset recovers, not the attempt's record) — so a *later* dead attempt on the same dataset, whose immediately-prior row is that earlier swept-and-recovered one, reads `'failed'` and is kept `'failed'` even though the asset was legitimately `'ready'` when it started. A further stored marker distinguishing "recovered by the sweep" from "never recovered" would close this, but it's a bigger schema surface than this fix's blast radius, and the safety direction is correct for a reconciler: never claim `'ready'` the moment there is doubt, at the cost of occasionally staying `'failed'` one cycle longer than strictly necessary — which self-corrects the moment an operator retries and it succeeds. Covered by 2 new database-backed tests (the exact failed→retry→abandon scenario, and its companion where the prior attempt genuinely completed).
- **The prior-attempt marker itself needed refining: raw generation status over-counts what "failed" means.** `regenerate_vrt_endpoint`'s own orphan-guard rollback marks a just-created `VrtGeneration` `'failed'` when the Procrastinate *enqueue* itself throws — the task never reached a worker — and in the same rollback reverts the asset to whatever it already was (commonly `'ready'`). That generation row genuinely never ran; reading its `'failed'` status as "the asset was not ready" would produce a false `'failed'` for an asset that never had a real attempt run against it. `heartbeat_at IS NOT NULL` is the fact that separates them: it's set in exactly one place in this codebase, `tasks_vrt.regenerate_vrt`'s Phase-1 claim (the first thing the *task* does once a worker actually picks it up) — never anything the endpoint touches at creation. An enqueue failure never reaches that line, so its generation keeps `heartbeat_at` NULL forever; the same is true of a generation stuck `'pending'` because no worker ever claimed it before the sweep's own cutoff, which is the identical "never actually ran" fact. Filtering the prior-attempt lookup to `heartbeat_at IS NOT NULL` finds the most recent OTHER generation a worker actually claimed and ran, skipping past any enqueue-failures or claim-starved rows in between. Covered by a new database-backed test.
- **The generation-staleness proof itself had a gap: `started_at` age alone cannot tell an orphaned `'pending'` generation from one still legitimately queued through a sustained worker backlog.** Queue waits are unbounded — the exact reason `stale_pending_clauses` (a few lines above, for `IngestJob`) already requires `no_live_procrastinate_job()` before declaring a pending row abandoned. The VRT sweep's staleness check lacked that guard, so it could reconcile a generation whose task was still going to run, and when it eventually did, its own Phase-1 claim (`tasks_vrt.regenerate_vrt`) would fail against a generation the sweep had already flipped to `'failed'` — losing a valid regeneration. Split the check by status: `'running'` keeps the heartbeat-only proof (still sufficient on its own — and *necessary*, since a crashed worker can leave its `procrastinate_jobs` row reading `'doing'` until the separate stalled-queue sweep prunes it, so requiring "no live row" there would make a genuinely-dead running generation unsweepable until that unrelated sweep runs); `'pending'` now additionally requires no live Procrastinate row referencing it, correlated via the `generation_id` kwarg every dispatch site passes (mirrors `_ABANDONED_RUN_SQL`'s `args->>'job_id'` correlation for refresh runs). Covered by 2 new database-backed tests (a queued-but-unstarted generation with a live job is left alone; the same shape with no live job is reaped).
- **That live-job correlation was itself incomplete: it only recognized the modern delivery shape.** `tasks_vrt.regenerate_vrt`'s `generation_id` argument is optional precisely so a delivery queued by pre-upgrade code (no `generation_id` in its args at all) survives a rolling deploy — on execution it falls back to adopting `RasterAsset.current_generation_id`. The correlation as first written only matched on `generation_id`, so it was blind to a live legacy delivery, which could still be reaped out from under it. Added a second correlation form: a live `procrastinate_jobs` row with no `generation_id` counts as live for a generation if that generation is *currently* the dataset's `current_generation_id` — the exact row such a delivery will claim once it runs. `vrt_dataset_id` is a required (non-optional) argument on every delivery shape, so it's always available to correlate through. Kept deliberately conservative in the direction the finding specified: an ambiguous or legacy-shaped live row blocks the sweep rather than being read as absence of one. Covered by 2 more database-backed tests (a live legacy delivery protects its generation; a dataset with no live row of either shape still gets reaped).
- The `VrtGeneration` stale-UPDATE's `RETURNING` widened from `id` to `(id, vrt_dataset_id)` — the pairing the storage-key reconstruction needs, and one the asset-side UPDATE can't supply back since its own `current_generation_id` is nulled in the same statement.
- A stale doc comment on the regenerate endpoint claiming "no stale-cleanup sweep exists" (true before GAP-002, false and now doubly false after this PR) is corrected to point at the real reconciliation path.
- Fixed a CI-only failure in `test_tenant_staging_storage_keys.py::test_cleanup_and_retention_reaper_delete_only_tenant_key`, outside this PR's original targeted test set. **Determination: stale mock, not an unwired caller.** That test builds `fail_stale_jobs`'s session manually with a fixed-length `side_effect` list, counted for the VRT sweep's earlier 2-statement shape; splitting the `RasterAsset` UPDATE into two left the fixture one result short, so the purge-DELETE mock landed on the wrong statement and its `RETURNING` data never reached the reap step — `storage.delete` was never awaited. `fail_stale_jobs` already threads the purge's `RETURNING` through to `_reap_committed_staged_paths` correctly (proven independently by this PR's own coverage); this was purely a stale statement count in an unrelated test file's hand-built mock. Widened it to match, and the tenant-scoping property the test exists to prove (delete only the tenant-prefixed key) still holds on the post-commit reap path.

Scope guard respected (ADR-001 D8): this projects VRT state out and repairs it; it does not rewire refresh into the VRT lifecycle.

## Follow-up candidate (not implemented here)

The root wound this PR works around: `vrt_source_links` mutation and VRT regeneration are two separate transactions with no compensation between them. This sweep can only detect the drift and refuse to hide it — it cannot repair it (reconciling the link set back against the published artifact, or vice versa, would be active repair logic inside a reconciliation sweep, which is more machinery than a sweep should own). Making the two atomic or compensable is a real fix worth its own issue.

## Tests

- `backend/tests/test_vrt_stale_sweep_gap002.py` — updated for the split ready/degraded contract and the resolve-then-defer storage contract. New: restore-vs-degrade via compiled SQL, key resolution without any storage I/O, a storage-failure-never-masks-reconciliation test, a commit-failure test proving a dead attempt's storage objects survive a rolled-back reconciliation, and 4 real-database tests for the composition-preserving/changed discrimination.
- `backend/tests/test_ingest_job_attempt_fencing.py` — `test_vrt_generation_heartbeat_fences_stale_recovery` updated to the new status and 3-tuple return.
- `backend/tests/test_regenerate_vrt_integration.py` — new assertion [16]: `Dataset.last_refreshed_at` is stamped and equals `generation.completed_at`.
- `backend/tests/test_worker.py` — startup-recovery mocks updated for the sweep's widened result shape.
- `backend/tests/test_tenant_staging_storage_keys.py` — the CI-caught mock-count fix described above.

## Verification

```
uv run pytest tests/test_vrt_stale_sweep_gap002.py tests/test_regenerate_vrt_integration.py tests/test_ingest_job_attempt_fencing.py tests/test_dataset_source_freshness.py tests/test_worker.py tests/test_jobs_router.py -v
```

No schema/API changes: no Alembic migration, `openapi.json` unaffected (no response-shape change).

Closes #1267

### DIFF
--- a/backend/app/modules/catalog/datasets/api/router_vrt.py
+++ b/backend/app/modules/catalog/datasets/api/router_vrt.py
@@ -489,11 +489,14 @@ async def regenerate_vrt_endpoint(
 
     await db.commit()
 
-    # Dispatch task with orphan guard.
-    # No stale-cleanup sweep exists for VRT ``status="regenerating"``
-    # or for ``VrtGeneration`` rows — a Procrastinate outage would
-    # leave the VRT permanently stuck and the generation row dangling
-    # until manual operator intervention.
+    # Dispatch task with orphan guard. This closes the SYNCHRONOUS failure —
+    # Procrastinate unreachable at enqueue time — by reverting the mutation
+    # below before it is ever visible. A worker that dies AFTER a successful
+    # dispatch is a different failure and is not this guard's job: fix(#1267)
+    # gives ``sweep_stale_vrt_assets`` (GAP-002) a periodic + startup pass
+    # that reconciles it — restoring the asset to ``'ready'`` and failing the
+    # orphaned ``VrtGeneration`` row — so neither state is stuck on manual
+    # operator intervention.
     async def _defer() -> None:
         await defer_async_with_tenant(
             get_catalog_port().regenerate_vrt_task(),
--- a/backend/app/platform/jobs/router.py
+++ b/backend/app/platform/jobs/router.py
@@ -523,29 +523,88 @@ async def _reap_committed_staged_paths(
     )
 
 
+async def _reap_stale_generation_storage(
+    stale_generations: list[tuple[uuid.UUID, uuid.UUID]],
+) -> None:
+    """Best-effort delete of a dead regeneration attempt's immutable objects.
+
+    feat(#1267): ``regenerate_vrt`` writes its rebuilt VRT + quicklooks to an
+    immutable ``rasters/{vrt_dataset_id}/generations/{generation_id}/...`` key
+    (the full generation UUID, unset by attempt-fencing convention A3 —
+    mirroring ``attempt_scoped_staging_table``'s full-hex rule for staging
+    tables) BEFORE the phase-2 transaction that would otherwise clean them up
+    via ``_cleanup_orphaned_storage_keys``. A worker killed between that write
+    and the commit leaves the objects with no reference anywhere — this sweep
+    is the only remaining owner of the key, so it reaps them the same way.
+
+    A dead attempt may have written none, some, or all three objects before
+    the worker died; deleting a key that was never written is a documented
+    no-op on every StorageProvider.
+
+    ``current_tenant_var`` carries the right tenant because both callers of
+    ``sweep_stale_vrt_assets`` (the startup recovery pass and the periodic
+    sweep) run this inside ``tenant_job_context`` per tenant in multi-tenant
+    mode — the same context ``regenerate_vrt`` itself reads to resolve these
+    same keys. A missing tenant context in multi-tenant mode skips cleanup
+    rather than raising: the asset/generation reconciliation above already
+    landed regardless of whether this best-effort pass can find the bytes.
+    """
+    from app.core.tenancy import is_multi_tenant
+    from app.core.db.tenant_session import current_tenant_var
+    from app.platform.storage.titiler_url import resolve_storage_key
+
+    tenant_id = current_tenant_var.get()
+    if is_multi_tenant() and tenant_id is None:
+        return
+
+    keys = [
+        resolve_storage_key(
+            f"rasters/{vrt_dataset_id}/generations/{generation_id}/{suffix}",
+            tenant_id=tenant_id,
+        )
+        for generation_id, vrt_dataset_id in stale_generations
+        for suffix in ("source.vrt", "quicklook_256.png", "quicklook_512.png")
+    ]
+
+    # Function-local: mirrors the deferred processing-import convention
+    # already used for RasterAsset/VrtGeneration below (D-17,
+    # test_platform_processing_imports_stay_deferred) — the exact same helper
+    # regenerate_vrt itself uses to reap its own orphaned writes.
+    from app.processing.ingest.tasks_raster import _cleanup_orphaned_storage_keys
+
+    await _cleanup_orphaned_storage_keys(keys, job_id="vrt-stale-sweep")
+
+
 async def sweep_stale_vrt_assets(
     db: AsyncSession,
     stale_cutoff: datetime,
 ) -> tuple[int, int]:
-    """Reset RasterAsset rows stuck in status='regenerating' past ``stale_cutoff``.
+    """Reconcile RasterAsset rows stuck in status='regenerating' past ``stale_cutoff``.
 
-    GAP-002: a worker crash mid-regeneration leaves the VRT asset permanently
-    stuck in ``status='regenerating'``, causing all future link/regenerate
-    calls to 409. This helper mirrors the IngestJob stale-recovery pattern and
-    uses the same ``stale_cutoff = now - JOB_TIMEOUT_SECONDS`` threshold.
+    GAP-002 / feat(#1267): a worker crash mid-regeneration leaves the VRT
+    asset permanently stuck in ``status='regenerating'``, causing all future
+    link/regenerate calls to 409. This helper mirrors the IngestJob
+    stale-recovery pattern and uses the same
+    ``stale_cutoff = now - JOB_TIMEOUT_SECONDS`` threshold.
 
-    Staleness is measured via the associated ``VrtGeneration.started_at`` for
-    the asset's ``current_generation_id``, falling back to querying all
-    ``VrtGeneration`` rows whose ``started_at`` predates the cutoff and whose
-    ``status`` is still pending/running.
+    Staleness is measured via ``VrtGeneration.heartbeat_at`` (falling back to
+    ``started_at`` for a generation whose worker died before its first
+    renewal) — the same self-reported liveness signal ``maintain_vrt_
+    generation_heartbeat`` renews every 30s while the worker is alive, so a
+    live job's row never crosses the cutoff and this sweep never touches it.
 
-    Recovery status: ``'failed'`` — mirrors the explicit failure-handler path
-    in ``tasks_vrt.regenerate_vrt`` (which sets ``status='failed'`` on any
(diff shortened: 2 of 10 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

Issue #1285, opened 2026-08-08
feat(frontend): wire the Source panel to the refresh endpoint


ishiland:
The read-only Source panel (#1280) shows origin, freshness, and health but has no actions. The refresh endpoint it needs landed in #1277 (`POST /datasets/{dataset_id}/refresh`), and the generated TS types plus the Python SDK method already exist, so this is frontend wiring only.

## Scope

- A "Refresh from source" action on the Source panel, gated to datasets whose origin supports it (the endpoint answers 409 `refresh_not_applicable` with `origin_kind` for the rest; hide or disable the control from the same information the panel already renders).
- Render the refusal taxonomy as distinct, actionable states rather than a generic error toast:
  - 409 `dataset_busy`: a run is already active; point at the run status the panel shows.
  - 409 `origin_changed`: the binding changed between check and dispatch; prompt to re-review the source.
  - 409 `origin_unavailable`: the stored binding is too incomplete to re-address the layer.
  - 503 `credential_store_unavailable`: this install cannot do credentialed refresh; public-source refresh still works.
  - 400: the stored URL failed SSRF validation.
  - 422 `invalid_service_token`: the token does not meet the service's token policy (the body names the allowed alphabet; it never echoes the input).
- Token entry for protected sources: a field that is never logged, never echoed back, never persisted client-side beyond the request.
- On 202, surface the accepted run (id is in the response) so the panel's run history picks it up.

## Constraints

- All API calls through `apiFetch()`; reuse `components/ui/` primitives.
- New `t()` keys in all four locales (en/es/fr/de); `npm run test:i18n` gates parity.

Refs: #1277 (endpoint), #1280 (panel), #1225 (original panel issue).

PR #1333, opened 2026-08-10
fix(frontend): map the STAC refresh refusal wordings to their taxonomy keys

### LINKED ISSUE #1332: STAC refresh refusals render the generic 409/400 fallback instead of mapped copy
Live QA of the merged #1266 surface (post-#1326) found that dispatching a refresh on a pre-#1326 STAC dataset shows the generic "The request conflicts with the resource's current state." instead of the mapped, actionable copy.

Cause: `frontend/src/lib/error-map.ts` maps refresh-door refusals by exact backend literal (the #1285 convention), and #1326 introduced new STAC wordings that were never added to the table:

1. 409 `refresh_not_applicable` — "This dataset was not imported from a STAC item, so there is no item to re-resolve." → `errors.refreshNotApplicable`
2. 409 `origin_unavailable` — "This dataset's source binding does not record the STAC item its asset was published in, …" → `errors.refreshOriginUnavailable`
3. 409 `origin_unavailable` — "GeoLens cannot tell this dataset's STAC item from another one its stored URL might serve: …" → `errors.refreshOriginUnavailable`
4. 422 `credential_not_applicable` — "Refreshing a STAC dataset re-reads a public item document and needs no credential. …" → `errors.refreshCredentialNotApplicable`
5. 400 SSRF — "This dataset's stored STAC item URL is not reachable: …" — the existing prefix regex anchors on "stored source URL" and misses the STAC wording → `errors.refreshSourceUrlBlocked`

All five target keys already exist in all four locales (`common.json`), with origin-agnostic copy — no i18n change needed. The impact is real for upgrades: every STAC dataset imported before v1.11.0 refuses with wording (2) or (3), and the generic fallback drops the "re-import" recovery instruction.

Found during Milestone 5 close-out live QA.

ishiland:
## What

Adds the four STAC refresh-door refusal literals introduced by #1326 to the exact-text table in `frontend/src/lib/error-map.ts`, and widens the SSRF prefix regex to also match the STAC door's "stored STAC item URL is not reachable:" wording.

## Why

The error-map keys refusals by exact backend literal (the #1285 convention). #1326 added new STAC wordings without table entries, so dispatching a refresh on a STAC dataset whose binding predates item-identity tracking rendered the generic "The request conflicts with the resource's current state." instead of the mapped, actionable copy ("Re-import or re-register the source to fix this."). Every STAC dataset imported before this release refuses through exactly that path, so the generic fallback is the first thing an upgrading user sees. Found during Milestone 5 close-out live QA; see #1332 for the enumerated set.

## Impact

Frontend only. No API, schema, or config change. All five target locale keys already exist in en/es/fr/de `common.json` with origin-agnostic copy — no i18n change.

## Verification

- `cd frontend && npx vitest run src/lib/__tests__/error-map.test.ts` — 23 passed (new: one case per STAC literal + the STAC SSRF wording)
- `npm run lint`, `npm run typecheck` — clean

Closes #1332

### DIFF
--- a/frontend/src/lib/__tests__/error-map.test.ts
+++ b/frontend/src/lib/__tests__/error-map.test.ts
@@ -92,6 +92,48 @@ describe('API error localization boundary', () => {
     ).toEqual({ key: 'errors.refreshDatasetBusy' });
   });
 
+  it('maps the STAC door refusal wordings introduced by #1266 (#1332)', () => {
+    expect(
+      classifyApiError(
+        {
+          code: 'refresh_not_applicable',
+          message: 'This dataset was not imported from a STAC item, so there is no item to re-resolve.',
+        },
+        409,
+      ),
+    ).toEqual({ key: 'errors.refreshNotApplicable' });
+    expect(
+      classifyApiError(
+        {
+          code: 'origin_unavailable',
+          message:
+            "This dataset's source binding does not record the STAC item its asset was published in, so GeoLens cannot ask the catalog where that asset is now. Re-import it from the STAC catalog to record one.",
+        },
+        409,
+      ),
+    ).toEqual({ key: 'errors.refreshOriginUnavailable' });
+    expect(
+      classifyApiError(
+        {
+          code: 'origin_unavailable',
+          message:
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

PR #1358, opened 2026-08-10
fix(datasets): stop returning raw exception text from single delete

### LINKED ISSUE #1317: fix(datasets): single-item delete forwards raw exception text to the client
Sibling of #1297 (fixed for bulk delete in #1309): the single-item delete endpoint still forwards raw exception text to the client.

## Problem

`delete_dataset_endpoint` (`backend/app/modules/catalog/datasets/api/router.py:515-525`, verified on `main` at 403027cba):

```python
except ValueError as exc:
    msg = str(exc)
    if "not found" in msg.lower():
        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=msg)
    raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=msg)
```

`delete_dataset()` (`domain/service_lifecycle.py`) raises three distinct ValueErrors, and this handler cannot tell them apart except by message-sniffing:

1. `DatasetTitleMismatchError("Dataset title does not match confirmation")` — intended-public, caller-caused. Fine to echo.
2. `ValueError("Dataset not found")` — a race (the router already 404s before calling); echoing is harmless but sniffing for it is fragile.
3. `ValueError(f"Invalid table name: {table_name}")` — corrupted persisted metadata. This is INTERNAL STATE (the stored physical table name) reaching the client as a 400, and it is not a caller error at all.

## Fix shape (mirror #1309 exactly)

Type-discriminate, never message-sniff for the leak case:

- `except DatasetTitleMismatchError` → 400 with `str(exc)` (public by design; the type exists since #1309 and is exported from the `domain/service.py` façade).
- `except ValueError` where the message is the not-found case → keep the 404 but return the fixed string `"Dataset not found"` (same literal the pre-check uses) instead of echoing.
- Any other `ValueError` → `logger.exception(...)` with the dataset id server-side, and a fixed generic detail (`"Dataset deletion failed unexpectedly"`, same literal as bulk delete) with a 500 — a malformed persisted table name is a server-state problem, not a bad request.

## Acceptance

- [ ] Monkeypatch `delete_dataset` to raise `ValueError("Invalid table name: secret_tbl_xyz")`; assert the response detail contains neither `secret_tbl_xyz` nor the raw message, and that the full exception is logged server-side (see `test_bulk_delete_unexpected_exception_hides_detail` in `backend/tests/test_datasets.py` for the working pattern, including the structlog capture note).
- [ ] Title-mismatch and not-found behavior unchanged from the API consumer's perspective (existing tests keep passing; frontend relies on the 400-vs-404 split).
- [ ] `uv run pytest tests/test_datasets.py -q` and `tests/test_layering.py -q` green.

## Traps

- Import `DatasetTitleMismatchError` via the `domain/service.py` façade, never from `service_lifecycle` directly (`test_layering.py` enforces).
- Do not change the `DependentVrtError` 409 branch above this handler — its structured detail is intended-public and the frontend consumes it.

prashit-vora:
## Summary

The single-dataset delete endpoint returned `str(exc)` for every `ValueError`
and identified not-found errors by searching the exception message. This could
expose internal details and report unexpected server failures as client errors.

- Catch `DatasetTitleMismatchError` explicitly and preserve its intended public
  400 response.
- Return the fixed `"Dataset not found"` detail for the deletion race as a 404.
- Log any other `ValueError` server-side and return a generic 500 response.
- Add regression coverage proving internal table information does not reach the
  HTTP response and the complete failure is still logged.

This mirrors the bulk-delete error handling introduced in #1309.

No schema or configuration changes.

Closes #1317

## Testing

- `uv run pytest tests/test_datasets.py::TestSingleDeleteDataset::test_unexpected_value_error_hides_detail -vv` — 1 passed
- `uv run ruff check .` — passed
- `uv run ruff format --check .` — passed (940 files already formatted)

### DIFF
--- a/backend/app/modules/catalog/datasets/api/router.py
+++ b/backend/app/modules/catalog/datasets/api/router.py
@@ -518,16 +518,24 @@ async def delete_dataset_endpoint(
                 "dependent_vrts": exc.dependents,
             },
         )
-    except ValueError as exc:
-        msg = str(exc)
-        if "not found" in msg.lower():
-            raise HTTPException(
-                status_code=status.HTTP_404_NOT_FOUND,
-                detail=msg,
-            )
+    except DatasetTitleMismatchError as exc:
         raise HTTPException(
             status_code=status.HTTP_400_BAD_REQUEST,
-            detail=msg,
+            detail=str(exc),
+        )
+    except ValueError as exc:
+        if str(exc) == "Dataset not found":
+            raise HTTPException(
+                status_code=status.HTTP_404_NOT_FOUND,
+                detail="Dataset not found",
+            )
+        logger.exception(
+            "Unexpected error during dataset delete",
+            dataset_id=str(dataset_id),
+        )
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail="Dataset deletion failed unexpectedly",
         )
 
     await audit_emit(
--- a/backend/tests/test_datasets.py
+++ b/backend/tests/test_datasets.py
@@ -2011,3 +2011,48 @@ class TestBulkDeleteDatasets:
             and record.get("dataset_id") == str(ds.id)
             for record in captured
         ), f"expected the exception to be logged server-side; got: {captured}"
+
+
+class TestSingleDeleteDataset:
+    """Tests for DELETE /datasets/{dataset_id}."""
+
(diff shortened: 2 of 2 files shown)


ishiland:
@codex review

ishiland (approved):
Nice work! Thanks for the PR.

PR #1360, opened 2026-08-10
fix(ingest): stamp file_hash on first-ingest raster origin

### LINKED ISSUE #1294: ingest: first-ingest raster origins lack file_hash, unlike the replace path
create_raster_dataset stamps the upload origin without a file_hash, while the replace path (reupload_raster) records the source sha256 on its origin row. ADR-002 Decision 7 names the provenance hash as part of the justification for deleting the pre-conversion source after a sample-preserving conversion, so first-ingest rasters are missing the provenance the deletion policy leans on.

Fix shape: compute and persist the hash in the first-ingest tail the same way the replace tail does (the source read already happens for metadata extraction, so the hash is one pass over bytes already being read). Both tails should derive it at the same point so the two cannot drift — the same one-authority pattern the rest of the pipeline uses.

Where to look:
- backend/app/processing/ingest/tasks_raster.py (first-ingest tail, create_raster_dataset call)
- backend/app/processing/ingest/tasks_raster_replace.py (the replace tail's origin stamping, as the reference)

Refs #1290, ADR-002 Decision 7.

ishiland:
## What / why

`create_raster_dataset` already computed `source_sha256` (it hashes the raster's own `RasterAsset.source_sha256` column) but never forwarded it to `set_dataset_origin`, so a raster dataset's `origin_ref.file_hash` was absent until its first replace. The replace tail (`_write_swapped_fields` in `tasks_raster_swap.py`) already passes `file_hash=source_sha256` through the same `set_dataset_origin` authority — first-ingest now does the same.

ADR-002 Decision 7 leans on this provenance hash as part of the justification for deleting the pre-conversion source after a sample-preserving conversion, so first-ingest rasters were missing that provenance.

The fix is one line: forwarding a value `create_raster_dataset` already receives as a parameter into the `set_dataset_origin` call it already makes. Both tails now derive/stamp the hash through the same shared helper (`app/platform/dataset_origin.set_dataset_origin`), so they cannot drift by construction — no new hashing pass or new helper needed.

None. `origin_ref` is a JSONB column and `file_hash` is already an allowed key for the `upload` origin kind (used by the replace path since #1218/#1290). No migration required.

## Verification

- `cd backend && uv run ruff check .` — pass
- `cd backend && uv run ruff format --check .` — pass
- Failing-first test added: `tests/test_raster_ingest.py::TestFirstIngestOriginFileHash::test_create_raster_dataset_stamps_origin_file_hash` — confirmed failing against pre-fix code (stashed the fix and re-ran), passing after the fix.
- `uv run pytest tests/test_raster_ingest.py -v` — 21 passed
- `uv run pytest tests/test_raster_replace_1221.py tests/test_manifest_apply_roundtrip.py tests/test_manifest_apply_service.py -q` — 148 passed

Closes #1294

### DIFF
--- a/backend/app/processing/ingest/tasks_raster_common.py
+++ b/backend/app/processing/ingest/tasks_raster_common.py
@@ -165,7 +165,13 @@ async def create_raster_dataset(
     # feat(#1218): a raster dataset IS the COG; the pre-conversion upload is a
     # transient input, so the origin is the uploaded file and there is no
     # remote URI to point at (ADR-002 Decision 7).
-    set_dataset_origin(dataset, "upload", filename=source_filename)
+    # fix(#1294): file_hash was missing here, unlike the replace tail's
+    # equivalent call in tasks_raster_swap.py — both go through this same
+    # set_dataset_origin authority, so passing the sha256 the caller already
+    # computed for `source_sha256` is the whole fix.
+    set_dataset_origin(
+        dataset, "upload", filename=source_filename, file_hash=source_sha256
+    )
     session.add(dataset)
     await session.flush()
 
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

PR #1363, opened 2026-08-10
fix(stac): recover an empty-string keyed asset after it moves

### LINKED ISSUE #1331: stac: an asset keyed by the empty string is stored but never recovered by key
Residue from PR #1326's final review round (review 4893066201, stac_resolve.py:384).

`""` is a valid JSON property name, so a STAC item may key its data-role asset as the empty string. Search surfaces that key, the import model's length bound accepts it, and the binding stores it — but the resolver's recovery path tests the stored key with truthiness, so `asset_key=""` is treated as no recorded key. While the asset href still matches, nothing is visible; once the publisher moves the asset, href matching cannot recover the entry and the refresh reports the asset unidentified (inaccessible) even though the durable key still names it.

Fail-closed availability edge, not a rebinding vector: pointers stay untouched and the run says inaccessible. Fix direction: test `asset_key is not None` wherever the binding's key is consulted, and decide at capture whether an empty-string key is worth recording at all — refusing it at capture (like the length bound does for over-long keys) would keep the resolver's truthiness reads honest instead of special-casing them.

ishiland:
## What

`""` is a legal JSON property name, so a STAC item can key its data-role asset under the empty string. Search surfaced that key and the import model's length bound accepted it, but the resolver's recovery path (`_bound_asset_key` in `stac_resolve.py`) tested the stored key with truthiness, so `asset_key=""` was treated identically to "no key recorded." While the asset's href still matched, nothing was visible; once the publisher moved the asset, href matching alone could not recover the binding and refresh reported the asset unidentified (inaccessible) even though the durable key still named it.

## Fix

- `stac_resolve.py`: both consultations of the binding's stored key (`_bound_asset_key` and the removed-vs-unidentified branch in `_resolve_from_item`) now test `asset_key is not None` instead of truthiness.
- `adapters/stac.py`: `storable_asset_key` now refuses `""` at capture, the same way it already refuses an over-long key — so from here forward nothing ever writes an empty-string key into a binding (including the resolver's own write-back after a successful resolve).
- `stac_router.py`: the import handler now routes `data_asset_key` through `storable_asset_key` too, closing the gap where a raw API caller bypassing search (whose response already normalizes `""` away) could still submit `data_asset_key=""` directly — Pydantic's `max_length` does not reject an empty string.

Net effect: a legacy binding that already recorded `asset_key=""` is honored on its next resolve (recovers a moved asset correctly), and that resolve's write-back leaves the binding keyless going forward — the same fallback-to-href-match state any other keyless binding is in. No binding can be captured with `asset_key=""` again after this change.

## Schema/API/config impact

None. No migration, no new fields, no API contract change.

## Verification

```
uv run pytest tests/test_stac_refresh_1266.py tests/test_stac_import.py tests/test_refresh_gate_1269.py -v   # 180 passed
uv run pytest tests/test_layering.py -q   # 40 passed (cap raised +13 LOC with rationale)
```

Closes #1331

### DIFF
--- a/backend/app/modules/catalog/sources/adapters/stac.py
+++ b/backend/app/modules/catalog/sources/adapters/stac.py
@@ -56,8 +56,17 @@ MAX_ASSET_KEY_CHARS = 255
 
 
 def storable_asset_key(key: str | None) -> str | None:
-    """The asset key if it is short enough to carry, else None."""
-    if key is None or len(key) > MAX_ASSET_KEY_CHARS:
+    """The asset key if it is non-empty and short enough to carry, else None.
+
+    fix(#1331): ``""`` is a legal JSON property name, so an item may key its
+    data asset under the empty string — and every consultation of the stored
+    key downstream (``stac_resolve.py``) tests it with truthiness, which
+    cannot tell a recorded ``""`` apart from no key at all. Refusing it here,
+    the same way an over-long key is refused, keeps those truthiness reads
+    describing a real invariant instead of quietly mismatching this one edge
+    forever.
+    """
+    if not key or len(key) > MAX_ASSET_KEY_CHARS:
         return None
     return key
 
--- a/backend/app/modules/catalog/sources/adapters/stac.py
+++ b/backend/app/modules/catalog/sources/adapters/stac.py
@@ -56,17 +56,20 @@ MAX_ASSET_KEY_CHARS = 255
 
 
 def storable_asset_key(key: str | None) -> str | None:
-    """The asset key if it is non-empty and short enough to carry, else None.
+    """The asset key if it is short enough to carry, else None.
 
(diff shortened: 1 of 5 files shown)


ishiland:
@codex review

ishiland:
Fixed in 0a4041e7e — good catch. `storable_asset_key` no longer refuses ``: it has no length problem, so once every downstream read tests `is not None` there is nothing left to special-case at capture. The import handler's redundant second capture-time check is reverted too. Re-triggering review.

ishiland:
@codex review

ishiland:
Fixed in 30a73a81a — the docstring was stale from the reverted first design. It now says capture preserves `` and distinguishes it from no key, matching the actual behavior.

ishiland:
@codex review

PR #1367, opened 2026-08-10
docs(quota): state the ownerless-dataset exemption policy and pin it

### LINKED ISSUE #1293: quota: datasets with NULL created_by count against nobody on every quota seam
Both the request-time admission checks (upload, replacement, presigned completion) and the worker-side reservation (reserve_storage_bytes / reserve_replacement_bytes) resolve the billed identity from dataset.record.created_by. For legacy datasets where created_by is NULL, every seam skips accounting entirely: the bytes are stored but never reserved against any user, and the door checks admit without projecting.

This is a pre-existing policy, deliberately mirrored (with a pinned test) rather than half-corrected during #1290 so the two seams could not disagree. The open question is what the policy should be: bill the instance-admin pool, refuse mutation until ownership is assigned, or leave ownerless datasets exempt but documented.

Where to look:
- backend/app/modules/quota/service.py (check_replacement_quota, reserve_storage_bytes)
- backend/app/processing/ingest/tasks_raster_swap.py (reserve_replacement_bytes)
- The parity tests in backend/tests/test_raster_replace_1221.py pin the mirroring.

Refs #1290 (review rounds 3 and 5 established the mirroring).

ishiland:
Closes #1293.

## The question, and the answer

#1293 asked which policy datasets with `records.created_by IS NULL` should get at the quota seams: bill the instance-admin pool, refuse mutation until ownership is assigned, or leave them exempt and say so. The answer is **exempt**, and this PR makes that a stated policy with tests behind it rather than a property of a SQL filter nobody wrote down.

An ownerless dataset's bytes and its dataset row count against nobody at every seam, and no seam substitutes a stand-in identity for it.

## Why exempt

**Refusing mutation has no remedy.** There is no ownership-assignment surface in the product, and 0019's adoption path is not reachable without a destructive downgrade (#998), so an operator told "assign an owner first" has nothing to do. It would permanently brick legacy datasets. Worse, ownerless is not only a legacy state — `catalog.records.created_by` is `ON DELETE SET NULL` and `AdminService.delete_user` relies on exactly that, so deleting one user orphans every dataset they created. Under a refusal policy, deleting a user would freeze their datasets for everyone, forever.

**Billing the admin misattributes storage.** The admin user list reads `get_user_quota_usage_bulk`, so one operator would appear to hold the entire orphaned catalog — and once caps are enabled, that phantom usage would refuse the admin's own legitimate uploads.

**Exempt is cheap.** Both caps default to 0 (unlimited), so only an instance that opts into enforcement has a gap at all. The code already mirrored the exemption at both seams deliberately (the `check_replacement_quota` docstring says so) so they could not diverge — this keeps that property and names it.

The durable fix is ownership adoption, tracked by #998.

## What changed

The policy is stated once, in `backend/app/modules/quota/service.py`'s module docstring. The per-seam docstrings (`check_replacement_quota`, `reserve_storage_bytes`, `reserve_dataset_slot`, `get_user_quota_usage`) and `reserve_replacement_bytes` in `tasks_raster_swap.py` point at it instead of each re-deriving it.

Two things the statement records that were not written down anywhere before:

- **The gap is capped but not frozen.** Creation always carries an authenticated uploader, so no new dataset can be born ownerless — but replacing an already-orphaned one writes uncounted bytes, so the exempt pool can grow by the size of what is already in it. Stated as a cost being accepted, not as harmlessness.
- **Usage reading zero is not a short-circuit.** Nothing accumulates, and the count cap can never refuse a NULL owner, but the byte cap still weighs the incoming reservation on its own — so one reservation larger than the whole cap is still refused. An `if user_id is None: return` would look equivalent and would change behaviour. There is a test on that boundary.

Signatures now admit the nullability they already had at runtime: `get_user_quota_usage`, `reserve_storage_bytes`, `reserve_dataset_slot`, `reserve_replacement_bytes`, and both presigned completion functions (plus their `CatalogPort` protocol declarations) annotated `user_id: uuid.UUID` while three callers already passed `dataset.record.created_by`.

## Seam coverage

`TestOwnerlessDatasetsAreExemptAtEverySeam` in `backend/tests/test_raster_replace_1221.py` — extending the suite that already pinned the door/worker mirroring rather than starting a parallel one. Seven tests: the aggregate that the exemption physically lives in, attribution (an orphan next to an identical owned dataset lands on neither the user nor the NULL identity), door admission (the same two datasets judged before and after orphaning — ownership is the only variable), `reserve_storage_bytes` including the residual edge, `reserve_dataset_slot`, the worker-side `reserve_replacement_bytes` pass-through, and a structural pin that all three replacement admission points bill `dataset.record.created_by`.

**A seam #1293 did not list:** `get_user_quota_usage_bulk`, the aggregate behind the admin user list. It resolves the same column by `GROUP BY`, so ownerless datasets appear in no user's row there either — that is the surface where a misattribution would show up first, and it is covered by the attribution test.

## Impact

**No behaviour change. No schema, API, or config impact.** Docstrings, widened type annotations, and tests only. No new endpoints, no migration, no OpenAPI or SDK regeneration.

## Verification

- `cd backend && uv run ruff check .` — pass
- `cd backend && uv run ruff format --check .` — pass
- `uv run pytest tests/test_raster_replace_1221.py` — 108 passed
- `uv run pytest tests/test_1224_quota_service.py tests/test_302_quota_reservation.py tests/test_quota_bulk_usage.py tests/test_1224_ingest_quota_enforcement.py tests/test_me_usage.py` — 30 passed
- `uv run pytest tests/test_reupload.py tests/test_tenant_staging_storage_keys.py tests/test_manifest_apply_service.py` — 92 passed
- `uv run pytest tests/test_presigned.py tests/test_presigned_completion_size.py tests/test_presigned_content_validation_1202.py tests/test_presigned_staging_reap_1202.py tests/test_reupload_idor.py tests/test_extension_slot_guard.py tests/test_openapi_overlay_boundary.py` — 108 passed, 1 skipped
- `uv run pytest tests/test_rule1_structural.py tests/test_rule2_structural.py` — 82 passed
- `uv run pytest tests/test_layering.py` — 40 passed

### DIFF
--- a/backend/app/core/catalog_port.py
+++ b/backend/app/core/catalog_port.py
@@ -53,6 +53,10 @@ class CatalogPort(Protocol):
         job_id: uuid.UUID,
     ) -> None: ...
 
+    # Here and on `finalize_presigned_object`, `user_id` is the dataset OWNER
+    # when the flow is a replacement and the uploader when it is a creation, so
+    # it is nullable: an ownerless dataset has no owner to name (#1293 — the
+    # policy is stated in app.modules.quota.service).
     async def verify_completed_presigned_upload(
         self,
         *,
@@ -60,7 +64,7 @@ class CatalogPort(Protocol):
         storage: Any,
         key: str,
         expected_size: Any,
-        user_id: uuid.UUID,
+        user_id: uuid.UUID | None,
         request: Any,
         job_id: uuid.UUID,
     ) -> int: ...
@@ -93,7 +97,7 @@ class CatalogPort(Protocol):
         logical_key: str,
         expected_size: Any,
         filename: str,
-        user_id: uuid.UUID,
+        user_id: uuid.UUID | None,
         request: Any,
     ) -> str: ...
 
--- a/backend/app/modules/quota/service.py
+++ b/backend/app/modules/quota/service.py
@@ -3,6 +3,65 @@
 Core check is authoritative for community and enterprise editions.
 The EntitlementPort enforce_limit calls are an additive cloud seam (QUOTA-03):
 in OSS/Enterprise the DefaultEntitlementPort is grant-all and never raises.
+
+Ownerless datasets are exempt (policy, #1293)
+---------------------------------------------
+Every seam in this module resolves the billed identity from
+``catalog.records.created_by``, and that column is nullable. A dataset whose
+``created_by`` is NULL is EXEMPT from quota accounting at every seam: its bytes
+and its dataset row count against nobody, and no seam substitutes a stand-in
+identity for it. That is the decided policy, not an accident of the SQL.
+
+The exemption is one mechanism repeated rather than six special cases.
+``get_user_quota_usage`` filters ``records.created_by = :user_id``, and
+``= NULL`` is never true, so the aggregate returns zeros for a NULL identity
+and skips ownerless rows for every real one. Every other function here reads
+its usage through that aggregate, so all of them inherit the same answer;
+``reserve_replacement_bytes`` in ``app.processing.ingest.tasks_raster_swap``
+inherits it by delegating to ``reserve_storage_bytes``. Nothing needs an
+``if owner_id is None`` branch, and adding one to a single seam is how the
+seams would start to disagree.
+
+Ownerless is a LIVE state, not only a pre-0019 legacy one:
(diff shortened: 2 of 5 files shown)


ishiland:
@codex review

PR #1368, opened 2026-08-10
fix(sources): request every ArcGIS field so imports keep their attributes

### LINKED ISSUE #1359: Service imports and analysis outputs populate column_info with only the id column
## What happens

A dataset imported from an ArcGIS FeatureServer ends up with `column_info` containing a single entry, `id`, even though the imported table has ~29 attribute columns (the import-time schema preview showed `OBJECTID`, `ID`, `MAG`, `EVENTTYPE`, `SIG`, `ALERT`, "+23 more columns", and magnitude-based styling works against the stored table). A materialized analysis output (`select_by_location`) shows the same single-column state, so this is not specific to the service path.

Observed on a 1.11.0 dev stack:

- `GET /api/datasets/?search=USGS` returns `column_info` with length 1 (`['id']`) for the service-imported dataset.
- The dataset detail stats bar renders "Columns 1" (`DatasetStatsBar` shows `column_info.length`) next to "Features 2,128", directly contradicting the Structure tab and the import preview.
- A refresh-from-source run (v1 -> v2, succeeded, before/after counts populated) does not repopulate `column_info` either.

## Why it matters

- The stats bar is wrong on every dataset page for these ingest paths, and it's the kind of contradiction a GIS user notices ("Columns 1" over a rich attribute table).
- Anything else consuming `column_info` (API clients, SDKs, MCP reads) under-reports the schema.

## Repro

1. Import a layer via Service URL (e.g. a public ArcGIS FeatureServer layer with many fields).
2. Open the dataset page: stats bar shows "Columns 1".
3. `GET /api/datasets/?search=<title>` and inspect `column_info`.
4. Run Refresh from source; `column_info` is unchanged.

File uploads appear to populate `column_info` correctly, so the likely home is the metadata-population step of the service-import (and materialize) paths, with the refresh path also missing a repopulation hook.

ishiland:
Closes #1359

## Root cause

The issue reported `column_info == ['id']` on an ArcGIS FeatureServer import whose preview listed ~29 fields. The stored value was **not** wrong about the table — the table itself only had one attribute column.

`build_gdal_source` (`backend/app/modules/catalog/sources/preview.py`) composed the ArcGIS `/query` URL as `?f=json&where=1=1[&orderByFields=…]` with no `outFields`. An ArcGIS `/query` with no `outFields` answers with the layer's **display field alone**, so ogr2ogr faithfully loaded geometry plus that one field and every other attribute was dropped before PostGIS ever saw it.

Verified against the exact service from the report:

```
$ curl '…/USGS_Seismic_Data_v1/FeatureServer/0/query?f=json&where=1%3D1&resultRecordCount=1'
"fields":[{"name":"id",…}]           # one field

$ curl '…/query?f=json&where=1%3D1&outFields=%2A&resultRecordCount=1'
fields: OBJECTID, id, mag, eventType, sig, alert, place, hoursOld, …   # 31 fields
```

And against the dev database that produced the report — the live table matches its stored `column_info` exactly:

```
 column_name |     data_type     
-------------+-------------------
 gid         | integer
 id          | character varying
 geom        | USER-DEFINED
 geom_4326   | USER-DEFINED
```

The preview reads the layer's native `?f=json` field list (`adapters/arcgis.py`) rather than this URL, which is why it promised 29 columns the import never asked for.

**The other two symptoms in the issue are consequences of this one, not separate defects:**

- The materialized `select_by_location` output showed one column because `_list_carry_columns` carries the source's *live* columns, and its source was the broken import. An analysis output built from a healthy source has always catalogued its full column list (the dev DB's `buildings_selected` has 9). A guard test now pins that.
- Refresh "never repopulated" because it re-fetched through the same builder and recomputed the same one-column answer. `_apply_reupload_swap` has always written `dataset.column_info` from the fresh staging metadata — so with this fix, a refresh now *repairs* an already-broken dataset rather than needing a new hook.

## What changed

**`backend/app/modules/catalog/sources/preview.py`** — the ArcGIS branch of `build_gdal_source` now sends `outFields=*`. This is the single site every service path composes its URL at, so first import (single-shot and paginated), refresh-from-source, and the ogrinfo preview are all fixed at once. `urlencode` renders the value as `outFields=%2A`, which ArcGIS decodes back to `*` (verified live, above).

**`backend/app/processing/ingest/service.py`** — `register_existing_table` derived no metadata at all for a non-spatial table, cataloguing it with `column_info` and `feature_count` NULL: the same "the stats bar contradicts the schema" state, reached a different way. Registration now runs the one derivation for every table it registers. `extract_metadata` already returns `srid`/`geometry_type`/`extent_wkt` as `None` when there is no geom column, so the spatial fields land exactly as before.

Deliberately **not** changed: the `source_columns` fallback in `_finalize_ingest` still fires only when the landed table has *no* attribute columns. Broadening it to cover a partial column set would write names from the upstream API schema that don't exist in the table, and `column_info` is what `validate_where_clause` and the feature-write property allowlist check against. It must keep describing the live table.

## Impact

- **Schema:** none. No migration, no model change.
- **API:** none. No route, request, or response shape changes; `backend/openapi.json` is untouched.
- **Config:** none.
- **Behavioural:** ArcGIS imports and refreshes now fetch every field, so the response payload per page grows to the layer's full width — which is what the preview always promised. Existing broken datasets are repaired by running Refresh from source; they are not backfilled automatically.

## Tests

New file `backend/tests/test_service_column_info_1359.py`. Five of the six fail on `main` with the issue's exact symptom and pass with the fix; the sixth passes on both, which is the evidence that the analysis symptom was a cascade:

| test | pre-fix |
|---|---|
| `test_arcgis_gdal_source_requests_every_field` | FAIL |
| `test_arcgis_gdal_source_requests_every_field_when_paging` | FAIL |
| `test_arcgis_service_import_lands_every_attribute_column` | FAIL — `['id'] != ['id','mag','eventtype','place','depth']` |
| `test_service_refresh_repopulates_wrong_stored_column_info` | FAIL — same |
| `test_register_non_spatial_table_populates_column_info` | FAIL — `[] != ['code','population']` |
| `test_materialize_output_column_info_matches_carried_columns` | PASS (guard) |

The two worker-driven tests stand in for ogr2ogr with a fake that materialises exactly the columns the composed URL asked the service for, so they fail the moment the URL stops requesting every field.

`backend/tests/test_layering.py` — `_MODULE_LOC_CAPS` for `processing/ingest/service.py` raised 1073 → 1077 (exact), with a comment recording what the four lines bought.

## Verification

All run from the worktree against Postgres at `localhost:5434`.

```
cd backend && uv run ruff check .                                    # All checks passed
cd backend && uv run ruff format --check .                           # 941 files already formatted
uv run pytest tests/test_service_column_info_1359.py                 # 6 passed
uv run pytest tests/test_layering.py                                 # 40 passed
uv run pytest tests/test_arcgis_auth.py tests/test_services_endpoints.py \
             tests/test_ingest_progress.py                           # 58 passed
uv run pytest tests/test_reupload_service.py tests/test_service_refresh_1220.py \
             tests/test_dataset_refresh_runs.py                      # 145 passed
uv run pytest tests/test_ingest.py tests/test_dataset_source_state.py \
             tests/test_staging_pipeline.py \
             tests/test_staging_pipeline_integration.py              # 180 passed
uv run pytest tests/test_analysis_materialize.py                     # 99 passed
uv run pytest tests/test_rule1_structural.py tests/test_rule2_structural.py \
             tests/test_postgis_refresh_1265.py \
             tests/test_tenant_attribution_materialize.py \
             tests/test_analysis_curved_sources.py \
             tests/test_service_token_redaction_1277.py \
             tests/test_url_redaction_sec.py                         # 305 passed
uv run pytest tests/test_processing_port.py tests/test_preview_token_sec021.py \
             tests/test_reupload_idor.py tests/test_reupload.py      # 62 passed
```

Pre-fix failure reproduction: `git stash` of the two source files, rerun of the new file — 5 failed, 1 passed, with the assertion messages quoted in the table above.

### DIFF
--- a/backend/app/modules/catalog/sources/preview.py
+++ b/backend/app/modules/catalog/sources/preview.py
@@ -47,7 +47,17 @@ def build_gdal_source(
             raise ValueError("ArcGIS layer preview requires a layer ID")
         safe_base_url = _encode_url_for_gdal(base_url.rstrip("/"))
         safe_layer_id = quote(str(layer_id).strip("/"), safe="")
-        params: dict[str, str | int] = {"f": "json", "where": "1=1"}
+        # fix(#1359): a /query with no outFields returns the layer's display
+        # field alone, so the import landed geometry plus ONE attribute column
+        # and dropped every other field. The preview reads the layer's ?f=json
+        # field list and therefore promised columns this fetch never asked
+        # for. urlencode renders the value as `outFields=%2A`, which ArcGIS
+        # decodes back to `*`.
+        params: dict[str, str | int] = {
+            "f": "json",
+            "where": "1=1",
+            "outFields": "*",
+        }
         if order_field:
             params["orderByFields"] = f"{order_field} ASC"
         if result_limit is not None:
--- a/backend/app/processing/ingest/service.py
+++ b/backend/app/processing/ingest/service.py
@@ -617,7 +617,6 @@ async def register_existing_table(
     # (_current_tenant_schema()='data', _current_tenant_role()='geolens_reader').
     _grant_role = _current_tenant_role()
 
-    metadata = {}
     if has_geom:
         if not has_4326:
             srid = await get_table_srid(session, table_name, schema=_schema)
@@ -648,11 +647,16 @@ async def register_existing_table(
                     f"Failed to linearize geom_4326 on '{table_name}': {exc}"
                 ) from exc
 
-        await grant_reader_access(session, table_name, schema=_schema, role=_grant_role)
-        metadata = await extract_metadata(session, table_name, schema=_schema)
-    else:
-        # Non-spatial table -- grant access but skip spatial metadata
-        await grant_reader_access(session, table_name, schema=_schema, role=_grant_role)
+    await grant_reader_access(session, table_name, schema=_schema, role=_grant_role)
+
+    # fix(#1359): one derivation for every registration, spatial or not. The
+    # non-spatial branch used to skip this entirely and register the table
+    # with column_info and feature_count NULL — the same "the stats bar
+    # contradicts the schema" state the ArcGIS import produced, reached a
+    # different way. extract_metadata already reports srid, geometry_type,
+    # and extent_wkt as None for a table with no geom column, so the spatial
+    # fields land exactly as they did before.
+    metadata = await extract_metadata(session, table_name, schema=_schema)
 
     # Extract sample values for attribute metadata example_values
     col_info = metadata.get("column_info", [])
--- a/backend/tests/test_layering.py
+++ b/backend/tests/test_layering.py
@@ -2623,7 +2623,12 @@ _MODULE_LOC_CAPS: dict[str, int] = {
     # applies. Deriving from the raw filename split the key: the logical URI
     # kept a path component the write stripped, so the counted row pointed at
     # nothing. Cap 1059 -> 1073, exact.
-    "backend/app/processing/ingest/service.py": 1073,
+    # fix(#1359): +4 — register_existing_table now derives metadata for every
+    # table it registers instead of only the spatial ones, so a non-spatial
+    # registration stops landing with column_info and feature_count NULL. The
+    # added lines are the comment explaining why the branch is gone.
+    # Cap 1073 -> 1077, exact.
+    "backend/app/processing/ingest/service.py": 1077,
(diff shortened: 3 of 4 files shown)


ishiland:
@codex review

PR #1371, opened 2026-08-10
fix(worker): initialize the tile cache in the shared bootstrap

### LINKED ISSUE #1315: fix(worker): worker-side MVT tile invalidation is a no-op — init_tile_cache never runs there
Found during the #1313 review (round 8). Pre-existing and repo-wide; not introduced by that PR.

`invalidate_tile_cache_for_table()` (`backend/app/processing/ingest/tasks_common.py:1628`) resolves the provider through `get_tile_cache()`, which returns the module-level singleton set by `init_tile_cache()`. The only call site for `init_tile_cache()` is the API lifespan (`backend/app/api/main.py:328`). The Procrastinate worker is a separate process and never calls it, so `get_tile_cache()` returns `None` there and the invalidation silently does nothing:

```python
tile_cache = get_tile_cache()
if tile_cache is not None:
    await tile_cache.invalidate_table(table_name)
```

Every worker-side caller is affected:

- `reupload_file` (`tasks_reupload.py:489`)
- `reupload_service` (`tasks_reupload.py:1051`)
- `refresh_postgis` (`tasks_postgis_refresh.py:891`, added by #1313)

The purge was added by fix(#394) B-019/VT-01 precisely because a re-upload swaps a table's contents under the same name and the MVT cache key has no content-version dimension. If this reading is right, that fix has never actually taken effect from the worker, and cached tiles keep being served for up to `tile_cache_ttl` after every re-upload — the exact failure it was written to prevent.

The `_v=` tile-URL cache-buster (`Dataset.bump_tile_cache_version`) is unaffected: it is a catalog column the frontend reads, so browser and CDN busting still works. What is lost is the server-side eviction, and the server cache key ignores `_v`, so the bump does not evict those entries either.

## What to check first

Confirm the reading before building anything — a worker-side `init_tile_cache()` may already happen somewhere indirect (a shared bootstrap import), in which case this is a non-issue and worth a comment saying so.

## Scope if confirmed

- Initialize the tile cache in the worker process alongside whatever `bootstrap()` already sets up, so the Redis-backed provider is live there. Note the in-memory fallback is NOT a valid substitute in the worker: a per-process LRU would evict entries nobody reads and leave the API's entries untouched, which is the current behaviour with extra steps. When `REDIS_URL` is unset, worker-side invalidation genuinely cannot reach the API's in-memory cache and the honest answer is to say so rather than pretend.
- Or make invalidation cross-process (a pub/sub channel, or a content-version dimension in the cache key so stale entries are simply never read).

## Acceptance

- [ ] A re-upload from the worker demonstrably evicts the API's cached MVT entries for that table (asserted against a live provider, not a mock)
- [ ] The `REDIS_URL`-unset case has a stated, tested behaviour rather than a silent no-op
- [ ] The three call sites above are covered

ishiland:
## What

`invalidate_tile_cache_for_table()` resolves its provider through the `get_tile_cache()` singleton, and `init_tile_cache()` ran in the API lifespan only. The Procrastinate worker is a separate process, so `get_tile_cache()` returned `None` there and all three post-swap MVT purges returned without evicting anything.

`init_tile_cache()` moves into the shared `bootstrap()` that both entrypoints already call, and the lifespan's copy goes away so the two cannot drift apart again.

## Step 0: the reading is confirmed

The issue asked for confirmation before building anything, in case some indirect bootstrap import already initialized the tile cache in the worker. It does not.

Statically, on `origin/main` at e515e0150, `init_tile_cache` had exactly one call site in the whole repo: `backend/app/api/main.py:328`. The shared `bootstrap()` calls `init_cache()` (the JSON cache) and never the tile cache. `worker.py:main()` calls `bootstrap(app=None)`, and the worker's only other startup surface is a bare Starlette health app (`observability/health/worker.py`) with no lifespan to hang an init off. `scripts/worker-entrypoint.sh` execs `python -m app.worker`, a shim over `app.platform.jobs.worker.main`.

Empirically, importing the worker's full module chain and asking:

```
bootstrap module mentions init_tile_cache: False
worker module mentions init_tile_cache: False
get_tile_cache() after worker-module import chain: None
```

And 8 of the 13 new tests fail on the unfixed tree, including the live-Valkey one.

## REDIS_URL unset

The worker deliberately does not take the PERF-01 in-memory fallback. It never reads tiles, so a process-local LRU there holds nothing; purging it evicts zero entries and logs `tile_cache_invalidated` anyway, which is the same no-op with a success message attached. With `REDIS_URL` unset the worker leaves the singleton unset and warns once at boot:

```
tile_cache_unavailable_in_worker
  reason="REDIS_URL is unset"
  consequence="worker-side MVT purges after a reupload or PostGIS refresh cannot
               reach the API process's in-memory tile cache; the API keeps serving
               pre-swap tiles for up to tile_cache_ttl"
  remediation="set REDIS_URL so both processes share one tile cache"
```

At boot rather than per purge, because the condition is a static deployment property. Repeating it on every swap would be the same line carrying no new information. `.env.example` gets the operator-facing version, since "leave unset to use in-memory caching (single-instance default)" was already glossing over the fact that a single instance is still two processes.

The API's behaviour with `REDIS_URL` unset is unchanged. It still gets the LRU, and a test pins that.

## Acceptance

**(a) A live provider, not a mock.** `test_worker_bootstrap_provider_evicts_api_tiles_live` caches a tile through one `TileCacheProvider`, builds the worker's singleton the way `bootstrap(app=None)` builds it, runs the real `invalidate_tile_cache_for_table`, and asserts the first provider's read now misses. It runs against a real Valkey (`docker run -d --rm --name w6-valkey -p 16379:6379 valkey/valkey`) and skips when nothing is listening.

Every live test asserts the tile is readable *before* the purge. That control is load-bearing: `TileCacheProvider` turns backend errors into cache misses, so without it an unreachable Valkey would be indistinguishable from a successful eviction and the test could pass while proving nothing.

The cross-process test also runs under `fakeredis` with a shared `FakeServer`, which is what covers this in CI, where no Valkey runs. Residual gap: CI exercises the eviction logic but never a real Redis wire.

Because the test suite still shares one process, I also ran it as three separate OS processes against the live Valkey, with the middle one calling the real `bootstrap(app=None)` rather than any test helper:

```
[A api]    cached tile -> b'pre-swap-mvt-bytes'
--- separate process ---
[B worker] get_tile_cache() after bootstrap -> <class '...tile_cache.TileCacheProvider'>
           tile_cache_invalidated  table=w6_crossproc_demo
[B worker] invalidate_tile_cache_for_table done
--- separate process ---
[C api]    read back -> None
           OK: the worker's purge evicted the API's cached tile
```

An earlier run where B aborted before the purge gave the negative control for free: C reported `read back -> b'pre-swap-mvt-bytes'` and failed. Booting B with `REDIS_URL` unset produces `get_tile_cache() -> NoneType` plus the warning quoted above.

**(b) REDIS_URL unset is stated and tested.** `test_worker_bootstrap_without_redis_logs_why_purges_cannot_land` asserts the warning event, `test_worker_bootstrap_refuses_in_memory_fallback_without_redis` asserts the singleton stays unset, and `test_in_memory_worker_cache_would_not_have_evicted_anything` asserts the rejected alternative genuinely fails: a live API-cached tile survives an in-memory worker purge.

**(c) All three call sites.** The fix is at the singleton, so all three are covered at once. `test_every_worker_swap_path_purges_the_tile_cache` pins the counts per module (2 in `tasks_reupload`, 1 in `tasks_postgis_refresh`), so a fourth swap path added without a purge fails there rather than in production tiles. The `refresh_postgis` site is now `tasks_postgis_refresh.py:840`, not the `:891` the issue cites.

Schema: none, no migration. API: none, and `backend/openapi.json` is untouched. Config: no new variables. `REDIS_URL` is already plumbed into the worker service in both compose files (`docker-compose.yml`, `docker-compose.prod.yml:363`), so nothing there changes either.

Runtime: a worker with `REDIS_URL` set now holds a second Redis client, the binary tile-cache one. It is not the worker's first: `bootstrap()` step 9 has always given it a `RedisCacheProvider` for the JSON cache under the same setting. Neither provider exposes a close method and the API lifespan does not close its own either, so worker shutdown is symmetric with what the API already does.

## Verification

- `cd backend && uv run ruff check .` and `uv run ruff format --check .`: pass
- `uv run pytest tests/test_worker_tile_cache_1315.py -v`: 13 passed, with the live-Valkey tests confirmed running rather than skipped
- the same file on the unfixed tree: 8 failed, 5 passed
- `uv run pytest -n 4 -q` (whole backend suite): result in a comment below
- `uv run python scripts/check_env_doc_drift.py`: aligned

## For review

`tasks_common.py` is deliberately untouched, because #1359 is in flight there.

`test_event_ingest.py` needed one line. Its `_fake_settings()` is a `SimpleNamespace` standing in for `Settings`, and `TestWorkerDeliveryProof` stubs out `init_cache` but now reaches the new `init_tile_cache`, which reads `settings.redis_url`. The whole-tree run caught it; the fake gains `redis_url=None`, which takes the worker branch. Nothing about the assertion changes. That stub's docstring already tracks which bootstrap fields it has to carry, so this is the same maintenance the S3 gate needed.

Multi-tenancy is fine. All three call sites sit inside `@tenant_task`, so `current_tenant_var` is set and `_invalidation_table_segment` resolves the tenant-scoped key prefix in the worker exactly as it does in the API. No cross-tenant eviction is introduced.

`backend/app/api/main.py` shrinks by 2 lines, so its exact `_MODULE_LOC_CAPS` entry drops from 1332 to 1330.

Closes #1315

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -648,6 +648,11 @@ FRONTEND_PORT=8080
 # --- Cache Provider ---
 # [OPTIONAL] Redis/Valkey connection URL for cross-instance caching
 # Leave unset to use in-memory caching (single-instance default).
+# Note (#1315): "single-instance" still means two processes. With this unset,
+# the API's MVT tile cache is process-local, so the worker cannot purge it
+# after a re-upload or PostGIS refresh — the API keeps serving pre-swap tiles
+# until they expire (TILE_CACHE_TTL). The worker says so at boot. Set this if
+# you re-upload datasets and cannot wait out the TTL.
 # Type: string | No default
 # Example: redis://redis:6379/0
 # REDIS_URL=
--- a/backend/app/api/main.py
+++ b/backend/app/api/main.py
@@ -28,7 +28,6 @@ from app.observability.metrics import (
     shutdown_worker_metrics,
     sweep_dead_worker_metrics,
 )
-from app.platform.cache.provider import init_tile_cache
 
 # settings already imported above for the tempdir override — do NOT reimport
 # fix(#909): async_session/engine are late-bound inside each function that
@@ -325,7 +324,6 @@ async def lifespan(app: FastAPI):
     # the worker's age-aware sweeper instead.
     sweep_orphaned_exports(exports_dir)
 
-    init_tile_cache()
     await init_tile_pool()
     await task_app.open_async()
 
--- a/backend/app/platform/cache/provider.py
+++ b/backend/app/platform/cache/provider.py
@@ -2,11 +2,15 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING, Any, Protocol
 
+import structlog
+
 if TYPE_CHECKING:
     from app.platform.cache.tile_cache import (
         InMemoryTileCacheProvider,
         TileCacheProvider,
     )
+
+logger = structlog.stdlib.get_logger(__name__)
 
 
 class CacheProvider(Protocol):
@@ -63,13 +67,21 @@ def get_cache() -> CacheProvider:
 _tile_cache: "TileCacheProvider | InMemoryTileCacheProvider | None" = None
 
 
-def init_tile_cache() -> None:
+def init_tile_cache(*, in_memory_fallback: bool = True) -> None:
     """Initialize the tile cache singleton.
 
     Uses the Redis-backed binary provider when ``REDIS_URL`` is set;
     otherwise falls back to an in-memory LRU provider (PERF-01,
     Phase 274) so smaller single-VPS deployments still get tile-cache
     benefits without running Redis.
+
+    fix(#1315): pass ``in_memory_fallback=False`` from a process that only
+    ever INVALIDATES tiles and never reads them — the Procrastinate worker.
+    A process-local LRU there holds nothing, because nothing in that process
+    ever caches a tile, so the post-swap purge would evict zero entries while
+    logging ``tile_cache_invalidated`` — a no-op wearing a success message.
+    Leaving the singleton unset instead makes the gap legible: the caller sees
+    ``None`` and the warning below says what is lost and how to fix it.
     """
     global _tile_cache
     from app.core.config import settings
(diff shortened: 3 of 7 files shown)


ishiland:
@codex review

ishiland:
Whole-tree gate, as promised in the body.

The local `uv run pytest -n 4 -q` run over the whole backend suite is what caught the only knock-on this change had: **8170 passed, 2 failed**, both in `test_event_ingest.py::TestWorkerDeliveryProof`. That class stubs `Settings` with a `SimpleNamespace` and patches `init_cache` away, so it sailed past bootstrap step 9 but hit the new step 9b, which reads `settings.redis_url`:

```
E  AttributeError: 'types.SimpleNamespace' object has no attribute 'redis_url'
app/platform/cache/provider.py:89: AttributeError
```

The fake gained `redis_url=None`, which takes the worker branch, and `tests/test_event_ingest.py` now passes 17 with 1 skipped. A targeted subset would have shipped that break. Worth recording, because it is the same shape as the bug this PR fixes: a second process reaching a bootstrap step nobody had told it about.

I stopped the local re-run partway rather than finish it. It had crawled to 57% in about 23 minutes against the shared dev Postgres at localhost:5434, which several parallel worktrees were using at the time, and it was slowing them down for a gate CI runs against an isolated database anyway. **Backend Tests on this PR is green**, and that is the authoritative whole-tree result: 28 checks pass, 10 skip, none fail.

Only `tests/test_event_ingest.py` changed between the failing local run and the pushed commit, and it is green both locally and in CI.

PR #1378, opened 2026-08-10
fix(raster): serialize crs_wkt as WKT2_2019 at both ingest paths and backfill

### LINKED ISSUE #1376: raster_assets.crs_wkt is WKT1_GDAL, but STAC's proj:wkt2 implies WKT2
Follow-up from #1334's review (codex, PR #1374).

\`rasterio.crs.CRS.to_wkt()\`'s default serialization is WKT1_GDAL (\`PROJCS[...]\`), not WKT2 (\`PROJCRS[...]\`) — confirmed live:

\`\`\`
default:   PROJCS["WGS 84 / UTM zone 21N",GEOGCS["WGS 84",DATUM["WGS_19...
WKT2_2019: PROJCRS["WGS 84 / UTM zone 21N",BASEGEOGCRS["WGS 84",ENSEMBL...
\`\`\`

\`RasterAsset.to_stac_properties()\` publishes \`raster_assets.crs_wkt\` verbatim as the STAC Projection Extension's \`proj:wkt2\` field, which by name and spec convention implies WKT2 content. A strict STAC consumer parsing that field as WKT2 could reject or misinterpret a WKT1 string.

This is NOT specific to remote/STAC-imported rasters. \`backend/app/processing/raster/cog.py:287\` (the local-upload ingest path) has called \`crs.to_wkt()\` with no explicit version for as long as \`crs_wkt\` has existed — every raster GeoLens has ever ingested, local or remote, stores WKT1 in this column.

Fixing only the remote-asset probe path (\`cog_info.py\`'s \`_georeferencing\`, added in #1334) to request \`version="WKT2_2019"\` while leaving the local path unconverted would split \`raster_assets.crs_wkt\` into two dialects depending on \`storage_backend\` — worse than the current uniform-but-technically-imprecise state, since a STAC consumer could no longer assume one format for the column.

Fix direction: convert to WKT2 (\`to_wkt(version="WKT2_2019")\` or similar) at BOTH call sites together, in one coordinated change, so the column means one consistent thing regardless of how the raster was ingested. Needs a decision on whether existing stored rows get backfilled or drift until re-ingested/refreshed.

Low severity — most STAC/GIS tooling accepts WKT1_GDAL leniently even under a \`wkt2\`-named field, but strict consumers may not.

ishiland:
## What

`RasterAsset.to_stac_properties()` publishes `raster_assets.crs_wkt` verbatim as the STAC Projection Extension's `proj:wkt2`, but both writers of the column asked rasterio for its default serialization — which is WKT1_GDAL (`PROJCS[...]`), not WKT2 (`PROJCRS[...]`). A strict consumer parsing a `wkt2`-named field as WKT2 can reject or misinterpret it.

Both call sites change together, because fixing either alone splits the column into two dialects with nothing on a row to say which one it holds:

- `backend/app/processing/raster/cog.py` — the local-upload path, which has written WKT1 for as long as `crs_wkt` has existed (every raster GeoLens has ever ingested locally).
- `backend/app/modules/catalog/sources/cog_info.py` — the remote-asset probe added in #1374.

## Backfill

Migration `0041_raster_assets_crs_wkt2` converts the rows already stored, so the dialect is not a function of ingest era either. Re-ingest is the only other way a row would ever change, and a stable raster has no reason to be re-ingested.

- Plain transactional statements only — no `CREATE INDEX CONCURRENTLY`, no autocommit block.
- Predicate matches the full set of WKT1 root keywords (`PROJCS`/`GEOGCS`/`GEOCCS`/`LOCAL_CS`/`COMPD_CS`/`VERT_CS`/`FITTED_CS`), anchored with `\[` so `GEOGCS` cannot also match a `GEOGCRS[` prefix.
- Idempotent by construction: every rewritten row leaves the predicate, since WKT2 roots are `PROJCRS`/`GEOGCRS`/`GEODCRS`/`ENGCRS`/`COMPOUNDCRS`/`VERTCRS`.
- Keyset-paged by `id`, which also steps past rows it deliberately leaves alone.
- A row PROJ cannot parse is left exactly as stored and logged as a warning; an unreadable CRS is not a reason to fail a deployment.
- Downgrade is a documented no-op: WKT1 is the less expressive dialect, so the original bytes are not recoverable, and both readers of the column (`core/geo.py`'s `wkt_is_geographic` / `wkt_has_degree_unit`) already handle either dialect.

Verified on the test database: the migration converted 64 pre-existing rows.

## Tests

- `tests/test_raster_ingest.py` — the local path stores a WKT2 root keyword, parametrized over a geographic and a projected CRS (`GEOGCRS[` / `PROJCRS[`), since WKT1 and WKT2 differ in the root keyword for each class.
- `tests/test_cog_info.py` — the same property on the remote probe.
- `tests/test_crs_wkt2_backfill_1376.py` (new) — a real `alembic downgrade -1` / `upgrade head` round trip over three committed row shapes in one migration run: a WKT1 row converts and keeps its EPSG identity, a structurally broken WKT1-shaped row survives untouched (and does not stop the loop before the others), and an already-WKT2 row comes out byte-identical.

Worth recording from writing that last one: PROJ is far more tolerant than it looks. An invented `LOCAL_CS["...",NONSENSE[1,2,3]]` parses cleanly and converts to `ENGCRS[...]`; only a structurally broken string (truncated mid-record) reaches the skip branch.

## Verification

```
  uv run pytest tests/test_cog_info.py tests/test_raster_ingest.py \
    tests/test_crs_wkt2_backfill_1376.py -q                      # 35 passed
    tests/test_rule2_structural.py -q                            # 122 passed
  uv run pytest tests/test_stac_import.py tests/test_stac_refresh_1266.py \
    tests/test_raster_tiles.py tests/test_wkt_is_geographic.py \
    tests/test_crs_degree_agreement.py tests/test_vrt_rewrite.py \
    tests/test_regenerate_vrt_integration.py tests/test_raster_schema.py -q   # 256 passed
  uv run pytest tests/test_email_verification_migration.py tests/test_alembic_helpers.py \
    tests/test_origin_ref_indexes.py tests/test_dataset_source_state.py \
    tests/test_raster_replace_1221.py tests/test_cog_band_unit_capture.py \
    tests/test_raster_antimeridian_887.py -q                     # 376 passed
  PYTHONPATH=. uv run alembic upgrade head                       # applied 0041
  PYTHONPATH=. uv run alembic check                              # no new upgrade operations
  uv run ruff check . && uv run ruff format --check .            # clean
```

No schema change, so no `openapi.json` / SDK regeneration.

Fixes #1376

### DIFF
--- a/backend/alembic/versions/0041_raster_assets_crs_wkt2.py
+++ b/backend/alembic/versions/0041_raster_assets_crs_wkt2.py
@@ -1,0 +1,130 @@
+"""Convert stored raster_assets.crs_wkt from WKT1_GDAL to WKT2:2019.
+
+fix(#1376): ``RasterAsset.to_stac_properties()`` publishes ``crs_wkt``
+verbatim as the STAC Projection Extension's ``proj:wkt2``, but both writers
+of the column asked rasterio for its DEFAULT serialization, which is
+WKT1_GDAL (``PROJCS[...]``, not ``PROJCRS[...]``). Both now request
+``version="WKT2_2019"`` — ``processing/raster/cog.py`` for local uploads and
+``modules/catalog/sources/cog_info.py`` for the remote-asset probe.
+
+Why the rows get backfilled rather than left to drift. Fixing only the
+writers would make the dialect a function of INGEST ERA: a consumer reading
+``proj:wkt2`` off this catalog could no longer assume one format for the
+field, and nothing in a row says which era wrote it. Re-ingest is the only
+other way a row would ever change, and a stable raster has no reason to be
+re-ingested. So the conversion happens once, here, and the column means one
+thing afterwards.
+
+Idempotent by construction: the predicate matches WKT1 root keywords only,
+and every row this migration rewrites leaves that predicate (WKT2 roots are
+``PROJCRS``/``GEOGCRS``/``GEODCRS``/``ENGCRS``/``COMPOUNDCRS``/``VERTCRS``).
+A row whose stored WKT PROJ cannot parse is left exactly as it was — an
+unreadable CRS is not a reason to fail a deployment, and the value is no
+worse off than before.
+
+Downgrade is a deliberate no-op. WKT1 is the strictly less expressive
+dialect, so the pre-migration bytes are not faithfully recoverable, and
+nothing reads this column expecting WKT1: the two consumers
+(``core/geo.py``'s ``wkt_is_geographic``/``wkt_has_degree_unit``) hand it to
+PROJ and their keyword fallback already recognises both dialects.
+
+Revision ID: 0041_raster_assets_crs_wkt2
+Revises: 0040_dataset_origin_ref_indexes
+Create Date: 2026-08-10
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "0041_raster_assets_crs_wkt2"
+down_revision: Union[str, None] = "0040_dataset_origin_ref_indexes"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+# Every WKT1 root keyword, not just the projected/geographic pair: each has a
+# DISTINCT WKT2 counterpart, so matching the whole set converts the column
+# completely instead of leaving a second dialect behind on the rarer CRS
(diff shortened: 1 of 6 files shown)


ishiland:
@codex review

ishiland:
@codex review

Issue #1314, opened 2026-08-09
fix(catalog): reconcile auto-generated distributions when a dataset's modality changes


ishiland:
Split out of #1313 (review round 7), which is where the gap became visible rather than where it was introduced.

`record_distributions` rows are generated once, at dataset creation, by `generate_distributions()` in `backend/app/modules/catalog/records/service.py`. Nothing re-derives them afterwards, so a dataset whose spatiality changes keeps the distribution set it was born with:

- a registered table that gains geometry keeps only its CSV download and OGC Features rows, and never advertises vector tiles, GeoPackage, GeoJSON or GeoParquet;
- a dataset whose geometry columns are dropped keeps advertising all of those against a relation that can no longer serve them.

This predates #1313. That PR narrows it — `build_assets()` computes from `record_type` live and now gets the right answer, because the registered-PostGIS refresh re-derives `record_type` from the measurement — but the persisted rows are exactly as stale as they were before.

## Why it is not a one-liner

`generate_distributions()` has merge semantics: it inserts rows that are missing and never removes any. So the promote direction is nearly free (call it again with the new geometry type), and the demote direction is the real work — it needs a reconcile that removes auto-generated rows the new modality excludes while preserving:

- rows with `auto_generated=False`, which a user authored;
- user edits to auto-generated rows (title, media type, `is_primary`), if those are meant to survive — the current model has no per-field provenance for distributions the way `attribute_metadata` has `user_modified_fields`, so this needs a decision, not just code.

It also needs a `ProcessingPort` seam. `generate_distributions` lives in `modules/catalog/`, and `backend/app/processing/` reaches catalog only through the port (`test_no_processing_imports_catalog`), so the refresh task cannot call it directly.

## Scope

- A `reconcile_distributions()` beside `generate_distributions()` that brings a record's auto-generated rows in line with a given geometry type, with an explicit and tested preservation policy for user-authored and user-edited rows.
- A `ProcessingPort` method exposing it.
- Call it from the registered-PostGIS refresh (`backend/app/processing/ingest/tasks_postgis_refresh.py`) in the write transaction, when the measured modality differs from the stored one.
- Consider whether the re-upload paths want it too: a file re-upload that replaces a spatial dataset with a non-spatial one has the same problem, and `_apply_reupload_swap` does not touch distributions either.

## Acceptance

- [ ] A dataset that gains geometry advertises the full spatial distribution set after a refresh
- [ ] A dataset that loses geometry no longer advertises spatial distributions
[...]

PR #1381, opened 2026-08-10
fix(catalog): stop user-authored distributions suppressing built-in export rows

### LINKED ISSUE #1370: fix(catalog): a user-authored distribution permanently suppresses the built-in export row for its pair
Found during review of #1369 (#1314).

`generate_distributions()` in `backend/app/modules/catalog/records/service.py` probes for existing rows by `(distribution_type, format)` and skips any pair already present. The probe does not filter on `auto_generated`, so a row a user authored through `POST /records/{id}/distributions/` counts as "already there".

Consequence: once someone adds their own `download`/`gpkg` entry pointing at their own file, the built-in `/datasets/{id}/export?format=gpkg` row can never be generated for that record — not by a later call, and not by the `reconcile_distributions()` promote that #1369 added. The dataset advertises the user's link and silently never advertises the platform's own export endpoint for that format. The endpoint still works; it is just absent from the catalog record, DCAT output and STAC assets.

## Why #1369 did not change it

The obvious fix — filter the probe to `auto_generated=True` so a user row no longer blocks the generated one — is not safe as a drive-by:

- The unique constraint is `(record_id, distribution_type, format, url)`. A user row whose URL happens to equal the template's (`/datasets/{id}/export?format=gpkg` is a perfectly guessable thing to type) would make the insert raise. On the create path that is a failed dataset creation; on the reconcile path it is an `IntegrityError` inside the write transaction of a registered-PostGIS refresh or a reupload swap, turning a metadata correction into a failed job.
- So closing this needs a conflict-tolerant insert (`ON CONFLICT DO NOTHING` against the four-column constraint) rather than a changed probe, plus a decision about what two rows advertising the same format should look like to DCAT and STAC consumers.

#1369 mitigated the sharpest edge instead: the `is_primary` normalization picks the primary from the rows that actually exist, so a record in this state keeps a primary distribution (CSV) rather than ending up with none.

## Acceptance

- [ ] A record carrying a user-authored `download`/`gpkg` row still advertises the built-in GeoPackage export link
- [ ] A user row whose URL collides with the template's does not raise on create, on `generate_distributions`, or on `reconcile_distributions`
- [ ] The DCAT and STAC representations of a record with two rows for one format are checked, not assumed

ishiland:
## The bug

`generate_distributions()` probed for existing rows by `(distribution_type, format)` with no filter on `auto_generated`, so a row a user added through `POST /records/{id}/distributions/` counted as "the pair is taken". Once somebody added their own `download`/`gpkg` entry, the built-in `/datasets/{id}/export?format=gpkg` row could never be generated for that record again — not by a later call, and not by the `reconcile_distributions()` promote #1369 added. The export endpoint kept working; the catalog record, the DCAT feeds and the STAC assets simply stopped naming it.

## The change

1. The existence probe reads only `auto_generated=True` rows, so a user row no longer withholds the platform's own export link for that format. Two rows advertising one format is the intended end state: one the user's, one the platform's.
2. The generated-row insert is `INSERT ... ON CONFLICT DO NOTHING` against `uq_record_distribution`, via `sqlalchemy.dialects.postgresql.insert(...).on_conflict_do_nothing(...)`. Rows skipped by a conflict are absent from `RETURNING`, so the returned `created` list stays a truthful record of what was inserted — which is what reconcile's `is_primary` normalization picks from.

All three paths that insert generated rows go through that one statement: dataset create (`service_create.py` calls `generate_distributions`), the direct call, and the reconcile promote.

### Why the probe change alone is unsafe

The unique constraint is four columns — `(record_id, distribution_type, format, url)` — verified in `RecordDistribution.__table_args__` and in `0001_baseline.py`, and pinned by a new test. With user rows dropped from the probe, a user row whose url happens to equal the template's (`/datasets/{id}/export?format=gpkg` is a guessable thing to type) stops being a skipped pair and becomes a live constraint violation. Catching `IntegrityError` and passing would be a worse mechanism than it looks: the reconcile caller runs inside the write transaction of a registered-PostGIS refresh or a reupload swap, so a raised violation aborts every write that job has already made, turning a metadata correction into a failed job. Resolving the conflict inside the statement never opens that hole.

This is measured, not asserted. With the probe fixed and `on_conflict_do_nothing` removed, three collision tests plus the #1314 fallback test fail; with `on_conflict_do_nothing` in place and the probe unfixed, five suppression tests fail.

### `is_primary` re-checked against the two-rows state

Normalization spans the generated rows, so exactly one auto-generated row is primary in the new state, and a promote beside a user row now promotes the real GeoPackage row instead of falling back to CSV. #1314's fallback is still reachable and still tested — the url-collision case leaves no generated GeoPackage row to promote — so the test that covered it was retargeted rather than deleted.

One thing the change does not alter: a user who flags their **own** row `is_primary=true` keeps that flag, and that record advertises two primaries. That is reachable today with a single POST on any dataset (nothing in `create_distribution` demotes the incumbent), so it predates this PR and fixing it here would widen scope mid-review. Filed as #1383 and referenced from `reconcile_distributions`' docstring.

The fallback is also where the new insert mechanism gets exercised hardest, so it has its own test. It is the only path that flips `is_primary` on a row `generate_distributions` just inserted — everywhere else the normalization writes to a survivor, or accepts the flag the template already carried. Those rows now arrive from `INSERT ... RETURNING` rather than `session.add`, and one that was not session-persistent would take the assignment and never emit the UPDATE.

## DCAT / STAC verification (checked, not assumed)

- **DCAT 3, DCAT-US 3.0, GeoDCAT-AP** each serialize `record.distributions` as a plain list of Distribution nodes (`dcat/service.py:346`, `dcat_us/service.py:107`, `geodcat_ap/service.py:170`). Nothing is keyed, and the two rows differ by `accessURL`, so both appear. Covered by a test asserting both gpkg URLs are in the DCAT output.
- **STAC** was the one worth checking, since assets are a keyed object. They are not built from `record_distributions`: `build_assets()` (`search/service_records.py:57`) derives `download_<fmt>` from the dataset id and the modality, and `ogc_record_to_stac_item()` copies that dict through and never reads `properties.distributions`. So there is no asset key for two rows to collide on, no last-writer-wins overwrite, and no suffixing needed. Both rows do survive in the OGC Record `properties.distributions`, which is the representation that carries them. Covered by a test that walks `dataset_to_ogc_record` → `ogc_record_to_stac_item`.

## Verification

```
uv run pytest tests/test_distribution_reconcile_1314.py -v                  # 20 passed
uv run pytest tests/test_dataset_source_state.py tests/test_dcat.py \
              tests/test_geodcat_ap.py -q                                   # 182 passed (with the above at 19)
uv run pytest tests/test_metadata_roundtrip.py tests/test_ogc_collection_metadata.py \
              tests/test_ogc_record_enrichment.py tests/test_postgis_refresh_1265.py \
              tests/test_processing_port.py tests/test_records_authz_sec001.py \
              tests/test_record_subresource_edits.py tests/test_records_related.py \
              tests/test_stac_record_output.py -q                           # 168 passed
uv run pytest tests/test_ingest.py tests/test_raster_ingest.py \
              tests/test_raster_replace_1221.py tests/test_regenerate_vrt_integration.py \
              tests/test_vrt_catalog_175.py tests/test_quicklook_predicate.py \
              tests/test_raster_dataset_assets_bug041.py -q                 # 214 passed
uv run pytest tests/test_layering.py -q                                     # 40 passed
uv run ruff check . && uv run ruff format --check .                         # clean
```

No migration: the constraint the insert targets already exists (`0001_baseline.py`). No schema, API or config impact — `record_distributions` gains rows on records that were missing them, and every response shape is unchanged.

Fixes #1370

### DIFF
--- a/backend/app/modules/catalog/records/service.py
+++ b/backend/app/modules/catalog/records/service.py
@@ -720,11 +720,9 @@ async def reconcile_distributions(
 
     Normalization spans the generated rows only, which is the same boundary as
     every other bullet here. A user who flags their OWN row primary keeps that
-    flag through a reconcile; that record advertises two primaries, and it
-    could already do so before #1370 put a generated row beside a user row for
-    one format — a single ``POST /records/{id}/distributions/`` with
-    ``is_primary=true`` reaches it, and nothing in ``create_distribution``
-    demotes the incumbent.
+    flag through a reconcile, so that record advertises two primaries — see
+    #1383, which is reachable from a single POST on any dataset and predates
+    #1370 rather than being introduced by it.
 
     The lost-edit case is narrower than it reads: ``update_distribution`` and
     ``delete_distribution`` both refuse to touch a row with
--- a/backend/app/modules/catalog/records/service.py
+++ b/backend/app/modules/catalog/records/service.py
@@ -496,6 +496,13 @@ _DISTRIBUTION_TEMPLATES = [
 # generated set, so reconcile has to see it.
 _VECTOR_TILES_PAIR = ("vector_tiles", "pbf")
 
+# The four-column unique constraint on ``record_distributions``
+# (record_id, distribution_type, format, url) — see RecordDistribution's
+# ``__table_args__``. Named here because the generated-row insert has to be
+# conflict-tolerant against it; a rename that missed this constant fails loudly
+# on the next insert ("constraint ... does not exist") rather than silently.
+_DISTRIBUTION_UNIQUE_CONSTRAINT = "uq_record_distribution"
+
 # Every (distribution_type, format) pair this module owns. A row outside this
 # set was written by something else — the raster and VRT ingest tails add their
 # own ``download`` rows — and reconcile must leave those alone even when they
@@ -520,7 +527,7 @@ def _pair_applies(dist_type: str, fmt: str, geometry_type: str | None) -> bool:
 # the template table marks primary; CSV is the fallback, both for a modality
 # that generates no GeoPackage row at all and — on reconcile — for a promote
 # where no generated GeoPackage row exists to promote because a user's own row
-# already occupies that pair.
+# already sits at the exact url the GeoPackage template would have inserted.
 _PRIMARY_PREFERENCE: tuple[tuple[str, str], ...] = (
     ("download", "gpkg"),
     ("download", "csv"),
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

Issue #1334, opened 2026-08-10
Remote raster asset rows never populate crs_wkt/res_x/res_y


ishiland:
Observed during the #1326 review loop (Milestone 5 close-out ledger). Raster asset rows created for remote references (STAC imports and refresh-adopted assets) leave `crs_wkt`, `res_x`, and `res_y` unset — the UI renders Resolution as "— x —" for these datasets (visible on any Sentinel-2 STAC import's Raster Properties card).

The COG info probe (`fetch_cog_info` in `backend/app/modules/catalog/sources/cog_info.py`) already retrieves enough metadata to fill these at import/refresh time; the values are simply never projected onto the asset row.

Low severity (display-only), but it makes remote rasters look less complete than uploaded ones.

PR #1384, opened 2026-08-10
feat(stac): give remote raster rows a measured resolution and rotation flag

### LINKED ISSUE #1375: Remote raster res_x/res_y still unset — Titiler's /cog/info exposes no affine transform
Follow-up from #1334's review (codex, PR #1374).

#1334 asked for \`crs_wkt\`/\`res_x\`/\`res_y\` to be projected onto remote raster asset rows from \`fetch_cog_info\`. \`crs_wkt\` shipped; \`res_x\`/\`res_y\` did not, because Titiler's \`/cog/info\` response carries no affine transform — only an axis-aligned \`bounds\` envelope plus pixel \`width\`/\`height\`.

Dividing that envelope by pixel dimensions is only the true per-pixel resolution for an axis-aligned (non-rotated, non-sheared) raster. For a rotated or sheared COG, the bounding envelope is larger than the raster's own footprint, so the computed value would silently overstate resolution — indistinguishable, from this payload alone, from a correct one. The local-upload path avoids this because it reads the real transform (\`rasterio\`'s \`transform.a\`/\`transform.e\`, plus \`transform.b\`/\`transform.d\` to detect and flag \`is_rotated\`); nothing in Titiler's REST response for a remote asset exposes that.

So remote raster rows (STAC imports and refresh-adopted assets) still show "Resolution: — x —" in the UI's Raster Properties card, and remote datasets still have no \`gsd\` in their STAC export.

Possible directions, none attempted yet:
- A Titiler/rio-tiler version or endpoint that exposes the geotransform or a rotation flag directly.
- A narrowly-scoped, Gate-2-respecting way to ask Titiler itself to report whether the source is axis-aligned (rather than fetching the transform into GeoLens's own process, which AGENTS.md Rule 2 rules out for a caller-controlled remote URL).
- Accepting the axis-aligned assumption as a documented approximation, IF this codebase decides that's an acceptable trade-off for remote assets (it currently does not — see \`cog_info.py\`'s \`_georeferencing\` docstring).

Low severity — display/metadata-only, same as #1334 itself.

ishiland:
## What

#1334 left `res_x`/`res_y` unset on remote raster assets, because Titiler's `/cog/info` carries no affine transform and dividing its bounding envelope by pixel dimensions is only correct for an axis-aligned raster. Remote rows therefore showed "Resolution: — x —" in the Raster Properties card and exported no `gsd`.

`/cog/stac` does publish the transform. rio-stac writes the projection extension's `proj:transform`, and its first six elements are the same six numbers `processing/raster/cog.py` reads off rasterio's `src.transform` on the local-upload path:

| affine elements | read as | how |
|---|---|---|
| 0, 3 (`a`, `d`) | `res_x` | `hypot(a, d)` — the column-step pixel vector's length |
| 1, 4 (`b`, `e`) | `res_y` | `hypot(b, e)` — the row-step pixel vector's length |
| 1, 3 (`b`, `d`) | `is_rotated` | nonzero shear in either term |

So this is parity with the local-upload path on identical inputs, not an approximation of it. Both paths now run those six numbers through one helper, `pixel_size_from_affine` in `core/geo.py`.

### The resolution formula changed under review

The first version of this PR read `res_x`/`res_y` straight off elements 0 and 4, matching what `raster/cog.py` had always done. Codex review caught that those are the pixel vectors' **components on the world axes**, which equal the pixel sizes only when the raster is axis-aligned — and that matching the local path's existing calculation merely reproduced its error on a newly supported path.

It is right. For the 30°-rotated 10 m fixture below the affine reads `a=8.66, d=5.0`, so the old form stored 8.66 m for a pixel that is 10 m across: a 13% understatement reaching the UI and STAC's `gsd`. Verified against the measured distance between adjacent pixel centres, which is 10.0.

**Both** paths are corrected here, not just the remote one this PR adds. Fixing only the remote half would leave the same scene reporting two different resolutions depending on how it was ingested, which is the exact contradiction this PR exists to remove. The change is inert for axis-aligned rasters — with `b` and `d` zero the vector length collapses to exactly `abs(a)`/`abs(e)` — so no existing raster's stored resolution moves, and rotated rasters are already barred from VRT sources by `_check_rotation` (VAL-07), so the new values cannot perturb VRT grid maths.

## Measured, against the pinned image

titiler 2.2.1 (`ghcr.io/developmentseed/titiler:2.2.1`, the tag `docker-compose.yml` pins) → rio-tiler 9.4.2, rio-cogeo 7.0.2, rasterio 1.5.0. A synthetic COG with 10 m pixels and its 30°-rotated twin:

```
axis-aligned  /cog/stac  proj:transform = [10.0,     0.0,  500000.0,  0.0,    -10.0,     8000000.0, 0, 0, 1]
rotated 30°   /cog/stac  proj:transform = [8.660254, -5.0, 500000.0,  5.0,    -8.660254, 8000000.0, 0, 0, 1]
```

`/cog/info` for that same rotated file returns `bounds` spanning 498720–502217 — a 3497 m envelope around a 2560 m footprint. Envelope-over-pixels would report ~13.7 m where the affine says 10 m, a 37% overstatement with nothing in that payload to distinguish it from a correct value. That is the number #1334 declined to publish, and it is why the resolution comes from a second endpoint rather than a smarter reading of the first.

`/cog/validate` (rio-cogeo) reports `GEO.Resolution`, and it is the same `(transform.a, transform.e)` pair — but it publishes no shear terms, so it cannot answer the rotation question, and it does a full COG-structure validation to report less. Not used.

## Cost

`/cog/stac` defaults to `with_raster=true&with_eo=true`, which makes it a **pixel** read — rio-stac downsamples to `max_size` to compute per-band statistics. The call sets both false, leaving a header-only read that still carries the whole projection block. Measured on a 2048x2048 3-band COG against the pinned image:

| request | latency | extensions returned |
|---|---|---|
| `/cog/stac` defaults | 90 ms | proj + raster + eo |
| `/cog/stac?with_raster=false&with_eo=false` | 7 ms | proj |

`fetch_cog_info` already makes its own `/cog/statistics` request, so the statistics half was redundant regardless.

## Availability

Not incidental: `titiler.application` 2.2.1 declares a hard dependency on `titiler-extensions[cogeo,stac]==2.2.1`, so `/cog/stac` ships in every deployment of the pinned tag rather than only where someone opted into an extra. No version bump, no compose change.

## Security

SSRF stance unchanged. This is a third request to the same internal Titiler service, for the same URL the function's caller already validated, from the same function, inside both gates of the SEC-OBSV-02 contract — Gate 1 caller-side `validate_url_for_ssrf`, Gate 2 Titiler's `CPL_VSIL_CURL_ALLOWED_EXTENSIONS` clamp. It reaches no origin `fetch_cog_info` did not already reach, and GDAL still never sees a caller-controlled URL inside GeoLens's own process, which is what AGENTS.md Rule 2 forbids.

## A correctness fix, not only display polish

`raster_assets.is_rotated` is `NOT NULL DEFAULT false`. Every remote row today therefore asserts axis-alignment with nothing behind it — a rotated remote COG is silently recorded as unrotated, and `processing/raster/validation.py` reads that flag. The transform closes that gap in the same change.

## Degradation

The georeferencing keys are **absent**, not `None`, when the transform cannot be read: the two callers write them straight onto the row, so an absent key leaves the column as it was where a `None` would assert "measured, and there is no value". A failing or missing `/cog/stac` leaves the rest of the probe intact, the same contract the optional `/cog/statistics` call already follows.

## Tests

- `tests/test_cog_info.py` — new `TestGeotransform`: resolution comes off the affine and not the envelope (asserted on the rotated fixture, the case the shortcut gets wrong), the rotated resolution is the pixel vector's length and explicitly *not* element 0, rotation set and cleared, magnitudes positive for a north-up transform, five malformed-transform shapes that must leave the keys absent, and a failing endpoint that must not fail the probe.
- `tests/test_raster_ingest.py` — the same vector-length property on the local path, against a real 30°-rotated GeoTIFF, plus an axis-aligned case pinning that the review fix is inert where the old form was already right.
- `tests/test_stac_import.py` — the import records the probed resolution and rotation onto the row and the API; its sibling keeps covering the no-transform case where they stay NULL.
- `tests/test_stac_refresh_1266.py` — the refresh repoint carries all three onto the moved object, beside the fields it already moved.

## Verification

```
  uv run pytest tests/test_cog_info.py tests/test_raster_ingest.py \
    tests/test_stac_import.py tests/test_stac_refresh_1266.py -q             # 182 passed
    tests/test_rule2_structural.py tests/test_wkt_is_geographic.py \
    tests/test_crs_degree_agreement.py tests/test_raster_tiles.py \
    tests/test_raster_validation.py -q                                       # 255 passed
  uv run pytest tests/test_raster_antimeridian_887.py tests/test_vrt_rewrite.py \
    tests/test_regenerate_vrt_integration.py tests/test_raster_replace_1221.py \
    tests/test_cog_band_unit_capture.py tests/test_crs_wkt2_backfill_1376.py \
    tests/test_vrt_source_management_174.py -q                               # 308 passed
  uv run ruff check . && uv run ruff format --check .            # clean
```

No migration, no schema change, no `openapi.json` / SDK regeneration — `res_x`/`res_y` were already in the API response and were simply always null for remote rows.

Closes #1375

### DIFF
--- a/backend/app/core/geo.py
+++ b/backend/app/core/geo.py
@@ -707,6 +707,52 @@ def wkt_has_degree_unit(crs_wkt: str | None) -> bool | None:
     return crs_has_degree_unit(_parse_crs(crs_wkt))
 
 
+def crs_metres_per_unit(crs: object | None) -> float | None:
+    """Metres per linear unit of a PROJECTED CRS, or None when that is not a
+    question with an answer.
+
+    fix(#1375 review): STAC's ``gsd`` is defined in metres, but a stored
+    resolution is in whatever unit its CRS measures. PROJ answers the
+    conversion for projected CRSs directly — ``units_factor`` reports
+    ``('metre', 1.0)`` for UTM and Web Mercator and ``('US survey foot',
+    0.3048006...)`` for the state-plane systems, so a foot-based raster
+    converts as readily as a metre-based one.
+
+    Returns None for a GEOGRAPHIC CRS on purpose, rather than a factor.
+    ``units_factor`` reports RADIANS per unit there, and an angular
+    resolution has no fixed length: a degree of longitude is 111 km at the
+    equator and nothing at the pole, so the conversion needs a latitude this
+    function is not given. Callers must read None as "cannot be expressed in
+    metres" and omit the value rather than publish it unconverted — see
+    ``RasterAsset.to_stac_properties``.
+    """
+    if crs is None:
+        return None
+    try:
+        if not crs.is_projected:
+            return None
+        _, metres_per_unit = crs.units_factor
+    except Exception:  # broad: exotic CRSs raise from PROJ rather than answering, and "no answer" is exactly the None case
+        return None
+    if not metres_per_unit or metres_per_unit <= 0:
+        return None
+    return float(metres_per_unit)
+
+
+def wkt_metres_per_unit(crs_wkt: str | None) -> float | None:
+    """:func:`crs_metres_per_unit` plus the WKT parse.
+
+    Same shape as :func:`wkt_has_degree_unit` over
+    :func:`crs_has_degree_unit`: one implementation, two entry points, so a
+    caller holding a live ``rasterio.crs.CRS`` need not round-trip through
+    WKT to reach it. None covers every uncertainty — no WKT, an unparseable
+    one, or a CRS whose units PROJ will not report.
+    """
+    if not isinstance(crs_wkt, str) or not crs_wkt:
+        return None
+    return crs_metres_per_unit(_parse_crs(crs_wkt))
+
+
 def pixel_size_from_affine(
     a: float, b: float, d: float, e: float
 ) -> tuple[float, float]:
--- a/backend/app/core/geo.py
+++ b/backend/app/core/geo.py
@@ -705,3 +705,30 @@ def wkt_has_degree_unit(crs_wkt: str | None) -> bool | None:
     if not isinstance(crs_wkt, str) or not crs_wkt:
         return None
     return crs_has_degree_unit(_parse_crs(crs_wkt))
+
+
+def pixel_size_from_affine(
+    a: float, b: float, d: float, e: float
+) -> tuple[float, float]:
+    """Per-pixel ground distances along a raster's OWN axes, from its affine.
+
+    fix(#1375 review): not ``abs(a)``/``abs(e)``. Those are the pixel vectors'
+    COMPONENTS on the world axes, which equal the pixel sizes only when the
+    raster is axis-aligned. A geotransform maps pixel (col, row) to world
+    ``x = a*col + b*row + c``, ``y = d*col + e*row + f``, so one step along the
+    column axis moves the world point by the vector ``(a, d)`` and one step
+    along the row axis by ``(b, e)``. The LENGTHS of those two vectors are the
+    resolutions; ``a`` and ``e`` alone are their projections onto x and y.
+
+    Rotate a 10 m-pixel raster by 30° and the affine reads ``a=8.66, d=5.0``:
+    ``abs(a)`` reports 8.66 m for a pixel that is 10 m across, a 13%
+    understatement that reaches the UI and STAC's ``gsd``. Confirmed against
+    the distance between adjacent pixel centres, which is 10.0.
+
+    For an axis-aligned raster ``b`` and ``d`` are zero and this returns
+    exactly ``abs(a)``/``abs(e)`` — it agrees with the old form everywhere the
+    old form was right, and differs only where it was wrong. Both raster
+    ingest paths call this so a rotated scene reports the same resolution
+    whether it was uploaded or imported from a remote catalog.
+    """
+    return math.hypot(a, d), math.hypot(b, e)
--- a/backend/app/modules/catalog/sources/cog_info.py
+++ b/backend/app/modules/catalog/sources/cog_info.py
@@ -63,12 +63,9 @@ def _georeferencing(info: dict) -> dict:
     transform to check that against, so a rotated or sheared remote COG would
     silently get a resolution inflated by however far its bounding envelope
     exceeds its own footprint — indistinguishable, in this payload, from a
-    correct value. The local-upload path can tell the two apart
-    (``raster/cog.py`` reads ``transform.b``/``transform.d`` to set
-    ``is_rotated``); nothing here can. A wrong number that LOOKS like a
-    measurement is worse than the blank "—" it would replace, so it stays
-    unset until there is a source that can rule rotation out. Tracked as
-    #1375.
+    correct value. fix(#1375): they now come from ``_geotransform`` and a
+    SECOND Titiler endpoint that does carry the transform, which is why this
+    one still refuses to guess at them.
 
     Individual failures degrade to None rather than raising: this is
     descriptive metadata for a UI card, not something a failed probe should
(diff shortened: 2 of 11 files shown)


ishiland:
@codex review

ishiland:
Fixed in f35a7582d, and the fix is broader than the finding.

You are right on the maths — verified against ground truth, the measured distance between adjacent pixel centres on the 30° fixture is 10.0 while `abs(a)` reports 8.66. `res_x`/`res_y` now come from `hypot(a, d)` / `hypot(b, e)`.

Your last sentence is the load-bearing one, so I acted on it rather than only on the cited line: correcting the remote path alone would have left the same scene reporting two different resolutions depending on whether it was uploaded or imported, which is the contradiction this PR set out to remove. So both ingest paths now go through one helper, `pixel_size_from_affine` in `core/geo.py` — that module rather than `processing/raster/cog.py` because `catalog/` cannot import `processing/`, and it is where the other shared CRS helpers already live.

Two things I checked before widening the blast radius:
- The change is inert where the old form was right. With `b` and `d` zero the vector length collapses to exactly `abs(a)`/`abs(e)`, so no existing axis-aligned raster's stored resolution moves. Pinned by a test.
- Rotated rasters are already rejected as VRT sources by `_check_rotation` (VAL-07), so the corrected values cannot reach VRT grid maths.

Tests: the rotated-vs-element-0 contrast is asserted on both paths — `test_cog_info.py` against the probe fixture and `test_raster_ingest.py` against a real 30°-rotated GeoTIFF — so a regression to `abs(a)` fails on whichever half reintroduces it. Re-ran the geo.py consumers too since that module is widely imported: 255 passed across the structural and CRS gates, 308 across the VRT/raster batch.

@codex review

ishiland:
Fixed in c36b5c70c, and you found a regression rather than only a gap — thank you.

Confirmed both mechanisms in the source before changing anything: `_check_rotation` errors only on `src.is_rotated` being truthy, `_check_grid_alignment` guards its comparison with `if src.res_x is not None and ref.res_x is not None`, and `router_vrt.py` explicitly admits `storage_backend == "remote"` members while `validate_sources` applies no backend filter — so the path is reachable, not theoretical.

The part that makes it a regression: before this PR the refresh never touched `is_rotated`. My version wrote `described.get("is_rotated", False)`, which downgrades a row previously measured as rotated to `False` whenever `/cog/stac` fails. That is a fabricated measurement, not a missing one, and the NOT NULL column cannot tell the two apart.

The three affine-derived columns now move as one fact on both paths — written together when the transform was read, left alone when it was not. Same argument `sources/cog_info.py` already makes for `crs_wkt`/`epsg` coming off one parsed CRS object: half a fact, written from a source that produced none of it, leaves a row that is neither the old truth nor the new one. Stale is the lesser wrong here because it is the conservative direction — a scene measured as rotated stays flagged rotated.

The new test seeds a rotated row, refreshes it with a probe that answers `/cog/info` but no transform, and asserts all three survive. Verified it is not vacuous: reverting `_pixel_geometry` to the unconditional form fails it on exactly the `is_rotated` assertion.

On the second half of your finding, which I have NOT fixed here and want to be explicit about rather than leave implied: VAL-05 and VAL-07 treating unmeasured geometry as passing is pre-existing and independent of this PR. Every remote raster row has had NULL `res_x`/`res_y` and a defaulted `is_rotated` for as long as STAC import has existed; this PR shrinks that population rather than creating it. Closing it properly means making "not verifiable" its own validation outcome, which starts returning 422 for VRT builds that succeed today — a user-visible admission change that wants its own PR and release note. Filed as #1385 with the mechanism, the suggested direction, and the nullable-`is_rotated` alternative. Say the word if you think it belongs in this PR instead.

@codex review

ishiland:
Fixed in d52a3e5ec. Right on the unit, and I went further than omitting.

Verified the claim: `to_stac_properties()` published `min(abs(res_x), abs(res_y))` with no unit handling at all, and `test_stac_asset_model.py` had an assertion pinning exactly the wrong number — `gsd == 0.001` for a raster whose resolution is 0.001 DEGREES.

PROJ's `units_factor` answers the conversion for projected CRSs directly, so rather than only omitting the geographic case I convert every projected one. That also fixes a class you did not name: the foot-based state-plane systems. EPSG:2263 measures in US survey feet, so a 10-foot pixel was being exported as `gsd: 10` — overstated by 3.28x — and now converts to 3.048.

Geographic CRSs omit the field, as you suggested rather than convert. An angular resolution has no fixed length without a latitude, `gsd` carries no unit of its own to qualify it, and an optional field left out is always conformant where a wrong number is not. Unknown and unparseable CRSs omit it for the same reason.

Deliberately scoped to the STAC serializer, and I want to flag the asymmetry rather than let it look like an oversight: the OGC Records path (`service_records.py`) keeps its CRS-unit value on purpose. It ships a `crs_is_geographic` flag beside `gsd` so its consumer can format the number — that flag is what fix(#569) added after the UI rendered a 60-arc-second DEM as "2 cm". STAC has no such companion field, which is precisely why it has to convert or abstain. I checked #569 before deciding, to be sure I was not overturning a recorded decision about the STAC surface; it is a UI-formatting fix on the Records surface and says nothing about `gsd` in STAC.

New tests cover metre-based, foot-based, geographic and unknown CRSs at the model level.

@codex review

ishiland:
Refuting this one, with evidence — and I think the underlying system property you are describing is real, just not attributable to this PR or fixable inside it.

**The residual case is the pre-#1375 status quo, unchanged.** Before this PR the refresh's UPDATE never wrote `res_x`, `res_y` or `is_rotated` at all:

```
$ git show origin/main:backend/app/processing/ingest/tasks_stac_refresh.py | grep -n 'is_rotated\|res_x\|res_y'
362:            # import of the same asset would record. `res_x`/`res_y` are
```

One comment, no code. A moved asset has always kept the previous row's pixel geometry, including an old `is_rotated=false`. `_pixel_geometry` returning `{}` reproduces exactly that, so the scenario you describe — old flag false, new object unmeasured, validator accepts — is reachable identically on `main` today and is not introduced here. What this PR changes is the other branch: when `/cog/stac` answers, the row now gets real measurements where it previously got nothing. It strictly shrinks the affected population.

**Both remedies you offer are outside what this PR can do.**

*Refuse the repoint when the transform is unknown* trades a rare failure of an optional descriptive probe for a dataset left pointing at a dead URL. That is against this module's stated invariant 10 — a refresh that cannot resolve leaves last-known-good intact — and against the design that makes `/cog/stac` optional in the first place. A Titiler hiccup should not strand a dataset on a URL its publisher has abandoned.

*Represent the geometry as explicitly unknown* is the right answer and needs a migration: `is_rotated` is `NOT NULL DEFAULT false` and cannot hold "unknown". Making it nullable, and teaching the validator a distinct `unknown_pixel_geometry` outcome, is #1385 — filed on your previous round, with that nullable option written up as the more complete of the two remedies. I have added this exact scenario to it as the concrete route: https://github.com/geolens-io/geolens/issues/1385#issuecomment-5246434692

I have deliberately not taken the third option — writing `is_rotated=true` as a conservative default when unknown — because it trades a silent wrong value for a loud one. It would make VAL-07 reject legitimate mosaics and show "rotated" in the UI and API for rasters that are not, corrupting a descriptive field to compensate for a validator that should be asking a different question.

Happy to be overruled if you or a maintainer reads the trade differently; the migration is small and #1385 has it scoped.

@codex review

PR #1386, opened 2026-08-10
chore(release): v1.11.1

ishiland:
Version bump to 1.11.1 across all 19 version sites (`make bump VERSION=1.11.1`) plus the CHANGELOG section.

Patch release carrying the post-1.11.0 fix wave:

- Raster tile cache staleness after replace/refresh (#1372) — tile URLs now carry the dataset content version so the shared nginx cache rolls.
- Reupload swap re-derives effective geometry and record type (#1361, #1373).
- Remote raster rows get measured `res_x`/`res_y`/`is_rotated` from Titiler's transform (#1374, #1384), rotated-raster pixel size corrected on both ingest paths, STAC `gsd` now published in metres or omitted for geographic CRSs (#1384).
- WKT2:2019 CRS serialization at both ingest paths + data migration 0041 backfill (#1376).
- User-authored distributions no longer suppress built-in export rows (#1370); auto-generated distributions reconcile on modality change (#1369).
- AuditExtension seam removed (#1303), origin_ref index (#1365), ArcGIS full-field imports (#1368), empty-string STAC asset key recovery (#1363), worker tile-cache bootstrap (#1371), first-ingest file_hash (#1360), sanitized delete errors (#1358), raster replace exposed in the reupload dialog (#1362).

No schema changes in this PR itself; migration 0041 (data migration, WKT1→WKT2 rewrite) already landed via #1378. Deploy logs will report `converted to WKT2:2019: N row(s)`.

Verification: `make version-check` green (19/19 sites at 1.11.1).

The changelog also picks up the missing `[1.11.0]` compare link that the last release bump omitted.

PR #1387, opened 2026-08-10
test(ogc): page through /collections in the versioned-tiles-link test

ishiland:
Fixes the flake that evicted #1386 (the v1.11.1 bump PR) from the merge queue.

`test_collections_list_raster_tiles_link_is_versioned` (added in #1379) scanned only the first page of `GET /collections`, which defaults to `limit=50` with unordered offset pagination. `test_ogc_features` runs in the `tenancy_global_state` xdist group, whose worker's shared DB accumulates datasets from sibling tests, so the fixture dataset can fall outside page one. The bare `next()` then raises StopIteration inside the coroutine, surfacing as `RuntimeError: coroutine raised StopIteration` (merge-group run 31439751677).

The test now pages through with `limit=200` and stops when a page carries only the catalog Records entry, which every page prepends (`collections=[catalog_collection] + dataset_collections`). Verified passing alongside its sibling detail-endpoint test; no app code touched.

### DIFF
--- a/backend/tests/test_ogc_features.py
+++ b/backend/tests/test_ogc_features.py
@@ -534,12 +534,26 @@ async def test_collections_list_raster_tiles_link_is_versioned(
     client: AsyncClient, raster_dataset: Dataset
 ):
     """fix(#1372 codex r2): the collections LIST advertises the same versioned
-    raster template as the collection detail."""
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review

PR #1388, opened 2026-08-11
fix(vrt): fail validation when pixel geometry was never measured

### LINKED ISSUE #1385: VRT validation treats unmeasured pixel geometry as safe (VAL-05/VAL-07)
Found during the codex review of PR #1384 (#1375). Pre-existing and separate from that PR, which is why it is filed rather than folded in.

`validate_sources` in `backend/app/processing/raster/validation.py` cannot tell "measured and fine" from "never measured", and treats both as passing:

- `_check_rotation` (VAL-07) errors only when `src.is_rotated` is true. The column is `NOT NULL DEFAULT false`, so it cannot express "unknown" — a source whose geometry was never read looks identical to one measured as unrotated.
- `_check_grid_alignment` (VAL-05) compares `res_x`/`res_y` only `if src.res_x is not None and ref.res_x is not None`. A NULL resolution silently skips the comparison instead of failing it.

A remote asset is an eligible VRT source — `router_vrt.py` handles `storage_backend == "remote"` members explicitly, and `validate_sources` runs over the found assets with no storage-backend filter — so a remote raster with unknown pixel geometry can pass both gates and produce a mosaic built on a rotated or grid-misaligned source, which is exactly what VAL-05 and VAL-07 exist to prevent.

Every remote raster row had NULL `res_x`/`res_y` and a defaulted `is_rotated` before #1375, so this has been reachable for as long as STAC import has existed. #1375 shrinks the affected population to remote rows whose transform probe failed, but does not close the hole.

**Suggested direction.** Make "not verifiable" its own outcome rather than a silent pass — a source with NULL `res_x`/`res_y` cannot be proven unrotated or grid-aligned, so it should raise a distinct validation code (e.g. `unknown_pixel_geometry`) instead of skipping the checks.

Worth deciding deliberately, because it changes VRT admission for sources that are accepted today: any existing flow that mosaics remote rasters with no measured resolution would start returning 422. That is the correct answer for a gate whose whole purpose is to refuse unverifiable sources, but it is a user-visible behaviour change and wants its own PR and release note rather than riding along with an unrelated fix.

Alternative worth weighing: make `is_rotated` nullable so "unknown" is representable directly, instead of inferring it from a NULL resolution. That is a migration and a wider blast radius, but it removes the proxy.

ishiland:
## What

`validate_sources()` in `backend/app/processing/raster/validation.py` could not distinguish "measured and fine" from "never measured" for a candidate VRT source:

- `_check_rotation` (VAL-07) only errors when `is_rotated` is `true`, but that column is `NOT NULL DEFAULT false` — a source whose geometry probe never ran looks identical to one confirmed unrotated.
- `_check_grid_alignment` (VAL-05) skips the `res_x`/`res_y` comparison entirely when either is `NULL`, instead of failing it.

A remote asset is an eligible VRT source with no storage-backend filter in `validate_sources`, so a remote raster with unmeasured pixel geometry could pass both gates and end up mosaicked despite being possibly rotated or grid-misaligned — exactly what VAL-05/VAL-07 exist to prevent.

Adds **VAL-08** (`unknown_pixel_geometry`): a source with NULL `res_x` or `res_y` now raises this distinct code instead of silently passing. It runs for both `mosaic` and `band_stack`, mirroring VAL-07 (which applies to both), not just VAL-05 (band-stack only). This is the direction from #1385 — `is_rotated` stays `NOT NULL`, no migration.

## Behavior change (release note material)

Remote rasters with unmeasured pixel geometry that previously passed VRT admission now fail with **422** (`unknown_pixel_geometry`) on both `POST /ingest/vrt/create` and the add-source endpoint. This is intentional — the gate's purpose is to refuse unverifiable sources — but it is user-visible: any existing flow that mosaics remote rasters with no measured resolution will start being rejected.

## Frontend

Checked how VAL-05/VAL-07 currently reach the user. The backend returns `SourceValidationError` objects (`source_id`/`code`/`message`/`field`/`severity`) as a raw array under the 422 `detail`. `error-map.ts`'s `classifyApiError()` array-detail branch expects FastAPI/Pydantic validation-error shape (`loc`/`type`/`ctx`); `SourceValidationError` doesn't match it, so **every** VAL-0X code — including the pre-existing `grid_misaligned` and `rotated_raster` — collapses to the generic `errors.validationFailed` ("The submitted values are invalid.") fallback today, not a code-specific string. (Separately, `VrtCreatorForm.tsx` runs its own pre-submission client-side heuristic from search-result metadata with local i18n keys for 5 of the 7 codes — but that never sees the actual backend response, and already has no equivalent for `single_band_required`/`rotated_raster`.)

Since none of VAL-08's siblings are wired to a code-specific i18n string in the actual backend-response path, `unknown_pixel_geometry` behaves consistently with them (same generic fallback) and needs no frontend change here. Wiring the `SourceValidationError` array shape into `error-map.ts` — mirroring the STAC refusal taxonomy pattern from #1333 — is a separate frontend follow-up worth filing on its own.

## Verification

- `cd backend && uv run pytest tests/test_raster_validation.py -q` — 48 passed (10 new VAL-08 tests: NULL res_x, NULL res_y, both NULL → single error, applies to first source, applies to mosaic and band_stack, measured/unrotated/aligned source passes, existing grid_misaligned/rotated_raster failures still fire unchanged and aren't masked by VAL-08).
- `cd backend && uv run pytest tests/test_layering.py -q` — 40 passed.
- `cd backend && uv run ruff check app/processing/raster/validation.py tests/test_raster_validation.py` — passed.
- `cd backend && uv run ruff format --check app/processing/raster/validation.py tests/test_raster_validation.py` — passed.
- `cd backend && uv run pytest tests/test_vrt_creation_173.py tests/test_vrt_source_management_174.py tests/test_vrt_schema_171.py tests/test_vrt_catalog_175.py tests/test_vrt_source_authz_1172.py tests/test_regenerate_vrt_integration.py tests/test_manifest_apply_vrt.py tests/test_vrt_ingest_tasks.py tests/test_raster_ingest.py tests/test_raster_replace_1221.py tests/test_stac_refresh_1266.py tests/test_refresh_gate_1269.py tests/test_vrt_stale_sweep_gap002.py -q` — 398 passed (no regressions; existing fixtures already use measured resolution).

Closes #1385

### DIFF
--- a/backend/app/processing/raster/validation.py
+++ b/backend/app/processing/raster/validation.py
@@ -1,8 +1,9 @@
 """Source compatibility validation for VRT creation.
 
 Validates candidate COG sources for VRT creation by running a series of
-checks (CRS, dtype, nodata, rotation, band count, grid alignment) and
-returning structured per-source errors.  All checks always run — no fail-fast.
+checks (CRS, dtype, nodata, rotation, band count, grid alignment, pixel
+geometry) and returning structured per-source errors.  All checks always
+run — no fail-fast.
 
 Usage::
 
@@ -149,7 +150,13 @@ def _check_nodata_consistency(sources: list[Any]) -> list[SourceValidationError]
 
 
 def _check_rotation(sources: list[Any]) -> list[SourceValidationError]:
-    """VAL-07: Rotated rasters are rejected."""
+    """VAL-07: Rotated rasters are rejected.
+
+    fix(#1385): `is_rotated` is `NOT NULL DEFAULT false`, so it cannot
+    represent "never measured" — a source whose geometry probe never ran
+    looks identical to one confirmed unrotated. `_check_pixel_geometry_known`
+    (VAL-08) catches that case via the NULL `res_x`/`res_y` it leaves behind.
+    """
     errors: list[SourceValidationError] = []
     for src in sources:
         if src.is_rotated:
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

Issue #1373, opened 2026-08-10
fix(ingest): an empty spatial reupload de-spatializes the dataset it replaces


ishiland:
Found during review of #1369 (#1314).

`extract_metadata()` in `backend/app/processing/ingest/metadata.py` derives the geometry type by sampling a row (`SELECT GeometryType(geom) ... WHERE geom IS NOT NULL LIMIT 1`), so a spatial table carrying zero features — or only NULL geometries — reports `geometry_type=None` even though its geometry column is right there.

`_apply_reupload_swap()` in `backend/app/processing/ingest/tasks_common.py` writes that sampled value straight onto the dataset:

```python
dataset.geometry_type = metadata["geometry_type"]
```

So re-uploading an empty (or all-NULL-geometry) spatial file over a spatial dataset reclassifies it as non-spatial, against a relation that is still spatial.

The consequences are the ones #1313 spelled out when it fixed the identical trap on the registered-PostGIS refresh path (review round 5): `_require_feature_table` refuses feature writes to a dataset whose `geometry_type` is None, so the API is locked out of ever repopulating the table through GeoLens, and the map builder drops the layer as unsupported.

## Why the refresh path is already correct

`tasks_postgis_refresh.py` resolves an **effective** geometry type before writing anything — `_declared_geometry_type()` reads `geometry_columns` for the column's declared type, and `_effective_geometry_type()` applies the precedence: a sampled row wins; no rows but a specific declared type is what the column accepts; a generic `GEOMETRY` column keeps what the catalog last measured; no `geom` column at all is the only case that yields None.

The reupload path has no equivalent. Closing this means deriving an effective type there too — probably by moving `_declared_geometry_type` / `_effective_geometry_type` into `tasks_common` (which `tasks_postgis_refresh` already imports from, so the dependency runs the right way) rather than writing a second spelling of that precedence.

## What #1369 already did

It stopped the auto-generated `record_distributions` rows from following the bad value: the demote there now requires the geometry COLUMN to be absent, not merely the sampled value to be None. So an empty spatial reupload no longer deletes GeoPackage, GeoJSON, Shapefile, GeoParquet and vector-tile rows. The `geometry_type` column itself is still written wrong, which is this issue.

## Acceptance

- [ ] Re-uploading an empty spatial file over a spatial dataset leaves `geometry_type` describing the column, not the absent rows
- [ ] Re-uploading a genuinely non-spatial file (a CSV with no geometry) still sets `geometry_type` to None
[...]

PR #1389, opened 2026-08-11
fix(ingest): retire the synthetic geom row when a reupload de-spatializes

### LINKED ISSUE #1380: fix(ingest): retire the synthetic geom attribute row when a reupload de-spatializes a dataset
Found while closing #1373 / #1361, which brought the reupload swap's `geometry_type` and `record_type` to parity with the registered-PostGIS refresh. A third field is still asymmetric.

`refresh_attribute_metadata()` in `backend/app/processing/ingest/metadata.py` touches the synthetic `geom` attribute row only when it is handed a non-null `geometry_type`, and it excludes `geom` from the removed-column sweep by name:

```python
    # Mark removed columns as is_current=False
    for field_name, am in existing.items():
        if field_name not in current_field_names and field_name != "geom":
            am.is_current = False
```

That is right for every caller that replaces a table's contents while keeping its shape, and wrong for the two callers whose relation can lose its geometry column while keeping its identity.

The refresh path handles it — `refresh_postgis` in `tasks_postgis_refresh.py` retires the row itself when the effective geometry type resolves to None (#1313 review round 7):

```python
            if effective_geometry_type is None:
                AttributeMetadata = port.get_attribute_metadata_orm_class()
                await session.execute(
                    update(AttributeMetadata)
                    .where(
                        AttributeMetadata.dataset_id == dataset.id,
                        AttributeMetadata.field_name == "geom",
                    )
                    .values(is_current=False)
                )
```

`_apply_reupload_swap` in `tasks_common.py` has no equivalent. A CSV reupload over a shapefile now correctly writes `geometry_type = None` and `record_type = "table"`, but `catalog.attribute_metadata` keeps a `geom` row at `is_current = true`, so the attributes API and the validation service go on advertising a geometry field the relation no longer has.

Deliberately left out of the #1373/#1361 PR: those two issues are the fields the swap already wrote, and this is a row it never touched. Filed so the asymmetry is a decision rather than an oversight.

## Acceptance

- [ ] A reupload that removes geometry retires the synthetic `geom` attribute row, exactly as the registered-PostGIS refresh does
- [ ] A reupload that keeps or gains geometry leaves it current
- [ ] The retirement is one shared implementation with the refresh path, not a second copy (the two now share `_effective_geometry_type` / `_derived_record_type` from `tasks_common`)

ishiland:
## What

`_apply_reupload_swap` now retires the synthetic `geom` attribute row when a reupload de-spatializes a dataset, through the same function the registered-PostGIS refresh uses.

`refresh_attribute_metadata` touches that row only when it is handed a non-null geometry type, and excludes `geom` from its removed-column sweep by name. That is right for a caller replacing a table's contents while keeping its shape, and wrong for the two callers whose relation can lose its geometry column while keeping its identity. The refresh path retired it inline from #1313 review round 7; the swap never did, so after #1382 a CSV reupload over a shapefile wrote `geometry_type = None` and `record_type = "table"` beside an attribute row still at `is_current = true`, and the attributes API and the validation service went on advertising a geometry field the relation no longer has.

## How

- `_retire_geometry_attribute_row` lands in `tasks_common.py`, beside `_effective_geometry_type` / `_derived_record_type` that #1382 already moved there, and `tasks_postgis_refresh` imports it in place of its inline `update(AttributeMetadata)` block. One implementation, two callers — a second copy of this UPDATE is how the two paths came to disagree about the row in the first place.
- The null check lives inside the helper, so both call sites are one unconditional line: a caller cannot hold half the pair (`refresh_attribute_metadata`) and forget the other half.
- Fed the **effective** geometry type, the same value handed to `refresh_attribute_metadata`. An empty spatial reupload measures None against a relation whose geom column is right there, and retiring on the sampled value would strike the row of a still-spatial dataset with nothing to restore it.
- `update` drops out of `tasks_postgis_refresh`'s sqlalchemy import (its only use is gone).

## Impacts

No schema, API, or config change. Behaviour change is confined to the reupload swap: a de-spatializing reupload now marks the `geom` attribute row `is_current = false`, which is what the refresh path already did. `tasks_postgis_refresh` shrank by 10 lines; the `tasks_common` cap in `_MODULE_LOC_CAPS` rises 2139 → 2178 with the note explaining what the lines bought.

## Tests

New `TestReuploadRetiresTheGeometryAttributeRow` in `backend/tests/test_dataset_source_state.py`, following the shape #1382 used:

- a non-spatial reupload (CSV over a shapefile) retires the row;
- a spatial reupload leaves it current;
- an **empty** spatial reupload leaves it current — the #1373 trap, on this row;
- a reupload that gains geometry leaves it current, so nothing is retired on the strength of the pre-swap stored value;
- `test_both_paths_share_one_retirement` pins the shared implementation by identity rather than by file path.

Negative control: with the new call commented out, `test_a_non_spatial_reupload_retires_the_row` fails and the three "leaves it current" guards still pass. The refresh path's existing retirement assertion (`test_a_dataset_that_loses_geometry_becomes_a_table` in `test_postgis_refresh_1265.py`) is unchanged and still green.

## Verification

Host-side, from `backend/` with `.env.test` sourced:

- `uv run pytest tests/test_dataset_source_state.py tests/test_postgis_refresh_1265.py tests/test_reupload_swap_lock_retry.py tests/test_attribute_metadata.py tests/test_reupload_service.py -q` → 214 passed
- `uv run pytest tests/test_tasks_common_phase_brackets.py tests/test_refresh_gate_1269.py tests/test_service_refresh_1220.py tests/test_staging_pipeline.py tests/test_phase_2_commit_boundary.py tests/test_raster_replace_1221.py -q` → 208 passed
- `uv run pytest tests/test_layering.py -q` → 40 passed
- `uv run ruff check .` / `uv run ruff format --check .` → clean

Closes #1380

### DIFF
--- a/backend/app/processing/ingest/tasks_common.py
+++ b/backend/app/processing/ingest/tasks_common.py
@@ -1802,6 +1802,46 @@ def _derived_record_type(current: str | None, geometry_type: str | None) -> str 
     return "table" if geometry_type is None else "vector_dataset"
 
 
+async def _retire_geometry_attribute_row(
+    session: Any, dataset_id: uuid.UUID, *, geometry_type: str | None
+) -> None:
+    """Retire the synthetic ``geom`` attribute row of a de-spatialized dataset.
+
+    ``refresh_attribute_metadata`` touches that row only when it is handed a
+    non-null ``geometry_type``, and excludes ``geom`` from its removed-column
+    sweep by name. That is right for a caller that replaces a table's contents
+    while keeping its shape, and wrong for the two callers whose relation can
+    lose its geometry column while keeping its identity: the
+    registered-PostGIS refresh, whose owner can drop the column out from under
+    the catalog, and the reupload swap, which installs a CSV over a shapefile.
+    Left current, the attributes API and the validation service go on
+    advertising a geometry field the relation no longer has.
+
+    fix(#1313 review round 7) established the retirement on the refresh path;
+    fix(#1380) gives the reupload swap the same behaviour from this one
+    function rather than a second copy of it. Feed it the EFFECTIVE geometry
+    type — the same value handed to ``refresh_attribute_metadata`` — and call
+    it unconditionally: the null check lives in here so that a caller cannot
+    hold one half of the pair and forget the other, which is exactly how the
+    two paths came to disagree.
+    """
+    if geometry_type is not None:
+        return
+
+    from app.platform.extensions import get_processing_port
+    from sqlalchemy import update
+
+    AttributeMetadata = get_processing_port().get_attribute_metadata_orm_class()
+    await session.execute(
+        update(AttributeMetadata)
+        .where(
+            AttributeMetadata.dataset_id == dataset_id,
+            AttributeMetadata.field_name == "geom",
+        )
+        .values(is_current=False)
+    )
+
+
 async def _apply_reupload_swap(
     session,
     *,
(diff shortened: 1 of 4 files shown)


ishiland:
@codex review

PR #1391, opened 2026-08-11
refactor(sec): invert Rule 2 safe-env credit to an allowlist of shapes

### LINKED ISSUE #1077: refactor(sec): invert Rule 2 safe-env credit to an allowlist of shapes

ishiland:
## What changed

`backend/tests/test_rule2_structural.py` used to grant safe-env credit by walking the enclosing scope: any canonical `gdal_safe_env` call the walk reached exempted every GDAL CLI argv in that scope. This replaces that with an allowlist. A helper call credits an argv only when two things hold.

**Position.** Every step from the call up to the lowest common ancestor it shares with the argv sits in an allowlisted eager position (`_EAGER_POSITIONS`: a `try` body, a `with` body or item, the value of an assignment or a `return`, a call argument, an element of a list/tuple/set display, an `await`, the outermost iterable of a comprehension). At that shared ancestor the two must meet in the same field, or both sit in eager positions of it, which is what lets `subprocess.run(argv, env=gdal_safe_env())` credit across the args/keywords split.

**Shape.** The env has to go somewhere. `env = gdal_safe_env()` (annotated or not) credits only when the target is read, and a bare `gdal_safe_env()` statement discards the result, so it credits nothing. `gdal_safe_env` is a pure function whose return value *is* the protection.

Anything else reports as unclamped: conditional, deferred, or a shape nobody has named yet. That is a false alarm somebody investigates instead of a silent all-clear in a security gate.

## Why

`if False: gdal_safe_env()`, an untaken ternary arm, and an unreached `match` case all execute eagerly *when reached*, so #996's deferred-body rule never touched them and all three credited silently. There is no finite set to patch here. Every conditional structure is another instance, which is why #1077 asked for an inversion rather than three more rules.

Two things fall out of it. The deferred set (generator expression, lambda, nested def/async def) is now denied by the same machinery at every nesting level rather than by its own rule, which closes the recursion hole #996 documented: `(x for x in (gdal_safe_env() for _ in ()))` credited before and reports now. And an env whose result goes nowhere stops reading as protection.

## Site counts

Measured with the gate's own collectors on `backend/app/`, before the change and after:

| | before | after |
|---|---|---|
| argv sites holding safe-env credit | 4 | 4 |
| argv sites on `GDAL_CLI_CALL_ALLOWLIST` | 7 | 7 |
| `rasterio.open` sites | 6 | 6 |
| violations | 0 | 0 |

The four credited sites are `prepare_with_overviews` (gdaladdo), `convert_to_cog` (gdalwarp and gdal_translate) and `_build_vrt` (gdalbuildvrt). They keep credit through three shapes: `env = gdal_safe_env(extras=...)` inside a `try`; an inline `env=gdal_safe_env()` keyword from inside an `if`, with the argv built in the same `if`; and the same inline form from inside a `try`, with the argv built in the straight-line body above it. `test_guard_real_tree_credit_shapes_are_all_allowlisted` pins all three, so a later narrowing fails in CI instead of in review.

A `finally` body is deliberately absent from that list, and it is the one place the two commits disagree. It passes the eager test (a `finally` runs whenever its `try` is entered) and still cannot clamp anything in the body, because it runs after it. Probing the new model turned up `try: run(argv, env={})` / `finally: env = gdal_safe_env()` crediting an argv the env never reached, so the second commit drops it. A `finally` that builds its own argv beside its own env keeps credit through the same-branch rule, and the regression covers both halves.

One position was widened deliberately while building this, not mid-review: `Return.value`. Without it, `return run_gdal(cmd, env=gdal_safe_env())` reports, and `_build_vrt` is one dropped local variable away from that shape.

## Tests

New regressions:

* the three conditional shapes from the issue (branch, ternary arm, `match` case), each paired with the unconditional control that must still credit
* the loop-body and `except`-handler siblings, plus an in-loop control that keeps its credit because the argv shares the branch
* the nested generator iterable `(x for x in (gdal_safe_env() for _ in ()))`
* a `finally` env that cannot reach the try body, paired with a `finally` that builds its own argv
* the three ways an env goes nowhere: dead store, annotated dead store, discarded result
* `test_guard_real_tree_credit_shapes_are_all_allowlisted`

#996's deferral tests (generator expression, lambda, nested def) pass unchanged, as does the genexp outermost-iterable case.

One existing test changes verdict on purpose. `test_guard_eager_comprehension_still_grants_safe_env_credit` becomes `test_guard_comprehension_body_no_longer_grants_safe_env_credit`: a comprehension body runs once per item and an empty iterable runs it never, so "eager" was never the same claim as "runs". Nothing in `app/` builds its env that way.

## Docstring

The limits section's conditionality paragraph is now a description of the allowlist, what falls outside it, and what widening it costs. It also corrects an over-promise in the old text. The third of the three "credit gap classes" it said #1077 would close is a *detection* gap, whether a GDAL-headed literal is seen as an argv at all (`commands = (["gdalinfo", path],)` then `for cmd in commands:`), not a credit gap. Credit decides which detected sites are exempt, and no exemption rule makes an undetected site appear. Worth its own issue; not this PR.

## Verification

Run from `backend/` with `.env.test` sourced:

* `uv run pytest tests/test_rule2_structural.py tests/test_rule1_structural.py tests/test_layering.py -q` → 130 passed
* `uv run ruff check .` → All checks passed
* `uv run ruff format --check .` → 949 files already formatted

Test infrastructure only. No `backend/app/` changes, no schema, API or config impact.

Closes #1077

### DIFF
--- a/backend/tests/test_rule2_structural.py
+++ b/backend/tests/test_rule2_structural.py
@@ -45,9 +45,10 @@ This module closes that gap structurally, in the spirit of
 2. **GDAL CLI subprocess argv is built next to the safe env or justified,
    per call.** Every argv list or tuple literal whose head names a
    gdal*/ogr* executable (the family, not a fixed list — codex round 10)
-   must sit in a FUNCTION that references ``gdal_safe_env``, or match a
-   ``GDAL_CLI_CALL_ALLOWLIST`` entry keyed (module, function, tool) with an
-   exact expected count and a justification.
+   must be covered by a ``gdal_safe_env()`` call in one of the ALLOWLISTED
+   SHAPES below (fix(#1077)), or match a ``GDAL_CLI_CALL_ALLOWLIST`` entry
+   keyed (module, function, tool) with an exact expected count and a
+   justification.
 
 Both allowlists are asserted EXACT in both directions and by count, so an
 entry whose site disappears (or becomes wrapped) fails loudly instead of
@@ -64,13 +65,13 @@ Known limits (accepted trade-offs, same posture as the Rule-1 guard):
   unsure and its calls are UNCLASSIFIED violations. Re-exports of rasterio
   callables through intermediate modules stay invisible; no such shape
   exists in the codebase.
-- The CLI check is function-scoped with exact counts, not full dataflow: an
-  argv is credited when its ENCLOSING FUNCTION references ``gdal_safe_env``,
-  and unclamped argvs are counted per (module, function, tool) against the
-  allowlist. A function that calls the helper for one subprocess and
-  hand-rolls a second env in the SAME function for the SAME tool would still
-  pass; verifying which env reaches which ``subprocess.run`` is reviewer
-  territory.
+- The CLI check is shape-scoped with exact counts, not full dataflow: an argv
+  is credited when a ``gdal_safe_env()`` call in one of the allowlisted shapes
+  covers it (see CREDIT IS AN ALLOWLIST below), and unclamped argvs are
+  counted per (module, function, tool) against the allowlist. A function that
+  calls the helper for one subprocess and hand-rolls a second env in the SAME
+  function for the SAME tool would still pass; verifying which env reaches
+  which ``subprocess.run`` is reviewer territory.
 - Argv built dynamically (``cmd = [tool_var, ...]``) is not matched. Every
   GDAL CLI call in the codebase starts from a string-literal argv head.
 - A GDAL-headed literal counts as an argv only when its value ESCAPES —
@@ -116,9 +117,10 @@ Known limits (accepted trade-offs, same posture as the Rule-1 guard):
   so their targets cannot shadow a name used elsewhere in the enclosing
   function. Two carve-outs match Python: the outermost iterable is evaluated
   before the comprehension's scope exists, and a walrus inside a
-  comprehension binds in the CONTAINING scope (PEP 572). They remain
-  transparent to CLI call-credit, where the question is whether the scope's
-  immediate body runs the helper and only def/lambda defer.
+  comprehension binds in the CONTAINING scope (PEP 572). For CLI call-credit
+  they are no longer transparent (fix(#1077)): only that outermost iterable is
+  an allowlisted position, because a comprehension body runs once per item and
+  an empty iterable runs it never.
 - Wrapping is judged lexically, with two execution-order rules (codex round
   6): credit stops at def/lambda boundaries (a callable defined inside a
   wrapped block runs after the context exits), and within one ``with``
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review

ishiland:
@codex review

PR #1397, opened 2026-08-11
feat(frontend): map source-validation codes to user-facing strings

### LINKED ISSUE #1390: feat(frontend): map SourceValidationError codes to user-facing strings
Found while closing #1385 (PR #1388).

`classifyApiError()` in `frontend/src/api/error-map.ts` has an array-detail branch that expects the Pydantic validation-error shape (`loc`/`type`/`ctx`). The VRT validation 422 payload carries `SourceValidationError` objects (`source_id`/`code`/`message`/`field`/`severity`), which do not match, so every VAL code — the pre-existing `grid_misaligned` (VAL-05) and `rotated_raster` (VAL-07) included, and now `unknown_pixel_geometry` (VAL-08) — falls through to the generic `errors.validationFailed` string. None of them reach the user as anything more specific than "validation failed".

## Proposal

Recognize the `SourceValidationError` array shape in `classifyApiError()` and map each code to a code-specific i18n string, mirroring the STAC refresh-refusal taxonomy mapping added in #1333. All 8 VAL codes get entries; new `t()` keys go to all four locales (en/es/fr/de) per the locale-parity gate.

## Acceptance

- [ ] A VRT validation 422 renders the per-code message (with source context) instead of `errors.validationFailed`
- [ ] All existing VAL codes and VAL-08 `unknown_pixel_geometry` are covered
- [ ] `npm run test:i18n` passes with the new keys in all four locales

ishiland:
## What

`classifyApiError()` in `frontend/src/lib/error-map.ts` only recognized the Pydantic validation-error array shape (`loc`/`type`/`ctx`). The VRT source-compatibility 422 (`validate_sources` in `backend/app/processing/raster/validation.py`) returns an array of `SourceValidationError` objects (`source_id`/`code`/`message`/`field`/`severity`), which don't match, so every VAL code fell through to the generic `errors.validationFailed` string.

Adds `sourceValidationDescriptor()`, which detects the shape by its `source_id`+`code` signature (a Pydantic entry never has `source_id`) and maps each of the 8 backend VAL codes to a code-specific i18n string carrying the offending `source_id` as context — mirroring the refresh-refusal taxonomy convention from #1333 (one user-facing key per stable backend code). A `SourceValidationError`-shaped entry with an unrecognized code (a future VAL check) still returns the generic fallback string rather than falling through to raw-key rendering.

Codes covered (from `backend/app/processing/raster/validation.py`):

| Code | Check |
|---|---|
| VAL-01 `crs_mismatch` | CRS must match across sources |
| VAL-02 `band_count_mismatch` | Mosaic sources must share band count |
| VAL-03 `single_band_required` | Band-stack sources must be single-band |
| VAL-04 `dtype_mismatch` | dtype must match across sources |
| VAL-05 `grid_misaligned` | Band-stack sources must share grid dimensions/resolution |
| VAL-06 `nodata_inconsistent` | nodata presence must be consistent |
| VAL-07 `rotated_raster` | rotated rasters rejected |
| VAL-08 `unknown_pixel_geometry` | pixel resolution must have been measured |

New `t()` keys (`errors.sourceValidation*`) are added to all four locales (en/es/fr/de).

## Impact

Frontend only. No API, schema, or config change.

## Verification

- `cd frontend && npx vitest run src/lib/__tests__/error-map.test.ts` — 27 passed (new: one case per VAL code, VAL-08 real-shaped payload, unrecognized-code fallback, Pydantic-shape regression guard)
- `npm run typecheck`, `npm run lint` — clean
- `npm run test:i18n` — clean (key parity + translated-value gates)
- `npm run test:coverage` — 385 files / 4772 tests passed

Closes #1390


ishiland:
@codex review

PR #1398, opened 2026-08-11
refactor(sec): follow For.iter containers in Rule 2 argv extraction

### LINKED ISSUE #1394: refactor(sec): Rule 2 argv extraction misses containers iterated with for
Found while closing #1077 (PR #1391).

`test_rule2_structural.py`'s argv extraction follows subscript and attribute loads when resolving GDAL CLI argvs, but not `For.iter`: a container of argvs iterated with a `for` loop reports zero argv sites, so the sites inside it are invisible to the gate entirely.

```python
commands = (["gdalinfo", path],)
for cmd in commands:
    subprocess.run(cmd)   # not counted as a GDAL CLI argv site
```

This is a DETECTION gap, not a credit gap — the #1077 docstring previously misfiled it as the latter (corrected in #1391). A site the collector cannot see can neither be credited nor reported, which is the one failure direction the allowlist inversion was built to eliminate.

## Proposal

Extend `_binding_targets` / `_argv_escape_kind` to follow `For.iter` over a container of argv literals, so each element counts as an argv site subject to the same allowlist credit rules.

## Acceptance

- [ ] The shape above reports as an argv site (and, unclamped, as a violation)
- [ ] `MIN_GDAL_CLI_ARGV_SITES` / `MIN_RASTERIO_OPEN_SITES` floors and the current 11/6 site counts are unaffected on today's tree
- [ ] A regression pins the loop shape, both clamped and unclamped

ishiland:
## The gap

`test_rule2_structural.py` decides whether a GDAL-headed literal is an argv by asking where its value goes. It follows the value out of a container through subscript and attribute loads, so this is tracked:

```python
commands = (["gdalinfo", path],)
subprocess.run(commands[0])
```

Iterating the same container was not:

```python
commands = (["gdalinfo", path],)
for cmd in commands:
    subprocess.run(cmd)          # zero argv sites
```

Zero sites means the gate never sees the call. It cannot credit it and it cannot report it — the one failure direction the #1077 credit inversion was built to remove. Detection, not credit: no exemption rule makes an undetected site appear.

The rule already existed for the spelling where the container is written in place. `for cmd in (["gdalinfo", path],):` has been detected since #996, via `_binding_targets`, which reads the `For` node straight off the AST. Giving the container a name is what broke it: at the load of `commands` the AST no longer says a container is being iterated, and only `_argv_escape_kind`'s alias walk knows.

## The fix

Iterating a container is the same question as subscripting it, so it gets the same answer, in the place that already answers it. `_argv_escape_kind`'s `if holds:` block extracted through `Subscript`/`Attribute`; it now also follows a loop target, binding each element as the vector.

`_loop_target_paths` is the single definition of what a loop binds, called by both readers: `_binding_targets` for the in-place container, `_argv_escape_kind` for the named one.

Getting that right needed four questions answered, all found in review and each with its own regression:

**What does iterating this container yield?** A list, tuple or set yields its elements, so the loop target receives the argv. A dict yields its KEYS, so `for key in {"inspect": ["gdalinfo", path]}` receives a string and the argv never moves. "Holds a container" was one bit and this needs two, so `_binding_targets` returns what iterating the wrapper hands over, and the kind travels with the path across alias hops. The rule reads WHERE the literal sits rather than testing for `dict` — a literal used as a KEY is exactly what iteration yields.

**What does the target receive?** A plain name or path receives the element whole. `for tool, arg in commands:` splits the argv into strings and gets nothing. `for *cmd, in commands:` gets everything — `list(element)`, the vector rebuilt — so it is not destructuring despite being spelled like it. A partial star (`for tool, *rest in commands:`) gets a suffix, which is not a GDAL-headed command.

**Which expression is being iterated?** `for cmd in commands or ():` iterates `commands`. `_depth_preserving_ancestor` names the property once: the outermost expression that is still the same value at the same container depth. That covers `or`, the ternary's two RESULTS (not its condition), `+`, and a star into a display. A CONTAINER wrapper is deliberately excluded — `[commands]` is a level deeper.

**How does `*` count?** Splicing PRESERVES depth. `[*argv]` is the vector; `[*(argv,)]` is a container of it. So a star and the display around it are ONE level, not two, and neither node moves the kind.

One more fix is not about loops at all. The escape worklist was a set of `(path, kind)` pairs collapsed each pass with a dict comprehension, so a path reached with two kinds kept whichever the set yielded last and the verdict moved with `PYTHONHASHSEED`. It predates this PR (the same collapse could pick either boolean). The worklist is now a path -> kind mapping merged by `_louder_kind`, which keeps the answer that licenses MORE following, so a conflict lands on the reported side. `_queue_path` re-queues only when a louder kind arrives late, so loudness rises monotonically and the walk terminates in at most three passes per path.

Credit semantics are untouched. The new sites go through the ordinary #1077 rules, which is worth reading out loud for one case: an env built in the loop BODY does not credit, because a loop body may run zero times. So the creditable spelling hoists the env above the loop, and an in-loop env reports — the same verdict the inline `for cmd in (["gdalinfo", path],):` form has had all along. That consistency is asserted rather than described.

## Site counts

Measured with the gate's own collectors on `backend/app/`:

| | before | after |
|---|---|---|
| `rasterio.open` sites | 6 | 6 |
| violations | 0 | 0 |
| `GDAL_CLI_CALL_ALLOWLIST` entries used | 6 | 6 |

No allowlist entry moved and neither floor (`MIN_GDAL_CLI_ARGV_SITES`, `MIN_RASTERIO_OPEN_SITES`) changed. Nothing in `app/` iterates a named container of argvs today, so the real tree is unaffected by construction — the change adds detection and removes none.

## Tests

Six new regressions, one per question above plus the determinism one:

* the issue shape three ways — unclamped reports, clamped with the env hoisted above the loop credits, clamped inside the loop body reports
* the control for the container gate: `for part in cmd:` and a named tool-name list stay inert
* the dict: inert when iterated, still reached when subscripted, reported when the argv is the KEY; and the kind surviving an alias hop, list beside dict
* destructuring inert (named, inline, partial star) and the all-consuming star reported (named, inline, `[*cmd]`)
* the wrapped iterable: `or`, ternary, `+` and star all report, the ternary CONDITION does not, and `_loop_target_paths` is pinned directly for the container-wrapper case, where end-to-end the container escapes either way
* `*` splicing: star of the vector inert when iterated, still a site when spawned, and star of a container reported both iterated and indexed
* `_louder_kind`'s table, pinned directly, because the end-to-end shape can only catch a last-wins regression on half the seeds

Mutation-checked: stubbing out the loop follow fails exactly the new detection test; making `_container_iteration_kind` uniform fails exactly the two dict tests.

## Residue

The docstring's silent-direction list gains three neighbours of this rule, all "no container-element tracking" in different clothes, each needing its own machinery rather than a wider rule:

- the container kind names what ONE wrapper yields, not a depth, so a container of containers (`groups = ((["gdalinfo", path],),)`) hands the inner container to the loop target as if it were the vector.
- a container reached through a METHOD (`for cmd in commands.values()`) is a call this analysis does not model, which is also why a dict holding argvs as values only ever reaches them by subscript here.
- a comprehension TARGET binds in the comprehension's own scope, which `_use_reaches_the_binding` reads as a shadow, so `[subprocess.run(cmd) for cmd in commands]` links only when the comprehension itself escapes. That one predates this PR and applies equally to the inline `for cmd in (["gdalinfo", path],)` spelling.

The "Still open" paragraph #1391 added to the credit section is rewritten to say this is closed and where.

## Verification

Run from `backend/` with `.env.test` sourced:

- `uv run pytest tests/test_rule2_structural.py tests/test_rule1_structural.py tests/test_layering.py -q` → 139 passed
- `uv run ruff check tests/test_rule2_structural.py` → All checks passed
- `uv run ruff format --check tests/test_rule2_structural.py` → 1 file already formatted
- the collectors over `PYTHONHASHSEED` 0,1,2,3,7,11,42,12345 on nineteen synthetic shapes plus the real tree → identical verdicts on every seed

Test infrastructure only. No `backend/app/` changes, no schema, API or config impact.

Closes #1394

### DIFF
--- a/backend/tests/test_rule2_structural.py
+++ b/backend/tests/test_rule2_structural.py
@@ -1791,17 +1791,19 @@ def _binding_targets(node: ast.AST) -> tuple[set[str], str | None]:
     index_in_parent: int | None = None
     container_kind: str | None = None
     while isinstance(parent, _TRANSPARENT_WRAPPERS):
-        # fix(#1394), codex round 4: `*` SPLICES rather than nests, so a display
-        # holding a starred vector holds its ELEMENTS and is itself a vector —
-        # `[*["gdalinfo", path]]` IS `["gdalinfo", path]`. Reading the display
-        # as a container of the argv made `for part in commands:` look like it
-        # handed over a command when it hands over strings. Checked on the node
-        # being climbed OUT of, so it lands on the display above the star.
-        if isinstance(current, ast.Starred):
-            container_kind = None
+        # fix(#1394), codex rounds 4 and 5: a star and the display around it
+        # are ONE level, not two. `*X` splices X's ELEMENTS into that display,
+        # which PRESERVES X's own depth rather than adding or removing one:
+        # `[*["gdalinfo", path]]` is `["gdalinfo", path]`, still the vector,
+        # while `[*(["gdalinfo", path],)]` is `[["gdalinfo", path]]`, still a
+        # container of it. Counting both nodes reported the first (iterating it
+        # yields strings, not commands); resetting to "is the vector" hid the
+        # second, which is the direction that actually matters. So neither the
+        # star nor the step out of it moves the kind.
+        starred = isinstance(parent, ast.Starred) or isinstance(current, ast.Starred)
         # Only a CONTAINER wrapper means the level above holds the vector
         # rather than being it (fix(#996 review)).
-        elif isinstance(parent, _CONTAINER_WRAPPERS):
+        if isinstance(parent, _CONTAINER_WRAPPERS) and not starred:
             container_kind = _container_iteration_kind(parent, current)
         if isinstance(parent, (ast.Tuple, ast.List)) and current in parent.elts:
             index_in_parent = parent.elts.index(current)
@@ -3324,7 +3326,7 @@ def test_guard_destructuring_loop_target_does_not_receive_the_argv():
 
 
 def test_guard_starring_an_argv_splices_it_into_the_display():
-    """fix(#1394), codex round 4: `*` flattens, so there is no container.
+    """fix(#1394), codex rounds 4 and 5: `*` preserves depth.
 
     ``commands = [*["gdalinfo", path]]`` IS ``["gdalinfo", path]``. Reading the
     outer display as a container of the argv made ``for part in commands:``
@@ -3332,6 +3334,12 @@ def test_guard_starring_an_argv_splices_it_into_the_display():
     reported, the #996 class. The vector is still a vector, so passing the
     spliced display to a subprocess is still a site (the second case), and
     nesting the star one level deeper puts a real container back (the third).
+
+    Depth PRESERVED, not removed, is the whole rule:
+    ``[*(["gdalinfo", path],)]`` splices a one-element tuple and is
+    ``[["gdalinfo", path]]``, a container of the argv either way it is read
+    (the last two cases). Collapsing that to "is the vector" hid a real argv,
+    which is the direction that matters.
     """
     inert, total = _collect_gdal_cli_violations(
         _mod(
@@ -3369,6 +3377,24 @@ def test_guard_starring_an_argv_splices_it_into_the_display():
     )
     assert total_nested == 1, (total_nested, nested)
     assert len(nested) == 1 and "(fn)" in nested[0]
+
+    wrapped, total_wrapped = _collect_gdal_cli_violations(
+        _mod(
+            "import subprocess\n"
+            "def iterated(path):\n"
+            "    commands = [*(['gdalinfo', path],)]\n"
+            "    for cmd in commands:\n"
+            "        subprocess.run(cmd)\n"
+            "def indexed(path):\n"
+            "    commands = [*(['ogrinfo', path],)]\n"
+            "    subprocess.run(commands[0])\n"
+        ),
+        {},
+    )
+    assert total_wrapped == 2, (total_wrapped, wrapped)
+    assert len(wrapped) == 2, wrapped
+    assert any("(iterated)" in v for v in wrapped)
+    assert any("(indexed)" in v for v in wrapped)
 
 
 def test_guard_conflicting_container_kinds_merge_to_the_louder():
--- a/backend/tests/test_rule2_structural.py
+++ b/backend/tests/test_rule2_structural.py
@@ -1585,8 +1585,8 @@ def _loop_target_paths(iterable: ast.AST) -> set[str]:
 
     ``for cmd in commands:`` binds ``"cmd"``; ``async for`` and a
     comprehension's generator bind the same way. Empty when ``iterable`` is not
-    in an iterated position, or when the target is a shape ``_target_names``
-    cannot name.
+    in an iterated position, when the target is a shape ``_target_names``
+    cannot name, or when the target DESTRUCTURES.
 
     fix(#1394): ONE definition of what a loop binds, because two callers need
     it and they know different things. ``_binding_targets`` calls it for a
@@ -1595,11 +1595,22 @@ def _loop_target_paths(iterable: ast.AST) -> set[str]:
     calls it for a NAME already known to hold a container of argvs
     (``commands = (["gdalinfo", path],)`` then ``for cmd in commands:``), where
     it is not.
+
+    The destructuring rule is codex round 2 on #1394, and it predates the
+    named-container half: ``for tool, arg in commands:`` splits the element
+    apart, so each name gets a PIECE of the argv and none of them gets the
+    argv. ``_target_names`` flattens a pattern to all its names, which is the
+    right answer for an assignment — where ``_positional_targets`` then matches
+    them up by position — and the wrong one here, where the value being split
+    is the command vector rather than a tuple of unrelated values. Returning
+    nothing is the accurate answer within this model, in which the element of a
+    container IS the vector: destructuring a vector yields strings.
     """
     parent = getattr(iterable, "_rule2_parent", None)
     if (
         isinstance(parent, (ast.For, ast.AsyncFor, ast.comprehension))
         and parent.iter is iterable
+        and not isinstance(parent.target, (ast.Tuple, ast.List))
     ):
         return _target_names(parent.target)
     return set()
@@ -3220,6 +3231,31 @@ def test_guard_iterating_a_dict_yields_keys_not_the_argv():
     )
     assert total_keyed == 1, (total_keyed, keyed)
     assert len(keyed) == 1 and "(fn)" in keyed[0]
+
+
+def test_guard_destructuring_loop_target_does_not_receive_the_argv():
+    """fix(#1394), codex round 2: unpacking splits the argv into strings.
+
+    ``for tool, arg in commands:`` binds ``"gdalinfo"`` and the path, never the
+    vector, so linking the literal to either name flagged data that cannot
+    execute. Both spellings of the container are checked, because the
+    destructuring bug predates the named-container half — the inline form went
+    through the same ``_target_names`` flattening on #996.
+    """
+    violations, total = _collect_gdal_cli_violations(
+        _mod(
+            "def named(path):\n"
+            "    commands = [['gdalinfo', path]]\n"
+            "    for tool, arg in commands:\n"
+            "        consume(tool)\n"
+            "def inline(path):\n"
+            "    for tool, arg in (['ogrinfo', path],):\n"
+            "        consume(arg)\n"
+        ),
+        {},
+    )
+    assert total == 0, (total, violations)
+    assert violations == []
 
 
 def test_guard_container_kind_survives_an_alias_hop():
--- a/backend/tests/test_rule2_structural.py
+++ b/backend/tests/test_rule2_structural.py
@@ -83,7 +83,13 @@ Known limits (accepted trade-offs, same posture as the Rule-1 guard):
   target over a container of argvs — chased through alias chains and out of
   containers to a fixed point. Consuming operations stop it: a single-index
   subscript of the VECTOR yields an element, while the same subscript of a
-  CONTAINER yields the vector. Inert data
+  CONTAINER yields the vector. ITERATING one is the same question as
+  subscripting it and gets the same two answers (fix(#1394)): a ``for`` over a
+  container binds each argv to the loop target, while a ``for`` over the vector
+  binds strings. That holds whether or not the container has a name —
+  ``for cmd in (["gdalinfo", path],):`` and ``commands = (["gdalinfo", path],)``
+  then ``for cmd in commands:`` are one rule, read off the AST in the first case
+  and off the followed path in the second. Inert data
   (``SUPPORTED_TOOLS = ["gdalinfo", "ogrinfo"]``, a constant only subscripted
   or compared) is not a command vector. A literal
   whose every element names a GDAL utility is a name list rather than a
@@ -108,6 +114,18 @@ Known limits (accepted trade-offs, same posture as the Rule-1 guard):
   neither is one extracted from an ANONYMOUS container that never gets a name
   (`cmd = ({"k": ["gdalinfo", path]})["k"]`), because the path machinery keys
   on a binding and there is none to key on.
+  Three more silent residues sit next to the iteration rule (fix(#1394)), all
+  of them "no container-element tracking" wearing different clothes, and each
+  needs its own machinery rather than a wider rule: `holds_a_container` is a
+  BOOLEAN, not a depth, so a container of containers (`groups = ((["gdalinfo",
+  path],),)`) hands the inner container to the loop target as if it were the
+  vector and the second loop follows nothing; a container reached through a
+  METHOD (`for cmd in commands.values()`) is a call this analysis does not
+  model; and a comprehension TARGET is bound in the comprehension's own scope,
+  which `_use_reaches_the_binding` reads as a shadow, so
+  `[subprocess.run(cmd) for cmd in commands]` links only when the comprehension
+  itself escapes — that one predates #1394 and applies equally to the inline
+  `for cmd in (["gdalinfo", path],)` spelling.
   Extending further means writing a static analyser, and this is a CI gate —
   the escape hatch for anything it misclassifies is an allowlist entry with a
   written justification, which is the same answer #974 reached for the wrapper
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review

ishiland:
Fixed in e03ab554a — dict iteration yields keys, so the loop-target follow now carries what iterating the wrapper hands over instead of a bare "holds a container" bit, and the kind survives alias hops. Subscripting a dict is unchanged. Regressions pin the inert dict, the subscript control, the argv-as-KEY case, and the alias hop.

@codex review

ishiland:
Fixed in fff45c2fc — `_loop_target_paths` returns nothing for a destructuring target. That is the accurate answer within this model: the element of a container IS the vector, so unpacking it yields strings. The bug predates the named-container half (the inline spelling went through the same `_target_names` flattening on #996), so the shared helper fixes both and the regression pins both.

@codex review

ishiland:
Fixed in aec413d0b — the worklist is now a path -> kind mapping merged by `_louder_kind` rather than a set of pairs collapsed last-wins. The merge keeps the kind that licenses more following, so a conflict lands on the reported side. `_queue_path` re-queues only when a louder kind arrives late, so loudness rises monotonically and the walk terminates in at most three passes per path.

The collapse predates the kinds (it could pick either boolean), so this closes a latent flake as well as the widened one.

Verified over PYTHONHASHSEED 0,1,2,3,7,11,42,12345: identical verdicts on every shape and on the real tree (11 GDAL CLI argv / 6 rasterio.open, zero violations). The merge rule itself is unit-pinned, because the end-to-end shape can only catch a last-wins regression on half the seeds.

@codex review

ishiland:
Fixed in 11de554f8 — the container kind resets when the climb steps out of a `Starred`, so `[*argv]` reads as the vector rather than a container of it. Checked on the node being climbed out of, so it lands on the display above the star; `[[*argv]]` still puts a real container back one level up, and a spliced display handed to a subprocess is still a site. All three pinned.

Verified over PYTHONHASHSEED 0,1,2,3,7,11,42,12345: identical verdicts on every shape, real tree still 11 GDAL CLI argv / 6 rasterio.open with zero violations.

@codex review

ishiland:
Fixed in e591a5054 — right, the reset removed one wrapper too many. Splicing PRESERVES depth: `[*argv]` is the vector, `[*(argv,)]` is a container of it. So the rule is now stated once — a star and the display around it are ONE level, not two, and neither node moves the kind. Counting both was the round-4 false positive; counting neither fixes both.

The regression gains the wrapped-star shape read two ways, iterated and indexed, since either must reach the argv.

Verified over PYTHONHASHSEED 0,1,2,3,7,11,42,12345: identical verdicts on thirteen shapes including `[*[*(argv,)]]`, real tree still 11 GDAL CLI argv / 6 rasterio.open with zero violations.

@codex review

ishiland:
Fixed in 21e7a5dc2 — `_depth_preserving_ancestor` names the property once (the outermost expression that is still the same value at the same container depth) and the loop rule reads the iterable through it, so the ternary, `+`, `or` and starred forms all reach the loop. A CONTAINER wrapper is deliberately not in it: `[commands]` is a level deeper. Pinned on the helper, since end-to-end the container escapes into the loop body either way and the two answers are indistinguishable from the verdict.

Verified over PYTHONHASHSEED 0,1,2,3,7,11,42,12345: identical verdicts on fifteen shapes, including the wrapped dict, wrapped vector and wrapped tool-name list staying inert. Real tree still 11 GDAL CLI argv / 6 rasterio.open with zero violations.

@codex review

ishiland:
Fixed in 84fdbc838, both of them.

`_intact_loop_target` now names the one pattern that is not destructuring despite being spelled like one: `for *cmd, in commands:` binds `cmd` to `list(element)`, the vector rebuilt. `for tool, arg in ...` still gets nothing, and neither does a PARTIAL star, whose target is a suffix of the argv rather than a GDAL-headed command.

`_depth_preserving_ancestor` climbs a ternary through its two RESULTS only; its condition is not part of its value.

Verified over PYTHONHASHSEED 0,1,2,3,7,11,42,12345: identical verdicts on nineteen shapes, including both ternary results still reporting and the partial star staying inert. Real tree still 11 GDAL CLI argv / 6 rasterio.open with zero violations. PR body rewritten to match what the change grew into.

@codex review

ishiland:
Fixed in cfafb4326 — both are one rule, so it is now written once. `_is_value_position` answers "can this wrapper evaluate to this child" for both climbs, so the loop's own iterable and the alias path cannot drift apart again (the ternary half was fixed at the iterable in round 7 and the alias path went around it).

`and` yields a non-final operand only when it is FALSY, so `commands and ()` can only ever be an empty `commands`. `or` decides on the opposite truthiness and the final operand is a result for both, so `commands or ()` and `() and commands` still report.

Verified over PYTHONHASHSEED 0,1,2,3,7,11,42,12345: identical verdicts on twenty-five shapes, including both live alias spellings still reporting and a dead alias not hiding a direct spawn beside it. Real tree still 11 GDAL CLI argv / 6 rasterio.open with zero violations.

@codex review

ishiland:
b212e474e — one fixed, one declined.

**Slice.** Fixed. `for cmd in commands[:]:` iterates argvs and the loop rule could not climb the subscript. #996 already reads a slice this way in `_escape_kind`, where only a SINGLE index stops the walk because it consumes a level; the loop rule and the alias extraction did not. Both now do — the alias half (`sliced = commands[:]`) matters on its own, since that is exactly how round 8 got around the round-7 fix. An index still yields the vector one level down; slicing the vector itself still yields strings.

**`commands * 0`.** Declined, documented in `_is_value_position`. Reading it as possibly non-empty is an over-report, and this gate does no constant folding anywhere — the same decision #1077 made when it chose to report `if False: gdal_safe_env()` rather than evaluate the condition. Both cost a false alarm; neither costs a silent pass. The ternary-condition and `and`-operand exclusions from round 8 are not folding: they read which FIELD an operator can return, with no claim about any operand's value. Adding truthiness evaluation for one literal spelling would be the first crack in that, for a shape no code writes, and the escape hatch for a genuine false alarm is an allowlist entry with a justification (the standing answer since #974).

Verified over PYTHONHASHSEED 0,1,2,3,7,11,42,12345: identical verdicts on twenty-nine shapes, real tree still 11 GDAL CLI argv / 6 rasterio.open with zero violations.

@codex review

ishiland:
Fixed in 647192829 — `_is_value_position` gains the `NamedExpr.value` field, which puts the walrus with the ternary results and the `or` operands where it belongs: the question is which field a wrapper can evaluate to, and a walrus can only evaluate to its value. The binding half genuinely does not help here, since the target is a Store and the alias walk only follows Loads.

Verified over PYTHONHASHSEED 0,1,2,3,7,11,42,12345: identical verdicts on thirty shapes, including the walrus stacked under an `or` and a walrus of the VECTOR staying inert. Real tree still 11 GDAL CLI argv / 6 rasterio.open with zero violations.

@codex review

ishiland:
Rebased onto main (7fed6c965) — content of the change is unchanged, no conflicts. Re-verified on the new base: 141 passed across rule2 + rule1 + layering, whole-tree `ruff check` and `ruff format --check` clean, and the real tree still measures 11 GDAL CLI argv / 6 rasterio.open sites with zero violations (#1400's ingest decomposition landed in between and moved nothing the gate keys on).

Re-triggering so the verdict names the current head.

@codex review

ishiland:
Fixed in c2f8a49b3 — an unpacked target now gets the kind from BELOW the wrapper the unpacking consumed, which is what unpacking actually hands it. `alias, = (argv,)` makes `alias` the vector; one level deeper still hands over a container; and `alias` being the vector means spawning it directly is still a site, so the fix does not turn the unpacking into a dead end. All three pinned, plus the #996 sibling-position pair, which is unaffected.

Verified over PYTHONHASHSEED 0,1,2,3,7,11,42,12345: identical verdicts on twenty-seven shapes, real tree still 11 GDAL CLI argv / 6 rasterio.open with zero violations.

@codex review

ishiland:
Fixed in fa3666fb9 — the one-level pairing is now a chain. The climb records every wrapper it passes (the index the value occupies and the kind of what is inside), and `_unpacked_binding` walks that chain outermost first against the target patterns, consuming one wrapper per pattern and stopping where the targets stop being patterns. Falling off either end is the same loop, so there is no rule per depth.

Verified over PYTHONHASHSEED 0,1,2,3,7,11,42,12345: identical verdicts on thirty-two shapes, including three levels of nesting, #996's sibling-position and chained-target pairs unchanged, and a dict unpacked out of a list keeping its mapping kind. Real tree still 11 GDAL CLI argv / 6 rasterio.open with zero violations.

@codex review

ishiland:
Fixed in 2f6b2a126, both.

The chain now records whether each wrapper is a LEVEL at all — the same condition that moves the kind, so the two cannot disagree — and the pairing steps over the ones that are not instead of ending there. `(alias,) = ((argv,) + ())` unpacks the concatenated tuple exactly as `(alias,) = (argv,)` does.

Each chained target is kept at the depth where its pattern ends, so the uneven chain no longer drops the shallow one. The caller speaks of one path set with one kind, so the endpoint kinds merge through `_louder_kind` and the disagreement lands on the reported side, as everywhere else here.

Verified over PYTHONHASHSEED 0,1,2,3,7,11,42,12345: identical verdicts on thirty-six shapes, including both ends of the uneven chain, an unpacking beside a plain target, three levels of nesting, and #996's sibling-position and chained-target pairs unchanged. Real tree still 11 GDAL CLI argv / 6 rasterio.open with zero violations.

@codex review

ishiland:
Fixed in 6b685d0e6 — `_binding_targets` now takes the kind its node already carries and prepends it to the wrapper chain as the innermost level, with no known position: a container the analysis knows only by KIND has none, and in a container of argvs every position is one. A MAPPING gets no such entry, because unpacking it yields keys and keys are not commands.

The answer comes back absolute, which collapses the three alias-site cases — a new wrapper, no wrapper, an unpacking — into one question with one answer, instead of a caller-side `or` that could only express two of them.

Verified over PYTHONHASHSEED 0,1,2,3,7,11,42,12345: identical verdicts on forty-two shapes, including the unpacked alias still reporting when spawned or sliced, a plain alias unchanged, and the index/slice extractions unchanged. Real tree still 11 GDAL CLI argv / 6 rasterio.open with zero violations.

@codex review

ishiland:
dfbac4ad1 — one fixed, one declined.

**Per-endpoint kinds.** Fixed, and by removing the constraint rather than patching the shape: `_binding_targets` returns `{path: kind}` instead of a set plus one kind. The merge existed only because the answer could not say two things; it is now left to the one case that needs it, a name that really does appear at two depths. `for part in deep:` is inert, `for cmd in alias:` still reports, and spawning `deep` still reports.

**Heterogeneous container through a name.** Declined, documented at `_ANY_POSITION`. A container reached through a NAME has no known position — positions are exactly what a kind drops, and this analysis has tracked no container elements since #996, which is the residue #996 declined to close by writing a static analyser. So unpacking one binds every name in the pattern and an inert position reports too. The alternative on the table is dropping the inherited level, and that reinstates the round-14 silent miss: it trades a false alarm somebody investigates for an all-clear nobody sees, which is the wrong direction for this gate.

Verified over PYTHONHASHSEED 0,1,2,3,7,11,42,12345: identical verdicts on forty-two shapes, including both ends of the uneven chain read both ways. Real tree still 11 GDAL CLI argv / 6 rasterio.open with zero violations.

@codex review

ishiland:
822a6a4e6 — both declined, with the reasoning recorded at the code that makes each call.

**`cmd, *rest = commands`.** Keeping the non-starred names and dropping the star is the obvious narrowing, and it is a trap: the star's target holds a container of argvs, so a later `for c in rest:` is a real path to an exec. Dropping it trades this false alarm for a silent miss, which is the one direction that is worse. Splitting the two kinds means the element pickers returning a kind per target rather than a level — the container-element tracking #996 explicitly declined to build.

**`cmd, = {("gdalinfo", path)}`.** A singleton set does hand its sole element over unambiguously, but reading that needs the wrapper's arity, which the chain does not carry. Sets are unordered, so nothing builds an argv in one.

Both are over-reports on shapes with zero occurrences in `backend/app`, which is the trade this module has documented since #974: the escape hatch for a genuine false alarm is an allowlist entry with a written justification, and the notes say what to widen if a real site ever asks.

Worth stating where this loop stands after sixteen rounds. Every finding in the SILENT direction has been fixed — the loop rule itself, dict iteration, destructuring, the value-wrapper positions, slices, the walrus, nested and uneven unpacking, the inherited level, per-endpoint kinds. What is left is a tail of over-reports on shapes nobody writes, and this gate chooses that direction on purpose. Three are now declined in-tree with their reasoning (`commands * 0`, the heterogeneous named container, and these two).

@codex review

ishiland:
Fixed in 999e71c2a — good catch, that one is in the silent direction. Position, not arity, is what decides it: a star FIRST keeps the head (`*cmd,` takes the whole element, `*cmd, ignored` everything but the tail, both still starting with the tool name), while a star anywhere else takes a suffix, which has no head and would be judged against the wrong tool. The rule reads that directly now and covers round 7's all-consuming case as its one-element instance rather than as a separate clause.

The assignment path keeps its round-16 answer, because it is a different question: there the star splits the CONTAINER, not the argv, so its target holds argvs rather than being one.

Verified over PYTHONHASHSEED 0,1,2,3,7,11,42,12345: identical verdicts on nineteen shapes, including a trailing and a middle star staying inert. Real tree still 11 GDAL CLI argv / 6 rasterio.open with zero violations.

@codex review

ishiland:
256715472 — declined, documented at the rule that makes the call.

`for *cmd, ignored in [("gdalinfo",)]:` binds `cmd` to `[]` because the pattern's fixed names eat the only element, so the site reports and nothing runs. Closing it means comparing the pattern's arity against the LITERAL's, which couples a target-shape question to a value `_intact_loop_target` never sees, for a shape needing a one-element argv and a starred pattern at once. And it is the loud direction on a literal the module already treats as a command by convention: a single-element `("gdalinfo",)` is documented as indistinguishable from a bare invocation, which is why `_tool_name_list` has a two-element floor.

Standing where this loop is after eighteen rounds: every finding in the SILENT direction has been fixed, and the remaining tail is over-reports on shapes with zero occurrences in `backend/app`. Four are now declined in-tree with their reasoning and a note saying what to widen if a real site ever asks — `commands * 0`, the heterogeneous named container, `cmd, *rest = commands` (where narrowing would create a silent miss), the singleton set, and this one. That is the trade this module has documented since #974, and the escape hatch for a genuine false alarm is an allowlist entry with a justification.

@codex review

Issue #1104, opened 2026-08-01
Curved geometries (MultiSurface/CompoundCurve) break tiles, feature reads, and pre-existing analysis ops — normalize geom_4326 at ingest


ishiland:
## Background

WFS ingest admits curved geometry types. GeoServer MultiSurface/CompoundCurve layers load as stored, and `_ABSTRACT_TO_CONCRETE_GEOMETRY_TYPE` in `backend/app/processing/ingest/metadata.py` classifies the dataset by the closest concrete linear type *without touching the binary geometry data* — its own comment says "Until the database layer normalizes the stored geometries themselves". This issue is that normalization.

PR #1097 linearizes at the analysis boundary (`linearized()` in `backend/app/platform/analysis_sql.py`) so the four operations it adds survive curved rows. But the underlying class is wider than analysis, and everything below is broken **on main today** for a curved dataset. Each failure verified against the running PostGIS 3.6:

## Confirmed broken surfaces

1. **Vector tiles** — `ST_AsMVTGeom(curved)` → `ERROR: lwgeom_get_basic_type: Invalid type (12)`. A curved layer renders no tiles.
2. **Feature reads** — `ST_AsGeoJSON(curved)` → `ERROR: 'CompoundCurve' geometry type not supported` (GeoJSON has no curved types). The features browse endpoints 500.
3. **buffer / centroid / clip** — `ST_MakeValid(curved)` → `ERROR: unsupported geometry type` (`render_geometry_expr`, `render_clip_layer_join`).
4. **dissolve** — same `ST_MakeValid` raise inside its `ST_Union` CTAS expression (`processing/analysis/tasks.py`).

Curve-tolerant, for reference: `&&`, `ST_Intersects`, `ST_Intersection`, `ST_Union`, `ST_Dimension`, `ST_IsEmpty`, `ST_Centroid` (the raise in the centroid op is the `ST_MakeValid` wrapper, not `ST_Centroid` itself).

## Proposed fix

Normalize at the one boundary every surface reads from: make `geom_4326` always linear.

- `add_4326_column()` (`processing/ingest/metadata.py`) applies `ST_CurveToLine` before `ST_Force2D`. It is an exact no-op on already-linear input (verified with `ST_Equals`), so non-curved ingest is unaffected. The original `geom` column keeps the curved source data, same as it keeps Z.
[...]

PR #1400, opened 2026-08-11
refactor(ingest): decompose processing/ingest/metadata.py

### LINKED ISSUE #1042: refactor(ingest): decompose processing/ingest/metadata.py (2002 lines, four cap carve-outs deep)

ishiland:
`backend/app/processing/ingest/metadata.py` was 2151 lines with five carve-outs on its `_MODULE_LOC_CAPS` entry, each one a correctness fix that had to argue for its lines (#888, #899, #906, #934, #961, then #1104/#1113's eight review rounds). #958 named the problem — the ratchet taxing correctness work on a module nobody had time to split — and filed this issue.

## What moved where

| module | LOC | contents |
|---|---|---|
| `metadata_sql.py` | 60 | `_TABLE_NAME_RE`, `_validate_table_name`, `_qtable`, `_sql_quote_ident` |
| `metadata_quality.py` | 190 | the four `_score_*` dimensions and `compute_quality_score` |
| `metadata_projection.py` | 266 | `add_4326_column`, `linearize_existing_4326`, `ensure_geom_4326_gist_index`, `grant_reader_access` |
| `metadata_attributes.py` | 352 | name/type inference plus `generate_attribute_metadata` / `refresh_attribute_metadata` |
| `metadata_mercator.py` | 361 | `clip_to_mercator_bounds`, `_shift_zero_to_360_longitudes`, `_mercator_envelope_degenerates`, the srtext CRS predicates |
| `metadata_geometry.py` | 424 | `construct_point_geometry`, `construct_wkt_geometry`, `ensure_geom_column`, `rename_reserved_columns`, `detect_dbf_truncation_collisions` |
| `metadata_extent.py` | 633 | `extract_metadata` and everything it consolidates: `get_extent`, `_seam_crossing_extent_wkt`, the four introspection helpers, `get_column_info` / `get_sample_values`, the 3D pair |
| `metadata.py` | 144 | re-export façade (49 names, `# noqa: F401` + explicit `__all__`) |

`metadata_sql` holds only what every other module needs and needs nothing back, which is what keeps the other six from importing each other.

## Why these seams

Two differ from the issue's sketch.

**The SQL CRS predicates stay with the clip** instead of becoming their own module. `_GEOGRAPHIC_SRTEXT_RE` and `_DEGREE_UNIT_SRTEXT_RE` have exactly one caller each (`_shift_zero_to_360_longitudes`), and the inline `GEOG(CS|CRS)` test lives inside `_mercator_envelope_degenerates`. #961's standing obligation is that the anchored gate and the degenerate-envelope floor *deliberately disagree* on wrapped CRSs — splitting the two halves of that disagreement into separate files would make the one thing the obligation is about unreadable. The issue asked for those cross-references to stay "true and in one place"; one place is this module.

**The extent producers keep company with the four introspection helpers.** `extract_metadata`'s CTE consolidates `get_feature_count` / `get_table_srid` / `get_geometry_type` / the extent query into one scan and falls back to calling those same four one at a time, so the CTE and the helpers have to be legible side by side.

## No behavior change

All 49 top-level symbols are byte-identical to their pre-split text — verified by AST segment comparison against the parent commit (zero missing, zero differing, zero extra). The façade re-exports every one, so existing importers and the `mock.patch("app.processing.ingest.metadata.<name>")` targets across the suite keep resolving. `ruff check --fix --diff` produces no diff, so autofix cannot strip the re-exports.

Two references had to move to the module a call physically lives in, because the façade name is not what the code reads:

- `tests/test_ingest_mercator_clip.py` monkeypatches `_mercator_envelope_degenerates` for the shift-before-a-skipped-clip ordering test. `clip_to_mercator_bounds` resolves that name from `metadata_mercator`'s globals, so a patch on the façade would have bound a name nothing reads and the probe would have run for real.
- `tests/test_crs_degree_agreement.py` imports the two srtext regexes; it is the gate on where those predicates live, so it names the file.

Every other patch target reaches `metadata` through a deferred (function-local) import and resolves the attribute at call time, so the façade keeps them working unchanged.

## Caps and gates

- `_MODULE_LOC_CAPS`: `ingest/metadata.py` **2151 → 144**, exact. No extracted module is near the 1000-line inclusion threshold, so none needs its own entry yet (`test_module_loc_cap_inclusion_rule_is_complete` passes). The carve-out history is condensed into the entry's comment rather than dropped.
- ruff `C901` baseline: the `ingest/metadata.py` entry follows `refresh_attribute_metadata` (McCabe 16, the file's only over-complex function) to `metadata_attributes.py` — retargeted, not new, same as the `defaults_ai_openai.py` precedent. The list still only shrinks in count.
- `test_analysis_provenance.py`'s lineage-summary allowlist follows `_score_metadata_completeness` to `metadata_quality.py`.

## Cross-references

Updated on both sides: `core/geo.py`'s two pointers at the extent writers, `ogr.py`'s four, `analysis_sql/shared.py`, `column_stats.py`, and the four test docstrings that named `metadata.py` by file.

None. No route, model or serializer is touched, so no OpenAPI or SDK regeneration.

## Verification

```
cd backend && uv run ruff check .            # All checks passed
cd backend && uv run ruff format --check .   # 963 files already formatted
cd backend && uv run ruff check --fix --diff app/processing/ingest/   # no diff (re-exports survive)
./run_tests_1042.sh -n 4 -q                  # 8339 passed, 92 skipped in 15:08
./run_tests_1042.sh tests/test_layering.py   # 40 passed
```

Plus a targeted re-run of the whole ingest-metadata surface after the last comment edits (343 passed): `test_layering`, `test_ingest_mercator_clip`, `test_extent_producers_934`, `test_crs_degree_agreement`, `test_analysis_provenance`, `test_phase_2_commit_boundary`, `test_ingest_metadata`, `test_staging_pipeline`, `test_ingest_column_preservation`, `test_geometry_detection`, `test_quality_score`, `test_dp_review_fixes`, `test_ingest_service_geometry_type`, `test_sql_safety`, `test_ensure_geom_column`, `test_ingest_ogr_pure`.

Closes #1042


ishiland:
@codex review

Issue #1236, opened 2026-08-06
ops(ingest): post-expiry sweep can orphan a staging object across a timeout-lowering restart


ishiland:
Known gap accepted in PR #1235 (review r5), documented in _sweep_expired_presigned_staging's docstring and the pending_job_timeout_seconds config comment.

Mechanism: an operator lowers pending_job_timeout_seconds and restarts while presigned URLs issued under the previous, longer value are still live. The post-expiry sweep derives its window from the CURRENT setting, so it can reap the staging object and set s3_key_reaped while an old URL is still valid; a later PUT through that URL recreates the object, and the reaped-marker suppresses every subsequent pass — a permanent orphan.

Why it was not fixed in #1235: persisting the issued deadline per job (user_metadata JSONB) contorts the sweep's WHERE clause with a cast that can throw and fail the whole bulk pass, and it cannot protect the motivating population anyway — jobs presigned BEFORE the persistence change carry no key and fall through to the same fallback.

Fix directions for whoever picks this up:
- A dedicated nullable timestamptz column (url_deadline) on ingest_jobs, written at presign time — schema migration, but the sweep predicate becomes a plain coalesce over a typed column that cannot throw.
- Or: make the reaped-marker non-permanent (a marked row older than the SigV4 ceiling, 604800s, gets one re-check pass) — bounds the orphan's lifetime without any schema change.

Operational guidance until then (already in the config comment): lower the setting only when no presigned uploads are in flight, or wait out the previous TTL before trusting the sweep.

Issue #1211, opened 2026-08-05
ops(storage): document AbortIncompleteMultipartUpload lifecycle rule for abandoned presigned multipart uploads


ishiland:
Found while closing the presigned staging lifecycle in #1202 (PR #1209, review round 8).

A client that initiates a presigned multipart upload and walks away — never completing, never aborting — leaves an incomplete multipart upload whose parts consume bucket storage. No object exists at the key, so none of the staging sweeps added in #1209 can see it, and nothing in the application aborts it (the app only aborts on an explicit failed/empty completion).

The standard remedy is operational, not app code: an `AbortIncompleteMultipartUpload` bucket lifecycle rule (S3 and MinIO both support it; e.g. abort after 1 day, comfortably past the 2h part-URL validity). Action: add it to RUNBOOK.md's storage section and the install docs' bucket-setup step; consider wiring it into the MinIO provisioning in the compose/installer path so self-hosters get it by default.

Pre-existing; unrelated to the #1202 validation fix beyond being found while enumerating its lifecycle. Not a v1.8.0 blocker.

Refs #1202.

PR #1401, opened 2026-08-11
fix(ingest): reconcile orphaned presigned-upload objects against tracking rows

### LINKED ISSUE #1249: ops(ingest): presigned-upload finalization has a residual orphan window under unenforced slow transfers
## Context

PR #1243 (#1236) closed the post-expiry sweep orphan bug through 4 rounds of Codex review, converging on a fixed finalization margin derived from S3's own 5GiB single-PUT ceiling (`_S3_SINGLE_PUT_MAX_BYTES`), after earlier rounds proved that scaling the margin from a client-declared `file_size` was unsound (nothing enforces that declared size against the real transfer).

Round 5 found the residual gap in that fixed-margin approach itself:

> "When a single-part PUT is accepted just before URL expiry and then transfers below 32 KiB/s, this finite margin can elapse while S3 or MinIO is still receiving the request; the subsequent sweep or retention purge deletes the key and retires its only tracking row, after which the PUT completes and recreates a permanently untracked object... the change replaces the previous duration guess with a larger one but still enforces neither that rate nor a maximum request duration. Keep the row recheckable until the provider guarantees termination, or enforce a hard transfer bound."

## Why this is a follow-up, not a round-6 patch

Any *time-based* margin, however generous, is theoretically defeatable by an arbitrarily slow (but still actively transferring) upload — this isn't a parameter to tune, it's a structural property of inferring "the transfer must be done by now" from elapsed time alone. Closing it for real needs one of:

1. **Verify-before-delete**: before the sweep/retention purge finalizes a row, do an actual S3/MinIO `HEAD` check to positively confirm the object doesn't exist yet, rather than inferring from elapsed time. Closes the gap unconditionally, at the cost of the sweep/purge making synchronous calls to the storage provider.
2. **A hard transfer bound enforced by the provider**, analogous to what #1211/#1237 (`dd11c47fd`) already did for abandoned multipart uploads via a bucket lifecycle policy. **Note: that existing mechanism does NOT cover this case** — it aborts *incomplete* multipart upload sessions; this bug is about a single-part PUT that *completes* after its tracking row is gone, which is a different problem (an orphan-object reconciliation gap, not an abandoned-upload gap). A bucket-level fix here would likely need a periodic bucket-listing sweep that reconciles real objects against known tracking rows, not a lifecycle rule.

Blast radius is a leaked storage object (cost/cleanup issue), not data loss or a security exposure — unlike the RUNBOOK retention findings (#1248) which risked deleting real audit data. Practically requires a sustained sub-32KiB/s transfer racing a URL right at its expiry boundary, a narrow window. Accepted as residual risk for PR #1243 rather than blocking Wave A on an architecture change.

ishiland:
Closes #1249.

## The residual gap

#1236 review r5 accepted a known hole and this PR is the follow-up it named. Every staging reaper in the tree starts from a job ROW and asks what object it owns — `_reap_committed_staged_paths`, `_sweep_expired_presigned_staging`, the two ingest task tails. An object that no row references at all is invisible to all of them.

The case that produces one: S3 validates a presigned signature when a request **starts**, not when it finishes transferring. An accepted single-part PUT can still be writing after every deadline derived from its URL has passed. The sweep deletes the key, the retention purge retires the row, and then the PUT lands — recreating an object with no tracking row anywhere.

## Why reconciliation and not another margin

`_RECHECK_TRANSFER_MARGIN_SECONDS` already took four rounds to settle, and the issue concedes the direction is structurally defeatable: any margin is a guess at a transfer rate nothing enforces (`generate_presigned_put_url` signs `ContentType`, never a content-length constraint). Widening it makes the guess rarer, not correct.

A HEAD-before-delete has the symmetric problem: it proves the object was absent when the check ran, not that it stays absent while the delete is in flight. It moves the race one step later.

Reconciliation removes the class. It never asks "is this upload probably done" — it asks a question with a durable answer, *is there a row that owns this key*, and an object whose answer is no is unreachable by any future PUT: a presigned URL for `staging/{job_id}/…` is only ever minted alongside the `ingest_jobs` row with that id, and the retention purge deliberately keeps such a row alive past `MAX_PRESIGNED_URL_LIFETIME_SECONDS + _RECHECK_TRANSFER_MARGIN_SECONDS` (`presigned_url_may_still_be_live`). It also catches orphans from causes no margin ever addressed — a best-effort delete that failed, a row purged while its object delete errored, a worker killed between writing an object and committing the row that names it.

Note on #1211/#1237: MinIO's stale-uploads sweep aborts *incomplete multipart sessions*. This is a single-part PUT that **completes**, so no lifecycle rule reaches it. The issue said a bucket-level fix would need a listing sweep that reconciles objects against tracking rows. That is what this is.

## Race analysis

Two races, both handled by construction rather than by ordering (there is no atomic delete-if-still-unreferenced across a database and an object store):

**1. An object uploaded after the listing snapshot.** Only objects whose last-modified is older than `staging_orphan_min_age_seconds` are candidates, and the age is **re-read from the provider immediately before the delete**. A listing entry is never authority for destroying anything — an object created or overwritten since the snapshot fails the re-read and is skipped.

**2. A row that lands after the batch row query.** Row absence is re-checked per object, immediately before that object's delete, in its own statement. Under READ COMMITTED each statement takes a fresh snapshot, so the recheck sees rows committed since the batch query rather than replaying it. It is a Core `select` over the id column, never an ORM load, so the identity map cannot answer from a stale instance.

Both rechecks fail **closed** — gone, rewritten, ambiguous, or unreadable all skip the object and leave it for the next pass. The cost of a missed pass is a day of leaked bytes.

Two further boundaries:

- **S3 storage only** (r8). `staging/` is the upload system's exclusive namespace in a *bucket*, and only in a bucket. On the local backend the same prefix sits inside `upload_staging_dir`, where an operator also stages manifest seed files — a local manifest source spelled `staging/{anything}/seed.geojson` resolves to an absolute path under that directory whenever the directory's basename is not itself `staging`, which no logical-key comparison here would match. Those files are operator-owned and legitimately staged *before* the manifest that names them is applied; deleting one is not a leaked byte, it is someone's input. Nothing is lost by declining — presigned uploads refuse anything but S3 at request time, so the orphan class this addresses cannot occur on local or Azure at all, and the s3:// manifest branch closes the mirror-image hole by refusing `staging/` keys outright.
- **Nothing outside `staging/` is ever listed**, so an object the upload system does not own cannot reach the delete path at all. A key under `staging/` whose owning job cannot be named (not `staging/{uuid}/…`) is counted and left alone: "I cannot tell whose this is" never becomes "so delete it".
- **Multi-tenant**: the prefix is resolved through `resolve_current_storage_key`, so a pass can only list its own tenant's namespace, and a call with no tenant context declines rather than sweeping the bucket while RLS shows it no rows (same fail-closed posture as `_stale_generation_storage_keys`).

## Ownership is "any row that still needs the key" (review r3, r4, r6)

A P1, and the case that found it is fan-out. `create_fan_out_jobs` clones the parent's `file_path` onto every child, so a child's only input is the **parent's** `staging/{parent_id}/frozen/...` object. The parent row reaches the retention cutoff on its own schedule — the purge's survivor query is what keeps the *object* alive for a still-pending, running, or retryable-failed child, not the parent's row. Reconciling on `IngestJob.id` alone looked up the purged parent, found nothing, and would have deleted the input the child was about to read.

The question is now *does any row still need this key*, which turns out to be two questions with two answers (r4 found the second half — the sibling of r4's own fix):

- **The row whose id the key names** is not a *reference* to the key, it is its *lifecycle*, so the question is not "does it exist" but "is anything still going to act on this key". Existence alone was a permanent shield, and r6 found it shielded the exact leak this change exists to close: once `_sweep_expired_presigned_staging` sets its final marker it never looks at that key again, and if the row is its dataset's latest complete job the retention purge exempts it indefinitely — so a PUT landing after that final delete recreated an object every row-driven reaper was finished with, and this reconciler then declined forever. `_owner_still_manages` now asks for a live claim: either the row can still consume the bytes, or it carries an `s3_key` the post-expiry sweep has not yet finalized. The `s3_key IS NOT NULL` half matters as much as the marker half — a direct upload never presigns, so no marker is ever set on it, and reading a missing marker as "not finalized" would shield those rows forever instead.
- **Any other row** counts only while it can still consume the bytes. Unconditional was wrong in the direction that never self-corrects: a fan-out child stays `complete` forever and is exempt from retention indefinitely as a dataset's latest complete job, so its inherited `file_path` would answer "still referenced" on every future pass and an object leaked by a failed parent-side delete could never be repaired — the reconciler would permanently decline the one case it exists for.

That is the same line the retention purge's survivor query already drew, so both now read `STATUSES_NEEDING_STAGED_INPUT` from `jobs/models.py` rather than each holding its own tuple. Two copies of "which statuses still need the bytes" drift into a leak in one direction and a deletion of live input in the other. `STAGING_REAPED_FINAL_MARKER` moved there for the same reason once the reconciliation became its second reader.

The per-page pre-filter is two queries built from the same two predicates the per-object recheck uses, not a Python re-derivation of them: a pre-filter that shields *more* than the recheck silently skips candidates the recheck never sees, which is how both r4 and r6 stayed hidden.

`user_metadata->>'s3_key'` is deliberately **not** consulted: it is cloned onto fan-out children wholesale, `owned_presigned_staging_key` already settles that an inherited copy is not ownership, and the only row for which it *is* ownership is the one whose id the key names. Comparisons are against the **logical** key, since `file_path` persists the tenant-agnostic form while the provider reports physical keys.

## Concurrency and the counter

The pass takes a `pg_try_advisory_xact_lock` before doing anything. Without it every API worker's sweeper loop would list the same prefix and count the same delete, reporting N times the truth under `UVICORN_WORKERS>1` — the fabricated-number class #1240 existed to remove. It also keeps the provider LIST to one per interval per tenant.

The lock lives on a **session of the pass's own**, not the caller's. A rollback expires every instance in the identity map regardless of `expire_on_commit`, and `fail_stale_jobs` still holds ORM objects when it calls this — an earlier draft that released the lock by rolling back the caller's session broke `test_presigned_staging_reap_1202.py`. The caller's session is now neither committed nor rolled back, and a test pins that.

## Bounded work per pass (review r1)

The first revision loaded the whole `staging/` prefix before applying any budget. Two P2s with one root: unbounded memory and lock/connection occupancy to attempt at most 200 deletions, and an expanding `IN` that could exceed the driver's bind-parameter ceiling on a large backlog — a failure the best-effort wrapper then swallowed on every subsequent pass, so the backlog could never drain.

`StorageProvider.list_objects` is now `iter_object_pages`, yielding one provider page at a time (ListObjectsV2 continuation on S3, `by_page()` on Azure, a single page for the local walk). A consumer that stops iterating stops the round trips with it, and a test asserts exactly that against moto rather than only asserting the results.

Both budgets are checked **between pages**, so the pass stops asking for more rather than merely stopping acting:

- `_MAX_DELETES_PER_PASS = 200` bounds the lock hold once there **is** work. On an unhealthy prefix this fires inside the first page.
- `_MAX_OBJECTS_SCANNED_PER_PASS = 20_000` bounds the walk when there is **not**, so a prefix of a million tracked-and-old objects is not paged through in full every cycle to delete nothing.

The live-row query moved inside the page loop: its id set is bounded by the provider's page size, so it can never approach the driver's parameter ceiling no matter how large the backlog grows.

## Scan cursor (review r2)

Two more P2s, both about the bound r1 introduced.

The local provider still returned the whole prefix as one page — an unbounded page rebuilds the same oversized `IN` the paging was meant to prevent. It now sorts by key and yields fixed-size pages, matching the ListObjectsV2 default so a per-page budget behaves identically on every backend.

And a scan cap without a cursor is a blind spot, not a bound: every pass re-walked the same lexicographically first window, so an orphan sorting after a window full of tracked, recent, or unattributable keys was never reached at all. `iter_object_pages` gains `start_after` (S3's own `StartAfter`; filtered client-side on the other two, neither of which can hold a presigned staging object anyway), and the pass records where it stopped:

- A budget stop advances the cursor; a completed walk clears it, so the next pass starts at the front.
- When the **delete** budget is what stopped the pass, the resume point is the last candidate actually *processed*, not the end of the page — otherwise candidates it never reached would be skipped until the wrap.

A follow-up round (r5) found the local walk was still eager behind that slicing — it built and sorted every matching object before yielding the first page, so a budget stop bounded the SQL batches but neither memory, filesystem traversal, nor lock occupancy. It is now a lazy generator: one directory level materialized at a time, any subtree sorting entirely at or before `start_after` skipped without being entered, and `stat` called only for entries a page actually needs. Directory entries sort as `name + "/"` so the sequence matches full-key lexicographic order — plain name order disagrees whenever a directory and a file share a stem, and `frozen/` sits next to a caller-chosen filename in every presigned staging layout.

The scan is **circular** (r7): the pass walks from the cursor to the end of the prefix, then spends whatever budget remains walking the front of the prefix back up to the cursor, bounded there so it ends instead of re-walking the first leg. A one-directional scan from a random start is a sample, not a rotation — a worker recycled before its next pass would only ever see suffixes, and an orphan near the front of a large prefix would be reached with probability ~`1/(objects+1)` per restart. With the wrap, one unbudgeted pass covers the whole prefix from wherever it begins.

That is what makes the cursor a starting **offset** rather than a window, and why process-local state seeded at a random point is enough: every pass is independently correct — the cursor never decides what may be deleted, only what is looked at first — so a lost cursor costs re-examining keys that were already cheap to skip.

## Changes

- `backend/app/platform/jobs/staging_reconcile.py` (new): the pass.
- `backend/app/platform/jobs/sweep.py`: called from `fail_stale_jobs` after the two row-driven reapers. Deliberately not folded into `StaleCleanupOutcome` — that dataclass is a published API and audit shape several callers reconstruct field by field.
- `backend/app/platform/jobs/router.py`: same call on the admin cleanup endpoint's single-tenant branch (the multi-tenant branch goes through `fail_stale_jobs` per tenant already).
- `backend/app/platform/storage/{provider,s3,azure,local}.py`: new `StorageProvider.iter_object_pages` yielding pages of `StoredObject(key, last_modified)`. `list` alone cannot age an object, and there is no head-with-timestamp call; a complete key is a valid prefix, which is how the pre-delete re-read works on every backend. `local.py`'s glob logic is now shared between `list` and the new call rather than duplicated.
- `backend/app/observability/metrics/jobs.py`: `geolens_staging_orphans_deleted_total`, incremented once per delete, strictly after the provider call returns. Per delete rather than once per pass (r5): the objects are already gone when a pass dies partway, and nothing later can recover a count for something that no longer exists.
- `backend/tests/test_layering.py`: `sweep.py` 1366 → 1380, `config.py` 1091 → 1104, both exact with comments. The new module is 340 lines, under the 1000-line inclusion threshold.

## Config surface

`.env.example` gains the matching entry — `scripts/check_env_doc_drift.py` (the Installer Tag Resolution job) requires every operator-facing Settings field to be documented there, and the backend suite does not cover it.

One new setting, `STAGING_ORPHAN_MIN_AGE_SECONDS` (`staging_orphan_min_age_seconds`), default `86400`, floored at `3600`. It is **not** an estimate of how long an upload takes — the row check is what decides ownership; this only has to be far enough past a listing that no object is reported old while its own tracking row is still being written. The floor exists so a misconfiguration cannot turn the sweep into a deleter of objects whose uploads are still landing. No schema change, no migration, no API change.

## Verification

- Full backend suite, `uv run pytest -n 4 -q` — **8396 passed, 92 skipped, 0 failed** (re-run after each review round)
- Targeted: `tests/test_staging_orphan_reconcile_1249.py` (43), `tests/test_presigned_staging_reap_1202.py`, `tests/test_storage.py`, `tests/test_storage_azure.py`, `tests/test_local_storage_containment.py`

Tests are deterministic — no sleeps, no bounded waits. Ages are facts each test writes (`os.utime` for the real-provider case, an injected `now=` everywhere). Negative controls run and reverted, twice: disabling the per-object row recheck and the pre-delete age re-read fails exactly the two race tests and nothing else; narrowing the reference clause back to `IngestJob.id` fails exactly the three fan-out tests; dropping the status restriction fails exactly the terminal-child test; making the owner shield unconditional again fails exactly the finally-reaped and direct-upload tests; dropping the wrap leg fails exactly the two circular-scan tests. Every guard here is load-bearing rather than decorative.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -715,6 +715,13 @@ FRONTEND_PORT=8080
 # Type: integer (minimum 0) | Default: 30
 # INGEST_JOBS_RETENTION_DAYS=30
 
+# How old an object under the upload staging prefix must be before the
+# reconciliation sweep deletes it for having no ingest-job row that still needs
+# it. Not an estimate of how long an upload takes — the row check decides
+# ownership; this only has to clear a listing snapshot. Floored at an hour.
+# Type: integer (minimum 3600) | Default: 86400
+# STAGING_ORPHAN_MIN_AGE_SECONDS=86400
+
 # --- Analysis Budgets ---
 # Statement timeout for the analysis materialize CTAS — the CREATE TABLE AS that
 # builds a buffer/dissolve/clip/centroid output. The default covers roughly 150k
--- a/backend/app/core/config.py
+++ b/backend/app/core/config.py
@@ -366,6 +366,19 @@ class Settings(BaseSettings):
     # 0 disables the purge (keep history forever).
     ingest_jobs_retention_days: int = Field(default=30, ge=0)
 
+    # fix(#1249): how old an object under the `staging/` prefix must be before
+    # the reconciliation sweep will delete it for having no ingest_jobs row.
+    # Not a guess at how long an upload takes — the row check is what decides
+    # whether an object is owned, and this only has to be far enough past a
+    # LISTING that no object can be reported old while its own tracking row is
+    # still being written. A day is orders of magnitude past that and costs
+    # nothing but a day of leaked bytes in the rare orphan case.
+    # Floored at an hour so a misconfiguration cannot turn the sweep into a
+    # deleter of objects whose uploads are still landing.
+    # Consumer: `reconcile_orphaned_staging_objects` in
+    # platform/jobs/staging_reconcile.py.
+    staging_orphan_min_age_seconds: int = Field(default=86400, ge=3600)
+
     # ---------------------------------------------------------------------------
     # Outbound Notification channels (Phase 1229 NOTIF-02 / NOTIF-03 / NOTIF-05)
     # ---------------------------------------------------------------------------
--- a/backend/app/observability/metrics/jobs.py
+++ b/backend/app/observability/metrics/jobs.py
@@ -37,6 +37,23 @@ jobs_failed_total = Counter(
     ["queue"],
 )
 
+# fix(#1249): staging objects deleted because no ingest_jobs row tracks them.
+# A true counter rather than a polled gauge, and safe as one for a reason
+# worth stating (the concern refresh.py's module docstring raises): the
+# reconciliation pass runs under a `pg_try_advisory_xact_lock`, so at most one
+# process per interval deletes — and therefore counts — any given object,
+# where a poll-and-increment design would report N times the truth under
+# UVICORN_WORKERS>1. Incremented only after the provider delete returns, so
+# the number counts completed deletions, not intentions.
+#
+# The series matters most when it is non-zero and STAYS non-zero: a steady
+# trickle means something is leaking staging objects faster than one-off
+# incidents explain.
+staging_orphans_deleted_total = Counter(
+    "geolens_staging_orphans_deleted_total",
+    "Staging objects deleted for having no ingest-job row tracking them",
+)
+
 # Track previous snapshot for counter delta computation.
 # NOTE(#655): on the first cycle after boot this is empty, so the counters seed
 # with the full historical procrastinate_jobs row counts — raw counter values
--- a/backend/app/platform/jobs/models.py
+++ b/backend/app/platform/jobs/models.py
@@ -35,6 +35,17 @@ from app.core.db import Base
 # sweep.py) and the staging-orphan reconciliation both read it.
 STATUSES_NEEDING_STAGED_INPUT = ("pending", "running", "failed")
 
+# `user_metadata` key the post-expiry presigned sweep sets once it has finished
+# with a row's `s3_key` for good — see `_sweep_expired_presigned_staging` in
+# sweep.py, which owns the whole story of when it may be written.
+#
+# fix(#1249 review r6): it lives here because a SECOND reader now depends on
+# it. Its presence is the fact that says "the row-driven reaper will never look
+# at this key again", which is exactly when the staging-orphan reconciliation
+# may take the key over — and a copy of the string in each module is one
+# rename away from a row that shields an object forever.
+STAGING_REAPED_FINAL_MARKER = "s3_key_reaped_final"
+
 
 def owned_presigned_staging_key(
     job_id: uuid.UUID | str,
--- a/backend/app/platform/jobs/models.py
+++ b/backend/app/platform/jobs/models.py
@@ -20,6 +20,22 @@ from sqlalchemy.orm import Mapped, mapped_column
 from app.core.db import Base
 
 
+# Statuses whose row still NEEDS the staged input its `file_path` points at:
+# pending and running are reading it now, and failed keeps it for
+# /jobs/{id}/retry (a failed-only endpoint). Every other status is done with
+# the bytes — a `complete` fan-out child keeps its metadata row but not a claim
+# on the shared original, which is why a successful fan-out's input is reapable
+# at all rather than pinned forever by children that are each a dataset's
+# latest complete job.
+#
+# fix(#1249 review r4): lives here rather than inline at either consumer,
+# because two of them now ask the same question and a drift between them is a
+# leak in one direction and a deletion of live input in the other. The
+# retention purge's survivor query (`_reap_committed_staged_paths`'s feeder in
+# sweep.py) and the staging-orphan reconciliation both read it.
+STATUSES_NEEDING_STAGED_INPUT = ("pending", "running", "failed")
+
+
 def owned_presigned_staging_key(
     job_id: uuid.UUID | str,
     user_metadata: dict[str, Any] | None,
--- a/backend/app/platform/jobs/router.py
+++ b/backend/app/platform/jobs/router.py
@@ -36,6 +36,7 @@ from app.platform.jobs.schemas import (
     ReservedRenameWarning,
     StaleCleanupResponse,
 )
+from app.platform.jobs.staging_reconcile import reconcile_orphaned_staging_objects
 from app.platform.jobs.sweep import (
     JOB_TIMEOUT_SECONDS,
     STALE_PENDING_BOUND_MESSAGE,
@@ -195,6 +196,11 @@ async def cleanup_stale_jobs(
             publish_refresh_reconciliation(outcome)
             outcome = await _reap_committed_staged_paths(outcome)
             outcome = await _sweep_expired_presigned_staging(db, outcome)
+            # fix(#1249): same object-driven reconciliation the background
+            # sweeper runs. The multi-tenant branch needs nothing here — the
+            # fleet helper runs fail_stale_jobs with its own commit per
+            # tenant, and that path already reconciles in tenant context.
+            await reconcile_orphaned_staging_objects(db)
             details = outcome.as_dict()
     except Exception as exc:  # broad: cleanup spans DB and artifact deletion
         await db.rollback()
--- a/backend/app/platform/jobs/staging_reconcile.py
+++ b/backend/app/platform/jobs/staging_reconcile.py
@@ -1,0 +1,351 @@
+"""Reconcile real staging objects against ingest-job rows (fix #1249).
+
+Every other staging reaper in this codebase starts from a ROW and asks what
+object it owns. This one starts from the OBJECT and asks whether any row still
+owns it — the only direction that can see an object no row references at all.
+
+### Why reconciliation rather than a wider margin
+
+#1235/#1236 closed the presigned-upload orphan window twice by widening the
+window a sweep waits before it may act, and review round 5 of #1236 found the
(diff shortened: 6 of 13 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

Issue #1241, opened 2026-08-06
feat(builder): "Save as dataset" for chat query previews (snapshot via existing ingest path)


ishiland:
## Problem

An AI chat query (e.g. "show me the latest earthquake" in the dataset chat) produces an ephemeral preview in the builder layer stack, and the only action on it is dismiss. The "Save as dataset..." link on the preview row exists but is wired only for analysis-backed previews (`MapBuilderPage.tsx` wires `onSaveAsDataset` from `ephemeralResult.analysis`, feat #675). Plain chat query results have no persistence path at all.

## Proposal

Extend the existing "Save as dataset..." affordance to chat query previews. Frontend only:

- The full `FeatureCollection` is already on the client in `ephemeralResult.geojson`. Wrap it in a `File` (`new File([JSON.stringify(geojson)], "<name>.geojson")`) and push it through the existing ingest flow in `frontend/src/api/ingest.ts` (`uploadFile()` -> preview -> `commitImport()`).
- Land the user in the existing import-commit dialog so naming and collection choice reuse that UI. Prefill the suggested title from the chat prompt, same pattern as the analysis prefill in `AnalysisPanel.tsx`.
- The resulting dataset is a normal dataset: it tiles, filters, and adds to maps like any other.

## Truncation guard

Chat previews can be server-truncated (`result.truncated`). Saving the client copy would persist "first N of M" as if it were the complete result, which is the same misrepresentation #674/#1076 fixed in the count label. Offer the save only when `!truncated`; when capped, render it disabled with the "N of M" explanation.

## Out of scope

A server-side "materialize this query" endpoint (re-running the generated SQL into a new table) is the only way to persist truncated results, but it would store LLM-generated SQL as a dataset source: a new security-sensitive surface (Rule 1 coverage, SQL review, a new job type). Deferred until capped-result saves demonstrably matter.

## Semantics note

This is a snapshot: a saved answer, frozen at save time. For predicate-shaped requests the live alternative (a persisted layer filter via `set_filter`) already exists; discoverability of that path is a separate issue.

PR #1402, opened 2026-08-11
feat(ai): offer the layer-filter path after predicate chat queries

### LINKED ISSUE #1242: feat(ai): offer the layer-filter path after predicate-shaped chat query results
## Problem

When a chat query result is a simple row predicate ("earthquakes above magnitude 5"), the durable form of that answer already exists: a persisted layer filter via the `set_filter` tool, which saves with the map, costs no storage, and stays live as the dataset refreshes. Nothing tells the user this. They get an ephemeral preview and, once #1241 lands, a snapshot save. Both are worse than a filter for this query shape.

## Proposal

Prompt/copy change only. Extend the verb-class routing guidance (the fix(#549) area in `backend/app/processing/ai/tools.py` and the system prompt it points to) so that after a `query_data` result that is expressible as a row predicate, the assistant offers a follow-up along the lines of "want this as a filter on the layer instead? It will persist when you save the map."

## Constraints

- This must stay an offer the user accepts, never automatic routing. #549 is the record of "show me the..." silently landing on a persistent layer filter, and that was deliberately reverted. Read-shaped questions must not mutate saved map state on their own.
- No SQL-to-MapLibre-expression compiler. The model already knows whether the request it just served was a predicate; the offer comes from routing guidance, not from translating the generated SQL back into an expression.
- Applies only where a target layer exists (builder chat with the dataset on the map). The dataset-page chat has no layer to filter.

## Relationship

Complements #1241: predicate queries get the live-filter suggestion, everything else (top-N, aggregates, joins, "latest X") persists via snapshot. Neither path subsumes the other.

ishiland:
## What changed

Extends the builder-chat system prompt (`build_chat_system_prompt` in `backend/app/processing/ai/chat_service.py`) so that after a `query_data` result that reads as a simple row predicate (no aggregate, top-N/ranked list, or multi-layer join), the assistant offers to persist it as a layer filter instead: "Want this as a filter on the layer instead? It persists when you save the map." This is prompt and routing-guidance only. There is no SQL-to-MapLibre-expression compiler: the model decides from the shape of the request it just served.

`backend/app/processing/ai/tools.py` gets a pointer comment only (no description-string change) in the `query_data` #549 area, documenting that the offer is response-shaping and lives in the system prompt, not a second tool-description classification site.

## Behavior impact

- The offer is additive text in the assistant's response, never an automatic tool call. It stays an offer the user must accept, matching the #549 precedent that a read-shaped question (`query_data`) must never silently mutate saved map state (that's what `set_filter` does, and it fires only if the user accepts in a follow-up turn).
- Scoped to builder chat with an editable map: gated on `can_edit`, because a read-only viewer never has `set_filter` in its tool set (`select_chat_tools`), and the readonly note already tells the model it cannot filter. Offering it there would be a promise the model can't keep.
- Dataset-page chat (`build_dataset_chat_system_prompt`) is untouched: no map, no layer to filter.
- Complements #1241 (snapshot save): predicate-shaped queries get the live-filter suggestion; everything else (top-N, aggregates, joins, "latest X") persists via snapshot. Neither path subsumes the other.

## Verification

- `cd backend && uv run ruff check .`
- `cd backend && uv run ruff format --check .`
- `cd backend && set -a && source ../.env.test && set +a && uv run pytest tests/test_chat_narrative.py tests/test_ai_chat_dataset.py tests/test_ai_chat_verb_routing_549.py tests/test_sql_engine.py -v` (107 passed)
- `cd backend && set -a && source ../.env.test && set +a && uv run pytest tests/test_layering.py -q` (40 passed; chat_service.py's ratcheted cap raised 456 -> 485 in the same commit)

Closes #1242

### DIFF
--- a/backend/app/processing/ai/chat_service.py
+++ b/backend/app/processing/ai/chat_service.py
@@ -155,6 +155,8 @@ def build_chat_system_prompt(
     When ``can_edit`` is False the caller may view but not edit the map (the AI
     is given read-only tools — see ``select_chat_tools``); a read-only directive
     is injected so the model declines edit requests instead of silently failing.
+    ``can_edit`` also gates the query_data filter-offer note (#1242): only an
+    editor has set_filter available to fulfil it.
     """
     # Cap layers to prevent unbounded prompt growth
     display_layers = layers[:_MAX_SYSTEM_PROMPT_LAYERS]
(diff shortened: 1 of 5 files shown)


ishiland:
@codex review

PR #1403, opened 2026-08-11
feat(builder): save a chat query preview as a dataset

### LINKED ISSUE #1241: feat(builder): "Save as dataset" for chat query previews (snapshot via existing ingest path)

ishiland:
## What changed

Extends the preview row's "Save as dataset…" to plain AI-chat query previews, which previously had no persistence path (the affordance was wired only for analysis-backed previews, #675).

The result's FeatureCollection is already client-side, so this adds no API surface. Clicking the link opens a dialog that reuses the import page's `ImportMetadataForm` (name, description, visibility), and submitting wraps the collection in `new File([JSON.stringify(geojson)], "<name>.geojson")` and runs it through the existing ingest flow in `api/ingest.ts`: `uploadFile()` -> `previewFile()` -> `commitImport()`. The preview step is kept because it is where an unreadable or over-budget payload is rejected with an actionable message, before a job is queued. What lands is an ordinary dataset with no special casing downstream.

The suggested name comes from the chat prompt, which now rides along in the `onQueryResult` meta (`ChatPanel` -> `use-ephemeral-layers`), mirroring how `AnalysisPanel` prefills its save form from a chat handoff. It is handed to the form as a *file* name (`"<prompt>.geojson"`), because the form seeds its title from `stripExtension(defaultName)` and that cuts at the last dot: a prompt like "buildings over 3.5m" only survives intact this way. The file's own name is separately bounded by UTF-8 bytes, since a multibyte prompt within the grapheme cap can still blow the filesystem's 255-byte limit.

### Truncation guard

A truncated preview is not saveable. Saving the client copy would file "first N of M" in the catalog as if it were the complete answer, which is the misrepresentation #674/#1076 closed on the count label, except a dataset outlives the session. Both preview surfaces render the affordance **disabled** with the reason stated and the "N of M" count still visible, rather than dropping it silently.

Review turned up that `truncated` alone does not answer the question. It is the SQL sandbox's row-cap flag, and `chat_actions.py` slices the overlay to its own 50-row render budget *after* the sandbox returns, so a 300-row answer arrives as 50 features with the flag false. `chatOverlayCompleteness` (new, in `lib/`) compares what the client holds against `row_count` and calls anything short of it clipped, whoever clipped it. All three chat producers go through it, so the count label and the save guard cannot disagree. The dataset-chat handoff carried geometry only, so it now carries that pair and the prompt too.

The #675 analysis handoff is deliberately exempt from the guard: it re-runs the operation server-side and materializes the complete result, so a capped *preview* says nothing about what it would save. `previewSaveMode()` in `components/builder/ephemeral-preview.ts` is the single place that decides, and is unit-tested directly. It allowlists rather than blocklists: a preview with a stamped `source` (the Analysis panel's own, today) is never snapshotted, since that surface owns its own save flow and its preview is bbox-scoped.

### Permissions

The snapshot needs `upload` **and** `export`. It hands the caller a durable, caller-owned copy of source attributes they may only be able to read, which is why analysis materialize requires both (fix(#692), `router_analysis.py`); the ingest endpoints ask only for `upload` because their file normally comes from the caller's own machine.

### Retry

The commit chain has three server steps and the dialog stays open when one fails, so the staged job survives and a retry resumes from the failed step instead of re-uploading. A repeat commit is safe (the worker claims a job with one conditional UPDATE, fenced on attempt_id), and a 400 on a re-commit is reconciled against the job's actual status: only a running or complete job counts as evidence the import exists. A dead job (dispatch failure marks it `failed` and deletes its staged file) is abandoned so the next attempt starts fresh.

## Backend change

One producer fix, no new endpoint: `_safe_value` in `processing/ai/chat_geojson.py` now emits integers outside JavaScript's safe range as strings. The browser's `JSON.parse` rounds them (9007199254740993 arrives as ...992), which was always visible in the chat table and becomes permanent once a snapshot is serialized from the parsed payload. The function already stringifies `Decimal`, `UUID`, and `datetime` for the same reason. `chat_geojson.py`'s cap in `_MODULE_LOC_CAPS` is raised 350 -> 370 in the same commit.

## UI impact

- Layer stack preview row: "Save as dataset…" now appears for plain chat previews; on a truncated preview it renders disabled with a muted explanation line beneath, wired to the button via `aria-describedby` (a disabled button gets no hover or focus, so a title tooltip would reach nobody).
- Builder below 1100px (`EphemeralBadge`, the rail's preview surface): same treatment, so the two surfaces cannot drift.
- New dialog: title, snapshot disclaimer, and the import metadata form. Lazy-loaded and mounted only while open.
- Public viewer: unchanged affordances (it never passes a save handler), but its count label now discloses a clipped overlay.

Screenshots not captured from this worktree, since the dev stack serves the main checkout; the orchestrator can capture them.

## i18n

Five new `builder` keys (`savePreview.*`) plus `ephemeralBadge.saveTruncatedReason`, added to all four locales (en/es/fr/de). No plural-suffix keys.

## Verification

```
cd frontend
npm ci
npm run typecheck                     # clean
npm run lint                          # clean
npx vitest run                        # 388 files, 4834 tests passed
npm run test:i18n                     # passed
npm run check:i18n:toast-strings      # passed
npm run check:i18n:api-errors         # passed
npm run check:i18n:changed            # builder.json present in all locales
npm run test:coverage                 # passed

cd backend                            # host run, .env.test sourced
uv run ruff check / ruff format --check   # clean
uv run pytest tests/test_ephemeral_geojson.py tests/test_collect_chat_action.py \
              tests/test_ai_chat_actions_phase1135.py -q   # 102 passed
uv run pytest tests/test_layering.py -q                    # 40 passed
```

No schema, API, or config changes. No Playwright run: this worktree's dev stack serves the main checkout, so an e2e result here would not describe this branch.

Closes #1241

### DIFF
--- a/backend/app/processing/ai/chat_geojson.py
+++ b/backend/app/processing/ai/chat_geojson.py
@@ -257,10 +257,28 @@ def _detect_geom_column(columns: list[str], rows: list[list]) -> int | None:
     return None
 
 
+# JavaScript numbers are IEEE-754 doubles, so an integer beyond this magnitude
+# cannot round-trip through the browser's JSON.parse.
+_JS_MAX_SAFE_INT = 2**53 - 1
+
+
 def _safe_value(v: object) -> object:
-    """Convert non-JSON-serializable types to str; pass through primitives."""
-    if v is None or isinstance(v, (str, int, float, bool)):
+    """Convert values the client cannot represent to str; pass through the rest.
+
+    fix(#1241 codex r5): an integer outside JavaScript's safe range is
+    JSON-serializable and still lossy — 9007199254740993 arrives in the browser
+    as 9007199254740992, because JSON.parse rounds it to the nearest double.
+    That silently wrong id was always on screen; it becomes permanent now that
+    the map builder can save a chat preview as a dataset, since the snapshot is
+    serialized from the parsed payload. Emitting the exact digits as a string
+    is the only shape that survives the trip. Every smaller integer stays a
+    number, so ordinary ids keep their type.
+    """
+    if v is None or isinstance(v, (str, float, bool)):
         return v
+    if isinstance(v, int):
+        # bool is an int subclass and already returned above.
+        return str(v) if abs(v) > _JS_MAX_SAFE_INT else v
     if isinstance(v, (datetime, date, Decimal, bytes, memoryview, UUID)):
         return str(v)
     return str(v)
--- a/backend/tests/test_ephemeral_geojson.py
+++ b/backend/tests/test_ephemeral_geojson.py
@@ -148,6 +148,19 @@ class TestSafeValue:
     def test_bytes_converted(self):
         assert _safe_value(b"raw") == str(b"raw")
 
+    # fix(#1241 codex r5): the browser's JSON.parse rounds an integer past
+    # 2**53-1 to the nearest double, so the exact digits only survive as a
+    # string. The builder can now save a chat preview as a dataset, which
+    # writes whatever the client parsed straight into it.
+    def test_int_beyond_js_safe_range_converted(self):
+        assert _safe_value(9007199254740993) == "9007199254740993"
+        assert _safe_value(-9007199254740993) == "-9007199254740993"
+
+    def test_int_at_js_safe_boundary_stays_numeric(self):
+        # 2**53-1 is exactly representable — an ordinary id must keep its type.
+        assert _safe_value(9007199254740991) == 9007199254740991
+        assert _safe_value(-9007199254740991) == -9007199254740991
+
 
 class TestExtractGeojson:
     def test_wkb_rows_to_feature_collection(self):
--- a/backend/tests/test_layering.py
+++ b/backend/tests/test_layering.py
@@ -1739,6 +1739,13 @@ def test_decomposed_service_modules_stay_within_size_budgets() -> None:
         # chat_*.py glob below, ~142 lines under the 350 default) rather than
         # grown here. Cap 530 → 550 (~14 headroom).
         "backend/app/processing/ai/chat_actions.py": 550,
+        # feat(#1241): +18 over the 350 default — _safe_value now emits
+        # integers outside JavaScript's safe range as strings (constant, the
+        # int branch, and the docstring explaining why), so a bigint id
+        # survives the browser's JSON.parse intact instead of arriving rounded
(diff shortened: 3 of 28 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1404, opened 2026-08-11
feat(demo): add the Hurricane Exposure showcase item and flip three maps to globe

### LINKED ISSUE #1214: demo: flip the three global showcase maps to globe projection
The three global showcase maps on the demo (Restless Earth, Everything That Fell From the Sky, Hurricane Alley) render in Mercator, which distorts exactly the stories they tell: plate boundaries, global meteorite distribution, and Atlantic storm tracks all read better on a globe. MapLibre GL v5 ships globe projection natively, so this is a map-config change, not an engine change.

Scope:

- Flip the three maps' projection to globe and QA each one in the builder and the viewer.
- QA gates before touching the demo: `basemap_config` is additive-or-422 (an unknown field rejects the whole config — verify the projection key round-trips through save/load on both editors), heatmap and cluster layers behave at the antimeridian and poles, and the ETOPO raster layer on Restless Earth drapes correctly.
- The Matterhorn and Manhattan maps stay Mercator — regional maps gain nothing from the globe.

Not in scope: making globe a default for new maps. That's a product decision with its own tradeoffs.

### LINKED ISSUE #1215: demo: add a Hurricane Exposure showcase map (buffer + intersect + provenance)
Nothing on the demo showcases the analysis operations shipped across v1.5-v1.7 (buffer, intersect, provenance-tracked derived datasets). Every showcase map is a styling/rendering story; a visitor evaluating the analysis features has nothing to look at.

Proposal: a "Hurricane Exposure" showcase map built from data already on the demo:

- Buffer the Category 3+ segments of the Hurricane Alley tracks (e.g. 100 km).
- Intersect the buffers with coastal admin polygons to produce an exposure layer, styled by how many distinct major storms touched each polygon.
- The derived dataset's provenance panel then shows the real operation chain — that's the actual feature being showcased, the map is the vehicle.

Seeding: extend `scripts/seed-showcase.py` with a `hurricane-exposure` item so demo resets rebuild it, same pattern as the existing showcase items.

Follow-on candidate for the demo README/marketing copy once live: it's the first showcase map whose headline is an analysis result rather than a rendering.

ishiland:
Two demo changes to `scripts/seed-showcase.py`, both verified end to end against a live stack.

## #1214 — globe projection on the three global maps

Restless Earth, Everything That Fell From the Sky and Hurricane Alley move to the globe. Manhattan, the Matterhorn and New York From Orbit stay Mercator — a globe buys a city or a massif nothing.

The set of globe maps is written down once (`GLOBE_PROJECTION_MAPS`) and applied by a post-builder pass (`apply_globe_projection`), the same shape as the existing `enrich_showcase_metadata`. Not set at map creation, deliberately: builders skip a map that already exists, so a creation-time setting would reach a fresh instance and never an already-seeded one — which is every instance that matters, the live demo included.

The pass reads the stored `basemap_config` and PUTs it back with one key changed. That is load-bearing rather than defensive: `PUT /maps/{id}` dumps the whole submodel with no `exclude_unset`, so a bare `{"projection": "globe"}` body returns 200 and silently resets `label_mode`, `opacity`, `background_color` and `sublayer_overrides` to their defaults. A map already on the globe is skipped entirely.

## #1215 — Hurricane Exposure showcase item

A seventh hero map, and the first whose headline is a computed result rather than a rendering. Three real provenance-tracked operations through `POST /datasets/{id}/analysis/materialize/`:

```
Cat 3+ legs     --buffer 100 km-->        corridors        (198, one per storm)
basin regions   --intersect corridors-->  exposed pieces   (1,343 region x storm)
exposed pieces  --dissolve by region-->   EXPOSURE         (289 regions)
```

The dissolve is what turns pair rows into the map's number. Intersect emits exactly one row per (region, corridor) pair, and one corridor is one storm, so the generated `source_count` counts **distinct** major storms per region, while the union collapses that region's overlapping pieces into the single coastal footprint the graduated fill is drawn on.

The map: exposure polygons on top graded by `source_count` (4 classes), the Category 3+ legs that generated them underneath, camera on the basin.

### Shape notes the API forced

- **No attribute filter exists on analysis**, on preview or materialize. "The Category 3+ segments" therefore has to be its own dataset before it can be buffered.
- The corridors are the intersect **overlay**, capped at 1,000 features, so the legs are merged one-per-**storm** (198) rather than one-per-leg (1,880). One-per-leg would also have made a region's row count "legs that reached me" rather than "storms that reached me".
- Intersect **refuses two layers sharing any column name**, so the admin-1 name lands as `region`, never `name` (the overlay side carries the HURDAT2 track columns).
- `region` is the dissolve key and dissolve groups on one column, so repeated names are qualified by country (50 of 480 regions, e.g. `La Paz (Honduras)`).
- Materialize registers its output **private**, and provenance is redacted per requester (`visible_derived_from` drops `derived_from` whole when any dataset in the chain is invisible). Every link is published — one private intermediate blanks the panel for the anonymous visitor the map exists for.

### Why Natural Earth admin-1 at 1:10m

At 1:50m Natural Earth subdivides only **nine** large countries (verified against the pinned file: 294 features, 9 countries), so the Caribbean and Central America — the heart of hurricane alley — have no admin-1 polygons at all. The 1:10m file is a ~39 MB one-off download (1.2 s from the GitHub CDN) that filters to 480 basin regions across 47 countries and ~3 MB ingested.

### Also

- The HURDAT2 record parser is now shared between the two feeds (`_hurdat2_storms`) rather than copied — the format is positional and undocumented in the file.
- `poll()` treats `cancelled` as terminal. It previously waited out the full timeout for a job that had already stopped.

## Verification

Run against the local dev stack (`localhost:8080`, v1.11.1) with the real functions imported from the script, not re-implementations.

**Offline, no API** — feed preconditions:

```
198 storms, 1880 Cat 3+ legs, 98 KB      PASS storms <= 1000 (overlay cap)
480 regions, 47 countries, 3040 KB       PASS region values unique
source (regions):  ['country', 'region']
overlay (corridors): ['landfall', 'major_legs', 'name', 'peak_category',
                      'peak_wind_kt', 'season']
                                         PASS no shared column names
                                         PASS dissolve by_field 'region' present
hurdat2_feed after the parser refactor:  198 storms / 9522 segments (unchanged)
```

**Live, #1214** — `apply_globe_projection` against the seeded maps, with a non-default `basemap_config` written first so the run had something to lose:

```
PASS  Restless Earth: projection='globe'
PASS  Everything That Fell From the Sky: projection='globe'
PASS  Hurricane Alley - 75 Years of Major Atlantic Storms: projection='globe'
PASS  sibling basemap_config keys preserved
      (label_mode=hidden, opacity=0.62, background_color=#101820 all survived)
PASS  Manhattan / Matterhorn / New York From Orbit / Private Embed: projection=None
PASS  second run is a no-op ("already on the globe")
PASS  basemap-only PUT leaves layers, camera and visibility untouched
      (Restless Earth 7 layers -> 7, Hurricane Alley 2 -> 2, meteorites 1 -> 1)
```

A `label_mode` outside the enum 422'd during setup, which incidentally re-proved the additive-or-422 behaviour on the same request.

**Live, #1215** — `build_hurricane_exposure` end to end:

```
buffer    -> 198 features   derived_from operation=buffer, distance_meters=100000.0
intersect -> 1343 features  derived_from operation=intersect, mask_dataset_id=...
dissolve  -> 289 features   derived_from operation=dissolve, by_field=region
            columns ['region', 'source_count']
all five datasets visibility=public; provenance readable ANONYMOUSLY on each
lineage_summary e.g. 'Dissolved from "Coastal Regions Inside a Major Hurricane
  Corridor" by region, created by admin on 2026-08-11.'
map: 2 layers, public, anonymous read 200, visibility-check has_non_public=False
second builder run returns "(skipped)"; a re-run reused all five datasets
```

Counts are right on their face — top regions by distinct major storms: Florida 23, Quintana Roo 16, Louisiana 15, Yucatán 13, Tamaulipas 11, Texas 10; 191 of the 480 offered regions were never reached and correctly dropped. The class breaks were then retuned against that measured distribution (`[3, 6, 10]` splits it 87 / 83 / 113 / 6) rather than left at a guess.

Anonymous signed MVT tiles return real content, so the map draws for a logged-out visitor:

```
Florida       z5/8/13: 200  9623 bytes application/vnd.mapbox-vector-tile
Quintana Roo  z5/8/14: 200  9475 bytes
Louisiana     z5/7/13: 200  5275 bytes
basin         z3/2/3:  200 25887 bytes
```

Persisted layer style read back anonymously:

```
fill-color: ["step", ["to-number", ["get", "source_count"], 0],
             "#fed976", 3, "#feb24c", 6, "#f03b20", 10, "#800026"]
style_config: graduated / source_count / manual / breaks [3, 6, 10]
```

**Gates:**

```
python3 -m py_compile scripts/seed-showcase.py
node scripts/check-readme-locales.mjs
python3 scripts/check_public_surface.py
python3 scripts/check_env_doc_drift.py
python3 scripts/check_docs_contract.py
cd backend && uv run python -m pytest ../tests/ -q      # 34 passed, 55 subtests
cd backend && uv run ruff check ../scripts/seed-showcase.py
cd backend && uv run ruff format --check ../scripts/seed-showcase.py
```

`scripts/` is outside the backend ruff scope in CI, and the file already carried four `C901` complexity findings and one long line on `main`; this change adds none of either and leaves both counts unchanged.

## Docs

`scripts/README.md` gains the new `--only hurricane-exposure` choice, the post-builder passes, and the analysis item's extra requirements (`upload` + `export` capabilities, slower than the upload-only items). The four root READMEs said "six showcase maps" — updated to seven in all of them, since the count is now wrong in each.

## Review rounds

Two findings, both the same shape and both fixed at the helper rather than the call site: a name/title lookup that several callers had always treated as a key, which this PR made **mutating** for the first time.

**Round 1 — `list_maps` resolved to an arbitrary duplicate.** A plain `name -> id` comprehension keeps whichever row the page ended on. `--force` creates a *second* map with the same name rather than replacing the first, and an admin token sees every user's maps. Harmless while callers only asked "does this exist"; not harmless once `apply_globe_projection` projects what it resolves. `list_maps` now keeps the first match, scopes to this account, sorts by `created_at` (the endpoint's `updated_at` default is mutable — the pass's own PUT reshuffles it), and paginates. `_map_exists` and `prune` had the same exposure and are fixed with it.

Verified live by reproducing the case: created a duplicate `Hurricane Alley` map, then touched the *older* one so `updated_at` ordering would prefer it — `list_maps` still resolved to the newest, and the projection pass flipped that one.

**Round 2 — `datasets_by_title` had the sibling gap.** It already kept the newest match (fix(#389)) but never asked whose. All three callers mutate: `_get_or_analyze` publishes the dataset it resolves and rewrites its summary before building the chain on it, `enrich_showcase_metadata` overwrites its license, `prune` deletes it. `list_datasets_full` is now `list_own_datasets`, filtered on `created_by` against the id from `/auth/me/`.

The pagination counter had to change with it — the loop used `len(out)` as its rows-seen counter, which is only the same number while nothing is filtered out. Proven offline with a stubbed client over three pages of interleaved foreign rows, including an entirely-foreign final page and a duplicate name spanning a page boundary:

```
PASS  list_own_datasets: 300 of 450 rows, pages requested at skip=[0, 200, 400]
PASS  list_maps keeps the NEWEST duplicate across a page boundary
PASS  list_maps returns exactly the owned names (299 of 450)
PASS  all-foreign final page still terminates (400 kept, skip=[0, 200, 400])
```

The two lookups scope on different keys because the endpoints expose different ones: datasets carry the owner's uuid (`created_by`), maps only the display name (`created_by_username`). The identity comes from `GET /auth/me/`, not the `--username` argument — a spelling that differs from the stored username would match nothing, and "none of my content exists" is what makes every builder rebuild from scratch.

Both fixes re-verified against the live stack with no regression in the #1214 and #1215 checks above.

## Note for landing

This PR is the seed-script side only. Applying it to the live demo — flipping the three production maps and running `--only hurricane-exposure` there — is a post-merge step for the orchestrator. Two things to know when doing it:

1. The demo's existing maps are picked up by the projection pass automatically; no `--force` needed.
2. The dev instance also carries a `Restless Earth (Live)` map, which is deliberately **not** in `GLOBE_PROJECTION_MAPS` (the constant names exactly the builder's map names). If the demo has a similarly renamed variant, it will not be flipped.

Closes #1214
Closes #1215

### DIFF
--- a/README.de.md
+++ b/README.de.md
@@ -265,7 +265,7 @@ Die [CLI-Anleitung](https://docs.getgeolens.com/guides/cli/) dokumentiert das ge
 
 ### Beispieldaten
 
-`scripts/seed-showcase.py` erstellt sechs Beispielkarten aus öffentlichen offenen Daten: globale Tektonik über echtem Meeresbodenrelief, Manhattans nach Bauzeit gefärbte 3D-Skyline (Titelbild), 75 Jahre atlantischer Hurrikanbahnen, gruppierte Meteoritenfälle, das Matterhorn als 3D-Gelände aus 2-m-Lidar und referenzierte Sentinel-2-Bilder von New York:
+`scripts/seed-showcase.py` erstellt sieben Beispielkarten aus öffentlichen offenen Daten: globale Tektonik über echtem Meeresbodenrelief, Manhattans nach Bauzeit gefärbte 3D-Skyline (Titelbild), 75 Jahre atlantischer Hurrikanbahnen, gruppierte Meteoritenfälle, das Matterhorn als 3D-Gelände aus 2-m-Lidar, referenzierte Sentinel-2-Bilder von New York und eine Karte der Hurrikan-Exposition, die direkt aus den Sturmbahnen mit Puffer, Verschneidung und Auflösung berechnet wird:
 
 ```bash
 pip install httpx
--- a/README.es.md
+++ b/README.es.md
@@ -266,7 +266,7 @@ Consulta la [guía de la CLI](https://docs.getgeolens.com/guides/cli/) para ver 
 
 ### Datos de muestra
 
-`scripts/seed-showcase.py` construye seis mapas de muestra a partir de datos abiertos públicos: una historia tectónica global sobre el relieve real del fondo oceánico, el horizonte 3D de Manhattan coloreado por época de construcción (la imagen principal anterior), 75 años de trayectorias de huracanes del Atlántico, caídas de meteoritos agrupadas, el Matterhorn como terreno 3D de lidar a 2 m e imágenes Sentinel-2 de Nueva York por referencia:
+`scripts/seed-showcase.py` construye siete mapas de muestra a partir de datos abiertos públicos: una historia tectónica global sobre el relieve real del fondo oceánico, el horizonte 3D de Manhattan coloreado por época de construcción (la imagen principal anterior), 75 años de trayectorias de huracanes del Atlántico, caídas de meteoritos agrupadas, el Matterhorn como terreno 3D de lidar a 2 m, imágenes Sentinel-2 de Nueva York por referencia y un mapa de exposición a huracanes calculado sobre las propias trayectorias con área de influencia, intersección y disolución:
 
 ```bash
 pip install httpx
--- a/README.fr.md
+++ b/README.fr.md
@@ -265,7 +265,7 @@ Le [guide de la CLI](https://docs.getgeolens.com/guides/cli/) présente le sché
 
 ### Données d’exemple
 
-`scripts/seed-showcase.py` construit six cartes de démonstration à partir de données ouvertes : une histoire tectonique mondiale sur le relief réel des fonds océaniques, la silhouette 3D de Manhattan colorée par époque de construction (visuel principal ci-dessus), 75 ans de trajectoires d’ouragans atlantiques, des chutes de météorites regroupées, le Cervin en terrain 3D lidar à 2 m et des images Sentinel-2 de New York référencées :
+`scripts/seed-showcase.py` construit sept cartes de démonstration à partir de données ouvertes : une histoire tectonique mondiale sur le relief réel des fonds océaniques, la silhouette 3D de Manhattan colorée par époque de construction (visuel principal ci-dessus), 75 ans de trajectoires d’ouragans atlantiques, des chutes de météorites regroupées, le Cervin en terrain 3D lidar à 2 m, des images Sentinel-2 de New York référencées et une carte d’exposition aux ouragans calculée à partir des trajectoires elles-mêmes par zone tampon, intersection et fusion :
 
 ```bash
 pip install httpx
--- a/README.md
+++ b/README.md
@@ -302,11 +302,12 @@ See the [CLI guide](https://docs.getgeolens.com/guides/cli/) for the full manife
 
 ### Seed data
 
-`scripts/seed-showcase.py` builds six showcase maps from public open data: a global
+`scripts/seed-showcase.py` builds seven showcase maps from public open data: a global
 tectonics story over real ocean-floor relief, the Manhattan 3D skyline colored by
 construction era (the hero above), 75 years of Atlantic hurricane tracks, clustered
-meteorite falls, the Matterhorn in 2 m lidar 3D terrain, and by-reference Sentinel-2
-imagery of New York:
+meteorite falls, the Matterhorn in 2 m lidar 3D terrain, by-reference Sentinel-2
+imagery of New York, and a hurricane-exposure map computed in place from the storm
+tracks with buffer, intersect and dissolve:
 
 ```bash
 pip install httpx
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -6,16 +6,20 @@ Utility scripts for GeoLens administration and data seeding.
 
 ### `seed-showcase.py`
 
-Builds the six showcase hero maps from public, openly licensed data against a
+Builds the seven showcase hero maps from public, openly licensed data against a
 running stack — Restless Earth (quakes + volcanic eruptions + plate boundaries
 over ETOPO global relief), Manhattan (3D skyline by construction era + the MTA
 subway), The Matterhorn (3D lidar terrain), Hurricane Alley (75 years of major
 Atlantic storms from HURDAT2), Everything That Fell From the Sky (clustered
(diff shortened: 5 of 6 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1405, opened 2026-08-11
fix(ops): reach tenant-ownership adoption forward-only at the head schema

### LINKED ISSUE #998: fix(ops): make 0019 tenant-ownership adoption reachable without a destructive downgrade

ishiland:
## What changed

`backend/app/core/db/tenant_adoption.py` (plus `tenant_adoption_sql.py`, which
holds the DDL ported from migration 0019) reconstructs tenant schema/role
ownership **at the current head schema**, so the destructive
`alembic downgrade 0016` is never required.

```bash
docker compose run --rm --no-deps -e DATABASE_URL_OVERRIDE="<migrator-url>" \
  migrate sh -c "uv run --no-dev python -m app.core.db.tenant_adoption --apply"
```

Without `--apply` it is a read-only report and exits non-zero while work is
pending, so it doubles as a post-restore check.

It is deliberately not a replay of 0019:

- **No function body is ever installed.** That is what made replaying 0019
  impossible on a restored dump — 0019 uses plain `CREATE FUNCTION` and collides
  with the functions the dump already carries. Adoption verifies both boundary
  functions are present, `SECURITY DEFINER`, and `search_path`-pinned, then
  repairs only what `pg_restore --no-owner --no-acl` strips. Bodies stay owned
  by the migrations; a missing or unexpected-shape function is a refusal.
- **Relations go straight to the per-tenant writer.** 0019 parked them on the
  provisioner so its provisioning function could rewrite their ACLs; 0024
  removed that pass, so at head the writer owns them and grants the paired
  reader `SELECT` itself — the contract ingest already follows.
- **Column-owned sequences are skipped in the transfer.** PostgreSQL rejects
  `ALTER SEQUENCE … OWNER TO` on a `serial`/identity sequence and moves it with
  its table instead. 0019's unconditional loop would have stopped dead on the
  `ogc_fid` serial that every `ogr2ogr` vector ingest produces; verified against
  a `bigserial` fixture.

Idempotence is keyed on database state, not on a marker or a timestamp: each
step reads what the cluster holds and issues DDL only for the gap, so an
already-adopted tenant issues none at all. Every tenant runs in its own
transaction, so an interrupted run resumes by re-running it, and a tenant that
refuses is named in the report while the rest continue.

## Security impact

A `--no-owner --no-acl` restore leaves both SECURITY DEFINER boundary functions
owned by the restoring login (typically the superuser) **with the PostgreSQL
default function ACL, which grants `EXECUTE` to `PUBLIC`**. Confirmed on a real
`pg_dump`/`pg_restore` roundtrip at PostgreSQL 18:

```
provision_tenant_data_schema | geolens | secdef=t | proacl=NULL | has_function_privilege('public',…)=t
```

Until adoption runs, every login in a restored database can call
`catalog.provision_tenant_data_schema` as the definer. Adoption re-owns both
functions to `geolens_tenant_provisioner`, revokes `PUBLIC`, and re-grants
`EXECUTE` to `geolens_tenant_control` only.

## The tenant boundary is introspected, not enumerated

The report reads the boundary from the live `trg_stamp_current_tenant_on_insert`
triggers, and cross-checks it against `app.core.db.rls.RLS_TABLES` — a table
stamped in the database but missing from that tuple never gets RLS enabled at
boot. Measured on a fresh head-schema database:

- **9 stamped tables** (the live boundary): `audit_logs`, `collections`,
  `datasets`, `embed_tokens`, `ingest_jobs`, `maps`, `oauth_accounts`,
  `records`, `users`. Migration 0018's frozen `_TABLES` tuple names six.
- `dataset_refresh_runs` carries a `tenant_id` column with no trigger and no
  policy — 0037's deliberately dormant column. It is reported as
  "tenant_id column only, outside the stamped boundary" so an accidental
  omission cannot hide behind an intentional one.

## Operational impact — DR procedure change

`RUNBOOK.md` § "Multi-tenant role reconstruction after a fresh-cluster restore":

- Step 2 is now restore-the-dump + run the adoption command. The old 2c
  (`alembic downgrade 0016 && alembic upgrade heads`) is gone.
- The tenant-attribution snapshot and restore that used to bracket that
  downgrade (old 2b/2d, there because 0022's `downgrade()` discards `tenant_id`
  on `audit_logs`/`ingest_jobs`) are gone with it. Nothing on this path
  discards `tenant_id`.
- Step 1's manual role-creation block is now a fallback: adoption creates the
  five fixed NOLOGIN cluster roles itself, with 0019's validation.
- The downgrade table is kept under a new "Do not reach for
  `alembic downgrade 0016`" heading, reframed as the thing not to do rather than
  the recipe. Honest limits are unchanged and a "What it does not do" paragraph
  was added: runtime login grants, RLS enablement (still the API's job at boot),
  and object storage are all still separate.

No schema, API, or config changes. `alembic check` reports no drift.

## Verification

```bash
# new tests (10) — DB-backed, host Postgres at localhost:5434
uv run pytest tests/test_tenant_ownership_adoption.py -q          # 10 passed

# gates
uv run ruff check . && uv run ruff format --check .               # clean, 968 files
uv run pytest tests/test_layering.py -q                           # 40 passed
uv run pytest tests/test_rule1_structural.py tests/test_rule2_structural.py -q
uv run pytest -n 4 -k "tenant or rls or runtime_db or migration or docs or alembic" -q
#                                                                 # 714 passed, 1 skipped

# schema drift, on a host DB migrated to head
uv run alembic check                                              # No new upgrade operations detected
```

Also exercised end to end by hand against a head-schema database:
`pg_dump -Fc --no-owner --no-acl` → `pg_restore --no-owner --no-acl` into a
second database → `--apply`. The tenant schema, table, view, and `bigserial`
sequence came back owned by the restoring login with no ACLs; after adoption the
schema is owned by `geolens_tenant_provisioner`, every relation by
`geolens_writer_t_<uuid>`, the reader holds `SELECT` (and `SELECT, USAGE` on the
sequence), the provisioner holds ADMIN-only non-SET membership in both
per-tenant roles, and the transaction-scoped writer edge the transfer needs is
handed back before commit. A second run left a full catalog snapshot
byte-identical.

Closes #998

### DIFF
--- a/RUNBOOK.md
+++ b/RUNBOOK.md
@@ -726,8 +726,10 @@ What it does, walking `catalog.tenants` with each tenant in its own transaction:
   act as, a `SECURITY DEFINER` or untrusted-language routine in a tenant schema
   — is reported with the exact statement to run and the role to run it as, and
   that tenant is left for the next run. The automatic membership PostgreSQL
-  gives a role's creator is the one exception it tolerates rather than reports:
-  nobody can revoke it, and it confers nothing on its own.
+  gives a role's creator is tolerated only for the login running adoption. On
+  any other login it is refused — `ADMIN` alone lets that login grant itself a
+  usable edge — and the report names the remedy: `DROP ROLE` the retired
+  login, or revoke the membership as its recorded grantor.
 
 The end state is the one migration 0019 produced. Idempotence is keyed on
 database state rather than on a marker or a timestamp: a tenant already in that
--- a/RUNBOOK.md
+++ b/RUNBOOK.md
@@ -643,9 +643,44 @@ writer/sandbox/tile SET-only shapes adoption requires. Adoption refuses a wrong
 shape rather than rewriting somebody else's grant, so leaving one here stops 2d
 before it starts. Re-issuing them is a no-op when they already match.
 
-These are the grants `.env.example` documents alongside
-`GEOLENS_RUNTIME_DB_ROLE`; substitute your own login names, and never give the
-tile login any of the first three:
+First clear every membership row the replay restored, whatever grantor it
+recorded. Re-granting on top does not rewrite a restored row — on PostgreSQL
+16+ it adds a second row under your grantor and the old default-INHERIT/SET
+row survives to make 2d refuse the topology. This block is version-aware and
+safe to run on any supported release:
+
+```bash
+docker compose exec -T db psql -U "$POSTGRES_USER" -d "$POSTGRES_DB" <<'SQL'
+DO $$
+DECLARE m RECORD;
+BEGIN
+  FOR m IN SELECT granted.rolname AS granted_name,
+                  member.rolname AS member_name,
+                  grantor.rolname AS grantor_name
+           FROM pg_catalog.pg_auth_members AS am
+           JOIN pg_catalog.pg_roles AS granted ON granted.oid = am.roleid
+           JOIN pg_catalog.pg_roles AS member ON member.oid = am.member
+           LEFT JOIN pg_catalog.pg_roles AS grantor ON grantor.oid = am.grantor
+           WHERE granted.rolname IN ('geolens_tenant_control',
+                                     'geolens_tenant_writer',
+                                     'geolens_tenant_sandbox',
+                                     'geolens_tile_gateway')
+  LOOP
+    IF current_setting('server_version_num')::int >= 160000
+       AND m.grantor_name IS NOT NULL THEN
+      EXECUTE format('REVOKE %I FROM %I GRANTED BY %I',
+                     m.granted_name, m.member_name, m.grantor_name);
+    ELSE
+      EXECUTE format('REVOKE %I FROM %I', m.granted_name, m.member_name);
+    END IF;
+  END LOOP;
+END $$;
+SQL
+```
+
+Then issue the canonical grants — these are the ones `.env.example` documents
+alongside `GEOLENS_RUNTIME_DB_ROLE`; substitute your own login names, and
+never give the tile login any of the first three:
 
 ```bash
 docker compose exec -T db psql -U "$POSTGRES_USER" -d "$POSTGRES_DB" <<'SQL'
--- a/RUNBOOK.md
+++ b/RUNBOOK.md
@@ -505,12 +505,11 @@ For managed/external Postgres there is no `db` container to exec into: drop the
 `docker compose exec -T db` prefix and pass the provider's `-h`, `-p`, and `-U`
 to `pg_dumpall`/`psql` directly.
 
-If no globals dump exists, create the five fixed NOLOGIN groups by hand. This
-must happen **before** the step 2 commands: the downgrade passes through 0024,
-whose `downgrade()` re-installs the provisioning function with
-`OWNER TO geolens_tenant_provisioner`, so it fails on a cluster where that role
-is missing. The block is idempotent; run it as the migrator (needs CREATEROLE),
-and the attributes match what 0019 creates and validates.
+If no globals dump exists, step 2 creates the five fixed NOLOGIN groups itself,
+so you can skip ahead. The equivalent by hand, for a cluster where you want them
+in place first or where step 2 refused on the existing topology — it is
+idempotent, needs CREATEROLE, and the attributes match what 0019 creates and
+validates:
 
 ```sql
 DO $$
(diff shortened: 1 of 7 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

On the "compare a normalized definition or versioned hash" finding for the boundary function bodies: declining deliberately, and the reasoning is in `BoundaryFunctionState.migration_shaped`. Pinning the migration's exact text (or a hash of it) in app code creates a drift trap in a module whose whole job is to still run years after a restore — the next migration that edits a body silently breaks disaster recovery. The markers are now matched against the body with SQL comments stripped, so the specific mechanism named (markers hiding in comments) is closed.

What bounds the residual exposure, and why this is a detection gap rather than an escalation: a `--no-owner --no-acl` restore leaves both functions owned by the restoring login — normally the superuser — with `EXECUTE` to `PUBLIC`. Adoption only ever moves them to a NOLOGIN role and narrows execution to `geolens_tenant_control`. It can fail to notice a tampered body; it cannot raise one's privilege. The isolation *policy* is compared exactly, because there the trade goes the other way: one expression, unchanged since 0006, and it is the cross-tenant visibility rule itself.

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

Repair-path findings from the last several rounds are answered by a boundary rather than another mechanism, stated in both module docstrings and in the commit above.

**Adoption rewrites only the grants it is itself the grantor of.** A pre-existing anomaly — a membership a third party granted, a duplicate row behind a canonical one, a default-privilege entry owned by a role this credential cannot assume — is now detected and refused with the exact statement to run and the role to run it as, and the tenant is left for the next run. It is not repaired.

That is deliberate, not a gap. Removing somebody else's grant means naming its grantor and being able to assume it, and PostgreSQL answers that differently for every combination of server version, grantor and dependent grant: `GRANTED BY` exists only from 16, a plain `REVOKE` silently removes nothing when the caller is not the grantor, `CASCADE` is required once the row has dependents, and impersonating the grantor is precisely the authority a least-privilege migrator does not have. Each of those was a real defect found in the repair path over the last few rounds, which is the argument for not having one. The operator has authority this process does not, and a refusal naming the remedy costs one re-run.

What still repairs in place, because adoption is the grantor or the owner: its own temporary edges, the provisioner membership it grants when one is missing or mis-shaped, and ACLs on objects it owns or can own for the duration of the run.

Two consequences worth flagging for this pass: an orphan tenant role is refused from a gateway whatever its membership options are, and the runbook's fixed-group grants are no longer gated on having rebuilt the roles by hand.

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

Issue #1407, opened 2026-08-11
chore(frontend): migrate AttributeTable to @tanstack/react-table v9


ishiland:
Dependabot proposed the v8 to v9 major in #1338 (9.1.1). That PR is closed by hand and the major is ignored in `.github/dependabot.yml` (#1408; two `@dependabot ignore` comments went unacked, so the ignore lives in config) rather than migrated: v9.1.0 only reached `latest` on npm on 2026-08-07, it shipped three patches in the following three days, and the v9 docs are still served from a branch named `beta`. There is no security pressure (zero open npm dependabot alerts at the time of closing) and 8.21.3 works. The major needs a hand migration, so it should happen deliberately once 9.x settles instead of riding in on a dependency bump.

## Scope

One component uses the library: `frontend/src/components/dataset/AttributeTable.tsx`, plus its test at `src/components/dataset/__tests__/AttributeTable.test.tsx`. Nothing else in the frontend imports `@tanstack/react-table`. `@tanstack/react-virtual` is a separate package and is not part of this.

## What changes at 9.x

From the CI failures on #1338 plus the v9 docs:

- `useReactTable(options)` becomes `useTable(options, selector)`. The second argument is new; `(state) => state` reproduces v8 re-render behavior.
- The `getCoreRowModel()` / `getSortedRowModel()` options are gone. v9 takes an explicit `features` option built with `tableFeatures({...})`, registering `rowSortingFeature`, the visibility feature, and `sortedRowModel: createSortedRowModel()`. Confirm exact export names against the v9 docs at migration time.
- Core types gain a leading `TFeatures` generic: `ColumnDef<typeof features, TData>`, and the same for `Table`, `Row`, and `Cell`.
- The `VisibilityState` type no longer exists under that name; find its v9 replacement.
- `flexRender(def, ctx)` becomes the `<table.FlexRender header={...} />` / `<table.FlexRender cell={...} />` component.
- v9 requires sort functions to be registered explicitly via `sortFns`; v8 auto-detected them. This is the risky part for us: AttributeTable builds its columns dynamically from the dataset's Postgres column list and relies on that auto-detection, so per-type sorting behavior needs live verification, not just a green typecheck.
- The `react-hooks/incompatible-library` eslint suppression sits on the `useReactTable` call. Recheck it against the renamed hook; if the rule does not know `useTable`, the suppression becomes an unused disable, which our lint config flags.

## Verification

[...]

PR #1408, opened 2026-08-11
chore(deps): ignore @tanstack/react-table majors pending hand migration

ishiland:
Adds an ignore rule to the frontend npm block of `.github/dependabot.yml`: no more major-version bump PRs for `@tanstack/react-table`. The backend block already runs this policy for all majors.

Why: v9 reached npm latest on 2026-08-07 and shipped three patches in its first three days. Our one consumer, `AttributeTable.tsx`, needs a hand migration with live re-verification of its sorting and virtualizer behavior; #1407 tracks that work with the full API map. The bump PR (#1338) is being closed by hand because two `@dependabot ignore this major version` comments went unacked today while dependabot was otherwise active in the repo, so the comment channel was not a reliable place to park this decision.

Config impact: dependabot automation only. No app code, no schema, no API surface.

Verification: YAML parse check on the edited file (`ruby -ryaml`).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -36,6 +36,11 @@ updates:
         patterns: ["react", "react-dom", "@types/react*"]
       tanstack:
         patterns: ["@tanstack/*"]
+    ignore:
+      # fix(#1407): react-table v9 is a hand migration (AttributeTable);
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review

PR #1420, opened 2026-08-12
fix(frontend): refetch refresh runs on window focus so external runs are observed

### LINKED ISSUE #1328: design(frontend): discover externally-started refresh runs on an already-open dataset page
Split out of PR #1323 (codex round 6, P2). The dataset page's refresh-run watcher (from #1285) polls while a non-terminal run is cached and invalidates all dataset-derived caches on the active→terminal transition — covering runs dispatched from this page and runs already active at page load. What it deliberately does not cover: a refresh started by ANOTHER actor (CLI, another editor) AFTER the page cached an idle runs state. With `refetchOnWindowFocus` globally disabled, that external run is never observed and the page can stay stale until remount.

Not fixed in #1323 because the obvious patch — a permanent slow idle poll on every open dataset page — trades a quiet edge case for standing background request load on every viewer of every dataset, and the right mechanism is a design choice:

- (a) slow idle poll (30-60s) only for users with edit rights on refreshable-origin datasets (bounds the population);
- (b) piggyback on an existing signal (e.g. refetch runs whenever the user returns focus to the tab — would mean re-enabling focus refetch for just this query);
- (c) server push (SSE/websocket) — the general answer, but new infrastructure;
- (d) accept the staleness (a page-level "data may have changed, reload" affordance).

Recommendation: (b) as the cheap 90% answer — a per-query `refetchOnWindowFocus: true` override on the refresh-runs query is one line and catches the realistic scenario (user comes back to the tab after a colleague/CLI refreshed).

Refs: #1285 / PR #1323 (the watcher and its invalidation set), #1227 (CLI refresh — the external actor).

ishiland:
## What

Adds a per-query `refetchOnWindowFocus: 'always'` override to `useDatasetRefreshRuns` (`frontend/src/components/dataset/hooks/use-dataset.ts`), the query behind both the dataset page's refresh-run watcher (`useDatasetRefreshWatch`) and the Source panel's refresh history section. Returning focus to a dataset page tab now refetches its refresh runs every time, regardless of whether the cached data is still within the query's 15s staleTime, instead of only ever refetching on remount or on this client's own dispatch.

## Why

The app disables `refetchOnWindowFocus` globally in `main.tsx`; most queries have no business refetching on every alt-tab. That default means a refresh dispatched from the CLI or another editor's session, after this page has already cached an idle "no active run" state, is never observed. Nothing invalidates or refetches the runs query, so the page can serve stale data indefinitely until the route unmounts and remounts.

This is option (b) from #1328: piggyback on the window-focus signal instead of adding a standing idle poll (option a) or new push infrastructure (option c).

The override has to be `'always'`, not plain `true`. Plain `true` only refetches when the cached data is already stale, but an external run can start and this tab can regain focus inside the same 15s window as this query's last fetch. That fresh-window case is the realistic "user checks back after a colleague ran a refresh" scenario this fix targets, not a rare corner case. Worse, once that window passes, becoming stale is passive: nothing schedules a fetch on its own since `refetchInterval` is off in the idle state, so the page would need a second, unrelated focus event or a remount to notice. `'always'` makes every focus event the trigger, independent of this query's own freshness clock.

## Face 2 coverage

The issue's follow-up comment describes a second face of the same problem. `useDatasetRefreshWatch`'s active-to-terminal transition guard only fires invalidation when a run was actually observed going from active to terminal by the current mount. A run whose entire active phase happens with no observer watching (route navigation unmount and remount, most obviously) surfaces on the next fetch as already terminal, which the guard ignores, so data-derived caches like rows, versions, and attributes fall back to their own staleTime to self-correct.

This change covers most of that in practice, but not all of it. A run that this mount saw as active before the tab lost focus, whether its own dispatch or an external run already active on mount, still gets caught correctly on refocus. `wasActiveRef` is a ref, not component state, so it survives a focus/blur cycle untouched; only a true unmount resets it. Once the refocus refetch reports that run terminal, `wasActive` is still `true` from before the blur, the transition fires, and the full invalidation sweep runs.

What's still missed is a run whose entire lifecycle, dispatch and completion both, happens while the tab is unfocused, with this mount never having observed it active. On refocus the refetch reports it already terminal in one jump. The transition guard has nothing to compare against, so no invalidation fires. The refresh-runs list itself is still correct, since the refetch did happen, so the history section and the busy/disabled gate stay accurate; the rest of the data-derived caches just wait out their own staleTime, same as before this change. Using `'always'` instead of `true` doesn't change this part: it only changes when the refetch happens, not what the transition guard does with the result once it arrives.

Navigating away from the dataset page and back is unaffected either way. A remount's initial fetch is governed by `refetchOnMount` and staleTime, not `refetchOnWindowFocus`, so this change doesn't touch that path. It was already accepted as bounded and self-healing when #1323 deferred this issue, and stays exactly as it was.

Per the issue's own framing, a persisted watermark would close the remaining gap for good, but that's more than this issue asks for.

## Verification

- `cd frontend && npm ci`
- `npm run typecheck`
- `npm run lint`
- `npx vitest run src/components/dataset/hooks/__tests__/use-dataset.test.ts`
- `npx vitest run src/components/dataset/` (full directory, checked for regressions in SourcePanel/SourceRefreshAction, which share this query)

Closes #1328

### DIFF
--- a/frontend/src/components/dataset/hooks/__tests__/use-dataset.test.ts
+++ b/frontend/src/components/dataset/hooks/__tests__/use-dataset.test.ts
@@ -1,8 +1,8 @@
 import { renderHook, waitFor } from '@/test/test-utils';
-import { act } from '@testing-library/react';
+import { act, renderHook as renderHookRTL } from '@testing-library/react';
 import { vi } from 'vitest';
-import { useRef } from 'react';
-import { useQueryClient, type QueryClient } from '@tanstack/react-query';
+import { createElement, useRef, type ReactNode } from 'react';
+import { useQueryClient, QueryClient, QueryClientProvider, focusManager } from '@tanstack/react-query';
 
 vi.mock('@/api/datasets', async (importOriginal) => {
   const actual = await importOriginal<typeof import('@/api/datasets')>();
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

ishiland:
@codex review

Issue #1372, opened 2026-08-10
Raster tile shared nginx cache can serve stale bytes for up to 1h after a replace


ishiland:
## Context

While building raster reupload UI (#1289, PR #1362), a codex review flagged that a completed raster replacement can leave stale tiles visible. #1362 fixed the parts that are fixable from the frontend:

- Remounts the hero map's MapLibre raster source when a replace completes (it was never re-added on prop change).
- Appends the dataset's `updated_at` as a `?v=` query param to the raster tile URL, which busts the **browser's own** HTTP cache (the raster tile response carries `Cache-Control: public, max-age=3600` for public datasets — `backend/app/processing/tiles/router.py`).

## The residual gap

That `?v=` param does **not** bust the shared nginx tile cache. `frontend/nginx.conf`'s `raster-tiles` location block explicitly whitelists the cache key:

```
proxy_cache_key "$dataset_id/$z/$x/$y.$fmt/$arg_colormap_name/$arg_stretch/$arg_pmin/$arg_pmax/$arg_sigma";
proxy_cache_valid 200 1h;
```

Any query param outside that named list (including a cache-busting version) is silently ignored for cache-key purposes. So for a **public** raster dataset whose tiles were already warm in the nginx `raster_cache` zone, any client — including a fresh browser session, or the same browser after its own cache has been cleared — can be served pre-replace tile bytes for up to an hour after a successful replace, independent of anything the frontend does.

This is not new behavior introduced by #1289/#1362 — the same gap already existed for any raster content change reachable through the existing #1221 replace API (e.g. via direct API calls, or conceivably VRT regeneration touching the same cache). #1289 just makes it much easier to hit, since any editor can trigger a replace from the UI.

Private/restricted raster datasets are unaffected (the backend sets `Cache-Control: private, no-store` for those, so nginx never caches them).

## Suggested directions (needs a design decision, not prescribing one)

- Extend `proxy_cache_key` to include a content-version segment (e.g. `last_refreshed_at` or similar), and have the backend emit it in the tile URL/query so cache keys naturally roll over on replace.
- Or: an active cache-purge step in the raster-replace job (nginx cache-purge module, or deleting the relevant cache files under `/var/cache/nginx/raster_cache`) triggered on job completion.
- Or: shorten `proxy_cache_valid` for raster tiles and accept a bounded staleness window as a documented tradeoff (this is closest to today's behavior, just tightened).

[...]

PR #1421, opened 2026-08-12
fix(tiles): fold tile_cache_version into the raster meta cache key

### LINKED ISSUE #1329: tiles: cross-process invalidation for _raster_meta_cache when a raster pointer swaps
`tiles/router.py` keeps `_raster_meta_cache`, a per-process LRU with a 60s TTL that caches the raster metadata (href, band shape, rescale params) the tile proxy builds requests from. When a raster's pointer or description is swapped in place, each API worker process can keep serving tiles from the pre-swap description until its own cache entry expires.

Three paths now perform such a swap, and each accepted this window explicitly rather than closing it for itself:

- raster reupload (#1290, where the decision was first made)
- VRT regeneration
- STAC refresh of a moved asset (#1326, where the same reasoning was recorded as a refuted review finding)

The reasoning each time: closing the window needs cross-process invalidation that none of the three has, and building it for one path while the other two stay stale would imply a guarantee that does not exist. That reasoning is correct per path and is exactly why this issue exists: the fix belongs in the tile router, once, for all three.

Likely shapes, in rough preference order:

1. Fold `tile_cache_version` into the cache key. All three swap paths already bump it in the same transaction as the pointer swap, so a bump would make every process miss on the next lookup with no new coordination channel. Needs a check that the version is available at cache-lookup time without an extra query per tile.
2. A keyed invalidation channel through Valkey (which the tile path already depends on for the tile cache itself).

Bounded impact, so not urgent: the window is at most the 60s TTL, self-heals, and serves stale-but-well-formed tiles. It becomes visible when a refresh changes band shape (RGB to single-band or the reverse), where the stale description can produce rendering errors rather than stale imagery.

ishiland:
Shape 1 of #1329. `tiles/router.py` keeps `_raster_meta_cache`, a per-process LRU with a 60s TTL holding the href, band shape and rescale params the raster proxy builds Titiler requests from. It was keyed on tenant and dataset alone, so when a raster's pointer swapped in place, every API worker that had served a tile in the last minute kept building requests from the pre-swap description until its own entry expired.

## Which lever, and why

The issue conditions shape 1 on the version being readable at cache-lookup time without an extra query per tile. Two candidates, only one of which the code supports.

The dataset row that feeds the raster path does carry `tile_cache_version`: `_resolve_raster_meta` already selects `d.tile_cache_version` and stores it on the `_RasterMeta` snapshot. That value is useless as a key component. It arrives through the very snapshot the key is meant to guard, so it is exactly as stale as the `asset_uri` beside it, and folding it in buys nothing. (The vector `_dataset_cache` near the top of the file is a separate 60s cache keyed on `table_name`, and the raster path never consults it. Its `_DatasetMeta` does not carry the version at all.)

The request does carry it, thanks to #1372. Every raster tile URL the API emits gets `?v=<tile_cache_version>`: the dataset detail payload, map layer payloads, search results, the OGC landing links, and the signed raster template. nginx forwards `$args` when it rewrites `/raster-tiles/...` to `/tiles/raster-proxy/...`, and the Vite dev proxy preserves the query too, so `raster_auth_check` sees the value on the request it is already parsing for the #1372 cacheability check. That is the lever this PR uses. No new query, no new coordination channel.

## What invalidates now

The key gains a `v` segment taken from the request. All three swap paths already bump `tile_cache_version` in the same transaction as the pointer swap (raster reupload #1290, VRT regeneration, STAC moved-asset refresh #1326), so a client that has refetched its metadata requests the new version, misses in every API process, and gets one indexed read of the new href and band shape. Nothing was added to the swap paths themselves; their existing bump is the whole mechanism. The two comments that recorded the window as deliberately open are updated to say what closed it.

The `v` is read once, at the top of `raster_auth_check`, using the same nginx `$arg_v` semantics the #1372 check needed (first occurrence, case-insensitive name), so the key and the cacheability decision can never disagree about which value the request carried.

Lookup and store use different keys, which is what keeps the request from naming the entry it writes (codex P1 on this PR). The lookup key is the request's `v`. The store key is the resolved row's `tile_cache_version`, taken from the snapshot being stored. The first draft used the request's value for both, and that reopened the bug from the other side: the counter is public and increments by one, so an anonymous caller on a public raster could ask for `v=N+1` while the row still read N, park the current snapshot on the key the imminent swap was about to make legitimate, and leave genuine post-swap requests reading the old href for a full TTL. Answering that request with `private, no-store` does not help, because that defends nginx while the poisoned entry is process-local. With the store key derived from the snapshot, key and content always agree: a mismatched request resolves fresh and files under the row's real version, and the first post-swap request at the new version finds nothing and reads the new pointer. Mismatched requests become permanent cache misses, one indexed read each.

The #1372 pre-warm defense is unchanged. Cacheability is still decided against the snapshot's own version, which comes from the database and never from the request, so priming a future `v` gets a `private, no-store` response and never populates the shared cache.

Sibling surfaces on the same axis, so the reviewer can see it is closed: the vector `_dataset_cache` in the same module is keyed on `table_name` with no request-supplied component, and its `_DatasetMeta` does not carry a version at all; the nginx layer already answers a mismatched `v` with `private, no-store`. This meta cache was the only place where a request parameter could name a cache entry, and it no longer can.

## Degradation

An absent `v` (a connect URL copied into desktop GIS, a direct API caller) keeps the unversioned key byte for byte, and with it the old 60s-bounded staleness. So does a malformed one: `_meta_cache_version_segment` accepts a short ASCII digit run and returns None for anything else, which puts the caller back on the old key rather than on an error.

Requests still carrying the old `v` keep hitting their own entry, stale but well formed, until it expires. That is the tradeoff #1329 states it accepts, and it matches what the shared nginx cache does with the same `$arg_v` segment.

Memory stays bounded by the existing `LRUCache(maxsize=256)`. A caller varying `v` to evict entries costs at most one extra indexed read per request, which is the same order as the uncached miss an unknown dataset id already produces on this route.

## Tests

`backend/tests/test_perf002_raster_meta_cache.py` gains a `TestRasterMetaCacheVersionKey` class covering the key derivation, a swap followed by a lookup at the bumped version, the accepted stale read at the old version, an absent `v`, and a junk `v`. Two of them drive the whole thing through `/tiles/raster-auth-check/` so the plumbing from query param to key is exercised, not just the helper.

The poisoning scenario has its own test: an anonymous request for `v=2` while the row reads 1, asserting the entry lands on `:v1` and not `:v2`, then a swap, then a genuine `v=2` request that must resolve the swapped URI. Against the request-keyed store it fails twice over, first on the key and then, with that assertion relaxed to let the test run to the end, on the post-swap request serving the pre-swap pointer. The same fail-first check was run for the three original fix-specific tests. The two degradation tests pass either way, which is the point of them.

`tiles/router.py` grew 77 lines, most of it the notes recording why the lookup uses the request's `v`, why the store does not, so the `_MODULE_LOC_CAPS` entry is raised from 2191 to 2268 with that rationale.

## Verification

```
uv run pytest tests/test_layering.py -q                                        # 40 passed
uv run pytest tests/test_perf002_raster_meta_cache.py -q                       # 10 passed
uv run pytest tests/test_raster_tile_url_version_1372.py tests/test_raster_tiles.py \
  tests/test_raster_colormap_proxy.py tests/test_tile_signing.py \
  tests/test_vrt_catalog_175.py tests/test_rule1_structural.py \
  tests/test_rule2_structural.py -q                                            # 263 passed
uv run pytest tests/test_tiles.py tests/test_tile_cache.py \
  tests/test_tile_cache_scope_sec009.py tests/test_raster_replace_1221.py \
  tests/test_regenerate_vrt_integration.py tests/test_vector_tile_auth.py \
  tests/test_dp03_cross_tenant_privilege_gate.py \
  tests/test_raster_antimeridian_887.py -q                                     # 327 passed
uv run pytest tests/test_stac_refresh_1266.py tests/test_raster_replace_1221.py -q  # 217 passed
```

No schema, API or config impact. No e2e run: the change is backend-only and the dev stack bind-mounts main, so a Playwright run from this worktree would not have exercised it.

Closes #1329

### DIFF
--- a/backend/app/processing/ingest/tasks_raster_swap.py
+++ b/backend/app/processing/ingest/tasks_raster_swap.py
@@ -580,14 +580,18 @@ async def _run_post_swap_followups(
 
     "No reader left" is true of the DATABASE. An API process that served a tile
     in the last minute may still hold this dataset in the tile router's
-    ``_resolve_raster_meta`` cache, whose entries carry the OLD asset_uri for up
-    to ``_RASTER_META_CACHE_TTL`` (60s) — so raster tiles can fail for that
-    window before the entry expires and the new pointer is read. Deliberately
-    not worked around: ``regenerate_vrt`` reaps its superseded generation the
-    same way against the same cache, the window is bounded and self-healing, and
-    closing it needs cross-process invalidation neither path has. The bumped
-    ``tile_cache_version`` already changes the tile URL, so browser and CDN
-    caches roll over immediately.
+    ``_resolve_raster_meta`` cache, whose entries carry the OLD asset_uri.
+    fix(#1329): that cache is now keyed on the request's ``v``, so the
+    ``tile_cache_version`` bump this swap already made in the write transaction
+    IS the invalidation — the first tile request carrying the new version misses
+    in every API process and reads the new pointer, with no separate
+    coordination channel (``regenerate_vrt`` and the STAC moved-asset refresh
+    bump the same counter and get the same effect). What is left is requests
+    still carrying the OLD ``v`` — a tab that has not refetched its tile URL:
+    those keep the pre-swap asset_uri until the entry expires
+    (``_RASTER_META_CACHE_TTL``, 60s), so their tiles can fail for that bounded,
+    self-healing window. The bumped ``tile_cache_version`` also changes the tile
+    URL, so browser and CDN caches roll over immediately.
     """
     from app.core.db import async_session
     from sqlalchemy.orm import joinedload
--- a/backend/app/processing/ingest/tasks_stac_refresh.py
+++ b/backend/app/processing/ingest/tasks_stac_refresh.py
@@ -609,19 +609,20 @@ async def refresh_stac(
                 # beside the content change it describes, which is the
                 # contract on the method.
                 #
-                # It does NOT reach `tiles.router._raster_meta_cache` (#1266
-                # review): that is a per-process LRU in the API workers, keyed
-                # on tenant and dataset alone, so an API process that served a
-                # tile in the last minute keeps the pre-refresh href until its
-                # 60s TTL expires — even for requests carrying the new `_v=`.
-                # Left as it is, deliberately and not for the first time:
-                # `reupload_raster` and `regenerate_vrt` have the same window
-                # against the same cache (see the note in
-                # tasks_raster_swap._run_post_swap_followups), and closing it
-                # needs cross-process invalidation none of the three has.
-                # Building that for one of them would leave the other two
-                # stale while implying otherwise; it belongs to the tile
-                # router, once, for all of them.
+                # It reaches `tiles.router._raster_meta_cache` too, which it
+                # once did not (#1266 review, when that per-process LRU was
+                # keyed on tenant and dataset alone and kept the pre-refresh
+                # href for a TTL even for requests carrying the new version).
+                # fix(#1329) keyed it on the request's `v` as well, so this
+                # bump is itself the invalidation: the first request carrying
+                # the new value misses in every API process and re-reads the
+                # href. `reupload_raster` and `regenerate_vrt` bump the same
+                # counter in their own write transactions and get the same
+                # effect (see the note in
+                # tasks_raster_swap._run_post_swap_followups) — the fix that
+                # belonged to the tile router, once, for all three. A request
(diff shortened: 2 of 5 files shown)


ishiland:
@codex review

ishiland:
@codex review

Issue #1291, opened 2026-08-08
design(raster): srid_override conflates CRS assignment with reprojection


ishiland:
`srid_override` on raster ingest is implemented as `gdalwarp -t_srs`, which reprojects. For a source whose declared CRS is genuinely wrong, reprojecting from the wrong CRS produces garbage coordinates; the correct remedy there is assignment (`-a_srs`), which relabels without resampling.

The field is currently coherent as "put this raster in this CRS", and the schema description ("Forces reprojection during ingestion") is accurate to the implementation, so nothing is lying. But the two use cases the description names ("missing or incorrect") want different operations:

- Missing CRS: assignment is the only option (there is nothing to reproject from). Today's behavior is effectively assignment semantics applied via warp against an absent source CRS.
- Incorrect declared CRS: the caller wants relabeling (`-a_srs`), not a resample from coordinates that were never in the declared system.
- Correct declared CRS, different target wanted: reprojection (`-t_srs`) is right.

Deciding which semantic `srid_override` means (or splitting it into two fields) is a schema-level design question: the description is a generated surface, both SDKs carry it, and PR #1290 deliberately did not touch it while fixing the adjacent bug (the override is now applied when a CRS is declared, per the documented contract).

Where to look: `resolve_crs_assignment` in `backend/app/processing/raster/cog.py` (shared by both raster tails as of #1290), and the conversion call sites in `tasks_raster.py` / `tasks_raster_replace.py`.

Refs: #1290 (the contract fix that surfaced this), #1221.

ishiland:
## Decision: (assignment) — `srid_override` becomes `-a_srs` relabeling, no resample

Researched 2026-08-12 against ce739825f. Three things point the same way:

1. Both use cases the schema description names (missing CRS, incorrect CRS) want assignment. Reprojecting FROM a wrong declared CRS produces garbage coordinates by construction, so nobody can be using the current behavior successfully for the 'incorrect' case, and a missing CRS has nothing to reproject from at all.
2. The internal parameter has been named `assign_crs` all along, and the `cog_preserves_source` audit (cog.py) carries a whole bullet explaining why the warp it actually runs is sample-altering and forces retention of the original upload. Assignment via `gdal_translate -a_srs` is metadata-only.
3. Deliberate reprojection at ingest (correct CRS, different target wanted) is speculative: Titiler reprojects at serve time and export can reproject at export time. If real demand appears it gets its own honestly named field (`target_srid`), not a repurposed one.

[...]

PR #1422, opened 2026-08-12
docs(refresh): document origin_kind as the run's door, not the dataset origin

### LINKED ISSUE #1325: design(refresh): run-ledger origin_kind mixes origin vocabulary with door vocabulary ('raster' is a door, 'created' is missing)
Surfaced twice now: by the 2026-08-07 milestone audit (unfiled gap) and independently by PR #1323's review of the Source panel wiring.

Two vocabularies share the name `origin_kind`:

- **Dataset origin kinds** (`ORIGIN_KINDS` in `backend/app/platform/dataset_origin.py:48`): `upload, postgis, service, stac, created`.
- **Refresh-run ledger kinds** (CHECK constraint in `backend/app/platform/refresh/models.py:60`): `upload, postgis, service, stac, raster` — `raster` in, `created` out.

The divergence exists because the raster replace door (#1290) labels its runs `raster` even though a raster dataset's ORIGIN is `upload` (it was uploaded). So a raster dataset's Source panel says origin "upload" while its run history rows say "raster" — same field name, different meaning (door vs origin). `DatasetRefreshRunResponse`'s description (`datasets/domain/schemas.py:1398`) documents the ledger vocabulary, and the docstring near `dataset_origin.py`'s vocabulary now reads as contradicting it.

## Decide one

(a) **Document as intentional**: the ledger's `origin_kind` is really "execution strategy / door"; rename nothing (the CHECK is deployed), but rename in prose — docstrings on both vocabularies point at each other and state the door-vs-origin distinction, and the frontend/CLI render it as the run's *mechanism*, not the dataset's origin. Cheapest, no migration.

(b) **Unify**: migrate the ledger column toward the origin vocabulary (raster replace runs become `upload`) and add a separate `door`/`mechanism` column if the distinction matters. Migration-bearing, touches #1290's fencing tests.

Recommendation: (a). The distinction is real and useful (a future `created`-origin snapshot-refresh would also want its own door label); it just needs to stop masquerading as the same vocabulary.

Refs: #1290 (raster replace door), #1313 (postgis strategy), PR #1323 (where it resurfaced).

ishiland:
## What

Implements option (a) from #1325: document the `origin_kind` name collision instead of renaming anything.

Two vocabularies share the field name `origin_kind`:

- `ORIGIN_KINDS` in `backend/app/platform/dataset_origin.py` (upload, postgis, service, stac, created) is the dataset's origin: how its data entered the catalog, recomputed fresh from the dataset's current `source_format`/`record_type` on every response. It is not stored and not fixed at ingest, but it does not move on every `source_format` write either: `classify_origin()` sorts every format into one of five buckets, and only a mutation that crosses a bucket boundary (a successful raster replace: `stac` to `geotiff`, reclassifying `stac` to `upload`) changes it. A same-bucket reupload, GeoJSON to CSV for example, leaves it `upload` both times.
- The CHECK constraint on `catalog.dataset_refresh_runs.origin_kind` in `backend/app/platform/refresh/models.py` (upload, postgis, service, stac, raster) is a run's execution door: which mechanism this particular refresh or reupload attempt went through.

Four of the five values happen to share a spelling across both vocabularies. That is a coincidence of which doors have been built so far, not a guarantee that a run's door and its dataset's origin ever agree (see the correction below for a case where they don't). The other two values don't even share a spelling: `created` has no ledger counterpart (nothing to refresh from), and `raster` has no origin counterpart (it is reserved for a future raster-replace door label; today's raster-replace runs use `upload`).

Changes:

- `dataset_origin.py`: comment above `ORIGIN_KINDS` states the door-vs-origin split and points at the ledger constraint.
- `refresh/models.py`: comment above the CHECK constraint states the same split from the other side, and what `raster` means.
- `schemas.py`: `DatasetRefreshRunResponse.origin_kind`'s field description now says door, not origin. This is an OpenAPI-visible string, so `backend/openapi.json`, both SDKs (`sdks/python`, `sdks/typescript`), and the frontend's generated type mirror (`frontend/src/types/api.generated.ts`) are regenerated and committed. The hand-maintained `frontend/src/types/api.ts` mirror gets the same note in a doc comment.
- `SourcePanel.tsx`: the run-history row's origin_kind badge gets a label ("Method: x") and a tooltip explaining it's the run's mechanism, not the dataset's origin, so it stops reading as a second, disagreeing origin claim next to the OriginBadge shown above it. Two new keys (`sourcePanel.refresh.history.mechanismLabel`, `mechanismTooltip`) added to all four locales (en/es/fr/de).
- CLI: checked `cli/geolens_cli/` for anywhere it prints `origin_kind`. The only place it does is the refresh-dispatch response (`DatasetRefreshResponse`, a different schema from the run-history one above), and that response only ever carries `postgis`/`service`/`stac`. `upload`, `created`, and `raster` datasets are refused before reaching it, so its `origin=` label was never ambiguous. No CLI change.
- Backend module-size gate: `schemas.py` was already at its exact cap; `_MODULE_LOC_CAPS` in `test_layering.py` moves from 1458 to 1468 across this PR's wording revisions, each bump with its own comment explaining the added lines.

No renames, no migration. `chk_refresh_runs_origin_kind` stays exactly as deployed.

## One correction to the issue's premise

The issue describes the visible symptom as a raster dataset's Source panel showing origin "Upload" while its run-history row says "raster". That specific pairing does not happen today. Tracing the actual code path (`reupload_commit` in `router_reupload.py`, the one place a raster replace creates its ledger row) shows raster-replace runs are stamped `origin_kind="upload"`, never `"raster"`. `test_raster_replace_1221.py`'s `_queue_replace_job` helper, which mirrors what the real commit door writes, asserts the same thing. `raster` has been present in the CHECK constraint since the ledger table's first commit, before raster replace existed, so it reads as a value reserved ahead of an implementation that never actually switched to it, the same pattern as this file's own `blocked` status and `scheduled` trigger comments. That half of the premise, a run row literally showing `"raster"`, is dormant, not live.

But the underlying contradiction the issue is pointing at, a run's `origin_kind` disagreeing with its dataset's own origin, is reachable today, just through a different value. A STAC-imported raster's origin is `"stac"`. Replacing it stamps the run `origin_kind="upload"` immediately, at commit time (`router_reupload.py`), while the dataset's origin is only rebound to `"upload"` inside a successful swap's field write (`tasks_raster_swap.py`, `_write_swapped_fields`). While that replace is pending, and permanently if it fails, the Source panel would show origin "STAC" next to a run row recorded "upload": the same shape of disagreement the issue describes, on a supported path, today.

This PR documents the general rule rather than the specific case the issue named: `origin_kind` on a run is the door it executed through, and it must never be read as, or presented as, the dataset's origin, whether or not the two happen to share a spelling. `dataset_origin.py`, `refresh/models.py`, the OpenAPI field description, and the two frontend comments all say so explicitly now, and the STAC-raster-replace path above is the concrete example each of them points at. This still does not change any stamping behavior; making the two vocabularies agree on every path would be a behavior change outside what option (a) covers.

Closes #1325.

## Notes for reviewers

- Another PR in this wave (#1291) also regenerates `backend/openapi.json`; merge order between the two is being coordinated so neither's regen artifacts get clobbered by the other's rebase.
- Nothing here is a schema/DB change. The CHECK constraint's string is unchanged; only comments, one field description, and frontend display strings moved.

## Verification

- `cd backend && uv run ruff check .` and `uv run ruff format --check .`: clean on touched files
- `uv run pytest tests/test_dataset_refresh_runs.py tests/test_origin_ref_indexes.py tests/test_refresh_gate_1269.py tests/test_raster_replace_1221.py tests/test_stac_refresh_1266.py tests/test_dataset_source_state.py tests/test_service_refresh_1220.py tests/test_source_health_1222.py tests/test_postgis_refresh_1265.py tests/test_reupload.py tests/test_dataset_source_freshness.py -q`: 726 passed
- `make openapi`, `make sdks`, `npm run types:generate`: regenerated and diffed to confirm the only change is the `origin_kind` description
- `cd cli && uv run --extra dev python -m pytest -q`: 308 passed (sanity check against the regenerated Python SDK)
- `cd frontend && npm run typecheck`: clean
- `cd frontend && npm run lint`: clean
- `cd frontend && npx vitest run` on the touched dataset test files: 58 passed
- `cd frontend && npm run test:i18n`: clean

### DIFF
--- a/backend/app/modules/catalog/datasets/domain/schemas.py
+++ b/backend/app/modules/catalog/datasets/domain/schemas.py
@@ -1422,7 +1422,14 @@ class DatasetRefreshRunResponse(BaseModel):
             "row is purged by retention; the run itself survives."
         ),
     )
-    origin_kind: str = Field(description="upload, postgis, service, stac, or raster")
+    origin_kind: str = Field(
+        description=(
+            "The run's execution door, not the dataset's origin: upload, "
+            "postgis, service, stac, or raster. 'raster' is the "
+            "raster-replace door and has no origin counterpart; a raster "
+            "dataset's own origin is 'upload'."
+        )
+    )
     trigger: str = Field(description="manual, api, or cli")
     status: str = Field(description="pending, running, succeeded, failed, or cancelled")
     triggered_by: uuid.UUID | None = None
--- a/backend/app/modules/catalog/datasets/domain/schemas.py
+++ b/backend/app/modules/catalog/datasets/domain/schemas.py
@@ -1425,9 +1425,10 @@ class DatasetRefreshRunResponse(BaseModel):
     origin_kind: str = Field(
         description=(
             "The run's execution door, not the dataset's origin: upload, "
-            "postgis, service, stac, or raster. 'raster' is the "
-            "raster-replace door and has no origin counterpart; a raster "
-            "dataset's own origin is 'upload'."
+            "postgis, service, stac, or raster. 'raster' is reserved for a "
+            "future, distinct raster-replace door label; today's "
+            "raster-replace runs are still recorded as 'upload', matching "
+            "the dataset's origin."
         )
     )
     trigger: str = Field(description="manual, api, or cli")
--- a/backend/app/modules/catalog/datasets/domain/schemas.py
+++ b/backend/app/modules/catalog/datasets/domain/schemas.py
@@ -1425,10 +1425,12 @@ class DatasetRefreshRunResponse(BaseModel):
     origin_kind: str = Field(
         description=(
             "The run's execution door, not the dataset's origin: upload, "
-            "postgis, service, stac, or raster. 'raster' is reserved for a "
-            "future, distinct raster-replace door label; today's "
-            "raster-replace runs are still recorded as 'upload', matching "
-            "the dataset's origin."
+            "postgis, service, stac, or raster. The two can visibly "
+            "diverge; for example a STAC-imported raster's pending or "
+            "failed replace run is recorded 'upload' while the dataset's "
+            "origin stays 'stac' until the replace succeeds. 'raster' "
+            "itself is reserved for a future, distinct raster-replace door "
+            "label, with today's raster-replace runs recorded 'upload'."
         )
     )
     trigger: str = Field(description="manual, api, or cli")
--- a/backend/app/platform/dataset_origin.py
+++ b/backend/app/platform/dataset_origin.py
@@ -45,6 +45,18 @@ SERVICE_SOURCE_FORMATS: frozenset[str] = frozenset(
 # fetched from anywhere. Both classify as None so the type badge speaks alone.
 _ORIGINLESS_RECORD_TYPES: frozenset[str] = frozenset({"collection", "vrt_dataset"})
 
+# fix(#1325): this is the dataset's ORIGIN — how its data entered the catalog,
+# derived once by classify_origin() from source_format/record_type and never
+# revisited afterward. It is a DIFFERENT vocabulary from
+# DatasetRefreshRun.origin_kind, the CHECK constraint in
+# platform/refresh/models.py: that column is the per-attempt execution DOOR a
+# refresh or reupload ran through, not the dataset's origin, and the two sets
+# diverge on purpose in both directions. 'created' is here but has no ledger
+# counterpart, because a dataset drawn in the app has nothing to refresh from.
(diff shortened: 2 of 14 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1423, opened 2026-08-12
fix(raster): srid_override assigns the CRS instead of reprojecting

### LINKED ISSUE #1291: design(raster): srid_override conflates CRS assignment with reprojection

ishiland:
Implements the decision recorded on #1291: https://github.com/geolens-io/geolens/issues/1291#issuecomment-5260372477

## What changes

`srid_override` reached `convert_to_cog` as a `gdalwarp -t_srs` prepend, which resamples every pixel onto a new grid. It now rides the existing `gdal_translate` as `-a_srs`, which writes a CRS tag while every band passes through.

Measured on the deployed GDAL, with a 32x32 EPSG:4326 source relabelled to 3857:

| | grid | bounds | pixels |
| --- | --- | --- | --- |
| `gdal_translate -a_srs` | 32x32 | unchanged | array-equal to the input |
| `gdalwarp -t_srs` | 42x42 | rewritten in metres | different values |

## Why this is a bug fix for the documented contract

The field's own description names two cases, and neither wanted a warp. A source with no CRS has nothing to reproject *from* — GDAL's own fallback there was to treat the source SRS as the target, so that case was already an assignment in everything but its retention accounting. A source whose declared CRS is *wrong* reprojects from a wrong starting point, so the output coordinates are wrong by construction; nobody can have been using that successfully.

The only use that changes is the off-label one — correct CRS, reprojection wanted at ingest — which the description never promised. Deliberate reprojection is not being added: Titiler reprojects at serve time and export reprojects at export time, and if real demand appears it gets an honestly named `target_srid` rather than a repurposed field.

## `cog_preserves_source`

The audit bullet claiming `assign_crs` is sample-altering is rewritten. `-a_srs` is the same kind of flag as `-a_nodata` sitting next to it: a tag, not a resample. So under a lossless codec the COG carries the uploaded samples and the pre-conversion upload is redundant, and both raster tails now call the predicate without `reprojected=`.

That inverts a "keep the original" into a "delete it", and the docstring notes the asymmetry that makes it worth arguing rather than just doing: a wrong `True` here permanently deletes the only faithful copy. Three things carry the conclusion.

1. The samples are provably intact — measured above, and asserted end to end against the published object in `test_the_cog_carries_the_uploaded_samples_and_grid`.
2. A relabel is the one conversion step that is **reversible from the stored artifact**. A caller who assigns the wrong EPSG assigns another one over the same untouched pixels; a warp could never be undone, which is exactly what made the retained upload irreplaceable while one ran.
3. The upload's own CRS declaration is not lost with the file — `Dataset.original_srid` records it, by the split #1290 introduced.

No caller was left unproven, so nothing needed a conservative carve-out. The `reprojected` parameter itself stays (with its tests): nothing sets it today, and a future reprojecting field must.

`check_and_prepare_cog` still treats any `assign_crs` as a custom option and always converts — verified and kept. The tag is an argument to the translate run, so there is no path that relabels an already-compliant COG in place, and a `verified` return would publish the source still carrying the CRS the caller asked us to replace.

## Footprint under an override

This is the subtlest part, and it was a real bug class here before (#1290). With assignment the pixel grid and the corner **numbers** survive the conversion unchanged; only what they mean changes. So the footprint is decided entirely by which CRS those numbers are read under, and reading the source would read them under the CRS the caller just called wrong — placing the dataset somewhere else on earth with no field visibly disagreeing.

Both tails already read their catalog metadata off the converted COG, so the math needed no change; the in-code comments that justified it in terms of a warp did. The existing tests were moved to assignment semantics rather than deleted: the replace tail now asserts the published extent lands where the assigned CRS puts it (a whole-world 4326 source assigned 3857 must record a sub-degree footprint at the origin, which excludes the source read, a source-CRS read of the COG, and a warp in one assertion), and the first-ingest tail's half is pinned structurally, the way #1290 pinned its sibling for the same runtime reason.

## API surface

Description text only — no field, type, or validation change:

> EPSG code to use when the source CRS is missing or incorrect. Assigns (relabels) the CRS the source is read under rather than reprojecting from it: raster pixel values are unchanged, and vector geometries are reprojected to EPSG:4326 from the assigned CRS as usual.

The vector subclass gets its own wording, because for vector the override has always been an assignment followed by the standard reprojection to 4326 that every vector source gets. Regenerated: `backend/openapi.json`, both SDKs, and `frontend/src/types/api.generated.ts`. Each is a one-line diff.

**Merge order:** #1325 in this wave also regenerates `openapi.json`. Whichever lands second needs a regen on top; the orchestrator is sequencing it.

Also updated because they documented the old behavior: `RUNBOOK.md` section 9 (a CRS override was listed as a reason the upload is kept), a `CHANGELOG.md` entry under Unreleased, and the import form's resampling help text in all four locales — with the warp gone, `resampling` only shapes overviews.

## Verification

```
backend: ruff check . / ruff format --check .            pass
backend: pytest tests/test_layering.py                    40 passed
backend: pytest tests/test_raster_replace_1221.py         112 passed
backend: pytest tests/test_cog_subprocess_env.py          11 passed
backend: pytest tests/test_rule2_structural.py            98 passed
backend: pytest tests/test_raster_ingest.py tests/test_raster_validation.py
         tests/test_raster_schema.py tests/test_cog_band_unit_capture.py
         tests/test_cog_info.py tests/test_raster_antimeridian_887.py
         tests/test_raster_ingest_orphan_cleanup_gap017.py
         tests/test_perf002_raster_meta_cache.py          241 passed
backend: pytest tests/test_ingest.py tests/test_ingest_metadata.py
         tests/test_manifest_apply_service.py tests/test_manifest_apply_roundtrip.py
         tests/test_dp01_ingest_write_path.py
         tests/test_presigned_staging_reap_1202.py        141 passed
backend: pytest tests/test_presigned_raster_stamp_1186.py
         tests/test_strict_cog_enforcement.py
         tests/test_commit_request_schemas.py             58 passed
python scripts/dump_openapi.py --check                    exit 0
make sdks                                                 clean regen, committed
```

Rule 2: the removed `gdalwarp` argv site is one fewer GDAL subprocess; the surviving `gdal_translate` argv keeps `gdal_safe_env()`, and the structural guard still sees 10 argv sites against its floor of 6. Playwright was not run — this is backend and generated-type work, and a worktree e2e run would have exercised `main`.

Closes #1291

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,22 @@ and releases use semantic versioning.
 
 ## [Unreleased]
 
+### Changed
+
+- **A raster CRS override now assigns the CRS instead of reprojecting to
+  it.** Supplying an EPSG code at import or replace time relabels the raster
+  in place: pixel values, the pixel grid and the corner coordinates all come
+  through the conversion untouched, and only the CRS they are read under
+  changes. That is what both documented uses of the field want — a file with
+  no CRS has nothing to reproject from, and a file whose declared CRS is
+  wrong reprojects from a wrong starting point and lands somewhere wrong.
+  Two consequences: a dataset ingested with an override now sits where those
+  coordinates put it in the CRS you named, and because the conversion no
+  longer resamples, the uploaded file is deleted after a successful lossless
+  ingest rather than kept as a second permanent copy. Deliberate
+  reproject-at-ingest is not offered; rasters are reprojected at serve time
+  and at export time (#1291).
+
 ## [1.11.1] - 2026-08-10
 
 ### Added
--- a/RUNBOOK.md
+++ b/RUNBOOK.md
@@ -1993,8 +1993,8 @@ Decision 7).
 
 ### When the conversion is lossy, the upload is kept
 
-Two things about a conversion can make the COG an unfaithful copy, and either
-one causes the upload to be kept.
+One thing about a conversion can make the COG an unfaithful copy, and it causes
+the upload to be kept.
 
 **Compression.** The import form lets the uploader choose. Four options are
 lossless — **DEFLATE** (the default), **LZW**, **ZSTD**, **LERC** — and two
@@ -2006,12 +2006,21 @@ be tuned to throw away precision, but that is the GDAL creation option
 it. So a LERC upload keeps its exact sample values and is treated like any other
 lossless conversion.
 
-**Reprojection.** Supplying a CRS override reprojects the raster, which
-resamples every pixel onto a new grid. The output is a faithful *rendering* but
-not the original measurements, so this counts as lossy even under DEFLATE.
+Under a lossy codec the uploaded file is the only copy of the original samples
+that will ever exist, so GeoLens keeps it.
 
-In either case the uploaded file is the only copy of the original samples that
-will ever exist, so GeoLens keeps it.
+**A CRS override is not lossy** (behavior change, issue #1291).
+Supplying an EPSG code at import time *assigns* that CRS: it relabels the raster
+where it already sits and leaves every pixel value and every corner coordinate
+alone. Only the meaning of those coordinates changes, which is the point — you
+reach for the override when the file's own CRS is missing or wrong. So an
+override under a lossless codec still deletes the upload, and it is safe to:
+the COG holds your samples exactly, and a mistaken override is corrected by
+assigning a different code over the same untouched pixels.
+
+Earlier versions reprojected instead, which resampled every pixel onto a new
+grid, and therefore kept the upload. If you are looking for a retained original
+from one of those ingests, it is still where it always was.
 
 ### Where the kept original lives, and for how long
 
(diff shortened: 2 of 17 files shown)


ishiland:
@codex review

PR #1424, opened 2026-08-12
fix(vrt): stage source-link mutations on the generation and apply them at swap commit

### LINKED ISSUE #1327: design(vrt): source-link mutation and regeneration are non-atomic — the sweep can only refuse to hide the drift
Root cause surfaced by PR #1322's review (rounds 1-2, both P1): `add_vrt_source`/`remove_vrt_source` commit their `vrt_source_links` mutation in the same transaction that flips the asset to `'regenerating'` — BEFORE the regeneration that would make the served artifact match. The two steps are separate, non-atomic, non-compensable transactions, so a death between them leaves the catalog's declared composition permanently ahead of the served bytes.

PR #1322 made the abandoned-attempt sweep HONEST about this state (it discriminates composition-preserving vs composition-changing dead attempts via `built_from` vs the live link set, and refuses to restore `'ready'` over drift), but the sweep deliberately does not repair it — active repair inside a sweep was rejected in review as too much machinery.

## Decide one

(a) **Compensable links**: stage the link mutation with the generation (e.g. write the new link set into the generation row, apply it to `vrt_source_links` only in the same transaction as the successful swap — the same commit-time pattern `built_from` itself uses, #1290). Death before swap then leaves catalog and artifact consistent automatically, and the sweep's degraded branch becomes near-unreachable.
(b) **Auto-retry**: the sweep (or a retry endpoint) re-dispatches the regeneration for composition-drifted assets instead of parking them in `'failed'`. Repairs forward but keeps the inconsistency window and adds sweep-side dispatch machinery.

Recommendation: (a) — it removes the inconsistent state rather than adding a repair path for it, and reuses an established pattern.

Refs: #1267 / PR #1322 (sweep honesty + the discrimination logic and its in-code rationale), #1290 (`built_from` commit-time pattern).

ishiland:
Implements option (a) of #1327.

## The invariant

`vrt_source_links` can no longer describe a composition the served VRT does not have.

Before this, `add_vrt_source`/`remove_vrt_source` committed their link mutation in the same transaction that flipped the asset to `regenerating`, then deferred the regeneration. Those are two transactions with nothing between them to compensate, so a worker death after the first left the catalog's declared composition permanently ahead of the served bytes. #1322 taught the abandoned-attempt sweep to detect that drift and refuse to hide it, which is as far as a reconciler can honestly go.

Now the endpoints stage the full intended post-mutation member set on the `VrtGeneration` row and touch no links. `regenerate_vrt` builds from that set and applies it to `vrt_source_links` in the same transaction that swaps `asset_uri` and writes `built_from`, the commit-time pattern from #1290. Death anywhere before that swap leaves the links exactly where the artifact is, with nothing to roll back. A full set is staged rather than a delta, so applying it is a replace: self-contained, and idempotent on retry.

Concurrency improves for free. Two adds that both pass the `regenerating` guard used to insert both links while only one artifact was ever built. Now the loser's set is never applied, because applying is fenced behind the `current_generation_id` check that already gates the publish.

## Schema

One new nullable column, `catalog.vrt_generations.staged_source_ids` (JSONB, an ordered array of source dataset ids, position implied by order). Migration `0043_vrt_generations_staged_source_ids`, additive, no backfill, no data rewrite. `backend/openapi.json` is unaffected: the column is internal and no response schema changed.

## NULL means "changes no membership"

A generation with no staged set builds from the live link rows and applies nothing. That covers a plain regenerate (`regenerate_vrt_endpoint` never changed membership) and any generation queued before this column existed, including one still sitting in the queue across a rolling deploy. One fallback, two producers, no version check.

## Sweep interaction

The composition-drift branch in `sweep_stale_vrt_assets` keeps its code and gains the reason it is now near-unreachable: a dead composition-changing attempt leaves the links matching `built_from` and takes the honest restore path, so the VRT resumes serving what it always served with the requested change simply not applied. The branch still guards rows written before this change, and any future writer of the link table outside a publish transaction. Deleting a check whose FALSE side went quiet would only make the next cause invisible.

## Set drift in both directions

A staged source can be created or deleted between the request and the apply. Both are decided rather than discovered.

Deletion is the harder one. A live link pins its source with `ON DELETE RESTRICT`, but a staged id has no link row yet, so the staging window is the one place a member can vanish. Two owners, two different remedies. The dataset delete guard now also refuses to delete a source that a `pending`/`running` generation has staged, which preserves the "you cannot delete a COG a VRT depends on" guarantee that staging would otherwise have punched a hole in. If a member disappears anyway, the task refuses to build a set it cannot fully load and fails the run whole, links untouched and served VRT intact, and the caller re-issues against the set that survived.

That guard asks membership with jsonb containment rather than by expanding the array, because containment is total. An explicitly assigned `None` lands in the column as the JSON scalar `null` (measured against Postgres, the #1322 shape), which an expansion would raise on, and nothing obliges the planner to evaluate a `jsonb_typeof` guard sitting beside it in the same `AND` first. A test holds a live generation in exactly that shape.

Only `pending`/`running` generations count, so the block cannot outlive the attempt. A terminal generation's staged set is ignored, and an attempt whose worker died is marked `failed` by the existing stale sweep, which releases the source on its own.

Creation is irrelevant by construction: the staged set is closed at request time, and the apply replaces the link rows with exactly that set.

The staged value itself is validated once, at claim time, before any bytes are built. Empty, unparseable or repeated ids fail the attempt rather than surfacing later as an opaque `ON CONFLICT` error.

## Deploy skew, both directions

Staging splits one write into a producer (the API) and a consumer (the worker), so the two can disagree about what the payload means while a deployment is in flight. The default install cannot get there: `api` and `worker` share an image and are recreated together by compose, so the window belongs to the chart's rolling update, where the API can be new while a worker is still old.

New worker, old API: the delivery carries no staged set, the column reads NULL, and the task builds from the live links and applies nothing. That is the same fallback pre-column rows take, so nothing special happens during the roll.

Old worker, new API: this is the one that could lose data, and it is why the staged mutations now go out under a second registered task name, `regenerate_vrt_staged`. A pre-#1327 worker does not know that column; handed the delivery it would rebuild from the live links, publish the composition that was already being served, and mark the generation complete, dropping an accepted add or remove with every state machine reporting success.

A marker kwarg cannot fence that, and the test says so: the pre-#1327 task signature ends in `**kwargs`, and so does `tenant_task`'s wrapper, so an unknown keyword is swallowed instead of raising. The task name is resolved against the worker's own registry, which is the one thing an old worker cannot fake. Measured against the pinned Procrastinate rather than assumed: a job whose task name is absent ends `failed` with one attempt, and no retry is scheduled, because there is no task object to ask for a retry strategy. That last fact also decides the retry question in this PR: adding a bounded retry to the new task would not make the window self-healing, since the retry policy lives on the worker that does not have the task. So `retry=0` stays.

The refused delivery lands in a state the existing machinery already owns rather than a new one. The task never ran, so the links are untouched, the generation stays `pending` with a NULL heartbeat, and the asset stays `regenerating` until `sweep_stale_vrt_assets` reconciles it: composition preserved, asset restored to `ready`, requested change simply not applied, generation row `failed`. The honest cost is that the VRT 409s further mutations until that sweep runs, and the operator re-issues the add or remove after the roll. Plain regeneration keeps the legacy name on purpose, so those deliveries keep flowing to old workers for the duration.

## Report surfaces

Two surfaces reported the composition and would have disagreed after this change.

The 202 message now reads "Source add queued" / "Source removal queued". The catalog's source list still describes the VRT being served; the change joins it when the artifact containing it is published.

The OGC record item's `raster.source_count` used to be read from the in-flight `VrtGeneration`. That was the post-mutation number back when the link write happened up front. Now it would be the pending composition, leaving that one surface claiming a member set the VRT is not serving. It counts the live links instead, like the dataset detail and list surfaces already did. The generation's own count is still reported per attempt by the VRT generations endpoint, and the port method that returns it is annotated with what it may and may not be used for.

## Tests

- `test_regenerate_vrt_integration.py`: five new database-backed tests over the real task. A successful run applies the staged set, and the stored VRT XML references one source per staged member, which is what proves the build read the staged set rather than the links it was about to overwrite. A build that dies mid-attempt leaves `vrt_source_links` untouched. A staged source deleted mid-flight fails the run cleanly. A NULL staged set builds from the live links, with an unlinked raster sitting there to prove the build did not wander. Applying a staged set twice is a no-op that also preserves a surviving member's `created_at`.
- `test_vrt_source_management_174.py`: the endpoints stage the set and issue no link write, on both add and remove; the orphan-guard rollback no longer compensates a link change; a 422 leaves no generation behind. A new class pins what the task accepts as a staged set (absent and JSON-null both mean "no membership change"; empty and repeated ids are refused before the build). Plus a database-backed class that runs both endpoints over HTTP and checks the JSONB column round-trips while the link table stays put, including the 404 path.
- `test_defer_orphan_guard.py`: the two VRT defer-failure tests converted. They now assert the rollback issues no link statement at all, and that the staged set is what the endpoint wrote.
- `test_vrt_stale_sweep_gap002.py`: new coverage for a dead staged mutation taking the restore branch. The two drift scenarios are kept and re-documented as what they now defend (pre-#1327 rows, unforeseen writers) rather than deleted.
- `test_vrt_delete_guard_174.py`: two database-backed tests. An in-flight staged member blocks the delete; a terminal generation's staged set does not.
- `test_vrt_source_query_count.py`: the record item reports the served member count while an add is in flight.
- `test_vrt_staged_mutation_1327.py`: new file for the delivery and skew half. Over HTTP, add and remove write the staged set and leave the link table alone (including the 404 path, which stages nothing), and both deliver under the staged task name while a plain regenerate keeps the legacy one. The skew mechanism is measured against the pinned Procrastinate: a worker holding only the legacy task succeeds the legacy delivery and fails the staged one with no retry. A last test walks the refused delivery through the stale sweep and shows it converging on the restore path with the links untouched.

  These are in their own module for a reason worth stating: they started life in `test_vrt_source_management_174.py`, whose autouse fixture patches authorization helpers, and a client-backed test there made `router_vrt` import for the first time mid-patch, permanently binding the mocks into that module and 404-ing unrelated tests in later files. The module docstring now says not to add DB or client tests there.

## Verification

```
set -a && source .env.test && set +a
uv run alembic upgrade head            # plus downgrade -1 and back up again
uv run alembic check                   # "No new upgrade operations detected."
uv run python scripts/dump_openapi.py --check
uv run pytest tests/test_regenerate_vrt_integration.py tests/test_vrt_source_management_174.py \
  tests/test_vrt_stale_sweep_gap002.py tests/test_vrt_delete_guard_174.py \
  tests/test_vrt_source_query_count.py tests/test_vrt_catalog_175.py \
  tests/test_vrt_source_authz_1172.py tests/test_vrt_lifecycle_188.py tests/test_worker.py \
  tests/test_ingest_job_attempt_fencing.py tests/test_jobs_router.py -q
uv run pytest -q -n 4
```

Closes #1327

### DIFF
--- a/backend/alembic/versions/0043_vrt_generations_staged_source_ids.py
+++ b/backend/alembic/versions/0043_vrt_generations_staged_source_ids.py
@@ -1,0 +1,48 @@
+"""Stage a VRT generation's intended member set on the generation row.
+
+fix(#1327). ``add_vrt_source``/``remove_vrt_source`` committed their
+``vrt_source_links`` mutation in the same transaction that flipped the asset to
+``'regenerating'`` — before the regeneration that would make the served artifact
+match. The two steps were separate, non-compensable transactions, so a worker
+death between them left the catalog's declared composition permanently ahead of
+the served bytes; #1322's sweep could detect that drift and refuse to hide it,
+but not repair it.
+
+``staged_source_ids`` holds the FULL intended post-mutation member set as an
+ordered array of source dataset ids. The link table is left alone at request
+time and the staged set is applied to it in the same transaction that swaps the
+artifact and writes ``built_from`` (the #1290 commit-time pattern). Death before
+that swap now leaves the links untouched, so catalog and artifact agree with no
+compensation to run.
+
+Nullable with no backfill, deliberately. NULL means the generation changes no
+membership — a plain regenerate, or anything queued before this column existed —
+and the task builds from the live links and applies nothing for it.
+
+Revision ID: 0043_vrt_generations_staged_source_ids
+Revises: 0042_record_distribution_single_primary
+Create Date: 2026-08-11
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+revision: str = "0043_vrt_generations_staged_source_ids"
+down_revision: Union[str, None] = "0042_record_distribution_single_primary"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "vrt_generations",
+        sa.Column("staged_source_ids", postgresql.JSONB(), nullable=True),
+        schema="catalog",
+    )
+
+
+def downgrade() -> None:
+    op.drop_column("vrt_generations", "staged_source_ids", schema="catalog")
--- a/backend/app/core/catalog_port.py
+++ b/backend/app/core/catalog_port.py
@@ -269,6 +269,11 @@ class CatalogPort(Protocol):
         self, session: AsyncSession, dataset_ids: list[uuid.UUID]
     ) -> dict[str, dict[str, Any]]: ...
 
+    # How many members ONE generation is assembling. fix(#1327): that is the
+    # attempt's intended set, which for a source add/remove is not what the VRT
+    # is serving until the attempt publishes — so this is a fact about the
+    # generation and must not be projected as a dataset's `source_count`. Count
+    # `catalog.vrt_source_links` for that, as every catalog surface now does.
     async def get_vrt_generation_source_count(
         self, session: AsyncSession, generation_id: uuid.UUID
     ) -> int | None: ...
--- a/backend/app/modules/catalog/datasets/domain/service_lifecycle.py
+++ b/backend/app/modules/catalog/datasets/domain/service_lifecycle.py
@@ -87,7 +87,29 @@ async def delete_dataset(
         from app.platform.storage.provider import get_storage
 
         if record_type == "raster_dataset":
-            # Guard: prevent deletion if any VRT still references this COG
+            # Guard: prevent deletion if any VRT still references this COG.
+            #
+            # fix(#1327): a reference is now either committed or in flight. An
+            # add staged by add_vrt_source has no vrt_source_links row until
+            # its regeneration publishes the artifact, so the link half of this
+            # guard alone would let a source be deleted out from under an
+            # in-flight attempt — a hole staging would otherwise open in a
+            # guarantee this repo already made. The second branch closes it by
+            # asking the same question of the not-yet-applied set. The
+            # regeneration task independently refuses to publish a set whose
+            # members it cannot load, so a delete that slips past this guard
+            # (deleted between this check and the build) fails the attempt
+            # instead of half-applying it.
+            #
+            # Membership is asked with `@>` rather than by expanding the array
+            # with jsonb_array_elements_text. Containment is TOTAL over jsonb:
+            # a column holding the JSON scalar `null` (which is what SQLAlchemy
+            # writes for an explicitly-assigned None — #1322's lesson) answers
+            # false instead of raising, and SQL NULL drops the row. An
+            # expansion would need a jsonb_typeof guard beside it in the same
+            # AND, and nothing obliges the planner to evaluate that guard
+            # first. Only 'pending'/'running' generations count — a terminal
+            # one will never apply its set.
             refs_result = await session.execute(
                 text(
                     """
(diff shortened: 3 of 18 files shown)


ishiland:
@codex review

ishiland:
@codex review

PR #1428, opened 2026-08-12
fix(maps): bound and offload the sprite-sheet render

ishiland:
Backend-audit follow-up on the map sprite sheet.

## What

`GET /maps/sprites/geolens.png` (and the `@2x` twin) takes no auth, and on a cache miss it decoded every icon in the catalog with PIL directly on the event loop — one storage read per icon in series, then `Image.open().convert("RGBA")`, a LANCZOS `thumbnail()`, and `save()` inline. Validation capped bytes but not pixels, and the two are only loosely related: a 9000x9000 PNG is 79 KiB on the wire, inside the 512 KiB cap, and costs roughly 320 MB and a third of a second to decode. Every other request in flight waits through that.

Two smaller holes in the same path: PIL raises `DecompressionBombError` from `Image.open` on absurd IHDR dimensions, and `verify()` raises `IndexError` on a PNG with no IDAT chunk. Neither is named in the handlers that were supposed to catch bad image bytes, so both surfaced as 500s — on upload, and on the sprite route for a row already in the database.

## Changes

Two pixel caps, deliberately different numbers because they answer different questions:

- `MAX_ICON_DIMENSION` (1024px) is a **product** bound on what may be uploaded from here on, checked off the IHDR header that `Image.open` parses lazily. It is far past what a 24px sprite cell (48 at `@2x`) can show, so it costs a new icon nothing.
- `MAX_RENDER_PIXELS` (24M px) is a **memory** bound on what is already stored, where the upload cap gets no say — those rows predate it. Checked between `Image.open` and `convert()`, so an oversized artifact never reaches the allocation. Re-using the upload cap here would blank icons that draw correctly on maps today; 24M px (~96 MB of RGBA, and decodes are sequential within a batch) lets anything plausible up to ~4900px render its real art while DoS-scale artifacts degrade to the placeholder. It bounds area rather than dimensions because cost tracks area: a 12000x160 strip is 1.9M px and cheaper than a 2048 square.

Alongside those:

- `validate_icon_upload` maps `DecompressionBombError` and `IndexError` to `ValueError`, which the route already turns into a 400.
- `_render_icon` catches `DecompressionBombError`, which fires from `open()` above ~179M px before the pixel check can measure anything, so that cell degrades instead of 500ing the sheet.
- The render works `_SPRITE_RENDER_BATCH` icons at a time: each batch's stored objects are fetched concurrently, then composited in a worker thread. The catalog has no size limit, so neither the read fan-out nor the icon bytes held at once may scale with it.
- The listing query that the render already runs now hands its rows down, so the render issues no second query and `_render_sprite_png` takes no session — which is what handing the composite to a thread needed to be true anyway.

## Impacts

- **API contract: none.** The PNG handlers have no `response_model` and their signatures are untouched; `openapi.json` and the SDKs are unchanged.
- **Behavior change on upload:** a PNG over 1024px on either axis is now rejected with 400 `PNG icon dimensions are too large`. Worth knowing before it reaches anyone uploading large source art.
- **Behavior change on render, bounded deliberately:** a stored icon above 24M pixels draws as the placeholder instead of its art. Nothing under that changes, so no icon a real map displays today is affected.
- No schema or config changes.

## Verification

```
cd backend && set -a && source ../.env.test && set +a && uv run pytest tests/test_map_sprites.py tests/test_phase_273_icon_csp.py tests/test_phase_273_svg_defusedxml.py tests/test_maps.py tests/test_maps_style_json.py -q
```

Twelve new tests, each failing against the code it targets for its own reason: the 4096px upload was accepted (`DID NOT RAISE`), the bomb-sized and IDAT-less uploads escaped as `DecompressionBombError` / `IndexError`, the stored 81M-px icon rendered its full art rather than the placeholder, the catalog read counted three round trips instead of one, the storage reads had no ceiling on how many ran at once, and the composite had no worker thread to run on. Two of them guard the property the render bound must not break — a 2048px and a 12000x160 stored icon still draw real art. The sheet route itself had zero coverage before this; it now asserts 200, `image/png`, and a decodable sheet one cell wide per icon.

### DIFF
--- a/backend/app/modules/catalog/maps/sprites.py
+++ b/backend/app/modules/catalog/maps/sprites.py
@@ -61,6 +61,12 @@ _PATH_TOKEN_RE = re.compile(
     r"[MmZzLlHhVvCcQqSsTtAa]|[-+]?(?:\d*\.\d+|\d+\.?)(?:[eE][-+]?\d+)?"
 )
 SPRITE_CELL_SIZE = 24
+# fix(#1428 codex r1): how many icons the sheet render holds in flight at once —
+# it bounds both the concurrent storage reads and the icon bytes resident while
+# they are composited. The catalog has no size limit, so this cannot be "all of
+# them": 8 icons at MAX_ICON_BYTES is 4 MiB, and still 8x the serial read it
+# replaced.
+_SPRITE_RENDER_BATCH = 8
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1430, opened 2026-08-12
refactor(platform): move SSRF safe-client module out of catalog

ishiland:
## The conflict this resolves

`backend/app/modules/catalog/sources/security.py` is the repo's single SSRF chokepoint: `make_safe_client()`, `make_safe_transport()`, and `validate_url_for_ssrf()`. AGENTS.md Rule 2 mandates it, a pre-commit hook enforces it, and 19 files import it, including `auth/oauth`, `platform/config_ops`, and five files under `processing/ingest`.

It also sat inside a catalog sub-domain, which made two codified rules contradict each other. `_PROCESSING_CATALOG_IMPORT_BURNDOWN` in `backend/tests/test_layering.py` listed `app.modules.catalog.sources.security` as shrink-only layering debt for four `processing/` files, and its failure message tells the next person to "Route the behavior through ProcessingPort instead". That indirection is what would stop the Rule 2 grep hook from matching at the import site. `processing/tiles/router.py:226` even carried a forward reference pointing a future security fix at the path the layering rule wanted gone.

Both rules are right. The module was in the wrong place.

## The move

`git mv` to `backend/app/platform/security.py`. The module imports only the standard library and `httpx`, with no catalog models or services, and every layer that uses it (auth, platform, catalog, processing) already sits above platform. There is no re-export shim at the old path: one unambiguous home is the point of the move, and a shim would keep the burndown entries alive.

Behavior is unchanged. Git reports the file as a pure rename, 0 insertions and 0 deletions.

## Path-pinned surfaces updated

Moves break guards that pin the old path, so every hit in the repo was swept:

| Surface | What changed |
|---|---|
| `.pre-commit-config.yaml` | Rule 2 hook name, guidance message, the "canonical factory" comment, and the `exclude:` regex. The exclude is load-bearing: the hook's `files:` filter is `^backend/app/.*\.py$`, so without retargeting it the factory would fail its own rule. |
| `AGENTS.md` Rule 2 | New path, plus a paragraph on why the module stays in `platform/`, so it does not drift back under a product domain. |
| `backend/tests/test_layering.py` | Deleted the 5 now-dead `_PROCESSING_CATALOG_IMPORT_BURNDOWN` edges (4 `sources.security` entries plus `manifest_sources.py`'s `app.modules.catalog.sources`, whose only catalog import was this module). The list is shrink-only, so deleting is the legal direction. |
| `backend/app/processing/tiles/router.py` | The forward-reference comment aiming a future fix at `make_safe_client`. |
| `backend/app/modules/catalog/sources/cog_info.py` | The SEC-OBSV-02 dual-gate docstring naming the Gate 1 validator. |
| 19 app importers, 21 test files | Import lines and `mock.patch` target strings. |

Checked and deliberately left alone:

- `backend/tests/test_rule2_structural.py` has no pin on this path (its only path-keyed entry is `modules/catalog/sources/preview.py`).
- `backend/pyproject.toml` per-file-ignores has no entry for the module.
- `_MODULE_LOC_CAPS` has no entry either. The file is 226 lines, well under the 1000-line inclusion threshold, so no cap to retarget.
- One comment in `.pre-commit-config.yaml` still contains the substring `security.py`, describing a *historical* grep pattern that was removed in #430. It is not a path reference and reads correctly as-is.

## Verification

The intermediate failure is the discriminator that the burndown edit was actually required. With the file moved and the burndown untouched, `test_processing_catalog_import_allowlist_is_current` failed with exactly the five stale edges listed above:

```
Failed: _PROCESSING_CATALOG_IMPORT_BURNDOWN lists edges that no longer exist. Delete them - the list only shrinks.
  ingest/manifest_service.py: app.modules.catalog.sources.security
  ingest/manifest_sources.py: app.modules.catalog.sources
  ingest/router.py: app.modules.catalog.sources.security
  ingest/tasks_reupload.py: app.modules.catalog.sources.security
  ingest/tasks_vector.py: app.modules.catalog.sources.security
```

The Rule 2 hook still fires after the path change. A throwaway file containing a raw `httpx.AsyncClient(follow_redirects=True)` was placed at the new location and rejected (the file was deleted, never committed):

```
$ pre-commit run ssrf-safe-client --files backend/app/platform/_hookcheck_throwaway.py
FAIL: backend/app/platform/_hookcheck_throwaway.py constructs httpx.AsyncClient( with follow_redirects=True
  Use make_safe_client() from backend/app/platform/security.py

$ pre-commit run ssrf-safe-client --files backend/app/platform/security.py
...(no files to check)Skipped        # the retargeted exclude covers the factory
```

`test_rule2_structural.py` was also checked against a deliberate break: replacing `gdal_safe_env(extras=...)` with a bare dict in `processing/raster/cog.py` failed `test_gdal_cli_argv_uses_safe_env_or_is_allowlisted`, confirming the suite is live rather than vacuously green. Reverted.

Commands run:

```
cd backend && uv run pytest tests/test_layering.py tests/test_rule2_structural.py -q          # 138 passed
cd backend && uv run pytest tests/test_ssrf_redirect.py tests/test_ssrf_ip_pinning_sec008.py \
  tests/test_sec013_ssrf_cgnat.py tests/test_phase_273_cog_redirect_revalidate.py \
  tests/test_oauth_destination_security.py tests/test_oauth_github_provider.py \
  tests/test_source_health_1222.py -q                                                          # with the two above: 274 passed
cd backend && uv run pytest tests/test_commit_revalidates_source_url.py tests/test_services_endpoints.py \
  tests/test_stac_import.py tests/test_stac_refresh_1266.py tests/test_refresh_gate_1269.py \
  tests/test_service_refresh_1220.py tests/test_service_column_info_1359.py \
  tests/test_service_token_redaction_1277.py tests/test_probe_classification.py -q             # 296 passed
cd backend && uv run pytest tests/test_ingest.py tests/test_ingest_progress.py \
  tests/test_manifest_apply_roundtrip.py tests/test_manifest_apply_service.py tests/test_reupload.py \
  tests/test_reupload_service.py tests/test_302_quota_reservation.py tests/test_vrt_lifecycle_188.py -q  # 183 passed
make openapi-check                                                                             # no drift
```

753 targeted tests pass. No schema, API, or config impact: this is an internal module relocation with no behavior change, and the OpenAPI snapshot is unaffected.


ishiland:
@codex review

PR #1431, opened 2026-08-12
refactor(seams): move the rate limiter, the role lookup, and the embed-token reads to their owning layers

ishiland:
Three structural moves. No behavior change, no API/schema/config impact — `backend/openapi.json` is byte-identical (`dump_openapi.py --check` exits 0 with no regen).

## The rate limiter no longer lives in an API-edge router

`limiter` was defined in `app/modules/auth/router.py`, a module with 28 route decorators, and ten modules imported it at load time: `settings/router.py`, `settings/router_public.py`, `admin/router.py`, `admin/router_operations.py`, `catalog/datasets/api/router_data.py`, `catalog/search/router.py`, `processing/tiles/router.py`, `processing/ai/router.py`, `processing/ai/query_router.py`, and `api/main.py`, which binds it to `app.state.limiter`. Importing a rate limiter executed those 28 route registrations plus auth's whole transitive import graph.

It now lives in `app/platform/ratelimit.py` with `_global_rate_limit`, the persistent-config-driven default callable. Measured in a fresh interpreter: importing the limiter's home loaded 49 `app.*` modules before and loads 26 after, with zero `app.modules.*`. Every importer (app code and tests) points at the new home; auth's router keeps its own `@limiter.limit` decorators and its login-specific limit.

This is the same defect class `fix(#836)` closed from the other side, where `test_platform_never_imports_processing_routers` stopped the platform ports importing `processing/ingest/router.py` for a constant.

## `get_user_roles` moved to the domain that owns the tables

It sat in `catalog/authorization.py` but selects `Role` joined to `UserRole` and touches nothing catalog, and `auth/dependencies.py` imported it back — so the two highest-fan-in modules in the backend depended on each other at domain level. The query moves to `auth/permissions.py`, next to `user_has_capability`, which was resolving roles through a function-local "LAZY — avoids circular" import of that same function. That workaround is gone with the cycle it was working around.

`catalog/authorization.py` keeps `get_user_roles` as a delegating def, so its ~40 callers and their import path are unchanged. It stays a `def` rather than a bare re-export for two reasons: `test_permission_chokepoints_use_extension` reads that `async def` as the end of the `apply_visibility_filter` block it inspects, and the tests that patch the catalog-side name keep working through it.

## Maps reads embed tokens through a seam instead of the ORM

`maps/service_public.py` imported the `EmbedToken` ORM and queried it inline twice: the active token's `allowed_origins`, which becomes the shared-map `frame-ancestors` CSP directive, and the per-map active-token count in the admin share-token listing.

The reverse direction already has this seam and documents why — `maps/sharing.py` exists so embed-token code can ask map questions without importing catalog ORM models, which is what keeps the pair free of an import cycle. This adds the mirror, `embed_tokens/sharing.py`, exposing `get_active_allowed_origins()` and `active_embed_count_subquery()`. The CR-04 rationale for why the expiry predicate must admit `NULL` explicitly now lives with the query rather than at one call site.

The other maps-to-embed_tokens edges stay, because none of them reach for the ORM: `maps/router.py` and `router_sharing.py` call the embed-tokens service to revoke tokens, and `maps/schemas.py` and `sharing_policy.py` import an error-message constant from its schemas.

## Tests

`backend/tests/test_module_import_seams.py` asserts both import seams in a fresh subprocess — inside pytest every module is already loaded, so a `sys.modules` check there would pass regardless of where anything lives. Both fail on `main` (49 modules / auth router present; `catalog.authorization` present), as does the delegation test. Two focused cases in `test_embed_framing_csp.py` pin that the seam returns the active token's origins and skips a revoked one, and `None` rather than `[]` when there is no token.

## Verification

```
cd backend && uv run pytest tests/test_layering.py tests/test_module_import_seams.py -q   # 43 passed
cd backend && uv run pytest tests/test_rate_limits.py tests/test_admin_rate_limit.py tests/test_middleware.py \
  tests/test_health.py tests/test_register_verification_flow.py tests/test_event_signup_health.py \
  tests/test_sandbox_query_endpoint_565.py -q                                            # 137 passed
cd backend && uv run pytest tests/test_permission_extension.py tests/test_batch_dataset_authz_1298.py \
  tests/test_vrt_creation_173.py tests/test_vrt_source_management_174.py tests/test_defer_orphan_guard.py \
  tests/test_permission_audience.py tests/test_processing_port.py tests/test_tenant_cache_partitioning.py \
  tests/test_auth.py -q                                                                  # 169 passed
cd backend && uv run pytest tests/test_embed_framing_csp.py tests/test_embed_tokens.py \
  tests/test_embed_frame_policy.py tests/test_embed_tokens_csp_no_wildcard.py tests/test_admin_share_token_sort.py \
  tests/test_embed_tokens_origin_bypass.py tests/test_embed_token_expiry_cache.py tests/test_csp_sec020.py -q
                                                                                         # 112 passed
cd backend && PYTHONPATH=. uv run python scripts/dump_openapi.py --check                 # exit 0, no drift
```

### DIFF
--- a/backend/app/api/main.py
+++ b/backend/app/api/main.py
@@ -44,7 +44,7 @@ from app.platform.extensions.bootstrap import (
 )
 from app.modules.auth.models import Role, User, UserRole
 from app.modules.auth.providers.local import hash_password
-from app.modules.auth.router import limiter
+from app.platform.ratelimit import limiter
 from app.processing.ingest.tasks import task_app
 from app.api.middleware.body_limit import RequestBodyLimitMiddleware
 from app.api.middleware.cors import DynamicCORSMiddleware
--- a/backend/app/modules/admin/router.py
+++ b/backend/app/modules/admin/router.py
@@ -41,7 +41,7 @@ from app.modules.admin.service import (
 from app.modules.quota.service import get_user_quota_usage_bulk
 from app.modules.audit.service import AuditEvent, audit_emit, audit_emit_durable
 from app.modules.auth.dependencies import require_mode_permission, require_permission
-from app.modules.auth.router import limiter  # HARDEN-01: shared rate-limiter instance
+from app.platform.ratelimit import limiter  # HARDEN-01: shared rate-limiter instance
 from app.modules.auth.models import User
 from app.modules.auth.schemas import UserResponse
 from app.processing.export.service import safe_content_disposition
--- a/backend/app/modules/admin/router_operations.py
+++ b/backend/app/modules/admin/router_operations.py
@@ -26,7 +26,7 @@ from app.modules.admin.service import AdminService
 from app.modules.audit.service import AuditEvent, audit_emit
 from app.modules.auth.dependencies import require_mode_permission, require_permission
 from app.modules.auth.models import ApiKey, User
-from app.modules.auth.router import limiter
+from app.platform.ratelimit import limiter
 from app.modules.auth.schemas import ApiKeyCreateResponse
 
 router = APIRouter()
--- a/backend/app/modules/auth/dependencies.py
+++ b/backend/app/modules/auth/dependencies.py
@@ -17,7 +17,7 @@ from app.core.config import settings
 from app.core.dependencies import get_db
 from app.core.identity import Identity
 from app.modules.auth.models import ApiKey, User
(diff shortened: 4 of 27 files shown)


ishiland:
@codex review

PR #1432, opened 2026-08-12
fix(stac): clamp over-limit item pages and share one bbox parser

ishiland:
## What

Two STAC conformance gaps, both cases of the same handler being left out of an earlier pass.

**Over-limit pages were rejected, not clamped.** `GET /stac/collections/{id}/items` declared `limit` with `le=200`, so `limit=1000` came back as a 400 while both Item Search handlers clamped the same value and returned 200. The STAC Item Search spec, and `stac-api-validator` with it, require an over-maximum limit to be clamped to the server maximum. The ceiling now lives in one module constant, `_STAC_MAX_LIMIT`, read by all three item-returning handlers, so the next change to it cannot miss one.

**STAC hand-rolled bbox parsing at two sites** instead of calling the shared `parse_bbox`, and the copies had drifted. STAC accepted a zero-height bbox that OGC Features, per-dataset features, export and catalog search all rejected with a 400. That duplication had already cost a security guard once: the SEC-FU-06 NaN/Inf check had to be re-applied to the STAC copy (#430 BA-12) because the copy existed. Both sites now call `parse_bbox`, which takes the POST body's list of floats as well as the GET query string. `SearchParams` carried a third copy of the same logic, without the finite guard, and now validates through `parse_bbox` too.

The degenerate-box disagreement is settled toward the spec rather than toward the majority. OGC API Features and STAC both define bbox bounds as lower <= upper, so a zero-height (line) or zero-area (point) box is a legal filter. `parse_bbox` relaxes `miny >= maxy` to `miny > maxy`. Antimeridian crossing (`west > east`) is unchanged, and the finite-coordinate guard still applies to every caller from its single home.

The one bbox validator deliberately left alone is `AnalysisPreviewRequest._validate_bbox`: it rejects antimeridian crossing on purpose, because it feeds `ST_MakeEnvelope` directly with no two-envelope split, and its comment says so. It already uses `>` and already has the finite guard, so it agrees with the new shared rule where the rules overlap.

## API contract impact

- `GET /stac/collections/{id}/items`: the `limit` parameter loses its `maximum: 200` constraint and gains the clamp description. A value above 200 now returns 200 with at most 200 items instead of a 400. Values below 1 are still rejected (`ge=1` is untouched).
- **Behavior relaxation across five surfaces**: a zero-height or zero-area bbox goes from 400 to 200 on OGC Features, per-dataset features, export, catalog search and saved-search params. STAC already accepted it. Nothing that was accepted before is rejected now.
- Regenerated artifacts committed: `backend/openapi.json`, `sdks/python/geolens/api/stac/get_collection_items_stac_collections_collection_id_items_get.py`, `sdks/typescript/src/client/types.gen.ts`, `frontend/src/types/api.generated.ts`. The whole regen diff is the one parameter above.
- `backend/tests/test_layering.py`: the `stac/router.py` LOC cap drops 1870 -> 1855, since deleting the two inline parsers and `_require_finite_bbox` shrinks the file.

## Verification

Failing-first, on the pre-fix tree:

```
7 failed, 6 passed
AssertionError: surfaces disagree on a zero-height bbox:
  {'stac_items': 200, 'stac_search': 200, 'ogc_items': 400, 'catalog_search': 400}
AssertionError: {"title":"Bad Request","status":400,
  "detail":"query.limit: Input should be less than or equal to 200"}
```

After:

```
  tests/test_stac_limit_bbox_conformance.py tests/test_stac_api.py \
  tests/test_stac_integration.py tests/test_stac_search_validation.py \
  tests/test_stac_search_dos_sec023.py tests/test_stac_validation.py \
  tests/test_stac_visibility.py tests/test_stac_datetime_filter.py \
  tests/test_stac_record_output.py tests/test_stac_serializer.py \
  tests/test_parse_bbox_isfinite.py -q                                  # 137 passed

uv run pytest tests/test_ogc_features.py tests/test_ogc_pagination.py \
  tests/test_ogc_discovery.py tests/test_ogc_public_access.py \
  tests/test_ogc_records_conformance.py tests/test_ogc_cql2_filtering.py \
  tests/test_ogc_collection_metadata.py tests/test_ogc_queryables.py -q  # 131 passed

uv run pytest tests/test_export*.py tests/test_search*.py \
  tests/test_saved_searches.py tests/test_features_crud.py \
  tests/test_antimeridian_extent.py tests/test_antimeridian_rollup.py -q # 293 passed

uv run pytest tests/test_layering.py -q                                  # 40 passed
uv run ruff check app tests && uv run ruff format app tests --check      # clean
make openapi-check && make sdks-check                                    # clean
```

### DIFF
--- a/backend/app/modules/catalog/features/service.py
+++ b/backend/app/modules/catalog/features/service.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import json
 import math
 import re
+from collections.abc import Sequence
 from typing import TYPE_CHECKING
 
 from shapely.errors import GEOSException
@@ -93,8 +94,13 @@ def _geom_write_exprs(
     return base, base
 
 
-def parse_bbox(bbox_str: str) -> list[float]:
-    """Parse a comma-separated bbox string into a 4- or 6-element list.
+def parse_bbox(bbox: str | Sequence[float]) -> list[float]:
+    """Parse a bbox into a 4-element ``[minx, miny, maxx, maxy]`` list.
+
+    Accepts either a comma-separated string (the query-parameter spelling) or
+    an already-split sequence of numbers (the JSON request-body spelling, e.g.
+    STAC Item Search POST), so every GeoLens surface answers a given bbox the
+    same way.
 
     Accepts:
       - 4 values: minx, miny, maxx, maxy (2D)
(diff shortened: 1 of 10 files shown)


ishiland:
@codex review

PR #1434, opened 2026-08-12
fix(api): correct sprite sdf typing, read-GET 403 docs, and tag casing

ishiland:
## Summary

Three OpenAPI contract-accuracy fixes, all in the wire contract:

- **Sprite `sdf` boolean type.** `router_assets.py`'s two sprite-index GET
  handlers annotated their return type as `dict[str, dict[str, int | float]]`
  — narrower than `sprites.py`'s own `SpriteIndex` type, which includes
  `bool`. FastAPI infers the response model from the return annotation, and
  since `bool` is a subclass of `int`, Pydantic silently coerced the
  arrow-right builtin icon's `sdf: true` to the JSON integer `1` on the wire.
  MapLibre's sprite spec defines `sdf` as boolean. Both handlers now annotate
  `-> SpriteIndex` (imported from `sprites.py`), matching the service's own
  type instead of a narrower ad hoc one.

- **Read-gated GETs publishing a write-flavored 403.** Routers built with
  `responses=ERROR_RESPONSES_WRITE` document every operation's 403 as
  "Forbidden — caller lacks write access" by default — correct for mutation
  endpoints, wrong for GETs gated by visibility/read access (or nothing at
  all). The override pattern already existed at 4 sites in
  `datasets/api/router.py` (`responses={403: FORBIDDEN_RESPONSE}`); this PR
  walks every `ERROR_RESPONSES_WRITE` router's GET operations and applies the
  same override to every one that isn't actually write-gated. Landed **32**
  new override sites across 10 routers (`datasets/api/router.py`,
  `router_metadata.py`, `router_vrt.py`, `router_data.py`, `maps/router.py`
  + its `router_assets.py`/`router_sharing.py` sub-routers, `records/router.py`,
  `collections/router.py`, `processing/ingest/router.py`) — plus the 4
  pre-existing sites, that's 36 read-gated GETs now correctly documented.
  GETs that genuinely require write/ownership access keep the accurate
  default (e.g. map history and embed-token listing require
  `edit_metadata` + ownership; connector discovery and the icon catalog
  require an upload/edit capability; `validate_dataset`'s `?refresh=true`
  path enforces ownership). Verified by walking the live FastAPI route table
  post-fix: zero GET operations incorrectly publish the write-flavored 403.

- **Two OpenAPI tag-casing outliers.** `config-ops` → `Config Ops` (24
  sibling tags are Title Case; also updated the matching `openapi_tags`
  metadata entry in `api/main.py` so the Swagger UI description stays
  attached) and `Datasets - Source health` → `Datasets - Source Health`
  (its seven `Datasets - *` siblings are Title Case).

## API contract impact

- `sdf` changes from `anyOf[integer, number]` to `anyOf[integer, number,
  boolean]` on `GET /maps/sprites/geolens.json`'s response schema — additive,
  no client break (any client already tolerant of `1`/`0` still accepts
  `true`/`false`).
- 32 GET operations' documented 403 description changes from "Forbidden —
  caller lacks write access" to "Forbidden — caller lacks access to this
  resource" — docs-only, no runtime/status-code change.
- 2 tag renames regroup the affected operations' SDK methods under
  `ConfigOpsService`/`Config Ops` and the renamed `Datasets - Source Health`
  namespace instead of the old casing.
- `backend/openapi.json`, both SDKs (`sdks/python`, `sdks/typescript`), and
  the frontend's generated client (`frontend/src/types/api.generated.ts`)
  are regenerated and committed.

## Test plan

- [x] Failing-first: `test_sprite_index_endpoint_serves_sdf_as_json_boolean`
      (new, `test_map_sprites.py`) — verified it fails against the pre-fix
      annotation (asserts JSON integer `1`) and passes after.
- [x] Failing-first: `test_read_gated_gets_do_not_publish_the_write_403_description`
      and `test_config_ops_and_source_health_tags_are_title_case` (new,
      `test_api_contract_openapi.py`) — verified both fail against the
      pre-fix router/tag state and pass after.
- [x] `cd backend && uv run ruff check app tests && uv run ruff format app tests --check`
- [x] `uv run pytest tests/test_layering.py -q` (3 router modules grew past
      their LOC cap from the per-route `responses=` additions; ratcheted in
      the same commit with inline comments)
- [x] `uv run pytest tests/test_map_sprites.py tests/test_api_contract_openapi.py tests/test_config_ops.py tests/test_config_ops_unit.py tests/test_openapi_overlay_boundary.py tests/test_openapi_property_order.py tests/test_reupload_completion_contract_1207.py tests/test_secondary_contract_bounds.py -q` — 119 passed, 1 skipped
- [x] `uv run pytest tests/test_maps.py tests/test_maps_bulk_layers.py tests/test_maps_og_image.py tests/test_maps_style_json.py tests/test_maps_visibility_check_authz_sec007.py tests/test_collections.py tests/test_records_authz_sec001.py tests/test_records_related.py -q` — 394 passed (behavioral regression check on the most heavily-touched routers)
- [x] `make openapi-check` / `make sdks-check` (incl. SDK build + test) / `cd frontend && npm run types:check` — all clean, no drift

### DIFF
--- a/backend/app/api/main.py
+++ b/backend/app/api/main.py
@@ -570,7 +570,7 @@ _OPENAPI_TAGS = [
         "description": "Saved map configurations, layers, AI styling, and sharing.",
     },
     {
-        "name": "config-ops",
+        "name": "Config Ops",
         "description": "Configuration export, import, dry-run, and connectivity validation.",
     },
     {
--- a/backend/app/modules/catalog/collections/router.py
+++ b/backend/app/modules/catalog/collections/router.py
@@ -44,7 +44,7 @@ from app.modules.catalog.collections.service import (
 from app.modules.catalog.datasets.domain.helpers import dataset_to_response
 from app.modules.catalog.datasets.domain.schemas import DatasetListResponse
 from app.core.dependencies import get_db
-from app.standards.ogc.errors import ERROR_RESPONSES_WRITE
+from app.standards.ogc.errors import ERROR_RESPONSES_WRITE, FORBIDDEN_RESPONSE
 
 router = APIRouter(
     prefix="/catalog/collections", tags=["Datasets"], responses=ERROR_RESPONSES_WRITE
@@ -131,8 +131,17 @@ _CATALOG_CACHE_TTL = 60  # seconds
 
 
 # ROUTE-01 (Phase 1092): dual-shape decorator — see POST above.
-@router.get("", response_model=CollectionListResponse, include_in_schema=False)
-@router.get("/", response_model=CollectionListResponse)
+@router.get(
+    "",
+    response_model=CollectionListResponse,
+    include_in_schema=False,
+    responses={403: FORBIDDEN_RESPONSE},
+)
+@router.get(
+    "/",
+    response_model=CollectionListResponse,
+    responses={403: FORBIDDEN_RESPONSE},
+)
 async def list_collections_endpoint(
     skip: int = Query(0, ge=0),
     limit: int = Query(50, ge=1, le=200),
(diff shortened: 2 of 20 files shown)


ishiland:
@codex review

PR #1436, opened 2026-08-12
fix: gathered dataset-detail sessions and an embed-token auth-path commit

ishiland:
## Summary

Two unrelated transaction-hygiene fixes found in review.

**F12 — `get_dataset_detail` gather ran on one shared session, not in parallel.**

`get_dataset_detail` fetches a raster asset, a VRT source-link count, and dataset-asset rows via `asyncio.gather`, with a comment claiming they run "in parallel through CatalogPort." All three coroutines were bound to the same shared `db` session. `AsyncSession` is not safe for concurrent use — the repo states this rule explicitly in `search/router.py` and every other gather site in the codebase (`search/router.py`, `stac/router.py`, `ingest/router.py`) already follows it by opening a fresh session per branch. Here, asyncpg's per-connection execute lock silently serialized the three branches instead of running them concurrently (measured: three `pg_sleep(0.3)` calls gathered on one session took 0.916s; on three sessions, 0.307s). Beyond the performance claim being false, sharing a session across a gather is latently unsafe — a different dialect or a tightened SQLAlchemy concurrency guard would turn this into an `InvalidRequestError` on `GET /datasets/{id}` for every raster or VRT dataset.

Fix: each branch now opens its own short-lived session via `app.core.db.async_session`, matching the established pattern. I took the "give each branch its own session" option rather than dropping the gather and going sequential — the three fetches only read plain, already-loaded columns downstream (no relationship traversal), so detaching them from their fetch session is safe under the sessionmaker's `expire_on_commit=False`, and this preserves the genuine-concurrency benefit the original comment was aiming for. `service_query.py`'s ratcheted line cap moves 438 -> 443 (exact) in the same commit.

**F13 — `validate_embed_token_access` committed the caller's session from a read-path auth helper.**

After authorization passed, the helper bumped `use_count`/`last_used_at` via `async with db.begin_nested(): ...` then `await db.commit()` **on the session its caller owns**. Every caller today is a read path (`features/router.py`, `tiles/router.py`), so this was benign in practice, but a read-path authorization helper committing the caller's transaction would silently flush whatever else that transaction was carrying the moment a write endpoint gates on an embed token. The codebase already solves the identical problem the opposite way one module over: `_resolve_api_key`'s `ApiKey.last_used_at` bump in `auth/dependencies.py` uses a dedicated side session specifically so an early commit doesn't release advisory locks or persist uncommitted state ahead of the route's own commit decision.

Fix: the bump now runs on its own side session via `app.core.db.async_session`, mirroring `_resolve_api_key` exactly. The helper no longer touches the caller's transaction at all (no commit, no `begin_nested` on `db`). Bump semantics are unchanged (still gated on the cache-miss `token is not None` branch, still one increment per DB lookup).

That side-session change surfaced a real test-infra gap: `TestEmbedTokenTenantPinning` exercises the cache-miss bump path from tests that otherwise stay entirely on their own raw, per-test `create_async_engine(...NullPool...)` instances (deliberately, for tenant-RLS isolation). The new side session was the first thing to bind a connection from the app's shared, pooled engine to one of those test-scoped event loops — the same "got Future attached to a different loop" hazard `test_worker`/`test_schema_skew_guard` already guard against in `conftest.py`. Reproduced it 3/3 runs before the fix, 3/3 clean after. Fixed by adding `test_embed_tokens` and `test_embed_token_expiry_cache` to `conftest.py`'s `_DISPOSE_SHARED_ENGINE_MODULES`, the same established mechanism those two modules already use.

## Contract impact

None — no OpenAPI/schema/config changes. Both fixes are internal transaction-handling changes with no behavior change to API responses.

## Verification

```
  tests/test_embed_tokens.py tests/test_builder_audit_p0p1_sharing.py \
  tests/test_embed_tokens_origin_bypass.py tests/test_embed_token_expiry_cache.py \
  tests/test_embed_revocation_by_map.py tests/test_tenant_cache_partitioning.py \
  tests/test_vrt_catalog_175.py tests/test_sql_engine.py tests/test_rule1_structural.py -q
```

Each new test was confirmed failing-first: reverted its corresponding production fix locally and re-ran to see the specific assertion fail, then restored the fix and re-ran clean.

### DIFF
--- a/backend/app/modules/catalog/datasets/domain/service_query.py
+++ b/backend/app/modules/catalog/datasets/domain/service_query.py
@@ -223,9 +223,9 @@ async def get_dataset_detail(
     # checkout while the caller's own connection is held for the rest of this
     # request: under the default (non-external-pooler) pool of 10 + 3
     # connections, ~13 concurrent raster/VRT detail requests exhaust it
-    # (codex review). These are three fast, single-row/point lookups, so the
-    # wall-clock cost of running them in sequence on the connection this
-    # request already holds is negligible next to that risk.
+    # (fix(#1436) codex review). These are three fast, single-row/point
+    # lookups, so the wall-clock cost of running them in sequence on the
+    # connection this request already holds is negligible next to that risk.
     record_type = getattr(dataset.record, "record_type", None)
     needs_raster = record_type in RASTER_FAMILY_RECORD_TYPES
     needs_vrt_count = record_type == "vrt_dataset"
--- a/backend/app/modules/catalog/datasets/domain/service_query.py
+++ b/backend/app/modules/catalog/datasets/domain/service_query.py
@@ -2,7 +2,6 @@
 
 from __future__ import annotations
 
-import asyncio
 import uuid
 from typing import TYPE_CHECKING, Any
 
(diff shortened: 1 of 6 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1437, opened 2026-08-12
test: consolidate six copy-pasted fixture/helper clusters in backend/tests

ishiland:
## Summary

Six mechanical DRY consolidations across `backend/tests/`. No production code
touched, and no test coverage removed — same assertions run against the same
data, just built through one shared implementation per cluster instead of N
divergent copies.

**1. `enterprise_migrations_present` (9 → 1)** — the SAML column drift gate's
guard read `version_locations` from `alembic.ini` at rest, which never
contains that key (it's set at runtime by `conftest.py`/`env.py`), so the
guard was permanently `False`. Eight sibling migration-test modules already
had the correct implementation (enumerate the `geolens.migrations`
entry-point group). Lifted one canonical copy into `tests/alembic_helpers.py`;
all nine modules now import it.

**2. `_init_tile_pool_for_tests` (5 → 1)** — byte-identical asyncpg pool setup
for tile-serving tests, previously requiring a retry fix to be hand-applied at
three of five copies separately. Moved into `conftest.py` as one non-autouse
fixture; the four class-scoped consumers keep `@pytest.mark.usefixtures`,
and `test_embed_tokens.py`'s file-scoped `autouse=True` becomes a
module-level `pytestmark` with the same mark, preserving its request-based
guard.

**3. Extension-registry reset (13 `_reset_registry` + 10 `_clean_registry` →
1 each)** — `_reset_registry` cleared varying subsets of four
`app.platform.extensions` globals across files (7 cleared two of four, 3
added a third, 3 worker-bootstrap files cleared all four) — a latent
cross-test pollution bug waiting for a file to touch a slot its own
copy-pasted subset didn't anticipate. One `_reset_registry` in `conftest.py`
now clears all four; the entry-points-patching `_clean_registry` fixture
that 8 of 13 files shared verbatim also collapsed into `conftest.py`. Two
files whose `_clean_registry` doesn't patch entry points, and three
worker-bootstrap files with their own combined registry+edition reset,
keep local fixtures but import the shared `_reset_registry`.

**4. Admin-identity resolution (15 `_get_admin_id` bodies, 3 shapes → 1) +
a boot-time guard** — bodies hardcoded `"admin"`, queried
`settings.geolens_admin_username`, or delegated to
`tests.factories.get_user_id`. That setting has no default, and roughly
1,160 other call sites elsewhere in the suite (~1,082 through
`factories.get_user_id`, ~79 querying the literal directly) assume it
resolves to `"admin"` — a mismatch would fail every one of those sites
individually with `NoResultFound` and no shared root cause. Collapsed all
15 onto `factories.get_user_id`, and added one assert in `conftest.py`'s
session-scoped `_test_db_lifecycle` that fails loudly, once, if the setting
is ever not `"admin"`. The ~1,160 direct call sites are unchanged — the
assert is what makes them safe, not a rewrite.

**5. Dataset factories (13 `_create_dataset` + 10 `_create_raster_dataset` +
7 `_fresh_query` → 3 canonical helpers)** — converted the vector copies to
thin wrappers over `factories.create_dataset`, extending it with the
optional fields each file needed (every field applies only when passed, so
existing callers are unaffected). Added `factories.create_raster_dataset`
for the 10 raster copies, which had three different return shapes (`Dataset`,
`dataset.id`, or a `(record, dataset, raster_asset)` tuple) and divergent
`RasterAsset` field sets; its `theme_category` stays `NULL` unless a caller
explicitly sets it, unlike the vector factory, since several raster callers
depend on that. Moved the AUTOCOMMIT-connection `_fresh_query` helper (needed
because subprocess alembic commits outside the test session's transaction)
into `alembic_helpers.py`.

Diffing the vector copies against the extended factory surfaced a real bug
before it shipped: the factory's own `theme_category` default of `["test"]`
changed an OGC API response's `"keywords"` field for six files that
previously left it `NULL`. Caught by a real test failure, fixed by passing
`theme_category=[]` explicitly in those wrappers.

**6. `_create_test_user` (private-in-conftest, imported by ~25 files) → 1
public factory** — moved its body to `tests/factories.py` as `create_user`;
`conftest.py` keeps `_create_test_user = create_user` so every existing
`from tests.conftest import _create_test_user` site keeps working unchanged.

## Verification

- `cd backend && uv run ruff check tests && uv run ruff format tests --check` — pass across all 518 files
- `cd backend && set -a && source ../.env.test && set +a && uv run pytest tests/test_layering.py -q` — 40 passed (run after every commit)
- Per-finding and cumulative smoke runs after every commit: 1,062 tests collected with zero import errors across every touched file; full execution passed for all of them, including a broad slice re-run after rebasing onto four commits that landed on `main` mid-work (two of which relocated modules these tests import — no behavior change, verified empirically)

### DIFF
--- a/backend/tests/alembic_helpers.py
+++ b/backend/tests/alembic_helpers.py
@@ -32,12 +32,36 @@ from __future__ import annotations
 import os
 import subprocess
 import sys
+from importlib.metadata import entry_points
 from pathlib import Path
 
 import sqlalchemy
 
 _BACKEND_DIR = Path(__file__).parent.parent.resolve()
 _ALEMBIC_INI = _BACKEND_DIR / "alembic.ini"
+
+
+def enterprise_migrations_present() -> bool:
+    """True when an enterprise/overlay migrations entry-point is installed.
+
+    conftest then migrates the per-worker test DB to the enterprise head (e.g.
+    e002_add_saml_columns), making the alembic environment MULTI-HEAD. The
+    core-only ``alembic`` subprocess these OSS-drift-gate roundtrip/check tests
+    shell out to can neither locate the enterprise revision nor disambiguate
+    ``head`` / ``-1`` across branches — so they are skipped under the overlay.
+    They still run (and gate drift) in the no-overlay Pytest Parallel Isolation
+    job. Core registers no ``geolens.migrations`` entry-point, so this is False
+    for community/OSS runs.
+    """
+    for ep in entry_points(group="geolens.migrations"):
+        try:
+            fn = ep.load()
+            if callable(fn) and any(Path(p).is_dir() for p in fn()):
+                return True
+        except Exception:
+            pass
+    return False
+
 
 # The remediation SQL that 0030's own error message prescribes. Collapsing a
 # two-ring seam extent to its envelope LOSES the antimeridian-crossing shape
--- a/backend/tests/alembic_helpers.py
+++ b/backend/tests/alembic_helpers.py
@@ -211,3 +211,27 @@ def run_alembic(
         env=env,
         cwd=str(_BACKEND_DIR),
     )
+
+
+async def fresh_query(query: str, params: dict | None = None):
+    """Run a query on a fresh AUTOCOMMIT connection, bypassing the test transaction.
+
+    The ``test_db_session`` fixture holds an open transaction around each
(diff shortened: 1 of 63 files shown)


ishiland:
@codex review

PR #1438, opened 2026-08-12
refactor(stac): route raster-asset reads through CatalogPort

ishiland:
## What

`backend/app/standards/stac/router.py` reached into processing-owned ORM directly: `from app.processing.raster.models import DatasetAsset, RasterAsset` at module scope, plus a deferred `from app.processing.raster.queries import fetch_raster_meta_bulk` inside a helper. All four read sites now go through `CatalogPort`, the same seam `backend/app/modules/catalog/` uses for the same data.

| Read site | Before | After |
| --- | --- | --- |
| `_fetch_dataset_asset_rows` | `select(DatasetAsset).where(...)` | `get_catalog_port().list_dataset_assets(...)` |
| `_fetch_raster_meta` | `fetch_raster_meta_bulk(db, ids, include_vrt=False)` | `get_catalog_port().fetch_raster_meta_bulk_without_vrt(...)` |
| `_fetch_projection_codes` (collections list) | module-scope `RasterAsset` | `get_catalog_port().raster_asset_orm_class()` |
| `_fetch_projection` (single collection) | module-scope `RasterAsset` | `get_catalog_port().raster_asset_orm_class()` |

## Why

`test_layering.py::test_no_catalog_imports_processing` holds `backend/app/modules/catalog/` to a strict zero on `app.processing` references, and catalog pays a real ergonomic cost to honor it. `standards/` was never held to the same rule, so the backend's largest router sat between the two domains as an unguarded bridge welding catalog ORM to processing ORM. A refactor of `processing/raster/models.py` broke STAC output with no layering signal — the failure showed up as wrong STAC responses rather than as a gate.

## No output-shape change, and how that was verified

Pure read-path indirection. The SQL is unchanged: the port's `list_dataset_assets` issues the same `select(DatasetAsset).where(dataset_id.in_(...))` the router had inline, and the two projection queries are byte-identical with the ORM class resolved through the port instead of imported.

- Full STAC suite before the change: **282 passed**. After: **283 passed** (282 + the new structural pin). No test changed its expectations.
- `dump_openapi.py --check` exits 0 — no schema drift, no SDK regeneration.

The one narrowing that is load-bearing is `include_vrt=False`: `vrt_type` becomes an item property when present (`search/service_records.py`), so a STAC Item would grow a property if the reader took the wider answer. That is why the new port method exists rather than the existing one being reused.

## Port compatibility and the version bump

`fetch_raster_meta_bulk_without_vrt` is an **additive method** rather than an `include_vrt` keyword on the existing `fetch_raster_meta_bulk`, because widening an existing port method's signature is a bump under `platform/extensions/version.py` (the 2 -> 3 entry) and because it leaves the wider reading working unchanged for the callers that need `vrt_type`.

The additive shape is **not** an exemption from the bump, which review round 1 correctly caught. `EXTENSION_API_VERSION` goes **5 -> 6**: the method is required, not optional. Every STAC item and item-page response calls it — including an empty page — so an overlay that replaces the `catalog_port` slot without it would load cleanly at version 5 and then raise `AttributeError` on the first STAC request. The "do NOT bump for new optional methods" carve-out means methods carrying a default no-op; a Protocol method a structural implementer must supply is required by definition. That is the same reading that took the constant 4 -> 5 for `reconcile_distributions`.

`test_catalog_port.py` states it as a test, mirroring the 4 -> 5 precedent: a port missing the method does not satisfy `CatalogPort`, so the loader refuses the overlay at boot rather than serving an AttributeError later.

Worth noting for whoever updates the overlay packages: the only overlay that claims the `catalog_port` slot wraps the prior implementation as a pure `__getattr__` delegate, so it inherits the new method regardless of what version it declares.

## Structural pin

New `backend/tests/test_standards_layering.py` walks every module under `backend/app/standards/` as an AST and fails on any `app.processing` import at module **or** function scope. AST rather than `git grep` for two reasons: an import is a syntactic fact where a comment mentioning the module name is not, and `git grep` is blind to untracked files. Relative imports are resolved against the file's own package, so `from ...processing.raster.models import ...` is caught too.

On the pre-change tree it fails, naming both offenders:

```
backend/app/standards/stac/router.py:52: from app.processing.raster.models import DatasetAsset, RasterAsset
backend/app/standards/stac/router.py:248: from app.processing.raster.queries import fetch_raster_meta_bulk
```

After the change it passes, and `rg "app\.processing" backend/app/standards` returns nothing.

## One guard widened, deliberately

`test_raster_replace_1221.py::test_every_dataset_asset_consumer_crosses_the_boundary` pins that any module fetching `DatasetAsset` rows either applies the `is_public_asset_key` allowlist itself or delegates to the builder that does, so internal keys like `archived_original:*` cannot reach a response. Moving the STAC fetch onto the port brought that router into the guard's scope **for the first time** — it previously hand-rolled `select(DatasetAsset)` and matched neither fetch name, so the guard never looked at it.

The router does cross the boundary: its rows go only to `dataset_to_ogc_record` and `build_assets`, and both hand them to `_build_stac_assets`, which drops every non-public key. Nothing iterates the rows into a payload on its own terms. The accepted-name set gains `build_assets` — the public facade over the private builder, and the name a module outside `search/` can actually call.

Net effect on the guard: it now inspects three modules instead of two, each crossing the boundary. It got stronger, not weaker.

## Verification

```
uv run pytest tests/test_layering.py tests/test_standards_layering.py tests/test_catalog_port.py -q
uv run pytest tests/test_stac_*.py -q                       # 283 passed
uv run pytest tests/test_extensions.py tests/test_cloud_overlay_side_by_side.py \
  tests/test_worker_bootstrap_parity.py tests/test_worker_overlay_resolution.py \
  tests/test_entitlement_port.py tests/test_notification_port.py \
  tests/test_processing_port.py tests/test_workflow_extension.py \
  tests/test_datasets.py -q                                 # 159 passed, 12 skipped
uv run pytest tests/test_raster_replace_1221.py tests/test_vrt_catalog_175.py \
  tests/test_quicklook_predicate.py tests/test_phase_274_async_parallelization.py \
  tests/test_raster_dataset_assets_bug041.py tests/test_dp01_ingest_write_path.py -q
PYTHONPATH=. uv run python scripts/dump_openapi.py --check  # exit 0
```

LOC ratchets adjusted in the same commit: `stac/router.py` 1855 -> 1854, `defaults_catalog_port.py` 445 -> 450, both exact with a comment saying what the lines bought.

Backend-audit follow-up (F7).

### DIFF
--- a/backend/app/core/catalog_port.py
+++ b/backend/app/core/catalog_port.py
@@ -269,6 +269,19 @@ class CatalogPort(Protocol):
         self, session: AsyncSession, dataset_ids: list[uuid.UUID]
     ) -> dict[str, dict[str, Any]]: ...
 
+    # The same rows without the VRT assembly fields. `vrt_type` is forwarded to
+    # a record's properties when present (search/service_records.py), and the
+    # STAC item surface must not grow that property — so its reader asks for
+    # the narrower answer rather than trimming the wider one.
+    #
+    # A separate method rather than an `include_vrt` keyword on the call above:
+    # widening an existing port method's signature is an EXTENSION_API_VERSION
+    # bump under platform/extensions/version.py (the 2 -> 3 entry bumped for
+    # exactly that shape), while an additive method is explicitly not.
+    async def fetch_raster_meta_bulk_without_vrt(
+        self, session: AsyncSession, dataset_ids: list[uuid.UUID]
+    ) -> dict[str, dict[str, Any]]: ...
+
     # How many members ONE generation is assembling. fix(#1327): that is the
     # attempt's intended set, which for a source add/remove is not what the VRT
     # is serving until the attempt publishes — so this is a fact about the
--- a/backend/app/core/catalog_port.py
+++ b/backend/app/core/catalog_port.py
@@ -274,10 +274,13 @@ class CatalogPort(Protocol):
     # STAC item surface must not grow that property — so its reader asks for
     # the narrower answer rather than trimming the wider one.
     #
-    # A separate method rather than an `include_vrt` keyword on the call above:
-    # widening an existing port method's signature is an EXTENSION_API_VERSION
-    # bump under platform/extensions/version.py (the 2 -> 3 entry bumped for
-    # exactly that shape), while an additive method is explicitly not.
+    # REQUIRED, which is why EXTENSION_API_VERSION went 5 -> 6: every STAC
+    # item and item-page response calls it, so an overlay that replaces the
+    # `catalog_port` slot without it serves AttributeError instead of a page.
+    # A separate method rather than an `include_vrt` keyword on the call above
+    # because widening an existing port method's signature is a bump under the
+    # same rule (the 2 -> 3 entry), and this shape keeps the wider reading
+    # working unchanged for the callers that need `vrt_type`.
     async def fetch_raster_meta_bulk_without_vrt(
         self, session: AsyncSession, dataset_ids: list[uuid.UUID]
     ) -> dict[str, dict[str, Any]]: ...
--- a/backend/app/platform/extensions/defaults_catalog_port.py
+++ b/backend/app/platform/extensions/defaults_catalog_port.py
@@ -419,6 +419,11 @@ class DefaultCatalogPort:
 
         return await fetch_raster_meta_bulk(session, dataset_ids)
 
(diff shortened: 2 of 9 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1440, opened 2026-08-12
fix(api): correct sprite-route error contract; dedupe test helpers

ishiland:
## Summary

Three small residuals surfaced by this wave's own reports, bundled since each is a one- or two-file cleanup.

**Sprite-route 401/403/409 contract fix.** The four map sprite routes (`geolens.json`, `geolens@2x.json`, `geolens.png`, `geolens@2x.png`) take only `Depends(get_db)` — no auth at all — yet inherited the maps router's write-flavored `ERROR_RESPONSES_WRITE` and published a 401/403/409 contract they can never emit. PR #1434 had reworded the JSON routes' 403 to the read-flavored description, but couldn't remove it.

Investigating why a route-level `responses={}` override can't drop an inherited key led to a second finding: FastAPI's router `responses=` merge is additive along the *whole* include chain, not just the immediate parent. A same-file sibling router (`sprites_router`, built with `responses=ERROR_RESPONSES_PUBLIC`) nested under the maps router still leaked 401/403/409 through, because the maps router's `ERROR_RESPONSES_WRITE` is unioned in at every level below it for any key the descendant route doesn't itself set. Verified against the FastAPI 0.135.2 source (`_RouterIncludeContext.for_include`/`.combine`) and empirically against the generated `openapi.json`.

The fix moves the four sprite routes onto `sprites_router`, mounted with `ERROR_RESPONSES_PUBLIC` directly on `api_router` (`backend/app/api/router.py`) as a **sibling** of the maps router rather than a descendant, so the write-flavored defaults never enter the chain. Paths, tags, and operation IDs are unchanged — only the two published sprite operations' response docs changed (401/403/409 dropped, 400's wording corrected to the public-profile text). `backend/app/modules/catalog/maps/router_assets.py`'s reviewed-cap ratchet in `test_layering.py` is bumped 132 → 142 with an explanatory comment.

The icon-asset endpoint (`/maps/icons/{icon_id}/asset`) has the same unauthenticated-yet-403-documented shape but is intentionally left untouched here — out of scope for this PR, noted for a follow-up.

**Test helper dedup.** `test_raster_tiles.py` carried a local `_get_auth_header` byte-identical to the one in `test_vector_tile_auth.py` and functionally identical to `tests.conftest.get_auth_header`. `test_raster_tiles.py` now imports the shared helper (matching `test_auth.py`, `test_ingest.py`, `test_api_key_scope_875.py`). `test_vector_tile_auth.py`'s copy turned out to have zero call sites in that file (it only exercises anonymous-access denial), so it's deleted outright rather than replaced with an unused import.

**Vestigial migration-test constants.** `_BACKEND_DIR`/`_ALEMBIC_INI` in five migration test files were dead — leftover from a guard that predates the current `alembic_helpers`-based approach. Removed, along with the now-unused `pathlib.Path` import each was the sole user of.

## Contract impact

`backend/openapi.json`, both SDKs, and `frontend/src/types/api.generated.ts` are regenerated. The only behavioral change is documentation: the sprite JSON/PNG GET operations no longer claim 401/403/409 responses they could never return, and their 400 description now matches the public-endpoint wording used elsewhere (e.g. dataset export). No runtime behavior changed — verified via `test_map_sprites.py` (all 14 sprite tests) and `test_api_contract_openapi.py`.

## Verification

- `cd backend && uv run ruff check app tests && uv run ruff format app tests --check`
- `cd backend && set -a && source ../.env.test && set +a && uv run pytest tests/test_layering.py -q` — 40 passed
- `cd backend && set -a && source ../.env.test && set +a && uv run pytest tests/test_map_sprites.py tests/test_raster_tiles.py tests/test_vector_tile_auth.py tests/test_tenant_rls_migration.py tests/test_email_verification_migration.py tests/test_embedding_vector_migration.py tests/test_oauth_github_migration.py tests/test_dormant_tenancy_migration_roundtrip.py tests/test_api_contract_openapi.py -q` — 140 passed
- `make openapi-check` and `make sdks-check` (including the TypeScript SDK's own build + tests) — both pass clean

### DIFF
--- a/backend/app/api/router.py
+++ b/backend/app/api/router.py
@@ -41,6 +41,7 @@ from app.processing.ingest.router import router as ingest_router
 from app.platform.jobs.router import router as jobs_router
 from app.modules.catalog.layers.router import layers_router
 from app.modules.catalog.maps.router import router as maps_router
+from app.modules.catalog.maps.router_assets import sprites_router as maps_sprites_router
 from app.standards.ogc.router import ogc_features_router, ogc_router
 from app.modules.catalog.records.router import router as records_router
 from app.modules.catalog.search.router import collections_router, search_router
@@ -87,6 +88,10 @@ api_router.include_router(collections_crud_router)
 api_router.include_router(ogc_features_router)
 
 api_router.include_router(maps_router)
+# fix: sprites are unauthenticated reads; mounted as a sibling of maps_router
+# (not nested inside it) so they don't inherit its ERROR_RESPONSES_WRITE
+# defaults — see the comment on sprites_router in maps/router_assets.py.
+api_router.include_router(maps_sprites_router)
 api_router.include_router(ai_router)
 # feat(#565): raw sandbox SQL endpoint (POST /query/), consumed by the MCP
 # `query` tool. Auth-only; see processing/ai/query_router.py.
--- a/backend/app/api/router.py
+++ b/backend/app/api/router.py
@@ -88,9 +88,10 @@ api_router.include_router(collections_crud_router)
 api_router.include_router(ogc_features_router)
 
 api_router.include_router(maps_router)
-# fix: sprites are unauthenticated reads; mounted as a sibling of maps_router
-# (not nested inside it) so they don't inherit its ERROR_RESPONSES_WRITE
-# defaults — see the comment on sprites_router in maps/router_assets.py.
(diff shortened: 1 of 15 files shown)


ishiland:
@codex review

ishiland:
@codex review

PR #1442, opened 2026-08-12
test(layering): broaden import guards to all domains

### LINKED ISSUE #836: chore(backend): maintainability batch — LOC gate globs, defaults.py split, duplicated prologues, unused dep

ishiland:
## Summary

Broadens four `backend/tests/test_layering.py` import-boundary guards past their original scope. Tests only — zero production-code changes.

**F17 — processing/ -> app.modules, all domains.** `_catalog_import_edges()` only saw `app.modules.catalog`, so a processing/ bypass into auth, audit, quota, or embed_tokens passed silently. Broadened the collector to `app.modules.*` and split the result into two shrink-only burndowns: the existing catalog one (unchanged, 13 edges / 9 files, still points fixes at ProcessingPort) and a new `_PROCESSING_OTHER_DOMAINS_IMPORT_BURNDOWN` (19 edges / 14 files). Kept separate rather than merged: ProcessingPort's surface (verified in `app/core/processing_port.py`) is dataset/record/map/grant methods only, so folding an auth/audit/quota edge into the catalog dict would point the fixer at a port that cannot serve it.

**F24 — private-name/private-module ban, past platform/.** `test_platform_does_not_import_private_module_names` only scanned `platform/` and only caught a private *name* (`from X import _y`). `processing/ai/router.py` imports two functions through `app.modules.catalog.maps._router_helpers` — a private *module* — which the old check happened to catch only because the two imported names were also underscore-prefixed; a public-looking name on that same private module would have passed. Renamed to `test_no_private_module_imports_from_app_modules`, extended to `platform/`, `processing/`, and `standards/`, and now resolves each import to its full dotted path so a private segment anywhere on it trips the guard (1 pre-existing edge, seeded).

**F7 — standards/ -> app.modules, frozen surface.** The sibling `backend/tests/test_standards_layering.py` (added by #1438) holds `standards -> app.processing` to a strict zero; nothing watched `standards -> app.modules`. STAC/OGC/DCAT construct queries directly against catalog ORM by design, so this isn't a burn-down toward zero — it's a frozen surface: today's 19 edges across 8 files are enumerated so growth requires a reviewed addition instead of happening as a side effect of a router change. Left the sibling file standalone (smallest diff) and cross-referenced it from this file's docstring.

**F6 — router-module import ban, package-wide.** `test_platform_never_imports_processing_routers` (fix #836) only caught platform/ importing `app.processing.*` routers. Added a general guard: any module-scope import of a router module (filename `router.py` / `router_*.py` / `*_router.py` — checked against every real `APIRouter(` instantiation in `backend/app/`, zero misses/false-positives) from outside its own package is flagged, anywhere in `backend/app/`. Composition root (`app/api/router.py`, `app/api/main.py`) and same-package sibling composition are allowed; function-local imports stay exempt (deferring avoids the route-registration side effect, so it isn't the bug this guard targets). The original platform-only, any-scope test is untouched and still passes — this is additive.

The F6 enumeration surfaced exactly one edge: `modules/admin/router.py` imports `app.platform.jobs.router` for `get_retry_capability`. This is **not** treated as routine reviewed debt — every other burndown entry in this file traces to a fix commit that named its tradeoff. This one predates any guard that could see it (source outside `platform/`, target outside `app.processing/` — invisible to both axes fix #836 checked) and was added in #476 ("harden lifecycle and tenant isolation"), a PR about tenant isolation with nothing suggesting the coupling was deliberate. Seeded so the guard is green today, but flagged in the code comment and here for a follow-up: `get_retry_capability` looks movable into a plain service module, the way `sweep.py` already did for this same router's other non-route helpers.

## Verification

Each guard was proven failing-first: green on the clean tree, then a synthetic violation injected (module scope, function-local where relevant, and a fabricated burndown entry for the staleness half) confirmed a precise failure message, then reverted. None of the synthetic violations are in this diff.

- `cd backend && uv run ruff check tests && uv run ruff format tests --check` — clean
- `cd backend && set -a && source ../.env.test && set +a && uv run pytest tests/test_layering.py tests/test_standards_layering.py -q` — 48 passed
- `git diff --stat daaf3561b HEAD` — touches only `backend/tests/test_layering.py`
- No `_MODULE_LOC_CAPS` entry needed: that ratchet only walks `backend/app/`, and `test_layering.py` has no self-cap (confirmed by grep — nothing in the test suite tracks its own LOC).
- No OpenAPI/SDK impact (tests only).

### DIFF
--- a/backend/tests/test_layering.py
+++ b/backend/tests/test_layering.py
@@ -4396,35 +4396,122 @@ def test_platform_does_not_import_modules() -> None:
         )
 
 
-@pytest.mark.architecture
-def test_platform_does_not_import_private_module_names() -> None:
-    """No `from app.modules.… import _private` anywhere in platform/, at any scope.
+# fix(#1438 F24): platform/ is not the only layer this rule protects — processing/
+# and standards/ reach into app.modules.* too, and a leading underscore is exactly
+# as fragile from either. `_STANDARDS_DIR` mirrors `_PLATFORM_DIR` / `_PROCESSING_DIR`
+# above; scoped here because this guard is its only reader.
+_STANDARDS_DIR = REPO_ROOT / "backend" / "app" / "standards"
+_PRIVATE_MODULE_IMPORT_ROOTS: tuple[Path, ...] = (
+    _PLATFORM_DIR,
+    _PROCESSING_DIR,
+    _STANDARDS_DIR,
+)
 
-    A leading underscore says the name can move or vanish without notice.
-    `platform/config_ops` imported the settings router's `_ENTERPRISE_ONLY_TABS`, so any
-    refactor of that router became a runtime break here.
+
+def _private_module_import_edges() -> dict[str, set[str]]:
+    """Every import reaching a `_`-prefixed segment under `app.modules.*`, at ANY
+    scope, across platform/, processing/, and standards/.
+
+    Two shapes count, because a leading underscore can sit on either side of the
+    `from X import Y` boundary. `from app.modules.X import _name` marks the NAME
+    private; `from app.modules.X._mod import name` (or `import
+    app.modules.X._mod`) marks the MODULE private, and the imported name can look
+    perfectly public while still being reached through a path that can move or
+    vanish without notice. Resolving each import to its full dotted symbol path
+    and checking every segment after `app.modules` catches both shapes with one
+    rule, rather than two independent checks that could drift apart.
     """
-    import ast
-
-    offenders: list[str] = []
-    for path in sorted(_PLATFORM_DIR.rglob("*.py")):
-        for node in ast.walk(ast.parse(path.read_text(encoding="utf-8"))):
-            if not isinstance(node, ast.ImportFrom) or not node.module:
-                continue
-            if not node.module.startswith("app.modules"):
-                continue
-            for alias in node.names:
-                if alias.name.startswith("_"):
-                    rel = path.relative_to(_PLATFORM_DIR)
-                    offenders.append(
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

Issue #1429, opened 2026-08-12
Tile caches are keyed on a reusable table name, so a deleted dataset's tiles can outlive it


ishiland:
Follow-up from the review on #1427. That PR added the missing tile-cache purge on dataset delete, which removes the durable window. Two residual windows survive it, and both come from the same root cause, so they want one fix rather than three patches.

## Root cause

Physical table names are reusable. `generate_table_name` (`backend/app/processing/ingest/service.py`) collides only against live `catalog.datasets` rows and live `pg_class` relations. A vector delete drops the table and the row, so the name is free again immediately, and the next dataset with a similar title draws it.

Three caches key on that name and none of them carry a dimension that distinguishes one occupant from the next:

| cache | key | lifetime |
|---|---|---|
| MVT tile cache (`platform/cache/tile_cache.py`) | `tile:{table}:{z}:{x}:{y}` | `tile_cache_ttl`, default 300s, tunable to 86400 |
| tile router dataset metadata (`processing/tiles/router.py`, `_dataset_cache`) | `{tenant}:{table}` | 60s, process-local |
| CDN / browser, via the `_v=` URL cache-buster | table name in the path | CDN policy |

So an entry written for occupant A can be read while serving occupant B, and B's authorization decision is made against B's visibility.

## The two windows #1427 does not close

**In-flight tile writer.** `_acquire_and_serve_tile` releases its database transaction as soon as it has `tile_data` (`processing/tiles/router.py:1886-1891`), then emits a usage event and gzips on a worker thread before calling `tile_cache.set()` (:1947-1950). The delete's `ACCESS EXCLUSIVE` lock only blocks readers up to the end of their transaction, so a reader that got past that point can write pre-delete bytes back after the purge has run.

**Per-worker in-memory caches.** `docker-compose.prod.yml` defaults to `UVICORN_WORKERS=2` with `REDIS_URL` empty, so `init_tile_cache()` gives each worker its own LRU. A DELETE is handled by one worker; the other keeps serving its own cached tiles until TTL. The same is true of `_dataset_cache`, which is process-local unconditionally and holds the deleted dataset's visibility and id for up to 60s.

## Options

[...]

PR #1444, opened 2026-08-12
fix(catalog): retire freed table names so a successor cannot inherit them

### LINKED ISSUE #1443: Stop reusing physical table names, so a successor can never inherit a predecessor's cached authorization
Follow-up from the review on #1441, which closed the tile-*bytes* half of #1429 by keying tile caches on the dataset UUID. The remaining half is the tile router's `_dataset_cache`, and it cannot be closed the same way.

## Why the cache-key trick does not work here

`_dataset_cache` maps `table_name -> _DatasetMeta`, and the vector tile route is addressed by table name (`/tiles/{table_path}/{z}/{x}/{y}.pbf`). The dataset id is the *result* of that lookup, not an input to it, so there is nothing to put in the key. The entry is also what decides authorization: `visibility`, `record_status` and `created_by` are read from the cached snapshot rather than re-queried.

The exploitable ordering, on the checked-in production default of two uvicorn workers:

1. Public dataset A occupies table `roads`. Worker 2 caches meta(A).
2. A is deleted. Worker 1 handles it and evicts its own entry; worker 2 is unreachable.
3. Private dataset B is created within 60s and draws `roads` again, because `generate_table_name` collides only against live rows and live relations.
4. An anonymous tile request lands on worker 2. It is authorized against A's `public` visibility, then queries `roads` — which now holds B's rows — and serves B's private geometry.

A second, narrower ordering exists even on one worker: a metadata lookup already in flight when the eviction runs writes its pre-delete result back afterwards. The race is one query wide, but the entry it leaves lives the full TTL.

## Why "just propagate the invalidation" is not the fix

There is no cross-process channel in the default topology:

- `REDIS_URL` is commented out in `.env.example` and defaults to empty in `docker-compose.prod.yml`, so Redis pub/sub would not cover the very configuration the scenario above describes.
- PostgreSQL `LISTEN`/`NOTIFY` needs a session-pinned connection. Transaction-mode PgBouncer is a supported topology here — the tile read path is written around it (`SET LOCAL ROLE` inside one transaction, T-1209-10) and there are routing tests for it — and it does not provide one.

Shortening or removing the TTL is the other obvious lever, and it is a deliberate, documented performance tradeoff (PERF-006 / PERF-002): the cache exists so tile serving does not hit the DB per request.

## The fix

Remove the precondition instead of chasing the propagation: **a freed table name should never be redrawn.** Then a stale entry can only ever describe the dataset it was cached for, which is the already-accepted 60s staleness window, and both orderings above become impossible — no channel, no TTL change, no per-request query.

Options, roughly in order of appeal:

1. **Tombstone the name.** Give `generate_table_name` a persistent record of retired names to collide against, alongside its existing live-row and `pg_class` probes. Needs a small table and a retention policy; the audit log is not a substitute, since its retention is not a security boundary.
2. **Suffix on reuse.** Keep names human-readable but make the collision probe consult history, so the second `roads` becomes `roads_2` exactly as a live collision already does. Same storage requirement as (1); the difference is only in what the operator sees.
3. **Defer the DROP.** Let the physical relation hold the name until a reaper removes it after a safe interval. No schema change — `generate_table_name` already probes `pg_class` — but it defers storage reclamation and needs a periodic job.

Table names are part of the public tile URL, so randomizing them is not on the table.

## Not affected

Raster and VRT: their metadata cache is keyed by `dataset_id` (`processing/tiles/router.py`), and the delete path drops no table for them, so no name is freed. The tile-bytes channel is closed by #1441 regardless of which option is chosen here.

ishiland:
## What

`generate_table_name` gains a third collision probe, against a new
`catalog.retired_table_names`. Deleting a dataset records the table name it
carried, in the same transaction, and the name generator then treats that name
exactly like a live collision — the successor of a deleted `roads` gets
`roads_2`.

## Why

GH-1429 had two halves. PR #1441 closed the tile-*bytes* half by keying tile
caches on the dataset UUID. The other half is the tile router's
`_dataset_cache`, and the same trick cannot reach it: the vector tile route is
addressed by table name (`/tiles/{table_path}/{z}/{x}/{y}.pbf`), the dataset id
is the *result* of that lookup rather than an input to it, and the cached entry
is what decides authorization — `visibility`, `record_status` and `created_by`
are read from the snapshot rather than re-queried.

On the checked-in production default of two uvicorn workers: public dataset A
occupies `roads` and worker 2 caches its metadata; A is deleted and worker 1
evicts only its own entry; a private dataset B draws `roads` within the 60s
TTL; an anonymous tile request landing on worker 2 is authorized against A's
`public` visibility and then served B's geometry.

Propagating the eviction is not available here. `REDIS_URL` is commented out in
`.env.example` and empty in `docker-compose.prod.yml`, so Redis pub/sub would
miss exactly the topology above, and PostgreSQL `LISTEN`/`NOTIFY` needs a
session-pinned connection that transaction-mode PgBouncer — a supported
topology the tile read path is already written around — does not provide.
Shortening the TTL is a documented performance tradeoff (PERF-006 / PERF-002).

So this removes the precondition instead of chasing the invalidation. Once a
freed name is never redrawn, a stale entry can only ever describe the dataset
it was cached for, which is the already-accepted 60s staleness window, and both
orderings in GH-1443 become unreachable with no channel, no TTL change, and no
per-request query. This is option 1 from the issue.

## Which sites record, and which do not

The rule: **retire a name if, at the moment it is freed, a `catalog.datasets`
row pointed at it.** That is exactly the condition under which the tile
router's metadata map could hold an entry for the name — the map is populated
from `catalog.datasets` and nothing else. A name no dataset row ever carried
was never cacheable, so retiring it buys nothing and costs the operator
readable names, because every failed ingest attempt would permanently burn a
slug. `backend/tests/test_retired_table_names_1443.py` asserts structurally
that exactly one site constructs a tombstone.

**Probes (must not hand out a retired name):**

- `generate_table_name` (`processing/ingest/service.py`) — every ingest,
  analysis output, and layer materialization.
- `register_existing_table` (same module) — the one path that takes a physical
  table name from the caller instead of generating one, so the probe is
  repeated there or the whole guarantee is bypassable: recreate a table under a
  deleted public dataset's name, register it as private, and the disclosure is
  back. It refuses rather than renaming, because registration copies no data
  and serves from the caller's own live table.
- Nothing else needs one. VRT, raster, and STAC datasets derive `table_name`
  from a fresh `record.id.hex[:16]`, which is never reissued; the vector,
  analysis, and layer paths all go through `generate_table_name`.

**Records:**

- `delete_dataset` (`catalog/datasets/domain/service_lifecycle.py`) — the only
  site where a name a live dataset row carried stops being carried by one.
  Both the single-delete route and the per-item bulk-delete route reach it, and
  both are covered end-to-end.
- Recording is not limited to the vector branch. Raster and VRT drop no
  physical table, but their dataset row goes, which frees the name in the
  catalog probe just the same — and `_resolve_dataset_meta` caches whatever row
  a `table_name` lookup finds without filtering on `record_type`, so a raster
  name is as cacheable as a vector one.

**Deliberately does not record:**

- Ingest staging tables (`ingest/tasks_vector.py`, `ingest/tasks_reupload.py`,
  `ingest/tasks_common.py`, `ingest/parquet.py`). These are attempt-scoped
  (`{base}_staging_{attempt_hex}`), no dataset row ever points at one, and they
  are dropped precisely because an attempt failed. Retiring them would burn the
  user's chosen name on every transient failure — the opposite of what GH-692's
  self-heal-to-a-suffix behaviour is for.
- The `{table_name}_old` intermediate inside the reupload swap
  (`ingest/tasks_common.py`). It exists between two statements of one
  transaction, is never any dataset's `table_name`, and the live name itself is
  *retained* by the same dataset across a reupload, so nothing is freed.
- Analysis output tables (`processing/analysis/tasks.py`). Every drop there is
  either gated on `_output_table_adopted()` returning false or runs on a
  failure path where no dataset row was ever committed, so no `catalog.datasets`
  row ever carried the name.

Recording is transactional with the operation that frees the name: the row goes
in with `session.add` inside the caller's transaction, beside the `DROP TABLE`
and the record delete. The failure direction is a delete that never happened,
never a name freed with nothing recording it.

## Schema impact

New table `catalog.retired_table_names` (migration `0044_retired_table_names`,
new head of the community chain): `id`, `table_name`, `tenant_id`, `dataset_id`,
`retired_at`, plus `ix_retired_table_names_table_name` backing the prefix probe.

- **Retention is forever.** A row is one name plus two ids, written once per
  deleted dataset, and its whole job is to outlive any cache — an expiry would
  re-open the window on precisely the names still likely to be cached. A
  deployment deleting a thousand datasets a day accumulates well under a
  megabyte a year.
- **Not unique.** `register_existing_table` adopts a physical table by its real
  name without consulting `generate_table_name`, so a retired name can
  legitimately return and be retired again; a unique constraint would turn that
  second delete into a failed delete. The probe is a set-membership test, so
  duplicates cost nothing.
- **`tenant_id` is dormant** — nullable, no FK, no RLS policy, not in migration
  0018's stamping set, the same shape `0037_dataset_refresh_runs` uses. Nothing
  reads it yet: the new probe is unscoped by tenant because the live-dataset
  probe beside it is unscoped, and the two have to agree. Over-collision costs
  a suffix; under-collision costs the guarantee. The column is there so that
  scoping both, if it is ever wanted, is a query change rather than a backfill.
- **`dataset_id` is diagnostic and deliberately not a FK** — the row it names is
  deleted in the transaction that writes the tombstone.
- **No backfill, and none is pretended.** Names freed before this deploy are
  already gone from the catalog and cannot be recovered, so they stay redrawable
  exactly once more; the caches that made that dangerous expired long ago.

`EXTENSION_API_VERSION` moves 6 -> 7. `get_retired_table_name_orm_class` is a
required `ProcessingPort` method, and the contract in
`platform/extensions/version.py` says a required Protocol addition bumps it —
otherwise an overlay declaring 6 loads cleanly and raises `AttributeError` on
the first upload, skipping the probe instead of being refused at boot. **Any
overlay replacing the `processing_port` slot must implement the accessor and
re-declare 7 before core ships.**

Both probes bind a name for **its own tenant plus the NULL scope**, matching
migration 0020's two partial unique indexes on `datasets.table_name`. A tenant
cannot inherit another's tombstone — the tile metadata cache keys on
`{tid}:{table}` and filters its query on `tenant_id`, and each tenant's tables
live in their own `data_t_{tid}` schema — so retiring globally would only cost
unrelated tenants suffixes, and with the suffix bound below it could exhaust a
shared budget and refuse a title outright. NULL binds everywhere because it is
the single-tenant namespace and where any row retired before a single-to-multi
transition sits, since nothing back-stamps this table.

## Identifier limit

Forever-retention makes long suffixes genuinely reachable, and at a 60-character
base `_100` is the first candidate past PostgreSQL's 63-byte identifier limit.
Postgres truncates it onto the same physical relation as `_10` while the catalog
keeps both untruncated strings, which puts two logical datasets on one table and
hands the disclosure back. So:

- `_with_collision_suffix` trims the base by the suffix length, keeping every
  candidate within 63.
- The collision probe's `LIKE` prefix is trimmed to match, because a candidate
  carrying a longer suffix has a *shorter* base than the one the probe used to
  key on. It over-matches for slugs longer than that prefix, costing those names
  a suffix they might not have needed — the safe direction.
- The walk refuses past `_MAX_COLLISION_SUFFIX` (9999) rather than emitting a
  name the database would silently truncate. That bound and the probe prefix are
  derived from each other and only correct as a pair.

The new test proves it against a real database, not by string assertion alone:
it creates the generated 63-character table and asserts `pg_class` holds it
under exactly that name.

Residue, named rather than papered over: a `catalog.datasets` row created before
this fix could already carry a name past 63 bytes, and its retirement records
the full stored string. Reaching that needed 99 concurrent live same-slug
datasets, so there is no realistic population, and nothing here makes it worse.

No API surface change, so `backend/openapi.json` and the SDKs are untouched.

## Failing-first evidence

A probe on the base commit, since deleted:

```
E  AssertionError: a deleted dataset's table name was handed straight back: 'roads_92f6f0'
E  assert 'roads_92f6f0' != 'roads_92f6f0'
```

## Also updated

The two residual scope notes PR #1441 shipped — the listener-seam block in
`platform/cache/provider.py` and the `_evict_dataset_meta` / `_generation_table_key`
notes in `processing/tiles/router.py` — plus the GH-1427 comment in
`delete_dataset`. All four stated "a freed name is immediately reusable" as a
live precondition; it is now removed.

`backend/tests/test_vrt_delete_guard_174.py` and
`test_dataset_delete_tile_cache_purge.py` now override `session.add` with a
plain `MagicMock`. `AsyncSession.add` is synchronous, so an un-overridden
`AsyncMock` turned the tombstone write into an un-awaited coroutine that
recorded nothing and raised a `RuntimeWarning` at collection.

## Known residue

A relation recreated **directly in the database** under a retired name, and not
registered, can still be served under the deleted dataset's cached
authorization for up to the 60s metadata TTL. That is pre-existing, unchanged by
this PR, and not API-reachable — every `CREATE TABLE` a caller can reach goes
through `generate_table_name`, and `register_existing_table` creates nothing. It
needs an actor with DDL on the `data` schema, who can already read every table
in it. Written up with both candidate remedies in GH-1451 rather than folded in
here, because each is larger than this change: verifying relation identity per
request is the round-trip `_dataset_cache` exists to avoid, and dropping the
schema-wide default `SELECT` grant changes the operator-facing runtime-role
contract for every deployment.

## Verification

```
cd backend && uv run ruff check app tests && uv run ruff format --check app tests
uv run pytest tests/test_retired_table_names_1443.py -v
uv run pytest tests/test_datasets.py tests/test_ingest_progress.py \
  tests/test_service_column_info_1359.py tests/test_dp01_ingest_write_path.py \
  tests/test_dataset_metadata_idor.py -q
uv run pytest tests/test_analysis_materialize.py tests/test_rule1_structural.py \
  tests/test_rule2_structural.py tests/test_dataset_refresh_runs.py -q
uv run pytest tests/test_processing_port.py tests/test_catalog_port.py \
  tests/test_extensions.py tests/test_extension_api_version.py \
  tests/test_vrt_delete_guard_174.py tests/test_analysis_curved_sources.py \
  tests/test_dataset_delete_tile_cache_purge.py tests/test_tile_cache_generation_key_1429.py -q
uv run pytest tests/test_tenant_cache_partitioning.py tests/test_dormant_tenancy_schema.py \
  tests/test_dp_single_tenant_byte_identical.py tests/test_embedding_tenant_isolation.py -q
# against a scratch DB at head:
uv run alembic upgrade head && uv run alembic check   # No new upgrade operations detected.
uv run alembic downgrade -1 && uv run alembic upgrade head && uv run alembic check
```

`_MODULE_LOC_CAPS` raised for `processing/ingest/service.py` (1077 -> 1199 across
four steps), `processing/tiles/router.py` (2329 -> 2341, comment only) and
`platform/extensions/defaults_processing_port.py` (494 -> 499). All exact, each
with a carve-out naming what the lines bought.

Closes GH-1443


ishiland:
@codex review

ishiland:
Round 1: both P1s fixed in 5b5a523.

**Registration bypass** — correct, and it was the only one. `register_existing_table` is the single path that takes a physical table name from the caller instead of generating it, so the probe is now repeated there and refuses with a 400 (`ValueError` reaches the single-register route as a 400 and the bulk route as a per-item error, beside the existing duplicate-registration refusal). It refuses rather than renaming: registration copies no data and serves from the caller's own live table, so renaming it would be this service reaching into storage it does not own to fix a name the caller chose. I checked every other site that writes a dataset's `table_name` — VRT, raster and STAC derive it from a fresh `record.id.hex[:16]`, which is never reissued, and the vector, analysis and layer paths all go through `generate_table_name`.

Discovery (`discover_unregistered_tables`) deliberately still lists such a table. The physical table is dropped by the delete, so it can only reappear if someone recreated it, and hiding it silently reads as a broken discovery; the refusal at register names the reason.

**Extension API version** — correct, bumped 6 -> 7 with the rationale in `version.py` and the pin in `test_extension_api_version.py` updated. Any overlay replacing the `processing_port` slot must implement the accessor and re-declare 7 before core ships; that is now called out in the PR body.

One knock-on: the migration's "not unique" rationale cited `register_existing_table` readopting a retired name, which the refusal makes unreachable. Restated to the reason that survives — nothing promises a name reaches the table only once (a future recording site, an operator insert, a restore that merges two catalogs), and uniqueness would turn any of those into a failed delete, which is the one outcome this table must never cause.

@codex review

ishiland:
Round 2: fixed in fa70b5c.

Correct, and it is my change that makes it reachable — a live-only probe needed 99 concurrent same-title datasets to get to `_100`, retirement gets there over time. The consequence is worse than truncation: `{60-char base}_100` and `{60-char base}_101` are distinct catalog strings that both address the physical relation `{base}_10`, so two logical datasets share one table and a stale metadata entry for one queries the other's rows.

`_with_collision_suffix` now trims the base by the suffix length, so every candidate is exactly within 63. Two things had to move with it:

- **The probe prefix.** A candidate carrying a longer suffix has a *shorter* base, so `LIKE '{base_slug}%'` would not match it and the walk could hand out a name already taken. The LIKE prefix is now `base_slug[:_COLLISION_PROBE_CHARS]`, which is a prefix of every candidate the walk can produce. It over-matches for slugs longer than that, costing those names a suffix they might not have needed — the safe direction.
- **A bound.** `_COLLISION_PROBE_CHARS` is only a valid prefix while the suffix tag stays short, so the walk refuses past `_MAX_COLLISION_SUFFIX` (9999) with a message rather than silently emitting a truncatable name. The two constants are derived from each other and the comment says they move together.

Tested against a real database rather than by string assertion alone: the new test creates the generated 63-character table and asserts `pg_class` holds it under exactly that name, which fails on the old code both on the length assertion and on the lookup.

Residue worth naming: a `catalog.datasets` row created before this fix could already carry a >63-byte name, and its retirement records the full stored string. Reaching that needed 99 live same-slug datasets, so there is no realistic population, and nothing here makes it worse.

@codex review

ishiland:
Round 3: fixed in 15dec56.

Accepted. I had argued for unscoped on the grounds that the live-dataset probe beside it is unscoped and the two should agree, but that reasoning does not survive the bound I added in round 2: unscoped retirement plus a finite suffix budget means one tenant's create/delete history can exhaust a title for everyone, which is a real availability regression rather than a cosmetic one. And scoping is safe — a tenant cannot inherit another's tombstone in the first place, since the tile metadata cache keys on `{tid}:{table}` and filters its query on `tenant_id`, and each tenant's tables live in their own `data_t_{tid}` schema. Migration 0020 already decided this namespace is per-tenant; unscoped retirement was quietly reversing it.

Both probe sites — `generate_table_name` and the `register_existing_table` check — now bind a name for its own tenant plus the NULL scope, matching 0020's two partial unique indexes exactly.

Two things I deliberately kept conservative:

- **NULL binds in every scope.** It is the single-tenant namespace, and it is also where a row retired before a single-to-multi transition sits, because nothing back-stamps this table (same dormant-`tenant_id` shape as `0037_dataset_refresh_runs`). Over-collision on a bounded historical set is cheap; missing those rows would reopen the window on the oldest names.
- **The string-to-UUID coercion is explicit.** `current_tenant_var` carries the id as a string and the column is `UUID(as_uuid=True)`; a comparison that silently matches nothing reads as "no name is retired" and hands one straight back, so this is not a place to rely on dialect coercion.

The live-dataset probe stays unscoped, which is pre-existing and over-restrictive rather than unsafe. Narrowing it changes name generation for existing deployments and belongs with whoever revisits 0020's probe side, not here.

@codex review

ishiland:
Rebased onto main (7a2e066, frontend-only) and pushed two follow-ups that the full backend suite surfaced after the last clean pass:

- `test_raster_ingest.py`'s `_MockDataset` had no `tenant_id`, which `delete_dataset` now reads to scope the retired-name row, and its `AsyncMock` session made the synchronous `AsyncSession.add` a coroutine that recorded nothing. Both are the same corrections already applied to the vrt and tile-cache delete tests. Focused test selection missed these; the full suite (8681 passed, 92 skipped) is green now.
- One comment rewrap in `ingest/service.py`, line-neutral so the LOC cap stays exact.

Re-requesting review because the head sha moved.

@codex review

ishiland:
Round 5: the facts check out, but I am refuting this one as out of scope rather than fixing it here. Tracked as GH-1451.

I verified both citations. `scripts/lib/configure-runtime-db-role.sh` does run `ALTER DEFAULT PRIVILEGES IN SCHEMA data GRANT SELECT ON TABLES TO geolens_reader`, and the cache-hit branch of `_resolve_dataset_meta` does return before touching the database. The sequence is real.

**It is not API-reachable.** I enumerated every `CREATE TABLE` under `backend/app/`:

- `analysis/tasks.py:1219` — `out_table` from `generate_table_name`
- `layers/service.py:104` — `generate_table_name` via the catalog port
- `datasets/domain/service_create.py:93` — same
- `ingest/parquet.py:470` — the ingest staging name, which derives from `generate_table_name`
- `ai/sql_generator.py:116` — prompt text for the NL→SQL model, never executed

All of them are now closed against retired names, and `register_existing_table` creates nothing. So reaching step 3 requires a database session with DDL on the `data` schema — an operator, not a caller.

**That actor already has the data.** Anyone who can `CREATE TABLE data.roads` can already `SELECT` every table in that schema and edit `catalog.records` directly. There is no escalation here; the only novel element is that anonymous third parties could read the recreated relation for up to 60 seconds, and only if its columns happen to match the cached `column_info`.

**It is pre-existing and unchanged by this PR.** The same sequence worked before retirement existed. This PR neither creates nor widens it — the registration refusal was never claimed to cover physical reuse, only to stop the catalog half from being the bypass.

**Both remedies you suggest are larger than the change that surfaced them.** Verifying relation identity before honouring cached authorization means a per-request `pg_class` lookup, which is precisely the round-trip `_dataset_cache` exists to avoid (PERF-006 / PERF-002) and a solution class GH-1443 ruled out by name. Dropping the schema-wide default privilege is the more appealing option and would fail the query closed, but it changes the operator-facing runtime-role contract for every deployment, and that default privilege appears to exist partly because logical restores omit ACLs — the REVOKE/GRANT repair block immediately above it in the same script says so. That deserves its own review and an upgrade note, not a fold-in here.

Written up with the full reasoning and a suggested shape in GH-1451.

PR #1445, opened 2026-08-12
chore(frontend): migrate AttributeTable to @tanstack/react-table v9

### LINKED ISSUE #1407: chore(frontend): migrate AttributeTable to @tanstack/react-table v9

ishiland:
## What

Migrates `AttributeTable.tsx` (the only frontend consumer of `@tanstack/react-table`) from v8.21.3 to v9.1.2, now that the major has settled per the issue's "When" criteria. Also removes the `@tanstack/react-table` dependabot ignore block, since the hand migration is done.

## Why

Dependabot proposed this major in #1338 (closed by hand, ignored via #1408) because v9 hadn't settled: docs were still on a `beta` branch and 9.1.0 had only just reached `latest`. It's since shipped several patches and the docs guide referenced below is current.

## v8 → v9 API changes absorbed

- `useReactTable(options)` → `useTable(options, selector)`. Passes `(state) => state` as the selector to reproduce v8's full-state re-render behavior.
- `getCoreRowModel()` / `getSortedRowModel()` table options are gone. Replaced with an explicit `features` object built via `tableFeatures({ rowSortingFeature, columnVisibilityFeature, sortedRowModel: createSortedRowModel(), sortFns: {...} })`, exported from the module so the new test below can reuse the real config instead of a re-implementation that could drift.
- `ColumnDef<TData>` → `ColumnDef<typeof features, TData>` (leading `TFeatures` generic, confirmed against the installed package's `.d.ts` — same shape applies to `Table`/`Row`/`Cell` but this component doesn't type those directly).
- `VisibilityState` → `ColumnVisibilityState`. The old name doesn't exist in v9's export list at all; confirmed by grepping the shipped `table-core` types rather than guessing.
- `flexRender(def, ctx)` → `<table.FlexRender header={header} />` / `<table.FlexRender cell={cell} />`. (The old `flexRender` function still technically exists in v9, but the component now uses the table-bound component form per the issue.)
- Sorting — the risky part per the issue, since columns here are generated dynamically from the dataset's Postgres column list with no per-column `sortFn`, relying on auto-detection:
  - Confirmed by reading the actual v9 source (`rowSortingFeature.js`, `rowSortingFeature.utils.js`) that `rowSortingFeature` still defaults every column's `sortFn` to `'auto'`, and `column_getAutoSortFn` uses the same "sample first 10 rows, check `[object Date]`, else split on digit runs" heuristic v8 used — unchanged, not a behavior regression.
  - v9 just requires the resolved name to be pre-registered, or it silently falls back to `basic`. Registered all six built-ins (`alphanumeric`, `alphanumericCaseSensitive`, `basic`, `datetime`, `text`, `textCaseSensitive`) by importing the individual `sortFn_*` functions — not the bundled `sortFns` export, which is marked deprecated upstream in favor of importing only what you use (tree-shaking doesn't help here since any of the six could be auto-selected for an arbitrary Postgres column).
  - Added a runtime test (not just a type check) exercising the real `useTable()`/`features` config against synthetic numeric, text, and ISO-timestamp-string rows, asserting sorted order. Worth calling out: timestamp columns arrive from the JSON API as ISO strings, never real `Date` instances, so they auto-detect as `alphanumeric`, not `datetime` — same as v8, and still chronologically correct for same-format zero-padded ISO strings, but worth having pinned down rather than assumed.
- `react-hooks/incompatible-library`'s suppression on the old `useReactTable` call became dead: I read the `eslint-plugin-react-hooks` source directly and its module-type provider hardcodes the property name `useReactTable` for `@tanstack/react-table` — it has no entry for `useTable`, so the rule no longer fires there. Removed the suppression.
  - Side effect: removing it unmasked the *same* warning on the unrelated `useVirtualizer()` call a few lines down — the rule only reports the first flagged hook per component, and `useReactTable` was previously occupying that slot. Verified this by running lint against the stashed original file (clean, 0 warnings) vs. this branch pre-fix (1 warning) to confirm it wasn't pre-existing. `@tanstack/react-virtual` itself is untouched and out of scope per the issue; added an equivalent suppression comment there so `npm run lint` output matches the pre-migration baseline (0 warnings).

## Incidental cleanup

Dropped `enableColumnFilter` from the column defs. It was dead even in v8 — never read via `column.getCanFilter()` or a filtered row model; the filter row is driven directly off the `NON_FILTERABLE_COLUMNS` set. v9's `ColumnDef` only exposes properties for features actually registered in `tableFeatures()`, and this component doesn't register `columnFilteringFeature` (filtering is local state + API calls, not TanStack's), so keeping the field would mean registering an unused feature just to keep a no-op property.

## Verification

All from `frontend/`:
- `npm run typecheck` — clean
- `npm run build` — clean
- `npm run lint` — clean, 0 warnings (matches pre-migration baseline; see the incompatible-library note above)
- `npm run test:coverage` — 388 files / 4838 tests passed (13 pre-existing AttributeTable tests + 3 new sort-auto-detection tests)
- No locale files touched, so `npm run test:i18n` wasn't required

Not run from this worktree, per the dev-stack bind-mount caveat (`localhost:8080` always serves `main`, not this branch): `e2e/attribute-table-ax.spec.ts` and the issue's requested manual check of column sorting on a live dataset's text/numeric/timestamp columns. The new unit-level sort test above exercises the real `useTable`/auto-detection mechanism end-to-end and is the strongest verification available outside the live stack, but it's a synthetic-data substitute, not the live check the issue asks for — flagging that as still open for CI/manual follow-up.

Closes GH-1407

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -36,11 +36,6 @@ updates:
         patterns: ["react", "react-dom", "@types/react*"]
       tanstack:
         patterns: ["@tanstack/*"]
-    ignore:
-      # fix(#1407): react-table v9 is a hand migration (AttributeTable);
-      # the major gets pulled deliberately, not via bump PR.
-      - dependency-name: "@tanstack/react-table"
-        update-types: ["version-update:semver-major"]
 
   # --- Root E2E (Playwright) — was untracked ---
   - package-ecosystem: "npm"
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -17,7 +17,7 @@
         "@maplibre/maplibre-gl-style-spec": "^26.2.1",
         "@tailwindcss/vite": "^4.3.3",
         "@tanstack/react-query": "^5.101.4",
-        "@tanstack/react-table": "^8.21.3",
+        "@tanstack/react-table": "^9.1.2",
         "@tanstack/react-virtual": "^3.14.9",
         "@turf/area": "^7.4.0",
         "@turf/distance": "^7.4.0",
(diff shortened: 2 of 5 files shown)


ishiland:
@codex review

PR #1453, opened 2026-08-13
fix(catalog): detach on registered-dataset delete instead of dropping the operator's table

### LINKED ISSUE #1452: delete of a registered dataset drops the operator's own physical table
`delete_dataset`'s vector branch drops the physical table unconditionally. For datasets created by ingest that is correct — GeoLens owns the table. But `register_existing_table` registers a table the operator created and owns, and GeoLens copies no data at registration. Deleting such a dataset therefore destroys the operator's original table, not a GeoLens-managed copy.

Surfaced while tracing the registration path during GH-1443 (PR review of the tombstone change). Pre-existing behavior, independent of that PR.

Possible directions, unassessed: record provenance (created-by-ingest vs registered) on the dataset row and skip the DROP for registered tables; or make the DROP opt-in at delete time for registered datasets. Either needs a decision on what existing deployments expect.

ishiland:
## What changed

Deleting a dataset that was registered from an existing PostGIS table now detaches instead of dropping. Registration copies no data, so the dataset row is a reference to a relation the operator built. `delete_dataset` dropped it unconditionally, which meant removing a catalog entry destroyed the original.

`geolens_owns_table` in `backend/app/platform/dataset_origin.py` is the single place that answers whether GeoLens created a dataset's table. Every origin but `postgis` did. `postgis` is the registered-in-place origin, where it did not. Raster and VRT rows are answered by record type before the origin rule is consulted, because registration only ever creates vector datasets, so a raster whose `source_format` were ever left NULL cannot read as registered-in-place.

One caller breaks the origin rule, and it is why this is more than an origin check. The analysis materialize path CTAS's its own output table and then registers it through the same `register_existing_table` an operator uses, producing an identical postgis origin with a null `source_format`. It now passes `managed=True`, which is stamped onto `origin_ref` at registration and is the only thing separating the two. The two register endpoints keep the default, `managed=False`.

An absent `managed` key reads as not managed, so every dataset registered before this change falls on the side that preserves the operator's table. That default leaks a table for analysis outputs registered before this ships. The other default would irreversibly destroy data GeoLens does not own, and that asymmetry decided it.

### Name retirement

The GH-1443 tombstone follows the **relation**, not the dataset's origin, because a name is freed only when nothing is left holding it. `delete_dataset` probes `pg_class` inside its own transaction and retires the name on every drop, plus on the detach of a table that had already vanished. It stays silent only for a detach that leaves a relation behind, which is the case that has to stay re-registerable: `register_existing_table` refuses a retired name by design, so tombstoning a surviving detached table would leave the operator permanently unable to re-adopt it.

The flag defaults to freed, so any path that does not lower it retires the name. The two mistakes are not symmetric: a missing tombstone is the GH-1443 disclosure, an extra one costs a rename.

### Delete dialog

The copy promised that all spatial data goes. It now has three cases, matching what the backend will actually do:

- GeoLens owns the table: unchanged warning that the data is destroyed.
- Registered: says GeoLens will not drop the PostgreSQL table it was registered from, that anyone who can upload can register it again, and that removing the data means dropping the table yourself.
- Registered with `source_health: 'missing'`: says GeoLens last saw the source table as missing, so there may be nothing left to remove.

Two properties of that wording are deliberate and cost several review rounds to get right. It states what **GeoLens does**, never what the table contains: `source_health` is a past observation, so a table dropped since its last probe still reads `healthy`, and a promise about the relation still existing would be wrong exactly when the backend's live `_relation_exists` check retires the name instead. And it is a contract on all four locale strings that no gate can check, since `test:i18n` verifies that keys exist, not what they say.

`deleteDetachesTable` mirrors `geolens_owns_table`, raster-family override included and in the same order. It also requires a readable `origin_ref` before promising anything, rather than reading an absent one as "not managed" — that field is redacted for non-owners, and the dataset response is cached under `['dataset', id]` with no identity in the key, so an owner can be handed a non-owner's redacted copy. Absent evidence falls back to the copy that warns the data is destroyed.

## Behavior impact

| Dataset | Before | After |
| --- | --- | --- |
| Registered, table survives | table dropped, name retired | table and rows untouched, name not retired, re-registration works |
| Registered, table already gone | name retired | name retired (unchanged) |
| Analysis output | dropped, name retired | unchanged (`managed=True`) |
| Uploaded / service / STAC / created / raster / VRT | dropped where applicable, name retired | unchanged |

Catalog-side deletion is unchanged in every case: the record, grants, tile-cache entries, and search and embedding rows all still go. Managed storage under `originals/{id}/` and `vectors/{id}/` is still reaped on a detach, since those are GeoLens artifacts keyed by dataset id, not the operator's table.

A detach leaves the table as *registration* left it, which is not the same as untouched. `register_existing_table` may add a `geom_4326` column and its GIST index, grant SELECT to the reader role, and linearize an existing `geom_4326` in place. None of that is undone, deliberately: each undo would be this path writing to a relation GeoLens does not own, and each is worse than the state it removes. Dropping the column rewrites the operator's table under a lock to delete a harmless one, revoking the grant removes a privilege that reaches nothing once the catalog row is gone, and the linearization is not reversible at all. Re-registering the table re-applies all three idempotently.

No schema change. No migration. No API surface change, so `backend/openapi.json` and the SDKs are untouched.

## Known residual, needs a decision

Three codex findings name work that is real and out of scope here. The first two converge on one missing capability and are documented at the call site rather than closed, because closing them needs a schema change this PR was scoped to avoid.

If the operator drops the relation **after** the delete transaction reads it, the name goes free with no tombstone, and a new ingest could draw it inside the metadata cache's 60-second window. Separately, a surviving detached table is listed by `/ingest/discover/` and can be registered by any user holding the editor `upload` permission.

Both want the same thing on `catalog.retired_table_names`: the relation's identity (`pg_class.oid`) and the previous owner's id. With those, registration could accept "the table I detached, re-adopted by its owner" and refuse "a new table wearing its name", while `generate_table_name` refused both. The owner's id in particular is unrecoverable after the fact, since the record row carrying `created_by` is deleted in the same transaction.

On the second finding specifically: the discovery exposure is the ambient property of a table sitting in the shared `data` schema, not something the detach introduces. The same relation was discoverable and registerable by the same users during the whole window before its owner registered it. Detach returns the table to the state the operator themselves put it in. What was cheap and worth doing was making the consequence explicit in the dialog rather than leaving it silent, which this PR does.

The third is older than this PR and belongs to the frontend cache: every identity-dependent dataset field is cached under an identity-independent key (`['dataset', id]`, 60s staleTime), while `use-auth.ts` clears only `['auth','me']` and `['auth','permissions']` on an identity change under BUG-021. Putting identity in the dataset key, or clearing `['dataset', *]` on login and logout, is the right fix and affects every consumer of a redaction-gated field. Not smuggled into a PR about dataset deletion; the dialog defends itself locally instead.

## Verification

```
# backend
cd backend && uv run ruff check .                  # clean
cd backend && uv run ruff format --check .         # clean
pytest tests/test_registered_delete_detach_1452.py -q          # 30 passed
pytest tests/test_layering.py -q                               # 47 passed
pytest <13 affected suites>                                    # 586 passed, 4 skipped
PYTHONPATH=. python scripts/dump_openapi.py --check             # no drift

# frontend
npm run lint                                       # clean
npm run typecheck                                  # clean
npm run test:i18n                                  # locale parity holds
npx vitest run src/components/dataset              # 329 passed
```

The new backend tests were confirmed non-vacuous by reverting each guard in turn and re-running: reverting the drop guard fails 4, reverting the retirement guard fails 2.

Two module LOC caps were raised with comments saying what the lines bought (`processing/ingest/service.py` 1199 to 1211, `processing/analysis/tasks.py` 1413 to 1420), and `datasets/domain/service_lifecycle.py` gained its first explicit cap, 350 to 415.

`_MockDataset` in `tests/test_raster_ingest.py` is the only hand-rolled Dataset double passed to `delete_dataset` (the two other delete-test files use `MagicMock`, which auto-supplies attributes). It gained `source_format` and `origin_ref`, defaulting to the ingested case those tests exercise.

Closes GH-1452

https://claude.ai/code/session_014tw9Frobcy3jR1ejkfjbCy

### DIFF
--- a/backend/app/modules/catalog/datasets/domain/service_lifecycle.py
+++ b/backend/app/modules/catalog/datasets/domain/service_lifecycle.py
@@ -19,6 +19,7 @@ from app.core.db.tenant_session import current_tenant_var
 from app.core.db.tenant_schema import tenant_data_schema
 from app.core.tenancy import is_multi_tenant
 from app.core.record_types import RASTER_FAMILY_RECORD_TYPES
+from app.platform.dataset_origin import geolens_owns_table
 from app.platform.storage.titiler_url import resolve_storage_key
 
 logger = structlog.stdlib.get_logger(__name__)
@@ -40,6 +41,24 @@ class DatasetTitleMismatchError(ValueError):
     public-safe case apart from other ValueErrors (e.g. a malformed
     persisted table name) can do so by type rather than by message content.
     """
+
+
+async def _reap_managed_storage(prefixes: list[str], tenant_id: str | None) -> None:
+    """Delete every object under GeoLens-managed prefixes for one dataset.
+
+    Extracted from ``delete_dataset``'s two branches (which reaped identically
+    from different prefix lists) when fix(#1452) added a branch of its own and
+    pushed the function past ruff's complexity ceiling. The import stays
+    function-local so tests keep patching the provider attribute.
+    """
+    from app.platform.storage.provider import get_storage
+
+    storage = get_storage()
+    for prefix in prefixes:
+        physical_prefix = resolve_storage_key(prefix, tenant_id=tenant_id)
+        keys = await storage.list(physical_prefix)
+        if keys:
+            await asyncio.gather(*(storage.delete(key) for key in keys))
 
 
 class DependentVrtError(Exception):
@@ -64,6 +83,12 @@ async def delete_dataset(
     Raises ValueError if dataset not found, name mismatch, or invalid table name.
     For raster datasets, storage cleanup happens before DB deletion so that a
     storage failure prevents any DB changes (no orphaned records).
+
+    fix(#1452): a dataset registered from an existing PostGIS table is DETACHED
+    rather than dropped — the catalog row, its grants, its tiles and its search
+    and embedding rows go, and the operator's table survives with its rows,
+    exactly as registration left it. See
+    :func:`app.platform.dataset_origin.geolens_owns_table`.
     """
     # Function-local import via the service.py façade is intentional — it lets
     # tests mock `service.get_dataset` to inject fixture datasets without DB.
@@ -84,9 +109,16 @@ async def delete_dataset(
 
     record_type = dataset.record.record_type
 
-    if record_type in RASTER_FAMILY_RECORD_TYPES:
-        from app.platform.storage.provider import get_storage
+    # fix(#1452): registration copies no data — it points the catalog at a
+    # table the operator built and keeps writing to. Deleting the dataset
+    # therefore has to detach, not drop, or the delete destroys the original
+    # rather than a GeoLens-managed copy. Resolved once here because the DROP
+    # and the name retirement below are two effects of this one answer.
+    owns_table = geolens_owns_table(
+        dataset.source_format, record_type, dataset.origin_ref
+    )
 
+    if record_type in RASTER_FAMILY_RECORD_TYPES:
         if record_type == "raster_dataset":
             # Guard: prevent deletion if any VRT still references this COG.
             #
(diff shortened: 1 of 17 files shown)


ishiland:
@codex review

ishiland:
Fixed in 4819b8eb4.

You are right, and the argument I wrote was wrong in exactly the way you name: "the relation still occupies the name" only holds while the relation exists. A registered dataset whose table the operator had already dropped released its name outright, and skipping its tombstone let `generate_table_name` hand that name to a new ingest inside the 60s metadata-cache window. GH-1443 through the ingest door.

The retirement now follows the RELATION rather than the dataset's origin. `delete_dataset` probes `pg_class` in its own transaction and retires the name whenever nothing is left holding it, which covers every drop plus the detach of a table that had already vanished. It stays silent only for a detach that leaves a relation behind, which is the case that has to stay re-registerable. The flag defaults to freed, so any path that does not lower it retires the name.

Three new cases in `test_registered_delete_detach_1452.py`, two of which fail against the previous commit: the vanished table's name is retired, `generate_table_name` refuses to hand it back, and a surviving table keeps its own name off new ingests without a tombstone.

On the second half of your finding, the operator dropping or renaming the relation AFTER this transaction reads it: that residual is real and is documented at the call site rather than closed. Telling "the table I detached" from "a new table wearing its name" needs the relation's identity (`pg_class.oid`) stored on `catalog.retired_table_names`, so registration can accept the first and refuse the second while `generate_table_name` refuses both. That is a migration, and this PR was scoped to avoid one. Flagged to the maintainer as a follow-up rather than decided here.

ishiland:
@codex review

ishiland:
Both addressed in cf8244789.

**P2 (missing source table) — fixed.** You are right that `tasks_postgis_refresh` maps SQLSTATE 42P01 to `source_health: "missing"`, and that the backend then detects the absent relation and retires the name rather than preserving anything. The dialog now uses missing-source wording for that case. It says "last saw" rather than "is missing", because the stored verdict is a past probe and the operator may have recreated the table since.

**P1 (reserve for the original owner) — partly refuted, partly mitigated, and the rest escalated.**

The listing and the permission are exactly as you describe: `/ingest/discover/` returns unregistered `data`-schema tables and `/ingest/register/` needs only `upload`. What I do not think holds is that the detach creates that exposure. The same relation was discoverable and registerable by the same set of users during the entire window before its owner registered it — it is an unregistered table in the shared `data` schema in both states. Detach returns the table to the state the operator themselves put it in, which is what makes it a detach.

Reserving it properly needs two things stored on `catalog.retired_table_names`: the relation's identity, and the previous owner's id. The second is not recoverable after the fact — the record row carrying `created_by` is deleted in the same transaction. So an owner-only re-registration rule is not implementable without the same schema change your round-1 residual needs, and this PR is deliberately scoped to avoid a migration. Escalated to the maintainer as one follow-up covering both.

What was cheap and worth doing is not leaving the consequence silent, so the registered-delete copy now states it: the table can be registered again by anyone who can upload, and removing the data means dropping the table yourself. The owner can act on that; they could not act on our silence.

ishiland:
@codex review

ishiland:
Fixed in 275b2ac06. Correct, and it is the frontend sibling of a guard the backend already carried: `geolens_owns_table` answers on record type before it consults the origin, and `deleteDetachesTable` skipped that step. The check now comes first on both sides, in the same order.

I checked the rest of the mirror for the same class of drift and found none. The frontend reads the SERVED `origin`, which is `classify_origin`'s own output, so the postgis test cannot disagree with the backend's; and a non-dict `origin_ref` resolves to "not managed" on both sides, since optional chaining on a string yields undefined the way the `isinstance` guard yields False. Covered by two new parameterized cases.

@codex review

ishiland:
Pushed d230ec80f, comment only: the drop-vs-detach block still claimed its answer decided both the DROP and the name retirement, which stopped being true when the retirement moved onto `name_is_freed` in 4819b8eb4. No behavior change.

@codex review

ishiland:
CI's Backend Tests found two failures the local affected-suite selection missed: `test_tenant_attribution_materialize.py` is in the serial group and I listed it as affected without running it. Fixed in 25e044f36, both following from the analysis path now registering with `managed=True`.

`_register_without_tenant_guc` was the third `register_existing_table` test double in the suite, sibling to the two in `test_analysis_materialize.py` that were already widened, and it now forwards `**kwargs`. The origin-pointer assertion pins the whole ref dict, so it gains the `managed` key. Neither is a behavior change.

@codex review

ishiland:
Fixed in a463e9be3, and the premises check out: `queryKeys.datasets.detail` is `['dataset', id]` with no identity, `staleTime` is 60s, and `use-auth.ts` invalidates only `['auth','me']` and removes `['auth','permissions']`. So an owner really can be handed the redacted response a non-owner or anonymous reader populated moments earlier.

The dialog no longer infers anything from an absent `origin_ref`. It requires a readable object before promising survival, and falls back to the data-is-destroyed wording otherwise. That also covers the rows migration 0036's backfill could not resolve, which carry the same null for an unrelated reason. It over-warns for a registered table whose provenance we cannot read, which is the side to be wrong on: a spurious warning costs nothing that a false assurance does not cost more. Three cases cover absent, null, and non-object refs.

I did not take the other two options you offered. Putting identity in the dataset query key, or invalidating `['dataset', *]` on login and logout, is the right fix for the underlying weakness — every identity-dependent dataset field is cached under an identity-independent key, and the auth hook already does exactly this for `['auth','me']` and `['auth','permissions']` under BUG-021 — but it changes cross-cutting cache behavior for every consumer, from a PR about dataset deletion. Raised for the maintainer as its own change rather than smuggled in here.

@codex review

ishiland:
Fixed in bbcf118a5, and this one improved the shape of the whole thing. You are right that `source_health` is a past observation and that the copy was making a claim about the world on the strength of it.

The registered copy now states what GeoLens will do, which is the only part that is ours to promise: it will not drop the table, so whatever the table holds is left alone. That holds whether or not the relation is still there, so it no longer depends on a stored verdict being current, and the live `_relation_exists` probe finding nothing cannot contradict it.

The `missing` variant stays, because "we will not drop it" is a strange thing to say about a table the refresh has actually seen disappear, but it is now a courtesy rather than the thing standing between the reader and a false promise. All four locales.

@codex review

ishiland:
Fixed in f85c04132. Straightforwardly right, and it is the sibling-set mistake: I corrected the English claim and left the same claim standing in the three translations. Spanish said the contents stay intact, French and German said the same.

All three now describe GeoLens's non-action on the table ("la dejará tal como esté", "n'y touchera pas", "lässt sie unangetastet"). I swept both registered keys across all four locales rather than fixing the three you named: the `missing` variants were already hedged past-tense observations everywhere and are unchanged.

The wording is a contract on the string that no gate can check, so it is now written down beside the selector: describe what GeoLens does to the table, never what the table holds. `test:i18n` checks that keys exist, not what they say, which is exactly how one corrected string and three stale ones passed CI.

@codex review

ishiland:
Fixed in 911cb72e4. It now reads `fix(#1452 review round 6)` like every other finding reference in the branch. I swept the rest of the diff for the same pattern: nine other review-round references, all already scoped to #1452, so this was the only one. Comment only.

@codex review

PR #1454, opened 2026-08-13
sec(tiles): refuse cached authorization for a dataset row that no longer exists

### LINKED ISSUE #1451: sec(tiles): a recreated-but-unregistered data-schema relation can be served under a deleted dataset's cached authorization
Raised during review of PR #1444 (which closes GH-1443). Recording it rather than folding it into that PR, because it is a different threat model with a different actor, and both obvious remedies are larger than the change that surfaced it.

## The residue

GH-1443 removed the product-reachable half: a freed physical table name is retired and never redrawn, so no dataset a user can create through the API — ingest, analysis output, layer materialization, or `register_existing_table` — can land on a name whose predecessor might still be in a tile worker's metadata cache. Every `CREATE TABLE` under `backend/app/` that a caller can reach goes through `generate_table_name`, and the registration path repeats the probe.

What remains needs **direct DDL on the `data` schema**:

1. Public dataset A occupies `roads`. A tile worker caches meta(A).
2. A is deleted. The physical table is dropped and `roads` is retired.
3. Someone with a database session recreates a physical `roads` holding private rows, and does **not** register it.
4. Within the 60s `_DATASET_CACHE_TTL`, an anonymous tile request for `/tiles/roads/...` hits the cached meta(A), is authorized against A's `public` visibility, and queries the recreated relation.

The cache-hit branch (`_resolve_dataset_meta`, `processing/tiles/router.py`) returns before touching the database, so registration having been refused is not consulted. `scripts/lib/configure-runtime-db-role.sh` runs `ALTER DEFAULT PRIVILEGES IN SCHEMA data GRANT SELECT ON TABLES TO geolens_reader`, so the recreated relation is readable by the role the tile path binds with `SET LOCAL ROLE`, without `grant_reader_access` ever running. The query only returns rows if the recreated table's columns are compatible with the cached `column_info`, which a restore-from-backup would satisfy.

## Why it was not fixed in #1444

- **Pre-existing and unchanged by that PR.** The same sequence worked before retirement existed. #1444 neither creates nor widens it.
- **The actor already has the data.** Anyone who can `CREATE TABLE data.roads` can already `SELECT` every table in that schema and edit `catalog.records` directly. There is no privilege escalation; the novel part is only that anonymous third parties could read the recreated table for up to 60s.
- **Both remedies are bigger than the change that found it.**
  - Verifying the relation identity (OID/generation) before honouring cached authorization means a per-request `pg_class` lookup — exactly the round-trip `_dataset_cache` exists to avoid (PERF-006 / PERF-002), and a solution class GH-1443 explicitly ruled out.
  - Dropping the schema-wide default privilege so an unregistered relation is simply unreadable by `geolens_reader` is the more appealing option and would fail the query closed. But it changes the operator-facing runtime-role contract for every deployment, and the default privilege appears to exist partly because logical restores omit ACLs (see the REVOKE/GRANT repair block directly above it in the same script). That needs its own review and an upgrade note.

## Suggested shape

Option 2 above, plus a check that every read path reaching a `data`-schema relation has an explicit grant behind it, and a reconciliation step for tables whose ACLs a restore dropped. Worth pairing with a decision on whether unregistered relations in `data` should be readable by the runtime role at all.

ishiland:
## What changed

`_resolve_dataset_meta` answers a cache hit without touching the database, so for up to 60 seconds (`_DATASET_CACHE_TTL`) a tile worker keeps authorizing against a dataset row that may already be deleted. GH-1443 retired freed table names, which closed the half a caller could reach through the API. Direct DDL on the `data` schema is not that half: `CREATE TABLE data.roads` needs no registration, and the schema-wide `ALTER DEFAULT PRIVILEGES ... GRANT SELECT ON TABLES TO geolens_reader` in `scripts/lib/configure-runtime-db-role.sh` hands the tile role read access to it without `grant_reader_access` ever running. Inside the TTL an anonymous request for `/tiles/data.roads/...` is authorized against the deleted dataset's `public` visibility and answered from a relation nobody registered.

New `_assert_dataset_still_registered()` in `backend/app/processing/tiles/router.py` asks the catalog whether the cached authorization still has a row behind it. Both tile endpoints call it on the first line past their byte-cache short-circuit.

## Which option this is

GH-1451 costed two remedies. This is neither verbatim.

**Option 1** (verify relation identity with a per-request `pg_class` lookup) was ruled out because it costs the round-trip `_dataset_cache` exists to avoid. That objection is about the hot path, not about consulting the database at all — a tile served from the byte cache returns above this check and still costs zero round-trips. Asking the catalog is also stronger than checking relation identity: it validates the authorization premise rather than a proxy for it, and it does not break after a `pg_dump`/restore the way a cached OID would.

**Option 2** (drop the schema-wide default privilege) changes the operator-facing runtime-role contract for every deployment and needs its own review, an upgrade note, and ACL reconciliation for restores. It remains worth doing on its own terms and nothing here blocks it.

Propagating the eviction was not available: the GH-1429 listener is process-local, `REDIS_URL` is unset in the checked-in default, and a delete therefore only evicts in the worker that served it.

## Where the check sits, and why it took four review rounds

Placement turned out to be most of this change, so it is written down at the call site rather than left to be re-derived.

The position is bounded on both sides. **No earlier**, or a byte-cache hit pays for a check that only matters when the relation is about to be read. **No later**, for two independent reasons:

- Everything past that line acts on the cached authorization, starting with the COLD-02 seam, which would enqueue a restore and return a 202 for a dataset the catalog no longer has.
- A tile request then takes three bounded resources in sequence: an API-pool connection, a FAIR-01 permit, a tile-pool connection. `get_db` holds the first until the response is written, so a metadata-cache miss carries `_resolve_dataset_meta`'s connection into both later waits. Any position that takes a permit or a tile connection *before* asking the API pool inverts a pair against that path, and two bounded resources acquired in opposite orders stall under ordinary mixed load with no attacker involved.

The check therefore runs before all of it and rolls back before the first acquisition. That release also retires the pre-existing half of the same problem, where a metadata-cache miss and an authenticated user lookup both carried their connection into the permit wait.

Asking on the tile connection instead would need neither pool twice, but it would need `geolens_reader` to hold SELECT on `catalog.datasets` — a runtime-role contract change for every deployment, and a widening of the one role whose narrowness is why this path binds it.

One finding was refuted rather than fixed: holding a row lock (`FOR KEY SHARE`) through the relation read. Reasoning is in a PR comment. Short version: it needs the same catalog grant, it makes ordinary tile traffic block dataset deletes, and the race it closes is microseconds wide against an actor who by GH-1451's own analysis already has full read access to the schema.

## Behavior impact

- A tile request whose cached authorization no longer has a catalog row behind it returns 404 instead of serving, and no longer wakes cold storage. The refusal evicts the entry that produced it, so the next request 404s in the resolver.
- No change for any registered dataset, and no change to byte-cache-hit latency.
- A request the FAIR-01 limiter is about to reject now pays one indexed primary-key lookup. That is the cost of a single consistent acquisition order; the connection is back in the pool before any wait begins.
- The probe pins id and table_name together, so a surviving row repointed at a different relation cannot authorize a read of the old name, and it mirrors the resolver's tenant filter.
- No migration, no config change, no API surface change.
- The raster proxy caches authorization the same way and is deliberately not covered: it is addressed by dataset id and resolves to an object-storage asset, so there is no relation an out-of-band `CREATE TABLE` could substitute. Its bounded staleness is the tradeoff already documented at `_RASTER_META_CACHE_TTL`.

## Tests

`backend/tests/test_tile_cached_authz_liveness_1451.py` (11 tests).

The scenario, on both endpoints: seed a public dataset over a real physical table, warm the metadata cache, delete through the real `delete_dataset` service without evicting, recreate the relation out of band, then issue an anonymous tile request. Driving the service rather than the endpoint is deliberate — GH-1429 moved the eviction to the delete endpoints after their commit, so this is the state every worker that did not serve the delete actually holds.

**I confirmed the tests fail without the fix.** With the check removed, both scenario cases return **200** and serve the unregistered relation.

The rest pin what a 404 assertion cannot see: a live dataset still serves its own tiles; a byte-cache hit reaches the database zero times; the recorded call order puts the catalog read and its rollback ahead of the cold seam and both pool acquisitions; a refusal takes no pool slot and wakes no storage; and an AST walk of both endpoints fails if the check is missing or falls after the cold seam or the serve helper. That last one replaces the structural funnel the check gave up by moving out of `_acquire_and_serve_tile`.

`test_data_serving_extension.py` needed its `get_db` override changed from `None` to a session double, since the endpoints now use the session they previously never touched. The double answers the probe with a row rather than patching the check out.

`tiles/router.py` grew 127 lines, so its `_MODULE_LOC_CAPS` entry moves 2341 to 2468 with the carve-out written beside it.

## Verification

```
uv run pytest tests/test_tile_cached_authz_liveness_1451.py -q            # 11 passed
              tests/test_rule2_structural.py -q
```

Plus every suite in the tree that touches a tile endpoint, found by grepping the
test tree for `/tiles/`, `_acquire_and_serve_tile` and the two endpoint names —
33 files including `test_tiles`, `test_vector_tile_auth`, `test_tile_signing`,
`test_raster_tiles`, `test_embed_tokens`, `test_dp02_read_path_binding`,
`test_dp03_cross_tenant_privilege_gate`, `test_tenant_session_guc` and
`test_retired_table_names_1443`: **885 passed, 2 skipped**.

No e2e or alembic-check: this touches no schema and no frontend.

Closes GH-1451

https://claude.ai/code/session_014tw9Frobcy3jR1ejkfjbCy

### DIFF
--- a/backend/app/processing/tiles/router.py
+++ b/backend/app/processing/tiles/router.py
@@ -1781,13 +1781,34 @@ async def _assert_dataset_still_registered(
     LRU, so a delete only evicts in the worker that served it. No process-local
     cache can answer for the others, so the check has to reach the database.
 
-    The cost is bounded by WHERE this runs. The caller is ``_acquire_and_serve_tile``,
-    which reaches this only after both tile caches have missed and before it takes
-    any capacity. A served-from-cache tile still costs zero round-trips, which is
-    the whole point of ``_dataset_cache`` (PERF-006 / PERF-002); a tile the pool
-    has to build pays one indexed primary-key lookup that is over, connection and
-    all, before the FAIR-01 permit is requested. The call site records why every
-    later position inverts an acquisition order against some other path.
+    WHERE this runs is the rest of the design, and four review rounds narrowed it
+    to one line in each endpoint: the first statement past the tile-byte-cache
+    short-circuit. Both bounds are tight.
+
+    No earlier, or the hot path pays for it. A tile answered from the byte cache
+    returns above this and still costs zero round-trips, which is the whole point
+    of ``_dataset_cache`` (PERF-006 / PERF-002).
+
+    No later, for two independent reasons. Everything below acts on the cached
+    authorization, starting with the COLD-02 seam, which would enqueue a restore
+    and hand back a 202 for a dataset the catalog no longer has. And a tile
+    request takes three bounded resources in sequence — this API-pool connection,
+    the FAIR-01 permit, then the tile-pool connection — so the check has to
+    complete and roll back before the first of them is requested. Every later
+    position inverts a pair against a metadata-cache MISS, which carries
+    ``_resolve_dataset_meta``'s connection into both waits: inside the tile
+    transaction it asks the API pool while holding a tile connection; after the
+    permit it holds a permit while asking the API pool. Both stall under ordinary
+    mixed load with no attacker involved.
+
+    Asking on the tile connection instead would need neither pool twice, but it
+    would need ``geolens_reader`` to hold SELECT on ``catalog.datasets``: a
+    runtime-role contract change for every deployment, and a widening of the one
+    role whose narrowness is why this path binds it at all.
+
+    ``test_both_tile_endpoints_ask_before_they_act`` pins the position, because a
+    third endpoint that resolved cached metadata and forgot this call would be a
+    silent regression.
 
     Pinning id AND table_name together is what makes it a liveness check rather
     than an existence check: a surviving row that has since been repointed at a
@@ -1976,8 +1997,6 @@ def _cluster_cache_table_key(
 async def _acquire_and_serve_tile(
     *,
     request: Request,
-    db: AsyncSession,
-    dataset_id: uuid.UUID,
     table_name: str,
     z: int,
     x: int,
@@ -2010,35 +2029,6 @@ async def _acquire_and_serve_tile(
     Callers keep their own cache-hit short-circuit and COLD-02 cold-rehydrate seam,
     which differ between the two paths.
     """
-    # fix(#1451): the catalog check, and the release that comes with it, run here
-    # — before any capacity is taken. Three resources are in play on this path and
-    # they are now acquired in one order by every request: the API-pool connection
-    # first, then the FAIR-01 permit, then the tile connection. Each is released
-    # before the next is requested, so no two paths can hold them in opposite
-    # orders (codex P1, rounds 1-3, which found one edge of that graph per round).
-    #
-    # `_assert_dataset_still_registered` ends by rolling back, which is what makes
-    # this position safe rather than merely early. The round-1 objection was
-    # retention, not placement: `db.execute` opens a transaction `get_db` would
-    # otherwise hold until the response is written, pinning an API connection
-    # through the 10s permit wait, the tile-pool wait, the PostGIS query and the
-    # gzip. Handing it back here also retires the PRE-EXISTING half of the same
-    # problem, since a metadata-cache miss already carried `_resolve_dataset_meta`'s
-    # connection into the permit wait, and an authenticated request carried the
-    # user lookup's.
-    #
-    # The two later positions both invert something. Inside the tile transaction
-    # it asks the API pool while holding a tile connection; after the permit it
-    # holds a permit while asking the API pool. Either one stalls against a
-    # metadata-cache miss taking the same pair the other way, under ordinary mixed
-    # load and with no attacker involved. Asking on the tile connection instead
-    # would need `geolens_reader` to hold SELECT on catalog.datasets: a
-    # runtime-role contract change for every deployment, and a widening of the one
-    # role whose narrowness is why this path binds it at all.
-    await _assert_dataset_still_registered(
-        db, dataset_id=dataset_id, table_name=table_name, tid=tid
-    )
-
     try:
         pool = get_tile_pool()
     except RuntimeError as exc:
@@ -2231,6 +2221,14 @@ async def cluster_tile_endpoint(
                 _serving_tile_headers(cache_scope, cache_ttl, _cluster_cache_control),
             )
 
+    # fix(#1451): first thing past the byte-cache short-circuit, because
+    # everything past it acts on the cached authorization — the cold seam below
+    # would enqueue a restore for a dataset the catalog no longer has. See the
+    # helper for why it cannot sit any later than this.
+    await _assert_dataset_still_registered(
+        db, dataset_id=meta.dataset_id, table_name=table_name, tid=_cluster_tid
+    )
+
     # COLD-02 (Phase 1214-04): cold-rehydrate seam — BEFORE cluster tile query.
     # Mirrors the tile_endpoint seam: uses cached meta.record_status (T-1214-18);
     # failure is broad-except-swallowed (T-1214-17).
@@ -2266,8 +2264,6 @@ async def cluster_tile_endpoint(
     # The same tenant concurrency budget governs vector and cluster DB reads.
     return await _acquire_and_serve_tile(
         request=request,
-        db=db,
-        dataset_id=meta.dataset_id,
         table_name=table_name,
         z=z,
         x=x,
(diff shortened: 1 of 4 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
Round 2 P1 (nested pool acquisition) is fixed in b5b3d3b — the check now runs
between the FAIR-01 permit and the tile connection and releases its API
connection before the acquire, so no path holds either pool while waiting on the
other. It also retires the pre-existing half of that inversion, where a
metadata-cache miss carried an API connection into the tile-pool wait.

Round 2 P2 (hold a lock through the relation read) I am not implementing, and I
want the reasoning on the record rather than silently dropped.

The finding is technically correct. After the probe commits its read and rolls
back, a delete can commit and an out-of-band session can recreate the relation
before the tile query resolves the name. Closing that needs the catalog row
locked (`FOR KEY SHARE`) in the same transaction as the relation read, held
until the read takes its own lock.

Three reasons that trade is worse than the window it closes.

**The actor already has the data.** GH-1451 says so in its own refutation of
GH-1444: recreating `data.roads` requires direct DDL on the data schema, and
anyone holding that can already `SELECT` every table in it and edit
`catalog.records` directly. Nobody with those rights needs to win a microsecond
race against a probe to read rows they can already read, or to flip a record to
public and read them at leisure. The disclosure being defended is a third party
reading the recreated table, and that requires the attacker to land a DROP plus
CREATE inside the gap between two statements on an already-acquired connection.

**The lock is only available at a price the issue already priced.** Same
transaction means the tile connection, and `geolens_reader` has no grant on
`catalog` — it holds USAGE on `data` and SELECT on tables in `data`, nothing
else. Granting it catalog access is a runtime-role contract change for every
deployment, needing the configure script, an upgrade note and reconciliation for
restores that dropped ACLs. GH-1451 costed that as its Option 2 and called it
"needs its own review". It also widens the one role whose narrowness is why the
tile path binds it, which trades a defence for a defence.

**A row lock on the tile read path makes tile traffic block deletes.** Every
uncached tile would hold `FOR KEY SHARE` on its dataset row for the duration of
a PostGIS query, and a dataset delete would wait behind whatever tiles are in
flight. That is a live availability cost on an ordinary operation, paid on every
request, to close a race that needs an attacker who does not need it.

What the change does close is the 60-second window in which a stale cached
authorization serves an unregistered relation to anonymous callers, which is the
disclosure GH-1451 describes. The residual is two adjacent statements wide.

Happy to be overruled if you would rather take Option 2 wholesale — the grant,
the upgrade note and the ACL reconciliation together — but that is a different
change from this one, and GH-1451 says as much.

ishiland:
@codex review

ishiland:
Round 3 P1 is fixed in 87a0e83, and this time as the whole class rather than the edge that was reported.

Three resources are in play on a tile request: an API-pool connection, a FAIR-01 permit, and a tile-pool connection. `get_db` holds the first until the response is written, so a metadata-cache miss carries it through both later waits. Every position the check has occupied inverted some pair against that: ahead of the permit it held an API connection through the 10s wait (round 1); inside the tile transaction it asked the API pool while holding a tile connection (round 2); after the permit it held a permit while asking the API pool (round 3).

The check now runs before any capacity is taken and rolls back before the permit is requested, so every request uses one order — API connection released, then permit, then tile connection. Releasing there also retires the pre-existing half, where a metadata-cache miss and an authenticated user lookup both carried their connection into the permit wait.

The consequence worth naming: a request the limiter is about to reject now pays one indexed primary-key lookup. That is the cost of a consistent order, and the connection is back in the pool before the wait begins, so it starves nothing. The limiter-rejection test asserts the rollback rather than the absence of the query.

Round 2's P2 remains refuted in the comment above.

ishiland:
@codex review

ishiland:
Round 4 P2 is fixed in f465d30. The check moved out of `_acquire_and_serve_tile` and into both endpoints, on the first line past the byte-cache short-circuit, which puts it above the COLD-02 seam. You were right that the seam acts on the same cached `record_status`: a deleted dataset left cold would have bought a restore job and a 202 with the catalog never consulted.

That position is now bounded on both sides. Earlier and a byte-cache hit pays for it, which is what `_dataset_cache` exists to avoid. Later and something has already acted on metadata the catalog may not back — the seam first, then the three resource acquisitions from the previous round, which all sit below it and keep the order they were given.

Moving up cost the structural funnel, so the convention is enforced instead of documented: `test_both_tile_endpoints_ask_before_they_act` walks each endpoint's AST and fails if the call is missing or falls after `_check_cold_rehydrate` or `_acquire_and_serve_tile`. `_acquire_and_serve_tile` is back to the signature it had before this branch.

Verification on this head: ruff check and format across backend, `test_tile_cached_authz_liveness_1451` (11), `test_layering`, `test_rule1_structural`, `test_rule2_structural`, and every suite in the tree that touches a tile endpoint — 885 passed, 2 skipped.

ishiland:
@codex review

PR #1457, opened 2026-08-13
fix(catalog): record freed relation oid and prior owner on tombstones

### LINKED ISSUE #1456: catalog: tombstones should record the freed relation identity and prior owner
Follow-up to GH-1444 (tombstones) and GH-1452 (detach on registered delete, PR #1453). Two review findings on #1453 converge on the same schema gap: `catalog.retired_table_names` records only the freed NAME, not the freed relation's identity or its prior owner.

## The two windows

1. **Post-transaction drop goes untombstoned.** Detach decides tombstone-vs-not by probing pg_class inside the delete transaction. If the operator drops the relation AFTER that read, the name is freed with no tombstone row, reopening the GH-1443 ingest window for that name.
2. **A surviving detached table is discoverable and re-registerable by any editor with `upload`** via `/ingest/discover/`. Partially refuted on the PR: this is the ambient property of any table in the shared `data` schema, identical to the window before its owner first registered it — the detach restores the pre-registration status quo rather than creating a new exposure class. The delete dialog now says so instead of being silent. Recording the prior owner would still allow a future ownership-aware re-registration flow.

## Why it needs a migration

Both windows want the tombstone (or a sibling record) to carry the relation's `pg_class.oid` and the previous owner's user id. The owner id is unrecoverable after the fact: the record row carrying `created_by` dies in the same transaction that frees the name. That is a schema change to a retain-forever table, so it was deliberately not folded into #1453; the call site documents the gap.

## Cost of doing nothing

Window 1 is narrow (operator drops the table in the seconds between the delete's probe and its commit) and its damage is the already-documented GH-1443 residual, bounded by #1441's cache keying. Window 2 is status-quo exposure. Neither is urgent; both are cheap to close properly once a migration is on the table.

ishiland:
## What changed

Every dataset delete now records the identity of the relation it released, and who owned the dataset that held it.

`catalog.retired_table_names` gains two nullable columns, `relation_oid` (the pg_class oid of the relation whose name the delete freed) and `previous_owner_id` (the `created_by` of the dataset that held it), for the deletes that free a name. A sibling table `catalog.detached_relations` takes the same two values plus the name, tenant and dataset ids for the delete that frees nothing: a detach that leaves the operator's table standing.

Nothing reads any of it. The tombstone-vs-not decision is exactly the one #1453 shipped, and the retirement set contains exactly the rows it would have contained before.

It is a PR of its own because none of these values can be recovered afterwards. The `catalog.records` row carrying `created_by` is deleted inside the delete transaction. The oid dies with its relation, either in the same transaction (a drop) or whenever the operator gets round to it (a detach). A closure written later would find both sources gone and would have to start collecting from its own deploy date.

## The two windows from #1456

Window 1 is the post-transaction drop. Detach decides tombstone-vs-not by probing pg_class inside the delete transaction, so an operator who drops the relation *after* that read frees the name with no tombstone, reopening GH-1443's ingest window for it. Closing that needs registration to accept "the table I detached" and refuse "a new table wearing its name" while `generate_table_name` refuses both, and telling those two apart is what the recorded identity is for. This is the window the data enables closing later. The closure itself is not here.

Window 1 lives entirely on the path that writes no tombstone, which is why `detached_relations` exists. The first version of this PR recorded identity alongside retirements only, which collected it for every case except the one it was collected for. Codex caught that; the `elif` branch in `delete_dataset` is the answer.

Window 2 is the surviving detached table, discoverable and re-registerable by any editor holding `upload`. #1453 partially refuted it: that is the ambient property of any table sitting in the shared `data` schema, identical to the window before its owner first registered it. `previous_owner_id` is what a later ownership-aware re-registration would key on, but no such flow exists and this PR does not propose one.

Neither window is closed here. Both stay as wide as they were.

## Why a sibling table and not a flag on the tombstone

`retired_table_names` has one API, and it is set membership. `generate_table_name` asks "is this name retired?" and `register_existing_table` refuses any name that answers yes. A row describing a relation that still holds its name is not a prohibition. Putting one in that set either burns the operator's own table name, if the readers ignore the discriminator, or makes every future reader of a security-load-bearing probe responsible for remembering a predicate whose failure direction is silent.

GH-1456 sanctioned either shape ("the tombstone or a sibling record"). This is the one that cannot be read wrong, and it needs no change to either consumer. `test_the_record_does_not_retire_the_name` asks both of them directly: after a detach, registration still accepts the operator's surviving table, and the collision walk still refuses the name for the reason it always did, because a live relation occupies it.

`detached_relations` mirrors 0044's shape and its reasoning: retain-forever, no unique constraint, dormant nullable `tenant_id`, no foreign keys, and an index on `table_name` because every plausible consumer arrives by name.

## What deliberately does not change

`name_is_freed` is computed from the same probe, in the same transaction, with the same fail-safe default of retiring. Nothing was added to or removed from the retirement set.

The DROP path does not consult the probe. A missing relation there means `DROP TABLE IF EXISTS` is a no-op, not that the name stays taken: the catalog row going is what frees it. Reading the oid on that path only records what was dropped.

A detach that leaves the relation standing still writes no tombstone. Retiring it would leave the operator holding a table their own catalog can never accept again, which is why #1453 keys on the relation rather than the origin.

No API surface changes. `backend/openapi.json` is untouched and `scripts/dump_openapi.py --check` agrees.

The one cost is a second pg_catalog lookup on the ingested-vector delete, which previously probed nothing. It is an indexed single-row read inside a transaction that is already dropping a table.

## Why neither owner column carries a foreign key

Both tables are retain-forever, and an FK to `catalog.users` gives only bad options. CASCADE erases the record when a user is deleted, silently re-arming GH-1443 for every name that user's datasets freed. RESTRICT lets a retain-forever row block a user deletion. SET NULL discards the durable half of the identity to preserve referential tidiness nothing reads.

The same reasoning already governs the `dataset_id` column beside it, whose referent dies in the transaction that writes the row. `test_deleting_the_prior_owner_leaves_the_tombstone_intact` asks the database directly: delete the owner, and the tombstone is still there with the uuid intact.

## The oid does not survive a restore

`pg_dump`/`pg_restore` does not preserve oids. Every row written before a RUNBOOK restore carries an oid that matches nothing on the far side, so a consumer has to read that as unknown rather than as a mismatch. Refusing a re-registration because a restored oid disagrees would be exactly backwards. `previous_owner_id` is the durable half and survives the restore intact. The caveat is repeated in both migration docstrings, on both models, and at the capture site, because each is a place where someone will decide what a stored oid means.

`BIGINT` rather than `INTEGER`: a pg_class oid is unsigned 32-bit, so int4 overflows past 2^31, and a cluster that has churned enough objects hands those out.

## Sibling findings preempted

Nullability of every new column. NOT NULL would make pre-#1456 tombstones unrepresentable and could fail a delete, which is the one outcome these tables must never cause. On the tombstone, `relation_oid` is legitimately NULL wherever no relation held the name: the raster/VRT branch, whose `table_name` is the synthetic `raster_<hex>` and names nothing in the data schema, and a detach whose table the operator had already dropped. NULL means "not recorded", never "no owner" and never "no relation". On `detached_relations`, `relation_oid` is nullable even though the only write site reaches it with a non-null oid, guarded by an explicit `is not None` so a row can never assert an identity nothing verified.

FK absence on both owner columns, covered above. `test_no_foreign_key_reaches_out_of_this_table` pins it against `pg_constraint`, so re-adding one fails a test rather than passing review.

Backfill impossibility. Every tombstone written before 0045 was written after its relation was dropped and its record row deleted, so neither value survives anywhere. Every relation detached before 0046 recorded nothing and its owner is already gone with the record row. Both migrations say so instead of pretending, the way 0044 handled names freed before it existed.

oid transience across dump and restore, covered above.

Write-site count. `test_only_delete_dataset_constructs_a_detach_record` walks the `backend/app/` ASTs and requires exactly one `DetachedRelation` construction, mirroring the enumeration #1444 already applies to `RetiredTableName`.

## Verification

```
cd backend && uv run ruff check .                       # clean
cd backend && uv run ruff format --check .              # 991 files already formatted
PYTHONPATH=. python scripts/dump_openapi.py --check     # exit 0, no drift

alembic upgrade head                                    # 0044 -> 0045 -> 0046, test DB
alembic downgrade -1 && alembic upgrade head            # both round-trip clean

pytest tests/test_tombstone_identity_1456.py -q         # 15 passed (new)
pytest tests/test_retired_table_names_1443.py \
       tests/test_registered_delete_detach_1452.py \
       tests/test_raster_ingest.py \
       tests/test_vrt_delete_guard_174.py \
       tests/test_dataset_delete_tile_cache_purge.py \
       tests/test_layering.py \
       tests/test_migration_discovery.py \
       tests/test_dormant_tenancy_schema.py \
       tests/test_alembic_helpers.py -q                 # 173 passed
pytest tests/test_datasets.py tests/test_raster_replace_1221.py \
       tests/test_dataset_metadata_idor.py tests/test_attribute_metadata.py \
       tests/test_tile_cache_generation_key_1429.py \
       tests/test_tile_cached_authz_liveness_1451.py \
       tests/test_raster_ingest_orphan_cleanup_gap017.py -q  # 255 passed
```

The tests were confirmed non-vacuous by breaking the implementation and re-running. Moving the probe below the DROP fails 1: the recorded oid becomes NULL while the delete still succeeds, which is the failure mode that equality exists to catch. Stamping `previous_owner_id=None` fails 4.

Two test doubles needed the new read. `_MockRecord` in `test_raster_ingest.py` gained `created_by`, and one mocked session in that file returned `None` from every `execute`, which the new `.scalar()` read tripped over. It now returns a result whose scalar is None, the honest answer for a session with no database behind it.

`service_lifecycle.py`'s LOC cap went from 415 to 483 across the two rounds, with a comment for each saying what the lines bought. `domain/models.py` is at 986, still under the 1000-line ratchet inclusion threshold.

Refs #1456

Closes #1456

### DIFF
--- a/backend/alembic/versions/0045_retired_table_names_identity.py
+++ b/backend/alembic/versions/0045_retired_table_names_identity.py
@@ -1,0 +1,72 @@
+"""Record the freed relation's identity and its prior owner on a tombstone.
+
+fix(#1456). ``catalog.retired_table_names`` recorded only the freed NAME. Two
+review findings on #1453 converge on the same gap: a tombstone cannot tell
+"the table I detached" from "a new table wearing its name", and it cannot say
+who owned the dataset that freed it. Both answers have to be captured inside
+the delete transaction, because both sources die in it — the relation is
+dropped, and the ``catalog.records`` row carrying ``created_by`` is deleted.
+
+This migration only makes the data exist. Nothing reads the new columns yet;
+the tombstone-vs-not decision is unchanged and still keyed on whether the
+delete freed the name.
+
+``relation_oid`` is BIGINT, not INTEGER: a pg_class oid is an unsigned 32-bit
+value, so int4 overflows on any oid past 2^31 and a cluster that has churned
+enough objects hands those out. It is forensic identity within ONE cluster
+lifetime and nothing more — oids are not preserved by pg_dump/pg_restore, so
+every row here reads as a stale oid on the far side of the RUNBOOK's backup
+and restore path. The durable half of the identity is ``previous_owner_id``.
+
+``previous_owner_id`` is deliberately NOT a foreign key to ``catalog.users``.
+This table is retain-forever, and an FK gives only bad options: ON DELETE
+CASCADE would erase tombstones when a user is deleted (silently re-arming
+GH-1443 for every name that user's datasets freed), RESTRICT would let a
+tombstone block a user deletion, and SET NULL would quietly discard the one
+durable half of the identity. The same reasoning already governs the
+``dataset_id`` column beside it, whose referent is deleted in the very
+transaction that writes the row.
+
+Both columns are nullable and no backfill is attempted, because none is
+possible by construction. Every pre-existing tombstone was written after its
+relation was dropped and its record row deleted, so neither value survives
+anywhere to be recovered. NULL means "not recorded", not "no owner" — a
+consumer must treat a NULL as unknown, never as a match.
+
+Revision ID: 0045_retired_table_names_identity
+Revises: 0044_retired_table_names
+Create Date: 2026-08-13
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+revision: str = "0045_retired_table_names_identity"
+down_revision: Union[str, None] = "0044_retired_table_names"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "retired_table_names",
+        sa.Column("relation_oid", sa.BigInteger(), nullable=True),
+        schema="catalog",
+    )
+    op.add_column(
+        "retired_table_names",
+        sa.Column("previous_owner_id", postgresql.UUID(as_uuid=True), nullable=True),
+        schema="catalog",
+    )
+    # No index on either column. Nothing queries them yet, and the probe this
+    # table exists to serve is still the name lookup that
+    # ix_retired_table_names_table_name backs. A reader that later matches on
+    # identity reaches these columns through that same name row.
+
+
+def downgrade() -> None:
+    op.drop_column("retired_table_names", "previous_owner_id", schema="catalog")
+    op.drop_column("retired_table_names", "relation_oid", schema="catalog")
--- a/backend/alembic/versions/0046_detached_relations.py
+++ b/backend/alembic/versions/0046_detached_relations.py
@@ -1,0 +1,108 @@
+"""Record a relation GeoLens released a claim on while it was still standing.
+
+fix(#1456 review round 1). 0045 stamped the freed relation's identity onto the
+tombstone, which covers every delete that RELEASES a name. It does not cover
+the case #1456's window 1 is actually about: a detach that leaves the
+operator's table standing writes no tombstone at all, so the oid and the prior
+owner were probed and then discarded. If the operator drops that relation after
+the delete commits, the name goes free with nothing recorded anywhere, which is
+precisely the window the identity was collected to let a later change close.
+
+Both values are as unrecoverable here as they are on the freed paths: the
+record row carrying ``created_by`` dies in the delete transaction, and the oid
+dies with the relation whenever the operator gets around to dropping it.
+
+A SIBLING TABLE, not a flag on ``catalog.retired_table_names``. That table's
+entire API is set membership: ``generate_table_name`` asks "is this name
+retired?" and ``register_existing_table`` refuses any name that answers yes.
+A row recording a relation that still holds its name is not a prohibition, and
+putting one in that set would either burn the operator's own table name (if the
+readers ignore the discriminator) or make every future reader responsible for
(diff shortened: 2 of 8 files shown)


ishiland:
@codex review

ishiland:
P1 was right, and it lands on this PR's own justification. Window 1 lives entirely on the path that writes no tombstone, so recording identity alongside retirements collected it for every case except the one it was collected for. Both values are as unrecoverable on that path as on the freed ones: `created_by` dies with the record row in the delete transaction, and the oid dies with the relation whenever the operator drops it.

Fixed in 8f2f1cc7a. Migration 0046 adds `catalog.detached_relations`, written by the `elif` branch in `delete_dataset` when the detach leaves the relation standing.

Sibling table rather than a flagged tombstone, deliberately. `retired_table_names` has one API and it is set membership: `generate_table_name` asks whether a name is retired and `register_existing_table` refuses any name that answers yes. A relation still holding its name is not a prohibition, so a discriminated row there would either burn the operator's own table name or make every future reader of that probe responsible for a predicate whose failure direction is silent. Neither consumer changes. `test_the_record_does_not_retire_the_name` asks both directly: registration still accepts the operator's surviving table, and the collision walk still refuses the name because a live relation occupies it, not because a record retired it.

ishiland:
@codex review

PR #1471, opened 2026-08-13
feat(demo): live-service quakes, showcase metadata enrichment, and 2025-season HURDAT2

ishiland:
Modernizes `scripts/seed-showcase.py`, plus the two rows of `THIRD_PARTY_DATA.md` that describe the sources this changes. No backend, no frontend, no API surface, no schema. The motivation is freshness (two of the showcase datasets had gone stale, and one map name had gone false) and surfacing capabilities that already shipped but nothing in the demo exercised.

Running any of this against the live demo is a separate operational step. This PR only changes the seeder and the provenance inventory that documents it.

## Part 1: the quakes datasets now read from a live service

They used to download a USGS GeoJSON feed and upload it, and `--refresh-quakes` re-downloaded and re-uploaded. Both datasets now bind to the USGS Recent Earthquakes MapServer, so a refresh is two POSTs and a poll with no data crossing this machine.

Conversion happens in place. The dataset id, record id, table name and every map layer survive it, so an instance that was seeded months ago upgrades on the next run without its maps being touched. A fresh instance takes the same request body through `/services/preview/` instead.

Two consequences follow. The connector hardcodes `where=1=1`, so there is no server-side filter and the choice is the whole layer or none of it: the feed is M2.5+, not M4.5+, and the datasets are retitled to say so. Lookups try the new title, fall back to the old one once, and patch it forward, so one run migrates a live instance and later runs are clean. The service also brings its own column names: `depth_num` and `event_time_utc_date_fmt`, where the seeder used to derive `depth_km` and `time_utc`. Styles speak the service's vocabulary permanently, because every refresh rebuilds `column_info` from the service and a column rename would not survive it. That is recorded as a gotcha in the module docstring.

`--refresh-quakes` keeps its name and is now a thin wrapper. Sentinel-2 scenes also get one refresh each at the end of a normal seed, which gives their Source panel a run history to show.

## Part 2: catalog metadata

`SHOWCASE_METADATA` now carries `source_organization`, `source_url`, `update_frequency`, `data_vintage_start`, `data_vintage_end` and `theme_category` alongside the license and keywords.

Frequencies come from the vocabulary `chk_records_update_frequency` actually enforces, so nothing here can fail the CHECK. Vintages are set only where the source states one, which is why several datasets have none: ETOPO 2022 is a compilation with no single coverage window, and the volcano database starts in 4360 BC, which a date column cannot hold. The three analysis-derived datasets deliberately carry no `source_organization`. No outside body published them, and crediting NOAA for a number GeoLens calculated would be wrong.

Both manifest ingests now pass `attribution`, and the Restless Planet collection description was reworded: it promised earthquakes and eruptions to a STAC client that only ever sees the collection's raster members.

## Part 3: HURDAT2 through the 2025 season

The file bumps to the 2025 release. NHC changed the filename date suffix from MMDDYY to MMDDYYYY, which is noted in a comment because a script pattern-matching the old form will not find the new file.

That makes "Hurricane Alley - 75 Years of Major Atlantic Storms" wrong, since 1950 to 2025 is 76 seasons. The map becomes "Hurricane Alley - Major Atlantic Storms Since 1950", which carries no year count and cannot rot again. Builders skip by name, so the rename is applied before the exists-check; without that the next run would build a duplicate beside the old map.

A plain seed reuses datasets by title and would never pull the new file, so `--refresh-hurdat2` does it: both track datasets swap in place with their ids intact, then the derived exposure chain is deleted and recomputed. The chain cannot refresh in place because materialize always registers a new dataset. The exposure map keeps its id, its derived layer is deleted before the datasets it points at, and it is re-added through `add_layer` with the full style body rather than a PATCH.

## Part 4: styling

A hatched context layer (`geolens-fill-hatch`) sits under the exposure choropleth showing the regions that were offered to the intersect but not reached; the graded fill stays clean. Folder groups fold Restless Earth into Hazards and Context and the Matterhorn into Terrain and Routes & Peaks. All seven maps get a legend title and notes. The meteorite and earthquake circle layers get `circle-pitch-alignment: map`, which lays them flat on the globe instead of turning them toward the camera.

All of it runs as a post-builder pass, following `apply_globe_projection`, because builders skip a map they already found and anything set at creation time would reach a fresh instance and never an existing one. Existing layers are restyled by reading the layer back from the server, applying the delta and re-posting it, rather than through the layer-diff PATCH path that can clobber `style_config`.

## Part 5: `--prune-userdata`

Dry run by default: it prints a report and deletes nothing. `--prune-userdata --execute` performs it.

Ownership decides, not titles. Maps and datasets owned by other accounts are the delete set. Admin-owned content the seeder does not recognize is reported as a stray and kept, because the live demo carries hand-uploaded datasets that predate this script. The three externally pinned datasets are held back whoever owns them. Maps are deleted before datasets, since a layer is a live reference. Collections are reported only.

## Verification

Against the local dev stack, with throwaway datasets deleted afterward:

- `python3 -m py_compile` after every edit batch, clean.
- Conversion smoke: ingested a 3-point dataset, converted it against the USGS URL, asserted dataset id, record id and table name unchanged, `origin` moved `upload` to `service`, 3 rows became 2308, and `depth_num` / `event_time_utc_date_fmt` / `mag` are present with `place` / `felt` / `tsunami` / `sig` surviving. Refreshed it twice; both runs recorded `succeeded`.
- Fresh service ingest verified separately, since a new instance has nothing to convert: probe, `/services/preview/`, commit, poll gives `origin: service` with the same columns.
- Layer restyle round-trip: every writable field survived a delete-and-re-add, including an expression filter, a nested `style_config` with `render_mode`, `label_config` and `show_in_legend: false`, with both deltas applied.
- `--prune-userdata` dry run: exit 0, correctly bucketed 11 admin-owned strays and the 3 pinned datasets, found no foreign content, and dataset and map counts were identical before and after.
- `THIRD_PARTY_DATA.md`: exactly two rows changed, the earthquakes row and the HURDAT2 revision. Verified that the license and attribution cells are byte-identical on both, since neither changed (US-gov public domain either way), and that every row still has the same column count.
- `ruff check` and `ruff format --check`: the file is formatted, and the lint result matches the base commit exactly, with the same four pre-existing C901 findings and no new ones, with `build_restless_earth` down from 18 to 16. The pre-commit ruff hooks are scoped to `^backend/.*\.py$`, so this path is not gated by them.

`--refresh-quakes` itself was not run end to end, because dev has no seeded quakes datasets. It goes through the same probe and refresh endpoints the smoke exercised.

The full seeder was not run against dev: it downloads a 466 MB GeoTIFF and 62 DEM tiles.


ishiland:
@codex review

ishiland:
Fixed the two code findings in 583b0749a.

The P1 was real and worse than it looked. `build_restless_earth` returned at its `_map_exists` check before `ensure_quake_datasets` ran, so every already-seeded instance, which is the only kind this upgrade is for, would have kept its uploaded M4.5+ datasets. `--refresh-quakes` could not rescue it either, for the reason you gave. The binding now runs ahead of the skip, matching what `build_meteorites` already does with its own dataset healing. Verified live: converted a throwaway dataset in place and confirmed the id, record id and table name survive while `origin` moves from `upload` to `service`.

The P2 restyle window is fixed by restoring the original layer body when the replacement POST fails, then re-raising. Verified live by forcing a timeout on the replacement: the original layer comes back byte-identical and the caller still reports the failure, so the worst case is now an unapplied style delta rather than a silently missing layer.

On `THIRD_PARTY_DATA.md`: you are right, and I have not fixed it here. I confirmed both rows are stale, line 16 on the magnitude, the retrieval mode and the transform, and line 19 on the HURDAT2 revision. This PR is deliberately scoped to `scripts/seed-showcase.py` only, so I am flagging the inventory update rather than folding it in. Deferred by scope, not disputed.

ishiland:
@codex review

ishiland:
Both round-2 findings were real and are fixed in 5deea23f2.

The P1 was a hard failure on fresh instances and my earlier smoke had missed it, because it only ever imported once. I confirmed the guard in `sources/router.py` and then confirmed both halves live: a second `POST /services/preview/` for the same layer returns 409 `duplicate_source`, while a second conversion through `reupload/service/preview` is accepted. So every service dataset the seeder creates is now a one-point stub that is immediately converted.

I took the "reuse one bound dataset for both layers" option off the table because the two datasets are separately titled, collected and referenced, and collapsing them is a much larger migration than the bug needs. I also did not keep a "first one imports, second one converts" rule: it breaks whenever only the *other* dataset exists, since the import then collides with the surviving binding. One path has no such edge. Verified live on a throwaway pair: both datasets bind from the one USGS layer, `placeholder` is gone after conversion, both land 2310 features with `origin: service`, and a second pass converts nothing.

On the exposure chain, agreed. The replacement is now materialized beside the old chain, the map layer is swapped onto it, and only then is the superseded chain deleted, so a worker or analysis failure leaves the map exactly as it was instead of stranded. `_build_exposure_chain` gained a `force_analysis` flag so the derived datasets recompute while the two ingested inputs stay reused, since `--refresh-hurdat2` has already swapped those in place. The tradeoff is that the three derived datasets exist twice for the length of the rebuild, which only matters where a dataset quota is tight; that is noted in the code.

`THIRD_PARTY_DATA.md` is still outstanding and still out of scope for this PR, as noted above.

ishiland:
@codex review

ishiland:
Both round-3 findings were real and are fixed in 8452747ba. This is where I stop the review loop: my budget was two rounds and this is the third, so I am handing the state over rather than spending another.

The P2 was the more serious of the two and I am glad it was caught. `build_restless_earth` converts the quake data and then returns early on an existing map, so an upgraded instance kept a popup listing `depth_km` and `time_utc` against a table that no longer has those columns. The styling pass now repairs the popup fields and the heatmap weight floor. Both values moved into shared constants that the builder and the repair read, so there is one definition of correct rather than two that can drift. Verified live against a throwaway map seeded with the exact pre-conversion styles: fields go from `[depth_km, time_utc, felt, tsunami]` to the service vocabulary, the weight floor goes from 4 to 2.5, and a second pass recreates no layer at all.

On the vintage, agreed, and it contradicted a rule this file already states about not writing a vintage the source does not support. The HURDAT2 datasets now carry `data_vintage_start` and no end. 1950 is a floor the script imposes itself so it holds for any release, while the end is only knowable once `--refresh-hurdat2` has run. The map note no longer names a season either.

Outstanding and deliberately not addressed here: the `THIRD_PARTY_DATA.md` rows from round 1. Still correct, still outside this PR scope, and flagged to the orchestrator.

ishiland:
Correction to my earlier replies: the `THIRD_PARTY_DATA.md` finding from round 1 is now **fixed in this PR**, not deferred. I had scoped this branch to the seeder alone; that constraint has been lifted, and since the inventory going stale is a direct consequence of this change it belongs in the same PR.

Fixed in 9b6e0bd1c. Exactly the two affected rows:

The earthquakes row now reads M2.5+, names the ArcGIS REST service endpoint and layer instead of the old 30-day GeoJSON feed, records the retrieval as a live service binding re-pulled server-side by `--refresh-quakes` rather than a seed-time download, and drops the transform description entirely, because there is no longer a client-side transform: the connector applies no attribute filter, so the layer is bound whole and the styling reads the service column names.

The HURDAT2 row now names the 1851-2025 revision with the new filename.

License and attribution cells are untouched on both rows, since neither changed; both remain US Government public domain. I verified that mechanically rather than by eye, along with the table shape staying intact at the same column count on every row.

ishiland:
@codex review

ishiland:
All four fixed in 8fd667f13. Three were clean bugs; the fourth I have addressed rather than solved, with reasoning below.

The duplicate-layer one was the real find, and it was a defect in the build-first flow I added last round rather than a pre-existing one. You are right that after a partial failure `datasets_by_title()` returns the orphaned newer ids, `old_layer_ids` comes back empty, and the retry stacks a second choropleth. The layer is now resolved by its display name, which is stable whatever dataset it points at. I also widened the cleanup: it now sweeps every chain-titled dataset except the three just built, so a failed run's orphans get collected instead of accumulating one set per attempt. Verified against a simulated partial-failure state: resolving by id finds nothing, resolving by name finds the layer, and the sweep deletes six datasets (old chain plus orphans) while keeping the new three and touching nothing else.

`--refresh-quakes` exiting 0 on a missing dataset was a straightforward miss, agreed, and it matters exactly because this is the cron entry point. It now counts as a failure. Verified live against dev: exit code 1.

On the pinned exemption, I have kept the title-based hard-keep and changed what it reports, rather than binding it to ids. The requirement this implements is that the three externally referenced datasets survive a cleanup whoever owns them, and narrowing to known ids would drop that protection on any instance where they were re-created. Your underlying point stands though, so a pinned-title dataset owned by another account now lands in its own reported bucket with an explicit note that titles are not unique, instead of hiding among the genuine three. It is still not deleted; it is just no longer invisible.

On the input consistency during `--refresh-hurdat2`, your reading is correct and I have not made it transactional, because I do not think it can be from here. Reupload is in-place by design and exposes no staging handle to roll back to, and the chain can only be computed from the already-updated legs, so there is no ordering that avoids the window. Staging the legs as a separate dataset would work but changes the legs dataset id, which the map layer and the callers depend on staying stable. What I have done instead is stop the failure from being misleading: it now says plainly that the tracks were replaced and the chain was not, that the map is showing new legs against an older choropleth, and that re-running finishes the job. If you think the id-changing variant is worth it despite the churn, say so and I will take it up separately rather than in this PR.

ishiland:
@codex review

ishiland:
All four fixed in 15d744ee7. This is the end of my review budget, so I am stopping the loop here and handing over.

Three of these were the same defect wearing different hats, and it is the defect this whole PR keeps tripping over: a correction that is only written at creation time never reaches the instances that carry the mistake. The Restless Earth description still said M4.5+ on any seeded instance, the Restless Planet collection kept its pre-STAC wording because `build_collections` reused by name and only topped up membership, and `--refresh-hurdat2` never reached the enrichment pass because it is a terminal mode, so every successful refresh reset the exposure chain to a proprietary license with no keywords. All three now migrate on an existing instance. The map description moved to a shared constant so the builder and the styling pass cannot drift, and both migrations were verified live against throwaway objects, including that a second pass is a no-op.

You were right about attribution and I had overclaimed it. I confirmed the backend stores it as `job_metadata["manifest_attribution"]` and that no map-layer response schema carries an attribution field, so nothing renders it over the tiles. I have kept passing it, since it is the documented manifest field and belongs in the provenance trail, but the comments now describe what actually happens instead of a viewer credit that does not exist. Wiring it through to the layer API is a backend change and does not belong in this PR; worth an issue if the swisstopo and NOAA credits need to be visible on the map itself, which their terms arguably require.

Tally across five rounds: fifteen findings, all real, none refuted, all fixed. That is a good hit rate and it means I would not read this PR as clean so much as bounded, so a further pass may well find more.

ishiland:
Before triggering the final round I did a targeted self-pass with the two lenses this loop has been teaching: is any remaining correction creation-time only, and does any migration path run twice destructively. Fixed in a397ea13e.

Three findings, all the first kind:

The circles dataset summary was create-time only. Its heatmap sibling has carried a summary in `SHOWCASE_METADATA` since #614 for exactly this reason, so the enrichment pass could rewrite it, and the circles dataset had no such entry. An upgraded instance would have gone on describing an M4.5+ download indefinitely. Both summaries now live in one shared constant that the creation path and the enrichment pass both read.

The Hurricane Alley rename ran only inside `build_hurricanes`. The styling and globe passes resolve that map by its current name, and that builder does not run under `--only` or after an earlier builder fails, so the map could keep its legacy name and quietly miss its legend, notes and projection. The rename now also runs in the post-builder path.

The collection description was PATCHed unconditionally on every seed, unlike every other migration here, bumping `updated_at` and reordering any listing sorted by it. It compares first now, verified live: first run migrates, second writes nothing and `updated_at` is unchanged.

Nothing surfaced under the second lens. Every migration compares or is guarded before writing, and the exposure orphan sweep keeps only the chain it just built, so repeat runs neither accumulate nor destroy.

ishiland:
@codex review

ishiland:
Both round-6 findings were real and are fixed in e743ac14f. Per the agreed budget I am NOT re-triggering; this is the hand-off state.

The popup gate is the important one, and it is a case of a repair being able to make things worse than it found them. The conversion runs inside an isolated builder while the styling pass runs afterwards either way, so a transient conversion failure left an upload-origin dataset carrying a popup that named service columns it does not have. The stale popup at least showed real values; the replacement would have shown blank rows. It is now gated on the dataset actually having those columns, read from the `dataset_column_info` the layer response already carries, so there is no extra request and it declines whenever the columns cannot be confirmed. I left the heatmap weight ungated deliberately: it reads `mag`, which both the upload and the service carry, and a 2.5 floor over M4.5+ data is inert rather than wrong.

You were right that the DELETE is ambiguous rather than binary. A failed delete now re-reads the map and lets the actual state decide: still present means nothing was lost and the delta simply does not apply, absent means the delete landed and the re-add becomes the recovery path.

Writing the test for the first fix turned up a third defect that neither of us had flagged. `MAP_LAYER_STYLE_FIXES` was missing from the styling pass reachability guard, so it reached Restless Earth only because that map happens to appear in two other tables. A style fix added for a map listed nowhere else would silently never have run, and my first test run passed for exactly the wrong reason: the map was being skipped entirely rather than the gate working. Every table that can carry work for a map is now in the guard, and all three cases verified live afterwards, including the negative one.

ishiland:
@codex review

ishiland:
All three round-7 findings were real and are fixed in 154f2dfbc. Per the agreed budget I am not re-triggering; this is the hand-off state.

The rename ordering was the significant one, and it is the same defect as the round-6 popup gate wearing a bigger coat. I gated the popup on the columns existing but left the title itself ungated, and the title is the key `SHOWCASE_METADATA` is looked up by. Since builders are isolated and the enrichment and styling passes run regardless, a conversion that died mid-way left the instance advertising a live M2.5 service over unchanged M4.5 rows: new summary, new description, new notes, old data. Finding and renaming are now separate operations and the rename runs last, so a failed conversion leaves the old title, which matches no metadata entry and therefore claims nothing. `--refresh-quakes` adopts the new title as well, but only once `origin` proves the conversion actually landed, which has the useful side effect of finishing a seed that converted and then failed to rename. Verified both directions live: forced conversion failure leaves the legacy title, success renames.

`refresh_and_report` reading the count outside its own error handling was a clean miss, and worse than it looks because that helper runs from `refresh_sentinel2_scenes`, outside `main()`s builder isolation, so a transient failure on a decorative lookup would abort a finished seed and skip every remaining scene. Verified: a failing count lookup after a successful refresh now still reports success and logs the count as unknown.

You were right that I fixed only half the ambiguity last round. The POST has exactly the shape the DELETE did, and creation is not idempotent, so a lost response after a committed create meant restoring produced a permanent duplicate that every later pass then preserved and restyled. The map is re-read before deciding. Where even that read fails I still restore, deliberately: a missing layer is a visible hole in a public map, a duplicate merely draws twice, and only one of those is obvious enough for an operator to notice and fix. Verified with a forced committed-then-lost create: one layer, no duplicate.

Seven rounds, twenty findings, all real, none refuted.

ishiland:
Class sweep done, in 057fd624a. Not a codex round: this enumerates the sibling set behind the defect codex found twice at different layers (the popup config in round 6, the dataset title in round 7), rather than waiting to be shown a third instance.

The question asked of every write that can execute downstream of an isolated builder failure: what does it advertise, what must have succeeded for that to be true, and what OBSERVED state gates it. Twenty-two writes, four of them still advertising a live service without evidence.

The Restless Earth notes and description both describe a live M2.5 feed and are written by a pass that runs whether or not the conversion succeeded. The heatmap dataset summary and its `continual` maintenance frequency claim the same thing and had no protection at all, because the rename-after-conversion gate only covers the dataset that has a legacy title to move off. The Restless Planet collection description called them live earthquake feeds.

The first three now check observed `origin`, which the datasets list already carries, so the gate costs nothing and reads the same class of evidence the popup gate reads from `column_info`. The collection description instead drops the claim: a collection is a label over datasets rather than a view of one, so it has no natural gate, and describing what it holds rather than how it is delivered is true on every instance. Same reasoning that stopped the hurricane map name counting years.

One row was already protected but only by inference. The circles dataset metadata is keyed on a title that is adopted after conversion, so the title was standing in as a proxy for evidence. It checks origin directly now, which is the point of the exercise.

Nothing that makes no claim was gated. Licenses, source organizations, theme categories and legend titles are properties of a dataset identity rather than of a builder succeeding, so they still apply unconditionally, and the heatmap weight repair stays ungated because it reads a column both vocabularies carry.

Verified in both directions against a simulated failed conversion: with origin still `upload`, no live claim is published anywhere while licenses and legend titles still land; with origin `service`, every held claim publishes. The full table is in my report.

ishiland:
@codex review

ishiland:
All three round-8 findings were real and are fixed in 4aaad3c31. Per the standing instruction I am not triggering round 9; the next step is a human review pass over the full diff.

The prune finding is the one that mattered most, and it is the only genuinely destructive defect this review has surfaced. I confirmed both `Map.created_by` and `Record.created_by` are `ON DELETE SET NULL`, so deleting a user strips ownership from everything they made and leaves it looking exactly like another account content to a "not mine" test. An operator whose account had been removed would have watched `--prune-userdata --execute` delete their own maps. Ownerless is unknown, not foreign, so those items now go to their own reported bucket and are never in the delete set.

The live-service gate was under-scoped and I should have caught that in the sweep. The wording it guards describes both quake layers, the map draws them from two datasets, and they convert in sequence, so the second can fail after the first succeeded. Checking only the circles published "read live" over a heat layer still holding the old upload. Both are checked now, verified across all four combinations including the missing-dataset case.

The README copy is the same class as the `THIRD_PARTY_DATA.md` rows from round 1: public text describing what the seeder produces, made false by the change. Fixed in `README.md`, its German, French and Spanish translations, and `scripts/README.md`. All five now name the start year rather than counting from it, so they cannot rot again, and the translations moved in step rather than drifting from the English.

Eight rounds, 23 findings, all real, none refuted.

ishiland:
Your finding is fixed in e2bd2a064. You were right that this was the active class on the one path nobody swept: my sweep enumerated writes that run *downstream* of a failed builder, and this write runs *inside* one, so the whole table looked past it.

`ingest_service` committed the stub with the live-service summary a step before the conversion that would make it true, and nothing downstream could retract it: the origin gates choose whether to overwrite a summary and cannot un-write one already committed. A transient failure on a fresh seed left a public dataset describing a live USGS feed over a single placeholder point at Null Island.

The stub now commits with a neutral summary that is true when written and stays true if the binding never completes, and the real description arrives from the origin-gated enrichment pass later in the same run.

One deliberate departure from your wording, flagged because you are re-reading this diff. Rather than passing a neutral summary, I removed the `summary` parameter from `ingest_service` entirely, so a caller cannot hand it a claim it has not yet earned. Same outcome for this call site; it also closes the seam for the next one, which felt like the point of the exercise. Say the word if you would rather have the smaller diff and I will put the parameter back.

The title still gets its real value at creation, as you said: it is the idempotency key every later lookup depends on, and with the summary gated it no longer carries the substantive claim.

Verified with the existing harness in both directions. Fresh seed with the conversion forced to fail: the stub is created, origin stays `upload`, the summary is the neutral text, and the enrichment pass holds both the live summary and `continual` while still applying the license. Then with origin `service`: both land. Full self-pass re-run clean, lint unchanged at the four pre-existing findings, CI green.

PR #1475, opened 2026-08-13
fix(standards): publish real raster access surfaces in the DCAT feeds

### LINKED ISSUE #1469: fix(standards): DCAT raster distributions expose raw storage keys; stac-origin datasets have none
Found during an anonymous audit of the public DCAT export (`GET /api/datasets/dcat`) on a v1.12.0 instance. Vector datasets serialize correctly (24 datasets, each with `OGC API Features` + `Vector Tiles` distributions carrying `dcterms:title` and `dcat:mediaType`). Raster datasets do not:

**1. COG-backed rasters advertise a raw internal storage key as `dcat:accessURL`.**

ETOPO 2022 and the swissALTI3D mosaic each expose exactly one distribution, with no `dcterms:title`, no `dcat:mediaType`, and an `accessURL` like:

```
rasters/6e9ca821-3345-4c30-bcf8-935bc6dbd61f/1954d0…c080b1/source.cog.tif
```

That is a relative object-storage key with no scheme or host: unresolvable by any DCAT consumer, and it leaks the internal storage layout plus a content hash.

**2. `stac`-origin rasters expose zero distributions.**

All four Sentinel-2 by-reference datasets appear in the feed as `dcat:Dataset` entries with no distribution at all, so DCAT consumers see them as data with no access method.

**3. No raster-tile distribution is advertised anywhere in DCAT**, even though `/raster-tiles/{id}/tiles/{z}/{x}/{y}.png` serves them and the STAC surface advertises exactly that template as a `raster_tiles` asset. Raster access is systematically under-published relative to STAC.

Suggested direction: the raster branch of the DCAT serializer should emit the same access surfaces the product actually exposes (raster tile template, and the download/export endpoint where one exists) with proper titles and media types, and never fall through to the storage key. The Sentinel case needs the tile-template distribution too, since there is no local COG to point at.

Verification once fixed, against any instance with a public COG raster and a STAC-imported scene:

```
curl -s $BASE/api/datasets/dcat | jq '..|.["dcat:accessURL"]? // empty' | grep -c '^"rasters/'   # expect 0
```

and every `dcat:Dataset` should carry at least one distribution.

ishiland:
Closes #1469

## Problem

The DCAT-family serializers mapped `record_distributions` rows straight onto distribution nodes. That works for vector datasets, whose rows are written by `generate_distributions` and hold root-relative API paths. Raster datasets have three separate defects:

1. **COG-backed rasters published a storage key.** The raster/VRT ingest tails write `RecordDistribution(url=cog_key)` — a bare object-storage key like `rasters/<id>/<hash>/source.cog.tif`, with no title and no media type. DCAT 3 emitted it verbatim (`_distribution_to_dcat` only prefixed urls starting with `/`). DCAT-US and GeoDCAT-AP were worse in a quieter way: their `_absolute_url` had a third branch that glued the key onto the API origin, producing `https://host/api/rasters/<id>/<hash>/source.cog.tif`. Neither form resolves for a consumer, and both expose the internal storage layout plus a content hash.

2. **STAC-origin rasters published nothing.** `stac_router.py` creates the Record and Dataset but writes no distribution row, so those datasets appeared in every feed as `dcat:Dataset` entries with no access method at all.

3. **The raster tile template appeared nowhere**, even though `build_assets` advertises exactly that template as the STAC `raster_tiles` asset for the same datasets. The two surfaces described one dataset differently.

## Fix

New `backend/app/standards/distributions.py` is the one place that decides what a feed may publish:

- `is_publishable_url` drops stored urls that are neither root-relative nor absolute http(s). The rule cannot drop a legitimate row: everything a user authors goes through `DistributionCreate`, whose validator requires an http(s) url, and everything `generate_distributions` writes is a root-relative API path. Storage keys are the only thing left.
- `published_distributions` adds the raster family's real access surfaces — the tile template (title, `image/png`), plus `/datasets/{id}/download/cog` for `raster_dataset` only, since `download_cog` rejects a VRT with 400. STAC-origin rasters get both: their raster asset lives on a remote host and the download endpoint redirects to it.

Both raster entries are derived per request rather than stored, because they depend on values a row cannot hold: the tile template is served at the **app** origin (`/raster-tiles/...` is nginx-rewritten to the tile proxy; the API origin has no such route) and carries the dataset's current `tile_cache_version`. That is why `app_base_url` is threaded into the six serializer entry points and resolved with `get_public_urls` at the 12 handlers in `router_export.py`, replacing `get_public_api_url` (which is literally `get_public_urls(...)[1]`, so the API half is unchanged).

`raster_tiles` joins `vector_tiles` in each profile's `SERVICE_DISTRIBUTION_TYPES`, so the two tile surfaces serialize alike.

The OGC Records profile (`properties.distributions`) republished the same storage key. It gets the drop half only — `build_assets` already publishes its raster access surface as the `raster_tiles` asset, so synthesizing a second copy there would duplicate it.

## Verification

```
uv run pytest tests/test_dcat.py tests/test_geodcat_ap.py tests/test_distribution_reconcile_1314.py \
  tests/test_metadata_roundtrip.py tests/test_ogc_discovery.py tests/test_layering.py tests/test_datasets.py -q
# 214 passed

uv run pytest tests/test_ogc_record_properties.py tests/test_ogc_record_enrichment.py \
  tests/test_ogc_records_conformance.py tests/test_ogc_collection_metadata.py \
  tests/test_distribution_primary_1383.py tests/test_raster_tile_url_version_1372.py \
  tests/test_record_subresource_edits.py tests/test_stac*.py -q
# 381 passed

PYTHONPATH=. uv run python scripts/dump_openapi.py --check   # no drift
```

New tests in `tests/test_dcat.py` cover the issue's acceptance criteria across all three profiles: no accessURL matches `^rasters/`, every dataset carries at least one distribution (including a STAC-origin raster), and the tile template appears for both a COG-backed and a STAC-origin dataset. `test_dcat_feed_never_exposes_a_storage_key` was confirmed non-vacuous by stubbing `is_publishable_url` to return True — it fails.

## Impact

No schema or API-contract change; no OpenAPI drift. `_MODULE_LOC_CAPS` for `search/service_records.py` goes 519 -> 524 (the filter plus its comment), and `_STANDARDS_MODULE_IMPORT_SURFACE` gains the new module's TYPE_CHECKING-only catalog-ORM edge — the same edge the three serializers already carry, now in one file instead of three.

### DIFF
--- a/backend/app/modules/catalog/datasets/api/router_export.py
+++ b/backend/app/modules/catalog/datasets/api/router_export.py
@@ -59,7 +59,7 @@ from app.modules.catalog.datasets.domain.service import get_dataset
 from app.core.dependencies import get_db
 from app.core.db.tenant_session import current_tenant_var
 from app.core.tenancy import is_multi_tenant
-from app.core.public_urls import get_public_api_url
+from app.core.public_urls import get_public_urls
 from app.platform.extensions import get_catalog_port
 from app.platform.storage import get_storage
 from app.platform.storage.titiler_url import resolve_storage_key
@@ -307,11 +307,12 @@ async def get_dcat_catalog(
     """DCAT 3 JSON-LD catalog feed. Respects dataset visibility."""
     datasets = await _get_visible_dcat_datasets(db, user, limit=limit, offset=offset)
 
-    base_url = await get_public_api_url(db, request=request)
+    app_base_url, base_url = await get_public_urls(db, request=request)
     preferred_languages = parse_accept_languages(request)
     catalog = catalog_to_dcat(
         datasets,
         base_url,
+        app_base_url=app_base_url,
         preferred_languages=preferred_languages,
         lineage_by_record_id=await _visible_lineage(db, datasets, user),
     )
@@ -345,10 +346,11 @@ async def validate_dcat3_catalog(
     """Validate the visible W3C DCAT 3 catalog feed."""
     datasets = await _get_visible_dcat_datasets(db, user)
 
-    base_url = await get_public_api_url(db, request=request)
+    app_base_url, base_url = await get_public_urls(db, request=request)
     catalog = catalog_to_dcat(
         datasets,
         base_url,
+        app_base_url=app_base_url,
         lineage_by_record_id=await _visible_lineage(db, datasets, user),
     )
     report = validate_dcat3(catalog, "Catalog")
@@ -380,10 +382,11 @@ async def get_dcat_us3_catalog(
     """DCAT-US Schema v3.0 catalog feed. Respects dataset visibility."""
     datasets = await _get_visible_dcat_datasets(db, user, limit=limit, offset=offset)
 
-    base_url = await get_public_api_url(db, request=request)
+    app_base_url, base_url = await get_public_urls(db, request=request)
     catalog = catalog_to_dcat_us3(
         datasets,
         base_url,
+        app_base_url=app_base_url,
         catalog_contact_email=settings.dcat_contact_email,
         lineage_by_record_id=await _visible_lineage(db, datasets, user),
     )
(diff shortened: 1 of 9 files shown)


ishiland:
@codex review

ishiland:
Round 1 finding applied in 6241eae: the COG download distribution is gone. You are right that it was unusable for this feed's audience — `_resolve_download_user` (backend/app/modules/catalog/datasets/api/router_export.py:812) 401s a caller with neither credentials nor a download-scoped `?token=`, and minting one is a separate POST to `/auth/download-token/{id}`.

Chose to stop publishing it rather than open anonymous access: the token handshake is deliberate (SEC-04), and dropping the entry restores exact parity with `build_assets` (backend/app/modules/catalog/search/service_records.py:128), which advertises `raster_tiles` and no COG download for the same datasets. The tile template has no such gate — a public, published raster serves tiles anonymously (`TestRasterAuthCheck::test_auth_check_returns_open_path_for_public_raster`).

@codex review

PR #1476, opened 2026-08-13
fix(installer): pull before the upgrade outage and restore the app if it fails

### LINKED ISSUE #1467: upgrade.sh applies data migrations while the old app is still serving writes
Found while reviewing #1466 (the OGC:WMTS relabel). The specific case is handled there; this is the general class.

## The window

`scripts/upgrade.sh` runs migrations before it replaces the app containers:

- Step 6 (`scripts/upgrade.sh:413-442`) brings up the one-shot `migrate` service and blocks on its exit code.
- Step 7 (`compose up -d`) is what actually replaces the running `api` and `worker` containers.

So between the migration committing and the new app starting, the PREVIOUS release is still live and still accepting writes. Any row the old code writes in that window is written with the old semantics, after the migration that was supposed to correct them has already run. Alembic will not run the revision again, so nothing repairs those rows.

The window is short (however long `compose up -d` takes to swap containers) but it is not zero, and an instance under real traffic can take an ingest in it.

## Why it is a class, not a one-off

Every data-only backfill in `backend/alembic/versions/` has the same exposure. A non-exhaustive list: 0013 (geoparquet distribution backfill), 0028 (oauth email_verified backfill), 0034 (geometry linearization), 0036 (dataset source state), 0048 (vector-tile protocol relabel). Each one assumes it sees the final state of the old data, and each is one-shot.

0048 works around it in the application layer: `generate_distributions` now repairs a surviving stale row when it next runs for that record. That is a per-migration patch, it only covers records that are later refreshed or reuploaded, and it is not something every future data migration should have to reinvent.

## Options

1. Stop the app before migrating and start it after: correct for every migration at once, but it converts a rolling-ish upgrade into a real outage window and needs a decision about that tradeoff.
2. Put the API into a read-only or maintenance mode for the migrate step, so reads keep serving and writes are refused.
3. Accept the window and require every data migration to ship an application-side repair, which is roughly what we do today by accident rather than by policy.

Worth deciding deliberately rather than per-PR. Option 1 is the honest default for a self-hosted single-instance product; option 2 is the one that keeps the demo and small deployments from looking down during an upgrade.

ishiland:
Closes #1467.

## What #1467 asked, and what was actually there

The issue describes a window where migrations run while the previous release is still serving writes. Checking the script first: that window does not exist. `scripts/upgrade.sh` has stopped `api` and `worker` before the dump since f1183a645, and nothing restarts them until the final `compose up -d` (the two intermediate `up` calls both pass `--no-deps`). The migrate step already runs with no application writer alive.

What was missing is that the property was incidental. The stop landed as a pg_dump consistency fix, no comment claimed the migration-safety guarantee, and no test pinned it, so any future edit could have reintroduced the window silently. Two adjacent things were genuinely broken.

## The outage included the download

The release-file sync and `compose pull` ran *after* the stop, so every operator paid the image download as downtime. On a self-hosted link that is usually the largest part of an upgrade, and it has no reason to be inside the window. Both now run before the stop. The outage is stop, dump, migrate, start.

New order: PG-major probe, sync, pull, **stop api+worker**, dump, verify, db config bounce, migrate, pin `.env`, **up -d**.

## A failed migration left the instance down

The migrate-failure path printed the rollback recipe and exited with `api` and `worker` still stopped. The instance stayed down until someone noticed.

It now restores the previous release first. Two details are load-bearing:

- `--no-deps` on the restore. The prod compose gives `api` a `depends_on: migrate: service_completed_successfully` edge, so a plain `up -d api worker` re-runs the one-shot that just failed and the app stays down.
- The restore lives in the `EXIT` trap, gated on a flag, so it covers every abort while the app is deliberately stopped, not only the paths someone wrote a message for.

The gate closes once migrations commit. Starting old containers on a new schema is not a rollback, so a health-gate failure keeps printing the recipe and leaves the new version running.

## The version pin moved too early

`update_env_value GEOLENS_VERSION` ran before the migrate step. After a failed migration `.env` named a version whose migrations had not run, so re-running the script compared the target against that pin, decided `same`, printed "Already on X, nothing to upgrade" and exited 0 with the app still stopped and the schema unmigrated.

The pin now moves after migrations succeed. Compose reads the environment ahead of `.env`, so the exported `GEOLENS_VERSION` is all the sync, pull and migrate steps need. A re-run after a failure now actually retries (covered by a new test).

## Docs

`RUNBOOK.md` gets section 10: what stops and for how long, what keeps serving (`db`, `frontend` and `titiler` stay up, but `titiler` is only reachable through `api`, so raster tiles stop with it), a per-failure table of end states, and the manual recovery for when the script's own shell dies mid-run.

`UPGRADING.md`'s step list described the old order and never mentioned that the upgrade is an outage. Both manual recipes are updated to match the script. The source-build recipe now stops `api` and `worker` **before** `git checkout`: `docker-compose.yml` bind-mounts `./backend/app` into the api container, so checking out the new tag under a running app swaps its code immediately. That is the same defect #1467 describes, in the hand-driven path.

## Verification

- `sh -n scripts/upgrade.sh` clean. `shellcheck -x -s sh scripts/upgrade.sh` (via `koalaman/shellcheck:stable`) reports zero findings, same as the base revision.
- `scripts/tests/test-upgrade-order.sh`: 70/70 (was 55). New assertions cover pull-before-stop, no writer restart between the stop and the app up, the pin moving only on success, the migrate-failure restore, the pin surviving a failed migrate, a re-run after a failed migrate actually retrying, and no restore after migrations commit. The stub logs `restore_app` only for `--no-deps ... api worker`, so a restore that dropped `--no-deps` falls through to `app_up` and fails the failure-path assertions.
- `scripts/tests/test-upgrading-doc-consistency.sh` 15/15, `scripts/check_env_doc_drift.py` OK, plus the other four `installer-test` job scripts and `pre-commit` on the changed files.
- Independent PATH shim logging full argv, to confirm ordering and the restore's image tag:

```
SUCCESS                                              MIGRATE FAILURE (exit 1)
compose exec -T db psql ... server_version_num       compose exec -T db psql ...
compose exec -T db cat /etc/postgresql/custom.conf   compose exec -T db cat ...
compose pull --ignore-buildable                      compose pull --ignore-buildable
compose stop api worker         <- outage starts     compose stop api worker
compose exec -T db pg_dump ...                       compose exec -T db pg_dump ...
compose exec -T db pg_restore -f /dev/null           compose exec -T db pg_restore -f /dev/null
compose up -d --no-deps migrate                      compose up -d --no-deps migrate
compose ps -aq migrate ; docker wait                 compose ps -aq migrate ; docker wait
compose up -d                   <- outage ends       compose logs --tail 30 migrate
compose ps --format ...                              compose up -d --no-deps api worker
.env => GEOLENS_VERSION=1.2.4                        .env => GEOLENS_VERSION=1.2.3
```

The restore call carries `GEOLENS_VERSION=1.2.3` in its environment, so it starts the previous release's images. No real stack was touched.

## Scope notes

Beyond `scripts/upgrade.sh` and `RUNBOOK.md` I had to touch two more in-repo files, both made wrong by the change: `scripts/tests/test-upgrade-order.sh` pins the exact call order in CI's `installer-test` job, and `UPGRADING.md` documents the script's steps and the hand-driven equivalents.

Not edited, for follow-up:

- The public installer repo, which syncs `scripts/upgrade.sh` at release time.
- `getgeolens.com` `docs/.../upgrade.mdx`, the user-facing upgrade guide. UPGRADING.md already carries a maintainer note flagging it as a cross-repo follow-up, and the doc-consistency test enforces that note.

One behavior change worth calling out for review: because the release-file sync now runs before the stop, `compose stop api worker` executes against the newly synced compose file rather than the installed one. Compose resolves running containers by project and service label, so this is fine unless a release renames or removes those services, in which case the stop fails and the script aborts before touching the database. That failure is safe, and arguably the right outcome.

### DIFF
--- a/RUNBOOK.md
+++ b/RUNBOOK.md
@@ -2200,7 +2200,7 @@ otherwise the largest part of an upgrade.
 | Release-file sync or image pull | Nothing stopped, nothing migrated, `.env` unchanged. Fix and re-run. |
 | Cannot stop `api`/`worker` | Aborts before the dump instead of dumping under live writers, and restarts whatever it managed to stop. Database untouched. |
 | Dump fails, is empty, or does not read back | Discards the unusable dump, restarts `api`/`worker` on the previous version, aborts. Database untouched. |
-| Migrations fail | Restarts `api`/`worker` on the previous version, prints the rollback recipe, exits non-zero. |
+| Migrations fail | Restarts `api`/`worker` on the previous version, waits to see whether they stay up, prints the rollback recipe, exits non-zero. |
 | New version starts but never becomes healthy | Leaves the new version running and prints the rollback recipe. It does not put the old containers back. |
 
 Two consequences are worth knowing before you meet them.
@@ -2212,6 +2212,12 @@ old code, but old code on a partly-migrated schema is not a state to settle
 into. Read `docker compose logs migrate`, then either fix the cause and re-run
 the upgrade, or restore the pre-upgrade dump the script just took. Its path is
 in the rollback recipe the script prints.
+
+That restart can also fail to hold, and the script says so when it does rather
+than reporting a recovery it did not get. The api image runs
+`alembic upgrade heads` on boot, so if the failed migration committed a revision
+the previous image has never seen, that image refuses to start and the restart
+policy loops it. When you see that message, the dump is the way back.
 
 **The `GEOLENS_VERSION` pin in `.env` moves only after the migrations succeed.**
 A failed upgrade leaves the file naming the version you are still running, so
@@ -2246,9 +2252,9 @@ docker compose -f docker-compose.prod.yml logs --tail 50 migrate
 
 Then take the case that matches.
 
-**`api`/`worker` are stopped and migrate never ran, or exited non-zero.**
-Nothing irreversible happened. Start the previous release again and re-run the
-upgrade when you are ready:
+**`api`/`worker` are stopped and migrate never ran.** Nothing irreversible
+happened. Start the previous release again and re-run the upgrade when you are
+ready:
 
 ```bash
 docker compose -f docker-compose.prod.yml up -d --no-deps api worker
@@ -2257,6 +2263,29 @@ docker compose -f docker-compose.prod.yml up -d --no-deps api worker
 `--no-deps` is not optional here. `api` declares a dependency on the `migrate`
 one-shot completing successfully, so a plain `up -d` re-runs the migration you
 may be trying to stay clear of.
+
+**`api`/`worker` are stopped and migrate exited non-zero.** Do not assume the
+database is untouched. Alembic runs the batch in a transaction, but a revision
+that does its own commits (anything creating an index `CONCURRENTLY`, for
+instance) can leave both schema changes and an advanced `alembic_version` behind
+when a later revision fails. Read the migrate logs and decide which way to go:
+
+- If the cause is fixable (a full disk, a lock timeout), fix it and re-run
+  `scripts/upgrade.sh`. `alembic upgrade heads` is idempotent, so it resumes.
+- Otherwise restore the pre-upgrade dump, below. Starting the previous release
+  is not a safe substitute: the api image runs `alembic upgrade heads` on boot,
+  so if the failed migration committed a revision the old image does not know,
+  it refuses to start and the restart policy loops it. A stopped instance is the
+  visible symptom; old code silently reading a half-migrated schema is the worse
+  one.
+
+```bash
+docker compose -f docker-compose.prod.yml exec -T db \
+  psql -U "$POSTGRES_USER" -d "$POSTGRES_DB" -c 'SELECT * FROM alembic_version'
+```
+
+If that revision is not in the release you were running, the dump is the only
+way back.
 
 **`api`/`worker` are stopped and migrate exited 0.** The schema is already on
 the new release, so finish the upgrade rather than reverting it. Edit the
--- a/RUNBOOK.md
+++ b/RUNBOOK.md
@@ -22,6 +22,7 @@ documentation.
 7. [Schema rollback](#7-schema-rollback)
 8. [Audit log retention](#8-audit-log-retention)
 9. [Uploaded source-file retention](#9-uploaded-source-file-retention)
+10. [Routine version upgrade: outage and rollback](#10-routine-version-upgrade-outage-and-rollback)
 
 ---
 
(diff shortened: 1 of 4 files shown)


ishiland:
@codex review

ishiland:
Fixed in 60b354890. `APP_DOWN=0` now runs as the first statement after the migration succeeds (`scripts/upgrade.sh:501-512`), before the `update_env_value` write, so a failed pin can no longer reach the trap with the restore armed.

Added a regression test for it (CASE 2d in `scripts/tests/test-upgrade-order.sh`). The docker stub drops write permission on the install directory during `docker wait`, which is the last call before the pin, so `update_env_value`'s `mv .env.tmp.$$ .env` genuinely fails rather than being simulated. Moving `APP_DOWN=0` back to its old position makes that case fail with `restore_app` at the end of the call log, which was the exact behaviour you described.

The app is left stopped in that path, which the printed rollback recipe covers. Keeping `update_env_value` as a bare call is deliberate: wrapping it in an `if` would suspend `set -e` inside the function, and a disk-full `awk > $tmp` would then fall through to the `mv` and replace `.env` with a truncated file.

ishiland:
@codex review

ishiland:
Both fixed in a48105a0f. Round-2 review budget for this PR is spent, so these fixes are unreviewed.

**P1 (restore not verified).** Confirmed the mechanism: `backend/scripts/api-entrypoint.sh:110-126` runs `alembic upgrade heads` whenever `GEOLENS_API_RUN_MIGRATIONS` is not `false`, and `docker-compose.prod.yml:293` leaves it defaulting to `true`, so the previous image refuses to start against a revision its own graph does not carry. Added `previous_app_settled()` (`scripts/upgrade.sh:379-400`): it polls the api container's `.State.Status` for up to 30s and fails on `restarting`/`exited`/`dead`, using the same `docker inspect` primitive `wait_for_healthy` already uses. `report_restore` now only claims a restore when that holds; otherwise it names the boot-time-alembic cause and points at the dump restore, which `print_rollback` prints immediately after.

I did not use `compose up --wait`: it would need `--wait-timeout` to avoid hanging on a crash loop, and that flag needs Compose 2.19+, while this script only asserts Compose v2.

Regression test is CASE 2c in `scripts/tests/test-upgrade-order.sh`. The stub answers `.State.Status` per container, so `cid-api` can report `restarting` while the migrate one-shot stays `exited`. It asserts the restore is still attempted, that the output does not say "api + worker are running again", and that it points at the dump.

**P2 (RUNBOOK).** You are right that it contradicted the partial-migration warning three paragraphs above it. Split into two cases: "migrate never ran" keeps restart-and-retry; "migrate exited non-zero" now says not to assume the database is untouched, explains that a revision doing its own commits (a `CONCURRENTLY` index, say) can leave both schema changes and an advanced `alembic_version` behind, gives the `SELECT * FROM alembic_version` check, and makes the choice between retrying forward and restoring the dump explicit. The failure table row and the UPGRADING.md summary lost their "never leaves the instance down" overclaim to match.

Gates after the change: order test 76/76, doc-consistency 15/15, env-doc drift OK, `shellcheck -x -s sh` clean.

PR #1477, opened 2026-08-13
feat(catalog): persist and display dataset attribution

### LINKED ISSUE #1472: feat(catalog): dataset attribution is stored at ingest but never surfaced in the viewer
Found while wiring `attribution` into the showcase seeder's manifest ingests (#1471): the field is accepted and stored, but nothing ever displays it.

**The path today**: `ManifestMetadata.attribution` is validated on `POST /api/ingest/manifest/apply` and lands in `job_metadata["manifest_attribution"]`. From there it goes nowhere: no dataset or record column carries it forward, and no map-layer response schema has an attribution field, so neither the dataset page nor the map viewer can show it.

**Why it matters**: attribution is a license obligation for several data sources a deployment is likely to ingest (swisstopo's terms require visible credit for swissALTI3D; NOAA asks for credit on ETOPO). Right now an operator can dutifully supply the credit line at ingest and the product silently drops it.

**Suggested shape**: promote attribution to a real record/dataset field (PATCHable like `license`), include it in the map-layer read model, and render it in the viewer the way web maps conventionally do (a compact attribution control line; MapLibre has native `AttributionControl` support that currently only carries basemap credits). Backfill from `job_metadata["manifest_attribution"]` where it exists is optional but cheap.

Related: the seeder currently smuggles attribution text into `license`/`summary` prose, which is exactly the workaround this feature would retire.

ishiland:
Closes #1472

## What

Dataset attribution was accepted at manifest ingest and then thrown away. `ManifestMetadata.attribution` was validated on `POST /api/ingest/manifest/apply` and written to `ingest_jobs.user_metadata["manifest_attribution"]`, where it stopped: no record column carried it forward, no read model exposed it, and nothing rendered it. An operator who dutifully supplied the credit line a source's terms require got no credit displayed anywhere.

This carries it end to end.

**Backend.** New `catalog.records.attribution` column (`Text`, nullable), mirroring how `license` is shaped. It is deliberately its own column rather than a reuse of `license` or `source_organization` — `license` names the terms, `source_organization` is a search facet whose values get grouped and counted, and attribution is verbatim prose that has to render unchanged. The dataset PATCH (`DatasetMeta`) accepts and clears it, `DatasetResponse` returns it on both the detail and list endpoints, and both map-layer read models carry it as `dataset_attribution` — `MapLayerResponse` for the builder and the authenticated viewer, `SharedLayerResponse` for share links and embeds.

The ingest tail reads the job ledger back through a new `apply_manifest_record_metadata` in `tasks_common.py`, called from the vector tail, the raster tail, and the reupload swap — so a manifest vector, a manifest raster, a manifest-driven VRT, and a manifest *re-apply* all land the same value. It copies only the manifest-namespaced keys; `title`/`summary`/`visibility` continue to reach the record through `create_dataset`'s own arguments on every ingest path, so the helper is a no-op for uploads, service ingests, and STAC imports. An absent key means "unchanged" rather than "clear", matching what the dataset PATCH does with the same field.

**Frontend.** Every surface that renders the data now credits it, through the mechanism that surface actually has:

- The **viewer** (share links, embeds, and the read-only `/maps/:id` route) collects the distinct non-empty credits of its currently visible layers and passes them to MapLibre's `AttributionControl` as `customAttribution`, deduped and in stacking order, alongside the basemap's own credit. Scoped to the live eye-toggle state rather than the saved `visible` flag, so hiding a layer drops its credit and showing it restores it.
- The **builder** and the **dataset preview** have no explicit `AttributionControl` to hand `customAttribution` to, and react-maplibre reads the `<Map attributionControl>` option once at mount. Both credit through MapLibre's native source-level `attribution` instead, which the default control recomputes from whichever sources are live. Every source kind either can create is covered — vector tiles, both GeoJSON shapes (a small 3D dataset and a bounded client-side cluster layer), raster, raster-dem, and the terrain mesh. The terrain source earns it in its own right: a terrain-mode DEM has no visible layer, so nothing references the attributed source its adapter built, and MapLibre counts the terrain one through `usedForTerrain` instead.
- **Exported `style.json`** carries it on every dataset source. That export is a published artifact rendered outside this instance, which is where a source's display obligation binds most.

**Attribution is treated as text, not markup.** MapLibre renders attribution as HTML — `AttributionControl` assigns the joined string to `innerHTML` — and its sanitizer only removes `<script>` elements, `on*` handlers, and `javascript:`/`data:` URLs. ``, `<iframe src>`, and inline `style` all survive it, so an editor-supplied credit line would otherwise be stored markup rendered on every public, shared, and embedded map: an anonymous-facing surface a dataset editor cannot otherwise reach. Both write paths (the dataset PATCH and the manifest schema) reject `<` and `>` through `core.text.reject_html_markup`, the CLI's packaged manifest schema carries the same rule so `geolens manifest validate` agrees with the API, and every render boundary escapes as well so a value predating the guard stays inert. Rejecting angle brackets rather than allowlisting elements leaves nothing to get wrong, and ampersands, quotes, and accents — which appear in real organization names — are untouched.

The dataset metadata editor gets an editable attribution field directly beside the license read-out in the Governance card — same `InlineEdit` + `EditableFieldShell` + capability pattern as its neighbours, so an operator can add a credit line a source started requiring without re-ingesting.

Two implementation notes worth flagging for review. The `key` on `<AttributionControl>` is load-bearing, not cosmetic: react-maplibre's `useControl` builds the control once with `useMemo(..., [])`, so a changed `customAttribution` prop never reaches MapLibre, and keying on the credit set remounts it exactly when that set changes. And `toViewerSyncInput` deliberately does *not* set the source-level field — the viewer owns an explicit control, and feeding both would be two mechanisms answering one question, with MapLibre's dedupe hiding any disagreement between them.

## Schema, API, and config impact

- **Schema:** migration `0049_records_attribution` adds one nullable `Text` column to `catalog.records`. No index (the column is only ever read through a record row the query already joins, never filtered or grouped on), no default, no data migration. Downgrade drops the column.
- **API:** four schemas gain a field — `DatasetMeta.attribution` (request), `DatasetResponse.attribution`, `MapLayerResponse.dataset_attribution`, `SharedLayerResponse.dataset_attribution`. All optional/nullable additions, so no client breaks. Exported `style.json` sources gain a standard MapLibre `attribution` key. `backend/openapi.json`, both SDKs, and `frontend/src/types/api.generated.ts` are regenerated and committed; the hand-typed mirrors in `frontend/src/types/api.ts` are updated to match.
- **Config:** none.

`DatasetMeta.attribution` is bounded at 5000 rather than the 1000 its neighbours use, because `ManifestMetadata.attribution` is already `NonEmptyString5000` and the ingest tail writes it straight to the column. A 1000 bound would have accepted a manifest value that the dataset PATCH then refused to round-trip.

## Deliberately not in scope

- **No historical backfill.** Existing `user_metadata["manifest_attribution"]` values are not copied into the new column. The job ledger is per-ingest and a dataset may have been reuploaded or edited since, so a blanket copy could resurrect a credit line for a dataset whose data no longer comes from that source. Operators who want the old values can PATCH them.
- **No DCAT/STAC serializer changes.** Neither profile has a field where attribution maps in one line the way `license` does (STAC expresses this through `providers`), so the mapping is a design question rather than a field rename and does not belong here.
- **The showcase seeder still smuggles attribution into `license`/`summary` prose.** Adopting the new field there is a separate PR.

Two known limitations, both raised in review and both argued in the thread rather than fixed here:

- The **dataset preview's** source is created once, so an attribution edit on that page does not reach the map above it until reload. That is a pre-existing property of `useMapLayers` shared with `tileVersion` (which is why `#1362` busts the raster cache through the tile URL instead), and changing it means a source recreate on a hook with other callers. Every published surface is correct immediately.
- **Rendered map images** — PNG export, gallery thumbnail, and OG image — carry no attribution. None of them draws the *basemap's* credit either (the thumbnail and OG paths are a raw canvas crop with no compositing step at all), so this is a pre-existing hole rather than a regression. Filed as #1486 with the design questions it raises: assembling basemap plus dataset credits into one line, placement and truncation at three very different sizes, a scrim for legibility over arbitrary imagery, and i18n.

## Verification

Backend (host, test Postgres at `:5434`):

```
uv run python -m alembic upgrade head        # applies
uv run python -m alembic check               # no new upgrade operations detected
uv run python -m alembic downgrade -1        # column dropped
uv run python -m alembic upgrade head        # column restored
uv run pytest -n 4 -q                        # 8848 passed, 91 skipped (FULL suite)
uv run ruff check . && uv run ruff format --check .   # clean
```

Contract artifacts:

```
make openapi-check              # no drift
make sdks-check                 # exit 0, tree clean
make cli-test                   # 8 passed, 2 skipped
make manifest-contract-check    # CLI half 74 passed; backend half 101 passed
```

Frontend:

```
npm run lint          # clean
npm run typecheck     # clean
npm run test:i18n     # 4 passed
npm run build         # built
npx vitest run        # 4944 passed, 397 files (FULL suite)
```

New tests: `backend/tests/test_dataset_attribution.py` (PATCH set/clear/omit/bound, dataset detail and list, both map-layer read models, the null case, and the manifest ledger read-back driven from `manifest_job_metadata`'s real output rather than a hand-written dict, including the re-apply replace/keep semantics), four cases appended to `backend/tests/test_maps_style_json.py` (all three exported source types plus the no-credit case), `frontend/src/components/viewer/__tests__/ViewerMap.attribution.test.tsx` (collection ordering, dedupe, toggle, blank handling, remount-key injectivity, and what reaches `AttributionControl`), `frontend/src/components/builder/__tests__/map-sync.attribution.test.ts` (vector, raster, and raster-dem source specs), and `frontend/src/components/maps/hooks/__tests__/use-map-layers.attribution.test.ts` (preview sources, including that the absent-argument case leaves other callers unchanged).

Seven `_MODULE_LOC_CAPS` entries and two decomposed-service budgets are raised across the commits, each with a comment recording what the lines bought.

Codex review ran seven rounds and ended on a clean pass matched to this head. Eleven findings: nine fixed, one answered with evidence at the cited lines (the preview's create-time source, a property it shares with `tileVersion`), and one split out as #1486.

No Playwright: this branch is a worktree, and the dev stack bind-mounts the main checkout, so an e2e run here would exercise `main` rather than this code.


ishiland:
@codex review

ishiland:
Round 1 — all three findings were real and are fixed in 5758ae9.

**Attribution remount keys collide (ViewerMap.tsx:1078).** Correct, and the failure mode is exactly as described. The derivation now lives in `attributionControlKey` in `frontend/src/components/viewer/layer-identity.ts`, serializing with `JSON.stringify`. Pulling it out of the JSX was deliberate: the key's injectivity is the property that makes `customAttribution` live at all, so it should be assertable directly rather than inferred from a rendered control. `ViewerMap.attribution.test.tsx` now pins `['A|B']` against `['A', 'B']`, order sensitivity, and `[]` against `['']`.

**Builder mapper drops the field (api.ts:1274).** Correct. `toSyncInput` now maps `dataset_attribution` into `SyncLayerInput.attribution` — the MVT-05 field that was declared and never fed. I went one step past the finding: `map-sync.ts` only spreads `attribution` into the *vector* source spec, so mapping it alone would have left raster and DEM layers as the one builder layer kind still rendering uncredited. `AdapterLayerInput.attribution` plus the `addSource` spreads in `raster-adapter.ts:174` and `hillshade-adapter.ts:136` close that. `map-sync.attribution.test.ts` asserts the credit reaches all three source types.

`toViewerSyncInput` deliberately does *not* set the field. The viewer owns an explicit `AttributionControl` and passes `customAttribution`; feeding the source-level property too would be two mechanisms answering one question, and MapLibre's dedupe would hide any disagreement between them.

**Preview map omits the credit (api.ts:246).** Correct, and you were right to quote the schema description back — "displayed wherever the data is rendered" was my wording, and the preview was rendering it. `useMapLayers` takes an optional `attribution` and applies it to both the vector and raster source specs; `DatasetMap` accepts it as a prop and `DatasetPage.tsx:584` passes `dataset.attribution`. Source-level rather than a control prop for the same reason as the builder: the preview has no explicit control, and react-maplibre reads the `<Map attributionControl>` option once at mount. Covered by `use-map-layers.attribution.test.ts`, including that the absent-argument case leaves every other caller's source spec unchanged.

Verification on 5758ae9: `npm run typecheck`, `npm run lint`, `npm run test:i18n`, and `npm run build` all clean; `npx vitest run src/components/maps src/components/viewer src/components/builder src/components/dataset src/pages/__tests__` → 2917 passed across 222 files.

ishiland:
@codex review

ishiland:
Round 2 — two fixed in fd9191d, two answered. That closes my review budget, so I am not re-triggering.

## Fixed

**P1, apply attribution during manifest reuploads.** Correct, and the severity is right. Confirmed the whole chain: `_classify_dataset` returns `"update"` (manifest_service.py:513) → `_create_reupload_job_and_queue` builds `{**manifest_job_metadata(...), "reupload": True, ...}` (manifest_service.py:648), so the job really does carry the *new* `manifest_attribution` → and `tasks_reupload.py` touched no `record.*` field anywhere, so it was dropped. `reupload_file` now calls `apply_manifest_record_metadata(dataset.record, user_metadata)` immediately after `_apply_reupload_swap`, in the swap's own transaction. `dataset.record` is safe to touch there — the SELECT that loads it uses `joinedload(Dataset.record)` (tasks_reupload.py:338-342), so no lazy load runs inside the task.

Only the file path, not the service one: `_validate_existing_dataset_update` (manifest_service.py:547) rejects raster and any non-`vector_dataset`/`table` record, so a manifest update cannot reach `reupload_service`.

Worth naming the semantics this settles, since it is a choice rather than a consequence: an *absent* `attribution` key leaves the existing credit alone rather than clearing it, matching what the dataset PATCH does with the same field. A manifest that never mentions attribution should not silently drop one. Both cases are pinned in `test_dataset_attribution.py` — `test_a_reapply_replaces_a_previous_credit` and `test_a_reapply_that_omits_attribution_keeps_the_existing_credit`.

**P2, include dataset attribution in exported style sources.** Correct, and this was the strongest of the three P2s: an exported style is a published artifact rendered outside this instance entirely, which is precisely where a source's display obligation binds. `_source_for_layer` now sets `source["attribution"]` on its **common tail**, next to the existing `source["metadata"]` assignment, rather than inside the three branches — so vector, raster, and raster-dem all carry it and a fourth source type cannot be added without it. Four tests in `test_maps_style_json.py` cover the three types plus the null/empty/whitespace case leaving the exported source byte-identical to before.

## Not fixed, with reasoning

**P2, refresh preview sources after attribution edits.** The mechanism you describe is real — `addVectorLayers` returns at the `vectorLayersAdded.current` guard, and the effect at use-map-layers.ts:217-222 re-enters it and hits that same guard. I am leaving it, for three reasons.

It is not specific to attribution. Every prop in that callback's dep list is applied at source-create time only, `tileVersion` included — which is exactly why `#1362 codex r2` busts the raster cache through the tile *URL* rather than by re-adding the source, and why `DatasetPage.tsx:572` passes `key={isRasterOrVrt ? mapKey : undefined}` to force a remount for rasters. My `attribution` behaves the same as its neighbour, so this is a pre-existing property of the hook, not a deviation I introduced.

MapLibre has no setter for a live source's attribution, so the fix is remove-and-re-add — which drops tiles and flickers the preview — or mounting a live control, on a hook with other callers. That is a real change to shared code.

And the blast radius is the author's own in-page preview lagging until reload. Every *published* surface is correct immediately: the viewer, share links, embeds, and now the exported style all read the field per render or per request. I would rather leave this as a known, documented limitation than reshape a shared hook inside this PR.

**P2, draw required credits into PNG exports.** Also real, and also not something this PR introduced: the export footer draws only `export.poweredBy` when `showBranding` is set (use-builder-save.ts:968-975). It draws no **basemap** attribution either — OpenStreetMap, CARTO, and every other basemap credit is equally absent from every PNG the builder has ever exported. So this is a pre-existing hole that my change makes more visible rather than a regression, and closing it properly means composing the whole attribution line (basemap plus datasets, with wrapping and its own i18n key) into the export canvas. That is its own change with its own layout decisions, and I would rather file it than bolt a dataset-only credit line onto a footer that still omits the basemap's.

## Verification on fd9191d

```
uv run python -m alembic upgrade head / check / downgrade -1 / upgrade head   # clean both ways, no drift
uv run pytest tests/test_dataset_attribution.py tests/test_maps_style_json.py \
  tests/test_layering.py -q                                                   # 160 passed
uv run pytest tests/test_reupload*.py tests/test_manifest_apply_*.py \
  tests/test_maps.py -q                                                       # 326 passed
uv run ruff check . && uv run ruff format --check .                           # clean
make openapi-check                                                            # no drift
make sdks-check                                                               # exit 0, tree clean
npm run typecheck && npm run lint && npm run test:i18n                        # clean
```

`tasks_reupload.py` and `style_json.py` both sat exactly on their `_MODULE_LOC_CAPS` entries; both are raised in the same commit with a comment recording what the lines bought.

ishiland:
@codex review

ishiland:
Round 3's finding is fixed in 9ad17f2, and enumerating the sibling set turned up a third gap next to the two named.

**The finding (both GeoJSON sources).** Confirmed and fixed. The cause was one line in the wrong scope: `const attribution = ...` was declared *inside* the vector-tile block, which neither GeoJSON shape reaches — a 3D dataset at or under the GeoJSON-Z threshold (`is_3d && feature_count <= 5000`), and a bounded client-side cluster layer. Hoisted above the `useGeoJsonSource` branch so all three source kinds `ensureVectorSource` can create read one definition, which is what stops a fourth shape being added uncredited.

**Third sibling, not in the report: the terrain source.** `ensureRasterDemTerrainSource` (map-sync.ts:100) builds a `raster-dem` source and had no attribution. It matters specifically because a terrain-mode DEM has **no visible layer** — `isDemTerrainVisualSuppressed` (map-sync.ts:64) suppresses it — so nothing references the attributed raster-dem source its adapter created, and MapLibre's `used` flag for that source is false (`_updateAttributions` only counts `used || usedForTerrain`). The terrain source is a different id and *is* counted, through `usedForTerrain`, which makes it the only place a terrain-only DEM's credit can come from on the builder. swissALTI3D — the example in #1472 — is exactly that shape. A changed credit now also replaces the source, which is cheap here because it carries a tile template and no data.

**Your known (a), the raster path: already covered.** `raster-adapter.ts:174` and `hillshade-adapter.ts:136` both got `...(attribution ? { attribution } : {})` in round 1, fed from `AdapterLayerInput.attribution`. Verified at the lines, not assumed.

### Full enumeration

Eleven `map.addSource` sites in `frontend/src`, every one classified:

| Site | Source | Verdict |
|---|---|---|
| `map-sync.ts:1050` | vector tiles | credited (round 1) |
| `map-sync.ts:1009` / `:1012` | GeoJSON, clustered + plain | **credited now** |
| `map-sync.ts:133` | terrain `raster-dem` | **credited now** |
| `raster-adapter.ts:167` | raster | credited (round 1) |
| `hillshade-adapter.ts:129` | `raster-dem` | credited (round 1) |
| `use-map-layers.ts:66` / `:245` | preview vector + raster | credited (round 2) |
| `use-map-layers.ts:271` | `drawn-overlay` | out of scope — user-drawn geometry, seeded `EMPTY_FC` |
| `MeasurementPlugin.tsx:135` | `MEASURE_SOURCE` | out of scope — measurement UI |
| `use-ephemeral-layers.ts:82` | analysis result | argued in a comment, see below |

### Deliberately uncredited, argued in place

**Your known (b), the `setData` branch.** Agreed, and it is now a comment at the branch rather than silence. MapLibre reads a GeoJSON source's `attribution` at construction and exposes no setter, so an attribution edited mid-session does not reach a source already on the map. Remove-and-re-add would drop the rendered features and refetch, for a credit that is correct again on the next builder load — and every *published* surface (viewer, share, embed, exported style) re-reads the field per render or per request and is never stale. I took the bound rather than engineering live-update, as you suggested.

**The ephemeral analysis source.** Commented at `use-ephemeral-layers.ts:82`. It renders the *output* of a spatial operation, not a dataset: the geometry is computed, the result is transient (cleared on the next run, never persisted or exported), and the hook receives only the result GeoJSON with no handle on which inputs produced it. Crediting derived analysis output is a real question that belongs with the analysis feature that knows its inputs.

**The viewer's terrain path.** Passes no attribution, and that is not an oversight. It credits through `customAttribution` off its layer list, and terrain is gated on `demLayerVisible`, which is the same live-visibility set `collectLayerAttributions` filters on — so whenever viewer terrain renders, the DEM's credit is already collected. The one case that list misses is a share-token embed whose DEM row was filtered out (`use-viewer-terrain.ts:117-119` still seeds terrain from the raster token), and there the layer object does not exist, so there is no credit string to read.

### Verification on 9ad17f2

```
npx vitest run          # 4932 passed, 396 files (FULL suite, not a selection)
npm run typecheck       # clean
npm run lint            # clean
npm run test:i18n       # 4 passed
npm run build           # built
```

CI is green on this head, all checks. One existing test needed the new key: `BuilderMap.a11y.test.tsx` asserts the exact terrain options object, so it gains `attribution: null` — the option is always passed and the source spec omits the key when there is nothing to say.

@codex review

ishiland:
Round 4's P1 is fixed in 98b6f0b. Confirmed at the source before fixing, because the finding's severity rests entirely on what MapLibre's sanitizer actually permits.

**Verified.** `AttributionControl._updateAttributions` ends in `this._innerContainer.innerHTML = DOM.sanitize(joined)`, and `sanitize` is: parse as HTML, remove `<script>` elements, then `removeAttributes` drops only what `isPossiblyDangerous` matches — `src`/`href`/`xlink:href` containing `javascript:` or `data:`, plus anything starting with `on`. Everything else stands. So ``, `<iframe src="https://…">`, and `style="position:fixed;inset:0"` all survive it. The report is right, and right about the severity: the writer is an authenticated dataset editor, the audience is anonymous viewers of a public, shared, or embedded map, and markup on that surface is not otherwise reachable by that role.

**Two layers, because they close different holes.**

*Write paths reject markup.* `core.text.reject_html_markup` is a Pydantic validator on both writers — `DatasetMeta.attribution` (the PATCH) and `ManifestMetadata.attribution` (manifest apply). It lives in `core/text.py`, which already hosts the `normalize_nfc` validator this same field uses, so the rule has one definition across the catalog and processing layers. Guarding the manifest schema rather than only the ingest tail matters: a manifest with markup now fails apply with a field-level 422, instead of being accepted and having its credit silently dropped at commit.

I rejected `<` and `>` rather than allowlisting elements. A credit line is prose, so there is no allowlist to get wrong, and the characters that actually appear in organization names — ampersands, apostrophes, quotes, accents — are untouched. `test_ordinary_credits_still_round_trip` pins that with "Rand & McNally", a quoted NOAA string, and an accented IGN string.

*Render boundaries escape anyway.* One helper, `lib/attribution-safety.ts`, at all four MapLibre entry points — the viewer's `customAttribution` via `collectLayerAttributions`, the builder's source attribution via `toSyncInput`, the terrain source in `BuilderMap`, and the preview's two sources in `useMapLayers` — plus `html.escape` on the exported `style.json`, whose `attribution` is an HTML string the consuming application renders in a context we do not control. That covers a value written before the guard existed or straight to the database.

With the write guard in place the escaping only ever touches an ampersand, which is the correct rendering for an HTML target regardless. Escaping order is `&` first, pinned by a test — otherwise a stored `&lt;script&gt;` would decode back into a live tag.

The viewer dedupes *after* escaping, so one credit stays one entry no matter which form it arrived in.

### Verification on 98b6f0b

```
uv run pytest -n 4 -q          # 8848 passed, 91 skipped (FULL backend suite)
npx vitest run                 # 4944 passed, 397 files (FULL frontend suite)
npm run typecheck / lint       # clean
make openapi-check             # no drift — a field_validator is not a schema constraint
make sdks-check                # exit 0, tree clean
```

CI is green on this head, every check. Two `_MODULE_LOC_CAPS` entries raised with comments.

@codex review

ishiland:
Real, verified at the cited lines, and now filed as **#1486** rather than fixed here. This is the same class as round 2's PNG-export finding, restated at a second call site, so I have converted the refusal into a tracked issue instead of arguing it a third time in prose.

**Verified.** `doCapture` (use-builder-save.ts:76-110) does no compositing at all — both targets are a straight `cropResize(map.getCanvas(), …)` into a 400×250 and a 1200×630 JPEG. That is a stronger version of the point than the report makes: it is not that the DOM control is missing from an otherwise-composed image, it is that there is no composition step in this path whatsoever. `handleExportPNG` at least composites title, legend, and a branding footer.

**Why it is not this PR.** The same crop omits the **basemap's** attribution — OpenStreetMap, CARTO, whatever the operator configured — and always has, for every image the product has ever produced, in every deployment, independent of whether any dataset carries a credit. So the gap is not one #1477 introduced; it is one this feature makes more visible by adding a second thing that ought to appear there.

That matters for the fix, not just for the blame: shipping only the dataset half would leave the images non-compliant while implying they were fixed. The real change assembles basemap credit plus distinct visible-layer credits into one line, decides placement and truncation at three very different sizes (a 400×250 thumbnail has little room for a legible credit), needs a scrim so the text survives on arbitrary map imagery, needs i18n keys for any separator or affix, and needs `doCapture` to grow the offscreen-composite step `handleExportPNG` already has. Each of those is a decision, not a mechanical port of the DOM path.

#1486 records all of it, including the suggestion that the composed line come from the same `collectLayerAttributions` the viewer uses, so the image and the on-screen control cannot drift, and the open question of whether a thumbnail too small for a legible credit should carry one at all.

Nothing changed on this head — 98b6f0b stands, CI green.

@codex review

ishiland:
Correct, and it is drift I introduced one round earlier — fixed in 99b74de.

Adding `reject_html_markup` to `ManifestMetadata` in round 4 made the backend stricter than the packaged CLI schema, which is a hand-maintained mirror rather than a generated artifact. The consequence is the one worth naming: `geolens manifest validate` and the preflight inside `geolens manifest apply` both called such a document valid, so the user learned about the rejection as a 422 from the API — after staging and upload, from the command whose entire purpose is to answer that question beforehand.

`metadata.attribution` in `geolens-manifest-v1.schema.json` now carries `"pattern": "^[^<>]*$"` plus a description that names the backend model it mirrors, so the next person editing either side sees the pairing.

The character class is deliberate rather than incidental: a `.`-based expression would not match a newline, and a credit line spanning two lines is legitimate. `test_manifest_attribution_accepts_ordinary_credit_lines` covers that case alongside ampersands, quotes, and accents, and `test_manifest_attribution_rejects_markup` covers the four injection shapes plus a bare stray `>`.

I checked `license` for the same drift while I was there and left it alone: it is displayed through React, which escapes, and never reaches an HTML render context — so the restriction would be unjustified there.

### Verification on 99b74de

```
make manifest-contract-check   # CLI half 74 passed; backend half run separately, 101 passed
make cli-test                  # 8 passed, 2 skipped
```

CI is green on this head, every check.

@codex review

PR #1478, opened 2026-08-13
fix(ogc): serve HEAD wherever the standards preflight advertises it

### LINKED ISSUE #1470: fix(ogc): HEAD /api/collections returns 405 while the CORS preflight advertises HEAD
On a v1.12.0 instance:

```
$ curl -sI https://<instance>/api/collections
HTTP/2 405
allow: GET
```

while the CORS preflight on the same route advertises HEAD as allowed:

```
access-control-allow-methods: GET, HEAD, POST, OPTIONS
```

So a browser client that preflights a HEAD (or any client that trusts the preflight) is told HEAD is fine and then receives 405. FastAPI emits HEAD handlers only when routes opt in; the OGC routers in `backend/app/standards/ogc/router.py` declare GET only, while the CORS middleware advertises a static method list.

Low severity, but the two surfaces should agree: either register HEAD on the public standards routes (FastAPI supports `methods=["GET", "HEAD"]`, and HEAD-for-GET is cheap and spec-friendly on OGC endpoints that clients probe), or stop advertising HEAD in the preflight response.

ishiland:
Closes #1470

## Problem

```
$ curl -sI https://<instance>/api/collections
HTTP/2 405
allow: GET
```

while the CORS preflight on the same path answers `access-control-allow-methods: GET, HEAD, POST, OPTIONS`. A browser client that preflights a HEAD is told it is allowed, then refused.

The mismatch is not an oversight in the CORS layer — that layer means it. `DynamicCORSMiddleware._is_anonymous_standards_request` (`backend/app/api/middleware/cors.py:130`) explicitly admits `HEAD` into the anonymous standards allowlist, and `_set_public_standards_cors_headers` advertises it. What is missing is on the route side: FastAPI's `APIRoute` does not add HEAD alongside GET the way starlette's plain `Route` does. That is why `/openapi.json` and `/docs` (plain Routes) answer HEAD and every standards route does not. All 48 GET routes on that surface were GET-only.

## Fix

Register HEAD by deriving it, in one pass after the routers are included, from the same `standards_api_path` classifier the CORS middleware already uses to decide what to advertise. That is the point of doing it this way: the advertised set and the served set now come from one function and can no longer disagree. Editing ~48 decorators across five routers would fix today's mismatch and leave the next route free to reintroduce it.

There is already precedent for this shape — `_add_trailing_slash_aliases` is the same kind of derived-route pass in the same file, added for the same reason (a mechanical property that should hold across a whole surface, not per-decorator).

Two details worth calling out:

- **Registering a route, not mutating `route.methods`.** fastapi 0.140 keeps included-router routes nested and matches through `RouteContext`, whose `.methods` is a *copy* of the route's (verified: `ctx.methods is ctx.route.methods` -> `False`). An in-place `route.methods.add("HEAD")` would leave the matcher answering 405 while the route object claimed otherwise.
- **`include_in_schema=False`.** A HEAD operation documents nothing its GET does not, and publishing 48 of them would churn `openapi.json` and every generated SDK. Confirmed: no OpenAPI drift.

The ~25-kwarg route copy both passes need is factored into `_clone_api_route`; `_add_trailing_slash_aliases` previously spelled that list inline and now shares it. A kwarg dropped from that list is a silent behaviour difference between a canonical route and its copy, so it is worth having exactly one copy of it.

### Why register rather than narrow the advertisement

The issue offered both. Narrowing is the wrong direction here: the middleware's own allowlist already treats HEAD as a first-class standards method, HEAD-probing a landing page or collection before fetching it is ordinary OGC client behaviour, and the same static header also advertises POST (true only for `/stac/search`), so "make the header exact" is a larger change than "make the routes match" and leaves clients worse off.

## Verification

```
uv run pytest tests/test_ogc_discovery.py tests/test_ogc_features.py tests/test_ogc_pagination.py \
  tests/test_ogc_public_access.py tests/test_ogc_queryables.py tests/test_ogc_records_conformance.py \
  tests/test_ogc_collection_metadata.py tests/test_stac*.py tests/test_dcat.py tests/test_geodcat_ap.py \
  tests/test_middleware.py tests/test_redirect_slashes.py tests/test_layering.py -q
# 515 passed

PYTHONPATH=. uv run python scripts/dump_openapi.py --check   # no drift
```

Route-table check after the pass: 48 standards GET paths, 48 standards HEAD paths, zero GET-without-HEAD, and zero HEAD routes outside the standards surface.

New tests in `tests/test_ogc_discovery.py`:
- `test_head_is_served_wherever_the_preflight_advertises_it` — the preflight advertises HEAD, and HEAD on `/collections`, `/`, `/conformance`, `/stac`, `/datasets/dcat/` returns 200 with the same content-type as GET and an empty body.
- `test_head_stays_off_non_standards_routes` — `HEAD /datasets/` still 405s, so the pass is scoped by the classifier rather than applied app-wide.

## Impact

No schema or API-contract change; no OpenAPI drift, no SDK regeneration. `_MODULE_LOC_CAPS` for `api/main.py` goes 1389 -> 1466 (the new pass plus the shared route-copy helper), recorded with the reason in the same commit.

---

## Review round 1: the `Allow` header

Review pointed out that HEAD answering is only half of it — a 405 for some *other* method still reported `Allow: GET`, which understates what the surface serves.

The suggested "register GET and HEAD on the same effective route" is not available in FastAPI. `get_openapi_path` derives a route's operation id from its name and path rather than its method, so one route carrying `{GET, HEAD}` produces a duplicate operation id per route plus 48 phantom `head:` operations. Measured both ways with `scripts/dump_openapi.py --check`:

| approach | duplicate operation ids | snapshot |
| --- | --- | --- |
| separate HEAD route (this PR) | 0 | clean |
| `{GET, HEAD}` on canonical route | 5 | out of date |

That is the same reason FastAPI does not auto-add HEAD the way starlette's plain `Route` does.

So the header is restated instead, in `backend/app/standards/ogc/errors.py` where the standards error contract already lives. Two things surfaced while doing it:

- `fastapi/routing.py:96` imports `HTTPException` from `starlette.exceptions` and raises **that** for the 405. The existing problem-detail handler is bound to fastapi's subclass, so it never saw this exception — the standards 405 was being served by fastapi's default JSON handler all along.
- Handler lookup walks `type(exc).__mro__` and prefers the most specific registration, so binding the base class beside the subclass leaves router-raised errors on their problem+json contract. Pinned by `test_standards_router_errors_keep_problem_details`.

The new handler only rewrites `Allow` and delegates everything else to fastapi's default. It deliberately does **not** convert framework 405/404 bodies to problem+json — that is a real pre-existing inconsistency, but a separate contract change.

Keyed off GET, since HEAD is registered for every standards GET: a standards route allowing only POST gains nothing, and `PUT /datasets/` still reports `Allow: GET`.

Added `test_standards_405_allow_header_reports_head` and `test_standards_router_errors_keep_problem_details`. Full re-run: 550 passed across the OGC/STAC/DCAT/middleware/layering suites, `dump_openapi.py --check` clean.

### DIFF
--- a/backend/app/api/main.py
+++ b/backend/app/api/main.py
@@ -956,6 +956,54 @@ def _iter_api_routes(target_app: FastAPI) -> list:
     ]
 
 
+def _clone_api_route(
+    target_app: FastAPI,
+    route,
+    *,
+    path: str,
+    methods: list[str],
+    name_suffix: str,
+) -> None:
+    """Re-register an existing APIRoute at *path* for *methods*.
+
+    Both derived-route passes below register a copy of a canonical route
+    rather than editing it, so every attribute a handler's behaviour depends
+    on has to be carried across. Spelled once here: a kwarg dropped from this
+    list is a silent behaviour difference between the canonical route and its
+    copy, and that is not a difference either caller wants.
+
+    Hidden from OpenAPI. A derived route documents nothing the canonical one
+    does not, and publishing it would churn every generated SDK.
+    """
+    target_app.add_api_route(
+        path=path,
+        endpoint=route.endpoint,
+        response_model=route.response_model,
+        status_code=route.status_code,
+        tags=route.tags,
+        dependencies=route.dependencies,
+        summary=route.summary,
+        description=route.description,
+        response_description=route.response_description,
+        responses=route.responses,
+        deprecated=route.deprecated,
+        methods=methods,
+        operation_id=None,  # MUST differ from canonical for uniqueness;
+        # FastAPI auto-generates when None.
+        response_model_include=route.response_model_include,
+        response_model_exclude=route.response_model_exclude,
+        response_model_by_alias=route.response_model_by_alias,
+        response_model_exclude_unset=route.response_model_exclude_unset,
+        response_model_exclude_defaults=route.response_model_exclude_defaults,
+        response_model_exclude_none=route.response_model_exclude_none,
+        include_in_schema=False,
+        response_class=route.response_class,
+        name=f"{route.name}__{name_suffix}" if route.name else None,
+        openapi_extra=route.openapi_extra,
+        generate_unique_id_function=route.generate_unique_id_function,
+    )
+
+
 def _add_trailing_slash_aliases(target_app: FastAPI) -> None:
     """ROUTE-01 (Phase 1092 review CR-01): register a hidden no-slash alias
     for every trailing-slash route in the app.
@@ -1013,33 +1061,12 @@ def _add_trailing_slash_aliases(target_app: FastAPI) -> None:
             # Register the alias. Inherit response_model, dependencies,
             # status_code, etc. from the canonical route — APIRoute
             # exposes these directly.
-            target_app.add_api_route(
+            _clone_api_route(
+                target_app,
+                route,
                 path=no_slash,
-                endpoint=route.endpoint,
-                response_model=route.response_model,
-                status_code=route.status_code,
-                tags=route.tags,
-                dependencies=route.dependencies,
-                summary=route.summary,
-                description=route.description,
-                response_description=route.response_description,
-                responses=route.responses,
-                deprecated=route.deprecated,
                 methods=[method],
-                operation_id=None,  # MUST differ from canonical for
-                # uniqueness; FastAPI auto-generates
-                # when None.
-                response_model_include=route.response_model_include,
-                response_model_exclude=route.response_model_exclude,
-                response_model_by_alias=route.response_model_by_alias,
-                response_model_exclude_unset=route.response_model_exclude_unset,
-                response_model_exclude_defaults=route.response_model_exclude_defaults,
-                response_model_exclude_none=route.response_model_exclude_none,
-                include_in_schema=False,  # ROUTE-01: hide aliases from OpenAPI
-                response_class=route.response_class,
-                name=f"{route.name}__no_slash_alias" if route.name else None,
-                openapi_extra=route.openapi_extra,
-                generate_unique_id_function=route.generate_unique_id_function,
+                name_suffix="no_slash_alias",
             )
             existing_paths.add((method, no_slash))
             added += 1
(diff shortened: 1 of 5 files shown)


ishiland:
@codex review

ishiland:
Fixed in b7b8ec9 — standards 405s now report `Allow: GET, HEAD`. You were right that the header understated the surface.

I did not take the "register GET and HEAD on the same effective route" half of the suggestion, because it is not available in FastAPI without corrupting the schema. `get_openapi_path` derives a route's operation id from its name and path, not its method, so one route carrying `{GET, HEAD}` produces a duplicate operation id per route plus 48 phantom `head:` operations. Measured both ways with `scripts/dump_openapi.py --check`:

- separate HEAD route (this PR): 0 duplicate-operation-id warnings, no snapshot drift
- combined `{GET, HEAD}` on the canonical route: 5 duplicate-operation-id warnings, snapshot out of date

That is also why FastAPI does not auto-add HEAD the way starlette's plain `Route` does, and why `/openapi.json` and `/docs` (plain Routes) already answer HEAD while every APIRoute did not.

So I took the other half — aggregate the methods at the 405 — in `backend/app/standards/ogc/errors.py`. Two things surfaced while doing it that are worth recording:

1. `fastapi/routing.py:96` imports `HTTPException` from `starlette.exceptions`, and line 1268 raises **that** for the 405. The existing problem-detail handler is bound to fastapi's subclass, so it never saw this exception at all — the standards 405 was being served by fastapi's default JSON handler.
2. Handler lookup walks `type(exc).__mro__` and prefers the most specific registration, so binding the base class beside the subclass leaves router-raised errors on the problem+json contract. Pinned by `test_standards_router_errors_keep_problem_details`.

The new handler rewrites `Allow` and delegates everything else to fastapi's default. It deliberately does not convert framework 405/404 bodies to problem+json — that is a real pre-existing inconsistency, but a separate contract change rather than something to slip into this PR.

Scoping is keyed off GET (`_allow_header_with_head`), since HEAD is registered for every standards GET: a standards route allowing only POST gains nothing, and `PUT /datasets/` still reports `Allow: GET`.

Verification: 550 passed across the OGC/STAC/DCAT/middleware/layering suites; `dump_openapi.py --check` clean.

@codex review

Issue #1480, opened 2026-08-13
test(e2e): a worktree run silently tests `main`, so results are meaningless in both directions


ishiland:
## Problem

The dev stack bind-mounts the **main** checkout (`./frontend` → `/app`, `backend/app` → `/app/app`), so `localhost:8080` serves `main` no matter which branch a worktree is on. Running `npx playwright test` from a worktree therefore exercises code the author did not write.

Both directions are wrong, and the quiet one is worse:

1. False failure. The worktree's fix is absent from the running stack, so a passing change reports red.
2. False pass. A worktree change that breaks e2e goes green, because the stack never had the code. Nothing distinguishes this from a real pass.

`AGENTS.md` documents the trap and the workaround: run Vite on `:5174` with `API_PROXY_TARGET=http://localhost:8001` and point `E2E_BASE_URL` at it. That documentation is currently the only thing preventing a meaningless test result, so anyone using `git worktree` who has not read that section hits the trap blind.

One case is legitimately safe and must keep working. A spec-only change (editing `e2e/*.spec.ts` with no app-code change) is validly testable from a worktree, because Playwright reads specs from the worktree while app code stays byte-identical to `main`.

## Proposed fix

A guard in the Playwright config that fails loudly when the worktree and the stack under test disagree.

## Open question: how to detect the mismatch

Three candidates, with different reliability and different false-positive costs.

1. Path comparison. Fail when `E2E_BASE_URL` is the default `:8080` and `git rev-parse --show-toplevel` is not the main checkout. Cheap and dependency free, but it cannot see the spec-only exemption, so it would block a legitimate run.
2. Git rev comparison. Compare the worktree's `HEAD` against a build marker the API exposes. Precise, and it handles the exemption correctly if the marker reflects the mounted source, but it needs a marker to exist and be trustworthy.
3. Dirty-diff scope. Allow the run when the worktree's diff against `main` touches only `e2e/`, block otherwise. This encodes the exemption directly and needs no server support, but it is a heuristic and gets confused by stacked branches.

Worth settling before someone implements it. Option 2 is the most correct if a suitable marker exists. Option 3 is the cheapest thing that still respects the exemption.

## Acceptance

- Running e2e from a worktree against the main stack, with app-code changes, fails with a message naming the actual problem and the `:5174` workaround.
- A spec-only worktree change still runs.
- Running from the main checkout is unaffected.

ishiland:
## Correcting two things in the description

[...]

PR #1481, opened 2026-08-14
chore(make): add env-test target for the gitignored host test env

ishiland:
## What

`make env-test` creates the gitignored `.env.test` from `.env.test.example`. That is the whole change.

## Why

`.env.test` is gitignored, so a fresh clone starts without it and so does every new worktree. The host pytest recipe documented in `AGENTS.md` then dies on missing env vars while booting `app.core.config`, before collecting a single test. The non-zero exit reads exactly like a gate failure, which is the part that costs time.

`AGENTS.md` already said to fix it with `cp .env.test.example .env.test`. Replacing that instruction with a target matters most in worktrees, where `AGENTS.md` separately tells you to copy `.env.test` in from outside the checkout. `make env-test` does it without reaching across checkouts.

## Scope reduced after review

The first revision also made `ai-evals` depend on `env-test` so its recipe could source `.env.test` unconditionally, on the theory that the `if [ -f ... ]` guard was hiding a silent-wrong path. Codex showed that backwards, and it was right:

- `.env.test.example` ships `POSTGRES_PASSWORD=geolens`.
- `make dev-init` generates a *random* `POSTGRES_PASSWORD` into `.env`, specifically because `geolens` is the known-public default (SEC-010, `scripts/install.sh`).
- Sourcing the template under `set -a` exports the public default over the real password, so the evals can no longer authenticate against the dev DB on a default install.
- With no `.env.test` present, `Settings` loads `.env` directly (`backend/app/core/config.py`) and the credentials match.

The guard was preserving a working path. `ai-evals` is now restored byte-for-byte against `main`, with a comment recording why it must stay that way so the next person does not repeat the same "cleanup".

## Verification

- `make env-test` in a clean worktree creates the file.
- A second run reports the file exists and leaves it untouched, confirmed by appending a sentinel line between runs and checking it survived.
- `git check-ignore` confirms `.env.test` still matches `.gitignore:3`, so it cannot be committed by accident.
- The `ai-evals` recipe is diffed byte-for-byte against `origin/main`: identical.
- pre-commit passed.

Not run: the backend suite and every docker target. Another session was mid `pytest -n 4` against the shared DB, and starting a second run would have collided with its cluster-global DDL.

## Related

#1480 covers the worktree e2e trap, a larger version of the same class of problem.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -25,7 +25,7 @@ The React/Vite frontend is in `frontend/src/`: `components/`, `pages/`, `hooks/`
 - Frontend: `cd frontend && npx vitest run src/path/foo.test.ts`.
 - E2E: `npx playwright test e2e/foo.spec.ts --project=chromium` (stack must be running).
 
-A focused selection is blind to the module-size gates. `backend/tests/test_layering.py` caps the size of the largest backend modules, and CI runs it on every PR that triggers `backend-test`, so a change that adds lines to a ratcheted file passes locally and fails there. If you touched anything under `backend/app/`, finish with `cd backend && set -a && source ../.env.test && set +a && uv run pytest tests/test_layering.py -q`; it needs no database, but it does boot `app.core.config`, so a bare run dies on missing env vars with a non-zero exit before collecting anything — which reads exactly like a gate failure and is not one. In a fresh clone `.env.test` does not exist yet (it is gitignored); create it once with `cp .env.test.example .env.test` from the repo root. Growth is allowed — raise the file's cap in `_MODULE_LOC_CAPS` in the same commit, with a comment saying what the lines bought.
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

Your P2 on `Makefile:107` was correct and is fixed in c9f7a4cdc.

Verified all three legs before acting: `.env.test.example:31` ships `POSTGRES_PASSWORD=geolens`, `scripts/install.sh:383` generates a random one into `.env` under SEC-010, and `backend/app/core/config.py:1062` points Settings at the root `.env`. Sourcing the template under `set -a` would have exported the public default over the real password on any `make dev-init` install.

The `if [ -f ../.env.test ]` guard was preserving a working path rather than masking a bug, which is the opposite of what I assumed when I removed it. `ai-evals` is restored byte-for-byte against main and now carries a comment explaining why the guard must stay.

Scope is now just the `env-test` target plus the two AGENTS.md pointers. Please re-review the reduced diff.

Issue #822, opened 2026-07-28
ci: Rule-1 pre-commit hook is file-scoped and misses non-get_dataset fetch paths


ishiland:
From the 1.6.0 pre-tag audit — flagged independently by three review passes.

The Rule-1 grep hook (`.pre-commit-config.yaml`) matches `@*router.<verb>` handlers that call `get_dataset(` and greps the same file for an access/visibility check. Blind spots:

- Fetch paths that don't go through `get_dataset(`: `db.get(Dataset, ...)`, raw selects, service-layer loads.
- File scope: a check anywhere in the file satisfies the grep even if a specific handler lacks one.
- `processing/` is excluded.

Replace with a structural check: a pytest that walks the FastAPI route table and asserts every handler touching `Record`/`Dataset`/`Map`/`RecordEmbedding` calls one of the sanctioned guards. A `get_dataset_or_404(...)` helper would help — the load-then-404-then-check sequence is duplicated at ~47 call sites, and centralizing it makes the guard the default path rather than a convention.

PR #1482, opened 2026-08-14
docs(agents): unfile the docs policy and make the security rules skimmable

ishiland:
## What

Two structural fixes to `AGENTS.md`. No content is removed.

1. The root repository docs policy is now its own `## Repository Docs Policy` section. It was filed under `## Security & Configuration Tips`, where it has nothing to do with security.
2. Rule 2 and the Enforcement block are broken from single paragraphs into a rule statement followed by a list.

## Why

`AGENTS.md` loads into every agent session and is the first thing a new contributor reads. The content is accurate, but two paragraphs had grown to 1,003 and 1,319 characters and interleaved three different things: the rule to follow, the reasoning behind it, and the issue history explaining how it got that way. A rule you cannot skim is a rule that gets skipped.

The docs policy being under a security heading is a plain misfiling that makes it hard to find.

## One deliberate change of emphasis

"Never re-add `GDAL_HTTP_FOLLOWLOCATION`" was the last sentence of the 1,003-character Rule 2 paragraph. It now leads that paragraph in bold, with a note that no structural test catches it.

That is not cosmetic. `test_rule2_structural.py` requires GDAL argv to go through the safe-env helpers, so adding the variable *inside* `gdal_safe_env()` passes the suite clean. The prose is the only guard against this specific mistake, and it was buried.

## Verification

- Every backticked identifier in the previous file is still present. Checked by `comm` over the sorted sets, empty diff.
- Every issue and requirement reference still present: #435, #822, #936, #937, #1234, REL-01, BKP-04, INST-01, plus the 2026-08-03 CodeQL policy date.
- Longest prose block in the security section: 1,003 to 366 characters.
- Long prose blocks across the file: 5 to 3. Both removed are in the security section.
- The file is 139 characters longer overall. This is a readability change, not a size reduction.
- pre-commit passed.

## Not in scope

The 904-character paragraph in "Running a single test" is the remaining worst offender. It is untouched here because #1481 edits that same region, and splitting the work keeps both diffs clean. Worth a follow-up once that merges.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -81,13 +81,30 @@ Pull requests should describe the change, call out schema/API/config impacts, li
 
 Brand assets (logos, color tokens, font references, brand-usage rules, press materials) live in the sibling [`geolens-io/branding`](https://github.com/geolens-io/branding) repository — not here. When an app feature needs a logo, palette token, or identity element, copy from a tagged branding release rather than re-authoring locally. The propagation order for any change that touches brand identity is **branding → this repo → marketing → docs**. Cross-surface brand canon lives in branding's `BRAND-GUIDE.md`.
 
+## Repository Docs Policy
+
+Keep root repository docs single-purpose:
+
+- `README.md` is the public overview.
+- `SUPPORT.md` is support routing.
+- `CHANGELOG.md` is the release-note source of truth.
+- `EDITIONS.md` is the open-core/commercial boundary. Sanctioned at the root because licensing transparency requires it in-repo (REL-01).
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review

Docs-only change to `AGENTS.md`. Two things worth checking specifically:

1. That no content was dropped. I verified every backticked identifier and every issue/requirement reference survives (#435, #822, #936, #937, #1234, REL-01, BKP-04, INST-01), but a second pass on meaning rather than tokens would be useful.
2. That moving the root-docs policy out from under `## Security & Configuration Tips` into its own `## Repository Docs Policy` section does not orphan any cross-reference elsewhere in the repo that pointed at the old heading.

PR #1483, opened 2026-08-14
docs(agents): break the module-size-gate paragraph into steps

ishiland:
## What

The paragraph under `### Running a single test` becomes a lead sentence, a fenced command block, and four bullets. No content changes.

## Why

It packed nine separate facts into 887 characters:

1. A focused selection is blind to the module-size gates.
2. `backend/tests/test_layering.py` caps the largest backend modules.
3. CI runs it on every PR that triggers `backend-test`.
4. So a change adding lines to a ratcheted file passes locally and fails there.
5. The command to run if you touched `backend/app/`.
6. It needs no database.
7. It boots `app.core.config`, so a bare run dies on missing env vars before collecting anything, which mimics a gate failure.
8. A fresh clone has no `.env.test` (now `make env-test`, added in #1481).
9. Raising a cap in `_MODULE_LOC_CAPS` belongs in the same commit with a comment.

Facts 5 through 9 are things you act on, and they were buried mid-sentence. The command in particular was inline backticks inside a longer clause, which is awkward to copy.

## Verification

- The command text is byte-identical, just moved from inline backticks to a fenced `bash` block. Confirmed with an exact-string grep.
- Every other backticked identifier in the file is unchanged. Confirmed with `comm` over the sorted sets; the only delta is the command itself moving into the fence.
- Code fences balanced (4, two pairs).
- Longest prose block in the file: 887 to 659 characters. Long non-list prose blocks: 3 to 2.
- Diff is one file, 12 insertions, 1 deletion.
- pre-commit passed.

## Related

Follows #1482, which did the same for the security section. Completes the pass I deferred there to avoid tangling with #1481.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -25,7 +25,18 @@ The React/Vite frontend is in `frontend/src/`: `components/`, `pages/`, `hooks/`
 - Frontend: `cd frontend && npx vitest run src/path/foo.test.ts`.
 - E2E: `npx playwright test e2e/foo.spec.ts --project=chromium` (stack must be running).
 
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review

Docs-only reformat of one paragraph in `AGENTS.md`. Specifically worth checking:

1. That the fenced command is byte-identical to the inline one it replaces. I verified with an exact-string grep, but a second pass is welcome.
2. That splitting the paragraph did not drop or reorder any of the nine facts it carried, particularly the `app.core.config` boot-failure caveat, which is the one that actually costs people time.
3. That the `bash` fence language tag is consistent with how other fenced blocks in this repo's docs are tagged.

PR #1487, opened 2026-08-14
fix(demo): keep MNMAP_PLUTO out of the prune delete set

ishiland:
The 08-14 `--prune-userdata` dry run listed `MNMAP_PLUTO` for deletion: it is visitor-uploaded, so the owner-based classifier reads it as foreign, and it was never in `PINNED_DATASET_TITLES` because the seeder did not create it. The geolens-examples MCP transcripts quote it, so deleting it breaks a published walkthrough the same way deleting a UUID-pinned dataset breaks an example page.

One line plus its comment: pin the title. `--prune-userdata` hard-keeps pinned titles regardless of owner, so `--execute` becomes safe to run.

Verification: `ruff check` clean, syntax-checked; the prune re-dry-run on the demo after merge is the acceptance check (MNMAP_PLUTO must move from the delete set to "externally pinned, hard-kept").

### DIFF
--- a/scripts/seed-showcase.py
+++ b/scripts/seed-showcase.py
@@ -513,12 +513,21 @@ PINNED_DATASET_TITLES = (
     "NYC Subway Lines (MTA)",  # title derives data.nyc_subway_lines_mta - NEVER rename
     "NYC Subway Stations (MTA)",
     "swissALTI3D Matterhorn DEM (2m mosaic)",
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1489, opened 2026-08-14
fix(audit): serialize audit details through JSON mode so date fields cannot 500 the mutation

### LINKED ISSUE #1484: PATCH dataset metadata with a vintage date 500s: audit details carry datetime.date into a stdlib-JSON column
## Repro (found live on the demo, 1.13.0, during the showcase ops run)

As a dataset owner:

    PATCH /api/datasets/{id}
    {"data_vintage_start": "1950-01-01"}

Response: 500 after a long hang (see companion logging issue). The metadata update itself is lost.

## Chain

1. Pydantic parses `data_vintage_start` into a real `datetime.date` (`DatasetMeta`, schemas.py).
2. The audit emit in the PATCH handler passes `details=meta.model_dump(exclude_unset=True)` — python-mode dump, so the `date` object survives into the details dict (`backend/app/modules/catalog/datasets/api/router.py:380`).
3. The audit table's JSON column serializes with stdlib `json.dumps` — no engine-level `json_serializer` is configured — so the INSERT raises `TypeError: Object of type date is not JSON serializable`, surfaced as `StatementError` at `_emit_insert_statements`.
4. The exception rolls back the whole request transaction, taking the already-flushed record UPDATE with it: license/keywords/theme in the same PATCH are lost too.

Verified by field bisection on the live demo: `license`- and `keywords`-only PATCHes on the same dataset return 200 in ~35ms; `data_vintage_start` reproduces the failure every time. The rendered StatementError traceback in the api log names the json encoder frames explicitly.

## Impact

Any user setting temporal coverage through the metadata editor loses the whole edit with a 500. On single-worker deployments the companion logging bug turns each occurrence into a multi-minute API outage.

## Fix direction

`model_dump(mode="json", exclude_unset=True)` at the call site. Sweep the sibling `details=body.model_dump(exclude_none=True)` at `backend/app/modules/catalog/collections/router.py:256` and any other `details=` dict that can carry non-JSON scalars (dates, datetimes, UUIDs) — this is a class, not an instance. An engine-level `json_serializer` with `default=str` would also mask the class but silently stringifies everything; the explicit per-site `mode="json"` is the honest fix.

ishiland:
## Problem

`PATCH /api/datasets/{id}` with `{"data_vintage_start": "1950-01-01"}` returned 500 and silently discarded every other field sent in the same body. Reported in #1484, found live on the demo at 1.13.0.

## Chain

1. Pydantic parses `data_vintage_start` into a real `datetime.date` (`DatasetMeta`).
2. The PATCH handler built its audit payload with a python-mode `model_dump`, so the `date` object survived into the details dict.
3. `audit_logs.details` is JSONB and the engine configures no `json_serializer`, so SQLAlchemy encodes it with stdlib `json.dumps` — which cannot encode a `date`.
4. The `TypeError` surfaces at **flush** time, not at `audit_emit` time. `audit_emit` isolates sink failures per AUDIT-03, but the sink only calls `session.add()`; the INSERT is emitted later inside the caller's commit, outside that `try/except`. The exception therefore rolls back the whole request transaction and takes the already-staged record UPDATE with it — license, keywords, and theme sent in the same PATCH are lost too.

## Sweep

`details` is typed `dict | None` and nothing checks JSON-safety at the boundary, so this is a class, not an instance. I walked the AST of every module under `backend/app/` and enumerated all 126 `details=` keyword arguments (97 passed directly to `AuditEvent`/`log_action`, 29 through forwarding helpers). Disposition:

| Shape | Count | Disposition |
|---|---|---|
| `meta.model_dump(exclude_unset=True)` — `datasets/api/router.py` | 1 | **Fixed.** `DatasetMeta.data_vintage_start`/`_end` are `date`. This is the live 500. |
| `body.model_dump(exclude_none=True)` — `collections/router.py` | 1 | **Fixed, preventive.** `CollectionUpdate` is `name`/`description`, both `str \| None`, so this site could not 500 today. Changed so a future field addition cannot make it. |
| Literal dicts of `str`/`int`/`bool` | 89 | No change needed. UUIDs are already `str(...)`-converted and datetimes already `.isoformat()`-converted at the call site (e.g. the audit-export `filters` dict, `router_operations.py`'s `expires_at`). |
| `details` bound to a literal dict in the same function (`admin/router.py`, `audit/router.py`, `records/router.py`, `refresh/service.py`, `jobs/router.py`) | 12 | No change needed. Traced each to its assignment; all are literal dicts of scalars. |
| `{**spread, ...}` from an internal result dict | 6 | No change needed. Traced each source: `StaleJobSweepOutcome.as_dict()` is annotated `dict[str, int]`, the rest are counts and fixed strings. |
| Reads out of the column, not writes into it (`AuditLogResponse(details=log.details)`) | 4 | Out of scope — these render a value the DB already returned as JSON. |
| `MapEditHistoryEvent` via `record_map_history_event` | 13 | Already safe. `service_history.py:47` runs its payload through `jsonable_encoder`, so the sibling history table never had this bug. |

`Visibility` is a `Literal` of strings, not an `Enum`, so no enum member reaches a details payload.

## Why per-site `mode="json"` and not an engine-level serializer

An engine-level `json_serializer=partial(json.dumps, default=str)` would also make the 500 go away, and it is the wrong fix: it silently stringifies *every* type in *every* JSON column, so a future mistake gets a plausible-looking `"2026-08-14 00:00:00+00:00"` in the audit trail instead of a loud failure. Per-site `mode="json"` produces the ISO-8601 form Pydantic already uses everywhere else in the API, and leaves unexpected types loud.

## Preventing the class

`tests/test_audit_details_json_1484.py::test_audit_details_payloads_never_use_a_python_mode_model_dump` walks every `details=` keyword argument under `backend/app/` and fails any whose value resolves to a `model_dump` without `mode="json"`. It resolves three shapes lexically within one enclosing function: inline, spread into a literal (`{**body.model_dump(), ...}`), and assigned-then-passed. `mode=<variable>` is not provable and is reported rather than assumed.

It covers `details=` on *any* callee, not just `AuditEvent`/`log_action`, because the forwarding helpers (`_propagate_record_write`, `record_map_history_event`) take the same kwarg into the same columns.

Known limit, stated rather than implied: a dump reaching a `details=` site with no lexical trace in the same function is not recognizable to any AST rule and the guard does not claim it. A companion test (`test_structural_guard_detects_a_python_mode_dump`) calls the real resolver against a synthetic sample, so gutting the detection logic cannot leave a permanently green assertion-of-absence.

## Verification

Both directions checked, not just the green one. Against `origin/main`'s version of the two files the guard reports exactly `router.py:380` and `collections/router.py:256` and nothing else. With the fix reverted in place, the regression test fails with the same error the issue's live traceback shows:

```
sqlalchemy.exc.StatementError: (builtins.TypeError) Object of type date is not JSON serializable
```

Commands run from `backend/` with `.env.test` sourced:

```
uv run pytest tests/test_audit_details_json_1484.py -q                      # 5 passed
uv run pytest tests/test_collections.py tests/test_metadata_roundtrip.py \
              tests/test_audit.py -q                                        # 60 passed (incl. the new file)
uv run pytest tests/test_datasets.py tests/test_dataset_history_redaction.py \
              tests/test_audit_sink.py tests/test_audit_column_ddl_feed.py \
              tests/test_dataset_attribution.py -q                          # 114 passed
uv run pytest tests/test_layering.py -q                                     # 47 passed
uv run ruff check . && uv run ruff format --check .                         # clean
```

The null-clear test pins that `mode="json"` did not disturb `exclude_unset` semantics — an explicit `null` still appears in the details payload, which the history feed depends on (#458 E-04).

## Out of scope, worth a follow-up

AUDIT-03 says audit sinks must not break callers, and `audit_emit` implements that with a `try/except` around each sink. That guarantee is only half-real: the sink stages an ORM object, so any error in *writing* it lands at the caller's flush, outside the isolation. This PR removes the known way to trigger it, but the structural gap — an audit-row defect can still destroy a user's mutation — remains. Worth deciding deliberately whether audit failures should fail closed (reject the mutation) or be dropped with a loud log, rather than inheriting whichever one flush ordering produces.

Fixes #1484

### DIFF
--- a/backend/app/modules/catalog/collections/router.py
+++ b/backend/app/modules/catalog/collections/router.py
@@ -253,7 +253,9 @@ async def update_collection_endpoint(
             action="collection.update",
             resource_type="collection",
             resource_id=collection_id,
-            details=body.model_dump(exclude_none=True),
+            # fix(#1484): mode="json" so a future non-JSON scalar on
+            # CollectionUpdate cannot reach the JSONB column as a raw object.
+            details=body.model_dump(mode="json", exclude_none=True),
             ip_address=request.client.host if request.client else None,
         ),
     )
--- a/backend/app/modules/catalog/datasets/api/router.py
+++ b/backend/app/modules/catalog/datasets/api/router.py
@@ -377,7 +377,11 @@ async def update_dataset_metadata(
             resource_id=dataset_id,
             # exclude_unset, not exclude_none: an explicit null clear
             # (#458 E-04) must appear in the audit/history details.
-            details=meta.model_dump(exclude_unset=True),
+            # fix(#1484): mode="json" — details lands in a JSONB column that
+            # serializes with stdlib json.dumps, and data_vintage_start/end are
+            # real date objects. A python-mode dump raised at flush time and
+            # rolled back the record UPDATE staged above it.
+            details=meta.model_dump(mode="json", exclude_unset=True),
             ip_address=request.client.host if request.client else None,
         ),
     )
--- a/backend/tests/test_audit_details_json_1484.py
+++ b/backend/tests/test_audit_details_json_1484.py
@@ -1,0 +1,286 @@
+"""Audit ``details`` payloads must be JSON-safe before they reach JSONB (#1484).
+
+``audit_logs.details`` is a JSONB column and the engine configures no
+``json_serializer``, so SQLAlchemy serializes it with stdlib ``json.dumps``.
+A python-mode ``model_dump()`` leaves ``date``/``datetime``/``UUID``/``Decimal``
+objects intact, and stdlib json cannot encode any of them.
+
+Why that is worse than a broken audit row: the ``TypeError`` surfaces at FLUSH
+time, not at ``audit_emit()`` time. ``audit_emit`` isolates sink failures
+(AUDIT-03), but the sink only calls ``session.add()`` — the INSERT is emitted
+later, inside the caller's commit, outside that try/except. So the failure
+rolls back the whole request transaction and takes the user's already-staged
+mutation with it. On the live demo a
+``PATCH /datasets/{id} {"data_vintage_start": "1950-01-01"}`` 500'd and silently
+discarded the license/keyword edits sent in the same body.
+
+Two regression tests pin the observable behaviour, and one structural test
+pins the class so a new call site cannot reintroduce it.
+"""
+
+import ast
+import pathlib
+import uuid
+from datetime import date
+
+import pytest
+from httpx import AsyncClient
+from sqlalchemy import select
+
+from app.modules.audit.models import AuditLog
+from tests.factories import create_collection_via_api, create_dataset, get_user_id
+
+pytestmark = pytest.mark.anyio
+
+
+async def _latest_details(session, action: str, resource_id: uuid.UUID) -> dict:
+    """Return the newest audit row's details for one action/resource."""
+    # The handler committed on its own connection; drop any snapshot this
+    # session is holding so the SELECT below sees that commit.
+    await session.rollback()
+    row = (
+        await session.execute(
+            select(AuditLog.details)
+            .where(AuditLog.action == action, AuditLog.resource_id == resource_id)
+            .order_by(AuditLog.created_at.desc(), AuditLog.id.desc())
+            .limit(1)
+        )
+    ).scalar_one()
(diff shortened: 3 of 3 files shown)


ishiland:
@codex review

ishiland:
@codex review

PR #1490, opened 2026-08-14
fix(observability): log production tracebacks plain so an exception cannot pin the event loop

### LINKED ISSUE #1485: Production exception logging rich-renders tracebacks with locals on the event loop: one exception pins the API for minutes
## Measured on the live demo (1.13.0, single uvicorn worker)

A dataset-metadata PATCH that raises (see the audit-details date bug) pinned the api container at 100% CPU for well over three minutes. py-spy against the worker mid-hang shows the loop stuck in rich, not in application code:

    Thread 58 (active+gil)
        split_and_crop_lines (rich/segment.py:344)
        render_lines (rich/console.py:1386)
        _render (rich/table.py:838)
        ...
        _proxy_to_logger (structlog/stdlib.py:280)
        exception (structlog/stdlib.py:252)
        dispatch (app/api/middleware/logging.py:40)

## Chain

1. A handler raises with ORM-heavy frame locals (session, record, dataset objects).
2. The logging middleware logs it via structlog `exception()`, whose console renderer pretty-prints the traceback with rich, including per-frame locals tables.
3. Rich's segment splitting goes quadratic over the huge reprs. This is synchronous CPU work on the event loop, so client disconnects cannot cancel it and every other request (health checks included) queues behind it.
4. With `--workers 1` (the prod compose default) the whole API is down until the render finishes. The victim request never reaches `request_completed`, so the outage is also invisible in the access log.

Effect: any reproducible unhandled exception is a multi-minute denial-of-service primitive against a single-worker deployment.

## Fix direction

In production (`ENVIRONMENT=production`), exception logging must use a plain formatter — no rich traceback, or at minimum `show_locals=False` with hard `max_frames`/width caps. Rendering could additionally be moved off-loop, but the primary fix is not paying a rendering cost proportional to local-variable size in the request path. Worth a regression test asserting the production log config never routes exceptions through rich.

ishiland:
## The incident

A dataset-metadata PATCH that raised pinned the demo's api container at 100% CPU for over three minutes (1.13.0, single uvicorn worker). py-spy against the worker mid-hang showed the loop inside rich, not application code:

```
split_and_crop_lines (rich/segment.py:344)
render_lines (rich/console.py:1386)
_render (rich/table.py:838)
...
_proxy_to_logger (structlog/stdlib.py:280)
exception (structlog/stdlib.py:252)
dispatch (app/api/middleware/logging.py:40)
```

Any reproducible unhandled exception was a multi-minute denial-of-service primitive against a single-worker deployment. The victim request never reaches `request_completed`, so the outage did not appear in the access log either.

## The chain

`setup_logging()` picked `structlog.dev.ConsoleRenderer()` whenever `LOG_JSON` was false — and `docker-compose.prod.yml` defaults it to `false`. That renderer takes its `exception_formatter` from `structlog.dev.default_exception_formatter`, which is `RichTracebackFormatter(show_locals=True, max_frames=100)` whenever rich is importable. rich is in the backend's transitive dependency set, so production had it.

Rendering a frame's locals costs `repr()` of every local in every frame, and rich's segment splitting is quadratic in the length of a single long line. A handler that raises with a session, a record and a dataset in scope hands rich a hundred frames of enormous single-line reprs. That is synchronous CPU on the event loop in the request path, so client disconnects cannot cancel it and health checks queue behind it.

`RichTracebackFormatter`'s `locals_max_length` and `locals_max_string` are not a fix: they truncate containers and `str` values, not the `repr()` of an arbitrary object, which is exactly what a SQLAlchemy session or model instance produces.

## What renders where now

`setup_logging()` takes a keyword-only `production` flag, wired at both entrypoints (`api/main.py`, `platform/jobs/worker.py`) to `settings.is_production`.

| ENVIRONMENT | LOG_JSON | renderer | exception path |
|---|---|---|---|
| production | false | `ConsoleRenderer` | `format_exc_info` in the chain + `plain_traceback` |
| production | true | `JSONRenderer` | `format_exc_info` in the chain |
| unset / development | true | `JSONRenderer` | `format_exc_info` in the chain |
| unset / development | false | `ConsoleRenderer` | rich, `show_locals=False` |

Production gets two independent guards, and the second is not decoration. `format_exc_info` resolves `exc_info` into a plain `exception` string inside the processor chain, so no renderer is ever handed a traceback object — and because that same list is the `ProcessorFormatter`'s `foreign_pre_chain`, it covers stdlib records (`uvicorn.error` and friends) as well as structlog's own. Pairing it with a non-plain `exception_formatter` would then make `ConsoleRenderer` emit a `warnings.warn` on every single exception, so `plain_traceback` is required alongside it; it also still owns any `exc_info` that reaches the renderer by another route.

Development keeps rich's syntax-highlighted frames, which are the reason the pretty renderer is worth having, and loses only the locals tables. There is no setting that bounds an arbitrary object's `repr()`, so the cap has to be `show_locals=False`.

Nothing is lost for debugging: exception type, message, and every stack frame with its source line still land in the log.

## Verification

Before writing the tests I reverted the fix in-place and confirmed each new assertion fails against the old code — five of the eight fail pre-fix, the other three are the JSON path (already plain) and the counterfactual below.

Measured through `RequestLoggingMiddleware` with one 5.8 KB frame local, same request both times:

| | log output | locals sentinel | box drawing |
|---|---|---|---|
| before | 83,062 bytes | present (2x) | yes |
| after | 3,675 bytes | absent | no |

New `backend/tests/test_exception_log_rendering.py`:

- **Structural** — builds the production config and walks every processor it can route a record through (structlog's chain, the `foreign_pre_chain`, and the render chain), asserting none is a `RichTracebackFormatter` and none carries one as its `exception_formatter`. Parameterized over both log formats, since `ENVIRONMENT` and `LOG_JSON` are independent settings. Separate tests pin `plain_traceback` on the console renderer, `format_exc_info` in both chains, and `show_locals=False` on the development formatter.
- **Behavioral** — an exception through the middleware must still log its type, message, raising frame and `Traceback (most recent call last)`, and must contain neither the sentinel that only a frame local's `repr()` can produce nor any of rich's box-drawing characters.
- **Cost independence, without a stopwatch** — the same exception raised with a 500x larger frame local emits a byte-identical traceback.
- **Counterfactual** — asserts structlog's stock rich formatter *does* render the sentinel, so the two absence assertions above cannot pass vacuously.

`caplog` sees no structlog records in this codebase, so capture is via the configured handler's stream and `capsys`, through the existing `tests/_logging_state.py` helper (extended to pass `production` through).

Commands run from the worktree backend with `.env.test` sourced:

- `uv run pytest tests/test_exception_log_rendering.py tests/test_phase_273_log_redaction.py tests/test_access_log_path_redaction.py tests/test_logging_conftest_guard.py tests/test_notification_channels.py tests/test_layering.py -q` — 109 passed (random ordering).
- `uv run ruff check .` — passed. `uv run ruff format --check .` — passed.

None. No new environment variable: the flag is derived from the existing `ENVIRONMENT` setting via `settings.is_production`, which already falls back to `LOG_JSON` when `ENVIRONMENT` is unset.

A second commit documents the consequence. `ENVIRONMENT=production` now controls three things rather than two, so the field comment in `config.py`, the `is_production` docstring, and `.env.example` all enumerate traceback rendering alongside docs exposure and the Secure cookie. This setting exists in the first place because security posture used to be keyed off `LOG_JSON`, a flag documented as controlling log format only (SEC-005); leaving a third behavior undocumented on the setting that replaced it would repeat that.

Two `_MODULE_LOC_CAPS` ratchets move, each in the commit that causes it: `backend/app/api/main.py` 1466 -> 1470 (the boot call exceeds 88 columns as one line) and `backend/app/core/config.py` 1104 -> 1107 (the comment and docstring above).

Fixes #1485

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -157,10 +157,13 @@ GEOLENS_ADMIN_PASSWORD=
 # PUBLIC_BASE_URL=
 
 # --- Deployment environment ---
-# Controls security-sensitive behavior: API docs exposure (/docs, /redoc) and
-# the Secure flag on the OAuth session cookie.
+# Controls security-sensitive behavior: API docs exposure (/docs, /redoc), the
+# Secure flag on the OAuth session cookie, and how tracebacks are rendered.
 #   development -> docs enabled, session cookie WITHOUT Secure (safe for HTTP)
 #   production  -> docs disabled, session cookie WITH Secure (HTTPS only)
+# Production also formats tracebacks with the stdlib traceback module instead
+# of rich, so logging an exception cannot spend minutes rendering frame locals
+# on the event loop (#1485).
 # Set this to "production" on every public-facing, TLS-terminated deployment.
 # If unset, the app falls back to using LOG_JSON as the production indicator
 # (backward compatibility with deployments configured before this setting).
--- a/backend/app/api/main.py
+++ b/backend/app/api/main.py
@@ -65,7 +65,11 @@ from slowapi.middleware import SlowAPIMiddleware
 import slowapi.middleware as slowapi_middleware_module
 
 # Configure structured logging before app creation so lifespan logs are structured
-setup_logging(json_logs=settings.log_json, log_level=settings.log_level)
+setup_logging(
+    json_logs=settings.log_json,
+    log_level=settings.log_level,
+    production=settings.is_production,
+)
 structlog.contextvars.bind_contextvars(service="api")
 
 logger = structlog.stdlib.get_logger(__name__)
--- a/backend/app/core/config.py
+++ b/backend/app/core/config.py
@@ -175,7 +175,10 @@ class Settings(BaseSettings):
     # behaviors — API docs exposure (/docs, /redoc) and the Secure flag on the
     # OAuth session cookie (SessionMiddleware https_only). Previously these were
     # keyed off LOG_JSON, an innocuously-documented log-format flag.
-    #   "production"  -> hardened posture (docs hidden, Secure cookie)
+    # fix(#1485): also selects plain traceback rendering, because rich's
+    # frame-locals tables made one exception a multi-minute event-loop stall.
+    #   "production"  -> hardened posture (docs hidden, Secure cookie,
+    #                    plain tracebacks)
     #   "development" -> open posture (docs shown, no Secure cookie)
     #   unset (None)  -> fall back to LOG_JSON for backward compatibility
     # Set ENVIRONMENT=production on any public, TLS-terminated deployment.
@@ -860,8 +863,8 @@ class Settings(BaseSettings):
 
     @property
     def is_production(self) -> bool:
-        """Whether to enforce the production security posture (API docs hidden,
-        Secure session cookie).
+        """Whether to enforce the production posture (API docs hidden, Secure
+        session cookie, plain tracebacks).
 
         SEC-005: driven by the explicit ENVIRONMENT setting. When ENVIRONMENT is
         unset, fall back to LOG_JSON (the de-facto production switch before this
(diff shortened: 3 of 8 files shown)


ishiland:
@codex review

ishiland:
@codex review

PR #1492, opened 2026-08-14
test(e2e): fail fast when a worktree run would silently test main

### LINKED ISSUE #1480: test(e2e): a worktree run silently tests `main`, so results are meaningless in both directions

ishiland:
Closes #1480.

## What

A guard in `playwright.config.ts` that throws before any test runs when the worktree you are in and the stack you are pointing at disagree. 64 lines added, nothing removed.

## The design, and a correction to the issue

The issue offered three detection strategies and called git-rev-against-a-build-marker "the most correct." That was wrong, and I [corrected it there](https://github.com/geolens-io/geolens/issues/1480#issuecomment-5292474577): `.git` is not mounted into any container and no build marker exists in `backend/app`, so the API cannot report which source it runs. Stamping one is also meaningless for a `--reload` bind-mount stack.

What shipped is entirely client-side:

1. Engage only when `E2E_BASE_URL` is unset or points at the shared `:8080` stack.
2. Detect a linked worktree via `--absolute-git-dir` vs `--git-common-dir` (equal in the main checkout, different in a worktree).
3. Diff against the merge-base with `main`, plus `git status --porcelain`, since uncommitted changes are equally absent from the stack.
4. Block only if a changed path is under `frontend/`, `backend/app/` or `backend/alembic/`.

Step 4 is the part worth reviewing. Deriving the blocklist from the actual mount list, rather than exempting only `e2e/`, keeps docs-only, `.github/`-only and `backend/tests/`-only branches runnable. `backend/tests` and `scripts` are bind-mounted but cannot change what the browser sees.

## Verification

Six cases, exercised with `playwright test --list`, which loads the real config without starting a browser or touching the stack:

| case | expected | result |
|---|---|---|
| worktree + `frontend/src/main.tsx` change | block | blocked, named the file |
| worktree + `e2e/*.spec.ts` change only | allow | 163 tests listed |
| worktree + `README.md` change only | allow | 163 tests listed |
| worktree + app change, `E2E_BASE_URL=:5174` | allow | 163 tests listed |
| worktree + app change, `E2E_ALLOW_WORKTREE=1` | allow | 163 tests listed |
| main checkout / CI | inert | `git-dir == git-common-dir`, returns before any further git call |

Also confirmed a non-git directory makes `git` fail into an empty string and the guard returns early rather than throwing.

This PR touches `playwright.config.ts`, which is in ci.yml's path filter, so CI will run the e2e suite here. That independently proves the guard is inert in CI rather than relying on my reasoning about it.

## Note for review

This is the first root-level `.ts` file to import a node builtin (`node:child_process`). There is no root `tsconfig` and `@types/node` is not installed at the root, so nothing typechecks this file; Playwright transpiles it with esbuild, which is what the six runs above exercised. Flagging it in case you would rather the guard live in a `globalSetup` file instead.


ishiland:
@codex review

Guard in `playwright.config.ts`. Worth checking specifically:

1. **The blocklist derivation.** I block only `frontend/`, `backend/app/`, `backend/alembic/` because those are what the dev stack bind-mounts and serves to the browser. Is there a mounted path I missed that could change a Playwright result? `docker-compose.yml` also mounts `./backend/tests`, `./scripts` and `./.github`, which I judged incapable of affecting a browser test.

2. **The merge-base fallback.** `git merge-base HEAD main` falls back to `origin/main`. If neither resolves, `mergeBase` is empty and only `git status --porcelain` contributes, so a committed-but-undiffed worktree change would slip through. Is that failure mode worth hardening, or is it acceptable since it fails open rather than blocking valid runs?

3. **`git status --porcelain` parsing.** I take `line.slice(3)`, which mishandles renames (`R  old -> new`). Worst case is a path that matches nothing and the guard under-blocks. Confirm that is the only consequence.

4. That the guard is genuinely inert in CI. `--absolute-git-dir` equals `--git-common-dir` in a normal checkout, so it returns before any other git call, including under a shallow clone where `main` may not exist.

ishiland:
@codex review

Both P1s were correct and are fixed in `ed00ab93a`. I verified each before acting rather than taking them on faith.

**P1 #1 — alternate config bypass.** Confirmed: `playwright.builder-hardening.config.ts` exists, is selected with `-c` by `npm run e2e:smoke:builder-hardening`, defaults to the same `:8080`, and never called the guard. Extracted to `playwright.worktree-guard.ts`; both configs now call `assertWorktreeMatchesStack()`. Tested: the hardening config with a `frontend/` change now blocks, where it previously listed 20 tests and ran.

**P1 #2 — the remediation was harmful for backend changes.** Confirmed against `docker-compose.yml:146` (`api — host :8001 -> container :8000`): `:8001` is the MAIN checkout's API. So the `:5174` recipe I printed would have produced worktree-frontend plus main-backend, and `E2E_BASE_URL=:5174` also disables the guard, so a `backend/app/` change would have false-passed with no second warning. The message now branches by offender: frontend-only gets the `:5174` recipe, backend/alembic gets an explicit statement that it will NOT exercise the change and needs a worktree-backed stack.

I also traced #2 to its source. `AGENTS.md` gave that recipe unqualified, which is where the first revision got it, so this commit annotates it as frontend-only and says why.

Two things I did not change, and would like your read on:

1. The empty-`mergeBase` case still fails open (only `git status --porcelain` contributes). I judged failing open correct for a guard whose false positives block valid work, but it does mean a committed change on a branch with no `main` ancestry slips through.
2. `git status --porcelain` parsing is still `line.slice(3)`, so a rename entry (`R  old -> new`) yields a path matching nothing. Under-blocks, never over-blocks. Confirm that is the only consequence.

Please re-review the reduced surface.

ishiland:
@codex review

Third P1 confirmed and fixed in `274d6474e`. I reproduced it on this branch before changing anything: main had advanced one commit touching `backend/app/...`, my merge-base diff saw only my own 4 files and would have allowed the run.

**Fix, with one deliberate departure from the suggestion.**

The comparison target is now the MAIN CHECKOUT's actual HEAD, derived from `--git-common-dir`, rather than `main` or `origin/main`. That checkout is what gets bind-mounted, so it is literally what the containers serve. This matters concretely: on this repo local `main` was at `021e40454` and `origin/main` at `fdaf3927c`, and the containers were serving `021e40454`. Comparing against `origin/main` reports divergence that is not actually being served, which would be a false block.

**Where I did not follow the recommendation.** You proposed comparing the worktree directly with main and treating any served-path difference as a failure. I split it by direction instead:

- My changes missing from the stack: still a hard block. The run reports on code I did not write, which is unambiguously meaningless.
- Behind the served tree: a printed warning naming the paths, plus a rebase remedy, and the run proceeds.

The reasoning is that those two are not equally bad. A stale worktree's specs run against NEWER app code, which is what CI will do against main after merge anyway, so the result is arguably more representative rather than less. And main here moves several times a day, so hard-blocking every stale worktree would block nearly every run including spec-only branches. A guard that blocks constantly gets switched off with `E2E_ALLOW_WORKTREE=1`, and then it guards nothing at all. I would rather it stay credible.

If you think the stale direction should also hard-block, say so and I will change it — I am making a judgement call about nuisance-blocking, not disputing the underlying analysis, which was correct.

**Verified after the change**, both configs: clean worktree allows (171 / 20 tests), frontend change blocks with the frontend recipe, backend change blocks with the backend warning, e2e-spec-only allows, `E2E_ALLOW_WORKTREE=1` allows. Stale case exercised with a throwaway worktree checked out before a frontend commit: warns, names 5 paths plus a count of the rest, and proceeds.

ishiland:
@codex review

Both fixed in `6c4fa0523`, and both were right.

**P1 — dirty served tree.** The comparison now folds `git status --porcelain` from the main checkout into `divergent`, so an uncommitted or untracked file there under a served path counts. You are correct that resolving only the committed HEAD was blind to it.

**P2 — shell splitting.** Fixed at the root rather than by quoting: `git()` now takes an argv array and uses `execFileSync`, with `cwd` instead of `-C`. That removes the class of bug rather than escaping around one instance. Verified against a repo at `/tmp/dir with spaces/repo`: `rev-parse --verify --quiet HEAD` and `--abbrev-ref HEAD` both resolve, where the shell form would have split the path and fallen back.

**Also closed the rename gap from round one.** Porcelain reports renames as `XY old -> new` on one line, so that is now split and both sides kept. I checked the parser against all three forms rather than assuming:

```
" M README.md"                                        -> "README.md"
"?? backend/app/newthing.py"                          -> "backend/app/newthing.py"
"RM frontend/src/a.tsx -> frontend/src/renamed.tsx"   -> both sides
```

`slice(3)` is correct for all three; an earlier suspicion that it was off by one turned out to be an artifact of my synthetic fixture, not a real defect.

**One limitation I want to be explicit about rather than let it read as verified.** I could not exercise the P1 fix end to end. Doing so means dirtying the shared main checkout under a served path, and another session is actively working in this repo — the user asked me not to disturb it. So I verified the mechanism in isolation (a scratch repo with a modified served file, an untracked served file, and an unserved file, confirming the served subset is selected correctly) and the wiring by construction, but not the full path against the real checkout. Worth a second pair of eyes on that specific branch.

Regression after the change, both configs: clean allows (171 / 20), frontend blocks with the frontend recipe, backend blocks with the backend warning, spec-only allows, override allows.

ishiland:
@codex review

Fixed in `f7e726695`, and this was a good catch on a bug I had already seen and talked myself out of.

`git()` trimmed. Porcelain's first line begins with a space for an unstaged change, so trim ate the status column and `slice(3)` then ate a real character. Because trim only affects the ends of the whole output, it struck only when a space-leading entry sorted first, which is why it hid.

Porcelain and diff now go through `gitRaw` (no trim) with `-z`, parsed NUL-terminated. That also drops path quoting for non-ASCII names, and makes renames explicit: an `R`/`C` entry is followed by one extra field holding the origin path, both sides kept.

Reproduced before and after, against a fixture whose only change is an unstaged served file, running the **exported** `porcelainPaths` rather than a copy:

```
git says            " M frontend/src/a.tsx"
old (trim+slice)    ["rontend/src/a.tsx"]    served: []
new (-z parser)     ["frontend/src/a.tsx"]   served: ["frontend/src/a.tsx"]
```

Worth recording why I missed it twice. In round two the same corruption appeared as `"EADME.md"` in my own output. I retested by piping `git status --porcelain` straight into node, saw correct parsing, and concluded the first result was a fixture artifact. But that retest never went through `.trim()` — I verified a path the real code does not take, then used the clean result to dismiss real evidence. The lesson is to reproduce through the shipped call path, which is why `porcelainPaths` is now exported and the check runs against it directly.

Regression, both configs: clean allows (171 / 20), tracked frontend change blocks, untracked served file blocks, spec-only allows, override allows.

Still outstanding and unverified end to end, as flagged last round: the dirty-main-checkout branch of this. Exercising it means dirtying the shared checkout while another session works in the repo, which I have been asked not to do.

ishiland:
@codex review

Fixed in `fb7690d5f`. Confirmed the behaviour first:

| form | output for a committed `frontend/a.ts` -> `docs/a.ts` |
|---|---|
| `--name-only -z` | `docs/a.ts` only |
| `--name-only --no-renames -z` | `docs/a.ts` **and** `frontend/a.ts` |
| `--name-status -z` | `R100 frontend/a.ts docs/a.ts` |

I went with `--no-renames` rather than parsing `--name-status`. This code only needs the *set* of touched paths, not the rename relationship, and disabling detection emits the delete and the add separately — which is that set, with no extra parser to get wrong. Applied to both the merge-base diff and the divergence diff.

Reproduced on this branch by committing a move of `frontend/src/main.tsx` out of `frontend/`: with detection the diff reports zero `frontend/` paths, with `--no-renames` it reports one, and the guard blocks with the frontend recipe.

One deliberate asymmetry: `git status --porcelain -z` keeps its existing `R`/`C` pair parser instead of also taking `--no-renames`. That path is already correct and tested, and an unsupported flag would make `gitRaw` return empty — which fails **open** and blinds the guard entirely. Not a trade worth making for symmetry.

Regression, both configs: clean allows (171 / 20), tracked frontend change blocks, untracked served file blocks, spec-only allows, override allows.

Still unverified end to end, unchanged from previous rounds: the dirty-main-checkout branch, because exercising it means dirtying the shared checkout while another session is working in this repo.

ishiland:
@codex review

Fixed in `eec4caf45`. This was the one item I flagged myself in round one as "fails open, is that acceptable" and argued for leaving. You were right that it is not acceptable, and the orphan-branch reproduction is what makes it concrete.

With no common ancestor there is no way to separate my divergence from the served tree's, so `mine` now falls back to comparing against the served tree directly (`mineBase = mergeBase || mainRef`) and treats all of it as mine. That fails closed.

Verified against a real orphan worktree carrying a committed `frontend/` change, with a counterfactual against the previous commit:

```
new  (mineBase fallback)    blocked = 1
prev (mergeBase only)       blocked = 0, warned = 1
```

Regression, both configs: clean allows (171 / 20), tracked frontend change blocks, untracked served file blocks, spec-only allows, override allows.

## Where this stands

Six rounds, six real defects, every one a genuine hole. I want to name the pattern rather than just keep patching: a 64-line guard is now ~185 lines of git plumbing, and each round has found another way the git layer lies about what changed — rename detection hiding an endpoint, porcelain's leading status column, missing merge bases, HEAD versus working tree.

That is a signal about the approach, not just the code. The alternative worth weighing is not more edge-case handling but a different oracle: have the stack report what it is serving. A build marker was rejected in the issue because `.git` is not mounted, but a far cheaper version exists — the API could expose a hash of the mounted source tree, or the frontend could serve a build id, and the guard compares one value instead of reconstructing the diff from git.

I am not proposing to swap approaches in this PR. But if round seven finds a seventh, that is the direction I would take rather than a seventh patch, and I would rather say so now than discover it at round ten.

Still unverified end to end, unchanged: the dirty-main-checkout branch, because exercising it means dirtying the shared checkout while another session works in this repo.

ishiland:
@codex review

Fixed in `841a4b34b`, and you were right that this one over-blocks.

The merge-base delta was being used as the blocking signal when it is only good for *classifying* divergence into "mine" and "theirs" for the message. Divergence between the working tree and the served tree is the ground truth, so blocking is now gated on membership in that set. `git diff` cannot see untracked files, so mine's untracked entries are folded into the divergence set separately — those are real divergence, being absent from the served tree entirely.

Verified with a genuine counterfactual on `backend/app/api/main.py`, taking main's exact content onto a branch forked 8 commits earlier:

```
merge-base delta lists it     1   (the trap)
differs from served tree      0   (byte-identical)
previous commit               blocked = 1   (wrong)
this commit                   blocked = 0   (correct)
```

My first attempt at that test was vacuous — the file variable came back empty and both arms trivially reported zero. I caught it and rebuilt the fixture rather than reporting the pass.

## Where this stands

Seven rounds, seven real defects. Last round I said a seventh finding would send me to a different oracle (have the stack report a hash of what it serves) rather than an eighth patch. I am revising that, and want the reasoning on the record so you can push back.

Rounds one through six were all the same shape: git misreporting what changed. This round's fix is structurally different. Making served-tree divergence the authority is the question the guard should always have asked, asked directly, and it removes the class those earlier bugs came from rather than another instance. The oracle alternative needs backend changes to expose a build hash and now buys considerably less than it would have before this restructure.

Worth noting the remaining failure modes have flipped direction. Rounds one through six were under-blocking — silent wrong results, which is the entire point of #1480. This round was over-blocking, which is annoying but recoverable with `E2E_ALLOW_WORKTREE=1`. That is the right direction to be wrong in.

If this round is clean I intend to enqueue. If you find another P1, that hits the project's standing escalation rule and a human reviews the whole diff instead of taking my eighth self-assessment.

Regression, both configs: clean allows (171 / 20), tracked frontend change blocks, untracked served file blocks, spec-only allows, override allows.

Still unverified end to end, unchanged across all rounds: the dirty-main-checkout branch, because exercising it means dirtying the shared checkout while another session is working in this repo.

ishiland:
@codex review

Fixed in the head commit. You were right, and this finding is the one that named the actual problem rather than another instance of it.

Both trees can differ from `mainRef` identically — the same dirty edit on each side, or an untracked file present in both — and nothing in the path-level union ever compared the two working trees to each other.

**That is the through-line of all eight rounds.** This file was reconstructing "do two directories differ" out of git metadata, and git metadata answers a different question: what changed relative to a COMMIT. Rename detection hiding an endpoint, porcelain's leading status column, the missing merge base, HEAD versus working tree, byte-identical branches, and now this — each was a new way for that reconstruction to be wrong. I said last round that a further finding would send me to a different oracle and then argued myself out of it. You have now supplied the oracle, and it is a better one than the build-hash idea I floated: compare the bytes.

Git now only nominates candidates. A byte comparison of the two working trees decides. Correctness no longer depends on git reporting changes the way this code expects — only on the candidate set being over-inclusive, which eight rounds of widening have made it. Cost is a read of the handful of changed paths, not the whole tree.

Semantics, verified directly against the exported comparator:

```
identical bytes in both        -> not divergent
different bytes                -> divergent
present in only one tree       -> divergent
absent from both               -> not divergent
unresolvable root              -> divergent (fails closed)
```

Regression, both configs: clean allows (171 / 20), tracked frontend change blocks, untracked served file blocks, spec-only allows, override allows.

## Process note

This is round eight without a clean pass, which trips this project's standing escalation rule, so I read the whole file myself rather than relying on another self-assessment. That personal read reached the same conclusion your comment did, independently: the git-metadata reconstruction was the defect generator, and every prior round treated a symptom.

I am not going to claim this one is certainly the last. What I will claim is narrower and checkable: after this change, a wrong answer requires the candidate set to omit a path entirely, rather than requiring git to report a change in the shape this code guessed. That is a much smaller surface, and it fails toward over-blocking, which is recoverable with `E2E_ALLOW_WORKTREE=1`, rather than toward the silent false pass #1480 exists to prevent.

Unverified end to end, unchanged across all rounds: the dirty-main-checkout branch, because exercising it means dirtying the shared checkout while another session is working in this repo.

ishiland:
@codex review

Fixed in `f3fb5ab85`. Correct, P1, and a silent false pass — the exact failure this PR exists to prevent.

Reproduced first. `git status --porcelain` in its default untracked mode collapses a new directory into one entry, `?? frontend/src/widget/`. The byte comparison read that as EISDIR here and ENOENT on the served side, `readOrNull` mapped both to `null`, and the pair was declared absent from both.

Two fixes, because either alone leaves a hole:

- `--untracked-files=all` on both status calls, so untracked files are listed individually and a directory never becomes a candidate path.
- `readOrNull` now separates genuinely absent (`ENOENT` -> `null`) from present-but-unreadable (anything else -> `undefined`), and an unreadable side fails closed. A directory, a permissions error, or any future surprise is treated as divergent rather than equal.

Counterfactual against the round-eight commit, with a new untracked file in a new directory under `frontend/`:

```
round eight   blocked = 0   (silent pass)
this commit   blocked = 1
```

Comparator semantics re-verified directly, including the new case: identical -> not divergent, only-one-side -> divergent, absent-both -> not divergent, directory-vs-missing -> divergent.

## An accurate attribution

This defect is mine, not an inherited one. It was introduced by the round-eight byte-comparison change; the collapse behaviour only became reachable once directory paths could be fed to `readFileSync`.

I should also record that I had seen this. In round four a synthetic fixture produced `?? backend/` and I dismissed it as an artifact of that fixture rather than as git's documented directory collapsing. That is the second time in this review I have explained away a real signal — the first was `"EADME.md"`, which was the porcelain trim bug. Both times the tell was that I re-tested through a path the shipped code does not take and trusted the clean result.

Regression, both configs: clean allows (171 / 20), tracked frontend change blocks, spec-only allows.

Unverified end to end, unchanged: the dirty-main-checkout branch, because exercising it means dirtying the shared checkout while another session works in this repo.

ishiland:
@codex review

Both addressed in `2e997d350`.

**P1 — conceded, and I am reversing my own round-three decision.** Stale served-path divergence now blocks instead of warning.

My argument for warning was that hard-blocking every stale worktree in a repo whose main moves hourly would get the guard switched off permanently. That argument was built on commit ancestry: "behind" was an inference, and I did not want to block on an inference. Since the byte comparison landed, these paths are *proven* to differ on disk. So the run really would report on code that is not in this worktree, which is precisely the failure this PR exists to prevent. The remedy is also one command — rebase — rather than standing up a separate stack, and `E2E_ALLOW_WORKTREE=1` remains documented in the message for the spec-only case where testing against the served checkout's newer app code is what you actually want.

Verified with a counterfactual on a worktree forked 8 commits back, differing at `backend/app/api/main.py`:

```
previous commit   blocked = 0, warned = 1
this commit       blocked = 1, message names the rebase remedy
override          still escapes (171 tests listed)
```

**P2 — real bug, fixed.** `readFileSync` followed symlinks, so two different links whose targets held identical bytes compared equal, and so did a link versus a regular file with matching content. `lstat` first: a symlink is compared by its stored target behind a sentinel prefix, and anything that is not a regular file fails closed. That also replaces the EISDIR special case from last round with an explicit `isFile()` check, which is the more honest form of it.

Verified directly against the exported comparator:

```
different symlink targets, same target content   -> divergent
regular file vs symlink, same content            -> divergent
```

Regression, both configs: clean allows (171 / 20), tracked frontend change blocks, untracked file in a new directory blocks, spec-only allows, override allows.

## Standing note

Eleven rounds, thirteen findings. Two of the thirteen were defects I introduced (the round-eight byte-comparison hole and the symlink gap in the same change); the rest were inherited from the git-metadata approach that rounds eight through ten replaced.

The remaining known gap is unchanged and I am not going to quietly let it read as covered: the dirty-main-checkout branch has never been exercised end to end, because doing so means dirtying the shared checkout while another session is actively working in this repo.

ishiland:
@codex review

Your P1 was correct — `docker-compose.yml:681` maps `${FRONTEND_PORT:-8080}` and `.env.example:302` sets it, so a stack on 8081 skipped the check entirely.

I have not fixed it. I have removed the approach that made it possible.

## What changed

The guard no longer tries to work out whether this worktree's changes are present in the stack. It asks the one question it can answer reliably — am I in a linked worktree — and requires `E2E_ALLOW_WORKTREE=1` to proceed. 297 lines become 81.

## Why

Eleven rounds, fourteen findings, every one real:

- rename detection hiding a served endpoint
- porcelain's leading status column eaten by `.trim()`
- missing merge base failing open
- HEAD compared instead of the served working tree
- byte-identical branches over-blocked
- collapsed untracked directories read as absent
- symlinks compared through their targets
- a configurable `FRONTEND_PORT` skipping the check

Two of those were mine, introduced by fixes in rounds eight and ten. The rest were inherent to reconstructing "do two directories differ" from git metadata and filesystem calls. That surface spans git reporting, filesystem semantics and compose configuration, and each round bought one fix and revealed another gap. Your findings did not get weaker — they kept being right, which is the evidence that the approach rather than the code was wrong.

The simple version cannot be wrong about whether your changes are in the stack, because it never asks. There is no metadata to misread, no bytes to compare, no port to detect.

## The trade, stated plainly

A legitimate run now needs one environment variable — including a spec-only branch, since the guard cannot distinguish one and deliberately does not try. That variable is exactly what the 297 lines existed to avoid.

I think that is the right trade for this specific guard. Blocking a run that would have been fine costs a variable. Allowing one that reports on code you did not write costs a wrong answer that looks identical to a right one, which is the failure #1480 is about.

## Verified

- blocks from a worktree, both configs, including with `E2E_BASE_URL` on a non-default port
- allows with `E2E_ALLOW_WORKTREE=1`, both configs (171 / 20 tests listed)
- inert in the main checkout and in CI, where git-dir equals git-common-dir

The error message keeps the frontend recipe and the explicit warning that it is frontend-only because `:8001` is the main checkout's API container. `AGENTS.md` is updated to match, including that spec-only branches also need the flag.

Please review the deletion as carefully as you reviewed the additions — if the simple check has a hole, it is a much cheaper one to find now.

ishiland:
@codex review

Correct, and thank you for reviewing the deletion as carefully as the additions — that is exactly where I expected a hole and could not see one myself.

`git()` swallowed every failure into an empty string, and the early return read that as "not a repository" and allowed the run. An older git without `--path-format`, or a `safe.directory` ownership rejection, disabled the guard silently from inside a worktree. Same false pass, new costume.

## Fix

Detection no longer shells out at all. A linked worktree's `.git` is a FILE containing `gitdir: /…/.git/worktrees/<name>`; the main checkout's `.git` is a DIRECTORY. Walking up from cwd for the nearest `.git` and testing which one it is answers the question with a single `stat`.

That removes the failure mode rather than handling it: there is no subprocess to fail, no git version to depend on, and no ownership check to be rejected by. The three outcomes are explicit — absent `.git` allows (a tarball install is legitimate and must keep working), a directory allows, and **anything else blocks**, so an entry we cannot classify fails closed by construction rather than by a branch someone has to remember to write.

## Verified against real paths, not fixtures

```
~/Code/geolens                          -> directory marker -> allow
~/Code/geolens/frontend/src             -> directory marker -> allow  (walk-up)
~/Code/geolens-worktrees/e2eguard       -> file marker      -> BLOCK
~/Code/geolens-worktrees/e2eguard/e2e   -> file marker      -> BLOCK  (walk-up)
/tmp                                    -> no marker        -> allow
```

The main-checkout and CI cases are the ones I cannot exercise by actually running Playwright, because that would mean placing files in the shared checkout while another session is working in this repo. Testing the exported detector against the real directories is the closest honest substitute, and I would rather say that than let a green regression table imply otherwise.

Behaviour otherwise unchanged: blocks from a worktree on both configs including a non-default `E2E_BASE_URL` port, allows with `E2E_ALLOW_WORKTREE=1` on both.

One residual I will name rather than leave implicit: a git submodule also has a `.git` file, so running Playwright from inside one would block. There is no submodule in this repo, the override is one variable, and distinguishing the two would mean parsing the file and reintroducing exactly the kind of conditional this rewrite removed. I judged that trade the wrong way round last time and would rather leave the simple rule.

ishiland:
@codex review

Correct, and the sharpest form of correct: you caught this module contradicting its own commit message. I wrote that unclassifiable entries "fail closed by construction", and then caught every stat error identically to a missing file and walked past it. An inaccessible or dangling `.git` allowed the run — the same fail-open the git version had, relocated to the filesystem. The claim was aspirational and the code was not.

## Fix

Only `ENOENT` and `ENOTDIR` continue the walk. Every other errno classifies the entry as `unreadable`, and `unreadable` blocks.

Detection also `lstat`s first, so a dangling symlink is an entry that EXISTS and must be classified rather than walked past as absent. A symlink that resolves is classified by its target; one that does not is `unreadable`.

## Verified against real fixtures, one per outcome

```
main checkout                  directory    allow
main subdirectory (walk-up)    directory    allow
linked worktree                file         BLOCK
no .git anywhere above         none         allow
dangling .git symlink          unreadable   BLOCK
.git under a chmod 000 dir     unreadable   BLOCK
```

The last two are the failure modes you named, built as actual fixtures rather than reasoned about.

End-to-end unchanged: blocks from a worktree on both configs, allows with `E2E_ALLOW_WORKTREE=1` on both.

## Note on the arc

Thirteen rounds. The last three findings have all been the same underlying mistake in different clothes — an error path swallowed into a value that reads as "fine". First `git()` returning `''`, then `catch {}` continuing the walk, and in both cases I asserted fail-closed behaviour in prose while writing fail-open code. That is worth naming because it is a pattern in my work here, not three unrelated slips: I keep treating the happy path as the specification and the error path as an afterthought, then describing the intent rather than the implementation.

The current version has exactly one mechanism and three outcomes, and every outcome now has a fixture behind it.

ishiland:
@codex review

Correct. Any non-empty string is truthy, so `E2E_ALLOW_WORKTREE=0` or `=false` bypassed the guard — acknowledging a risk the operator was explicitly trying to decline, which is the worst possible reading of that variable.

Only `1`, `true`, `yes` and `on` acknowledge now, case-insensitively and after trimming. Everything else, including `0` and `false`, does not.

Verified across the values that matter, run from a linked worktree:

```
unset   BLOCK      1      allow
0       BLOCK      true   allow
false   BLOCK      TRUE   allow
FALSE   BLOCK      yes    allow
no      BLOCK
off     BLOCK
```

## Process note

I should say plainly that I did not see this review until the user asked whether I was still watching the PR. The watcher had completed and I had not checked its output. That is a gap in how I have been running this loop — treating the completion notification as the only trigger rather than checking after each round — and it is worth recording next to the code findings, because a review loop that depends on me noticing is not a loop.

## Where this sits

Fourteen rounds, seventeen findings. The current guard is one mechanism, an `lstat` on `.git`, with three outcomes and a fixture behind each. The last two findings have been about input validation at its two edges — the errno branch and the opt-in value — rather than about the mechanism, which is the shape I would expect near the end. I am not going to predict a clean pass, having been wrong about that repeatedly, but the surface left to be wrong about is now small and enumerable: what `.git` is, and what the one environment variable says.

ishiland:
@codex review

We found this one independently and converged on the same defect, including the same reproduction from `/tmp`. It was fixed and pushed in `a8a0b86a3` before your comment landed, so your verdict is against the superseded `cb06d0695`.

I reached it by asking what your round-one finding about `-c` implied for the rest of the file, rather than waiting to be told — which is the first time in fifteen rounds that has happened, and worth recording as a fact about the process rather than a claim about the code.

## What shipped

Both anchors are checked — this module's own directory and cwd — and either one reporting a worktree blocks.

I deliberately did not simply swap cwd for `__dirname`, which is the narrower reading of your suggestion. If a loader ever provides no `__dirname`, a fallback to cwd would silently restore exactly this hole, and that fail-open-fallback shape is precisely what rounds twelve and thirteen were about (`git()` returning `''`, `catch {}` walking past). Two independent checks degrade to the old behaviour rather than to no behaviour.

Verified:

```
__dirname under Playwright's loader     resolves to the worktree root, cwd elsewhere
run from /tmp with -c into a worktree   previous commit: allowed
                                        this commit:     BLOCKED
five marker classifications             unchanged
```

## Residuals I am consciously accepting, so they are not mistaken for oversights

- A git **submodule** also has a `.git` file, so running Playwright from inside one blocks. No submodule exists in this repo, the override is one variable, and distinguishing them means parsing the file and reintroducing conditional complexity this rewrite deleted.
- On a **case-insensitive filesystem** a `.GIT` entry would match. Not a realistic checkout layout.
- If a loader provides **neither** `__dirname` nor a usable cwd, both anchors are empty and the guard allows. That requires `process.cwd()` to throw, which means the working directory was deleted mid-run.

Please review `a8a0b86a3`.

Issue #1286, opened 2026-08-08
fix(catalog): key duplicate-source guards on the canonical origin_ref pair


ishiland:
Both duplicate-source guards (service preview and STAC import) are keyed on the `origin_uri` string spelling. PR #1277's review surfaced how fragile that is: the refresh path briefly wrote a different spelling of the same source (bare base URL instead of `base_url/typename`) and the guard silently stopped catching duplicates for any dataset that had been refreshed. That instance was fixed at the writer (refresh now preserves the import's spelling, with a test asserting the two composition expressions against each other), but the class remains: any future writer that produces a different spelling of the same origin degrades the guards without failing anything.

## Scope

Re-key both guards onto the canonical structured identity: the `origin_ref` URL/layer pair, compared via the same `service_layer_identity` round-trip the refresh path uses, instead of comparing pointer strings.

- Keep the `source_url` fallback for un-backfilled datasets (rows that predate `origin_uri`), which is the guards' existing fallback contract.
- Failing-first coverage for the paths PR #1277's fix did not need: a never-refreshed dataset must still be caught, an un-backfilled dataset must still be caught, and a hypothetical writer that respells `origin_uri` must no longer be able to open the hole (that last one is the test that fails today with a spelling-keyed guard).

## Where

- Service preview guard: `backend/app/modules/catalog/sources/router.py` (the branch codex anchored at line 816 in PR #1277 round 11).
- STAC import guard: same shape, same file family.

Refs: PR #1277 (round-11 finding and the writer-side fix), #1222 (origin_uri backfill).

PR #1493, opened 2026-08-14
fix(demo): record the STAC self link so seeder imports can refresh

ishiland:
## What

Two changes in `scripts/seed-showcase.py`, found while completing the demo enrichment pass:

1. **`build_sentinel2` now passes `item_href`** (the STAC item's `rel=self` link) in each import item. The backend records it as `origin_ref.item_href` (#1222) and it is the one stored value a refresh can re-resolve against. The seeder queries Element84 directly (the in-app search proxy captures this link server-side; the direct path must supply it) and never sent it — so every seeder-imported scene was born unrefreshable: `refresh` 409 `origin_unavailable` and an empty Source-panel run history on every instance this script seeded. None-tolerant for catalogs that publish no self link.

2. **`refresh_and_report` surfaces the structured refusal code** (`[origin_unavailable]` vs `[refresh_not_applicable]`) instead of only the bare httpx status line. The bare line is why this took three probes to diagnose live: it hides the difference between "binding incomplete, re-import" and "wrong dataset kind, give up".

## Why now

The 08-14 demo completion pass rebuilt the sentinel map with freshly imported scenes and the refresh pass still 409'd — proving the gap is in this script's import payload, not (as previously recorded) in legacy bindings predating item recording.

## Verification

- `python3 -m py_compile` clean.
- Extraction expression run against a live Element84 `/search` response: both features yield proper `https://earth-search.aws.element84.com/v1/collections/.../items/...` self links, which pass the backend's `_validate_optional_stac_http_url` (plain `HttpUrl`, `None` passthrough).
- After merge: re-run the sentinel surgery + `--only sentinel2` on the demo; the seed-end refresh pass reporting success is the end-to-end proof.

### DIFF
--- a/scripts/seed-showcase.py
+++ b/scripts/seed-showcase.py
@@ -4516,6 +4516,13 @@ def build_sentinel2(api: Api, force: bool = False) -> str:
         #   dedupe-skip into a missing scene or an all-conflict raise after
         #   the maps were already gone. Keyed the same way the backend guard
         #   is (fix(#1286)): origin_ref.asset_href, source_url as fallback.
+        # * Deletes datasets BEFORE maps (round 7): a scene serving a
+        #   committed VRT or an in-progress generation makes its DELETE 409
+        #   DependentVrtError, and that check is server-side only (origin_ref
+        #   and derived_from expose no VRT sources client-side) - the delete
+        #   itself is the only authoritative probe. Failing on it aborts with
+        #   every stale map still standing, instead of discovering the block
+        #   after the showcase is gone.
         stale_maps = [
             m["id"]
             for m in api.list_all_maps()
@@ -4588,13 +4595,30 @@ def build_sentinel2(api: Api, force: bool = False) -> str:
                 f"ownerless, or own-but-renamed/manual) - they would "
                 f"dedupe-skip and leave the rebuilt map incomplete: {names}"
             )
+        for ds_id, title in doomed.items():
+            try:
+                api.delete_dataset(ds_id, title)
+            except httpx.HTTPStatusError as e:
+                blockers = ""
+                if e.response.status_code == 409:
+                    try:
+                        det = e.response.json().get("detail")
+                        if isinstance(det, dict) and det.get("dependent_vrts"):
+                            blockers = (
+                                f" (dependent VRTs: {det['dependent_vrts']})"
+                            )
+                    except ValueError:
+                        pass
+                raise RuntimeError(
+                    f"force recreate aborted: scene {title!r} could not be "
+                    f"deleted{blockers}; the stale showcase maps were NOT "
+                    f"deleted - resolve the dependency and rerun"
+                ) from e
+        if doomed:
+            print(f"  [force] deleted {len(doomed)} attached scene datasets")
         for map_id in stale_maps:
             api.delete_map(map_id)
             print(f"  [force] deleted existing map {map_id}")
-        for ds_id, title in doomed.items():
-            api.delete_dataset(ds_id, title)
-        if doomed:
-            print(f"  [force] deleted {len(doomed)} attached scene datasets")
     print(f"  importing {len(items)} TCI COGs by reference (no download)...")
     results = api.stac_import(SENTINEL_STAC, items, visibility="public")
     errored = [x for x in results if x.get("status") == "error"]
--- a/scripts/seed-showcase.py
+++ b/scripts/seed-showcase.py
@@ -4553,6 +4553,28 @@ def build_sentinel2(api: Api, force: bool = False) -> str:
                     and title.startswith("Sentinel-2 TCI ")
                 ):
                     doomed[ds_id] = title
+        if doomed:
+            # fix(#1493 review round 5): a scene someone layered into ANY
+            # other map is not ours to cascade away - keep it and say so. The
+            # kept scene then rides the dedupe skip-fallback below if the
+            # fresh search reselects its asset (reattached, still
+            # unrefreshable, and flagged [origin_unavailable] by the
+            # seed-end refresh report) - visible partial repair over silent
+            # imagery loss in an unrelated map. Rewiring the other map's
+            # layer to the replacement was rejected: mutating a map this
+            # seeder does not own is the same overreach in a different coat.
+            stale_set = set(stale_maps)
+            for m in api.list_all_maps():
+                if m["id"] in stale_set or not doomed:
+                    continue
+                for layer in api.get_map(m["id"]).get("layers", []):
+                    ds_id = layer.get("dataset_id")
+                    if ds_id in doomed:
+                        print(
+                            f"  [force] keeping scene {doomed[ds_id]!r} - "
+                            f"also layered in map {m.get('name', m['id'])!r}"
+                        )
+                        doomed.pop(ds_id)
         for map_id in stale_maps:
             api.delete_map(map_id)
             print(f"  [force] deleted existing map {map_id}")
--- a/scripts/seed-showcase.py
+++ b/scripts/seed-showcase.py
@@ -570,6 +570,7 @@ class Api:
         me = r.json()
         self.username = me["username"]
         self.user_id = me["id"]
+        self.roles = me.get("roles") or []
 
     @classmethod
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review

Small seeder-only change. Worth checking specifically:

1. **The `next()` over links.** Element84 items carry exactly one `rel=self`; is there any STAC catalog shape where multiple selfs or a relative href would slip through HttpUrl validation into a useless binding?
2. **The error-code surfacing.** `e.response.json()` inside the except - is there a path where `response` is unset on `HTTPStatusError`, or where `detail` is a list?

ishiland:
@codex review

The P2 was correct and is fixed in `e660db6a7`. I verified the claim at the cited flow before changing anything: an ordinary rerun exits at the map-exists skip, and under `--force` the dedupe's `status="skipped"` resolves back to the old dataset ids without touching `origin_ref` - so a rerun could never repair what this PR's first commit fixes forward.

Rather than an origin_ref backfill (there is no API surface for mutating a system-managed binding, by design), `--force` now deletes this seeder's own map + `Sentinel-2 TCI ` scene datasets before importing. That makes `--only sentinel2 --force` the supported repair path, and it fixes a defect the old `--force` had regardless: it stacked a second same-named map on top of the old datasets. Deletion uses the existing own-scoped helpers (`list_maps`/`list_own_datasets`), so a same-titled stranger dataset is never touched.

ishiland:
@codex review

Round-2 P2 confirmed and fixed in `cf91803f1`. Verified at the cited helper first: `list_maps()` keeps the FIRST match on a created_at-desc sort, so with stacked duplicates only the newest died and the cascade emptied the survivors. The force path now enumerates via `list_all_maps()` filtered by the same `created_by_username == api.username` predicate `list_maps` applies, and deletes every owned map under the name.

Scope note, to close the sibling class: other builders' `--force` still recreates beside same-titled predecessors - that behavior is documented at `datasets_by_title` and predates this PR. Sentinel2 is the one builder whose `--force` is now a documented REPAIR path (it must record fresh origin bindings), which is why delete-first is implemented here and only here.

ishiland:
@codex review

Both round-3 P2s confirmed and fixed in `1b127c950`, as one reshaped block.

**Ordering:** the force cleanup now sits after `if not items: raise` - search failure, empty results, or no usable visual assets all abort the builder before anything is deleted. Deletion still precedes the import call, which is the boundary that matters for the source_url dedupe.

**Selection:** scene datasets are now derived from the matched maps' own layers, intersected with owned + `raster_geolens` + `Sentinel-2 TCI ` prefix. The intersection is deliberate in both directions: your case (an operator's unrelated same-prefixed import, never attached to the showcase map) is spared, and the converse hazard - a shared context dataset attached to the map by a styling pass, which a bare attached-layer sweep would have deleted - is spared too.

One consequence stated openly rather than hidden: a prefix-titled ORPHAN from a half-failed old run (attached to no map) is no longer swept. If it shares a source_url with a fresh item, the dedupe skip-path reattaches it and that scene stays unrefreshable - and the seed-end refresh report now prints `[origin_unavailable]` for exactly that scene, so the state is visible, not silent. Sweeping unattached datasets by title alone is the data-loss shape this round removed; I chose visibility over reach.

ishiland:
@codex review

Round-4 P2 confirmed and fixed in `9f235b7c0`. I verified the premise at the guard first: the batch duplicate check in `stac_router.py` filters only on `source_format == "stac"` plus the href keys - no `created_by` scope - while the skip-fallback resolves through own-scoped `datasets_by_title()`. Both halves of your chain (silent partial map, all-conflict destruction) are real.

The fix is a preflight abort, not import-first ordering: importing before deleting would dedupe-skip against OUR OWN old scenes in the common no-new-acquisition repair case (same hrefs), and "preserve the old map until results are known" preserves only a husk - deleting the scenes cascades the old map's layers away regardless. So: selected hrefs are checked against every stac dataset NOT owned by this account (ownerless rows included - they block the dedupe identically), keyed `origin_ref.asset_href` with `source_url` fallback exactly as the backend guard keys (fix(#1286)), and a conflict aborts with owners named before anything is deleted. Deleting our own old scenes frees our hrefs; a foreign row is not ours to free, so refusing pre-destruction is the only honest outcome. Comprehension edge cases (foreign hit, ownerless fallback hit, own-row exclusion, non-stac exclusion, null-href exclusion) pinned by an offline harness before pushing.

ishiland:
@codex review

Round-5 P2 confirmed and fixed in `742980b25`. Doomed scenes found layered in ANY non-stale map are excluded from deletion, with the sheltering map named in the run output. I took the exclude arm of your either/or deliberately: rewiring the other map's layer onto the replacement dataset mutates a map this seeder does not own, which is the same overreach round 5 exists to remove, just in a different coat.

Downstream consistency checked rather than assumed: a kept scene holds its asset href, so if the fresh search reselects it the import dedupe-skips and the existing by-title fallback reattaches the kept (own) dataset - map complete, other map untouched, and the scene's unrefreshable state is printed as `[origin_unavailable]` by the seed-end refresh report. The kept scene is own-owned, so the round-4 foreign preflight correctly does not fire on it. Exclusion sweep shape (shared scene kept once with no double-report, unshared scene still deleted) pinned by an offline harness.

For the ledger at round 5 of 8: every finding so far has been confirmed at the cited code and fixed - none refuted.

ishiland:
@codex review

Round-6 P2 confirmed and fixed in `b80ab71d7`, taking exactly the predicate you proposed: a holding conflicts unless it is in the deletion set or the fallback provably resolves it (own-owned AND still carrying the seeder title for that asset). The ownership test alone was the wrong axis; resolvability is the right one, and foreign/ownerless/own-renamed/own-manual all land on the abort side of it now.

The restructure this required is worth noting for review: the preflight moved BELOW the doomed-set construction and the round-5 shared-scene exclusion, because "in the deletion set" is only meaningful against the FINAL doomed set - a kept-shared scene is not freed, and it passes preflight through the resolvable arm (it keeps its seeder title), not the freed arm. All destructive operations remain strictly after the preflight, so the round-3 ordering guarantee (nothing deleted until the search succeeded AND the preflight cleared) is intact. Six row classes (freed, own+titled kept, own-renamed, foreign, ownerless via source_url fallback, unrelated-asset) pinned by an offline harness.

Round 6 of 8.

ishiland:
@codex review

Round-7 P2 confirmed and fixed in `2bd7b5e3e`. I checked the preflight arm of your either/or first and it is not implementable: the dependency check in `service_lifecycle.py` is a raw SQL join over VRT source-links plus `staged_source_ids` on pending/running generations, and nothing client-readable exposes it - I probed a live VRT dataset and `origin_ref`/`derived_from` are both null. The delete itself is the only authoritative probe, and it also covers the in-progress-generation case a client-side snapshot would race against.

So the fix is the ensure-deletable-before-maps arm: doomed datasets are deleted FIRST, and a `DependentVrtError` 409 aborts immediately - naming the blocking VRTs from the response body - with every stale map still standing. Stated residual: scenes deleted before the failing one stay deleted, so an abort leaves the old map present but missing those layers. That is recoverable by rerunning after resolving the dependency, versus the previous shape where the block was discovered with the showcase already gone.

Round 7 of 8.

ishiland:
Both round-8 P2s confirmed and fixed in `316e9eac6`. The admin premise checked out (`--username` exists and non-admin map listings are visibility-filtered), so force recreate now fails fast without the admin role rather than running the protection sweep blind. And skipped imports now bind to the preflight-proved href holder before the by-title fallback, closing the newest-same-title mismatch; the binding is recorded only for rows the preflight kept (doomed rows are excluded before that arm), so it can never point at a deleted dataset.

Closing the review loop here per our standing process: eight rounds, ten findings, every one verified at the cited code and fixed, none refuted. As the round-8 gate I re-reviewed the complete diff in its final state (not the increments): the force path now orders admin-check -> search -> scan -> preflight -> datasets -> maps, with every destructive step behind every check, and the two harnessed resolution tables (preflight row classes, skipped-result binding) pin the logic offline. Known residuals, stated rather than hidden: the preflight is client-side and inherently racy against a concurrent import (the abort semantics keep the race non-destructive), and an abort mid-dataset-deletion leaves already-deleted scenes gone with the maps intact. Merging.

PR #1497, opened 2026-08-15
fix(audit): isolate each audit sink in a savepoint so a rejected row cannot destroy the caller's mutation

### LINKED ISSUE #1491: Audit sink isolation is flush-deep only: a failing audit INSERT still destroys the caller's mutation
Found during the #1489 review sweep (the #1484 fix).

AUDIT-03's intent is that audit sinks must not break the operation they record. The implementation try/excepts each sink inside `audit_emit` — but the DB sink only calls `session.add()`, which cannot fail. The actual INSERT executes at the caller's `flush()`/`commit()`, outside the guarded scope, so any defect in the audit row (the #1484 date was one instance of the class) rolls back the user's mutation along with the audit record.

#1489 removes the known trigger (non-JSON scalars via `model_dump`), but the structural gap remains: the guarantee is only as strong as the row-construction path, and nothing enforces it at the seam where the failure actually happens.

Fix direction is a product decision before an implementation one: either audit-fail-open (emit the row via a mechanism whose failure cannot reach the caller's transaction — separate session/savepoint with a loud error log) or explicitly document fail-closed as intended for audit-integrity reasons. A savepoint around the audit flush is the narrow mechanical option; a separate session changes durability ordering. Whichever way, a regression test should pin the chosen semantics with a deliberately poisoned audit row.

Related: `backend/app/modules/catalog/maps/service_history.py:47` already runs its history payload through `jsonable_encoder` — the codebase has the safe pattern in one sibling and not the audit path.

ishiland:
## What

`audit_emit()` runs each sink inside its own `SAVEPOINT` and flushes only what that sink staged, so an audit row the database rejects rolls back alone instead of taking the caller's mutation with it.

Closes #1491.

## Why the old guard did not hold

AUDIT-03's contract is that an audit sink must never break the operation it records, and `audit_emit()` has always try/excepted each sink. That guard was flush-deep only. `DefaultAuditSink.emit()` bottoms out in `session.add()`, which cannot fail, and the INSERT it stages runs at the **caller's** flush/commit, outside the guarded scope. So the guarantee was only ever as strong as the row-construction path.

#1484 was one instance: a `datetime.date` in a stdlib-JSON column 500'd a dataset PATCH and silently discarded the other fields in the same body. #1489 removed that trigger and left the structure alone.

## The load-bearing detail

`begin_nested()` flushes the whole session as the first act of taking its snapshot, and that flush is the caller's work, not the sink's. It therefore has to happen **outside** the guard.

Left inside it, a caller whose own mutation is broken would have their `IntegrityError` raised in the audit savepoint, rolled back with it, and logged as an audit failure. The user's edit vanishes behind a 200 and the log names the wrong culprit. That is #1484 inverted and much harder to diagnose, so it gets the more important of the two tests.

Measured against that exact wrong implementation, every other test in the file still passes and only `test_caller_side_failure_propagates_and_is_not_swallowed_as_audit` goes red.

## Fail-open, deliberately

The drop log now carries the whole event — actor, address and payload — instead of action and resource alone.

[NIST SP 800-53 AU-5](https://csf.tools/reference/nist-sp-800-53/r5/au/au-5/) is allocated to LOW, MODERATE and HIGH, and asks for an *alert* on an audit-logging process failure. Fail-closed is [AU-5(4)](https://csf.tools/reference/nist-sp-800-53/r5/au/au-5/au-5-4/), which sits in **no baseline** and is waived "unless an alternate audit logging capability exists". A complete log line is that capability: a rejected row lands in the application log rather than vanishing.

`details` is safe to log — already admin-readable through the audit API and CSV/JSON export, and carries fingerprints, token hints, ids and field names, never a credential.

`platform/config_ops/router.py` is the only `audit_emit()` call site in the app that commits for durability, so it is the only place the semantics changed. Its comment now says so. The `EnterpriseSiemSink` swallows its own transport errors, so a SIEM outage never reached this guard in the first place.

## Consequences at the ~96 call sites

- A caller relying on a later `commit()` to raise its `IntegrityError` now sees it raised at `audit_emit()`, because the savepoint cannot open without flushing first.
- Each event costs a `SAVEPOINT` / INSERT / `RELEASE` round trip that used to ride along with the caller's commit.

## Also

`log_action()` normalizes `details` through `jsonable_encoder`, matching the sibling `record_map_history_event`. The #1484 payload class is now *recorded correctly* rather than merely survived — the savepoint is the second line of defence instead of the only one. `None` is preserved as SQL NULL; an absent payload and an empty one are different rows.

A missing session returns early rather than raising. Pre-fix that input degraded by accident, raising inside `sink.emit()` where the try/except swallowed it. Moving the flush out of the guard turned the same input into a 500 on the OAuth generic-error path, whose entire job at that point is a clean 302 with `Referrer-Policy: no-referrer`.

## Verification

```
uv run pytest tests/test_audit_savepoint_isolation_1491.py tests/test_audit_sink.py tests/test_layering.py -q
58 passed
uv run ruff check app/ && uv run ruff format --check
All checks passed!
```

8 tests in the new file. The two log-content tests were checked against a counterfactual: reverting the field list makes `test_dropped_event_is_logged_in_full` fail with `KeyError: 'user_id'`, so they are not vacuous.

### DIFF
--- a/backend/app/core/logging_config.py
+++ b/backend/app/core/logging_config.py
@@ -23,7 +23,7 @@ instance produces.
 """
 
 import logging
-from collections.abc import MutableMapping
+from collections.abc import Mapping, MutableMapping
 from typing import Any
 
 import structlog
@@ -50,6 +50,10 @@ _SENSITIVE_FIELDS: frozenset[str] = frozenset(
     }
 )
 
+# Depth ceiling for redact_nested(). Audit payloads are two or three levels at
+# most; this only exists so a caller-supplied cyclic structure terminates.
+_MAX_REDACT_DEPTH = 8
+
 
 def _redact_sensitive_fields(
     _logger: Any, _method_name: str, event_dict: MutableMapping[str, Any]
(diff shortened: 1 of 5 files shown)


ishiland:
@codex review

ishiland:
@codex review

PR #1498, opened 2026-08-15
feat(auth): declare the cookie-mode headers as OpenAPI parameters

### LINKED ISSUE #1496: OpenAPI: declare the cookie-mode auth headers as operation parameters
Found during the docs API-reference sync (getgeolens.com#94, codex review): the cookie-mode browser auth flow shipped in v1.13.0 (#1446) negotiates with `X-GeoLens-Auth-Mode: cookie` and enforces double-submit CSRF via `X-CSRF-Token` (`backend/app/modules/auth/cookies.py:29-30`), but the OpenAPI spec mentions both only in prose. At v1.13.1, `/auth/login/`, `/auth/refresh/`, and `/auth/logout/` declare zero operation parameters, so generated SDKs and the docs reference's request UI cannot express the inputs needed to establish, refresh, or revoke a cookie session.

Fix direction: declare the headers as optional `Header(...)` parameters on the three auth routes (optional keeps the default body-token contract byte-identical for CLI/SDK/API consumers), regenerate `backend/openapi.json` and the SDKs (lockstep, REL-04), and let the docs snapshot pick it up at the next release sync.

Out of scope for getgeolens.com#94: the vendored snapshot must byte-match the shipped release (parity gate), so the spec change has to originate here.

ishiland:
## What

The cookie-mode browser flow (#1446, v1.13.0) negotiates with `X-GeoLens-Auth-Mode` and enforces double-submit CSRF with `X-CSRF-Token`, but the spec named both only in prose. `/auth/login`, `/auth/refresh/` and `/auth/logout/` published zero operation parameters, so generated SDKs and the docs reference's request UI had no way to express the inputs that establish, refresh, or revoke a cookie session.

This declares the headers as optional `Header(...)` parameters and regenerates every artifact derived from the spec.

## Header map, which differs from the issue

The issue asked for "the headers" on all three routes. I followed the handlers instead:

| Operation | `X-GeoLens-Auth-Mode` | `X-CSRF-Token` |
| --- | --- | --- |
| `POST /auth/login` | yes, `wants_cookie_auth` | no, it issues the CSRF cookie rather than reading one |
| `POST /auth/refresh/` | yes, `wants_cookie_auth` | yes, `enforce_csrf` |
| `POST /auth/logout/` | no | yes, `enforce_csrf` |

Logout never calls `wants_cookie_auth`. It accepts the refresh cookie and clears both cookies unconditionally, so an auth-mode parameter there would document an input the handler ignores.

Two smaller drifts from the issue text, both cosmetic: the published login path is `/auth/login` with no trailing slash (the slash form is a hidden ROUTE-01 alias), and the header constants sit at `backend/app/modules/auth/cookies.py:29-31` rather than 29-30.

## Compatibility

Optional is the whole argument. Neither parameter carries a default, a pattern, or a required flag, so a caller that omits both gets what it got before:

- the `backend/openapi.json` diff is 78 insertions with zero deletions
- `openapi-python-client` emits them as keyword-only arguments defaulting to `UNSET`, and sends no header when unset
- the frontend types move from `header?: never` to an optional `header?: { ... }`

`TestProgrammaticCallersUnchanged` in `backend/tests/test_auth_refresh_cookies_1302.py` is the existing behavioral guard for the no-header path, and it still passes. Four new cases in the same file cover the risk this PR introduces, which is a declared parameter that starts validating: an unrecognised auth mode, an empty auth mode, and a stray CSRF header on the body-token refresh and logout paths all behave as they did before.

## Enforcement does not move

The handlers still do not read these values. Every cookie and CSRF decision stays in `cookies.py`, which needs the `Request` anyway to reach the paired cookies. FastAPI has no way to document a header without binding it, so an optional unread parameter is the declaration. Two things keep it from drifting away from the enforcement:

- each `alias` is the `cookies.py` constant itself, so renaming a header cannot leave the spec stale
- `test_cookie_session_operations_declare_their_negotiation_headers` derives the expected header set per operation from the helper calls each handler actually makes, by AST. Declare a header no handler reads, or start enforcing CSRF somewhere without declaring it, and it fails.

I checked that test is not vacuous: with the logout declaration removed it fails with `assert set() == {'X-CSRF-Token'}`.

## Regenerated in lockstep (REL-04)

- `backend/openapi.json`
- `sdks/python/geolens/api/auth/{login,logout,refresh}_*.py`
- `sdks/typescript/src/client/types.gen.ts`
- `frontend/src/types/api.generated.ts`

The vendored snapshot in getgeolens.com is untouched, per the issue: it syncs at release time and has to byte-match a shipped release.

No schema, database, or config impact. The only runtime surface is three additional optional request parameters.

## Verification

| Command | Result |
| --- | --- |
| `pytest tests/test_api_contract_openapi.py tests/test_auth_refresh_cookies_1302.py tests/test_layering.py` | 106 passed |
| `pytest tests/test_auth.py tests/test_auth_refresh_logout.py tests/test_auth_revocation_horizon_1455.py tests/test_api_key_auth.py tests/test_openapi_property_order.py tests/test_openapi_overlay_boundary.py` | 98 passed, 1 skipped |
| `ruff check .` / `ruff format --check .` | passed, 1001 files already formatted |
| `make openapi-check` | exit 0 |
| `make sdks-check` | exit 0 |
| `make cli-test` | exit 0 (318 CLI + 8 round-trip) |
| `npm run typecheck` / `npm run types:check` | exit 0 |
| `npx vitest run src/lib/__tests__/auth-transport.test.ts src/api/__tests__/refresh-cookie-1302.test.ts` | 27 passed |

## Unrelated finding, left alone

The new test resolves routes through `_iter_api_routes(app)` rather than `app.routes`, because fastapi 0.140 keeps routes included from a router nested: a plain `app.routes` scan sees 133 of the app's 486 API routes and none of the 32 under `/auth`. `test_direct_route_http_exceptions_are_documented`, already in the same file, still scans `app.routes` and so silently covers about a quarter of the API. Out of scope here, worth its own issue.

Closes #1496

### DIFF
--- a/backend/app/modules/auth/router.py
+++ b/backend/app/modules/auth/router.py
@@ -3,12 +3,26 @@
 import secrets
 import uuid
 
-from fastapi import APIRouter, Depends, HTTPException, Query, Request, Response, status
+from fastapi import (
+    APIRouter,
+    Depends,
+    Header,
+    HTTPException,
+    Query,
+    Request,
+    Response,
+    status,
+)
 from fastapi.security import OAuth2PasswordRequestForm
 from sqlalchemy import func, select
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from app.modules.auth.cookies import (
+    AUTH_MODE_HEADER,
+    COOKIE_AUTH_MODE,
+    CSRF_COOKIE_NAME,
+    CSRF_HEADER_NAME,
+    REFRESH_COOKIE_NAME,
     clear_browser_session,
     enforce_csrf,
     issue_browser_session,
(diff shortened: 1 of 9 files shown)


ishiland:
@codex review

Issue #1486, opened 2026-08-14
fix(builder): rendered map images carry no attribution — not the basemap's, not the datasets'


ishiland:
Split out of #1477 review (rounds 2 and 5), where it was raised twice against the dataset-attribution feature. It is a real gap, and it predates that feature: **no image this product renders carries any attribution at all, including the basemap's**, which is a licensing obligation of its own.

## The three paths

**PNG export** — `handleExportPNG` in `frontend/src/components/builder/hooks/use-builder-save.ts:829`. Copies `map.getCanvas()` into an offscreen canvas and composites title, legend, and — only when `showBranding` is set — a "Powered by GeoLens" footer (lines 968-975). No attribution line is drawn.

**Thumbnail and OG image** — `doCapture` in the same file (lines 76-110). Worse: it does no compositing at all. Both targets are a straight `cropResize(srcCanvas, …)` of the WebGL canvas into a 400×250 JPEG and a 1200×630 JPEG. DOM controls are not part of the WebGL canvas, so the MapLibre attribution control is absent by construction.

## Why it matters

All three are distributed outside the application. The OG image is served to social-media crawlers as the map's public preview, the thumbnail is shown on the browse gallery, and an exported PNG goes wherever the user puts it. Each renders basemap tiles — OpenStreetMap, CARTO, or whatever the operator configured, every one of which requires visible credit — and, since #1477, may also render dataset tiles whose source terms require credit (swisstopo swissALTI3D, NOAA ETOPO).

The basemap half is the older and broader problem: it has been true for every image the product has ever produced, for every deployment, regardless of whether any dataset carries attribution.

## Why it was not fixed in #1477

That PR wires dataset attribution through MapLibre's attribution mechanism, which is a DOM control. Compositing credits into a raster image is a different mechanism with its own decisions, and doing only the dataset half would leave the images non-compliant anyway while implying they were fixed:

- **What text.** The full attribution line is basemap credit + distinct visible-layer credits. It has to be assembled from both, not just the layer list.
- **Where it goes, at three sizes.** A 400×250 thumbnail has little room for a credit line that stays legible; 1200×630 and the full-resolution export have different budgets. Truncation and wrapping rules are needed.
- **Legibility over arbitrary pixels.** The line sits on top of map imagery, so it needs a scrim or halo to stay readable on both light and dark tiles.
- **i18n.** Any separator or ellipsis affix needs keys in all four locales.
[...]

PR #1499, opened 2026-08-15
fix(settings): correct misleading basemap attribution copy

ishiland:
## Summary

`SettingsMapTab.tsx` labeled the custom-basemap `attribution` field's help text as "Optional." That's wrong for the common case: most tile providers (OpenStreetMap, CARTO, OpenFreeMap/OpenMapTiles) require visible credit as a licence condition, so leaving it blank on a custom XYZ basemap produces a map with no credit, not a cosmetic gap.

There's a real asymmetry, though, and the new copy reflects it rather than overstating:

- `toMaplibreStyle` (`frontend/src/lib/basemap-utils.ts:194`) returns style JSON URLs (`.json` suffix or containing `/styles/`) unchanged. MapLibre reads that remote style's own per-source attribution, so the stored `attribution` field is **not rendered** for those, and leaving it blank is genuinely harmless.
- For a plain XYZ tile URL, the field is threaded onto the raster source and **is** what gets rendered, so a blank field there means genuinely uncredited tiles.

New help text (`settings.basemaps.attributionHelp`):

> Required by most tile providers; not needed for style JSON URLs, which include their own attribution. HTML allowed for links.

This is copy-only. No validation was added — the field is still optional to save, and runtime attribution behavior is untouched (that's #1486's territory, which surfaced this in a design review: once rendered map exports carry attribution, the same blank field would also produce uncredited PNG exports/thumbnails/OG images).

## Locales

Updated all four (`frontend/src/i18n/locales/{en,es,fr,de}/admin.json`) with real translations, not English copies. Confirmed via `check-i18n-source-keys.mjs` that no key was added or removed — only the existing `attributionHelp` value changed, so the locale-parity gate isn't exercised differently, but I translated real prose in each anyway.

Note for later: es/fr/de's `basemaps.customDescription` and `urlError` text is already stale relative to en (missing the "or MapLibre GL style JSON URLs" / "/styles/" language that en has). Pre-existing drift, out of scope here — flagging in case it's worth a follow-up.

## Proposal (not built)

If a stronger nudge is wanted later — e.g. a visible warning when a plain XYZ URL is entered with attribution left blank — that's a UX/validation decision I'd rather leave to a separate PR rather than fold into a copy fix.

## Test plan

From `frontend/`:
- [x] `npm run lint` — clean
- [x] `npm run typecheck` — clean
- [x] `npm run test:i18n` — passes (key-parity check + resources test, 4/4 tests)
- [x] `npx vitest run src/pages/admin/__tests__/AdminSettingsPage.test.tsx` — 5/5 passing

### DIFF
--- a/frontend/src/components/admin/settings/SettingsMapTab.tsx
+++ b/frontend/src/components/admin/settings/SettingsMapTab.tsx
@@ -254,7 +254,7 @@ export function SettingsMapTab({ settings, envOnly, onSave, onReset, isSaving, s
                   value={newAttribution}
                   onChange={(e) => setNewAttribution(e.target.value)}
                 />
-                <p className="text-xs text-muted-foreground">{t('settings.basemaps.attributionHelp', "Most tile providers require credit; check your source's terms. For XYZ URLs this field is what renders. Style JSON URLs ignore it and show whatever credit the style itself declares. HTML allowed for links.")}</p>
+                <p className="text-xs text-muted-foreground">{t('settings.basemaps.attributionHelp', "Most tile providers require credit; check your source's terms. For XYZ/TMS tile URLs this field is what renders. Style JSON URLs ignore it and show whatever credit the style itself declares. HTML allowed for links.")}</p>
               </div>
               <div className="space-y-1.5 max-w-md">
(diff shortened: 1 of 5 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #205, opened 2026-06-07
docs(readme): sharper front-door positioning + refreshed screenshots

ishiland:
Refreshes the README front door per a positioning audit and reworks the screenshot set so it shows the differentiators, not just the map builder.

## Copy
- Lead the opener with the open-source / self-hosted / your-infrastructure value and a one-line "who it's for".
- Resolve the "production-grade" vs "early release" mismatch in the maturity note (now consistent with the docs' "features/APIs may change" framing).

## Screenshots — now a Find → Inspect → Build → Ask AI story
All four captured fresh from a live stack:
- **Find** — semantic search for `hydrology` (ranks subwatersheds, lakes, river networks)
- **Inspect** — dataset detail with a typed, filterable attribute table (1,473 river centerlines, 38 columns)
- **Build** — the Matterhorn as a 3D terrain mesh in the builder
- **Ask AI** — natural-language map editing adds county labels to a NY income choropleth

Removed the catalog and DEM-relief images; added the search, Matterhorn, and AI-labels images. (`-2 assets, +3 assets, 1 updated`.)

## Working examples
Both "See It in Action" snippets were broken and are now verified end-to-end against a live stack:
- Semantic search queried `rivers near mountains`, which returned **zero** results → now `hydrology`.
- The OGC Features example used a table-name slug where the path requires a dataset **UUID** (always 422'd) → now derives the collection id from the catalog, then calls the auth-free items endpoint with the bbox filter.

Translated READMEs (de/es/fr) only reference the unchanged Manhattan hero, so no broken links there.

PR #198, opened 2026-06-06
docs(readme): in-builder Manhattan hero + trim to a maintainable reference

ishiland:
## What

Refreshes the README hero and trims the file to a lean, maintainable reference that defers detail to the docs site.

### Hero
- New hero: an **in-builder Manhattan 3D skyline** with the layer style editor open on the buildings layer — 3D extrusion, data-driven graduated styling by `height_roof`, height legend. Captured live from the `scripts/seed-showcase.py` Manhattan map.
- Dropped the stale empty-world "Map Builder" screenshot; repurposed the Adirondack builder shot as the map-builder still.

### Trim (359 → 300 lines)
- Merged the overlapping **Why GeoLens? / Features** content.
- Moved the raw pgvector SQL example out to the search guide.
- Consolidated redundant docs-pointer sections.
- Condensed the Seed Data section.
- All documentation links verified against the live docs site.

### Branding
- Removed the Branding section — it referenced the **private** `geolens-io/branding` repo plus an internal asset-bump procedure. Kept a one-line brand-asset license note under License.

PR #1500, opened 2026-08-15
ci(readme): gate the committed README screenshot dimensions

ishiland:
## What

Adds `scripts/check-readme-assets.mjs`, a gate over the 15 committed images in
`.github/assets/`. It pins each one to an exact pixel size, fails if a size moves,
and fails if an image is committed without being pinned. Wired into `ci.yml` as an
always-run job feeding `ci-ok`, plus a pre-commit hook.

The `image-sanitizer` editor plugin hooks the **Read** tool and runs
`magick <file> -resize "${CLAUDE_IMAGE_MAX_DIMENSION:-1200}x...>"` in place. Shrink
only, aspect preserved, 1200px cap. So a committed image loses resolution the moment
somebody opens it to look at it, usually long after whatever produced it exited. It is
synchronous and agent-triggered, not a race.

Docs (`verify-build.sh` QUICK-03) and marketing (`scripts/verify-screenshots.mjs`) both
have a gate for this. The README shots did not, because they are captured in
getgeolens.com (`scripts/capture-readme.ts`) and committed here, so neither repo owned
the check. The gate belongs where the files are committed and where CI sees every PR.

Three of the source-panel crops are 1232px wide, so they were live-vulnerable to a
single `Read` right now. That is the immediate value.

## What the assets' history actually is

`capture-readme.ts` declares `const VP = { width: 1600, height: 900 }` for all five
targets and runs with `enforceDimensions: true`. `.github/assets/README.md` line 74
also documented 1600x900. Neither matches what is committed.

Measured with `sips`, the committed five are 1200x750 (1200x800 for `ai-labels`).
Those are 1.6 and 1.5 aspect; 1600x900 is 1.778. The sanitizer preserves aspect, so a
1600x900 master lands at 1200x**675**, never 750. These are therefore not downscales of
the current recipe.

Walking the blobs back through git confirms it:

| commit | date | search/dataset/matterhorn/ai-labels |
| --- | --- | --- |
| `47b13549` (#198) | 2026-06-06 | hand-captured, inconsistent (hero 1200x750, dataset 1280x800) |
| `4f983f81` (#205) | 2026-06-07 | **1200x675**, i.e. genuinely 1600x900 sanitized |
| `683c4cd7` (#404) | 2026-07-05 | matterhorn moves to 1200x750 |
| `ee99ba93` (#407) | 2026-07-06 | all 1200x750, ai-labels 1200x800; unchanged since |

So #205 is direct evidence of the sanitizer operating on 16:9 masters, and the current
set came later from a differently shaped capture that was then downscaled on top. They
predate the automated recipe and were shrunk as well, which is what @team-lead
suspected.

## Which path I chose

I recaptured, verified the output, and then did **not** commit it.

`npm run capture:readme` against the local stack produced all five at a clean 1600x900.
I QA'd them through throwaway copies (never `Read` on a master) and three were not
publishable:

- `geolens-dataset.png` captured a **page error**: `The requested module
  '/node_modules/.vite/deps/@tanstack_react-table.js' does not provide an export named
  'columnVisibilityFeature'`.
- `geolens-search.png` showed 9 results with "Significant Volcanic Eruptions" listed
  twice.
- The hero opened the layer editor on a junk `buildings - selected` layer with an
  unconfigured style panel, instead of the era-colored 3D buildings layer.

The cause is the dev stack's state, not the recipe, and it is two separate things.

**The dataset route is broken by a stale dependency install.** `frontend/node_modules`
is a named Docker volume, not the bind mount, and it still has
`@tanstack/react-table@8.21.3` while `frontend/package.json` declares `^9.1.2` and the
lockfile pins `9.1.2`. #1445 did that v8 to v9 migration on 2026-08-12; the volume was
populated before it and has never been reinstalled. `columnVisibilityFeature` is a v9
export, so the optimized dep cannot provide it. Reproducible across two page loads.
Source code is unaffected and current: the bind mount serves `layer-icons.tsx`
containing `gradientUnits="userSpaceOnUse"`, which #1494 introduced and which is absent
from its parent commit.

**The local catalog has duplicate seed data.** Three copies each of four datasets,
including "Significant Volcanic Eruptions", which is the duplicated row visible in the
search shot.

Both need changes to a stack other agents are actively using, so neither is mine to
make. This PR is the gate only, and it lands green:

- the five `capture:readme` targets are pinned at their committed size, with a
  `capturedAt: { width: 1600, height: 900 }` field recording the declared size
- the gate prints that gap on every run instead of silently blessing 1200x750
- recapture becomes an unblocked follow-up: update `width`/`height`, drop `capturedAt`

Pinning them at 1600x900 today would have gone red on `main` and blocked every PR. The
sequencing also happens to be right: at 1200 wide the five are currently below the
sanitizer's cap and cannot be shrunk further, so it is the recapture back to 1600x900
that makes them vulnerable again. The gate should exist before that, not after.

## Scope

All 15 images, not just the 5 script targets, and an unpinned image is a failure. The
six admin/dataset-chat shots from #659 have no capture script and are pinned at the
1200x750 that `.github/assets/README.md` documents for them. The four source-panel crops
from #1279/#1280 are the ones currently at risk.

## Non-vacuity proof

Five deliberate failures, each restored afterward. Proof 1 uses the plugin's actual
resize command on a genuinely at-risk file.

| # | Fault injected | Result |
| --- | --- | --- |
| 1 | `magick source-panel-service.png -resize "1200x1200>"` (1232x522 -> 1200x508) | exit 1, `source-panel-service.png is 1200x508 on disk; expected 1232x522` |
| 2 | `geolens-ai-labels.png` moved away | exit 1, `missing from .github/assets/` |
| 3 | new `geolens-brand-new-shot.png` added | exit 1, `committed ... but not pinned` |
| 4 | a real 1600x900 recapture dropped in without a manifest update | exit 1, `is 1600x900 on disk; expected 1200x750` |
| 5 | proof 1 repeated through `pre-commit run readme-assets` | hook fails |

Green on the correct set: `OK: all 15 README images are at their pinned size.` (exit 0).

Proof 4 matters because it makes the pin bidirectional: an unreviewed recapture cannot
land silently either.

`git diff --stat origin/main -- .github/assets/` shows only `README.md`, so no image
byte changed despite the resize experiments.

## Verification

```
npm run check:readme-assets                    # exit 0, 15 pinned
pre-commit run readme-assets --all-files       # Passed
cd backend && uv run pytest tests/test_ci_merge_queue.py -q   # 22 passed
```

`test_ci_merge_queue.py` is the one that matters for the CI wiring: `readme-assets` has
no `if:`, so it runs in the merge queue and satisfies
`test_every_required_gate_runs_in_the_queue`.

The script is stdlib-only Node with its own PNG/JPEG header parser, so the job needs no
`npm ci`, no database, no browser, and no network.

## Follow-ups, not in this PR

1. Recapture the five once the stack can produce publishable shots, then update the pins
   to 1600x900 and drop `capturedAt`.
2. The dataset detail route is broken on the shared dev stack for everyone, not just for
   screenshots, until the frontend `node_modules` volume is reinstalled against the
   current lockfile.
3. The footer version is stamped from `package.json` at Vite config load, so it reports
   whatever the version was when the container started (v1.11.1 here, from a
   2026-08-12T02:49Z start). It is not a signal for what code is being served, and it
   misled me before I checked the served module directly.
4. `npm run check:readme-locales` exists but is not wired into any workflow. Same class
   of gap, out of scope here.

### DIFF
--- a/.github/assets/README.md
+++ b/.github/assets/README.md
@@ -69,9 +69,29 @@ The three content shots (Find / Inspect / Ask AI panel) are deterministic — th
 are the ones that drift with catalog data, which is exactly what the automated
 recipe keeps in parity with the marketing site.
 
+## Keeping the committed files intact
+
+`npm run check:readme-assets` (from the repo root) pins every image in this
+folder to an exact pixel size and fails if one moves. CI runs it on every PR, and
+there is a pre-commit hook for it.
+
+It exists because these files get silently downscaled. The `image-sanitizer`
+Claude Code plugin hooks the **Read** tool and runs `magick -resize` in place on
+anything wider than 1200px, so a committed image loses resolution the moment
+somebody opens it to look at it — long after whatever produced it exited. Never
+open a master with an image-previewing tool. Measure with `sips -g pixelWidth -g
+pixelHeight`, and copy to a throwaway path first if you want to see one.
+
+Adding an image here means adding it to `scripts/check-readme-assets.mjs` too;
+the gate fails on an unpinned file rather than ignoring it.
+
 ## Capture notes
 
-- Sources are captured at 1600×900; the README renders them at `width="900"`.
+- The recipe captures at 1600×900 and the README renders at `width="900"`.
+  The five `capture:readme` shots currently committed are **not** at that size —
+  they are 1200×750 (1200×800 for `geolens-ai-labels.png`), predating the
+  automated recipe and downscaled on top of it. `check-readme-assets.mjs` records
+  the gap and pins them where they are; recapturing is a separate change.
   Optionally strip/optimize before committing (e.g. `magick in.png -strip out.png`;
   JPG at quality ~88 for terrain).
 - These images are public. Keep them free of private/draft dataset titles,
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -885,6 +885,35 @@ jobs:
       - name: Surface gate regression tests
         run: python -m unittest tests.test_check_public_surface tests.test_check_deployed_surface
 
+  # ---------- README assets ----------
+  # Gate over the committed README screenshots in .github/assets/.
+  #
+  # The image-sanitizer Claude Code plugin resizes images IN PLACE from a
+  # PreToolUse hook on the Read tool (shrink-only, 1200px cap), so any committed
+  # image wider than that silently loses resolution the first time someone opens
+  # it to look at it. docs and marketing (both in getgeolens.com) each grew a gate
+  # against this; the README shots did not, because they are produced in that repo
+  # and committed in THIS one, so neither side owned the check. Three of the
+  # source-panel crops are 1232px wide and were live-vulnerable.
+  #
+  # No path filter on purpose: the corruption edits the images, so a filtered job
+  # would still fire, but this also catches a manifest edit that quietly unpins a
+  # file. Pure stdlib Node — no DB, browser, or network. Runs on every push/PR.
+  readme-assets:
+    name: README Assets
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020  # v7.0.0
+        with:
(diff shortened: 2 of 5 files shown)


ishiland:
@codex review

PR #1501, opened 2026-08-15
feat(scripts): backfill thumbnails for maps a browser never opened

ishiland:
## The gap

A map thumbnail has exactly one producer: the builder's client-side auto-capture in `use-builder-save.ts`. It fires on first load when `hasThumbnail` is false, crops the WebGL canvas to 400x250, and PUTs it to `/maps/{id}/thumbnail/`.

There is no server-side path, and **`seed-showcase.py` creates none** — it does not mention thumbnails at all.

So a map has a thumbnail if and only if a browser has opened it. On a seeded instance that splits in a way nobody intends:

| | opened by | thumbnail |
|---|---|---|
| public showcase maps | visitors, capture runs, QA | yes |
| private maps | only the owner, who is a seeding script | **never** |

That is the demo's symptom: private maps render a grey placeholder in the gallery indefinitely. The same cause left two *public* maps blank locally until someone happened to open them during a screenshot refresh.

## What this does

Lists the maps the credential can see, opens each one missing a thumbnail, and lets the app's own code path produce it. Nothing here fabricates an image or writes the column directly, so the result is byte-identical to a human opening the map.

```bash
GEOLENS_URL=https://demo.example.com \
GEOLENS_ADMIN_USERNAME=... GEOLENS_ADMIN_PASSWORD=... \
node scripts/backfill-map-thumbnails.mjs [--dry-run]
```

Two details worth noting:

- **Completion is polled, not slept on.** The upload is fire-and-forget inside the app, so its only observable effect is `thumbnail_url` flipping non-null. A fixed wait would either be too short on a slow map or waste time on every fast one.
- **Exits non-zero if any map is still missing one**, while still reporting how many landed, so a cron or CI caller notices without the run pretending it achieved nothing.

## Verification

Non-vacuous, against a live instance:

1. `--dry-run` on a fully-populated instance → "9 maps visible, 0 without a thumbnail. Nothing to do."
2. Nulled `thumbnail_uri` on a private map (`Private Embed Demo`) directly in the database.
3. Re-ran → detected exactly that one, opened it, `Filled 1/1`.
4. Confirmed the result is real rather than a blank frame: `200 image/jpeg`, **400x250**, stddev **26.3**.

## Scope

Deliberately not fixing the deeper gap, which is that thumbnail generation depends on a browser visiting at all. Doing it server-side is a larger change and overlaps #1486's work on compositing attribution into rendered images — both concern the same three outputs (thumbnail, OG image, PNG export). This script makes the current design survivable in the meantime.

### DIFF
--- a/scripts/backfill-map-thumbnails.mjs
+++ b/scripts/backfill-map-thumbnails.mjs
@@ -1,0 +1,159 @@
+#!/usr/bin/env node
+// backfill-map-thumbnails.mjs — give seeded maps the thumbnail they never got.
+//
+// WHY THIS EXISTS
+//
+// A map thumbnail has exactly one producer: the builder's client-side
+// auto-capture (`use-builder-save.ts`), which fires on first load when
+// `hasThumbnail` is false, crops the WebGL canvas to 400x250, and PUTs it to
+// /maps/{id}/thumbnail/. There is no server-side path, and `seed-showcase.py`
+// creates none — it never mentions thumbnails at all.
+//
+// So a map has a thumbnail if and only if a browser has opened it. That splits
+// a seeded instance in a way nobody intends:
+//
+//   public showcase maps  -> opened by visitors and capture runs -> have one
+//   private maps          -> only the owner can open them, and the owner is a
+//                            seeding script -> never get one, and the gallery
+//                            renders a grey placeholder forever
+//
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review

ishiland:
@codex review

Issue #1502, opened 2026-08-15
Auto-captured map thumbnails can be blank, and one blank upload is permanent


ishiland:
Found while investigating grey placeholders on the demo's private maps. The thumbnails were not missing — two of them were **blank images**, uploaded once and kept forever.

## Evidence (demo, 2026-08-15)

| map | bytes | luminance stddev |
|---|---|---|
| Manhattan (private) | 1,359 | **0.00** — one solid colour |
| Recent Earthquakes - Heatmap source Map (private) | 1,468 | **1.88** — effectively blank |
| all seven public showcase maps | 15k–27k | 35–100 |

A 400x250 JPEG at ~1.4 KB is a flat fill; the two blank ones are the only maps whose owner is a seeding script rather than a person who would have noticed.

## Mechanism: two timeouts that capture anyway, then a gate that makes it permanent

The auto-capture path (`use-builder-save.ts`) is fail-open at every stage:

1. `waitForVisibleLayerSources` polls for the visible layers' **sources to exist** — not for tiles to paint — and its 5s deadline proceeds to capture anyway: `if (sourcesReady || Date.now() >= deadline)`.
2. `whenMapIdle` waits for `idle` with a 3s safety timeout that fires `fn()` regardless, and short-circuits immediately when `map.loaded()`.
3. `doCapture` reads whatever pixels the canvas has. Nothing checks that the frame has content — there is no variance/blank test anywhere in the file.
4. The auto-capture gate is `state.hasThumbnail !== false` plus a module-level LRU. Existence, not quality — so one blank upload permanently disqualifies the map from another attempt.

Steps 1–3 are each a reasonable liveness compromise on their own; step 4 is what turns a transient slow render into a persistent artifact. On a small VM, a map with a heavy layer (the heatmap source has ~2,300 points) can blow both deadlines on first open, and first open is the only chance the map gets.

Both affected maps also have `center=None, zoom=None` — no saved viewport — so they open at the world view with a flyTo pending, widening the not-yet-painted window.

## What does and does not already mitigate this

- **Manual save**: `handleSave` calls `captureThumbnail` unconditionally, so opening the map and saving replaces a blank thumbnail. That is the workaround, and it is how nobody hits this on curated maps.
- **#1501** (`backfill-map-thumbnails.mjs`) does NOT catch this class: it selects on `thumbnail_url == null`, and these rows had a URL. It fixes *missing*, not *blank*.
[...]

PR #1504, opened 2026-08-15
fix(builder): never persist a blank auto-captured thumbnail

ishiland:
## The bug (#1502)

The auto-capture pipeline is fail-open at every stage before the upload — the source wait proceeds on its 5s deadline, the idle wait fires on its 3s timeout, and `doCapture` reads whatever the canvas holds. On a slow first render that is a frame nothing painted. The upload then makes the failure **permanent**: the auto-capture gate is `hasThumbnail` (existence, not quality), so one blank upload disqualifies the map from every future attempt.

The demo served two such thumbnails — luminance stddev **0.00** and **1.88** — for months. The affected population is exactly the maps no human opens, which is also the population that cannot self-heal through a manual save.

## The fix

`doCapture` measures the luminance stddev of the already-small 400x250 crop and skips **both** uploads when the frame is effectively a solid fill (the OG crop comes from the same `srcCanvas`, so it is blank whenever the thumbnail is). A skipped upload leaves `hasThumbnail` false, so the next open retries naturally — the permanent failure becomes a transient one, with no new state and no changes to the debounce/LRU machinery.

**Threshold.** The observed gap is wide: blanks at 0.00 / 1.88, the flattest real thumbnail at 22.9, a bare world basemap ~24. `BLANK_LUMINANCE_STDDEV = 4` sits far from both sides.

**Fail-open.** No 2D context or unreadable pixels report "not blank", so the upload proceeds exactly as before this guard existed. "Cannot judge" must never block an upload.

## Tests

9 cases in `capture-blank-guard.test.ts`, on synthetic pixel data since jsdom has no canvas: both real-world failure shapes (solid fill, near-uniform noise at the 1.88 scale), real-content and pure-chroma variation (luminance weighting, not channel average), the empty-data NaN edge, and both fail-open paths (null context, throwing `getImageData`).

## Verification

- `npm run typecheck` / `npm run lint` — clean
- builder-hooks suite: 28 files, **398 passed** (9 new)

The live demo repair that motivated this (2026-08-15) used the same discriminator externally; this PR moves it to where the damage originates.

### DIFF
--- a/frontend/src/components/builder/hooks/__tests__/capture-blank-guard.test.ts
+++ b/frontend/src/components/builder/hooks/__tests__/capture-blank-guard.test.ts
@@ -13,12 +13,19 @@
  *  fail-open path, which must never block an upload just because pixels were
  *  unreadable.
  */
-import { describe, it, expect } from 'vitest';
+import { describe, it, expect, beforeEach } from 'vitest';
 import {
   luminanceStddev,
   isEffectivelyBlank,
+  shouldAutoCapture,
+  rearmAutoCapture,
+  __resetThumbnailDebounceForTests,
   BLANK_LUMINANCE_STDDEV,
 } from '../use-builder-save';
+
+beforeEach(() => {
+  __resetThumbnailDebounceForTests();
+});
 
 /** RGBA buffer of `n` pixels produced by `fn(index) -> [r,g,b]`. */
 function rgba(n: number, fn: (i: number) => [number, number, number]): Uint8ClampedArray {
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1505, opened 2026-08-15
fix(admin): scope embedding coverage stats to the active model

### LINKED ISSUE #1503: embedding-stats reports 100% coverage from vectors the current model cannot use
Found while hardening the marketing capture guard (getgeolens.com#101, round 5).

`AdminService.get_embedding_stats` (backend/app/modules/admin/service.py:798-802) joins `record_embeddings` on `record_id` alone:

```sql
LEFT JOIN catalog.record_embeddings AS embedding
ON embedding.record_id = visible_record.id
```

But semantic search reads only vectors matching the **current** model (`service_semantic.py:192`, `RecordEmbedding.model_name == model_name`), and the table is keyed `(record_id, model_name)` precisely because vectors from different models are incomparable — the settings guide says a model change requires **Regenerate All**.

So after an embedding model or dims change, the two disagree:

- `/admin/embedding-stats/` still reports **100% coverage** (old-model rows count)
- search's vector arm matches **nothing** (no rows for the new model)

An operator looking at the admin panel sees full coverage while semantic ranking is silently dead. Verified against a live instance: renaming all stored vectors' `model_name` left stats at 98/98 embedded while the discriminator query (one that returns 0 under keyword-only) dropped from 3 results to 0.

## Suggested fix

Scope the join to the active model, the same way `has_embeddings()` in `processing/embeddings/helpers.py` partitions its cache:

```sql
ON embedding.record_id = visible_record.id
AND embedding.model_name = :current_model
```

Then after a model swap the panel honestly reports 0% and the existing backfill/regenerate flow is the visible remedy. Optionally also surface `stale_records` (rows whose only embeddings are non-current models) so the panel can distinguish "never embedded" from "needs regeneration".

ishiland:
Closes #1503.

## The disagreement

`/admin/embedding-stats/` and semantic search read the same table two different ways.

`AdminService.get_embedding_stats` joined `catalog.record_embeddings` on `record_id` alone, so a record counted as covered if any vector existed for it. Semantic search filters on `RecordEmbedding.model_name == model_name` (`backend/app/modules/catalog/search/service_semantic.py`), and the table is keyed `(record_id, model_name)` because vectors from different models cannot be compared.

Change the embedding model and the two answers diverge. Stats keep reporting full coverage from rows search cannot touch, the vector arm matches nothing, and the panel shows a healthy progress bar over dead semantic ranking. The issue records a live check: renaming every stored vector's `model_name` left stats at 98/98 while a semantic-only discriminator query went from 3 results to 0.

## The fix

The coverage join is now scoped to the active model, resolved through the same helper `has_embeddings()` uses to partition its cache. That helper is promoted from `_resolve_embedding_model_name` to `resolve_embedding_model_name` now that admin stats is its second caller.

If the model cannot be resolved, the helper returns `"__model_unknown__"`, which matches no stored row, so coverage reports 0. That is deliberate and documented on both the helper and the service method: search's vector arm resolves the same value and degrades to FTS-only when it cannot, so zero coverage is the honest reading of that state. Over-reporting is the failure being fixed.

`EmbeddingStatsResponse` gains `stale_records`: records whose only vectors belong to other models. `embedded_records` and `missing_records` keep their meaning (`missing = total - embedded`), and `stale_records` is a subset of `missing_records`. It is the number that separates "never embedded" from "needs Regenerate All".

## Counterfactual proof

`backend/tests/test_embedding_stats_model_scope.py` seeds a record with a current-model vector, checks it counts as covered, then flips only its `model_name` and checks that coverage drops while `stale_records` rises.

I reverted the scoping predicate back to the unscoped join and ran the file. Both tests failed exactly where the bug lives:

```
E   assert 1 == 0
E    +  where 1 = EmbeddingStatsResponse(total_records=1, embedded_records=1, missing_records=0, stale_records=0, coverage_percent=100.0).embedded_records
```

100% coverage on a vector search could not use. The scoped query then restores both to passing. The three new frontend tests were checked the same way against the pre-fix component and all three failed.

## Frontend

The API fix alone would have made the admin panel worse, so the Settings AI tab changed too. `Regenerate All` used to render only when `embedded_records > 0`, which after a model swap is 0, hiding the one control that clears the superseded rows. It now renders when there are stale records as well.

`Generate Missing` renders only when `missing_records - stale_records > 0`. The non-force backfill selects on `RecordEmbedding.id IS NULL`, so it skips records that already have a row and would be a no-op on a stale-only catalog.

A warning names the state: how many vectors came from a different model and that regenerating restores coverage. New key `ai.staleEmbeddingsWarning` in en/es/fr/de.

## Regenerated artifacts

`backend/openapi.json`, `frontend/src/types/api.generated.ts`, `sdks/python/geolens/models/embedding_stats_response.py`, `sdks/typescript/src/client/types.gen.ts`. No generated file was hand-edited. The hand-maintained mirror in `frontend/src/types/api.ts` was updated to match.

## Verification

```
backend  pytest tests/test_embedding_stats_model_scope.py            2 passed
backend  pytest -k "embedding or admin_stats or hybrid_search or related_datasets or semantic"
                                                                   158 passed, 1 skipped
backend  pytest tests/test_layering.py                              47 passed
backend  ruff check . && ruff format --check .                       clean
make     openapi-check / sdks-check / cli-test                       clean
frontend npm run typecheck / lint / test:i18n / types:check          clean
frontend npx vitest run                                4960 passed (397 files)
```


ishiland:
@codex review

PR #1509, opened 2026-08-15
fix(seed): let --force repair the Matterhorn showcase

### LINKED ISSUE #1508: Matterhorn showcase lost its OSM overlay layers, and the seeder's restore path cannot rebuild them
Found during the README recapture review (#1507).

**State**: "The Matterhorn in 3D" (the seeded showcase) now carries only the terrain group and basemap (`layer_count: 1`). The Climbing routes (OSM), Route casing, and Peaks overlay layers that `seed-showcase.py` defines (~L4286-4356) are gone, and their OSM source datasets no longer exist in the catalog (searching "climbing routes" returns nothing). The map's own description still promises "Dashed red lines are real climbing routes… flags mark the named summits", so the map contradicts itself in-product.

**The restore path is broken**: `seed-showcase.py --only matterhorn --force` fails before reaching the overlay layers:

```
ERROR in [matterhorn]: 62/62 swissALTI3D manifest entries failed:
swissalti3d-2613-1090: Manifest raster updates are not supported; create a new raster dataset instead. [x62]
```

The builder re-pushes manifest entries onto the existing tile datasets, the API (correctly) refuses manifest updates, and the builder treats that as fatal instead of skipping already-ingested tiles. So on any already-seeded stack, `--force` cannot repair the Matterhorn showcase at all.

**Suggested fix**: make the matterhorn builder treat "dataset exists with identical manifest" as ok/skip (the `_get_or_ingest` contract the other builders follow), so the run proceeds to the OSM fetch and layer restoration. Then re-run it on the capture stack and demo to restore the overlays, and recapture `geolens-matterhorn-terrain.jpg` + the marketing `raster-vrt.png` with routes and peaks in frame (their alt texts were downgraded to match the overlay-less frames in the meantime; restore the richer wording when the overlays return).

Acceptance:
- `--only matterhorn --force` completes on an already-seeded stack.
- The map carries Terrain + Climbing routes + Route casing + Peaks again, matching its description.
- Counterfactual: run the fixed builder against the current broken state and watch it restore the three layers.

ishiland:
`seed-showcase.py --only matterhorn --force` could not perform the one repair it exists for. The builder re-pushed the 62-tile swissALTI3D manifest onto existing raster datasets, the API (correctly) refused with `Manifest raster updates are not supported` for every entry, and the builder aborted before the map and OSM overlay steps. That left no supported way to restore the Climbing routes / Route casing / Peaks layers after they went missing — the state the capture stack was in (#1508), with the map's own description promising overlays it did not carry.

The fix reuses the DEM mosaic and its tiles whenever they exist, `--force` included. Recreating the tiles for real would mean deleting 62 datasets and re-downloading gigabytes; `--force` keeps meaning "rebuild the map", the same way no builder re-pulls ETOPO under force.

## Verified against the exact broken state

- Before: `--only matterhorn --force` on the seeded dev stack → `62/62 swissALTI3D manifest entries failed` → builder dead before layers, map stuck at one layer.
- After: same invocation → `[reuse] existing DEM mosaic` → OSM fetch + ingest → map rebuilt with all four layers (Climbing routes, Route casing, Peaks, swissALTI3D relief), styling pass grouped them, and the map renders with the overlays its description promises.
- The one-layer duplicate map and an orphaned routes dataset from the aborted runs were cleaned up manually (the newest-wins duplicate behavior under `--force` is pre-existing and out of scope here).

The demo instance picks the repair up whenever the fixed seeder next runs against it. Recapturing the two Matterhorn screenshots with the restored overlays (and re-richening their alt text) follows in the capture repos once this lands.

Closes #1508.

### DIFF
--- a/scripts/seed-showcase.py
+++ b/scripts/seed-showcase.py
@@ -4164,7 +4164,16 @@ def build_matterhorn(api: Api, force: bool = False) -> str:
     print("\n[matterhorn] The Matterhorn (3D terrain via regional VRT mosaic)")
     by_title = api.datasets_by_title()
     vrt_title = "swissALTI3D Matterhorn DEM (2m mosaic)"
-    if not force and vrt_title in by_title:
+    # fix(#1508): the DEM mosaic and its 62 tile rasters are reused even under
+    # --force. The manifest endpoint refuses updates to existing raster
+    # datasets ("Manifest raster updates are not supported"), so a forced
+    # re-push aborted the whole builder before the map and OSM overlay steps —
+    # which made --force useless for exactly the repair it exists for
+    # (restoring the routes/peaks layers). Recreating the tiles for real would
+    # mean deleting 62 datasets and re-downloading gigabytes; --force keeps
+    # meaning "rebuild the MAP" here, like the vector builders rebuild cheap
+    # datasets but nobody re-pulls ETOPO.
+    if vrt_title in by_title:
         print("  [reuse] existing DEM mosaic")
         vrt_ds = by_title[vrt_title]
     else:


ishiland:
@codex review

PR #1510, opened 2026-08-15
fix(embeddings): scope the non-force backfill to the active model

### LINKED ISSUE #1506: Non-force embedding backfill cannot regenerate stale-model vectors
Follow-up to #1503 / PR #1505.

`get_records_without_embeddings` selects on `RecordEmbedding.id IS NULL` (`defaults_processing_port.py:254`), so the non-force backfill defines "missing" as "no embedding row of any model". After an embedding-model swap, every record still has its old-model row, so **Generate Missing is a no-op on exactly the catalog state where coverage is broken** — the stale rows #1505's `stale_records` field now surfaces.

#1505 handles this at the UI level: the Settings AI tab hides Generate Missing when `missing_records - stale_records == 0` and routes operators to Regenerate All (force mode re-embeds everything and clears stale rows). That is correct but blunt — on a large catalog with 2% stale records, Regenerate All re-embeds the other 98% for nothing, at provider-token cost.

The proper fix is to make the backfill's "missing" model-aware: select records with no row for the **active** model (`NOT EXISTS (... WHERE model_name = :active_model)`), which makes non-force mode regenerate stale rows incrementally without touching current-model coverage. That changes the semantics of an existing behavior (some deployments may rely on "missing" meaning "no row at all"), which is why it was deliberately left out of #1505.

Acceptance:
- Non-force backfill embeds records whose only vectors belong to non-active models.
- Current-model rows are untouched (no re-embed, no token spend).
- The `missing_records`/`stale_records` split from #1505 reads 0/0 after a non-force run on a stale-only catalog.
- Counterfactual test: seed one stale-model row, run non-force backfill, assert the record gains an active-model row (fails against today's `IS NULL` predicate).

ishiland:
Follow-up to #1503 / #1505.

## The disagreement

`catalog.record_embeddings` is keyed `(record_id, model_name)`, and semantic search reads only the rows matching the active model. Two places disagreed about what that means:

- `AdminService.get_embedding_stats` (fixed in #1505) counts a record as covered only when it has a row under the **active** model.
- `get_records_without_embeddings` selected on `RecordEmbedding.id IS NULL`, so it treated a record as covered when it had a row under **any** model.

After an embedding-model swap every record keeps its old-model row. The panel therefore reported 0% coverage and N stale records, while Generate Missing selected nothing and reported "no records without embeddings found". The non-force backfill was a no-op on precisely the catalog state where coverage was broken.

#1505 worked around this in the UI by hiding Generate Missing when `missing_records - stale_records <= 0` and pointing operators at Regenerate All. That is correct but expensive: on a catalog with 2% stale records, Regenerate All re-embeds the other 98% at provider-token cost.

## The change

The non-force branch now selects records with no row under the active model, using a correlated `NOT EXISTS`. The outer join went with it, since `NOT EXISTS` correlates on its own and the join only produced per-embedding duplicate `Record` rows that `.unique()` collapsed again.

A non-force run on a stale-only catalog now brings `missing_records` / `stale_records` to 0/0. Worth stating plainly: it does that by **adding** the active-model row, not by deleting the superseded one. `stale_records` is computed as `any_model_records - embedded_records`, so the new row closes the gap while the old row stays in the table. Regenerate All remains the only path that deletes superseded vectors and reclaims that storage. The stale warning in Settings now says so.

`backfill_embeddings` also passes its own `force` flag through to the port instead of hardcoding `False`. On the force path the delete has already emptied the table, so both flags select the same rows today, but routing a post-delete regenerate through the model-aware branch would make a run that already destroyed its input depend on model resolution succeeding. `force=True` asks for "every record" directly, which is what that branch has always meant.

## The sentinel decision

`resolve_embedding_model_name` returns `"__model_unknown__"` when the active model cannot be resolved. That name matches no stored row, which reads in opposite directions depending on which side of the query you are on:

- For a coverage **count** it under-reports. #1505 accepts that: search's vector arm is equally unusable in that state, and over-reporting coverage was the bug being fixed.
- For a "which records still need work" **select** it over-reports. Scoping by the sentinel would mark the entire catalog missing.

This PR fails closed instead: an unresolvable model returns an empty list and logs `backfill_skipped_unresolved_embedding_model`. The reason is what happens downstream. `backfill.py` stamps its rows from a separate `EMBEDDING_MODEL.get(session)` call, and `RecordEmbedding.model_name` is `nullable=False`. So a run that selected the whole catalog in that state would embed every record at provider-token cost and then fail every insert, either on the NOT NULL constraint or, if config resolution recovered mid-run, on `uq_record_embedding_model` for the records that already had a current row. Unbounded token spend on a transient config read failure is the exact cost #1506 exists to avoid.

Selecting nothing is the recoverable error. The operator re-runs once config resolution works, and the admin panel already reads 0% coverage in the meantime, so the state is visible rather than silent.

The sentinel is now a named constant, `helpers.UNKNOWN_EMBEDDING_MODEL`, instead of a literal repeated at three sites.

## UI

#1505's gating assumed the old backfill semantics, so it went wrong the moment the backend change landed.

- Generate Missing is gated on `missing_records > 0` (was `missing_records - stale_records > 0`).
- `staleEmbeddingsWarning` names the cheaper remedy first and no longer sends operators straight to Regenerate All. New copy written into all four locales (en/es/fr/de).

While editing the test file I also changed two pre-existing assertions from `{ name: 'Generate Missing' }` to `{ name: /Generate Missing/ }`. Testing Library treats a string `name` as an exact match, and the button's accessible name is "Generate Missing Embeddings", so those two assertions could never have failed.

## Counterfactual (backend)

Predicate reverted to `stmt.where(RecordEmbedding.id.is_(None))` with the outer join restored, everything else unchanged:

```
FF.F                                                                     [100%]
=================================== FAILURES ===================================
E   AssertionError: assert UUID('d38f7b51-0320-44bd-9531-f6497558570f') in set()
tests/test_embedding_backfill_model_scope.py:90: AssertionError
E   AssertionError: assert UUID('ef3942f8-331a-4a3f-a961-5aa57e480933') in set()
tests/test_embedding_backfill_model_scope.py:112: AssertionError
E   AssertionError: assert {UUID('c2f9db...eed6593b783')} == set()
tests/test_embedding_backfill_model_scope.py:159: AssertionError
=========================== short test summary info ============================
FAILED tests/test_embedding_backfill_model_scope.py::test_record_with_only_a_superseded_vector_reads_missing
FAILED tests/test_embedding_backfill_model_scope.py::test_a_swap_moves_a_covered_record_back_into_the_selection
FAILED tests/test_embedding_backfill_model_scope.py::test_unresolvable_model_selects_nothing
```

`test_force_still_selects_records_the_active_model_covers` passes in both directions, which is the point: force semantics are unchanged.

Restored: `4 passed in 4.86s`.

## Counterfactual (frontend)

Gate reverted to `missing_records - stale_records > 0` and the en copy reverted to "Regenerate all embeddings to restore coverage.":

```
 × offers Generate Missing when every missing record is stale 84ms
 × names Generate Missing as the remedy before Regenerate All 18ms
⎯⎯⎯⎯⎯⎯⎯ Failed Tests 2 ⎯⎯⎯⎯⎯⎯⎯
TestingLibraryElementError: Unable to find an accessible element with the role "button" and name `/Generate Missing/`
TestingLibraryElementError: Unable to find an element with the text: /Generate missing embeddings to re-embed just those records/.
 Test Files  1 failed (1)
      Tests  2 failed | 16 passed (18)
```

Restored: `18 passed (18)`.

## Gates

| Gate | Result |
| --- | --- |
| `pytest tests/test_embedding_backfill_model_scope.py` | 4 passed |
| `pytest tests/test_layering.py` | 47 passed |
| `pytest -k "embedding or backfill or semantic"` | 175 passed, 1 skipped |
| `pytest tests/test_embedding_tenant_isolation.py tests/test_embedding_backfill.py tests/test_processing_port.py` | 16 passed |
| `ruff check .` | All checks passed |
| `ruff format --check .` | 1004 files already formatted |
| `python scripts/dump_openapi.py --check` | exit 0, no drift |
| `npm run typecheck` | clean |
| `npm run lint` | clean |
| `npx vitest run SettingsAITab.test.tsx` | 18 passed |
| `npm run test:i18n` | 4 passed, all locales in parity |
| pre-commit (Rule 1, Rule 2, ruff, secrets) | all passed |

`test_layering.py`'s cap for `defaults_processing_port.py` moves 499 to 537 in this commit, with the justification comment inline.

No API schema change, so `backend/openapi.json` and the SDKs are untouched.

Closes #1506.


ishiland:
@codex review

PR #1512, opened 2026-08-15
chore(readme): Matterhorn shot with restored overlays

ishiland:
Closes the loop on #1507's review finding. The seeder repair (#1509) restored the Climbing routes / Route casing / Peaks layers on the showcase map, so the Build shot is recaptured with the labeled summits (Matterhorn 4478 m, Picco Muzio, Zmuttzähne) and the dashed-red Lion Ridge climbing routes its alt text originally promised — and the alt text gets its richer wording back. Matches the marketing-side recapture in getgeolens.com#106.

Verification: `node scripts/check-readme-assets.mjs` 15/15 at pinned size; frame eyeballed at full size.


ishiland:
@codex review

PR #1514, opened 2026-08-16
docs: link the client examples hub and place it in security scope

ishiland:
Docs only. Two commits: the README link-back, then the SECURITY scope line.

## The README never pointed at the examples hub

The only "examples" pointer in the reference table was `examples/manifests/`, which is internal manifest templates. A reader of the main README had no route to [geolens-io/geolens-examples](https://github.com/geolens-io/geolens-examples), now the public showcase.

Two placements, not three:

- **A `Client examples` row in the Reference table**, next to `Manifest examples`. The two sit together and the labels tell them apart, which the old single row could not do.
- **One paragraph at the end of `See it in action`**, after the QGIS line. That section is curl against `localhost:8080`; a reader who has just watched the API answer is exactly where wanting the same endpoints from a map library starts, and the hub answers it.

Both landed in all four locales. The reference table is maintained row-for-row across `README.md`, `README.es.md`, `README.fr.md` and `README.de.md`, so an English-only row would have been drift on the first pass.

The description names what is actually merged over there: MapLibre, Leaflet, OpenLayers, ArcGIS JS, a Python/GeoPandas analysis, and MCP. TypeScript SDK, Python SDK and embed examples are in flight and are deliberately not described.

## SECURITY scope now says where the examples repo sits

Both policies already route to security@getgeolens.com, so nothing was mis-routed. What was missing was which half lands where. The examples repo ships sample client code and nothing else, so it splits cleanly:

- A server, image, package, or demo bug found **while running an example** is in scope here.
- An unsafe pattern **in the example code itself** belongs to that repo's own policy, which takes this same address alongside GitHub private vulnerability reporting.

That mirrors what [the examples repo's SECURITY.md](https://github.com/geolens-io/geolens-examples/blob/main/SECURITY.md) already says in the other direction, so the two now agree rather than each being silent about the other.

## Verified

| Claim | How |
|---|---|
| `https://github.com/geolens-io/geolens-examples` resolves | `curl -o /dev/null -w %{http_code}` returns 200 |
| The gallery URL resolves | Same, against `https://geolens-io.github.io/geolens-examples/`, returns 200 |
| The examples repo's SECURITY.md exists and routes as described | Read from `origin/main` of that repo; it names GitHub private vulnerability reporting plus security@getgeolens.com, and it points server-side reports back here |
| The hub covers exactly what the row says | Directory listing on `origin/main`: `maplibre/`, `leaflet/`, `openlayers/`, `arcgis-js/`, `python/`, `claude-mcp/` |
| CI there re-runs the examples against the live demo weekly | `.github/workflows/verify.yml`: push, pull_request, and a Monday 06:00 UTC cron |
| `examples/manifests/` still holds the four templates its row names | Directory listing in this repo: `public-cog`, `url-source.yaml`, `s3-source.yaml`, `publication-states.yaml`, plus `first-catalog` |
| No docs-contract violation | `python scripts/check_docs_contract.py` passes |

## One wording note, left alone on purpose

`## Features` opens with "Each example above has a full guide in the docs." The examples-hub paragraph now sits above that line, and the hub is not a docs guide. The sentence stays true of the curl examples it was written for, and tightening it would mean touching four locales for a sentence that is not wrong, so I left it. Worth a second opinion if you disagree.

### DIFF
--- a/.github/SECURITY.md
+++ b/.github/SECURITY.md
@@ -36,6 +36,8 @@ This policy covers:
 - Official packages published to PyPI (`geolens`, `geolens-cli`, `geolens-mcp`) and npm (`@geolens/sdk`)
 - The public demo at demo.getgeolens.com
 
+That includes a server, image, package, or demo bug you find while running one of the client examples in [geolens-examples](https://github.com/geolens-io/geolens-examples), which ships sample client code and nothing else. An unsafe pattern in the example code itself belongs to that repo's own policy, which routes to this same address alongside GitHub private vulnerability reporting. See [its SECURITY.md](https://github.com/geolens-io/geolens-examples/blob/main/SECURITY.md).
+
 Out of scope here (report to the respective repo or contact): the marketing/docs website (getgeolens.com), and Helm/deployment packaging ([geolens-deployments](https://github.com/geolens-io/geolens-deployments)).
 
 ## Recognition
--- a/README.de.md
+++ b/README.de.md
@@ -106,6 +106,8 @@ PostGIS und pgvector teilen sich eine Datenbank. Bei aktivierter semantischer Su
 
 Direkt aus QGIS verbinden: **Layer > Add WFS / OGC API Features**, Ziel `http://localhost:8080/api/`.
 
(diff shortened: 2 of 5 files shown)


ishiland:
@codex review

ishiland:
@codex review

PR #1523, opened 2026-08-16
fix(tiles): let a genuine 404 through on the public raster-tile route

### LINKED ISSUE #1516: fix(tiles): /raster-tiles/ answers 204 for a missing dataset, so a bad id renders as a blank map
Point a map client at a raster dataset id that does not exist and it renders a blank map with no diagnostic anywhere. No error, no console message, no failed request. The tiles come back `204 No Content`, which is exactly what a correct dataset answers for a tile outside its footprint, so the client has nothing to distinguish "this dataset is not here" from "you have panned off the data".

This came up building error states for the browser examples in geolens-io/geolens-examples. The page could not report a misconfigured dataset id, because at the tile route there is nothing to report.

## Repro

Two routes serve the same raster tile. They disagree, and only on the case that matters. Against the live demo, 2026-08-15:

```
$ DEM=6f03bafa-34b3-4902-9351-40ce09a8181f
$ BOGUS=00000000-0000-4000-8000-000000000000
```

| request | `/raster-tiles/{id}/tiles/...` | `/api/tiles/raster-proxy/{id}/...` |
|---|---|---|
| real dataset, tile in footprint | 200, 89816 b | 200, 89816 b |
| real dataset, tile outside footprint | 204, 0 b | 204, 0 b |
| dataset id that does not exist | **204, 0 b** | **404, 84 b** |

```
$ curl -s -o /dev/null -w "%{http_code} %{size_download}\n" \
    "https://demo.getgeolens.com/raster-tiles/$BOGUS/tiles/12/2135/1457.png?v=1"
204 0

$ curl -s -o /dev/null -w "%{http_code} %{size_download}\n" \
    "https://demo.getgeolens.com/api/tiles/raster-proxy/$BOGUS/12/2135/1457.png"
404 84
```

The API already draws the distinction correctly. It returns 204 for a tile outside the extent and 404 for a dataset that is not there. The public path loses it, and the public path is the one every client gets: it is what `helpers.py` puts in dataset metadata, what `style_json.py` writes into saved-map styles, and what search results hand back.

## Root cause

`frontend/nginx.conf:137-138`, in the `/raster-tiles/` location:

```nginx
proxy_intercept_errors on;
error_page 404 = @empty_tile;
```

Every upstream 404 becomes an empty tile. The block's own comment says what it was for:

> a tile outside the raster's extent 404s upstream and arrives here through `error_page 404 = @empty_tile`

That was true when it was written. The API now answers 204 for an out-of-extent tile on its own, as the table above shows, so the rule no longer catches the case it was added for. What it still catches is every other 404 the handler raises, including the missing dataset and the `No raster asset` path at `backend/app/processing/tiles/router.py:1248`.

## Client side

MapLibre gives a 204 tile the same treatment as a 200: the tile is marked `loaded` and carries an image. Instrumented against the demo, panning a real dataset off its footprint produced 95 tiles of 204 with zero `error` events and zero console output. That is correct behaviour for an empty tile, and it is why nothing surfaces when the dataset is missing instead.

The only way a client can currently tell the two apart is a second call to a different API, `/api/collections/{id}`, which answers 404 for a missing dataset and 200 for a real one. `maplibre/imagery.html` in the examples repo now does exactly that, purely to have something true to say when the map is blank.

## Same class on vector tiles

An unknown table answers 404 with a body:

```
$ curl -s -o /dev/null -w "%{http_code} %{size_download}\n" \
    "https://demo.getgeolens.com/api/tiles/data.no_such_table_xyz/11/602/770.pbf"
404 84

$ curl -s -o /dev/null -w "%{http_code} %{size_download}\n" \
    "https://demo.getgeolens.com/api/tiles/data.nyc_subway_lines_mta/11/602/770.pbf"
200 20760
```

The status is right here, but it does not help a browser client: MapLibre treats a 404 tile as an empty tile too, so a typo in a table name is also a silently blank layer.

## One thing to weigh before changing it

Returning 404 for an unknown dataset id at an unauthenticated route lets a caller enumerate which dataset ids exist. The current mapping hides that, whether or not it was chosen for the reason.

It does not hide much, though. `/api/collections/{id}` already answers 404 for a missing dataset and 200 for a real one anonymously, so the same enumeration is available one endpoint over. If existence disclosure is the concern, the tile route is not where it is being protected.

## Suggested direction

Have the two routes agree, in whichever direction you prefer:

- Narrow `error_page 404 = @empty_tile` so it maps only the out-of-extent case, and let a genuine "no such dataset" through as 404. Since the API already returns 204 for out-of-extent, the rule may simply be removable, though `@empty_tile` is also what repeats the CORS header from #1464 and that has to survive.
- Or keep 204 everywhere and give clients a documented way to check, in which case the examples' current workaround is the supported pattern and should be written down rather than discovered.

Either is better than the two paths answering differently for the same request.

ishiland:
Closes #1516

## What changed

`frontend/nginx.conf` no longer maps upstream 404s on `/raster-tiles/` to an empty tile. `proxy_intercept_errors on;` and `error_page 404 = @empty_tile;` are gone, and the `@empty_tile` location goes with them, since that mapping was the only thing routing to it.

Two new test files pin the result: `backend/tests/test_nginx_raster_tile_404_1516.py` for the mapping's absence, and a rewritten `test_nginx_raster_cors_1464.py`, whose two `@empty_tile` cases were describing a mechanism that no longer exists. The invariant they protected (CORS on a non-200 tile) now rests on the `always` in the `/raster-tiles/` block, which is what the rewritten file asserts.

## The premise, verified locally

The issue's claim that the api answers 204 for an out-of-extent tile was dated 2026-08-15 against the demo. Confirmed against a local stack, using a private swissALTI3D DEM whose footprint is about 1 km across, so out-of-extent is a real tile rather than a hypothetical:

```
### API directly (127.0.0.1:8001/tiles/raster-proxy/)
DEM, tile INSIDE footprint (z16)               200 59609b
DEM, tile OUTSIDE footprint (z14)              204 0b
dataset id that does not exist                 404 84b
private DEM, ANONYMOUS                         401 93b
```

So `raster_tile_proxy` converts Titiler's out-of-bounds 404 into a 204 on its own (`processing/tiles/router.py:1379`), and the `error_page` rule had stopped catching the case it was added for. Nothing regresses by removing it, because nothing out-of-extent was reaching it.

## Before and after, through nginx

A throwaway `nginx:alpine` container renders `frontend/nginx.conf` through the same four `envsubst` variables `frontend/docker-entrypoint.sh` uses and attaches to `geolens_default`, so the upstream is the real dev api. The dev stack on :8080 runs Vite, not this config, and would not have shown any of it.

| request | api route | `/raster-tiles/` before | `/raster-tiles/` after |
|---|---|---|---|
| real dataset, tile in footprint | 200 59609b | 200 59609b | 200 59609b |
| real dataset, tile outside footprint | 204 0b | 204 0b | 204 0b |
| **dataset id that does not exist** | **404 84b** | **204 0b** | **404 84b** |
| real dataset, private, anonymous | 401 93b | 401 93b | 401 93b |

The Vite dev proxy has no equivalent rule, so `localhost:8080` already answered `404 84b` for a missing dataset. Production nginx was the only one of the three disagreeing.

## The two traps

**CORS survives on the out-of-extent path.** It was never coming from `@empty_tile` in the first place: an out-of-extent tile is a pass-through 204 from the api, so it picks up the `/raster-tiles/` block's own `add_header ... always`. Measured with `Origin: https://example.com`, before and after the change, all three outcomes carry the wildcard:

```
in-footprint       status=200  Access-Control-Allow-Origin: *
out-of-extent      status=204  Access-Control-Allow-Origin: *
missing-dataset    status=404  Access-Control-Allow-Origin: *      (was 204, also with the header)
```

The security headers `@empty_tile` used to re-declare also survive, from the api's `SecurityHeadersMiddleware`:

```
HTTP/1.1 404 Not Found
x-content-type-options: nosniff
referrer-policy: strict-origin-when-cross-origin
x-frame-options: DENY
Access-Control-Allow-Origin: *
```

`X-Frame-Options` is `DENY` here where `@empty_tile` said `SAMEORIGIN`. That is stricter, and a tile image is not a framing target.

**Nothing floods.** A 404 tile is a new response for these clients, so I checked what MapLibre does with one. A real raster source pointed at a missing dataset, same page, before and after:

| | tile requests | statuses | maplibre `error` events | console errors |
|---|---|---|---|---|
| before | 16 | 204 x16 | 0 | 0 |
| after | 30 | 404 x30 | 0 | 30 |

The 30 is one retry per visible tile and it is bounded: 30 at 6 s, 30 at 12 s, 30 at 30 s. A real dataset still loads 16 tiles, all 200, with no errors.

Worth knowing for the client side that prompted the issue: MapLibre does **not** raise a `map.on('error')` event for a 404 raster tile, though the browser logs the failed request. It does raise one for a 401 (16 tiles, 16 `AJAXError: Unauthorized` events, measured on the same page). So a page listening on the map's error channel still will not hear about a missing dataset, but the request is now visible in devtools and readable by anything that inspects the response, which was not true before.

## Dataset-id enumeration

The issue's argument holds. Anonymously, against the local stack:

```
collections: real dataset (public), anon        200 729b
collections: bogus id, anon                     404 126b
collections: PRIVATE dataset, anon              404 126b
```

`/api/collections/{id}` already distinguishes a real public dataset from one that does not exist, so the tile route is not where that is being protected. It answers 404 for a private dataset as well, so it does not disclose private ids.

One correction to the issue's framing. It says the current mapping hides existence "whether or not it was chosen for the reason". It mostly does not: on the tile route today a private dataset answers 401 while a nonexistent one answers 204, so existence is already distinguishable for exactly the ids most worth protecting. What this PR changes is that a public dataset is now distinguishable from a nonexistent one there too, which `/api/collections/{id}` has always been.

## Not in scope

Vector tiles have the same client-side symptom, and the issue says so. `/api/tiles/data.no_such_table/...` already returns the right status, so there is no nginx rule to remove and nothing here to change. Left alone deliberately.

## Verification

```
cd frontend && npm run build           # exit 0
cd frontend && npm run lint            # exit 0
cd frontend && npm run typecheck       # exit 0
cd frontend && npm run test:coverage   # exit 0, 398 files, 4980 tests
cd backend && uv run pytest tests/test_nginx_raster_tile_404_1516.py tests/test_nginx_raster_cors_1464.py   # 7 passed
cd backend && uv run pytest tests/test_csp_sec020.py tests/test_docker_audit_regressions.py \
  tests/test_nginx_apikey_log_scrub_sec006.py tests/test_phase_272_dockerfile.py \
  tests/test_nginx_raster_cors_1464.py tests/test_nginx_raster_tile_404_1516.py   # 39 passed
cd backend && uv run pytest tests/test_embed_frame_policy.py tests/test_raster_tile_url_version_1372.py   # 14 passed
```

`nginx -t` passes on the rendered config, and the container serving it produced every number above. The three new tests in `test_nginx_raster_tile_404_1516.py` were run against the pre-change `nginx.conf` and all three fail there, so they are not vacuous.

No Playwright suite was run. This worktree's specs execute against the main checkout's stack, which does not serve this config. The browser numbers come from a standalone Chromium script pointed at the harness container.

No backend `app/` code changed, so `test_layering.py` does not apply.

### DIFF
--- a/backend/tests/test_nginx_raster_cors_1464.py
+++ b/backend/tests/test_nginx_raster_cors_1464.py
@@ -4,12 +4,19 @@ Raster tiles are served from the app origin at ``/raster-tiles/...`` and never
 pass through the api's CORS handling, so the header has to come from nginx.
 Two properties matter and neither is visible from a single grep:
 
-* the header must reach the ``@empty_tile`` fallback as well, because
-  ``error_page 404 = @empty_tile`` internal-redirects and nginx then emits only
-  the FINAL location's ``add_header`` set; and
+* the header must be declared ``always``, because the outcomes that need it
+  most are not 200s. Out-of-extent tiles are a 204 from the api, a missing
+  dataset is a 404 (#1516), and the ``limit_req`` ceiling is a 503; a plain
+  ``add_header`` would skip every one of them and a sparse raster would fail
+  CORS on exactly its edge tiles.
 * the value must stay the static ``*``, because the route sits behind the
   ``raster_cache`` proxy cache (and a CDN tile cache in the demo), where a
   reflected ``$http_origin`` would be replayed to the wrong origin.
+
+Until #1516 the 204 came from an internal redirect to a ``@empty_tile``
+location, which had to repeat the header because nginx emits only the FINAL
+location's ``add_header`` set. That redirect is gone: the api answers 204
+itself, so the header now reaches every outcome from this one block.
 """
 
 import re
@@ -19,7 +26,6 @@ from tests.repo_paths import repo_root
 NGINX_CONF = repo_root(__file__) / "frontend" / "nginx.conf"
 
 RASTER_TILES_LOCATION = r"location\s+~\s+\^/raster-tiles/"
-EMPTY_TILE_LOCATION = r"location\s+@empty_tile\b"
 
 _ACAO_DIRECTIVE = re.compile(
     r'add_header\s+Access-Control-Allow-Origin\s+"\*"\s+always\s*;'
@@ -60,35 +66,37 @@ def test_raster_tiles_location_emits_cors_header():
     )
 
 
-def test_empty_tile_fallback_repeats_the_cors_header():
-    """Out-of-extent tiles 404 upstream and are served from @empty_tile.
+def test_raster_cors_header_is_declared_always():
+    """A plain add_header would drop the header on 204, 404 and 503.
 
-    nginx emits the add_header set of the location that finally handles the
-    request, so the header on the /raster-tiles/ block does not survive the
-    internal redirect. Without this, sparse rasters fail CORS on every edge tile.
+    Those are the statuses this route actually returns for an out-of-extent
+    tile, a missing dataset (#1516) and the rate-limit ceiling. Measured
+    against nginx serving this config: all three carry the wildcard.
     """
     conf = _without_comments(NGINX_CONF.read_text())
-    block = _location_block(conf, EMPTY_TILE_LOCATION)
+    block = _location_block(conf, RASTER_TILES_LOCATION)
 
-    assert _ACAO_DIRECTIVE.search(block), (
-        "@empty_tile must repeat the CORS header; add_header sets are not "
-        "inherited across the error_page internal redirect"
+    plain = re.search(
+        r'add_header\s+Access-Control-Allow-Origin\s+"\*"\s*;',
+        block,
     )
+    assert plain is None, (
+        "Access-Control-Allow-Origin must be declared `always`; without it "
+        "nginx emits the header on 2xx/3xx only and every empty, missing or "
+        "rate-limited tile fails CORS in the browser"
+    )
+    assert _ACAO_DIRECTIVE.search(block)
 
 
 def test_raster_cors_header_is_static_never_reflected():
     """A reflected origin is cache-poisonous behind raster_cache and the CDN.
 
-    Scoped to the two raster locations on purpose. The cache-safety invariant
+    Scoped to the raster location on purpose. The cache-safety invariant
     belongs to this cached route, and an uncached location elsewhere may have
     good reason to answer a specific origin.
     """
     conf = _without_comments(NGINX_CONF.read_text())
-    values = [
-        value
-        for pattern in (RASTER_TILES_LOCATION, EMPTY_TILE_LOCATION)
-        for value in _ANY_ACAO_VALUE.findall(_location_block(conf, pattern))
-    ]
+    values = _ANY_ACAO_VALUE.findall(_location_block(conf, RASTER_TILES_LOCATION))
 
     assert values, "expected at least one Access-Control-Allow-Origin header"
     for value in values:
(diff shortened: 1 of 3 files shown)


ishiland:
@codex review

PR #1547, opened 2026-08-17
fix(settings): evict a settings batch's cache keys as one step

### LINKED ISSUE #1543: PersistentConfig evicts cache keys one at a time, so paired settings have a mismatch window after commit
`PersistentConfig` invalidates one cache key at a time, so any two settings that must agree have a window where a reader sees one new value and one old one.

`update_settings` commits the whole batch in one transaction, then loops:

```python
for key, value in validated_settings.items():
    await registry_map[key].apply_side_effects(value)   # -> await cache.delete(f"config:{key}")
```

Each `apply_side_effects` awaits its own `cache.delete`. A reader landing between two of them gets the new value for the evicted key (cache miss, DB read) beside the old value for a key not yet evicted (cache hit). `embedding_model` / `embedding_dims` is the pair where that is known to matter, but nothing about the mechanism is specific to them.

Found by review on #1538, which made the *database* write atomic for that pair. The database half is fixed; this is the cache half.

## Measured

Instrumented `cache.delete` to run a reader on a separate session immediately after each eviction, with both entries warmed to the old pair first:

```
cache provider = InMemoryCacheProvider
warmed cache: model='cache-win-a' dims=1536
PUT /settings/ {"embedding_model": "cache-win-b"}  -> 200

after delete(config:embedding_model): model='cache-win-b' dims=1536   <-- MISMATCHED PAIR
after delete(config:embedding_dims):  model='cache-win-b' dims=768
```

Reachable, and it is the same mismatched pair #1529 was about, one layer up.

**Window width, and it amplifies.** `embedding_dims` is evicted *last* on the auto-detect path, because `update_settings` appends the detected width to `validated_settings` after the validation loop. Every other key in the same PUT therefore adds one eviction to the window:

| request | evictions between the two | window (in-memory cache) |
|---|---|---|
| `embedding_model` alone | 0 | **1.8 us** |
| `embedding_model` + 4 other settings | 4 | **40.0 us** |

Those numbers are a floor. `InMemoryCacheProvider.delete` is a dict pop; with `REDIS_URL` set, every one of those is a Valkey round-trip, so a realistic production window is roughly one round-trip for a model-only save and N round-trips for an N-key save. Sub-millisecond to a few milliseconds, versus the 30s+ provider call that made the original #1529 window dangerous.

## Why ordering alone cannot fix it

Swapping the eviction order mismatches the other way: the reader gets the old model beside the new width instead. Both are pairs that never existed in any configuration, and both are rejected by the provider. There is no order of two independent evictions that has no gap between them.

Evicting *before* the commit is also wrong, and worse: a concurrent reader repopulates the cache with the pre-commit value, which is the hazard fix #430 codex r3 documented when it moved side effects to after the commit.

## Real options, with costs

1. **`delete_many` on `CacheProvider`.** Valkey's `DEL` is variadic and atomic, so `DEL config:embedding_model config:embedding_dims` in one command has no gap. Cheapest complete fix, but it touches the `CacheProvider` protocol, both implementations (`redis.py`, `memory.py`), the tenant wrapper, and splits `apply_side_effects` (whose per-key `_on_change` hooks and `_update_sync_cache` warm still have to run individually). Note `delete_pattern` is **not** a shortcut: `redis.py` implements it as `scan_iter` plus a per-key `delete`, so it is both non-atomic and wider than what we have now.
2. **One cache entry holding the pair.** Removes the question rather than answering it, but changes the key model and every reader of either value.
3. **Generation/version stamp that readers check.** Correct for any number of keys, and the largest change: it touches `PersistentConfig.get()` for every key in the registry.

All three are changes to `PersistentConfig` or the cache contract generally, affecting every setting rather than these two, which is why #1538 does not carry them.

A cheap interim that is **not** a fix: evict the embedding pair adjacently (or first) in the `update_settings` loop. That removes the amplification in the table above but leaves the one-round-trip gap. Recorded here as a decision, not a recommendation, because a partial mitigation in the generic loop reads as "handled" when it is not.

## Severity, honestly

Lower than the bug it descends from, and the difference is the reason #1529 was worth an atomic publish while this can wait:

- **Transient, not durable.** The original window was a committed state that lasted for a provider call and returned the same mismatched pair on every read. This one closes microseconds later when the pending `delete` lands, and self-heals without intervention.
- **Loud, not silent.** A provider handed model B with model A's width rejects the call. The result is a failed request that succeeds on retry, not a wrong answer.
- **The destructive path is already guarded.** `backfill_embeddings(force=True)` resolves the pair once via `_snapshot_embedding_config` and then proves a real vector with `_preflight_embedding` *before* its DELETE (#1519). A force backfill that reads a mismatched pair in this window aborts with nothing deleted, so the "full coverage becomes zero coverage" outcome is covered while this issue is open.

The one genuinely silent consequence: `generate_and_store_embedding` catches provider errors and returns `False`, so a record ingested precisely inside the window gets no vector until the next backfill. Self-correcting, but it does not announce itself.

## Not just the embedding pair

Any two settings a reader combines have this window. The embedding pair is where it was noticed because those two values are consumed together on every embedding call. Worth a sweep for other pairs when this is picked up, rather than fixing only the one that was reported.

ishiland:
Closes #1543.

## The window

`update_settings` commits the whole batch, then invalidates each key's cache entry one at a time. Between the first and last eviction a reader resolves the already-evicted keys from the committed row and the rest from the still-warm cache, so it sees a pair of settings that was never committed together. Measured at 1.8µs for an in-memory model-only save and 40µs at four keys in one PUT; on Valkey every eviction is a network round-trip, so it is materially wider there.

The embedding model/dimensions pair is what surfaced it, but the window belongs to `PersistentConfig`, so the tests use an unrelated pair to say so.

## What changed

A variadic `delete_many` on the `CacheProvider` contract, with the batch writers going through a new `apply_side_effects_batch`. Both provider paths are made atomic by the same change, which is what made this the smallest option that works:

- Valkey: `DEL k1 k2 ...` is one command, a single round-trip no client can observe half-applied. `delete_pattern` is not a shortcut here. It is `scan_iter` plus per-key deletes, so it is non-atomic *and* wider than the caller asked for.
- In memory: the pops run with no await between them, so nothing can be scheduled mid-batch.

The process-local half of `apply_side_effects` (the `_on_change` hook, the sync rate-limit warm, the public-URL memo clear) moved into a synchronous `_apply_local_side_effects` for the same reason: no await, no interleaving.

Four call sites: `PUT /settings`, `POST /settings/reset`, config import, and the embedding rebuild-failure rollback. That last one arrived with #1538 while this branch was open; it commits both halves of the pair in one transaction and then evicted their two keys in sequence, putting the mismatched pair back into readable state on the way out of a rollback that exists to prevent it.

## Two fixes that look right and are not

Reordering the deletes mismatches the other way.

Evicting before the commit is worse. A concurrent reader repopulates the cache with the pre-commit value, and that entry then lives its full 30s TTL instead of being transient.

## Scope, and what is still open

This makes the **writer** atomic. It does not make a **reader** atomic. A caller resolving two keys with two `get()` calls samples at two different instants and can straddle the whole commit-then-evict step, seeing one key's new value beside another's old one. That residual is reachable on Valkey, where every `cache.get` yields, and it is why `generate_embeddings`, `resolve_runtime_config` and `probe.py` all still read the pair with two separate awaits. No writer-side change can close it; the reader has to resolve the set in one operation.

### Which half of #1539's endpoint residue this closes

#1539 makes the backfill gate read the embedding model and dimensions uncached and leaves the endpoint cached, because the only path to the endpoint is `EmbeddingProviderExtension.resolve_runtime_config` (`backend/app/platform/extensions/protocols.py:371`), which has no uncached variant. The shipped implementation reads `EMBEDDING_BASE_URL` and `OPENAI_BASE_URL` through `PersistentConfig.get` (`backend/app/platform/extensions/defaults_ai_openai.py:435-438`). #1543 is that residue's named owner, so to be exact about the split:

**Closed here: the eviction span.** A PUT changing `embedding_base_url` alongside `embedding_model` used to evict the two keys in sequence. For the length of that span the gate paired an uncached (therefore already-new) model with a base URL still served from a cache entry the writer had committed past. That span is now zero. At any instant the endpoint entry is either present and pre-commit, or absent.

**Not closed here, and not closable by an eviction fix.** The dominant exposure is the 30s `_CACHE_TTL`, which is not an eviction problem. An uncached read is always current; a cached one may lag it by a full TTL however atomically the entry was evicted, because the next reader is free to repopulate. The stale-repopulation race behaves the same way: a reader whose DB read landed pre-commit writes that value back after the eviction, and it then lives its full TTL. Only reading the endpoint uncached closes either, and that needs `EmbeddingProviderExtension` widened. That is a protocol change and belongs in its own issue, not here.

`_capture_embedding_config` in `backend/app/processing/embeddings/backfill.py` defends against the general reader straddle per-caller, by reading the pair twice and comparing. `get_uncached` is the other per-key opt-out. #1539's uncached reads are untouched and still needed.

## Tests

`backend/tests/test_persistent_config_batch_eviction_1543.py`. The window is microseconds wide, so racing for it is a coin flip; the tests drive it instead. A probe cache wraps the live provider and suspends the writer inside the eviction, which is where a real Valkey round-trip suspends it, then reads the pair at that exact point through the real `PersistentConfig.get` on a separate session.

Restoring the per-key eviction turns both endpoint tests red:

```
E  AssertionError: a reader inside the eviction saw a pair that was never committed
   together: [('anthropic', 'claude-old-model'), ('openai_compatible', 'claude-old-model')]
E  AssertionError: a reader inside the reset's eviction saw a mix of overridden and
   default values: [('anthropic', 'claude-old-model'), ('openai_compatible', 'claude-old-model')]
```

Each test also asserts the probe actually ran (otherwise it proves nothing) and that the new pair is visible afterwards (otherwise "never evict anything" would pass). Four provider-level tests cover the `delete_many` contract: one variadic DEL against fakeredis, the empty-batch guard (`DEL` with no arguments is a protocol error), the circuit-open fallback, and the in-memory path driven by hand to show it never suspends.

## Verification

- `uv run pytest tests/test_layering.py -q` — 47 passed (two ratchets raised: `persistent_config.py` enters `_MODULE_LOC_CAPS` at 1007 having crossed the 1000-line inclusion threshold, `config_ops/service.py` 1201 → 1207)
- `uv run pytest tests/test_persistent_config_batch_eviction_1543.py tests/test_settings_router.py tests/test_persistent_config.py tests/test_cache.py tests/test_config_ops_unit.py tests/test_config_ops.py tests/test_settings_admin.py -q` — 189 passed
- New file run three times with random ordering and once under `-n 4`: stable

https://claude.ai/code/session_01UM7M9D4sWb23hzNeCHQnyB

### DIFF
--- a/backend/app/core/persistent_config.py
+++ b/backend/app/core/persistent_config.py
@@ -10,6 +10,7 @@ from __future__ import annotations
 import logging
 import time
 import uuid
+from collections.abc import Sequence
 from typing import Any, Generic, TypeVar, cast
 
 import structlog
@@ -269,15 +270,20 @@ class PersistentConfig(Generic[T]):
     async def apply_side_effects(self, value: T) -> None:
         """Post-commit cache invalidation + runtime hooks for a value change.
 
-        Called automatically by set()/reset() when they own the commit; callers
-        batching with ``commit=False`` invoke this per key AFTER their terminal
-        commit (fix #430 codex r3 — never before, see set() above).
+        Called automatically by set()/reset() when they own the commit. A caller
+        batching with ``commit=False`` must NOT loop over this per key — use
+        ``apply_side_effects_batch`` after its terminal commit, which evicts the
+        whole batch in one step (fix #1543) and never before the commit
+        (fix #430 codex r3, see set() above).
         """
-        # Invalidate cache
-        cache = _get_cache_safe()
-        if cache is not None:
-            await cache.delete(f"{_CACHE_PREFIX}{self.key}")
+        await apply_side_effects_batch([(self, value)])
 
+    def _apply_local_side_effects(self, value: T) -> None:
+        """The process-local half of apply_side_effects.
+
+        Deliberately synchronous: this runs once per key of a batch, and having
+        no await in it is what keeps a whole batch's worth uninterleavable.
+        """
         # BUG-025: the public-URL keys are also memoized in a separate 60s
         # module cache in public_urls. Clear it so the new value is reflected
         # immediately (PUT response, tile-config, OGC self-links, share links).
@@ -361,6 +367,51 @@ class PersistentConfig(Generic[T]):
             "basemap_proxy_rate_limit",
         ):
             _sync_rate_limit_cache[self.key] = (value, time.monotonic())
+
+
+# ---------------------------------------------------------------------------
+# Batched post-commit side effects
+# ---------------------------------------------------------------------------
+
+
+async def apply_side_effects_batch(
+    items: Sequence[tuple[PersistentConfig[Any], Any]],
+) -> None:
+    """Apply the post-commit side effects of a whole settings batch at once.
+
+    fix(#1543): the per-key loop this replaces evicted ``config:`` entries one
+    at a time. Between the first and last eviction the cache held the new value
+    for the keys already evicted (``get`` falls through to the committed row)
+    and the old value for the rest, so a concurrent reader could resolve a pair
+    of settings that was never committed together — the embedding
+    model/dimensions pair being the case that surfaced it. One ``delete_many``
+    collapses that span to nothing: a single variadic ``DEL`` on Valkey, an
+    await-free loop of pops in memory.
+
+    Ordering the deletes differently is not a fix (it mismatches the other way)
+    and evicting before the commit is worse (a concurrent reader repopulates
+    the cache with the pre-commit value, which then survives its full TTL
+    instead of being transient). Hence: commit first, then evict, as one step.
+
+    Everything after the eviction is synchronous by construction — see
+    ``_apply_local_side_effects``.
+
+    Scope, stated plainly: this makes the WRITER atomic. It does not make a
+    reader atomic. Callers that resolve two keys with two ``get`` calls take
+    their samples at two different instants and can still straddle this whole
+    step, seeing one key's committed-new value and another's cached-old one.
+    Closing that needs the reader to resolve the set in one operation;
+    ``get_uncached`` is the per-key opt-out available today (#1539).
+    """
+    if not items:
+        return
+
+    cache = _get_cache_safe()
+    if cache is not None:
+        await cache.delete_many(*(f"{_CACHE_PREFIX}{cfg.key}" for cfg, _ in items))
+
+    for cfg, value in items:
+        cfg._apply_local_side_effects(value)
 
 
 # ---------------------------------------------------------------------------
--- a/backend/app/core/persistent_config.py
+++ b/backend/app/core/persistent_config.py
@@ -402,6 +402,17 @@ async def apply_side_effects_batch(
     step, seeing one key's committed-new value and another's cached-old one.
     Closing that needs the reader to resolve the set in one operation;
     ``get_uncached`` is the per-key opt-out available today (#1539).
+
+    That distinction decides what this function does for #1539's residue. The
+    backfill gate reads model and dimensions uncached but takes its endpoint
+    from ``EmbeddingProviderExtension.resolve_runtime_config``, which has no
+    uncached variant, so the endpoint stays cached. What is fixed here is the
+    eviction span: a PUT changing ``embedding_base_url`` beside
+    ``embedding_model`` no longer serves a base URL the writer has committed
+    past. The larger exposure is untouched and is not an eviction problem —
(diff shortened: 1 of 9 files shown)


ishiland:
@codex review

ishiland:
@codex review

Issue #1532, opened 2026-08-16
fix(export): serve export ranges from a stable cached artifact


ishiland:
`GET /api/datasets/{id}/export` runs a fresh conversion on every request, including every range request. It advertises `Accept-Ranges: bytes`, so a range-probing client issues a sequence of GETs against what it believes is one representation, and each one gets a newly built artifact. While the data is unchanged that is merely wasteful. While the data changes it is incoherent.

This is not a regression from #1522. `accept-ranges`, `etag` and `last-modified` all come from starlette's `FileResponse` and are on `main` today, verified against the dev stack with none of #1522's code present:

```
$ curl -sI ".../datasets/{id}/export?format=geojson"
HTTP/1.1 200 OK
accept-ranges: bytes
content-length: 1157
last-modified: Sun, 16 Aug 2026 20:41:58 GMT
etag: "94144557a59bfe6c3b5662c945420e7c"
```

#1522 adds a HEAD that advertises the same thing, which is what surfaced it.

## The cost, from two angles

**Wasted work.** A single `/vsicurl/` open costs roughly ten conversions, because every probe and every range slice re-enters the endpoint and rebuilds the file.

**Incoherent bytes.** Those are also ten *different* artifacts whenever the data moves under them. Two range probes on `main` straddling a mutation reported `bytes 0-0/4045733` and then `bytes 0-0/2697263`: same URL, same client, two sizes.

## What is not broken

Worth stating so the fix is not over-scoped.

**The export is byte-deterministic for unchanged data.** Two consecutive conversions of the same dataset produce identical sha256:

```
sha256 #1: 5ab03d3304e46bf1c6db649a9164c8324fcb1de49781ad012917682d5b02c3c0
sha256 #2: 5ab03d3304e46bf1c6db649a9164c8324fcb1de49781ad012917682d5b02c3c0
```

So the sequence is only incoherent if the data actually changes mid-read.

**It fails loudly when it does.** Both a size-preserving UPDATE and a size-changing DELETE produced `ERROR 1: Unexpected character` / `ERROR 4: Failed to read GeoJSON data`, rc=1, and no output file. No 416, no silently truncated layer, no plausible-looking wrong answer. A same-length UPDATE still reorders the Postgres heap, so the splice lands mid-token and the parse dies rather than succeeding on mixed content.

That is the good case, and any fix must keep it. Loud failure on changed data is worth more than a fix that makes the failure quiet.

## Why the ETag route is a trap

[...]

PR #1574, opened 2026-08-17
fix(cog): honour If-None-Match: * on a COG with no stored digest

### LINKED ISSUE #1554: Honour If-None-Match: * on COG downloads for rows with no stored digest
Follow-up home for further HTTP conditional-request refinements on the COG download route, so #1540 can land rather than absorb an open-ended series.

## Where #1540 got to

Seven review rounds, every finding real. The route began answering 405 to `HEAD`, which left GDAL `/vsicurl/` unable to open a dataset without fetching it whole. It now:

- answers `HEAD` on every backend, from object metadata rather than a GET-signed redirect, so the method never has to survive a hop;
- serves byte ranges bound to the version they started reading, via a strong ETag from the persisted asset SHA, with `If-Range` evaluated server-side before an S3 redirect because S3 cannot honour it;
- honours `If-Match` and `If-None-Match`, returning 412 and 304 rather than silently serving bytes from a replacement;
- issues one object-store read per range instead of one per megabyte, on the stale-resume path and the ordinary path, both of which previously amplified a single request into thousands;
- excludes `image/tiff` from gzip, so one strong ETag can no longer name both the compressed full representation and the raw ranged one;
- accepts a case-insensitive range unit, so a conforming `Bytes=` no longer transfers the whole object;
- carries a test asserting the CORS allowlists stay in step with the conditional headers the route evaluates, because that pair drifted twice during review.

## Deferred from that PR

**Honour `If-None-Match: *` when no digest is stored.** For legacy `RasterAsset` rows where `sha256` is NULL the computed `etag` is `None`, and the precondition guard skips evaluation entirely — so a request sending `If-None-Match: *` receives 200 or 206 and may transfer a multi-gigabyte COG where 304 was correct. The wildcard matches any current representation regardless of whether the server can emit an entity tag, and a 304 does not need an ETag when the corresponding 200 would not carry one. Evaluate `*` independently of whether a digest exists.

Deferred rather than declined: it is not a regression, because the route had no conditional-request handling at all before #1540, so those rows were never served a 304. It is an incompleteness in new functionality, on rows that predate digest storage.

## Related

- #1528 — the original issue.
- #1532 — serve export ranges from a stable cached artifact. Different route, adjacent concern.
- #1552, #1553 — the same arrangement for two other PRs in this wave, filed for the same reason.

ishiland:
`RasterAsset` rows that predate the `sha256` column carry no digest, so `_cog_etag`
returns None and this route publishes no validator for them. The precondition block
#1540 landed read that as "nothing to compare against" and skipped `If-None-Match`
outright:

```python
if etag is not None and _if_none_match_matches(request.headers.get("if-none-match"), etag):
    return _cog_not_modified(etag)
```

That is the right guard for a specific tag and the wrong one for `*`. RFC 9110
§13.1.2 defines the wildcard as a question about the resource: "if the field value
is `*`, the condition is false if the origin server has a current representation
for the target resource." No clause about the server being able to *name* that
representation. A client revalidating a legacy COG with `If-None-Match: *` was
answered 200 and handed the whole object back. For a multi-gigabyte raster that is
the entire cost the header exists to save.

Three lines of code:

1. the wildcard is evaluated whatever `etag` is;
2. a specific tag against a digest-less row still gets the bytes, because a
   representation this route holds no entity-tag for matches no entity-tag, so the
   condition is true;
3. the 304 carries no ETag when the 200 it replaces would not have carried one
   either. §15.4.5 asks for the header fields the 200 would send, and `_cog_headers`
   omits the validator on these rows for its own reasons. Minting one to fill the
   slot would be worse than leaving it blank: it would authorize the resume
   `_range_bound_to_this_version` deliberately refuses here.

## The ordering matters, and it is the wrong fix that reads most naturally

`*` matches a representation that **exists**. A catalog row can outlive its stored
bytes, and the unconditional GET has always answered 404 for that, so answering the
wildcard from the row alone would tell a cache to keep serving a representation that
is gone. #1540's last round moved the stat above the precondition block for exactly
this reason (§13.2.1 puts preconditions after the normal request checks); the
wildcard stays behind it, and
`test_a_wildcard_revalidation_for_bytes_that_are_gone_is_still_404` fails if anyone
hoists it.

The unconditional 304 is licensed because this path serves GET and HEAD and nothing
else. §13.1.2 gives `*` the other answer, 412, for a method that would create or
replace a representation. `test_the_cog_download_answers_only_safe_methods` walks
the flattened route table so a future PUT reusing this block fails there before it
can inherit the wrong one.

## The other two conditionals were checked and left alone

Both were already right, so this PR adds their missing s3 coverage rather than
changing them.

**If-Match, §13.1.1.** `*` passes on a digest-less row (a representation exists), a
specific tag is 412 (this route holds no entity tag, so nothing can match, and
unverifiable is not a pass). `test_a_row_with_no_digest_refuses_a_specific_if_match_but_allows_star`
covered local storage; the s3 half is new, because the s3 pass-path is a presigned
redirect and "evaluated for local, skipped for s3" is a shape this route has had to
be corrected into twice.

**If-Range, §13.1.5.** A validator this route cannot check means ignore the Range
and serve the complete representation with 200. Never a 206, and never a 412, which
the section gives `If-Range` no license for at all: it is an optimization a server
may decline, so refusing outright would break a client that was only trying to save
bandwidth. `test_a_row_with_no_digest_refuses_a_conditional_range` covered local; on
s3 "ignore it" additionally has to mean *not redirecting*, since the client re-sends
its Range across the hop and the bucket honours it. That is asserted with
`history == []`.

## Every reader of the two values

Per the wave protocol, the values are the published validator and the object's
existence, and these are all the places either is read.

- `_cog_headers` (200, 206, HEAD) and the 416 and 412 handlers already omit the ETag
  when there is none. Unchanged.
- `_cog_not_modified` did not, and now does. It was the only reader that assumed a
  non-None validator.
- `_if_match_passes` and `_range_bound_to_this_version` already took `str | None`.
  Unchanged.
- The audit row: a 304 writes none, which is the documented rule (nothing was
  transferred, same call HEAD makes). Legacy rows now reach that path, so a wildcard
  revalidation on one stops writing a `dataset.download_cog` row. It was never a
  download, and it stops being logged as one.
- CORS: no header name is added or removed, and `test_cors_range_headers_1540.py`
  re-derives the coupling from the route's own source and passes.
- The object store never sees a 304 request. It is answered before the storage
  branching, so nothing about the bucket's own validators is involved.
- nginx adds no ETag on `/api/` and strips none of these headers (established in
  #1540, unchanged here).
- Future readers: if the digests are ever backfilled onto these rows, the wildcard
  path keeps working and the 304 simply starts carrying an ETag again. Nothing here
  depends on `sha256` staying NULL.

## Items from #1554

Done: the one deferred item, plus the If-Match and If-Range audits above.

Left out, with reasons:

- **The s3 "ignore the Range" branch proxies the whole object through the API.** Real,
  but it is #1540's design and not specific to this issue: any foreign `If-Range`
  selects it on any row, digest or no digest. The alternative is a redirect the
  bucket answers with a 206 of whatever it currently holds, which is the splice
  #1540 exists to prevent. Changing it means either serving every s3 range from this
  process or publishing the bucket's ETag, both larger than #1554.
- **Variant-specific validators and Last-Modified.** #1540 declined both on the
  record (a `Vary` story for the first, one-second granularity for the second) and
  nothing in the wildcard case changes that reasoning.
- **#1532**, cached export artifacts, is a different route.

`openapi.json` does not move. The HEAD route is still `include_in_schema=False` and
the published GET description is untouched, so neither SDK nor the CLI churns.

## Verification

`test_layering.py` raises `router_export.py` 1656 to 1686 in this commit, exact, with
the reason in the comment. Three of the 30 lines are code.

Red, with the fix reverted:

```
FAILED test_a_wildcard_revalidation_is_304_on_a_row_with_no_digest
E  AssertionError: a wildcard revalidating HEAD on a digest-less row got 200 and 0
   bytes; the object exists, so `*` matches it whether or not this route can name it.
E  assert 200 == 304

FAILED test_a_wildcard_revalidation_is_304_on_s3_without_a_digest_too
E  AssertionError: the wildcard GET was redirected to
   ['https://cog-1540.s3.amazonaws.com/rasters/8ffacfa7/src.cog.tif?AWSAccessKeyId=...'];
   the bucket does not evaluate this route's preconditions, so that redirect ends in
   the whole COG.
E  assert [<Response [302 Found]>] == []

2 failed, 6 passed
```

The other five new tests pin behaviour this PR did not change, so there is no fix to
revert for them. Each was checked against the wrong fix it exists to catch, by
mutating the source and confirming it goes red:

```
# _if_none_match_matches: `if etag is None: return True`  (unverifiable = unchanged)
FAILED test_a_wildcard_is_the_only_tag_that_matches_a_digest_less_row
E  AssertionError: If-None-Match with a foreign strong tag returned 304 for a row
   with no digest. Nothing was compared, so nothing matched, and the client needs
   the bytes.

# download_cog: answer the wildcard from the row, above the stat
FAILED test_a_wildcard_revalidation_for_bytes_that_are_gone_is_still_404
E  AssertionError: a wildcard GET answered 304 for an object that is gone.
   `*` matches a representation that exists; there is none.

# _range_bound_to_this_version: `if etag is None: return True`  (cannot check, honour it)
FAILED test_a_row_with_no_digest_refuses_a_conditional_range
E  AssertionError: a conditional range with nothing to check against returned 206;
   unverifiable must not mean honoured.
FAILED test_an_unverifiable_if_range_on_s3_is_ignored_rather_than_honoured
E  AssertionError: the unverifiable resume was redirected to [...]; the client
   re-sends its Range across a 302 and the bucket answers 206, which is the partial
   response this branch just decided not to send.

# _this_service_owns_the_bytes: `== "local"`  (preconditions skipped for s3)
FAILED test_a_row_with_no_digest_refuses_a_specific_if_match_on_s3_too
E  AssertionError: a tag this route cannot check returned 302 on s3; unverifiable
   must not mean honoured, and a 302 here hands the precondition to a bucket that
   does not evaluate it.
FAILED test_head_cog_on_s3_is_answered_from_object_metadata
FAILED test_if_none_match_answers_304_on_s3_too
FAILED test_s3_keeps_redirecting_everything_that_is_not_a_stale_resume
FAILED test_a_wildcard_revalidation_is_304_on_s3_without_a_digest_too
```

`test_the_cog_download_answers_only_safe_methods` demonstrated its own failure mode
on the first run: written against `app.routes` it reported `answers []`, because
fastapi's `include_router` is lazy. It goes through `iter_route_contexts` now, the
way `test_rule1_structural.py` does, and the assertion is written so an empty match
fails rather than passes.

Green:

```
$ cd backend && uv run pytest tests/test_cog_head_ranges_1528.py -q
82 passed in 52.37s

$ uv run pytest tests/test_cog_head_ranges_1528.py tests/test_cors_range_headers_1540.py \
    tests/test_cog_vsicurl_1528.py tests/test_storage_get_stream.py \
    tests/test_rule1_structural.py tests/test_rule2_structural.py -q
203 passed in 63.46s

$ uv run pytest tests/test_layering.py -q
47 passed in 9.13s

$ uv run ruff check . && uv run ruff format --check .
All checks passed!
1024 files already formatted

$ PYTHONPATH=. uv run python scripts/dump_openapi.py --check
EXIT=0
```

`test_cog_s3_head_wire_1540.py` still skips its four cases without a live S3 endpoint,
as designed.

Closes #1554

### DIFF
--- a/backend/app/modules/catalog/datasets/api/router_export.py
+++ b/backend/app/modules/catalog/datasets/api/router_export.py
@@ -1109,9 +1109,13 @@ async def download_cog(
                 detail="COG has changed since the version you hold",
                 headers={"ETag": etag} if etag is not None else None,
             )
-        if etag is not None and _if_none_match_matches(
-            request.headers.get("if-none-match"), etag
-        ):
+        # fix(#1554): evaluated whatever `etag` is. The old `etag is not None`
+        # guard was the right test for a specific tag and the wrong one for
+        # `*`, which asks whether a current representation exists rather than
+        # which one it is — so a legacy row with no `sha256` answered a
+        # wildcard revalidation with the whole COG. The stat above is what
+        # makes that existence question answerable here.
+        if _if_none_match_matches(request.headers.get("if-none-match"), etag):
             return _cog_not_modified(etag)
 
     # 6. Audit log. user_id may be None for anonymous downloads (KNOWN-01).
@@ -1334,7 +1338,7 @@ def _if_match_passes(if_match: str | None, etag: str | None) -> bool:
     return etag is not None and etag in candidates
 
 
-def _if_none_match_matches(if_none_match: str | None, etag: str) -> bool:
+def _if_none_match_matches(if_none_match: str | None, etag: str | None) -> bool:
     """Does the client already hold this representation? RFC 9110 section 13.1.2.
 
     WEAK comparison, unlike ``If-Range`` above, and the difference is not an
@@ -1342,15 +1346,34 @@ def _if_none_match_matches(if_none_match: str | None, etag: str) -> bool:
     equivalent, while a resumed range needs them byte-identical. The spec picks
     a different comparison function for each, so this route does too.
 
-    ``*`` matches any current representation, which by the time this is called
-    is exactly the one whose ETag was passed in. A list is a list: a client may
-    hold several versions and offer all of them.
+    ``*`` matches any current representation, and fix(#1554) is that it does so
+    even when ``etag`` is None. The section says "if the field value is '*',
+    the condition is false if the origin server has a current representation" —
+    a statement about the RESOURCE, with no clause about the server being able
+    to name it. A row predating the ``sha256`` column has bytes in storage, the
+    caller stat'd them before asking, and answering 200 there transfers a
+    multi-gigabyte COG to tell a cache what it already knew.
+
+    The wildcard is checked before ``etag`` for that reason, and the specific
+    tags after it are still refused when there is nothing to compare: a
+    representation this route holds no entity-tag for matches no entity-tag,
+    so the condition is true and the client gets the bytes it asked for.
+
+    A match is answered 304 unconditionally because this path serves GET and
+    HEAD only (``test_the_cog_download_answers_only_safe_methods`` pins that).
+    The same section makes ``*`` a 412 for a method that would create or
+    replace a representation, which is a different answer this route has no
+    caller for.
+
+    A list is a list: a client may hold several versions and offer all of them.
     """
     if not if_none_match:
         return False
     candidates = [tag.strip() for tag in if_none_match.split(",")]
     if "*" in candidates:
         return True
+    if etag is None:
+        return False
     return any(_without_weak_prefix(tag) == etag for tag in candidates)
 
 
@@ -1358,7 +1381,7 @@ def _without_weak_prefix(tag: str) -> str:
     return tag[2:] if tag.startswith("W/") else tag
 
 
-def _cog_not_modified(etag: str) -> Response:
+def _cog_not_modified(etag: str | None) -> Response:
     """304: the client's copy is current, so send it nothing else.
 
     The ETag goes back per RFC 9110 section 15.4.5, and ``Accept-Ranges``
@@ -1367,11 +1390,18 @@ def _cog_not_modified(etag: str) -> Response:
     starlette's ``init_headers`` skips the length for 304 (and 204) exactly as
     the spec requires, which is why this response does not need the explicit
     header ``_cog_head_response`` has to pass.
+
+    fix(#1554): ``etag`` may be None, and then the 304 carries none. Section
+    15.4.5 asks for the header fields a 200 would have sent, and the 200 for a
+    row with no stored digest sends no validator either (``_cog_headers`` omits
+    it for the same reason). Minting one to fill the slot would be worse than
+    the blank: it would authorize the resume ``_range_bound_to_this_version``
+    refuses on exactly these rows.
     """
-    return Response(
-        status_code=status.HTTP_304_NOT_MODIFIED,
-        headers={"ETag": etag, "Accept-Ranges": "bytes"},
-    )
+    headers = {"Accept-Ranges": "bytes"}
+    if etag is not None:
+        headers["ETag"] = etag
+    return Response(status_code=status.HTTP_304_NOT_MODIFIED, headers=headers)
 
 
 def _cog_etag(raster_asset) -> str | None:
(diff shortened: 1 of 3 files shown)


ishiland:
@codex review

Issue #1530, opened 2026-08-16
release.yml's git-log notes fallback fires silently — third recurrence of the same class


ishiland:
`.github/workflows/release.yml` extracts the release body from `CHANGELOG.md` by matching the version header:

```sh
NOTES=$(awk "/^## \[${CHANGELOG_VERSION}\]/{found=1; next} /^## \[/{if(found) exit} found{print}" CHANGELOG.md)

if [ -z "$(printf '%s' "$NOTES" | tr -d '[:space:]')" ]; then
  # Fallback: generate from git log
  NOTES=$(git log --format="- %s" "${PREV}..${TAG}" --no-merges | grep -v "^- docs(" | grep -v "^- chore(" | head -30)
fi
```

When the header is absent for any reason, `found` is never set, `NOTES` is empty, and the fallback publishes a raw commit-subject list **with no warning anywhere** — no annotation, no step summary, nothing in the job log distinguishing "curated notes" from "we could not find them."

## Why this is worth changing

This is the third recorded case of the fallback firing when it should not have:

1. **#715** — every prerelease tag published the fallback, because an RC tag has no CHANGELOG section of its own and never will.
2. **#715 review** — a section containing only whitespace made `[ -z ]` false, so a release published effectively blank notes.
3. **2026-08-16, during #1518** — a new Unreleased entry landed on top of `## [1.13.1] - 2026-08-14` and deleted the header. The Version Coherence gate caught it before merge, so nothing shipped.

Each was fixed by making the *matching* smarter. The pattern across three says the problem is the silence, not the matcher: every failure mode of the extraction looks identical to success from the outside, and is found by someone noticing rather than by the pipeline reporting it.

## The current defense, and what it does not cover

`check_version_coherence.py` validates the header set and is what caught case 3 — the defense worked. The residual gap is narrower but real:

- A release whose notes come from the fallback is indistinguishable, in the published output, from one whose notes were curated.
- The fallback's filter drops every `docs(` and `chore(` subject and caps at 30 entries, so a fallback release is quietly incomplete as well as uncurated.

## Suggested change

Keep the fallback — a failed extraction must not block a release — but make it loud:

- `::warning::` annotation naming the version whose CHANGELOG section could not be found.
- A line in the job summary saying the notes are auto-generated rather than curated.
- Optionally a marker in the release body itself, so the state is visible to readers rather than only to whoever reads the workflow log.

The goal is that "the fallback fired" becomes something the pipeline announces, instead of something discovered three times by accident.

PR #1591, opened 2026-08-18
ci(release): dispatch a geolens-release event to the site after the Release publishes

ishiland:
## Summary

CI-1: getgeolens.com already declares `repository_dispatch: [geolens-release]` for its release-drift check, and a companion PR there (geolens-io/getgeolens.com#108) adds a `release-sync.yml` that opens a docs-contract PR on that same event. Nothing on this side ever fired the event, so the site's contract, OpenAPI snapshot, and installer sync stayed a manual step after every release.

This adds one step to `release.yml`'s `release` job, after the GitHub Release and its uploads (install script, SHA256SUMS) succeed. It checks that the dispatched tag is actually the latest published release, then fires the dispatch:

```yaml
- name: Notify the site of the release
  if: ${{ !contains(env.TAG, '-') }}
  env:
    RELEASE_READ_TOKEN: ${{ secrets.GITHUB_TOKEN }}
    SITE_DISPATCH_TOKEN: ${{ secrets.SITE_DISPATCH_TOKEN }}
  continue-on-error: true
  run: |
    LATEST="$(GH_TOKEN="$RELEASE_READ_TOKEN" gh api "repos/${GITHUB_REPOSITORY}/releases/latest" --jq .tag_name)"
    if [ "$LATEST" != "$TAG" ]; then
      echo "::notice ...skipping, not the latest release..."
      exit 0
    fi
    if [ -z "$SITE_DISPATCH_TOKEN" ]; then
      echo "::notice ...SITE_DISPATCH_TOKEN not set..."
      exit 0
    fi
    GH_TOKEN="$SITE_DISPATCH_TOKEN" gh api repos/geolens-io/getgeolens.com/dispatches \
      -f event_type=geolens-release \
      -f "client_payload[version]=${TAG}"
```

A few things worth calling out:

- **A monotonicity guard.** `release.yml` has no `concurrency:` group and `prod-smoke.yml` runs tag smokes in parallel, so an older tag's smoke can finish after a newer tag's release already published, or a re-run can replay an old tag's smoke. Dispatching in either case would tell the site to sync backwards. The step reads this repo's actual latest release first and skips (with a visible `::notice::`) unless the dispatched tag matches it.
- **Only GA tags dispatch.** `!contains(env.TAG, '-')` mirrors the `case "${TAG}" in *-*)` prerelease check already used a few lines up in the "Create release" step. A prerelease never gets promoted to the "latest" release that the site's install one-liner and archive link follow, so there's nothing for the site to sync.
- **Missing secret skips, but visibly.** An earlier version of this gated the whole step's `if:` on the secret being present, since `secrets` isn't in the context list GitHub documents for `jobs.<job_id>.steps.if` (`github, needs, strategy, matrix, job, runner, env, vars, steps, inputs`, per the [contexts reference](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts)), so it went through `env` instead. The problem: a step skipped by `if:` shows as a bare grey "skipped" in the run, which is exactly how a missing token would go unnoticed. This file already hit that failure mode once, with the CHANGELOG-derived release notes fallback a bit further up (`#1530` and its history). Now the step always runs, and the missing-secret and stale-tag cases log a `::notice::` plus a `$GITHUB_STEP_SUMMARY` line from inside the script instead.
- **Two tokens, because they read different repos.** Checking this repo's latest release and dispatching to the site both need `gh api` calls, but `SITE_DISPATCH_TOKEN` has no read access here and the default token has no access on the site repo. The step takes `RELEASE_READ_TOKEN` (the default `GITHUB_TOKEN`) and `SITE_DISPATCH_TOKEN` as separate env vars and picks the right one per call.
- **A dispatch failure with the secret present warns instead of failing the job** (`continue-on-error: true` plus a `::warning::` and a step-summary line), because the release is already published by this point. Worst case, the site's own scheduled `release-drift` check catches the gap.

Also documents the secret in `RELEASE.md`'s "Cutting a release" steps, and cites the receiver PR (geolens-io/getgeolens.com#108) from the inline comment per the repo's finding-id convention.

## What Ian needs to add

Secret **`SITE_DISPATCH_TOKEN`** on this repo (geolens-io/geolens): a fine-grained PAT scoped to `Contents: read/write` on `geolens-io/getgeolens.com`, since a repository dispatch write requires contents write on the target repo.

## Verification

- `actionlint .github/workflows/release.yml`: one pre-existing SC2016 info finding on line 66 (unrelated to this change, confirmed present on `main` before this diff by stashing and re-running, and again after this revision). No new findings from the step.
- Extracted the new step's script into a standalone file with fake tokens and ran `shellcheck` on it directly: clean.
- `git diff --stat`: only `.github/workflows/release.yml` and `RELEASE.md` changed.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -185,20 +185,20 @@ jobs:
         # INSIDE the run script rather than in this `if:`, on purpose: a step
         # this `if:` skips shows as a bare grey "skipped" in the run, which is
         # how a missing SITE_DISPATCH_TOKEN would go unnoticed indefinitely
-        # (this file already learned that lesson once, see the release-notes
-        # fallback around line 90 for #1530 and its history). Running the
-        # step and having it log a ::notice::/::warning:: and a step-summary
-        # line keeps the same "never fail the release over this" behavior
-        # visible instead of silent.
+        # (this file already learned that lesson once, for #1530). Running
+        # the step and having it log a ::notice::/::warning:: and a
+        # step-summary line keeps the same "never fail the release over
+        # this" behavior visible instead of silent.
         if: ${{ !contains(env.TAG, '-') }}
         env:
           # Two different tokens for two different targets: reading THIS
-          # repo's latest release needs only the default token; dispatching
-          # to getgeolens.com needs a PAT scoped there (SITE_DISPATCH_TOKEN
-          # has no read access here, and the default token has no access
-          # there). Fine-grained PAT, Contents: read/write on
-          # geolens-io/getgeolens.com; absent = skipped with a ::notice::.
-          RELEASE_READ_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # repo's latest release needs only the default GITHUB_TOKEN;
+          # dispatching to getgeolens.com needs a PAT scoped there
+          # (SITE_DISPATCH_TOKEN has no read access here, and the default
+          # token has no access there). Fine-grained PAT, Contents:
+          # read/write on geolens-io/getgeolens.com; absent = skipped with
+          # a ::notice::.
+          DEFAULT_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           SITE_DISPATCH_TOKEN: ${{ secrets.SITE_DISPATCH_TOKEN }}
         # The release is already published by this point in the job. A failed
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

PR #1599, opened 2026-08-19
fix(settings): reject falsy non-string public URL values instead of clearing them

ishiland:
`validate_public_app_url` and `validate_public_api_url` opened with `if not v or (isinstance(v, str) and not v.strip()): return ""`. `not v` is true for any falsy value, not just an empty string, so a JSON `false`, `0`, `[]`, or `{}` sent for either key silently cleared the configured URL instead of failing validation. #1592 fixed the same bug for `privacy_url`; this applies the identical shape to the other two URL validators: `v is None` clears, a non-string falls through to a 422, and everything else keeps its existing normalization.

Refs #1592

## Verification

- Red-first: reverted the schema change, ran the new tests, and confirmed all 8 falsy-rejection cases failed with 200 instead of 422 (the null/empty-clear cases already passed, since that path was never broken). Reapplied the fix and reran; all 10 pass.
- `tests/test_settings_router.py tests/test_public_app_url_shape_contract_1548.py tests/test_branding_settings.py`: 126 passed
- `tests/test_layering.py`: 47 passed, no cap changes needed
- `make openapi-check`: no diff (validator-only change, no schema shape change)
- Checked the frontend admin form (`SettingsGeneralTab.tsx` / `useSettingsForm.ts`): both fields are plain text inputs, so `e.target.value` is always a string, sent unmodified through `handleSave` and `updateMutation.mutate`. Nothing there relied on the old falsy-clear behavior.

### DIFF
--- a/backend/app/modules/settings/schemas.py
+++ b/backend/app/modules/settings/schemas.py
@@ -453,8 +453,14 @@ def validate_public_app_url(v: Any) -> str:
     does nothing. The ``/api`` clause is checked explicitly first only to keep
     its specific message; ``canonical_host_error`` supplies one for a host.
     """
-    if not v or (isinstance(v, str) and not v.strip()):
+    # `v is None`, not `not v`: JSON null is a legitimate clear (matching
+    # validate_privacy_url below), but a falsy non-string -- False, 0, [], {}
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

Issue #1464, opened 2026-08-13
fix(tiles): /raster-tiles/ responses carry no CORS headers, blocking external map clients


ishiland:
Found 2026-08-13 while building the first verified example for the new [geolens-examples](https://github.com/geolens-io/geolens-examples) repo. Directly relevant to the ArcGIS JS client-consumption story: `WebTileLayer` hits this exact wall.

## The defect

Responses on the `/raster-tiles/{dataset_id}/tiles/{z}/{x}/{y}.png` route carry no CORS headers, so every fetch-based map client is blocked from consuming GeoLens raster tiles cross-origin. The `/api/` surface does this correctly.

Evidence against the live demo (same `Origin: http://localhost:8899` header on both):

```
GET /api/collections                      → 200, access-control-allow-origin: *
GET /raster-tiles/6f03bafa.../13/4270/2914.png?v=1   → 200 image/png, NO access-control-* headers
GET same with a cache-busting query       → cf-cache-status: MISS, still NO access-control-* headers
```

The MISS case matters: this is not Cloudflare replaying stale cached copies — the origin response chain (demo nginx → `/tiles/raster-proxy/` API route → Titiler) itself never emits the header. Reproduced end to end in a browser: a MapLibre `raster` source pointed at the correct advertised tile template fails every tile with `blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present` (108 console errors), while the vector-tile (`/api/tiles/...`) and Features (`/api/collections/.../items`) examples on the same page origin render fine.

## Why it was never noticed

The route was designed for the app's own map: `search/router.py` documents that raster tiles are served at "the public APP origin (/raster-tiles/...), not the /api origin" — same-origin consumption, where CORS never applies. And the one popular client that would have surfaced it externally, Leaflet, loads raster tiles through `<img>` tags, which bypass CORS for display — so casual "it works in Leaflet" checks mask the gap. Everything canvas/WebGL-based (MapLibre, ArcGIS Maps SDK `WebTileLayer`, OpenLayers) fetches tiles and fails.

## Fix direction, with the cache subtlety stated

[...]

PR #1610, opened 2026-08-19
fix(nginx): keep the raster-tile cache from fragmenting on Vary: Origin

### LINKED ISSUE #1605: nginx raster-tile proxy cache should ignore the upstream Vary: Origin
Follow-up to #1602 / #1603. The API now sends \`Vary: Origin\` on every response that passes through \`DynamicCORSMiddleware\`, including \`/tiles/raster-proxy/...\`. \`frontend/nginx.conf\` serves those through \`location ~ ^/raster-tiles/...\` with a real \`proxy_cache raster_cache\` and deliberately overrides the CORS value with a static \`*\` (fix(#1464)), so the cached representation there does NOT vary by origin, yet nginx honours the upstream \`Vary\` when it stores the entry and will keep one copy per distinct \`Origin\` request-header value.

Impact is bounded: same-origin tile fetches carry no \`Origin\` header at all, so the app's own maps are unaffected; only cross-origin embeds fragment, one cache copy per embedding site. Cloudflare (the demo's tile cache rule) ignores \`Vary\` except \`Accept-Encoding\`, so the demo is unaffected either way.

Remedy: in that location block add \`proxy_ignore_headers Vary;\` (and \`proxy_hide_header Vary;\` so clients are not told about a variance nginx does not serve), with a structural test beside the existing #1464 ones. \`proxy_hide_header\` alone is not enough: it hides the header from the client but nginx still keys the cache on it.

ishiland:
## Summary

`location ~ ^/raster-tiles/...` in `frontend/nginx.conf` now ignores the upstream `Vary` for caching and hides it from the client:

```nginx
proxy_ignore_headers Set-Cookie Vary;
proxy_hide_header Vary;
```

Plus `backend/tests/test_nginx_raster_cache_vary_1605.py`, a brace-aware structural test.

## Why

Since #1603 the api's CORS middleware merges `Vary: Origin` onto every response it returns, including `/tiles/raster-proxy/...`, which is what this block proxies to. nginx keys a cache entry on the upstream `Vary`, so `raster_cache` kept one copy of the same tile bytes per distinct `Origin` request header. Same-origin tile fetches send no `Origin` at all and were never affected, so the app's own maps were fine; cross-origin embeds each warmed their own set of tiles.

Nothing in this block varies by origin. It answers every caller with the static `Access-Control-Allow-Origin: *` (#1464), which is exactly why the cached copy is shareable in the first place.

`proxy_hide_header Vary` on its own is not the fix. It changes what the client is told, not what nginx stores under.

## Verified

Behaviour, against nginx 1.31.3 (the version the frontend image ships) serving a throwaway harness with this shape: a cache key that does not include the origin, in front of an upstream answering `Vary: Origin` and a body unique per request.

| config | origin A | origin B | third request |
| --- | --- | --- | --- |
| `proxy_ignore_headers Set-Cookie;` (before) | MISS | MISS | A again: HIT |
| `proxy_hide_header Vary;` only | MISS | MISS | A again: HIT, no `Vary` sent |
| both (this PR) | MISS | HIT, A's body | origin C: HIT, no `Vary` sent |

The third column is the control on the first two rows: caching works there, so B's MISS is fragmentation rather than a cache that never stored anything.

Two more nginx behaviours measured, because the comment asserts them:

- Two `proxy_ignore_headers` at the same level accumulate. `Vary` stayed ignored with a later `Set-Cookie` directive after it.
- A directive inside a `location` replaces the value inherited from `http`/`server` instead of merging with it. A location declaring only `Set-Cookie` under an http-level `Vary` went back to MISS/MISS, while a sibling location with no directive of its own kept the HIT.

So the one line is the complete set for that block, which is why the test requires a single directive listing every field.

Gates:

- `test_nginx_raster_cache_vary_1605.py` red first against the conf as it stood (5 failed, 1 passed), then green. It enumerates every location with a real `proxy_cache` rather than naming the raster route, and it has counterfactuals for each half of the fix, for splitting the directive in two, and for the block parser itself.
- `pytest tests/test_nginx_raster_cache_vary_1605.py tests/test_nginx_raster_cors_1464.py tests/test_nginx_api_slash_redirect_1597.py tests/test_nginx_raster_tile_404_1516.py tests/test_nginx_assets_cors_1515.py tests/test_nginx_apikey_log_scrub_sec006.py tests/test_csp_sec020.py tests/test_embed_frame_policy.py tests/test_raster_tile_url_version_1372.py`: 38 passed.
- `tests/test_phase_272_dockerfile.py tests/test_docker_audit_regressions.py`: 26 passed. `tests/test_layering.py`: 47 passed, no cap moved.
- `frontend/src/__tests__/nginx-range-route-gzip-guard.test.ts`, the other reader of this file: 7 passed.
- `nginx -t` on the real conf, rendered through `envsubst` with the same variable list and defaults the entrypoint uses, inside `nginxinc/nginx-unprivileged:1.31.3-alpine`. Passes for the default `TRUSTED_PROXY_CONFIG` and for the CIDR-rendered one. An invalid field added to this same directive is rejected (`invalid value "Bogus"`), so the check is live rather than vacuous. The one warning it prints (`duplicate extension "wasm"`) is present on `main` at the same line.
- `ruff check .` and `ruff format --check .` clean; pre-commit hooks pass on both changed files.

`/raster-tiles/` is the only location in the file with a cache enabled. `location /api/` sets `proxy_cache off`, so it never stores anything and needs nothing here.

AGENTS.md rules 1, 2 and 3 are untouched: no handler, no `httpx` client, no credential literals.

Closes #1605

### DIFF
--- a/backend/tests/test_nginx_raster_cache_vary_1605.py
+++ b/backend/tests/test_nginx_raster_cache_vary_1605.py
@@ -1,0 +1,216 @@
+"""The nginx raster-tile cache must not fragment on the upstream ``Vary`` (#1605).
+
+The api's CORS middleware sends ``Vary: Origin`` on every response it touches
+(#1602/#1603), and ``/raster-tiles/...`` is proxied to it. nginx honours an
+upstream ``Vary`` when it stores a cache entry (the ``Vary`` parameter of
+``proxy_ignore_headers`` exists since 1.7.7 precisely because of that), so
+``raster_cache`` keeps one copy of the same tile bytes per distinct ``Origin``
+request header — a fresh set of tiles for every site that embeds a map.
+
+Nothing about the representation varies by origin: the block answers every
+caller with the static ``Access-Control-Allow-Origin: *`` (#1464). So the
+correct shape is to ignore the header for caching AND hide it from the client,
+and both halves are load-bearing. Measured against nginx 1.31.3 serving a
+harness with this exact shape (cache key without the origin, upstream sending
+``Vary: Origin`` and a unique body):
+
+* ``proxy_ignore_headers Set-Cookie;`` only — origin A MISS, origin B MISS,
+  origin A again HIT. The third request is the control: caching works, and B's
+  MISS is fragmentation.
+* ``proxy_hide_header Vary;`` only — origin A MISS, origin B MISS, origin A
+  again HIT, and no ``Vary`` on the wire. Hiding changes what the client is
+  told, not what nginx keys on.
+* both — origin A MISS, origins B and C HIT with A's body, no ``Vary`` header.
+
+The assertions are structural (they read the conf) and brace-aware, and they
+enumerate every location with a real ``proxy_cache`` rather than naming the
+raster route, so a second cached route inherits the requirement instead of
+quietly reintroducing the bug.
+
+One directive per block, listing every field. nginx's merge rules for
+``proxy_ignore_headers`` were measured too, because they are not symmetric:
+two directives at the SAME level accumulate (``Vary`` stayed ignored with a
+later ``Set-Cookie`` directive after it), but a directive in a location
+REPLACES the value inherited from ``http``/``server`` (a location declaring
+only ``Set-Cookie`` under an http-level ``Vary`` went back to MISS/MISS, while
+a sibling location with no directive at all kept the HIT). The directive in a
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

Issue #1517, opened 2026-08-16
fix(monitoring): GeoLensApiInteractiveLatencyP95 fires on bulk-download endpoints, which are slow by design


ishiland:
`GeoLensApiInteractiveLatencyP95` has fired twice in one day on the demo VM, both times from ordinary example-repo verification rather than from the demo being slow. It excludes tile endpoints as "the legitimate slow tail" but not the bulk-download endpoints, which are slow for exactly the same reason.

## The rule

`infra/monitoring/alerts.yml`:

```yaml
- alert: GeoLensApiInteractiveLatencyP95
  # fix(#658): replaces the all-handler p95 alert (#466), which fired
  # during plain map browsing — tile requests are the legitimate slow
  # tail, so they get their own rule below.
  expr: |
    histogram_quantile(0.95,
      sum by (le) (rate(http_request_duration_seconds_bucket{handler!~"/tiles/.*"}[5m]))) >= 1
  for: 10m
```

#658 already established the principle: an endpoint that is slow by nature should not fire an alert meant to catch interactive slowness. It applied that to `/tiles/.*` and stopped there.

## Why the same reasoning covers more handlers

Measured on the demo:

| handler | mean |
|---|---|
| `/datasets/{dataset_id}/features.geojson` | 1334 ms |
| `/collections/{dataset_id}/items` | 1197 ms |
| `/datasets/{dataset_id}/export` | 397 ms |

These are bulk downloads. `?limit=2000` on the NYC subway stations collection returns roughly 10 MB. Over a second is the correct, healthy time for that, the same way a tile taking a while is correct.

Because the handler-labelled histogram tops out at `le=1.0`, `histogram_quantile` clips at 1.0, so a handful of these pins p95 at the ceiling regardless of how much other traffic is healthy.

## Two incidents, both at negligible request rates

| | first | second |
|---|---|---|
| firing | 00:15Z, 00:30Z | 03:00Z, 03:15Z, 03:30Z |
| peak non-tile rate | 4.55 req/s | **0.1–0.35 req/s** |
| non-tile p95 | pinned at 1s ceiling | pinned at 1s ceiling |
| demo health | 200 throughout | 200 throughout |

The second one is the informative one. At a third of a request per second the alert still fired for thirty minutes, because a few bulk requests are all it takes. In the first incident 463 non-tile requests exceeded 1s in an hour, 971 of them to `/collections/{id}/items`.

## Why this will recur

[...]

PR #1612, opened 2026-08-19
ci(release): dispatch geolens-release to geolens-examples too

### LINKED ISSUE #1609: release: dispatch geolens-release to geolens-examples too, so its pins get a bump signal
`.github/workflows/release.yml` (around line 240) sends a `geolens-release` repository dispatch to `geolens-io/getgeolens.com` so the site can vendor the contract, installer and OpenAPI snapshot. Nothing notifies `geolens-io/geolens-examples`, which pins every client (`geolens`, `@geolens/sdk`, `geolens-mcp`, `geolens-cli`) to one version and replays its examples against the demo. Its `ci/check-pins.py` asks the demo's `/api/health` on the weekly run and only warns when the demo is ahead of the pin, so a release shows up there as a log line nobody reads.

Suggested change: in the same step, also dispatch to `geolens-io/geolens-examples` (same `SITE_DISPATCH_TOKEN` if it has scope there, or a second secret), with the same `client_payload[version]`. The examples repo can then add a small workflow that opens a "bump pins to X and re-sweep" issue on that event. Keep it `continue-on-error` like the site dispatch so a missing token never fails a release.

Low priority; the weekly warning and the examples repo's new CLAUDE.md rule ("demo ahead of pin is a task, not noise") cover the gap by hand. Filed as an issue rather than a PR because another branch is active in this repo.

ishiland:
`.github/workflows/release.yml` dispatches `geolens-release` to `geolens-io/getgeolens.com` when a GA tag publishes. The same step now also dispatches it to `geolens-io/geolens-examples`, carrying the same `client_payload[version]`.

## Why

`geolens-examples` pins `geolens`, `@geolens/sdk`, `geolens-mcp` and `geolens-cli` to one version across sixteen files, and replays every example against the demo, which is redeployed from the release. Its only signal that a release had happened was `ci/check-pins.py` on the weekly schedule printing `WARNING: the demo runs X, this repo pins Y` and exiting 0. A warning that passes is a task nobody is assigned.

## Tokens

`EXAMPLES_DISPATCH_TOKEN` when it is set, `SITE_DISPATCH_TOKEN` otherwise. So one fine-grained PAT with Contents: read/write on both target repos is enough, and two separate secrets work just as well.

The fallback runs one way only. An examples-scoped token is never tried against the site, because a dispatch that 404s on a scope problem is indistinguishable, from inside this step, from the site being down.

Neither secret is required. With neither configured the step logs two `::notice::` lines and passes. That case stays inside the run script rather than in the step's `if:` on purpose: a skipped step is a grey row nobody reads, which is how a missing token would go unnoticed indefinitely. This file already learned that once, in #1530.

## Failure isolation

The two dispatches cannot mask each other. A failure sets a flag instead of exiting, both targets are attempted on every run, and the step exits non-zero once at the end. `continue-on-error: true` stays, so a downstream repo being unreachable never turns an already-published release red.

Each target gets its own `::notice::` for a missing token, its own `::warning::` on failure, and its own row in the job summary. The heading is now "Release dispatches" and is written once however many rows follow, rather than repeating above each line.

The GA-only `if:` and the latest-release guard are untouched.

## Tests

`backend/tests/test_release_dispatch_targets_1609.py` parses the workflow and asserts both dispatch URLs, one `event_type=geolens-release` and one `client_payload[version]=${TAG}` per target, both token env names, the fallback expression, that no `exit 1` sits between the two dispatches, `continue-on-error: true`, `timeout-minutes`, and both guards.

It pins the shape rather than only the new endpoint, because the regression worth catching is a later edit re-coupling the two targets, not the examples dispatch never having been written in the first place.

Red first, against `main`'s `release.yml`:

```
8 failed, 6 passed in 9.91s
```

The 6 that passed are the site-dispatch and unchanged-guard assertions, which already held. Green on this branch: `14 passed`.

`.github/workflows/release.yml` is also added to `ci.yml`'s `backend` paths-filter. Without it a release.yml-only PR skips Backend Tests, and the test file above becomes a gate that cannot run on the change it exists to catch. Same class as fix(#561), fix(#1088) and fix(#1517) already in that filter.

## Behaviour, not just shape

The structural test asserts the script's text. To check it actually behaves, the step's `run:` was extracted and executed under `bash -e -o pipefail` against a stubbed `gh`:

| scenario | site | examples | exit |
| --- | --- | --- | --- |
| both tokens, both succeed | dispatched | dispatched | 0 |
| site dispatch fails | warning | still dispatched | 1 |
| examples dispatch fails | dispatched | warning | 1 |
| both fail | warning | warning | 1 |
| only `SITE_DISPATCH_TOKEN` set | dispatched | dispatched via fallback | 0 |
| only `EXAMPLES_DISPATCH_TOKEN` set | notice, skipped | dispatched | 0 |
| neither token set | notice, skipped | notice, skipped | 0 |
| tag is not the latest release | not attempted | not attempted | 0 |

Row two is the one that matters: before this change a site failure `exit 1`d out of the step, so the examples repo would never have been told.

The job summary comes out as one `### Release dispatches` section with one row per target in every case.

## Verification

- `actionlint .github/workflows/release.yml`: one SC2016 info at line 66, byte-identical to the finding on `main` (the new step adds none)
- `actionlint .github/workflows/ci.yml`: 8 findings, the same 8 as on `main`
- `pre-commit run --files .github/workflows/release.yml .github/workflows/ci.yml backend/tests/test_release_dispatch_targets_1609.py`: all hooks pass
- `uv run pytest tests/test_release_dispatch_targets_1609.py tests/test_ci_alembic_filter_paths.py tests/test_layering.py -q`: 70 passed
- `uv run ruff check` and `ruff format --check` on the new test: clean

Not verified: the dispatch itself, which only fires on a real GA tag. The `EXAMPLES_DISPATCH_TOKEN` secret does not exist yet, so until it is added (or `SITE_DISPATCH_TOKEN` is rescoped to cover both repos) the examples half logs its `::notice::` and skips.

## Receiver

geolens-io/geolens-examples#34 adds the workflow that listens for this event and opens the pin-bump issue. That side is inert until this one lands.

Closes #1609

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -196,6 +196,12 @@ jobs:
               # must run on PRs that touch it — otherwise CI-config changes ship
               # unvalidated by the very tests they configure.
               - '.github/workflows/ci.yml'
+              # fix(#1609): test_release_dispatch_targets_1609.py reads
+              # release.yml by literal path and asserts both downstream
+              # dispatches (the site and geolens-examples) stay wired and
+              # independent. Without this, a release.yml-only PR skips the
+              # only suite that would notice one of them going away.
+              - '.github/workflows/release.yml'
             frontend:
               - 'frontend/**'
               - 'backend/openapi.json'
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -175,7 +175,12 @@ jobs:
       # CI-1 (see geolens-io/getgeolens.com#108): the site repo listens for
       # this dispatch to run release-drift immediately and to open its
       # docs-contract sync PR (getgeolens.com#108, release-sync.yml).
-      - name: Notify the site of the release
+      # #1609: geolens-examples listens for the same event. It pins every
+      # published client (geolens, @geolens/sdk, geolens-mcp, geolens-cli)
+      # to one version across a dozen files, and otherwise hears about a
+      # release only from a weekly ci/check-pins.py line that warns and
+      # passes. On this event it opens a "bump the pins and re-sweep" issue.
+      - name: Notify the site and the examples repo of the release
         # Only GA tags (no `-` suffix) dispatch — same shape as the
         # IS_PRERELEASE case pattern in "Create release" above. A prerelease
         # never becomes the "latest" release the site's install one-liner and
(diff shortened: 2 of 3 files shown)


ishiland:
@codex review

PR #1616, opened 2026-08-21
fix(config): accept ambient AWS credentials for S3 storage

ishiland:
## What

`STORAGE_PROVIDER=s3` required `S3_ACCESS_KEY_ID` and `S3_SECRET_ACCESS_KEY` at boot, so a Kubernetes deployment had to keep long-lived IAM user keys in a Secret. This accepts an ambient credential source (EKS IRSA / Pod Identity, ECS/EKS container credentials) in their place.

## Why this is a validator-only change

Nothing below `validate_provider_settings` needed the keys:

- `S3StorageProvider` (`platform/storage/s3.py:94-97`) passes `aws_access_key_id` / `aws_secret_access_key` to boto3 **only if set**.
- `derive_gdal_s3_env` guards every `AWS_*` key the same way, and its docstring already anticipates "ambient AWS credentials".

## Verified on real infrastructure

Run on an EKS 1.33 cluster against a real S3 bucket, with the ServiceAccount annotated for IRSA:

| probe | result |
| --- | --- |
| boto3 in the api pod, no explicit credentials | resolved `assumed-role/…/botocore-session-…`, listed the bucket, read an object |
| GDAL/rasterio in the titiler pod, `AWS_ACCESS_KEY_ID` unset | opened `/vsis3/<bucket>/rasters/…/source.cog.tif` |
| booting with s3 and no static keys (before this change) | `FATAL: STORAGE_PROVIDER=s3 requires: S3_ACCESS_KEY_ID, S3_SECRET_ACCESS_KEY` |
| control: with keys | boots |

After the change, with the patched settings deployed and **no credential anywhere in the cluster**, the whole path worked end to end: vector ingest, raster ingest, tile rendering, export. The app logs the source itself — `credential_source=assume-role-with-web-identity`.

## Design notes

- Modelled as four settings fields rather than raw `os.environ` reads, per the CONF-03/CONF-04 convention already in this file.
- A **half-configured** key pair stays an error under either source — that is always a mistake, and boto3 fails it far less legibly at runtime.
- A lone `AWS_ROLE_ARN` does not count as ambient: without the token file it resolves nothing.
- EC2 instance profiles are deliberately **not** probed. That needs an IMDS round trip during settings construction, and a node role broad enough to reach the bucket is not a posture to boot into silently.
- `_make_settings` in the tests pins the four fields off, so the suite cannot flip behaviour when it happens to run inside EKS or ECS.

## Verification

```
backend: pytest tests/test_config.py tests/test_gdal_env.py   209 passed
backend: pytest tests/test_layering.py                         47 passed
ruff check / ruff format --check                               clean
```

Counterfactual: reverting `config.py` alone fails 3 of the 5 new tests (the other 2 are negative guards, correct in both versions).

`_MODULE_LOC_CAPS` for `core/config.py` moves 1375 → 1423 with the rationale recorded inline, per AGENTS.md.

## Companion change

The Helm chart side is geolens-deployments — `storage.s3AmbientCredentials` plus a ServiceAccount to hang the role on. That chart value requires this fix to be released before it does anything.

### DIFF
--- a/backend/app/core/config.py
+++ b/backend/app/core/config.py
@@ -474,6 +474,18 @@ class Settings(BaseSettings):
     s3_allow_http: bool = False
     s3_addressing_style: Literal["auto", "path", "virtual"] = "auto"
 
+    # Ambient AWS credential markers, injected by the runtime rather than by an
+    # operator: EKS IRSA / Pod Identity set AWS_ROLE_ARN +
+    # AWS_WEB_IDENTITY_TOKEN_FILE, and the ECS/EKS container credential
+    # providers set one of the AWS_CONTAINER_CREDENTIALS_* pair. Modelled as
+    # fields rather than read through os.environ, matching CONF-03/CONF-04.
+    # Read only by has_ambient_aws_credentials below; boto3 and GDAL resolve
+    # the actual credentials from these themselves.
+    aws_role_arn: str | None = None
+    aws_web_identity_token_file: str | None = None
+    aws_container_credentials_full_uri: str | None = None
+    aws_container_credentials_relative_uri: str | None = None
+
     # Azure Blob Storage (STOR-01 / Phase 1210)
     azure_storage_container: str | None = None
     azure_storage_connection_string: SecretStr | None = (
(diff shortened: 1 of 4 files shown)


ishiland:
@codex review

PR #1619, opened 2026-08-21
chore(deps): bump pip to 26.2.1 for PYSEC-2026-3721

ishiland:
PYSEC-2026-3721 (pip < 26.2 mishandles doubly-encoded package URLs from indexes) published overnight and now fails `pip-audit --strict` in Security Scan on every backend-touching run — it ejected #1616 from the merge queue this morning. This bumps the transitive pin in backend/uv.lock to 26.2.1.

Lock-only change; no dependency of the app itself moves.

Verification: `cd backend && uv lock --upgrade-package pip` reproduces the diff; Security Scan on this PR exercises the fix directly.

### DIFF
--- a/backend/uv.lock
+++ b/backend/uv.lock
@@ -1554,11 +1554,11 @@ wheels = [
 
 [[package]]
 name = "pip"
-version = "26.1.2"
+version = "26.2.1"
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review

Issue #846, opened 2026-07-28
maplibre-gl v6 upgrade: migration inventory and payoffs


ishiland:
## What

maplibre-gl v6.0.0 shipped 2026-07-22 (we run 5.24.0). Not a jump to make while it is days old, but the migration surface is small and the payoffs are real, so this issue captures the audited inventory for when the ecosystem settles.

## Gate

`@vis.gl/react-maplibre` 8.1.2 (latest) still predates v6. Its peer range (`maplibre-gl >=4.0.0`) lets v6 install, but v6's Map/Camera refactor and typed event classes have not been vetted by a wrapper release. Wait for a react-maplibre release that declares v6 support (or vet it ourselves against a scratch branch).

Re-checked 2026-07-31: npm `maplibre-gl` latest is 6.1.0, `@vis.gl/react-maplibre` latest is 8.1.2 with the peer range unchanged. Our lockfile resolves `maplibre-gl` 5.24.0 and `@vis.gl/react-maplibre` 8.1.1. The maplibre half of the recommendation below (wait for ~6.1) is now satisfied; the wrapper half is not.

Re-checked 2026-08-12: `maplibre-gl` latest is 6.3.0; `@vis.gl/react-maplibre` latest is still 8.1.2 with the peer range unchanged. Wrapper half still unmet.

## Our migration inventory (audited 2026-07-28, line refs refreshed 2026-08-11)

- **The `styleimagemissing` handler uses the removed provide-an-image pattern.** #835 consolidated what used to be three per-surface handlers into one shared factory, `makeStyleImageMissingHandler` at `frontend/src/lib/basemap-utils.ts:157`; the `map.addImage` of a transparent 1×1 now lives in exactly one place (`:164`). Three registration sites wire it up: `BuilderMap.tsx:843-844`, `ViewerMap.tsx:451`, `DatasetMap.tsx:626`. v6 makes the event notify-only; migrate to `Map.setMissingStyleImageResolver`.

  One constraint the conversion must preserve: the factory's `knownImagesOnly` parameter encodes a deliberate divergence. BuilderMap passes `true` so only known-noisy remote-sprite ids are stubbed and an editor authoring a bad icon reference still sees the missing-image warning; ViewerMap and DatasetMap pass `false` and stub everything to keep the public/preview console clean. `setMissingStyleImageResolver` is one resolver per map, so the resolver has to carry that per-surface behaviour rather than the registration site.

[...]

PR #1624, opened 2026-08-22
feat(frontend): upgrade maplibre-gl to v6.5.0

ishiland:
Upgrades `maplibre-gl` from 5.24.0 to 6.5.0. Closes the engine half of #846.

5.24.0 is the last release of the 5.x line, so we were sitting on a branch that gets no further fixes. The wrapper gate cleared on 2026-07-29: `@vis.gl/react-maplibre` 8.1.2 supports v6 through runtime feature detection, and we already depend on it.

## Two commits

**`refactor(frontend): make maplibre call sites v6-ready`** is types only, compiles under both engines, no runtime change.

**`feat(frontend): upgrade maplibre-gl to v6.5.0`** is the version bump and the two API migrations it forces.

Split that way so a bisect can tell a type change from a runtime one.

## Scope correction

#846 described the migration surface as small. It wasn't. v6 made `setPaintProperty`/`setLayoutProperty` generic over the property name, and our layer adapters pass names read from stored style JSON, so 47 call sites stopped compiling across 15 files. The issue's inventory missed that entirely, along with the removal of the default export (`import maplibregl from 'maplibre-gl'` is now a hard error, not just a type-namespace problem).

Where a type could be narrowed at its source, it was. `getExpressionSafeOpacity` now returns `DataDrivenPropertyValueSpecification<number>`, which fixed the circle, cluster, fill, line and mixed adapters without touching any of them. The irreducibly dynamic remainder goes through two pass-through helpers in `layer-adapters/shared.ts`. No `as any`, no `ts-ignore` anywhere in the diff.

## The worker, and why `?worker&url` matters

v5 shipped a UMD bundle that inlined its worker as a string, blobbed it at load time, and registered the URL with itself. Consumers wired nothing. v6 is ESM-only and ships the worker as a real second file resolved relative to `import.meta.url`, and `vite.config.ts` folds maplibre into the `map-vendor` chunk, which moves `import.meta.url` away from the package directory. So `main.tsx` now hands the URL over explicitly.

The query string is load-bearing. A plain `?url` emits the worker without its shared sibling and fails silently in production builds: no error, no console output, vector tiles simply never request. That's live upstream right now as maplibre-gl-js#8186. Confirmed in the build output that the worker emits as its own hashed asset and `map-vendor` references it.

## The missing-image resolver

`styleimagemissing` is notify-only in v6. The response batch is assembled before the event fires, so a listener calling `addImage()` only affects a *later* request for the same id, which surfaces as an icon that intermittently isn't there. Moved to `setMissingStyleImageResolver`, which MapLibre awaits before treating the image as missing.

Each surface registers exactly one resolver, so the builder-vs-viewer `knownImagesOnly` divergence survives the one-resolver-per-map constraint. The setter accepts `null`, so BuilderMap's symmetric teardown is preserved.

## New spec

Nothing in the suite asserted that MVT tiles are ever requested. Every map spec checks DOM and canvas presence, all of which a map with a dead tile worker satisfies: the canvas mounts, the basemap renders, nothing throws. The full 136-test suite was verified green against exactly that condition before this change.

`e2e/vector-tile-pipeline.spec.ts` asserts `.pbf` traffic exists and comes back clean, and that it survives a style swap. Added to `e2e:smoke:core`. It runs against the densest available vector dataset (32,186 features here) and logs which fixture it picked, so it can't quietly degrade into a trivial check.

The showcase console filter matched v5's warning text, which v6 rewrote; it now anchors on the prefix both versions share.

## Verification

Every gate was run against the dev stack **rebuilt onto v6**. Worth stating because the frontend container mounts an anonymous volume over `node_modules`, so a host-side bump alone leaves the container on the old version and Playwright silently tests that instead. Caught mid-review; the container was still on 5.24.0 while the host was on 6.5.0.

- e2e at parity with the pre-change baseline, plus the two new tests
- 411 test files / 5447 unit tests, zero failures
- typecheck, eslint, i18n clean; build succeeds
- `map-vendor` 1,184 kB → 1,144 kB (312.5 → 295.4 kB gzip)

Also captured a v5-vs-v6 screenshot diff across globe, terrain, and the z11→z12 projection handoff, since saved globe maps are invisible to CI. Mercator + terrain is pixel-identical (mean diff 0.000%); globe and the transition band are within run-to-run tile-timing variance. That closes upstream #7660, which was filed against a v6 pre-release and never answered when the maintainer asked whether it predated v6.

Ruled out by inspection rather than assumption: v6's `AJAXError` still carries `status`/`url`, so the tile-auth recovery path reads the same fields (only the declared type narrowed); `_vectorTileFeature` survives and is now actually in the public typings; the `__$json__` nested-property encoding change decodes transparently on every read path we use; and feature-state's array refactor is moot for us, since nothing in the frontend calls `setFeatureState`.

## Deliberately not here

- `fill-layer-opacity`/`line-layer-opacity`. A real fix for the builder's opacity slider double-darkening overlaps, but it touches two frontend allowlists, six adapter call sites, a backend allowlist that silently strips unregistered keys, and it leaves circle layers visibly inconsistent since v6 ships no `circle-layer-opacity`. Follow-up.
- Exported style.json never applies `layer.opacity` to the primary fill/line layer. Pre-existing, same class as #527. Follow-up.
- `FeaturePopup` renders object-valued properties as `[object Object]` while the accessible data panel JSON-stringifies them. Pre-existing. Follow-up.

https://claude.ai/code/session_01GF3ioig7XEMieK26eEbHk7


ishiland:
@codex review

e2e/vector-tile-pipeline.spec.ts
```diff
@@ -0,0 +1,170 @@
...
+    // "> 0 requests" assertion far weaker than it looks.
+    const res = await fetch(`${BASE_URL}/api/datasets/?limit=100`, {
+      headers: { Authorization: `Bearer ${getAuthToken()}` },
+    });
+    expect(res.ok).toBe(true);
+    const data = await res.json();
+    const datasets = data.datasets ?? data.items ?? data;
+    const vector = datasets
+      .filter((ds: { record_type?: string }) => ds.record_type === 'vector_dataset')
+      .sort(
```
ishiland, line 73: Good catch, and it was material rather than cosmetic. Measured it on the dataset page: of 35 `.pbf` responses, 19 are OpenFreeMap glyph ranges under `/fonts/` and 8 are basemap planet tiles, leaving 8 from the dataset. So the assertion could pass with the dataset's tile pipeline entirely dead, which is the exact failure the spec exists to catch.

Both assertions now match `/api/tiles/...data.<table>/....pbf` for the selected dataset (covering the plain and clustered routes), with the table name read from the API rather than inferred. The ignored third-party count is included in the failure message so the exclusion is visible to whoever reads a future failure.

Verified by counterfactual rather than by re-running the happy path: pointing the predicate at a table that serves no tiles fails both tests, where the old predicate passed on basemap traffic alone.

Your finding also exposed a second problem in the same test. Once scoped correctly it failed, because a basemap change does not re-request dataset tiles at all — this app mutates basemap layers instead of rebuilding the dataset source, so cached tiles are correctly left alone, and the old assertion was measuring the basemap's own new traffic. It now swaps the basemap and then zooms, which is what actually exercises the worker still serving that source under a changed style. The zoom goes through the NavigationControl button rather than `mouse.wheel`, since `DatasetMap` sets `scrollZoom={isFullscreen}` and a wheel zoom outside fullscreen silently does nothing.

Fixed in 6093091d98f4e83948629bf185a4259a63dfb712.

ishiland:
@codex review

e2e/vector-tile-pipeline.spec.ts
```diff
@@ -0,0 +1,190 @@
...
+    // assuming nth(0) is. BasemapToggle marks the active one with aria-current
+    // (BasemapToggle.tsx:97), and which one that is depends on theme — under
+    // the dark theme nth(1) IS the active basemap, so clicking it leaves the
+    // basemap identity unchanged, DatasetMap never calls setStyle, and the zoom
+    // below still populates afterSwap. The test would pass having swapped
+    // nothing.
+    const inactive = group.locator('button:not([aria-current])');
+    const inactiveCount = await inactive.count();
+    expect(inactiveCount, 'basemap picker exposed no inactive option to switch to').toBeGreaterThan(0);
+    await inactive.first().click();
```
ishiland: Correct, and it's the same failure mode as your first finding: the test could pass while exercising nothing. Verified the premise before fixing it — `BasemapToggle.tsx:97` sets `aria-current={isActive ? true : undefined}`, so which index is active does depend on theme.

Now selects the first option without `aria-current` rather than assuming nth(0) is active.

I also added the check that was missing underneath the bug. Rather than only asserting the active option changed in the DOM, the test now counts third-party `.pbf` traffic as proof the style genuinely changed, since a real swap necessarily pulls fresh basemap tiles. That makes the vacuous-pass mode detectable regardless of how the option gets selected, which the DOM assertion alone would not.

Verified by counterfactual: clicking the already-active basemap now fails on that guard ("setStyle did not run") instead of passing.

Fixed in 5924a79381b7b7ac6b5ec7b42b5c593102015b9c.

ishiland:
@codex review

e2e/vector-tile-pipeline.spec.ts
```diff
@@ -0,0 +1,198 @@
...
+ * under /fonts/, 8 basemap planet tiles). A `> 0` assertion over all of them
+ * passes with the dataset's own tile pipeline completely dead, which is the
+ * exact failure this spec exists to catch.
+ *
+ * Covers both the plain and clustered routes:
+ *   /tiles/data.<table>/{z}/{x}/{y}.pbf
+ *   /tiles/clusters/data.<table>/{z}/{x}/{y}.pbf
+ *
+ * Matched by ROUTE, not by origin: buildSignedTileUrl()/buildClusterTileUrl()
+ * use TILE_BASE_URL or the server-provided cdn_base_url when configured, so a
```
ishiland, line 52: Both correct, and CI failed E2E Smoke (core) on the second one independently, so this is confirmed rather than theoretical.

Origin: the predicate now matches by route (`/tiles/` + `data.<table>/` + `.pbf`) instead of requiring the BASE_URL prefix. That holds when `TILE_BASE_URL`/`cdn_base_url` puts tiles on another origin, and `data.<table>` is still unambiguous since no third-party basemap URL contains it.

Swap proof: counting basemap `.pbf` assumed every basemap is vector. CI's is raster, so the guard fired on a swap that had genuinely happened. It now asserts `aria-current` moved to the option that was clicked, which is what actually means setStyle ran and is independent of what format the basemap serves.

CI surfaced one more thing neither of us had: its fixture is a 1-feature dataset, not the 32k-feature one available locally, so the spec has to hold at both ends of that range.

Fixed in 6c0355d142346047524f70b025cd1def828160d4.

e2e/vector-tile-pipeline.spec.ts
```diff
@@ -0,0 +1,198 @@
...
+    expect(inactiveCount, 'basemap picker exposed no inactive option to switch to').toBeGreaterThan(0);
+    const targetName = (await inactive.first().textContent())?.trim();
+    await inactive.first().click();
+    await page.waitForTimeout(3000);
+
+    // The swap must have actually occurred, or everything below is vacuous.
+    // Asserted against the DOM rather than by counting basemap tile traffic:
+    // a configured basemap may be a raster XYZ style that emits no .pbf at all
+    // (CI's does), so tile-format-based proof fails on a perfectly healthy
+    // swap. aria-current moving is what actually means setStyle ran.
```
ishiland: Both correct, and CI failed E2E Smoke (core) on the second one independently, so this is confirmed rather than theoretical.

Origin: the predicate now matches by route (`/tiles/` + `data.<table>/` + `.pbf`) instead of requiring the BASE_URL prefix. That holds when `TILE_BASE_URL`/`cdn_base_url` puts tiles on another origin, and `data.<table>` is still unambiguous since no third-party basemap URL contains it.

Swap proof: counting basemap `.pbf` assumed every basemap is vector. CI's is raster, so the guard fired on a swap that had genuinely happened. It now asserts `aria-current` moved to the option that was clicked, which is what actually means setStyle ran and is independent of what format the basemap serves.

CI surfaced one more thing neither of us had: its fixture is a 1-feature dataset, not the 32k-feature one available locally, so the spec has to hold at both ends of that range.

Fixed in 6c0355d142346047524f70b025cd1def828160d4.

ishiland:
@codex review

frontend/src/main.tsx
```diff
@@ -22,6 +22,10 @@ import { ReportProblemHost } from '@/components/report/ReportProblemHost';
 import { appRoutes } from './App';
 import './index.css';
```
ishiland, line 23: Confirmed and fixed — this was the most serious of the four, and the only one in product code rather than the test.

Verified in the build rather than assumed: the entry chunk carried a static `from"./map-vendor"` and `index.html` listed map-vendor in modulepreload. On `main`, `main.tsx` imports no maplibre at all and every route in `App.tsx` is lazy, so this was introduced by the upgrade, not pre-existing. ~295 kB gzip on login, admin and search.

Moved to `@/lib/maplibre-worker`, side-effect imported by the five surfaces that actually construct a map (BuilderMap, ViewerMap, DatasetMap, BboxMapPicker, SpatialFilterPanel). FeaturePopup is deliberately excluded since it imports only `Popup` and renders inside an existing map.

I considered `basemap-utils` as a tidier single choke point and rejected it: it imports maplibre types only, so a value import there would newly drag map-vendor into all fifteen of its importers.

Verified three ways:
- Build graph: static map-vendor reference 1 -> 0, modulepreload entry gone, only lazy `import()` remains.
- Browser: on `/` and `/admin/settings`, zero maplibre-gl requests. (One hit on admin is `@maplibre/maplibre-gl-style-spec`, a separate small validation package our pre-existing `maplibre-filter-utils.ts` uses — not the engine.)
- Runtime: the tile spec still passes, which is the direct check that moving the init did not break worker resolution, plus a browser pass over dataset/globe/terrain with zero console errors and zero missing-image warnings.

The vitest maplibre mock also needed `setWorkerUrl`/`getWorkerUrl`, since those five surfaces are rendered in unit tests.

Fixed in 2f315d9b7ceaa3e54c1fa31d3bdf2421fc2d0ef2.

ishiland:
@codex review

e2e/vector-tile-pipeline.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+    // assertion on the swap alone measures the basemap's traffic instead of the
+    // dataset's. Zooming after the swap is what actually exercises "the worker
+    // still serves THIS source once the style has changed underneath it."
+    // Zoom OUT, not in. fix(#1624): DatasetMap fits the dataset's extent, and a
+    // single-point dataset has a zero-area extent that fits at MAX zoom, which
+    // leaves the zoom-in control disabled — `.click()` then hangs on
+    // actionability until the test times out. CI's seeded fixture
+    // (e2e/fixtures/sample.geojson) is exactly one point, so zoom-in can never
+    // work there. Zooming out is available from any fitted extent and requests
+    // tiles at a new z either way.
```
ishiland: Confirmed by CI, and the mechanism is exactly as you described. The CI log shows `locator.click: Test timeout of 60000ms exceeded` at the zoom-in line rather than an assertion failure — the control is visible but disabled at max zoom, so the click hung on actionability and a healthy worker never got measured.

Fixed by zooming OUT instead of in. That is available from any fitted extent and requests tiles at a new z either way, so it holds for both a single point and a dense dataset, without needing a differently-shaped fixture. Also asserts the control is enabled rather than merely visible, so a disabled control fails fast with a clear reason instead of timing out.

I verified against your specific scenario rather than only re-running the happy path: forcing the spec onto the 1-feature shared fixture locally (the same sample.geojson single point CI seeds) passes both tests.

Fixed in 1c75e5ec34b8de5a76dcfd89b0c3ded85bbd396e.

ishiland:
@codex review

ishiland:
Correcting my previous reply on the zoom thread: the zoom-out fix is superseded, and **the style-swap test is removed** in e80ef8713abc24a27417926cc5b642647d274639.

Zoom-out was right for CI's single-point fixture and wrong for a dense one. A point fits at max zoom, so zooming out demands new tiles; a globally-fitted dataset sits at world view, where zooming out toward z0 demands nothing new. Exactly the mirror of the zoom-in failure it replaced, which I only found because I ran it against both fixture shapes instead of just CI's.

The root cause is structural, not a bad selector. Forcing new tile requests needs a viewport change, and for a dataset whose extent already fills the viewport there is no direction, zoom or pan, that requires a tile the map does not already hold. The assertion is fixture-shape dependent by nature, and it was sitting in `e2e:smoke:core` gating every PR in this repo.

What it guarded did not justify that cost. v6's worker pool is a global lazy singleton rather than per-Style, so a basemap change cannot structurally orphan it. What Style does re-read is the missing-image resolver, and a broken resolver announces itself with a console warning that the showcase console gate already covers.

The failure actually worth gating on is #8186: a dead worker producing zero tile requests, no error, no console output. That test survives, stays in `e2e:smoke:core`, and still fails its counterfactual (point the predicate at a table that serves no tiles) while passing when healthy.

Thanks for the five rounds on this one. Every finding was real, and the last two are what surfaced that the test's premise, not its implementation, was the problem.

ishiland:
@codex review

PR #1632, opened 2026-08-23
fix(dataset): keep an inline cell edit alive across re-renders

### LINKED ISSUE #1628: test(e2e): feature-editing.spec.ts fails intermittently at roughly 1 run in 5
Measured while verifying #1624. Not caused by that PR: I first suspected it was, then ran it both ways under settled conditions and got the same result.

## Measurement

Ten runs against the dev stack, five with the #1624 change present and five with it reverted, alternating and after the stack had settled:

| condition | runs | runs with failures |
|---|---|---|
| with the change | 5 | 1 |
| without the change | 5 | 1 |

Identical rate, so the change is exonerated and the flake is pre-existing.

## Shape of the failure

Which test fails moves between runs. Observed: `integer cell edit persists across reload (E-03 live)` (`:158`), `non-numeric input into an integer cell is rejected with a message` (`:169`), and `text cell edit persists across reload`. A failing run takes out one to three of them.

The failure is always the same shape: a cell button that should carry the seeded value is not visible within the 15s timeout, e.g.

```
Locator: getByText(/not a valid/i).first()
Error: element(s) not found
```

The spec is correctly structured otherwise. `beforeEach` calls `createEditingDataset()` and `afterEach` deletes it, so each test does get a fresh dataset and this is not order dependence or leaked state between tests. The likely cause is a race between dataset creation/ingest completing and the Data tab rendering the seeded rows, which would make it load-sensitive rather than deterministic.

## Why it has not been noticed

`feature-editing.spec.ts` is in no smoke subset, so it does not run on PRs. It only runs in the nightly unfiltered `npx playwright test`, where a single flaky run is easy to write off.

## Worth knowing before fixing

An earlier misreading cost real time here: three consecutive failures immediately after a container rebuild and a full suite run looked like a deterministic regression. It was load. Any fix should be validated over several runs on a quiet stack, not one.

ishiland:
## What

`e2e/feature-editing.spec.ts` failed about one run in five. The reading in #1628, a race between dataset creation and the Data tab rendering the seeded rows, turns out to be wrong. The seeded cell always rendered. Every failure I captured happened after `input.press('Enter')`, at the assertion for either the "Cell updated" toast or the rejection message, with the table still showing the original value and no editor open.

## Root cause

`AttributeTable` rebuilt its `columns` array on every render, and TanStack Table renders `columnDef.cell` as the React component *type* (`FlexRender` calls `React.createElement(def.cell, ctx)`). A new `columns` array is a new element type at every cell position, so React unmounts and remounts the cell subtree instead of re-rendering it, and `InlineCellEditor`'s `value` state goes back to the stored cell value.

The rebuild was unconditional. `columns` listed `handleCellSave`, which in turn lists `updateFeature`, the object react-query's `useMutation` returns. That object is rebuilt on every render (`return { ...result, mutate, mutateAsync: result.mutate }`, react-query 5.101 `useMutation.ts`).

So any render at all between opening a cell editor and pressing Enter discarded what had been typed. Enter then compared the restored value against `initialValue`, took `commit()`'s unchanged-value branch, and closed the editor. No PATCH, no validation message, nothing to assert on. The window between the click and Enter is roughly 100 ms, which is why this was load-sensitive rather than deterministic.

I measured it with a requestAnimationFrame probe injected into the page that stamped every ancestor of the editor input. At the failure the `<td>` kept its identity and only the `<input>` was replaced, with the value back at `100`:

```
1174 ms  val='not-a-number'  input#1  < td#2 < tr#3 < tbody#4 ...
1242 ms  val='100'           input#18 < td#2 < tr#3 < tbody#4 ...
```

No request landed in that window, so the trigger was a plain client-side render.

Real users hit this too. The same reset defeats the "keep the editor open so the value can be corrected" behaviour of #458 E-03 and E-21: a value the type check or the backend rejected was silently replaced by the old one.

## Change

- `columns` now depends on the column list and `t` only. Everything else the renderer needs (`editingCell`, `editError`, `isSaving`, `canEdit`, `compact`, the save handler) is read from one ref during render, which extends the pattern the component already used for `editingCell`.
- `InlineCellEditor`'s focus effect is keyed on `isSaving`. The saving branch swaps the input for a spinner, and the accidental remount is what used to put the caret back afterwards.
- The e2e spec asserts the editor holds the new value before pressing Enter. That is diagnostics, not a fix: it caught the reset in 2 of the 5 reverted-control failures, and the other 3 slipped through in the gap between that assertion and the keypress. The spec is still in no smoke subset.
- New vitest suite `AttributeTable.cell-editor.test.tsx`. It does not mock `useUpdateFeature` on purpose: the real react-query identity churn is the trigger under test, and a stub returning one frozen object would pass against the bug.

None. Frontend plus the e2e spec.

Closes #1628

## Verification

Backend untouched, so no backend gates.

| command | result |
|---|---|
| `cd frontend && npm run typecheck` | pass |
| `cd frontend && npm run lint` | pass |
| `cd frontend && npm run build` | pass |
| `cd frontend && npx vitest run` | 412 files, 5448 tests, all pass |

E2E was `npx playwright test e2e/feature-editing.spec.ts --project=chromium`, one full spec run per row. Worktree frontend code needs the host-Vite recipe from AGENTS.md, so the branch rows went through Vite on `:5175` with `API_PROXY_TARGET=http://localhost:8001`. The baseline row ran against the shared `:8080` stack, which serves `main`.

| condition | stack | runs | runs with failures |
|---|---|---|---|
| `main`, before the fix | `:8080` (main checkout) | 8 | 4 |
| with the fix | `:5175` (this branch) | 15 | 0 |
| fix reverted, spec kept | `:5175` (this branch) | 10 | 5 |

The failures covered all three tests named in the issue, in the same shape every time: the assertion after Enter, with the table still showing the seeded value. The reverted control failed at a higher rate than the one-in-five reported in the issue because the machine was running three other agents at the time. The reverted block ran between the two with-fix blocks (5 with, 10 reverted, 10 with) on the same stack within the same hour, so the two rows are comparable.

Unit counterfactual: with `AttributeTable.tsx` reverted (patch file, not a stash) the new vitest case fails on both assertions, `expected <input> to be <input>` for element identity and `expected the element to have value 250, received 100` for the typed value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,17 @@ and releases use semantic versioning.
   text. Both components now call one shared formatter, so an object or array
   property renders the same way in either place (#1627).
 
+- **Editing an attribute cell no longer loses what you typed.** The dataset
+  Data tab rebuilt its column definitions on every render, and because
+  TanStack Table renders a column's `cell` function as a React component type,
+  that rebuild remounted every cell instead of re-rendering it. An open cell
+  editor was reset to the stored value, so a save started from a re-render —
+  the map above finishing its load, a background query settling — wrote
+  nothing and showed nothing: no update, no rejection message, the editor just
+  closed. Column definitions are now stable, so an in-progress edit survives
+  unrelated renders, and a value the backend or the type check rejects stays
+  in the box to be corrected (#1628).
+
 ## [1.14.2] - 2026-08-21
 
 ### Added
--- a/e2e/feature-editing.spec.ts
+++ b/e2e/feature-editing.spec.ts
@@ -138,6 +138,12 @@ test.describe('Feature editing round-trips', () => {
     const input = page.locator('tbody input');
     await expect(input).toBeVisible();
     await input.fill(nextValue);
+    // fix(#1628): assert the editor is actually holding the new value before
+    // committing. When a re-render used to remount the cell it silently reset
+    // the input to the stored value, and Enter then took the unchanged-value
+    // branch — no PATCH, no toast. The failure surfaced 10s later as "Cell
+    // updated never appeared", which named the wrong thing.
+    await expect(input).toHaveValue(nextValue);
     await input.press('Enter');
     await expect(page.getByText('Cell updated').first()).toBeVisible({
       timeout: 10_000,
@@ -175,6 +181,9 @@ test.describe('Feature editing round-trips', () => {
     await cell.click();
     const input = page.locator('tbody input');
     await input.fill('not-a-number');
+    // fix(#1628): same precondition as editCell — a reset editor would make
+    // Enter a no-op cancel and the rejection message would never be reached.
+    await expect(input).toHaveValue('not-a-number');
     await input.press('Enter');
     await expect(page.getByText(/not a valid/i).first()).toBeVisible({
       timeout: 10_000,
--- a/frontend/src/components/dataset/AttributeTable.tsx
+++ b/frontend/src/components/dataset/AttributeTable.tsx
@@ -103,10 +103,16 @@ function InlineCellEditor({
   const [value, setValue] = useState(initialValue);
   const errorId = useId();
 
+  // fix(#1628): keyed on isSaving, not []. The saving branch below swaps the
+  // input out for a spinner, and a save that fails leaves the editor open —
+  // before #1628 the editor was remounted wholesale on every render, so this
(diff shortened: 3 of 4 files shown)


ishiland:
@codex review

ishiland:
@codex review

PR #1636, opened 2026-08-24
fix(raster): skip PREDICTOR for sub-byte NBITS sources in COG conversion

ishiland:
## What

Fixes a production ingest failure on the demo (2026-08-21, `Peshawar_City_LULC_2050.tif`):

```
gdal_translate failed: ERROR 1: source.cog.tif: PREDICTOR=2 is only supported with 8/16/32/64 bit samples.
```

`_predictor_for_dtype` in `backend/app/processing/raster/cog.py` picked `PREDICTOR=2`/`3` from the rasterio dtype string alone. That's not enough: GDAL's `IMAGE_STRUCTURE` `NBITS` tag sub-byte-packs 1/2/4-bit samples (LULC/palette rasters, common) or non-standard widths (12/14-bit sensor data) into a wider dtype container, so a source that reads as plain `uint8` in rasterio can still be a 4-bit raster underneath. `gdal_translate` hard-refuses `PREDICTOR=2`/`3` on anything outside 8/16/32/64 bits, which failed the whole `cog_convert` step (and the ingest job with it).

The `NBITS` tag lives at the **band level**, not the dataset level: `src.tags(ns="IMAGE_STRUCTURE")` with no band index returns only `INTERLEAVE` for a packed source and silently misses it.

## Why this shape

`convert_to_cog` now checks each band's actual bit width (`_predictor_supported`, a new helper next to `_predictor_for_dtype`) and drops the predictor when any band declares NBITS outside `{8, 16, 32, 64}`. A source with no NBITS tag is trusted at its rasterio dtype (unchanged behavior).

I checked whether `prepare_with_overviews`'s `gdaladdo` call needed the same treatment. It doesn't pass `PREDICTOR`/`PREDICTOR_OVERVIEW` at all (only `COMPRESS_OVERVIEW`), so it's unaffected by this bug.

The probe fails closed: if it can't determine bit width (unreadable file, unexpected tag value), it reports "not supported" rather than risk reproducing the exact failure this fixes. Worst case is a slightly larger COG, not a failed ingest.

## Changes

- `backend/app/processing/raster/cog.py`: new `_predictor_supported(file_path)` helper; `convert_to_cog` calls it before putting `-co PREDICTOR=<n>` on the `gdal_translate` argv.
- `backend/tests/test_rule2_structural.py`: new `RASTERIO_OPEN_ALLOWLIST` entry for the new `rasterio.open` call in `_predictor_supported` (local staged/temp path only, same trust class as the sibling entries already in this file).
- `backend/tests/test_cog_subprocess_env.py`: new `TestPredictorLowBitDepth` class. It builds real NBITS=1/NBITS=4 GeoTIFF fixtures with rasterio's `nbits` creation option (no GDAL CLI dependency, since the CLI isn't guaranteed on the test host) and runs them through `convert_to_cog` with only the GDAL subprocess layer mocked, so the fixture and the per-band tag detection are both genuine.
- `CHANGELOG.md`: `[Unreleased] > Fixed` entry.

No schema/API/config impact.

## Verification

```
uv run pytest tests/test_cog_subprocess_env.py -v          # 15 passed
uv run pytest tests/test_rule2_structural.py -q             # 98 passed
uv run pytest tests/test_layering.py -q                     # 47 passed
uv run pytest tests/test_raster_replace_1221.py tests/test_vrt_rewrite.py \
  tests/test_cog_band_unit_capture.py tests/test_crs_wkt2_backfill_1376.py \
  tests/test_raster_antimeridian_887.py -q                  # 277 passed
```

Also reproduced the underlying GDAL behavior directly on the host GDAL 3.13.0 to confirm the error text and that dropping `PREDICTOR` alone fixes the conversion:

```
$ gdal_translate -of GTiff -co COMPRESS=DEFLATE -co PREDICTOR=2 -co TILED=YES nbits4.tif out.tif
ERROR 1: out.tif: PREDICTOR=2 is only supported with 8/16/32/64 bit samples. ...

$ gdal_translate -of GTiff -co COMPRESS=DEFLATE -co TILED=YES nbits4.tif out.tif
... done.
```

### Counterfactual

Committed the fix, then `git checkout HEAD~1 -- backend/app/processing/raster/cog.py backend/tests/test_rule2_structural.py` to put the pre-fix source back while keeping the new tests, and re-ran the new tests:

```
FAILED tests/test_cog_subprocess_env.py::TestPredictorLowBitDepth::test_nbits4_source_drops_the_predictor
FAILED tests/test_cog_subprocess_env.py::TestPredictorLowBitDepth::test_nbits1_source_drops_the_predictor
FAILED tests/test_cog_subprocess_env.py::TestPredictorLowBitDepth::test_predictor_supported_reads_band_level_tag_not_dataset_level
3 failed, 1 passed, 11 deselected
```

(the 4th test, "ordinary uint8 source keeps the predictor", correctly passed both before and after: it pins pre-existing behavior, not the fix). Then restored the fix with `git checkout HEAD -- ...` and the full suite passed again (160/160 across the three focused files).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,19 @@ and releases use semantic versioning.
 
 ## [Unreleased]
 
+### Fixed
+
+- **Low-bit-depth rasters (NBITS < 8) no longer fail COG conversion.**
+  LULC and palette rasters commonly pack 1/2/4-bit samples into a rasterio
+  `uint8` container, via GDAL's `IMAGE_STRUCTURE` `NBITS` tag, which lives
+  at the band level rather than the dataset level. The COG converter picked
+  a `PREDICTOR` from the dtype string alone, so it handed `gdal_translate` a
+  `PREDICTOR=2` the tool hard-refuses on anything outside 8/16/32/64-bit
+  samples, failing the whole conversion step and the ingest job with it.
+  This is what broke `Peshawar_City_LULC_2050.tif` on the demo.
+  `convert_to_cog` now checks each band's actual bit width before deciding
+  whether a predictor creation option is safe to pass.
+
 ## [1.15.0] - 2026-08-23
 
 ### Added
--- a/backend/app/processing/raster/cog.py
+++ b/backend/app/processing/raster/cog.py
@@ -529,10 +529,54 @@ def _predictor_for_dtype(dtype: str, compression: str = "DEFLATE") -> str | None
 
     Predictors only work for DEFLATE, ZSTD, and LZW.
     Returns None for JPEG, WEBP, LERC (no predictor applicable).
+
+    A rasterio dtype string alone is not the whole story: see
+    ``_predictor_supported`` for the sample-width check ``convert_to_cog``
+    layers on top of this before it lets a predictor onto the argv.
     """
     if compression.upper() not in ("DEFLATE", "ZSTD", "LZW"):
         return None
     return "3" if _is_float_dtype(dtype) else "2"
+
+
+def _predictor_supported(file_path: str) -> bool:
+    """Whether every band's actual sample width supports a GDAL PREDICTOR.
+
+    A rasterio dtype like ``uint8`` does not say how many bits a sample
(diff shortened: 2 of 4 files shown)


ishiland:
@codex review

PR #1637, opened 2026-08-24
test(export): capture evidence when the gpkg determinism gate is about to fail

### LINKED ISSUE #1633: test(export): gpkg byte-determinism gate flaked once in a merge-group run (unrelated PR ejected)
Observed once, 2026-08-23 19:40Z, merge-group run 32661141194 (queue entry for #1629, a frontend-only diff): 

```
FAILED tests/test_export_artifact_cache_1532.py::test_every_export_format_is_byte_deterministic[gpkg]
assert '1bd8de9a5c3c...733635ae81f2e' == '401fd1f0d75d...2dc33cd2c29a2'
```

Two gpkg builds of unchanged data hashed differently despite `normalize_gpkg_timestamps` running on both. The same test passed 10/10 consecutive runs on a dev machine immediately afterwards (same GDAL invocation path, host ogr2ogr), and passed again on the re-enqueued queue entry. So this is a rare, environment- or time-dependent nondeterminism, not a broken gate.

Why it matters more than a normal flake: this test IS the byte-determinism gate for the export artifact cache (#1532), and it runs in every merge-group. A false failure ejects an unrelated PR from the merge queue (that is how it was found); a true rare nondeterminism means a gpkg selection can silently become permanently contested in production, which disables range serving for the default export format.

Not diagnosed. Candidates for whoever picks this up:

- a timestamp column ogr2ogr writes that `_GPKG_TIMESTAMP_COLUMNS` does not cover, stamped only when the conversion crosses some boundary;
- SQLite page-image divergence after the normalizing UPDATE (freelist/page reuse sensitive to the original cell bytes);
- GDAL version skew between CI's container ogr2ogr and the dev host.

First step is evidence, not a fix: capture both artifacts on the next failure (upload as a workflow artifact when the assert fires) and diff the sqlite dumps; the differing bytes name the cause. Do not loosen the assertion.

ishiland:
## Summary

First step for #1633 is evidence, not a fix: this adds an evidence-capture mechanism to `test_every_export_format_is_byte_deterministic` and wires CI to upload it, so the next flake of the gpkg byte-determinism gate leaves artifacts instead of just a hex-prefix mismatch in the log. It does not touch the assertion the gate is built on.

- **Test-side capture** (`backend/tests/test_export_artifact_cache_1532.py`): when the two builds' digests are about to differ, the test now persists both artifact files plus `manifest.json` to `backend/test-artifacts/determinism/<format>_<ts>_<uuid>/` before the (unchanged) `assert` raises. The manifest carries: sha256 + size of each artifact, the `ogr2ogr --version` string, the GPKG normalize step's before/after digest (via a new optional `evidence` side-channel on `_build_export`, default `None` so every other caller is unaffected), and a pure-stdlib `sqlite3` structural comparison — per-table row counts, `PRAGMA page_count/page_size/freelist_count`, `gpkg_contents`/`gpkg_metadata` rows, and a dump-diff of the first 5 differing tables. Capture is wrapped in try/except so a bug in it can never mask the real assertion failure, and it costs one `if` on the pass path.
- **CI-side upload** (`.github/workflows/ci.yml`): added an "Upload determinism evidence" step to the `backend-test` job, gated on `if: failure()` (explicit status function, not a bare `if:`) with `if-no-files-found: ignore` and `retention-days: 14`. Verified `backend-test` runs `uv run pytest` directly on the GitHub Actions runner — Postgres is started via a plain `docker run`, not `docker compose exec` into a container — so `backend/test-artifacts/determinism/` is already on the runner filesystem and needs no copy-out-of-container step.
- **`.gitignore`**: added `backend/test-artifacts/`.
- In-source comments anchored `fix(#1633): ...` per repo convention.

No CHANGELOG.md entry — precedent checked: sibling flake-hardening PRs #1634 (`test(auth): derive the grace-window race tolerance...`), #1587, #1392 are test-only and none touch CHANGELOG.md.

## Verification

- `cd backend && uv run ruff check .` and `uv run ruff format --check .` — pass.
- `uv run pytest tests/test_layering.py -q` — 47 passed (host recipe, `.env.test` via `make env-test` inside the worktree; DB not required for this file).
- `uv run pytest tests/test_export_artifact_cache_1532.py -q` (host recipe, dev DB at `localhost:5434`) — 128 passed.
- Forced-mismatch proof: temporarily appended one byte to the second gpkg build (`second.write_bytes(second.read_bytes() + b"\x00")`), ran `tests/test_export_artifact_cache_1532.py::test_every_export_format_is_byte_deterministic[gpkg]` — it failed as expected and printed `determinism evidence captured: .../test-artifacts/determinism/gpkg_<ts>_<uuid>`, with a manifest containing both digests, sizes, GDAL version, pre/post-normalize hashes, and the sqlite structural diff (correctly reporting zero differing tables/rows for this specific forced case, since the extra byte lands outside the compared table data — itself informative). Reverted the forcing edit (`git checkout --` after committing the real change), re-ran the same test — passed, and confirmed no `test-artifacts/` directory was created on the pass path.
- `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/ci.yml'))"` — YAML parses.

No schema/API impact. Does not close #1633 — diagnosis is still open; this only ensures the next occurrence leaves evidence.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1456,6 +1456,27 @@ jobs:
           path: backend/htmlcov/
           retention-days: 14
 
+      # fix(#1633): the gpkg byte-determinism gate
+      # (test_export_artifact_cache_1532.py::test_every_export_format_is_byte_deterministic)
+      # flaked once in a merge-group run and passed 10/10 locally right after —
+      # a rare, environment- or time-dependent nondeterminism the issue asks
+      # to gather evidence on at the next occurrence rather than guess at now.
+      # This job runs pytest directly on the runner (Postgres is started via
+      # a plain `docker run`, not docker compose exec into a container), so
+      # `backend/test-artifacts/determinism/` — written by the test itself
+      # only when a digest assertion is about to fail — is already on the
+      # runner filesystem; nothing needs to be copied out of a container.
+      # `if: failure()` (an explicit status function, not a bare `if:`, which
+      # silently ANDs success()) so this only fires when a step above failed.
+      - name: Upload determinism evidence
+        if: failure()
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a  # v7
+        with:
+          name: determinism-evidence
+          path: backend/test-artifacts/determinism/
+          if-no-files-found: ignore
+          retention-days: 14
+
   # ---------- alembic clean-DB upgrade smoke ----------
   # Exercise the full migration chain against a freshly-initialized PostGIS
   # container via the clean-DB upgrade
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,12 @@ coverage.xml
 .coverage
 coverage/
 
+# Test evidence capture. fix(#1633): backend/tests/test_export_artifact_cache_1532.py
+# writes both artifacts plus a manifest here when a byte-determinism assertion
+# is about to fail; CI uploads this dir as a workflow artifact on failure, and
+# it must never be committed.
+backend/test-artifacts/
+
 # Build output
 build/
 dist/
--- a/backend/tests/test_export_artifact_cache_1532.py
+++ b/backend/tests/test_export_artifact_cache_1532.py
@@ -23,6 +23,7 @@ Requirements:
 """
 
 import hashlib
+import json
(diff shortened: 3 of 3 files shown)


ishiland:
@codex review

PR #1638, opened 2026-08-24
ops(demo): retain container logs and bridge Prometheus alerts to Azure Monitor

ishiland:
## What / why

Two ops gaps on the demo VM found during the v1.15.0 redeploy audit:

1. **Logs die on container re-creation.** `docker logs` is tied to the
   container's `json-file` log, which is deleted when the container is
   re-created (any `docker compose up -d`/`--build`/image bump). Today's
   redeploy destroyed a week of api/nginx history mid-incident-review.
2. **Prometheus alerts fire into the void.** Prometheus records `ALERTS` but
   nothing was ever notified — despite the VM already having a working
   channel (Azure Monitor action group `ag-geolens-demo-ops`, email, used by
   the disk alert `geolens-demo-disk-90`) and a system-assigned managed
   identity that can use it.

This PR is repo-side artifacts only; the orchestrator applies them to the VM.

### Part 1 — `infra/demo/logging-override.yml`

Opt-in compose override, **not** a change to the shipped root
`docker-compose.prod.yml` (a journald default would silently break any
non-systemd self-host). Sets `logging: {driver: journald, options: {tag:
"{{.Name}}"}}` for every service in the base file. Activated on the VM via
`COMPOSE_FILE=docker-compose.prod.yml:logging-override.yml` in
`/opt/geolens/.env`, so a bare `docker compose` picks it up. `docker logs`
keeps working; history now also survives re-creation via
`journalctl CONTAINER_NAME=geolens-api-1 --since ...`.

### Part 2 — `infra/demo/monitoring/prom-alert-bridge.sh` (+ `.cron`)

Cron job, every 5 minutes: GETs `http://127.0.0.1:9090/api/v1/alerts`, counts
`state=="firing"` alerts (python3 — jq isn't guaranteed on the VM), fetches an
IMDS token scoped to `https://monitoring.azure.com/` plus the subscription id
(also via IMDS), and POSTs the count as an Azure Monitor custom metric
(`geolens/prometheus` / `alerts_firing`) against the VM's own resource ID.
Always POSTs, even at count=0 — a zero series is what lets the alert rule's
"no data" state mean "bridge is down" instead of double-duty as "all quiet".
Logs one line per run to `/var/log/geolens-prom-bridge.log`; any HTTP failure
logs and exits 0 (cron stays noise-free).

`infra/demo/monitoring/README.md` documents install steps for both parts, the
already-provisioned Azure side (see mid-review update below), and a redeploy
note: neither of these syncs to the VM automatically on release — same
manual-copy convention as `infra/demo/monitoring/alerts.yml` relative to the
tracked `infra/monitoring/alerts.yml`.

**Mid-review update incorporated:** the orchestrator validated the exact
request shape live against the real endpoint and reported that no
`Monitoring Metrics Publisher` role assignment is needed (a VM's managed
identity may emit metrics against itself), and that the metric alert rule
(`geolens-demo-prom-alerts`, avg `alerts_firing` > 0 over 15m, eval 5m, action
group `ag-geolens-demo-ops`) already exists. The script and README reflect
that — no role-assignment step, alert rule documented as existing rather than
as a command to run.

### A repo-structure note worth flagging

`infra/demo/` has never been tracked in this repo (`git log --all -- infra/demo`
is empty) — it's local-only content excluded via `.git/info/exclude`, and its
top-level `README.md` (untracked, not touched by this PR) is explicitly
marked "Private: do not commit to the public repo" (it carries the VM's
public IP and an admin-credential pointer). Only `infra/monitoring/*` (the
generic, VM-agnostic reference config) was previously tracked. The four files
in this PR contain no VM-identifying secrets (subscription id comes from IMDS
at runtime; nothing hardcoded beyond the already-public resource *names*), so
I force-added them (`git add -f`, per the local exclude) rather than folding
them into the private README. Flagging this so it's a deliberate call, not a
missed one — happy to fold `infra/demo/monitoring/README.md` differently if
preferred.

## Verification

```
docker compose -f docker-compose.prod.yml -f infra/demo/logging-override.yml config --quiet
# exit 0, no errors (only expected "variable is not set" warnings for
# POSTGRES_*/JWT_SECRET_KEY/etc — no .env in this environment)

# merged services == base services, both profiles:
diff <(docker compose -f docker-compose.prod.yml config --services | sort) \
     <(docker compose -f docker-compose.prod.yml -f infra/demo/logging-override.yml config --services | sort)
# -> identical (7 core services)
diff <(docker compose --profile cloud-dev -f docker-compose.prod.yml config --services | sort) \
     <(docker compose --profile cloud-dev -f docker-compose.prod.yml -f infra/demo/logging-override.yml config --services | sort)
# -> identical (10 services incl. minio/minio-setup/valkey)

# confirmed every service in the merged config actually carries the
# journald logging block (parsed the rendered YAML)

shellcheck infra/demo/monitoring/prom-alert-bridge.sh   # clean, no findings
bash -n infra/demo/monitoring/prom-alert-bridge.sh      # syntax OK
```

Functional dry run against a local mock Prometheus + IMDS + Azure Monitor
endpoint (script's URLs are env-overridable for exactly this):
- 0 firing alerts -> logs `OK count=0 firing=none metric_post=200`, posts a
  zero-valued series.
- 2 firing + 1 pending -> logs `OK count=2
  firing=HostDiskUsageHigh,ContainerMemoryNearLimit metric_post=200` (pending
  correctly excluded); posted body:
  `{"time":"...","data":{"baseData":{"metric":"alerts_firing","namespace":"geolens/prometheus","dimNames":[],"series":[{"dimValues":[],"min":2,"max":2,"sum":2,"count":1}]}}}`
  — matches the shape validated live by the orchestrator against the real
  endpoint.
- Prometheus unreachable -> logs the curl error, exits 0.
- Metric POST unreachable -> logs the failure detail, exits 0.

This is new-file/new-capability work rather than a fix to existing broken
logic, so the counterfactual is that none of this exists on `main` today:
there is no `logging-override.yml` to merge (services stay on `json-file`),
and no `prom-alert-bridge.sh` to run at all.

No CHANGELOG entry — checked precedent on the only tracked `infra/` history
(`a26f3cdf6` shipping the original monitoring config, `1fbd72f06` splitting an
alert rule): neither touched `CHANGELOG.md`.

### DIFF
--- a/infra/demo/logging-override.yml
+++ b/infra/demo/logging-override.yml
@@ -1,0 +1,98 @@
+# ==============================================================================
+# Demo-VM log retention override — NOT part of the shipped root compose.
+# ==============================================================================
+# Root docker-compose.prod.yml (and docker-compose.yml) leave every service on
+# Docker's default `json-file` log driver with no explicit `log-opts`. On a
+# fresh host that means the daemon's own default applies, which most self-host
+# installs never touch — switching the SHIPPED compose to journald by default
+# would silently break any non-systemd host (bare Docker on a distro without
+# journald, a Docker-in-a-VM appliance, etc.), so that default stays put. This
+# file is an opt-in override, demo-only.
+#
+# The problem it fixes: `docker logs` output is tied to the CONTAINER's log
+# file, not the image or the compose project. `docker compose up -d --build`
+# (or any command that recreates a container — a new image tag, a changed
+# `environment:` block, `--force-recreate`) allocates a fresh container ID and
+# a fresh, EMPTY log file; the old container's log file is deleted with it.
+# The 2026-08-23 v1.15.0 redeploy did exactly that and took a week of api/
+# nginx history with it mid-incident-review. journald instead retains log
+# entries by CONTAINER_NAME/CONTAINER_TAG independent of container churn, so
+# history survives recreation (bounded by journald's own retention policy on
+# the host, e.g. SystemMaxUse in journald.conf).
+#
+# Activate on the VM via COMPOSE_FILE in /opt/geolens/.env so a bare
+# `docker compose ...` (no -f flags) picks it up automatically:
+#   COMPOSE_FILE=docker-compose.prod.yml:logging-override.yml
+# (compose reads COMPOSE_FILE as a `:`-joined list, applied in order — the
+# override merges on top of the base file exactly as -f/-f would.)
+#
+# `docker logs <container>` keeps working unmodified: Docker's journald log
+# driver still backs the `docker logs`/`docker compose logs` API, it just also
+# persists into the systemd journal outside the container's lifecycle.
+#
+# Querying history directly via journald (survives `docker compose up -d`
+# recreating the container, unlike `docker logs`):
+#   journalctl CONTAINER_NAME=geolens-api-1 --since "2026-08-23 00:00:00"
+#   journalctl CONTAINER_NAME=geolens-frontend-1 --since "-1 week"
+# CONTAINER_NAME is always populated by the journald driver regardless of the
+# `tag` option below; `tag` additionally sets SYSLOG_IDENTIFIER for tools that
+# key off that field instead.
+#
+# IMPORTANT: only override services that exist in docker-compose.prod.yml — a
+# service name here that ISN'T in the base file creates a NEW service instead
+# of overriding an existing one. Verify after any edit:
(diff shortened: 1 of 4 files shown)


ishiland:
Closing without merge: infra/demo/ is deliberately untracked (private demo-ops area; infra/ is in .git/info/exclude and the demo README is marked do-not-commit). The four files are kept locally and installed on the VM directly — publishing demo-VM-specific operational config to the public repo would break that convention. Content itself was fine.

Issue #1676, opened 2026-08-26
fix(import): token-handling copy overpromises — import and reupload-commit doors still place service tokens in durable task args


ishiland:
## The gap

The import UI promises: "Service tokens are used during import only and are not persisted" (`frontend/src/components/import/WorkflowRail.tsx:154`, `frontend/src/i18n/locales/en/import.json:242`).

That is true for the refresh door, which passes only a `credential_ref` into durable task args via the one-use Valkey lease (#1220, hardened in #1277; claim at `backend/app/processing/ingest/tasks_reupload.py:685`). It is not true for the other two doors:

- initial service import: `token` rides in durable Procrastinate task args (`backend/app/processing/ingest/service.py:1149`)
- reupload commit: same (`backend/app/modules/catalog/datasets/api/router_reupload.py:620`)

This split is deliberate — wiring the lease into those doors unconditionally would break protected imports on installs without Valkey (see the `credentials.py` module docstring) — but the UI copy predates the split and now overpromises.

Exposure is bounded: the worker deletes successful jobs (`delete_jobs="successful"`), so retained plaintext lives in failed/pending job rows, not indefinitely for every import.

## Ask

Pick one (or both):

1. **Copy fix (cheap):** scope the promise honestly — e.g. "used for this import and deleted with the job" — or per-door copy, since the refresh panel's "Used once for this refresh and never stored" (`en/dataset.json:780`) is already accurate.
2. **Lease parity (better):** wire the import and reupload-commit doors through the same lease when Valkey is reachable, falling back to the current behavior (with honest copy) when it is not.

Either way the copy and the mechanism should stop disagreeing. i18n note: any copy change lands in all four locales.

PR #1678, opened 2026-08-26
fix(refresh): page large ArcGIS layers with the import path's progress guard

### LINKED ISSUE #1675: fix(refresh): service refresh fetches ArcGIS layers unpaged, without the import path's page-progress guard
## The gap

The initial-import path and the refresh path fetch an ArcGIS FeatureServer layer differently, and only one of them is guarded.

- **Initial import** (`ingest_service`) pages explicitly: `resultOffset`-based chunks ordered on the OID field, with a no-progress check that aborts rather than looping or silently stopping short (`backend/app/processing/ingest/tasks_vector.py:886-935`, chunk size at `:49`). Pagination info is only trusted when the server declares `supportsPagination` (`backend/tests/test_arcgis_auth.py` covers the probe side).
- **Refresh** — both doors (`POST /datasets/{id}/refresh` and the reupload-commit dialog) dispatch `reupload_service`, which performs a single unpaged `run_ogr2ogr_service` fetch (`backend/app/processing/ingest/tasks_reupload.py:903-921`) and relies on GDAL's driver-side paging behavior.

## Why it matters

A refresh of a layer larger than the server's `maxRecordCount` depends on GDAL driver behavior the import path deliberately does not trust: servers that omit or misreport pagination support, sparse OIDs, and transfer-limit responses are exactly the cases the import loop's no-progress guard exists for. The failure mode is the bad one — a smaller-than-source copy that swaps in cleanly, since run rows only record `feature_count_before/after` and nothing verifies the count against the source.

The dataset being refreshed was, by construction, imported through the guarded path, so the refresh path is strictly weaker on the same data.

## Ask

Bring the refresh fetch up to import parity: either route the service-origin refresh fetch through the same guarded pagination loop, or extract the guarded fetch into a helper both paths share (the run-lifecycle and staging/swap layers are already shared, so the fetch is the remaining divergence). A regression test should cover a paged fetch through the refresh door.

Related: #1218 (source-state columns), #1277 (one-request refresh), and the guarded loop's history in `tasks_vector.py`.

ishiland:
Closes #1675; the copy half of #1676 (lease parity stays open there).

Both findings come from the pipelines Phase 0B delta audit (2026-08-26), which reconciled the July handoff against the shipped Refresh feature.

## Paging parity (#1675)

A refresh of a large ArcGIS layer performed one unpaged ogr2ogr fetch and trusted GDAL driver paging — the exact behavior the initial-import path's guarded loop exists to distrust (servers that omit or misreport `supportsPagination`, sparse OIDs, transfer limits). The dataset being refreshed was necessarily imported through the guarded path, so refresh was strictly weaker on the same data, and the failure mode was the bad one: a short copy that swaps in cleanly, since runs only record before/after feature counts.

- The guarded `resultOffset` loop moved verbatim (no-progress check intact) to `tasks_common.run_paged_arcgis_service_fetch`.
- The import path calls it with its per-page progress writer via `on_page`; behavior pinned by the existing `test_ingest_progress` suite, unchanged.
- The refresh executor runs the same criteria and loop through `_fetch_service_layer_with_paging_guard` (module-level so the branches stay out of `reupload_service`'s complexity budget), forwarding the origin-contact spawn hook to every page (monotonic OR, so repeated arming is harmless).
- The page-info probe resolves through `tasks_vector`'s module attribute, so one monkeypatch covers both doors in tests.

New tests drive the real `reupload_service` worker entry point (the `test_refresh_gate_1269` recipe, ogr2ogr boundary faked): five appended pages for 4,500 features with correct `resultOffset`/`resultRecordCount` per call; an empty page fails the run (recorded on the run row, then re-raised for the queue) with nothing swapped; a small layer keeps the single fetch.

## Token copy (#1676, copy half)

The import rail promised "Service tokens are used during import only and are not persisted" while the import and reupload-commit doors still place tokens in durable task args (deliberately — the one-use lease needs Valkey). The copy now states the guarantee that is true at every door: tokens are never stored with the dataset. All four locales updated; the refresh panel's already-accurate "used once and never stored" copy is untouched.

## Verification

- `tests/test_refresh_pagination_1675.py` (new, 3 tests), plus `test_ingest_progress.py`, `test_refresh_gate_1269.py`, `test_service_refresh_1220.py`, `test_reupload_service.py`: 88 passed
- `tests/test_layering.py` green (three LOC ratchets synced with rationale)
- `cd frontend && npm run test:i18n && npm run typecheck` green; import component suites 106 passed
- No API/schema impact; no migration; no OpenAPI change

### DIFF
--- a/backend/app/processing/ingest/tasks_common.py
+++ b/backend/app/processing/ingest/tasks_common.py
@@ -893,11 +893,27 @@ async def run_paged_arcgis_service_fetch(
                 _text(f"SELECT COUNT(*) FROM {_qtable(staging_table, schema=schema)}")
             )
             next_imported_rows = int(result.scalar_one())
-        if next_imported_rows <= imported_rows:
+        grew = next_imported_rows - imported_rows
+        if grew <= 0:
             raise ogr.IngestionError(
                 "ArcGIS service import made no row-count progress "
                 f"at offset {offset}; upstream pagination may be "
                 "unsupported or returned an empty page."
+            )
+        expected = min(page_size, feature_count - offset)
+        if grew != expected:
+            # fix(#1675 codex r2): a server that returns SOME rows but fewer
+            # than the requested page while the offset still advances by
+            # page_size would silently skip records — positive growth is not
+            # enough, the growth must be exact or a truncated copy swaps in
+            # cleanly. A mid-fetch source mutation trips this too, which is
+            # the safe direction: fail and retry against fresh counts.
+            raise ogr.IngestionError(
+                f"ArcGIS page at offset {offset} returned {grew} rows where "
+                f"{expected} were expected; the server may cap responses "
+                "below its advertised page size or the source changed "
+                "mid-fetch. Refusing to continue with a potentially "
+                "incomplete copy."
             )
         imported_rows = next_imported_rows
         if on_page is not None:
--- a/backend/app/processing/ingest/tasks_common.py
+++ b/backend/app/processing/ingest/tasks_common.py
@@ -824,6 +824,87 @@ async def _archive_original_file(
         return False
 
 
+async def run_paged_arcgis_service_fetch(
+    *,
+    service_type_raw: str,
+    service_type: str,
+    source_url: str,
+    layer_name: str,
+    layer_id: "int | str | None",
+    token: "str | None",
+    staging_table: str,
+    db_conn_str: str,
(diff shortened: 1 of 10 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

Issue #1666, opened 2026-08-25
OpenAPI schema declares FastAPI defaults where the app overrides them at runtime (3 cases)


ishiland:
The generated OpenAPI schema describes FastAPI's defaults in three places where the app overrides them at runtime. Anyone generating a client from the contract inherits all three, and the docs site vendors this snapshot, so it also misleads anyone reading the published API reference.

Found while auditing docs.getgeolens.com against v1.15.1. All three verified against a live 1.15.1 instance.

### 1. Validation errors are declared `HTTPValidationError`, but served as RFC 7807

Every route with request validation declares:

```json
"422": {"content": {"application/json": {"schema": {"$ref": "#/components/schemas/HTTPValidationError"}}}}
```

`HTTPValidationError` is `{detail: [{loc, msg, type}, ...]}`. But `backend/app/standards/ogc/errors.py:292-309` registers an app-wide handler that overrides it:

```python
@app.exception_handler(RequestValidationError)
async def validation_exception_handler(request, exc) -> JSONResponse:
    status_code = 400 if _is_standards_path(request) else 422
    detail = "; ".join(...)                      # flattened to a single string
    return JSONResponse(
        status_code=status_code,
        content=ProblemDetail(title=..., status=status_code, detail=detail).model_dump(),
        media_type="application/problem+json",
    )
```

Live:

```
GET /api/search/datasets/?limit=notanumber           422  application/problem+json  {type,title,status,detail}
GET /api/datasets/{id}/export?format=bogus           422  application/problem+json  {type,title,status,detail}
GET /api/collections/datasets/items?limit=notanumber 400  application/problem+json  {type,title,status,detail}
```

Two mismatches: the media type and body shape, and `detail` being a flattened string rather than an array. The 400-vs-422 standards split is already normalized in the generated description (`api/main.py:1235-1238`), so the mechanism for fixing this exists — it just does not cover the response schema.

### 2. `keywords` is declared as a request body on a GET

`GET /collections/datasets/items` declares 26 query parameters, and `keywords` is not one of them — it appears as an `application/json` request body titled `Keywords`.

It is really a query parameter: `backend/app/modules/catalog/search/router.py:437` declares `keywords: list[str] | None = Query(None, ...)`. The handler reads it off the raw query string at `:528-531` and `:1135-1136`, under the comment *"list[str] may not bind via Depends"* — the workaround for the same binding quirk that produces the wrong schema.

Live, against a 29-record catalog:

```
[...]

PR #1679, opened 2026-08-26
fix(maps): publish the style.json contract and let exports round-trip

### LINKED ISSUE #1672: GET /maps/{map_id}/style.json declares an empty 200 schema
Split out of #1666, whose other three cases were fixed in #1670. This one was deliberately left, because it needs a different and larger piece of work.

## The defect

```json
"200": {"content": {"application/json": {"schema": {}}}, "description": "Successful Response"}
```

`export_map_style_endpoint` (`backend/app/modules/catalog/maps/router_sharing.py:223`) returns a `JSONResponse`, so FastAPI has no model to infer a schema from. The route returns a complete MapLibre style document.

## Why it matters beyond a useless generated model

The `sprite` field's shape is undiscoverable from the contract, and the two sides disagree:

- The export always emits an **array** — `style_json.py` hardcodes `[{"id": ..., "url": ...}]`.
- `MapStyleImportRequest.sprite` on the import side is `string | null`.

A reviewer reading only the contract reasonably concludes the export may return a string. It cannot.

## Why it was not fixed alongside the others

The fix is a response model for a full MapLibre style document. Hand-writing one is the exact trap #1670 spent its effort avoiding — that PR deliberately *copied* parameter definitions from a sibling operation rather than restating them, because a hand-written mirror of a schema is the thing that drifts from it. A hand-typed MapLibre style model would be a much larger mirror with much more room to drift, and this repo has been bitten by hand-typed API mirrors before.

## Options worth weighing before starting

1. **A Pydantic model for the style document.** Most correct, largest surface, and the drift risk above. MapLibre's style spec is versioned upstream, so the model dates.
2. **A narrower model covering only the fields the export actually emits**, documented as non-exhaustive. Smaller and honest, but "non-exhaustive schema" is awkward in a generated client.
3. **A hand-written `responses={200: ...}` schema block** naming just the top-level shape plus `sprite`. Cheapest, and fixes the specific misreading, but publishes a partial schema.
4. **Reference the upstream MapLibre style spec by URL** in the description and leave the schema open, fixing only the `sprite` asymmetry. Cheapest of all; the generated model stays untyped.

Option 4 plus a `sprite` note may be the right amount of work — the concrete harm reported is the `sprite` misreading, not the absence of a model. Worth a decision before implementing.

## Also worth settling

Whether the import side should accept the array form too, so export output round-trips through import. Right now it does not, and that is arguably the more user-visible half of the asymmetry.

## Related

- #1666 — the original report, cases 1/2 fixed in #1670.
- #1670 — the contract-correctness pass, including why hand-typed mirrors were avoided there.

ishiland:
Closes #1672.

Split out of #1666 as its own issue because the fix needed a decision on how much MapLibre style schema to publish. The issue's own analysis converged on option 4 — reference the upstream spec, don't mirror it — plus fixing the sprite asymmetry, and that's what this implements.

## What changed

**Round-tripping (the user-visible half).** `GET /maps/{map_id}/style.json` always emits `sprite` in the MapLibre array form (`[{id, url}]`), but `MapStyleImportRequest.sprite` only accepted a string — so an exported style document failed re-import with a 422 on the sprite field. The field is now `string | MapSpriteEntry[] | null`, and a new test pins the full import → export → import round-trip (it fails on the old schema).

**The contract (the reported half).** The export operation's 200 response was an empty schema (`{}`), from which a reviewer reasonably concluded the sprite might be a string. It now declares an object whose description names the MapLibre Style Specification and pins the one guarantee beyond the spec: sprite is always the array form, and import accepts it. Deliberately NOT a field-by-field model — the issue's write-up explains why a hand-typed mirror of a versioned upstream spec is the drift trap #1670 avoided, and this PR follows that reasoning.

## Verification

- New tests: sprite form matrix (string ✓ / array ✓ / number → 422), export→import round-trip, and contract assertions on the regenerated snapshot (`test_phase_275_maps_import_typed_body.py`)
- `test_sdks_round_trip.py`, `test_maps_style_json.py`, `test_layering.py` (two LOC ratchets raised with rationale): 183 passed
- `make openapi-check`, `make sdks-check` (post-commit), `npm run typecheck` green
- Contract impact is additive: new `MapSpriteEntry` component, widened `sprite` union, documented 200 schema; SDKs and `api.generated.ts` regenerated


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1681, opened 2026-08-26
feat(export): add FlatGeobuf export format

ishiland:
## Summary

Adds `fgb` (FlatGeobuf) to the dataset export endpoint's format list, alongside the existing GeoPackage, GeoJSON, Shapefile, CSV, and GeoParquet options.

- `ExportFormat` gains `fgb`; `FORMAT_MAP` gains a `fgb` entry (driver `FlatGeobuf`, ext `.fgb`).
- FlatGeobuf is a single-file format, like GeoJSON/GPKG/CSV — it deliberately does **not** go through the `format_key == "shp"` zip special-casing in `export/service.py`.
- Geometry-required gate in the router (`?format=fgb` on a non-spatial dataset returns 400, same as gpkg/geojson/shp/parquet) and the endpoint docstring both updated.
- Frontend: `ExportButton`'s format picker gains a FlatGeobuf option; `export.fgb` i18n key added to all four locales (en/es/fr/de) since it's user-visible.
- `fgb` also joins the catalog discovery surfaces every other export format is already in: a download row in `generate_distributions`' DCAT template table (`catalog/records/service.py::_DISTRIBUTION_TEMPLATES`), the STAC/OGC asset media-type map (`catalog/search/service_records.py::_FORMAT_MEDIA`), and a new migration (`0054_backfill_flatgeobuf_distributions`) that backfills the row onto spatial datasets created before this deploy.

**Media type**: `application/vnd.flatgeobuf`. FlatGeobuf has no IANA registration; this is the vendor-prefixed type the format's own maintainers proposed after an OGC standardization attempt stalled ([flatgeobuf/flatgeobuf#112](https://github.com/flatgeobuf/flatgeobuf/discussions/112)), and it mirrors the pattern this codebase already uses for GeoParquet (`application/vnd.apache.parquet`), which also has no IANA registration.

**Concurrent-work note**: #1682 landed while this PR was in review and independently enabled `.fgb` uploads/ingest (allowlist default, magic-byte content validation, compose/env defaults, CLI `scan`). This PR is rebased on top of it and does not duplicate any of that — it is scoped to the export side only. The backfill migration here revises `0053_source_format_fgb` (#1682's migration) rather than a core revision authored alongside it. A follow-up (#1683) tracks extending the manifest-apply door to the four tier-1 formats #1682 added, deliberately deferred as its own contract-affecting change.

## Verification

Empirically confirmed byte-determinism (two builds of the same source hash identically) via the real `ogr2ogr -f FlatGeobuf` writer before relying on it — `_DETERMINISM_FORMATS` in `test_export_artifact_cache_1532.py` is derived from `FORMAT_MAP`, so `fgb` was automatically picked up by `test_every_export_format_is_byte_deterministic` with no test-file changes needed.

```
uv run pytest tests/test_export.py tests/test_export_access.py tests/test_export_hardening.py tests/test_export_head_1513.py -q
uv run pytest tests/test_export_artifact_cache_1532.py -q
uv run pytest tests/test_distribution_reconcile_1314.py tests/test_dataset_source_state.py tests/test_postgis_refresh_1265.py tests/test_records_related.py -q
uv run pytest tests/test_ogc_record_properties.py tests/test_flatgeobuf_distribution_backfill_1681.py -q
uv run pytest tests/test_sdks_round_trip.py -q

cd frontend && npm run typecheck && npm run lint && npm run test:i18n
npx vitest run src/components/dataset/__tests__/ExportButton.test.tsx

make openapi-check
make sdks-check
```

Migration round-trip (`0054`'s upgrade/downgrade) verified against a from-scratch bootstrapped Postgres DB with `alembic upgrade head` + `alembic check` (exit 0, no drift) — `make alembic-check` itself can't run against a worktree's migrations since the api container's bind mount is main-checkout-relative.

All pass locally. CI green.

## API/contract impact

`ExportFormat` enum gains `fgb`. Regenerated and committed: `backend/openapi.json`, both SDKs (`sdks/python`, `sdks/typescript`), and `frontend/src/types/api.generated.ts`.

### DIFF
--- a/backend/alembic/versions/0054_backfill_flatgeobuf_distributions.py
+++ b/backend/alembic/versions/0054_backfill_flatgeobuf_distributions.py
@@ -1,0 +1,83 @@
+"""Backfill FlatGeobuf download distributions for existing spatial datasets.
+
+fix(#1681 codex review): _DISTRIBUTION_TEMPLATES now generates a FlatGeobuf
+download distribution at dataset creation (records/service.py), but those rows
+are materialized — existing spatial datasets created before this deploy would
+never advertise the fgb download in DCAT feeds until an unrelated geometry-mode
+flip happened to run reconcile. This backfills one row per spatial dataset that
+lacks it, so DCAT distributions match what a fresh ingest produces. Mirrors
+0013_backfill_geoparquet_distributions.py, the precedent for adding a format to
+this materialized list.
+
+Idempotent: the NOT EXISTS guard (plus ON CONFLICT DO NOTHING against
+uq_record_distribution) makes re-running a no-op. Downgrade removes only the
+auto-generated fgb download rows.
+
+Non-spatial datasets are excluded (geometry_type IS NULL) — FlatGeobuf
+requires geometry, matching generate_distributions' spatial-only filter.
+
+fix(#1681 codex review round 2): the NOT EXISTS guard reads only
+auto_generated rows, matching generate_distributions' own existence probe
+(fix(#1370) there) — a user-authored download/fgb row must not suppress the
+platform's own export row; the intended end state is both rows coexisting.
+Narrowing the guard reopens the case that broad check was incidentally
+absorbing: a user's row at the exact guessable URL this INSERT targets
+(`/datasets/{id}/export?format=fgb`) collides on uq_record_distribution, so
+the INSERT carries `ON CONFLICT DO NOTHING` — the same resolution
+`reconcile_distributions` uses for the identical collision at runtime (see
+`TestATemplateUrlCollisionDoesNotRaise` in
+test_distribution_reconcile_1314.py).
+
+Revises 0053_source_format_fgb (#1682, merged concurrently) rather than a
+core revision authored alongside this one: that PR's own migration already
+covers letting 'fgb' through chk_datasets_source_format, so this file is
+scoped to the (unrelated) DCAT distribution gap only.
+
+Revision ID: 0054_backfill_flatgeobuf_distributions
+Revises: 0053_source_format_fgb
+Create Date: 2026-08-26
+"""
+
+from typing import Sequence, Union
+
+from alembic import op
+
+revision: str = "0054_backfill_flatgeobuf_distributions"
+down_revision: Union[str, None] = "0053_source_format_fgb"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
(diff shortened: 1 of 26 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
Investigated the FlatGeobuf media-type finding above. The claim that "the linked FlatGeobuf discussion proposes [application/flatgeobuf]" doesn't hold up — I re-fetched flatgeobuf/flatgeobuf#112 twice independently and it only ever proposes `application/vnd.flatgeobuf` (m-mohr, 2022-07-01: "The most reasonable thing is probably to just define one in the vendor prefixed area"), and reports the OGC standardization effort stalled with no final recommendation either way.

Checked the actual ecosystem instead of relying on the discussion thread:

- `pystac` (the reference STAC library) hardcodes it:
  `FLATGEOBUF = "application/vnd.flatgeobuf"  # https://github.com/flatgeobuf/flatgeobuf/discussions/112#discussioncomment-4606721`
  (`core/pystac/media_type.py`, stac-utils/pystac) — same discussion, same conclusion I reached.
- GeoServer's FlatGeobuf WFS output format (`FlatGeobufOutputFormat.java`) does use the bare `application/flatgeobuf` as its actual Content-Type, so the ecosystem is genuinely split — there's no single settled convention, which is consistent with the discussion's "no final recommendation."

Given the split, I'm keeping `application/vnd.flatgeobuf`: it's what the STAC-ecosystem reference implementation uses (directly relevant since this value flows into this codebase's own STAC/DCAT surfaces), it follows RFC 6838's vendor tree — the correct mechanism for an unregistered type — and it matches this codebase's own existing choice for GeoParquet (`application/vnd.apache.parquet`), which has the identical no-IANA-registration situation. Switching to the non-`vnd.` form would be internally inconsistent with that precedent for no compensating standards benefit.

Not changing this in the current diff. Flagging here in case Ian wants to make a different call than I did.

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1688, opened 2026-08-27
feat(map): support PMTiles basemaps (protocol, validator, wiring)

ishiland:
## Summary

Self-hosters can now point a basemap at a single PMTiles archive with no tile server and no API key -- a protomaps world build, or a GeoLens PMTiles export (#1686). This completes the offline basemap loop for v1.16.0: export a dataset to PMTiles, drop the file behind any static file host, and point a basemap URL at it.

Three pieces:

1. **Protocol registration** (`frontend/src/lib/maplibre-worker.ts`): adds the `pmtiles` npm package and registers maplibre-gl's `pmtiles://` protocol via `addProtocol('pmtiles', new Protocol().tile)`. Registered in this module (not `main.tsx`) for the same reason `setWorkerUrl` lives there (#1624): this module only loads as a side effect of the lazy map-vendor chunk, from every surface that constructs a map (`DatasetMap`, `SpatialFilterPanel`, `BboxMapPicker`, `ViewerMap`, `BuilderMap`). An eager import from the app entry would pull the ~295 kB map-vendor chunk into every route.

2. **Backend validator** (`backend/app/modules/settings/schemas.py`, `BasemapEntry.validate_tile_url`): now also accepts a bare `https://....pmtiles` archive URL and a `pmtiles://`-prefixed one, alongside the existing `.json` / `/styles/` / `{z}/{x}/{y}` shapes. The inner URL still has to be a well-formed http(s) URL -- nothing else about the validator loosens. `BasemapEntry` never surfaces in the OpenAPI schema (only the api-key-stripped `BasemapPublicResponse` does, and its description is unrelated to URL shape), so `openapi.json` and the generated SDKs/frontend types are unaffected -- confirmed via `make openapi-check` and a clean `git diff` on those artifacts.

3. **Basemap wiring** (`frontend/src/lib/basemap-utils.ts`, `toMaplibreStyle`): a style JSON whose sources reference `pmtiles://` URLs already renders correctly once the protocol is registered -- no code change needed, since MapLibre resolves those sources itself once fetched. For a bare archive URL configured directly as a basemap, `toMaplibreStyle` wraps it in a single-layer raster `StyleSpecification` whose source uses `url` (a TileJSON-style pointer the registered protocol resolves into synthesized metadata from the archive header) instead of `tiles` (a raw XYZ template array).

### Scope decision: (a) style JSON vs (b) bare archive

- **(a) Style JSON with `pmtiles://` sources** -- supported as-is, vector or raster, once the protocol is registered.
- **(b) Bare archive URL as a basemap** -- scoped to **raster PMTiles archives only**. A bare vector PMTiles archive carries no styling information (layer names, paint rules) for `toMaplibreStyle` to render sensibly, so a vector archive must be used via a style JSON that references it (scope a) rather than configured directly as a basemap URL.

The admin basemap form's client-side URL gate (`SettingsMapTab.tsx`) mirrors the backend acceptance so a PMTiles URL isn't blocked before it reaches the API, and the URL format hint/error copy is updated in all four locales (en/es/fr/de).

## Config example

A world basemap from a protomaps build, hosted on any static file server (no tile server, no API key):

```json
{
  "id": "protomaps-world",
  "label": "Protomaps World (offline)",
  "url": "https://example.com/basemaps/protomaps-basemap.pmtiles",
  "enabled": true
}
```

For a vector archive, publish a style JSON alongside it whose sources point at `pmtiles://https://example.com/basemaps/protomaps-basemap.pmtiles`, and configure the basemap URL as that style JSON's URL instead.

## Verification

- `cd frontend && npm run typecheck` -- clean
- `cd frontend && npm run lint` -- clean
- `cd frontend && npx vitest run` -- 417 files / 5555 tests passed, including new coverage:
  - `frontend/src/lib/__tests__/maplibre-worker.test.ts` -- protocol registration
  - `frontend/src/lib/__tests__/basemap-utils.test.ts` -- `isPmtilesArchiveUrl` + `toMaplibreStyle` archive-source mapping
  - `frontend/src/components/admin/settings/__tests__/SettingsMapTab.test.tsx` -- admin "Add basemap" acceptance
- `cd frontend && npm run test:i18n` -- clean (no new keys added; existing `urlError`/`tileUrlPlaceholder` values updated in all four locales)
- `cd backend && uv run ruff check . && uv run ruff format --check .` -- clean
- Focused backend tests (run against `.env.test` via `make env-test`): `backend/tests/test_basemap_pmtiles_validator.py` (17 new acceptance-matrix cases), `test_settings_router.py`, `test_basemap_config_drift.py`, `test_basemap_sublayer_overrides.py`, `test_layering.py` -- all pass
- `make openapi-check` -- clean; confirmed no diff on `backend/openapi.json`, the generated SDKs, or `frontend/src/types/api.generated.ts` (the changed field never surfaces in the OpenAPI schema)

Not verified: an actual browser render of a PMTiles basemap. This was implemented and tested from an isolated worktree, where the dev stack serves `main`'s frontend/backend regardless of worktree branch -- a real render check needs the post-merge stack.

## Known limitation

The vector-archive rejection (a bare `.pmtiles` basemap must be raster) is enforced in the admin Settings UI by reading the archive header before accepting a new URL -- it is **not** enforced by the backend validator, which only checks URL shape (as it does for every other basemap type: a broken `.json` style URL or a 404-ing XYZ template is accepted the same way today). An admin who writes `basemaps` directly via `PUT /settings/`, bypassing the UI, can still save a vector archive; the worst case is their own basemap renders blank, since MapLibre's registered `pmtiles://` protocol still attempts the fetch either way. Closing that gap at the API layer would mean a new server-side outbound fetch of an admin-supplied URL (SSRF review, `make_safe_client()`, Rule 2 compliance) for a config-correctness nicety on a trusted, admin-only, self-inflicted failure mode -- disproportionate to this feature's scope. Flagged by codex review; left as a documented residual limitation rather than fixed.

## Notes for reviewers / ops

- New npm dependency (`pmtiles`, plus its one dependency `fflate`): deployed frontend containers need a `-V` rebuild (`docker compose up -d --build -V frontend`) since the anonymous `node_modules` volume masks a host-only `npm ci`.
- No database migration.


ishiland:
@codex review

ishiland:
Fixed the P1 finding (bare vector PMTiles archive silently mis-rendered as raster) in e8d63b4fc: the admin "Add basemap" form now reads the archive header via pmtiles.PMTiles before accepting a bare archive URL, and rejects a vector one with a message pointing at the style-JSON path. Fails open on a header-read error since MapLibre's own protocol will attempt the same fetch at render time regardless.

ishiland:
@codex review

ishiland:
Fixed the round-2 P1 finding (api_key placeholder not substituted before probing) in c874f25d2: the header probe now substitutes {api_key} the same way router_public.py does before reading the archive, and a new test asserts the PMTiles constructor receives the substituted URL, not the literal placeholder.

ishiland:
@codex review

ishiland:
Fixed the round-3 P2 finding (replace vs replaceAll) in 36155989f: the probe now uses replaceAll for the {api_key} placeholder, matching Python's all-occurrences replace in router_public.py. Added a test with the placeholder appearing twice.

ishiland:
@codex review

ishiland:
Round 4 findings, addressed:

**Finding 2 (basemap-utils.ts:253, /styles/ vs pmtiles ordering)** -- fixed in 5ff1ec607. `toMaplibreStyle` now classifies a bare PMTiles archive before the generic `/styles/` substring heuristic, so a raster archive hosted under a path containing `/styles/` is correctly wrapped through the `pmtiles://` protocol instead of being handed to MapLibre as a raw style-JSON URL. Added a regression test for the path-collision case.

**Finding 1 (schemas.py:76, write-boundary enforcement)** -- accepted as a real gap, not fixing it in this PR. The vector-archive check lives in the admin Settings UI (reads the archive header before accepting a new URL); it is not enforced by the backend validator, which only checks URL shape -- consistent with every other basemap type today (a broken `.json` style URL or a 404-ing XYZ template is accepted the same way, no reachability/content check exists for any of them). An admin bypassing the UI via a direct `PUT /settings/` call can still save a vector archive; worst case their own basemap renders blank, since MapLibre's registered protocol attempts the identical fetch regardless of whether this check ran. Enforcing it server-side would mean a new outbound fetch of an admin-supplied URL at the API layer (SSRF review, `make_safe_client()`, Rule 2 compliance) for a config-correctness nicety on a trusted, admin-only, self-inflicted failure mode -- disproportionate scope for this PR. Documented as a known limitation in the PR body.

ishiland:
@codex review

PR #1693, opened 2026-08-27
fix(catalog): allow anonymous COG download for public+published rasters

ishiland:
## The asymmetry

`GET /datasets/{id}/export` (vector) lets an anonymous caller download a public+published dataset outright. `GET /datasets/{id}/download/cog` (raster) 401'd every anonymous caller unconditionally — even for a public+published raster whose tiles already render anonymously — unless the caller first minted a download-scoped token via `POST /auth/download-token/{id}`. That broke "download this raster and open it in QGIS" for exactly the datasets meant to be showcased.

The route's *own* access gate (`check_dataset_access_or_anonymous` + a public-visibility defense-in-depth check) was already correct and already matched the export route's — the bug was one level up, in the `_resolve_download_user` dependency: when there was no Authorization header AND no `?token=`, it raised 401 unconditionally, before that gate ever ran.

## The fix

`backend/app/modules/catalog/datasets/api/router_export.py`:

- `_resolve_download_user`: the "no auth signal at all" case now `return None` (mirroring `get_optional_user`, which is what `/export`'s dependency uses directly) instead of raising 401. `download_cog`'s existing anonymous branch — `check_dataset_access_or_anonymous(...)` then a `visibility != "public"` 403 — now actually runs for a plain anonymous GET, exactly mirroring `processing/export/router.py`'s `export_dataset_endpoint` anonymous branch. An auth signal that's actually invalid (bad token bytes, wrong `typ`/`scope`, expired, unresolvable `sub`) still 401s explicitly — that branch's implicit fall-through was made an explicit `raise` so it isn't accidentally swept into the new `return None`.
- `download_cog`'s **authenticated** branch previously inlined the per-role matrix check (`any(matrix.get(role, {}).get("export", False) for role in user_roles)`), while the export route routes the same check through `get_permission_extension().check_permission(...)`. That's a real difference: a deployment registering a custom `PermissionExtension` would get its policy applied on `/export` but bypassed on `/download/cog`. Mirrored it — `download_cog` now calls the same extension point. `DefaultPermissionExtension.check_permission` reduces to the identical matrix check, so OSS behavior is unchanged.

## Sibling-route audit

`router_export.py` (the module holding `download_cog`) registers exactly one download-shaped route pair (`HEAD`+`GET` sharing the `download_cog` handler); its other routes are DCAT/DCAT-US/GeoDCAT-AP JSON-LD feeds, not asset downloads. `_resolve_download_user` is used by no other handler in the backend (`grep -rn resolve_download_user backend/app`). No sibling instance of this asymmetry found.

`backend/tests/test_dcat.py`'s `test_dcat_raster_does_not_advertise_the_cog_download_url` (renamed from `..._token_gated_cog_download`) had a now-stale docstring premise ("`/download/cog` 401s an anonymous caller") for its public+published fixture — updated the docstring; the assertion itself (DCAT never advertises a `/download/cog` URL) is unrelated to auth and still holds.

## Tests

New `backend/tests/test_cog_download_access.py`, mirroring `test_export_access.py`'s matrix for the COG route:
- anon + public+published → 200, streams the real stored bytes, `image/tiff`
- anon + public+unpublished (`record_status=internal`) → denied (401/403/404)
- anon + private → **404** specifically (existence hiding, not 401/403)
- anon + restricted → denied (401/403/404)
- authenticated non-owner (viewer) + private → denied (401/403/404)

## Contract

No OpenAPI surface change. The route's published docstring (its OpenAPI description) is untouched on purpose — the new anonymous-access behavior is explained in a plain source comment instead, following the pattern the file already uses for its HEAD/GET docstring split, specifically to avoid churning `openapi.json`/SDKs/CLI for prose. `make openapi-check` passes clean (no diff), confirmed by running `scripts/dump_openapi.py --check` before and after.

`backend/tests/test_layering.py`'s `_MODULE_LOC_CAPS` ratchet for `router_export.py` raised 1490 → 1538 (exact), with a comment describing what the lines bought.

## Verification

```
  tests/test_cog_download_access.py tests/test_export_access.py \
  tests/test_rule1_structural.py tests/test_layering.py -q
# 67 passed

  tests/test_cog_head_ranges_1528.py tests/test_download_token.py \
  tests/test_phase_273_download_token.py tests/test_phase_273_cog_redirect_revalidate.py \
  tests/test_dcat.py tests/test_cors_range_headers_1540.py tests/test_cog_s3_head_wire_1540.py \
  tests/test_cog_vsicurl_1528.py tests/test_api_key_scope_875.py tests/test_alert_rules.py -q
# 203 passed, 4 skipped (pre-existing, unrelated)

cd backend && PYTHONPATH=. uv run python scripts/dump_openapi.py --check
```

### DIFF
--- a/backend/app/modules/catalog/datasets/api/router_export.py
+++ b/backend/app/modules/catalog/datasets/api/router_export.py
@@ -60,7 +60,7 @@ from app.core.dependencies import get_db
 from app.core.db.tenant_session import current_tenant_var
 from app.core.tenancy import is_multi_tenant
 from app.core.public_urls import get_public_urls
-from app.platform.extensions import get_catalog_port
+from app.platform.extensions import get_catalog_port, get_permission_extension
 from app.platform.http.ranges import (
     RANGE_UNSATISFIABLE,
     if_match_passes,
@@ -778,9 +778,23 @@ async def _resolve_download_user(
     downstream consumer (``download_cog``) is responsible for enforcing
     public visibility when user is None.
 
-    401 is reserved for: no auth signal at all (no header AND no token),
-    invalid token bytes, wrong typ, wrong scope, expired token, or a
-    sub-bearing token whose user no longer exists / is inactive.
+    No auth signal at all (no header AND no ``?token=``) also returns
+    ``None`` rather than raising 401: mirrors ``get_optional_user``, which is
+    what ``/datasets/{id}/export`` (``processing/export/router.py``) depends
+    on directly. Before this, a plain anonymous GET here — no Authorization
+    header, no minted token — hit the unconditional 401 below regardless of
+    the dataset's visibility, so a public+published raster's COG could not be
+    opened directly in QGIS/GDAL the way its tiles and vector export already
+    can; only a caller that first minted a download token could get through.
+    ``download_cog`` runs the same ``check_dataset_access_or_anonymous`` +
+    public-visibility gate the export route runs, so this closes that
+    asymmetry without loosening anything: a private/restricted/unpublished
+    dataset still denies (404, to hide existence) once ``download_cog``
+    applies that gate.
+
+    401 is reserved for an auth signal that is actually invalid: bad token
+    bytes, wrong typ, wrong scope, expired token, or a sub-bearing token
+    whose user no longer exists / is inactive.
     """
     if user is not None:
         return user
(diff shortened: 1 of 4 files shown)


ishiland:
@codex review

ishiland:
@codex review

Addressed both round-1 findings:
- Anchored the new inline comments to #1693 instead of the bare `fix(anon-raster-download)` branch-name label.
- `_resolve_download_user` now checks `qt is not None` instead of truthiness, so a present-but-empty `?token=` 401s through the existing `PyJWTError` path instead of falling through to the new anonymous branch. Added `test_anon_cog_download_with_empty_token_param_is_401` to pin it.

Issue #1553, opened 2026-08-17
Close the remaining gap between exported attribution and MapLibre's on-screen control


ishiland:
Follow-up home for further attribution-rendering findings after #1541, so that PR can land on its merits rather than absorb an unbounded series of refinements.

## Where #1541 got to

Nine review rounds, every finding real. Rendered map images originally carried no credits at all — not the basemap's, not the datasets'. They now:

- carry every credit the code can turn into text, wrapping rather than truncating a provider name;
- count the ones it cannot name in an explicit `+N more credits` marker, so nothing is dropped silently;
- never paint outside the canvas, and never grow the export past the tightest documented browser canvas limit rather than the most generous one;
- order credits for rendered content — visible layers, in-zoom-range layers, and the terrain source, which renders with no style layer — ahead of hidden content, so a bounded overlay elides what is not on screen first;
- resolve HTML attribution through the accessible-name rules, with `aria-label` outranking `alt` and `title` as fallback.

## What remains, and why it is here

The gap between the exported attribution and MapLibre's on-screen control is narrow but not closed, and the remaining items are all "which credit, in which order, rendered how" rather than "a credit is missing".

The priority set is also still a **proxy** for MapLibre's live `used`/`usedForTerrain` state rather than that state itself. It models layer visibility, zoom ranges and terrain usage. It does not model layer `filter` expressions, and it will drift from the control whenever MapLibre changes what counts as used. The module documents this as an approximation with its known gaps named, which is the right shape — but deriving priority from the live state, if it can be reached in the headless export path, would remove the class.

## Candidates

- Layer `filter` expressions that exclude every feature at the current view, making a source unused in practice while visibility and zoom both pass.
- Any further divergence from `AttributionControl`'s own dedupe and suppression rules as MapLibre changes them.
- `Reflect`-style indirection in attribution HTML is not a concern here, but arbitrary embedded markup is: the text derivation handles images, inline SVG and CSS-backed logos, and other embedded forms may need the same treatment.

## Related

- #1486 — the original issue.
- #1552 — the same arrangement for the web-storage gate, filed for the same reason.

ishiland:
Adding the item declined during #1541 review:

[...]

PR #1702, opened 2026-08-28
fix(builder): derive export attribution priority from MapLibre's live used flags

ishiland:
Advances #1553 (the rolling follow-up home; leaves it open).

The export/thumbnail/OG attribution priority set was a model of MapLibre's `used`/`usedForTerrain` state. This PR reads the live state where it is reachable and settles the rest of the issue's candidate list against the installed maplibre-gl 6.5.0 source.

## Candidate 1: derive priority from the live `used`/`usedForTerrain` state. Done.

The premise held. The "headless" export path is not actually headless: both capture paths (`doCapture` and `handleExportPNG` in `use-builder-save.ts`) hold the live builder `Map`, and the state the on-screen control renders from is reachable through it. In 6.5.0, `_updateAttributions` shows exactly the sources for which `map.style.tileManagers[id].used || .usedForTerrain` holds.

`shownSourceIds` now reads those flags first (`liveUsedSourceIds`), so the shown/hidden partition cannot drift from the control, including when MapLibre next changes what counts as used. Both captures run inside a render callback and the flags are recomputed per rendered frame, so they are at most one frame old. That is fresher than any model can be.

The approximation is demoted rather than deleted. `Style.tileManagers` is internal (typed, but not API) and has been renamed once already: it was `sourceCaches` until 5.24. The read is defensive. A collection that is absent, renamed again, or reshaped returns null, and the modeled tier (visibility, zoom range, terrain) answers exactly as before, instead of degrading to an everything-hidden answer. A reachable but empty collection is a live answer and is distinguished from an unreachable one. On a real map the live tier is what runs, so the approximation class of drift is gone from every real capture; removing the fallback outright would trade that for a silent regression on the next upstream rename, which is why the issue's "remove the approximation class" lands as "demote it to the degradation tier".

Ordering-not-filtering is unchanged: an unused source's credit is still rendered when there is room and elided first when there is not (the #1541-review decision). The control filters, so a bounded output with room to spare remains a deliberate superset of the control's line, which errs in the direction licensing wants.

## Candidate 2: layer `filter` expressions. Resolved by parity, with no filter evaluation.

Verified in the 6.5.0 source: `used` is set from `!layer.isHidden(zoom)` alone (`Style.update`), and `isHidden` is zoom range plus `layout.visibility`. Filters play no part. A source whose every feature is filtered out still shows its credit in the on-screen control, so parity with the control means counting it shown here too. The issue guessed this class "falls out for free if candidate 1 works", and it does, in an unexpected way: the control itself over-credits this class, and now the image does too, identically, by construction. Evaluating filters ourselves would open a divergence rather than close one. A test pins the semantics and the module comment records the finding.

## Candidate 3: dedupe/suppression alignment with current v6. Verified; one divergence documented and pinned.

`_updateAttributions` in 6.5.0 does: exact-duplicate skip at collection, whitespace-only filter, length sort, then drop any entry a later (post-sort, longer) entry `includes`. Comparing line by line:

- Strictly-longer containment suppression matches. After the sort and the exact dedupe, an equal-length distinct entry can never contain another, so our `dedupe` plus strictly-longer rule produce the same survivor set. Existing tests already cover this, including order-independence.
- The ` | ` join matches.
- The length sort is deliberately not mirrored: the shown-first order is load-bearing for elision (documented since #1541).
- One divergence is deliberate, and is now documented in the module and pinned by a test: the control compares raw HTML while this module compares derived text. Two anchors differing only in `href` render `© OSM | © OSM` on screen and collapse to one line in the image. Text is what the image renders, and both directions of the difference only ever remove a line whose text another line carries whole.

## Candidate 4: other embedded markup forms. Done.

MapLibre's control sanitizer (`DOM.sanitize`) strips only `<script>`, `on*` handlers, and `javascript:`/`data:` URLs, so `<object>`, `<embed>`, `<iframe>`, and `<video>` all render visibly in the control while deriving no DOM text. They now take the same treatment as an unnamed ``: accessible alternatives first (the generic-element rule already read `aria-label`/`title` on them), object/iframe fallback content next (which is DOM text), then the host-derived `(image credit: host)` placeholder. `data:`/`blob:` URLs fall through to the bare placeholder as before. `` keeps its own branch because `alt=""` is a decorative opt-out these elements do not have. Elements that embed nothing external (an empty `<span>`, a bare `<canvas>`) deliberately stay out of the new map; the source-level floor still counts a credit made only of them. No new i18n keys, since the two existing placeholder keys are reused.

## Remaining in #1553

- The rename fragility of `Style.tileManagers` itself. The live read degrades to the modeled tier on the next upstream rename, and nothing here can detect the rename beyond falling back. Re-verifying against new maplibre-gl majors is the standing task.
- The raw-HTML vs derived-text comparison basis, documented as deliberate; revisit only if MapLibre changes its own dedupe.
- Suppression across priority groups stays intentionally off (a hidden superstring must not displace shown text into the elision marker).

## Diff shape

Two files: `frontend/src/lib/map-image-attribution.ts` and its test file. No hook changes; `AttributionMapLike` gained an optional `style?: unknown` that the live `Map` both call sites already pass satisfies structurally. No backend contact, no i18n changes.

## Verification

- `cd frontend && npm ci && npm run lint && npm run typecheck`: clean.
- `npx vitest run src/lib/__tests__/map-image-attribution.test.ts`: 108 tests, 15 of them new (live-flag preference, live overruling a hidden layer, usedForTerrain via the flag, fallback on absent/renamed/reshaped collections, empty-collection-is-an-answer, filter parity, raw-distinct/text-identical dedupe, and the object/embed/iframe/video derivation set).
- `npx vitest run src/components/builder/hooks/__tests__/use-builder-save.test.ts`: 90 tests; the unchanged mocks exercise the fallback tier.
- `npm run test:coverage`: full suite green.

### DIFF
--- a/frontend/src/lib/__tests__/map-image-attribution.test.ts
+++ b/frontend/src/lib/__tests__/map-image-attribution.test.ts
@@ -455,6 +455,74 @@ describe('readRenderedAttribution', () => {
     expect(readRenderedAttribution(sourcesMap('   ', '\n\t'))).toEqual([]);
   });
 
+  /* feat(#1553 candidate 4): <object>, <embed>, <iframe> and <video> render
+   * perfectly visibly in the on-screen control — MapLibre's sanitizer strips
+   * only <script>, on* handlers and javascript:/data: URLs — and none of them
+   * derives DOM text. They take the same treatment as an unnamed <img>:
+   * accessible alternatives first, then the host-derived placeholder. */
+
+  it('names an unnamed object logo by its host, like an image', () => {
+    expect(
+      readRenderedAttribution(sourcesMap('<object data="https://assets.acme.com/logo.svg"></object>')),
+    ).toEqual(['(image credit: assets.acme.com)']);
+  });
+
+  it('names embed, iframe and video credits the same way', () => {
+    expect(
+      readRenderedAttribution(
+        sourcesMap(
+          '<embed src="https://a.example.com/logo.svg">',
+          '<iframe src="https://b.example.com/badge"></iframe>',
+          '<video poster="https://c.example.com/still.jpg"></video>',
+        ),
+      ),
+    ).toEqual([
+      '(image credit: a.example.com)',
+      '(image credit: b.example.com)',
+      '(image credit: c.example.com)',
+    ]);
+  });
+
+  it('lets an embed\'s accessible alternative outrank its host placeholder', () => {
+    // The generic-element rule already reads aria-label/title on these; the
+    // placeholder is only for an embed nothing names.
+    expect(
+      readRenderedAttribution(
+        sourcesMap('<object data="https://assets.acme.com/logo.svg" aria-label="© Acme 2026"></object>'),
+      ),
+    ).toEqual(['© Acme 2026']);
+  });
+
+  it('lets object fallback content, which is DOM text, win over the placeholder', () => {
+    expect(
+      readRenderedAttribution(
+        sourcesMap('<object data="https://assets.acme.com/logo.svg">© Acme Provider</object>'),
+      ),
+    ).toEqual(['© Acme Provider']);
+  });
+
+  it('keeps text around an unnamed embed rather than replacing or dropping it', () => {
+    // The inline mirror of the <img> rule: the embed is on screen beside the
+    // text, so the derivation says so.
+    expect(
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

PR #1704, opened 2026-08-28
feat(settings): restrict public visibility to admins behind an instance flag

### LINKED ISSUE #1691: feat(settings): instance setting to restrict public visibility to admins
On an open instance where SSO sign-in is enabled (like the public demo), any signed-in visitor can create datasets, maps, and collections and set them public. Public content lands in every visitor's search facets and browse UI. On the demo this produces a steady accumulation of junk ("Desktop" datasets, empty single-letter collections) that has to be pruned by hand — content was purged on 2026-08-25 and again on 2026-08-27, and new junk arrived within a day of the first prune.

## Proposal

An instance setting (default OFF, preserving current behavior for private self-hosted installs) that restricts the `public` visibility level to admins. When enabled:

- Non-admin users can still create and edit content, but visibility is capped at `internal` (or `private`, per the instance's choice).
- Attempts to set `visibility: public` on create or update by a non-admin are rejected with a clear error.
- Enforcement lives server-side at every visibility-writing path (dataset, map, collection create/update), not in UI affordances.

## Notes

- The demo would enable this, which structurally ends the junk-in-facets problem instead of treating it with prune cadence.
- Enforcement points should be enumerated rather than patched instance-by-instance: grep every handler that writes a visibility field and gate them through one shared check.
- Interacts with the existing anonymous-access contract: anonymous read of public+published content is unchanged; this only governs who may mark content public.

ishiland:
Implements #1691: an instance setting (default OFF) that restricts the `public` visibility level to admins. Built for open instances like the demo, where any signed-in visitor can currently mark junk content public and pollute every visitor's search facets.

## What changed

**Setting.** `RESTRICT_PUBLIC_VISIBILITY` (`restrict_public_visibility`, bool, default `false`) joins the persistent-config registry on the General tab. No migration — it rides the existing `app_settings` KV surface like every other flag.

**One shared gate.** `check_public_visibility_allowed(db, user, visibility, *, user_roles=None)` in `backend/app/modules/catalog/authorization.py`. When the setting is ON and a non-admin requests `public`, it raises 403 with a clear message; every other value passes through. Existing public content is untouched — the gate fires only on a mutation that requests public, so enabling the setting never retroactively narrows anything.

**Enforcement points** (enumerated, all through the one gate):

| Surface | Route |
|---|---|
| Dataset metadata PATCH | `PATCH /datasets/{id}` |
| Map update | `PUT /maps/{id}` |
| STAC import | `POST /services/stac/import` |
| Ingest commit | `POST /ingest/commit/{job_id}` |
| Ingest fan-out | `POST /ingest/commit-fan-out/{job_id}` (gates the parent job's inherited `user_metadata.visibility` — defense-in-depth; the request body itself has no visibility field) |
| Table register | `POST /ingest/register/` |
| Bulk register | `POST /ingest/register/bulk/` (one public item 403s the whole batch before anything runs) |
| VRT create | `POST /ingest/vrt/create` |
| Manifest apply (CLI `geolens apply`) | `POST /ingest/manifest/apply` — publication intent `published` maps to `public`; gated in `_classify_dataset` before staging, so it surfaces as that entry's error and dry runs report it too |

Surfaces confirmed to need **no** gate: collections (no visibility field at all), records router (reads only — `audience_visibility` is a counterfactual query param), map create (`MapCreate` has no visibility; maps are born private), map sharing routes (require an already-public map; they never write visibility).

**Structural test.** `backend/tests/test_visibility_gate_structural.py` walks the flattened FastAPI route table, recurses each POST/PUT/PATCH body-model tree for a field named `visibility`, and asserts every discovered handler calls the gate (7 discovered, all gated). The manifest and fan-out paths — whose bodies carry no `visibility` field — are pinned explicitly, and a gate-integrity test keeps the shared check from being hollowed out. `backend/tests/test_public_visibility_gate.py` covers the runtime matrix (editor 403 / admin 200 / internal ok / already-public untouched / default-off unchanged / feature-flag exposure).

**Frontend.** The flag rides the public `GET /settings/feature-flags/` bundle. A new `useCanSetPublicVisibility()` hook (feature flag + `isAdmin`) drives the affordances: the dataset Access tab disables the Public move with an explanatory note, the map SharePanel disables the Public choice and swaps its description, and the import/register pickers hide the Public option. The admin Settings → General tab gains the toggle. New `t()` keys are in all four locales (en/es/fr/de).

## API / config impact

- `FeatureFlagsResponse` gains `restrict_public_visibility` (default `false`) — additive. `backend/openapi.json`, both SDKs, and `frontend/src/types/api.generated.ts` regenerated.
- New settings key `restrict_public_visibility` (General tab). Default OFF: zero behavior change for existing installs until an admin enables it.
- Visibility-writing mutations can now return 403 (`"Public visibility is restricted to administrators…"`) when the setting is enabled — for non-admins requesting `public` only.

## Verification

```
cd backend && uv run pytest tests/test_public_visibility_gate.py tests/test_visibility_gate_structural.py -v   # 21 passed
cd backend && uv run pytest tests/test_layering.py tests/test_rule1_structural.py -q                            # 53 passed
cd backend && uv run pytest tests/test_manifest_apply_api.py tests/test_manifest_apply_service.py \
  tests/test_manifest_apply_roundtrip.py tests/test_manifest_apply_vrt.py tests/test_stac_import.py \
  tests/test_settings_router.py tests/test_settings_admin.py tests/test_api_contract_openapi.py \
  tests/test_sdks_round_trip.py -q -n 4                                                                         # 202 passed, 1 skipped
cd backend && uv run pytest tests/test_dataset_metadata_idor.py tests/test_maps.py tests/test_vrt_schema_171.py \
  tests/test_ingest.py tests/test_ingest_fan_out.py -q -n 4                                                     # 254 passed
cd backend && uv run ruff check . && uv run ruff format --check .                                               # clean
make openapi-check && make sdks-check                                                                            # clean
make public-surface-check && make deployed-surface-check                                                         # clean
cd frontend && npm run lint && npm run typecheck && npm run test:i18n                                            # clean
cd frontend && npx vitest run src/components/dataset/__tests__/AccessTab.test.tsx \
  src/components/builder/__tests__/SharePanel.test.tsx src/api/__tests__/hand-typed-mirror-drift.test.ts \
  src/components/import/__tests__/ src/components/admin/settings/__tests__/                                      # 246 passed
pre-commit run --from-ref main --to-ref HEAD                                                                     # all hooks pass
```

Closes #1691

### DIFF
--- a/backend/app/core/persistent_config.py
+++ b/backend/app/core/persistent_config.py
@@ -563,6 +563,19 @@ ENABLE_DATASET_EDITING = PersistentConfig[bool](
     label="Enable Dataset Editing",
 )
 
+# feat(#1691): when ON, only admins may set `visibility: public` on datasets and
+# maps. Default OFF so private self-hosted installs keep current behavior. The
+# server-side gate is `check_public_visibility_allowed` in
+# `modules/catalog/authorization.py`; existing public content is untouched
+# (the gate fires only on a mutation that REQUESTS public).
+RESTRICT_PUBLIC_VISIBILITY = PersistentConfig[bool](
+    key="restrict_public_visibility",
+    type_=bool,
+    env_default=False,
+    tab="general",
+    label="Restrict Public Visibility to Admins",
+)
+
 # FRONT-01 (Phase 1223): landing-first flag.  Default OFF so existing
 # self-hosters see no change.  When ON, the frontend root guard redirects
 # unauthenticated visitors from "/" to "/login" (the marketing landing page)
--- a/backend/app/modules/catalog/authorization.py
+++ b/backend/app/modules/catalog/authorization.py
@@ -536,3 +536,50 @@ async def require_dataset_editing_enabled(db: AsyncSession) -> None:
             status_code=status.HTTP_403_FORBIDDEN,
             detail="Dataset editing is disabled by the administrator.",
         )
+
+
+async def check_public_visibility_allowed(
+    db: AsyncSession,
+    user: Identity,
+    visibility: str | None,
+    *,
+    user_roles: set[str] | None = None,
+) -> set[str] | None:
+    """Enforce the `restrict_public_visibility` instance setting (#1691).
+
+    The ONE shared gate for every mutation that accepts a `visibility` value:
+    dataset metadata PATCH, ingest commit/fan-out/register/bulk-register/VRT
+    create, STAC import, manifest apply, and map update. When the setting is
+    ON, a non-admin requesting `public` gets a 403; every other visibility
+    value passes through untouched. Existing public content is unaffected —
+    the gate fires only on a mutation that REQUESTS public.
+
+    Pass ``user_roles`` when the caller already resolved them (e.g. after
+    ``check_dataset_write_access``) to avoid a second lookup. Returns the
+    resolved roles when a lookup happened (or the caller-provided set), so
+    callers can reuse them downstream; returns ``None`` untouched when the
+    fast paths made a role lookup unnecessary.
+    """
+    if visibility != "public":
+        return user_roles
+
+    # Local import mirrors the other persistent_config call sites in this
+    # module (require_dataset_editing_enabled) and avoids any import cycle.
+    from app.core.persistent_config import RESTRICT_PUBLIC_VISIBILITY
+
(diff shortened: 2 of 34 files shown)


ishiland:
@codex review

PR #1740, opened 2026-08-31
fix(ingest): refuse registration when geometry is not named geom (#1737)

### LINKED ISSUE #1737: A geometry column not named `geom` registers silently as a non-spatial table
## What happens

`register_existing_table` decides whether a table is spatial by looking for two exact column names (`backend/app/processing/ingest/service.py:731-742`):

```sql
SELECT column_name FROM information_schema.columns
WHERE table_schema = :schema AND table_name = :table_name
AND column_name IN ('geom', 'geom_4326')
```

A table whose geometry lives in any other column returns neither, so `has_geom` is false, the whole `add_4326_column` / `linearize_existing_4326` branch is skipped, and registration continues. `extract_metadata` reports srid, geometry_type and extent as None for a table with no column it recognizes, so the dataset is created as a non-spatial attribute table. No error, no warning.

`discover_unregistered_tables` does not filter these out either. Its join is `LEFT JOIN geometry_columns ... AND gc.f_geometry_column = 'geom'` (`service.py:127`), so the table appears in the discovery list with `geometry_type` and `srid` null. It looks like a legitimately non-spatial table, which is a shape the register path supports on purpose (#1359).

## Why this is easy to hit

`ogr2ogr -f PostgreSQL` names its geometry column `wkb_geometry` by default. So the most obvious way to get a table into the data schema produces a table that registers successfully and renders nothing, and the only way to find out is that the map is empty. Getting it right needs `-lco GEOMETRY_NAME=geom`, which is not documented anywhere I can find in this repo.

The same applies to a table restored from a dump written by any tool that uses a different convention (`the_geom` from older MapServer/GeoServer stacks, `shape` from Esri exports).

## Suggested fix

At registration, when neither `geom` nor `geom_4326` is present, ask `geometry_columns` whether the table has a geometry column under some other name, and refuse with a message naming it:

> Table 'facilities' has a geometry column named 'wkb_geometry'. GeoLens reads geometry from a column named 'geom'. Rename it, or re-export with `-lco GEOMETRY_NAME=geom`.

That is a few lines beside the two `ValueError`s already in that function (the retired-name refusal at `:726` and the duplicate-registration refusal at `:703`), and it keeps the deliberate non-spatial path intact: a table with no geometry column anywhere still registers as an attribute table.

Whether discovery should also flag these is a separate call. Listing them with a marker seems better than hiding them, since hiding a table an operator can see in psql is its own confusion.

ishiland:
Closes #1737.

`register_existing_table` decided "is this spatial?" by looking for the exact
column names `geom`/`geom_4326`. A table whose geometry lived under any other
name left `has_geom` false, skipped the `add_4326_column` branch, and still
registered: `extract_metadata` reports srid, geometry_type and extent as None
for a column it does not recognize, so the dataset was catalogued as an
attribute table with no error anywhere.

Discovery hid it too. Its `geometry_columns` join is a LEFT JOIN pinned to
`f_geometry_column = 'geom'`, so the table appears in the picker looking
legitimately non-spatial.

`ogr2ogr -f PostgreSQL` names its geometry column `wkb_geometry` by default,
which made the most ordinary way of landing a table in the data schema produce
a dataset that rendered nothing and explained nothing.

## Change

When neither name is present, probe `geometry_columns` for a geometry column
under some other name and refuse with the one that was found:

> Table 'facilities' stores geometry in a column named 'wkb_geometry'. GeoLens
> reads geometry from a column named 'geom'. Rename the column, or re-export
> the table with `-lco GEOMETRY_NAME=geom`.

The deliberate non-spatial path from #1359 is untouched: a table with no
geometry column anywhere has no `geometry_columns` row, the probe finds
nothing, and registration proceeds as before. Both halves are pinned in
`backend/tests/test_register_geom_column_name_1737.py`.

## Verification

```
uv run pytest tests/test_register_geom_column_name_1737.py \
  tests/test_service_column_info_1359.py::test_register_non_spatial_table_populates_column_info -q
3 passed

47 passed

uv run ruff check / ruff format --check    clean
```

Counterfactual: with the new branch replaced by `if False:`, the refusal test
fails and the no-geometry test still passes, so the guard is what the first
test is measuring and it is not propping up the second.

## Notes

No schema change, no API change, no new dependency. `service.py`'s cap in
`test_layering.py` goes 1376 -> 1402 for the probe, commented in place.

### DIFF
--- a/backend/app/processing/ingest/service.py
+++ b/backend/app/processing/ingest/service.py
@@ -741,6 +741,32 @@ async def register_existing_table(
     has_geom = "geom" in geom_cols
     has_4326 = "geom_4326" in geom_cols
 
+    # fix(#1737): a spatial table whose geometry lives under any other name
+    # used to register SILENTLY as a non-spatial attribute table -- the branch
+    # below is gated on `has_geom`, and extract_metadata reports srid/
+    # geometry_type/extent as None for a column it does not recognize. The
+    # non-spatial registration path is deliberate (#1359), so the two cases
+    # have to be told apart here rather than by widening that path: a table
+    # with NO geometry column anywhere is still a legitimate attribute table.
+    # `ogr2ogr -f PostgreSQL` names its column `wkb_geometry` by default, so
+    # the most ordinary way to land a table in the data schema hit this.
+    if not has_geom:
+        other_geom = await session.execute(
+            text(
+                "SELECT f_geometry_column FROM geometry_columns "
+                "WHERE f_table_schema = :schema AND f_table_name = :table_name "
+                "LIMIT 1"
+            ).bindparams(schema=_schema, table_name=table_name)
+        )
+        found = other_geom.scalar_one_or_none()
+        if found is not None:
+            raise ValueError(
+                f"Table '{table_name}' stores geometry in a column named "
+                f"'{found}'. GeoLens reads geometry from a column named "
+                "'geom'. Rename the column, or re-export the table with "
+                "`-lco GEOMETRY_NAME=geom`."
+            )
+
     from app.processing.ingest.tasks_common import (
         _current_tenant_role,
     )
--- a/backend/tests/conftest.py
+++ b/backend/tests/conftest.py
@@ -218,6 +218,7 @@ _TENANCY_GLOBAL_STATE_MODULES = {
     #     concurrently with each other or the group above (pg_shdepend contention).
     "test_analysis_materialize",  # register_existing_table → grant_reader_access
     "test_registered_delete_detach_1452",  # same: registers real tables
+    "test_register_geom_column_name_1737",  # same: its accept-case registers
     "test_embed_tokens",
     "test_features_crud",
     "test_features_geojson_z",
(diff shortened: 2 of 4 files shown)


ishiland:
@codex review

ishiland:
Fixed in `7531fb57c7`. Verified the premise before acting: `conftest.py:218-221` keeps an explicit section for "cluster-global `geolens_reader` role GRANT/REVOKE mutators", with `test_analysis_materialize` annotated `# register_existing_table → grant_reader_access` and `test_registered_delete_detach_1452` as `# same: registers real tables`. The accept-case test here registers a real table and reaches `grant_reader_access` at `service.py:784`, so it belongs in that set. Added beside them with the same annotation.

Exactly the class of thing that hides from a local run: I ran the file on its own, where there is no second worker to contend with.

One thing found while checking, not fixed here. `test_service_column_info_1359` registers a table on the same path (`test_register_non_spatial_table_populates_column_info`) and is also absent from the set. Looks like the same latent hazard, but widening this PR for it seemed wrong; flagging it for a separate call.

ishiland:
@codex review

Issue #1715, opened 2026-08-30
No admin path to reset a user's password, though the login page promises one


ishiland:
## The gap

The login page tells a locked-out user to ask an administrator:

> Need access or a password reset? Contact a GeoLens administrator.

(`frontend/src/i18n/locales/en/auth.json:87`, `supportHint`)

An administrator has no way to do it. Nothing in the API can set another
user's password:

- `PATCH /admin/users/{user_id}` (`UserUpdate`) accepts `email`, `is_active`,
  `status`, `role`. No password field, so the admin UI cannot expose one either.
- `POST /auth/change-password/` (`ChangePasswordRequest`) requires
  `current_password` — self-service only, useless to someone locked out.
- There is no forgot-password, reset-token, or invite flow. The only `/reset/`
  routes in `backend/openapi.json` are for dataset attributes and settings.
- `POST /admin/users/` accepts a `password`, but only at creation.

## Why the workarounds do not close it

**Delete and recreate destroys data.** `AdminService.delete_user`
(`backend/app/modules/admin/service.py:664`) is a hard delete whose docstring
reads "FK SET NULL handles audit_logs, datasets, ingest_jobs automatically".
So the user's datasets survive as **ownerless rows**, their `UserRole` and
`ApiKey` records are deleted outright, and the recreated account gets a new
UUID that reclaims none of it.

**SSO only helps if it is configured.** An instance running password login
alone — the default — has no second door.

**That leaves psql.** Hashing is bcrypt via `pwdlib`'s `BcryptHasher`
(`backend/app/modules/auth/providers/local.py`), so recovery means generating a
hash by hand and writing it into `users.password_hash`, bypassing the app's own
validation and audit trail.

## Why this matters more than it looks

This is a self-hosting failure mode, not an enterprise one. A single-admin
instance whose admin forgets the password is unrecoverable without database
access, and the product's own UI sends them to a person who cannot help. The
copy promises a capability that does not exist.

## Suggested shape

An admin-initiated reset on the existing user-management surface, audited like
the other lifecycle actions (`user.create`, `approve`, `deactivate`, `reject`
are already in `AUDIT_ACTIONS`). Either:

1. Accept `password` on `PATCH /admin/users/{user_id}`, gated owner-or-admin
   and force-rotating the user's sessions; or
2. A dedicated `POST /admin/users/{user_id}/reset-password/` returning a
   one-time credential the admin hands over, which pairs better with a
   "must change on next login" flag.

[...]

PR #1741, opened 2026-08-31
fix(auth): stop the login page promising a password reset nobody can do (#1715)

ishiland:
Partial mitigation for #1715. **Does not close it.**

The login page tells a locked-out user:

> Need access or a password reset? Contact a GeoLens administrator.

An administrator cannot do the second half. `UserUpdate` has no password
field, `POST /auth/change-password/` requires the current password, and there
is no forgot-password, reset-token or invite flow. The workarounds all cost
something: `delete_user` is a hard delete, so the user's datasets survive as
ownerless rows while their `UserRole` and `ApiKey` records are dropped.

This removes the promise, in all four locales. Asking an administrator is
still the right first step for access, and a login page is not the place to
document operator-level recovery.

I am deliberately not building the reset flow here. That spans `UserUpdate`,
an admin UI control, session and API-key invalidation semantics, and four
locales, on an auth path. #1715 stays open for it. Fixing the copy is the
part that costs nothing to get right today.

Also adds the opening `¿` the Spanish string was missing.

## Verification

```
npm run test:i18n     4 passed   (locale parity; no keys added or removed)
npm run lint          clean
npm run typecheck     clean
```

No key changes, so nothing to translate beyond the four edited strings, and
no risk to the parity gate.

### DIFF
--- a/frontend/src/i18n/locales/de/auth.json
+++ b/frontend/src/i18n/locales/de/auth.json
@@ -84,7 +84,7 @@
   "configLoadError": "Registrierungseinstellungen konnten nicht geladen werden. Bitte erneut versuchen.",
   "hidePassword": "Passwort ausblenden",
   "showPassword": "Passwort anzeigen",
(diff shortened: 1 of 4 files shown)


ishiland:
@codex review

Issue #1712, opened 2026-08-30
fix(import): switching tabs mid-import strands the job on the Upload, Service and STAC tabs


ishiland:
Follow-up to #1705 / PR #1708 (siblings: #1710, #1711).

PR #1708 fixed this for the URL tab only. The same shape exists on the other Import tabs, and it was found while fixing that one — filing it so the exposure does not evaporate with the PR.

## Mechanism

`frontend/src/pages/ImportPage.tsx` renders each tab conditionally:

```tsx
{activeTab === 'upload' && <UploadForm onPhaseChange={setUploadPhase} />}
{activeTab === 'url' && <UrlImportForm />}          // fixed in #1708
{activeTab === 'register' && <RegisterForm />}
{activeTab === 'service' && <ServiceUrlForm />}
{activeTab === 'stac' && <StacImportForm />}
```

Switching tabs therefore **unmounts** the active form. None of the unfixed forms has unmount handling. Any request already in flight keeps running — the server does not stop work when a client disconnects — so its response lands in dead component state and whatever it created becomes unreachable from the UI.

Aborting on unmount is **not** a fix on its own: the server keeps going regardless, so an abort strands the same server-side state *and* discards the identifier needed to find it. The response has to be kept, not dropped.

## What each tab actually strands

Worth distinguishing, because the severity is not uniform:

- **Upload** (`UploadForm`) — the direct analogue of the URL case, and the one worth fixing first. `uploadFile` / `uploadPresigned` return a `job_id` for a job that is `pending` with **staged bytes** (local staging file, or an S3 staging object). Unmount mid-upload and that job plus its bytes are unreachable until the stale-pending sweep collects them — `pending_job_timeout_seconds`, default 3600s. Repeated switches accumulate them.
- **Service** (`ServiceUrlForm`) and **STAC** (`StacImportForm`) — `previewServiceLayer` returns a `job_id` too, so an unmount mid-preview strands a `pending` job. No staged file (these jobs carry `source_url` rather than `file_path`), so the cost is the job row rather than storage; the same sweep collects it.
- **Register** (`RegisterForm`) — mildest. `bulkRegister.mutateAsync` creates datasets synchronously; losing the response loses the confirmation UI, not the work, and the datasets are discoverable in the catalog. Arguably a UX papercut rather than a leak.

## The fix shape

[...]

PR #1742, opened 2026-08-31
fix(import): keep a visited tab mounted so a switch cannot strand its job (#1712)

ishiland:
Partial fix for #1712. **Does not close it** — see the scope note below.

## The change

The Import panels rendered as `activeTab === x && <Form />`, so a tab switch
mid-import unmounted the running form. The request carried on, because the
server does not stop when a client disconnects, and its response landed in
dead component state: an unreachable job plus staged bytes waiting on the
stale-pending sweep.

A tab is now mounted on first visit and never unmounted. An in-flight XHR's
progress callbacks keep writing into live state because the component is
still there, so the whole class of bug goes away without touching any form's
internals.

Two details the change forces:

- Inactive panels use `hidden`, which keeps them out of the a11y tree and the
  tab order while leaving them alive.
- The error boundary moves inside the loop, one per panel. A single shared
  boundary would let a hidden tab's error blank the tab the user is looking at,
  which is a blast radius the old one-panel-at-a-time render did not have.

Mount-on-first-visit rather than mounting all five: these forms do real work
when they mount, and `RegisterForm` would list unregistered tables for a tab
nobody opened.

## Why not the module-session port the issue describes

The issue prescribes porting #1708's out-of-React session to the remaining
tabs. That is the right long-term shape and it covers strictly more, including
navigating away from the Import route entirely. It is also a much larger piece
of work than it sounds: `UploadForm` is 548 lines tracking a *batch* of
entries with per-entry progress callbacks writing into component state, and
the single-job URL version took ~25 review rounds to get the lifecycle right.

Not what I would land on the highest-traffic import path days before a
release. This closes the reported failure, tab switching, for Upload, Register,
Service and STAC in one file and ~20 lines, and it cannot regress any form's
internals because it does not touch them.

**#1712 stays open** for the batch session work and for route-navigation
survival.

## Verification

```
npx vitest run src/pages/__tests__/ src/components/import/
  39 files, 235 passed

npm run lint       clean
npm run typecheck  clean
```

New `ImportPage.keep-mounted.test.tsx` pins three properties. Counterfactual
against the old conditional render: "keeps a visited panel mounted" and "hides
the inactive panel from the a11y tree" both fail, so they measure this change;
"does not mount a panel the user has never opened" passes either way, which is
correct, since it guards behaviour this preserves rather than introduces.

The existing `ImportPage.layout.test.tsx` passes unmodified.

### DIFF
--- a/frontend/src/pages/ImportPage.tsx
+++ b/frontend/src/pages/ImportPage.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import { useCallback, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { Upload, Link, Database, Globe, Satellite } from 'lucide-react';
 import { PageHeader } from '@/components/layout/PageHeader';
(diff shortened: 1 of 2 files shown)


ishiland:
Codex is right on both, and they undercut the reason I chose this design. Verified both premises in the code rather than taking them on analogy.

**Quota (P2 #1).** Worse than it reads. `use-ingest.ts:149-156` documents `staleTime 0` + `refetchOnMount: 'always'` as the fix for a prior Codex P2 on PR #274, with the stated purpose of re-reading the live count after an import or delete elsewhere. Keeping the panel mounted defeats that for tab switches, and `RegisterForm` creates datasets, so register a table then switch to Upload and `FileDropzone` enforces a quota that is already spent. I regressed a previously-fixed issue.

**Portalled dialogs (P2 #2).** Confirmed. `DialogPortal` renders to `document.body`, so `hidden` on my wrapper cannot reach it, and `BulkTrackingList.tsx:103-111` auto-opens `VrtCreateDialog` from a completion effect. A finishing upload can drop a modal over whichever tab the user moved to.

The scope check that follows: only `UploadForm` has dialogs (8 sites; the other four forms have none), and only the auto-opening ones are reachable while hidden, since a hidden panel takes no clicks. So it is fixable. But fixing it means threading an `active` prop into `UploadForm` and `BulkTrackingList` and gating the auto-open paths, plus invalidating the upload config on activation, plus tests for both.

That is the whole argument for this PR gone. I picked keep-mounted over porting the #1708 session because it was ~20 lines in one file and touched no form internals. It is now neither. And the failure it introduces (a modal hijacking the screen, a stale quota gate) is more visible to a user than the one it fixes (a stranded background job that the stale-pending sweep collects).

My PR description says this "cannot regress any form's internals because it does not touch them." That was wrong. It does not touch their code, but it changes their lifecycle, and portals leave the subtree entirely.

Recommending this be closed rather than fixed forward, and #1712 done properly with the module session after the release. Moved to draft so it cannot land in the meantime.

ishiland:
Closing rather than fixing forward, per the analysis above.

The design traded a stranded background job for a modal that can hijack the screen and a quota gate that reads stale, and the "20 lines, touches no form internals" justification for preferring it over the #1708 session port no longer holds once `active` has to be threaded into `UploadForm` and `BulkTrackingList`.

#1712 stays open. The right fix is the module-scoped session the issue describes, done after the release with room to get the batch lifecycle right. The branch `fix/1712-import-tabs-keep-mounted` is left in place if any of it is useful later.

Issue #1229, opened 2026-08-05
docs(positioning): source modes and freshness docs page plus hub language pass


ishiland:
Rides the release that ships this milestone; positioning follows capability, so none of this lands before the features do. Best taken by whoever wants a low-risk, high-visibility change; no backend knowledge required beyond reading #1217.

## Scope

- Docs site: a "Where your data lives" concept page covering the five origins (Uploaded, Created, Registered table, Service, STAC), owned-copy vs registered-in-place vs remote-reference semantics, refresh, freshness, and health. Tracked here; the page itself lands in the docs repo during the release's site-sync pass.
- README: fold source modes and freshness into the feature list and "Why GeoLens", and nudge the one-liner toward the data-hub framing. No federation claims: external-database registration does not exist and is out of scope per #1217.
- EDITIONS.md: one clarifying line that manual refresh is Community while stored credentials and scheduled sync are paid, consistent with the existing pledge that free capabilities stay free.
- Screenshots refreshed if the Source panel changes the dataset-page hero.

## Where to look

- The accurate semantics to describe, straight from the code (do not trust older prose): service imports are one-shot copies; "Register Table" serves a table already in GeoLens's own database in place; STAC datasets reference the remote asset live. The import-page copy in `frontend/src/i18n/locales/en/import.json` (keys around 236-249) states the copy-vs-in-place distinction well and is the tone to match.
- README sections to touch: the one-liner (:5-7), "Why GeoLens?", and the features list. Keep the existing voice; it is deliberately concrete.
- EDITIONS.md: the free/paid paragraph (~:17) is where the one-line clarification belongs.
- Root-doc rules: no new root markdown files, no root `docs/` directory; detailed product docs live on the docs site (see AGENTS.md "Keep root repository docs single-purpose").

## Suggested path

1. Draft the docs-site page early (it can be reviewed while features are in flight) but merge it only in the release's site-sync pass, since the site tracks released product.
2. README and EDITIONS changes go in one small PR against the release branch window, after the Source panel merges, so screenshots and claims match reality.
3. Get an edit pass on all new copy before the PR is marked ready.

## Acceptance

- [ ] Docs page live in the release's site-sync pass
- [ ] README and EDITIONS updates land in the release PR
- [ ] Copy makes no claim the code cannot back (specifically: no external-database federation, no scheduling in Community)
- [ ] Screenshots current if the dataset page hero changed

## Verify

[...]

Issue #1227, opened 2026-08-05
feat(cli): geolens refresh command and freshness in status output


ishiland:
Rewritten 2026-08-09. The previous body predated the accepted ADR-002 Amendment 1 and the landed run ledger (#1274) and refresh endpoint (#1277), and contradicted both; see the edit history for the original. Nothing blocks this issue anymore: the endpoint, the run ledger, and both generated SDK methods exist on `main`.

The CLI has no refresh verb. Re-pulling a source today means editing a manifest so its fingerprint changes and re-running `geolens apply` (the `update` branch of `_classify_dataset` in `backend/app/processing/ingest/manifest_service.py`), which is right for CI and wrong for a human at a terminal.

## Settled decisions (do not re-litigate in the PR)

1. **Two new top-level commands**, matching the existing flat typer tree in `cli/geolens_cli/main.py`:
   - `geolens refresh DATASET_ID [--wait] [--timeout SECONDS] [--json] [--token-stdin]`
   - `geolens status DATASET_ID [--json]` — read-only; shows origin kind, origin pointer, source freshness, source health, `last_refreshed_at`, `last_checked_at`, and the most recent refresh run.
2. **`--wait` polls the run ledger, not the jobs API.** The 202 from `POST /datasets/{id}/refresh` returns both `run_id` and `job_id`. Jobs purge on retention (30-day default), and terminal detail the CLI needs (`error_code`, `error_message`, and the `cancelled` state, which has no job equivalent) exists only on the run row. Poll `GET /datasets/{id}/refresh-runs` and match `run_id`. Do not reuse the `GET /jobs/{job_id}` pattern from `publish.py`.
3. **No value-carrying `--token` flag, ever.** Argv lands in shell history, `ps`, and CI logs, which defeats the one-time credential handoff the endpoint feeds. `--token-stdin` reads the secret via `_read_secret_from_stdin()` (`cli/geolens_cli/main.py:345`, already used by `login`). Never echo it, never log it, never include it in `--json` output or error text.
4. **No dispatch retry.** The refresh endpoint has no idempotency key, so a re-POST after a timeout can queue a duplicate run (deduplicated only if the first is still active). On a dispatch timeout the CLI prints the situation and tells the user to run `geolens status`; it must not re-POST automatically.
[...]

PR #1743, opened 2026-08-31
docs(cli): drop the cron blessing that contradicts ADR-002 A9

ishiland:
`cli/README.md` called `geolens refresh` "the supported command for scheduled
cron or CI refresh jobs". ADR-002 §A9 decided the opposite and named this exact
surface:

> no GeoLens surface publishes a cron or automation recipe. That covers #1227's
> CLI docs, #1229's positioning copy, and the docs site alike. Where automation
> copy is unavoidable, it states that unattended refresh is unsupported until
> verification ships.

The README line landed after that decision, so the README is the side that is
wrong. No ADR change needed.

## What changed

Gate 1 still stands, so this does not pretend the API is closed. `--wait` and
`--timeout` exist for automation and stay documented as such. What goes is the
word "supported", replaced by the statement §A9 asks for.

Invariant 9 is why it matters:

> The invariant's protection survives v1 only because **every refresh is
> human-triggered and human-reviewable**. The human is the schema gate. That is
> a real safeguard and it is also the entire safeguard.

There is no completeness verification and drift does not block the swap, so a
reader who wires this into cron on the strength of that sentence has removed
the only thing standing between a truncated source and their live data.

## Sweep

Checked the other public surfaces for the same shape. What is left is fine and
untouched:

- `RUNBOOK.md`, `.github/ARCHITECTURE.md` — the backup service's own schedule.
- Generated SDK docstrings — the ops-only stale-job cleanup endpoint.
- `EDITIONS.md:17` — names scheduled refresh as a paid capability, which is the
  gate-4 edition boundary rather than an automation recipe.

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -50,6 +50,11 @@ thing standing between a truncated or reshaped source and live data. Nothing in
 the API stops you calling this from a scheduler, but until completeness
 verification ships, you are the check.
 
+`--wait` reports the job's terminal status, not what changed in the data. Drift
+is recorded on the dataset: read `schema_drift_status` from `GET /datasets/{id}`,
+or open the dataset's Source panel in the web app. Neither `geolens refresh` nor
+`geolens status` surfaces it today.
+
 Protected services can receive a transient credential with `--token`. Use bare
 `--token` to open a hidden-input prompt, which keeps the value out of terminal
 output and shell history. Supplying a token value directly is supported for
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review

ishiland:
Fixed in `a4402f3ec3`, with one correction to the suggested remedy.

The finding is right: the paragraph told operators they are the schema gate without saying where to look, and `--wait` reports only the job's terminal status while `geolens status` omits drift entirely. Unactionable as written.

The suggested destination does not work though. `/datasets/{id}/refresh-runs` exists but carries no drift field — `DatasetRefreshRunResponse` has no drift property in `openapi.json`. The field is `schema_drift_status` on `DatasetResponse`, surfaced in the UI by `SourcePanel.tsx`. The new text points at `GET /datasets/{id}` and the Source panel, and says plainly that neither CLI command shows it today.

ishiland:
@codex review

Issue #1746, opened 2026-09-01
ArcGIS authenticated-service import: 17 findings from the 2026-09-01 live test


ishiland:
Tracking issue for the findings from the 2026-09-01 live import of an org-only ArcGIS Online FeatureServer (probe -> preview -> commit, then refresh and reupload with and without a credential store, plus a static trace of the WFS/OGC bearer path). One PR per lane; each PR lists the finding numbers it closes.

Established facts the fixes rely on:
- ArcGIS Online `generateToken` now requires `client=referer` plus a `referer` value. A `requestip` token passes the portal but every hosted service on `services6.arcgis.com` returns 498. A referer-bound token works on hosted services with or without a `Referer` header. The HTML form at `/sharing/rest/generateToken` no longer renders.
- Viewer-tier accounts cannot mint API keys; generateToken (15-day max) is their only route.
- ArcGIS returns 498/499 error envelopes inside HTTP 200 bodies.
- With `REDIS_URL` set, import commit, refresh and reupload commit all stash the token and pass a `credential_ref`; the token never reaches the job row.

Findings:

Security
1. httpx logs the ArcGIS token in plaintext at INFO (`HTTP Request: GET ...?token=...`) in api and worker. The structlog redactor is key-based and never sees the message string.
2. Failed Procrastinate rows keep the raw token forever when no credential store is configured (the default). `delete_jobs="successful"` never touches failed rows and nothing sweeps them. Procrastinate's start/end log lines also print `task_kwargs` with the token.
10. The `cloud-dev` Valkey service cannot start: `cap_drop: ALL` breaks the entrypoint's `setpriv` (exit 127). This is what makes finding 2 the default posture.
14. Import-commit and reupload doors skip the strict header-token policy that the refresh door applies (PR #1277), so a WFS/OGC token with `+` or `/` burns the single-use credential and fails in the worker.
15. Preview accepts what commit rejects for WFS/OGC (weak validator at preview).

User-facing
3. Refresh dead end for auth-required sources: a credential-less refresh is accepted (202) and fails 0.5 s later with import-door copy; with a token on a default install it returns 503 `credential_store_unavailable`. No dataset-level auth-required marker exists.
12. Health probe false positive: `POST /datasets/{id}/source-health/` with no token reports `healthy` for an org-only service because the error envelope rides on HTTP 200.
4. Token input is `type="password"` with no `autocomplete`, so Chrome offers the admin password and asks to save the token.
[...]

PR #1747, opened 2026-09-01
fix(compose): run the cloud-dev valkey service as the valkey user

ishiland:
The `cloud-dev` valkey service never starts on main. Both compose files give it `cap_drop: [ALL]`, and the image entrypoint runs `setpriv` to drop from root to the valkey user. Without CAP_SETUID that call dies with "setresuid failed: Operation not permitted" and the container exits 127. Because the credential store cannot come up, every self-hosted install runs without `REDIS_URL`, which is what leaves raw service tokens in failed job rows (finding 2, separate PR).

This adds `user: "valkey"` to the valkey block in `docker-compose.yml` and `docker-compose.prod.yml`. The entrypoint skips `setpriv` when it is already non-root, so the hardening stays as it was. Nothing else in either file changes.

Closes finding 10 of #1746.

Verification, run from the branch on a throwaway compose project:

```
docker compose -p gl1746valkey --profile cloud-dev up -d --no-deps --wait valkey
docker compose -p gl1746valkey logs valkey | tail -5
docker compose -p gl1746valkey --profile cloud-dev down -v
docker compose -f docker-compose.prod.yml --profile cloud-dev config --services
```

The container reports healthy and logs "Ready to accept connections tcp". `backend/tests/test_phase_272_compose.py` checks the valkey healthcheck and port binding only, so no test changed.

Config impact: the cloud-dev valkey container now runs as the image's `valkey` user instead of starting as root and dropping privileges. `cap_drop: [ALL]` is unchanged.

### DIFF
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -759,6 +759,8 @@ services:
     logging: *bounded-logs
     image: valkey/valkey:8.1.7-alpine
     profiles: ["cloud-dev"]
+    # fix(#1746): cap_drop ALL removes CAP_SETUID, so the entrypoint's setpriv fails; a non-root user skips setpriv.
+    user: "valkey"
     security_opt:
       - no-new-privileges:true
     cap_drop:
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

PR #1751, opened 2026-09-01
fix(ingest): keep GDAL header tempfiles in staging and carry the ArcGIS feature count into preview

ishiland:
Two contained fixes from the ArcGIS authenticated-import test.

Preview dropped the feature count. Probe reports `feature_count` per layer (9,567 on the test layer) but the ArcGIS preview hardcoded `None`. `fetch_arcgis_layer_preview` in `adapters/arcgis.py` now calls the existing `fetch_arcgis_feature_count` helper (same safe client, one `returnCountOnly=true` request) and degrades to `None` if that request fails. The preview response schema already had the field, so `openapi.json` and the SDKs are unchanged.

GDAL bearer header files were written with `mkstemp` and no `dir=`. In a deployed process this was mostly latent, because `main.py` and `worker.py` both call `redirect_tempfile_to_staging` at boot, but a direct call or a future refactor lands the file in the system tempdir, and a SIGKILL or OOM between write and the `finally` unlink leaves it behind either way. `run_ogr2ogr_service` in `ogr.py` now passes `dir=settings.upload_staging_dir`, and a new `sweep_stale_gdal_header_files` in `core/runtime/staging.py` unlinks `gdal_auth_*.hdr` files older than an hour. It runs at the three places `sweep_orphaned_exports` already runs: API boot, the API's periodic sweeper, and worker boot. The matching `preview.py` line is in the lane D PR, which edits the same block for the strict token check.

Closes finding 7 and the `ogr.py` half of 16 of #1746.

Verification, from `backend/` with `.env.test` sourced:

```
uv run pytest tests/test_staging_runtime.py tests/test_preview_token_sec021.py tests/test_ogr_subprocess_env.py tests/test_rule2_structural.py tests/test_layering.py -q
make openapi-check
```

`test_ogr_subprocess_env.py` calls `run_ogr2ogr_service` directly and bypasses the client fixture that points `upload_staging_dir` at a tmp dir, so it gained an autouse fixture doing the same. `ogr.py` and `api/main.py` caps in `_MODULE_LOC_CAPS` are raised to the measured line counts.

No schema or API change. No config impact.

### DIFF
--- a/backend/app/api/main.py
+++ b/backend/app/api/main.py
@@ -46,6 +46,7 @@ from app.core.runtime.staging import (
     ensure_staging_ready,
     sweep_orphaned_exports,
     sweep_orphaned_write_scratch_occasionally,
+    sweep_stale_gdal_header_files,
 )
 from app.platform.extensions.bootstrap import (
     assert_enterprise_ports_resolved,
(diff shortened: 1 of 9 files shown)


ishiland:
@codex review

Issue #1755, opened 2026-09-01
Follow-ups from the #1746 authenticated-service import wave


ishiland:
Items noticed while fixing #1746 that were out of scope for those PRs. None is a regression from the wave; each is recorded here so review threads have a stable anchor.

Deferred by design

1. Refresh door, race branch (#1754): when a token-less refresh of an unmarked ArcGIS dataset is admitted and an authenticated re-upload of the same origin marks it inside the reservation window, the post-reservation recheck refuses with 422 `service_token_required` without a probe. If that re-upload used a token the service did not need, this is one spurious 422; the immediate retry goes through the normal pre-reservation probe and succeeds. Probing inside the window would hold the run slot and a pooled DB connection across network I/O, which is what the round-2 finding on that PR asked us not to do. Accepting one 422 in a race of that width.
2. The probe endpoint (`POST /services/probe`) still accepts WFS/OGC API Features tokens outside the header-token charset, because its httpx bearer path carries them safely; preview and commit now refuse them (#1752). Users learn at preview rather than at probe. Product call whether to refuse at probe as well.
3. `_validate_safe_token` remains the pydantic validator on `ServicePreviewRequest`, `ProbeRequest` and `ReuploadCommitRequest`, while the strict header-token policy runs at every door. Two rules on one field, now confined to the schema layer. Worth consolidating.

Frontend

4. `error-map.ts` does not map the new 422 `service_token_required` (#1754), and `SourceRefreshAction.tsx` does not special-case it, so the dialog renders the message generically and offers no inline token field.
5. `SettingsMapTab.tsx` and `SettingsAuthTab.tsx` have `type="password"` fields for admin secrets without the autocomplete and password-manager opt-out hints added to the service-token inputs in #1750.

Copy

6. The `unauthorized` source-health detail copy says the source "now requires access", which reads wrong for a service that was always private. `auth_required` (#1754) covers only the ArcGIS envelope case.
7. `origin_probe.py`'s module docstring still describes the three-value vocabulary and the 401/403 split only; it does not mention that the ArcGIS branch now parses a provider body.

Backend hygiene

8. `alembic/env.py` calls `logging.config.fileConfig()` with the stdlib default `disable_existing_loggers=True`. Once the test-DB migration fixture has run, any already imported third-party logger (`httpx` for one) is silently `.disabled` for the rest of the session. `test_1746_stdlib_log_redaction.py` resets the flag in a fixture; nothing else does.
[...]

PR #1769, opened 2026-09-02
fix(settings): password-manager attributes and unauthorized copy (#1755 items 5, 6)

ishiland:
Addresses items 5 and 6 from #1755.

Item 5: SettingsMapTab's basemap API key inputs (the per-row edit field and the new-basemap field) and SettingsAuthTab's OAuth client secret input are `type="password"` fields for admin secrets, but did not have the autocomplete and password-manager opt-out attributes the service-token inputs gained in #1750. All three now carry `autoComplete="new-password"` plus `data-1p-ignore`, `data-lpignore="true"`, and `data-bwignore`, matching the exact set from #1750.

Item 6: the `unauthorized` source-health detail copy said the source "now requires access", which misdescribes a service that was always private (the common case for a general 401/403, distinct from the ArcGIS-specific `auth_required` code added in #1754). Rewritten in all four locales to state the refusal without implying a change: the source refused the request because it requires credentials the dataset does not carry, and to refresh with credentials to continue.

No schema, API, or config impact. Frontend only.

Verification, from `frontend/`:
- `npm run typecheck` - passed
- `npm run lint` - passed
- `npm run test:i18n` - passed (4 tests)
- `npx vitest run src/components/admin/settings/__tests__/SettingsMapTab.test.tsx src/components/admin/settings/__tests__/SettingsAuthTab.test.tsx src/components/dataset/__tests__/SourcePanel.test.tsx` - passed (45 tests)

Left alone: `backend/app/modules/catalog/sources/origin_probe.py` has a comment above `AUTH_REQUIRED` referencing the old `unauthorized` copy ("whose shipped copy says the source 'now requires' access"). That comment is now slightly stale since the copy changed, but it's backend and out of this lane's scope.

### DIFF
--- a/frontend/src/components/admin/settings/SettingsAuthTab.tsx
+++ b/frontend/src/components/admin/settings/SettingsAuthTab.tsx
@@ -557,6 +557,12 @@ function OAuthProvidersSection({ envOnly }: { envOnly: boolean }) {
                 onChange={(e) => setForm((prev) => ({ ...prev, client_secret: e.target.value }))}
                 placeholder={editingProvider ? '********' : ''}
                 disabled={envOnly}
+                // fix(#1755): admin secret, not a login credential, so opt out
+                // of every password manager explicitly.
+                autoComplete="new-password"
+                data-1p-ignore
+                data-lpignore="true"
+                data-bwignore
               />
             </div>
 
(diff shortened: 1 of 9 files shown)


ishiland:
@codex review

ishiland:
@codex review

PR #1771, opened 2026-09-02
chore(search): sunset the two pre-#1666 search compatibility forms

### LINKED ISSUE #1671: Sunset the two search compatibility forms kept for pre-#1666 SDK clients
#1666 corrected three places where the published OpenAPI contract described FastAPI's defaults rather than what the server does. Two of those corrections would have silently broken clients generated from the previous contract, so #1670 kept the old forms working. Both are accepted but **not published**, and both should be removed.

This issue exists so that removal is tracked somewhere other than a CHANGELOG paragraph.

## What is being kept alive

**1. The `cql2_filter_lang` spelling of `filter-lang`**

`backend/app/modules/catalog/search/router.py` — `_resolve_filter_lang`, `_LEGACY_FILTER_LANG_PARAM`.

This was not merely the published name. Under the old `Depends()` binding it was the *only* spelling that bound — `filter-lang` worked solely because the handler read it off the raw query string:

```
/dep (old Depends)  ?cql2_filter_lang=cql2-json -> cql2-json
/dep                ?filter-lang=cql2-json      -> cql2-text
/qm  (new binding)  ?cql2_filter_lang=cql2-json -> cql2-text
/qm                 ?filter-lang=cql2-json      -> cql2-json
```

**2. `keywords` as a JSON request body on a GET**

`backend/app/modules/catalog/search/router.py` — `_legacy_keywords_body`.

The pre-fix contract declared it as an `application/json` body, so the generated Python client's parameter was named `body`. The exact reverse of the above:

```
/dep (old Depends)  JSON body ['osm','roads']    -> bound
/dep                ?keywords=osm&keywords=roads -> null
/qm  (new binding)  JSON body ['osm','roads']    -> null
/qm                 ?keywords=osm&keywords=roads -> bound
```

`collection_items` needs no shim for this — it keeps `Depends()`, so FastAPI still binds that body natively.

## Why they should go rather than stay

A request body on a GET is the defect #1666 fixed. It is not reliably forwarded by proxies and CDNs, so a client depending on it may already be broken in some deployments without anyone knowing. The `cql2_filter_lang` spelling is more benign but is still an undocumented second name for a published parameter, and every reader of the router has to be told why it is there.

## When

Both are safe to remove once clients have had a release cycle or two to regenerate. The published surface has been correct since the release containing #1666, so anyone who regenerates gets the right names automatically.

## What removal involves

- Delete `_legacy_keywords_body` and its call in `search_datasets_endpoint`.
- Drop the `_LEGACY_FILTER_LANG_PARAM` fallback from `_resolve_filter_lang`.
- Remove the compatibility tests in `backend/tests/test_ogc_cql2_filtering.py`: `test_legacy_keywords_json_body_still_filters`, `test_query_keywords_win_over_the_legacy_body`, `test_non_keyword_json_body_is_ignored_on_search`, and the `cql2_filter_lang` parametrize cases in `test_cql2_json_parses_under_either_filter_lang_spelling` / `test_bogus_filter_lang_rejected_under_either_spelling`.
- Lower the `_MODULE_LOC_CAPS` entry for `search/router.py` in the same commit — the ratchet fails on a file that shrinks below its cap, not only one that grows past it.
- Note the removal in `CHANGELOG.md`, where the deprecation was announced.

## Related

- #1666 — the contract defects.
- #1670 — the fix, and where both shims were added (codex rounds 2 and 3).

`test_malformed_legacy_body_still_validates_on_collection_items` should **stay**. It pins a separate runtime/contract asymmetry: `collection_items` still binds and validates the legacy body because that is inseparable from `Depends()`, so a malformed body answers a 400 the published contract does not describe. That is unrelated to these shims and is not resolved by removing them.

ishiland:
Closes #1671.

Removes the two search compatibility forms #1670 kept alive for one release cycle after #1666 corrected the published OpenAPI contract: the `cql2_filter_lang` spelling of `filter-lang`, and a JSON `keywords` request body on `GET /search/datasets/`. #1666 shipped in 1.16.0 (2026-08-27); 1.16.1 and 1.17.0 have shipped since, so the deprecation window has passed. Both forms were accepted but never published, so no client generated from the current contract is affected.

Deleted `_legacy_keywords_body` and its call in `search_datasets_endpoint`, and the `_LEGACY_FILTER_LANG_PARAM` fallback and constant from `_resolve_filter_lang`, along with the comments that only existed to explain them.

Fixing this exposed a second gap in `_resolve_filter_lang`: `collection_items` binds its query model with a bare `Depends()`, under which pydantic's synthesized `__init__` cannot bind the hyphenated `filter-lang` alias at all, so it falls back to binding the field's own Python name, `cql2_filter_lang`, from the raw query string. That meant `?cql2_filter_lang=...` kept working on that one route through the native binding alone, independent of the explicit shim just removed. `_resolve_filter_lang` now reads only the published wire name (`filter-lang`) and never falls back to `params.cql2_filter_lang`, so the sunset spelling has no effect on either route.

Removed the compatibility tests #1671 named: `test_legacy_keywords_json_body_still_filters`, `test_query_keywords_win_over_the_legacy_body`, `test_non_keyword_json_body_is_ignored_on_search`, and the `cql2_filter_lang` parametrize cases (renamed the two surviving tests to `test_cql2_json_parses_under_filter_lang` and `test_bogus_filter_lang_rejected` now that only one spelling exists). `test_malformed_legacy_body_still_validates_on_collection_items` stays, as it pins a separate runtime/contract asymmetry unrelated to these shims. Added two tests pinning the sunset behavior: `test_cql2_filter_lang_param_is_ignored` and `test_json_keywords_body_does_not_filter_on_search`.

Schema/API impact: none. Both forms were never published, and `make openapi-check` confirms the snapshot is unchanged.

Lowered the `_MODULE_LOC_CAPS` entry for `search/router.py` from 1536 to 1489, its exact new `wc -l`.

Verification run from the worktree:
- `uv run pytest tests/test_ogc_cql2_filtering.py tests/test_layering.py -q` -> 77 passed
- `uv run ruff check .` -> all checks passed
- `uv run ruff format --check .` -> all files formatted
- `PYTHONPATH=. uv run python scripts/dump_openapi.py --check` -> passed, no diff (backend/openapi.json untouched)

Nothing left out.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,17 @@ and releases use semantic versioning.
 
 ## [Unreleased]
 
+### Removed
+
+- **Two undocumented search compatibility forms are gone.** `?cql2_filter_lang=...`
+  (an alternate spelling of the published `filter-lang` parameter) and a JSON
+  `keywords` request body on `GET /search/datasets/` were kept working after
+  #1666 corrected the published OpenAPI contract, so a client generated from
+  the older contract would not silently break. Both were accepted but never
+  published, and the deprecation window has passed since #1666 shipped in
+  1.16.0. A client generated from a pre-1.16.0 contract must regenerate
+  against the current one. (#1671)
+
 ## [1.17.0] - 2026-08-30
 
 ### Added
--- a/backend/app/modules/catalog/search/router.py
+++ b/backend/app/modules/catalog/search/router.py
@@ -415,69 +415,29 @@ search_router = APIRouter(prefix="/search", tags=["Search"])
 _SUPPORTED_FILTER_LANGS = ("cql2-text", "cql2-json")
 
 
-_LEGACY_FILTER_LANG_PARAM = "cql2_filter_lang"
-
-
-async def _legacy_keywords_body(request: Request) -> list[str] | None:
-    """Read ``keywords`` from a JSON request body on a GET, for older clients.
-
-    compat(#1666 codex P2): the pre-fix contract declared ``keywords`` as an
-    ``application/json`` request BODY on a GET, so that is what SDKs generated
-    before this change send — the Python client's parameter was literally named
-    ``body``. Under the old ``Depends()`` binding FastAPI consumed it, and the
-    query-string form worked only because the handler read it separately. The
-    corrected query-parameter binding reads the opposite one, so an unchanged
-    client would silently receive UNFILTERED results, which is a worse failure
-    than an error.
-
-    Deliberately not republished: a request body on a GET is the defect being
-    fixed, is not reliably forwarded by proxies and CDNs, and should be sunset.
-    This honours what old clients send without offering it to new ones.
-
-    Anything that is not a JSON array of strings is ignored rather than
-    rejected. This runs on a normal search request, and the only job here is to
-    recognise the one shape the old generator emitted.
-    """
-    if not request.headers.get("content-type", "").startswith("application/json"):
-        return None
-    try:
-        body = await request.json()
-    except Exception:  # broad: any unreadable/!JSON body simply is not the legacy shape
-        return None
-    if isinstance(body, list) and body and all(isinstance(v, str) for v in body):
-        return body
-    return None
-
-
 def _resolve_filter_lang(
     params: SearchQueryParams, request: Request
 ) -> SearchQueryParams:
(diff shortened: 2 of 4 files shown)


ishiland:
@codex review

Issue #1459, opened 2026-08-13
auth: a login beginning inside the revoking transaction still outlives its logout


ishiland:
Found by codex review on PR #1458 (P1, `backend/app/modules/auth/service.py`) and confirmed by reproduction. Filing rather than folding into #1458 because closing it needs a design decision, not a patch: it is the timestamp horizon's premise that fails, not its implementation.

## The interleaving

`revoke_all_tokens` stamps the horizon with `func.now()`, which is the *transaction* timestamp, and `refresh_tokens.created_at` defaults to `func.now()` in the inserting transaction. Both are transaction-start times, so they order transactions by when they BEGAN, not by when they committed.

A login that begins after the revoking transaction begins, but before it commits, therefore gets a `created_at` LATER than the horizon while still reading the PRE-bump `token_version`:

1. Logout takes the owner lock, revokes the visible rows, stamps `sessions_revoked_at = R`, bumps `token_version`, and has not committed.
2. Login begins at `L > R`. It reads `token_version` and sees the old value, because the logout is uncommitted. It mints an access JWT carrying that stale value.
3. Login inserts its refresh row. The insert blocks on the FK `KEY SHARE` lock against the logout's `FOR UPDATE`.
4. Logout commits. The login's insert proceeds and commits.

The row is unrevoked, because the logout's UPDATE ran before the row existed, and it is stamped `created_at = L > R`, so it passes the `created_at > sessions_revoked_at` predicate #1458 added. Its access JWT is dead on the `token_version` check, but rotating the refresh row mints a new one carrying the CURRENT version.

Both password and OAuth login issue refresh tokens without taking the owner-row lock, so nothing serializes this.

## Reproduction

Confirmed on `fix/1455-revocation-horizon` by stalling the logout handler inside `audit_emit`, which runs after `revoke_all_tokens(commit=False)` and before `db.commit()`, so the logout holds its lock and its uncommitted revocation while a login is fired 0.5s later:

```
PROBE logout=204 login=200
PROBE horizon=2026-08-13 11:41:03.975377+00:00 token_version=2
PROBE row created_at=2026-08-13 11:41:03.790822+00:00 revoked=True  after_horizon=False
PROBE row created_at=2026-08-13 11:41:04.538396+00:00 revoked=False after_horizon=True
PROBE rotate_status=200
PROBE me_after_rotate=200
```

The second row is the racing login's. It rotates, and the rotated access token authenticates against `/auth/me/`. The session outlived its logout.

[...]

Issue #1455, opened 2026-08-13
auth: time-scope logout revocation so cross-tab login/logout races cannot span it


ishiland:
Follow-up to GH-1302 (PR #1446), which made logout a real server-side security event: `revoke_all_tokens` bumps `token_version` and revokes every refresh row for the user, "as of now".

## The class

A login (or refresh) that is in flight while a logout's server round-trip commits is inside that revocation set by definition. The client serializes every path one tab can observe: `pendingLogout` gates same-tab login and the OAuth redirect, and the cross-tab storage listener bumps the session epoch and aborts an in-flight refresh. None of that can span tabs reliably or outlive the client-side timeout.

Two concrete members were flagged by review on PR #1446 and deferred with reasoning there:

- A login in tab B racing a logout in tab A. `pendingLogout` is module-local to tab A, so tab B cannot wait on it (review round at commit 3b1ab32d5).
- A suspended tab that misses the intermediate cleared-storage state. Both queued storage events observe the latest non-empty state, so the abort branch never runs, and a late refresh response from the ended session can overwrite the successor login's cookie with an already-revoked token (review round at commit 45e20ccfd). Inspecting the storage event's `newValue` would narrow this member, but browsers coalesce or drop events for suspended tabs, so client-side detection stays best-effort whatever the listener reads.

## Residual today

Liveness only, in both directions: a session that ends early re-prompts for login, and an overwritten cookie is already revoked server-side, so it can never authenticate. No session outlives its logout.

## The durable fix

Scope revocation server-side: revoke refresh rows issued at or before the logout request's own timestamp, with matching treatment for the `token_version` bump, so a successor login's credentials sit outside the revocation set by construction rather than by client-side choreography.

Deliberately NOT narrowing logout's revoke-everything semantics: revoking every session is what makes logout a security event, and SEC-S15 builds access-JWT invalidation on top of it.

ishiland:
Third member of the class, from PR #1446 review round at 6744a48eb: zustand persist omits `refreshToken` from the persisted blob rather than writing an explicit null, so a sibling tab's epoch-update write can re-persist its in-memory body token after another tab's logout cleared storage. Material only when the logout's server-side revocation also failed; otherwise the re-persisted token is already revoked. When revocation failed, the sibling tab continues from its in-memory copy regardless of storage, so the persist shape is not the load-bearing part — time-scoped server-side revocation is.

PR #1772, opened 2026-09-02
feat(admin): let an admin reset another user's password

### LINKED ISSUE #1715: No admin path to reset a user's password, though the login page promises one

ishiland:
Closes #1715.

## What changed

`POST /admin/users/{user_id}/reset-password/` sets another account's password. Admin-only (`manage_users`, the same dependency the other user lifecycle routes use), body is `{"password": str}` validated by the same policy `POST /admin/users/` applies at creation, response is the updated `UserResponse` like `deactivate/` and `approve/`. 404 for an unknown user through the shared `_raise_on_error` path; 422 when the target signs in through an identity provider, because there is no local password to replace and minting one silently would be worse.

The reset revokes the account's credentials: `revoke_all_tokens(commit=False, bump_key_epoch=True)`, folded into the same transaction as the new hash and the audit row. That is copied deliberately from `POST /auth/change-password/`, which has done exactly this since SEC-S15 and #821. So anyone holding the old password does not keep a live access JWT, refresh row or API key.

`user.password_reset` joins `AUDIT_ACTIONS` (and `CURRENT_AUDIT_ACTIONS` on the frontend, which the registry test pins), emitted with the acting admin as `user_id` and the target as `resource_id`, `details` limited to `{"username": ...}`. The submitted value reaches the bcrypt hash column and nowhere else.

Frontend: a "Reset password" item on the user row's action menu opens a one-field dialog with a confirm button and a success toast. No generated password display and no clipboard copy, so the value only ever exists in the field the admin typed it into. The field carries `autoComplete="new-password"` plus the three password-manager opt-outs the service-token inputs from #1750 use, since the credential belongs to someone else's account and should not be filed under this admin.

Ten new keys in all four locales (en/es/fr/de).

## Two decisions worth flagging

**The action is offered on the admin's own row.** Every other destructive item in that menu is hidden for self, but this one is the only in-app way back for an admin who is still signed in and has forgotten their password: `change-password` requires the old value. The dialog says a self-reset signs them out here too, and it does, because the revocation does not exempt the caller.

**No confirm-password field.** The dialog has a reveal toggle (the shared `PasswordInput`), which is how the admin checks what they typed before handing it over. A second field would be the alternative; it does not fit "one password field" and the reveal is the cheaper check.

## Session invalidation in change-password

The brief asked whether `POST /auth/change-password/` invalidates existing JWTs today, and whether a revocation mechanism needed building. It does invalidate them, so nothing new was built. `auth/router.py` calls `AuthService.revoke_all_tokens(current_user.id, commit=False, bump_key_epoch=True)`, which takes the owner row lock, marks every unrevoked refresh row revoked, and in one UPDATE bumps `token_version` (rejecting prior access JWTs), bumps `key_epoch` (#821, so API keys minted before the change stop resolving), and stamps the revocation horizon (#1455, a use-time predicate that also covers a refresh row this transaction cannot see yet). `commit=False` keeps it in the caller's transaction. The reset endpoint calls the same method with the same arguments; `convert_saml_user_to_local` was already the third caller. No follow-up issue filed.

## Schema/API/config impact

- New route, so `backend/openapi.json`, both SDKs and `frontend/src/types/api.generated.ts` are regenerated and committed.
- No schema change, no migration, no new config.
- One refactor inside `admin/schemas.py`: the four-line password-policy `field_validator` was pasted into `AdminUserCreate` and `SamlToLocalConversion`, and the new schema would have made three. It is now one `_enforce_password_policy` helper the three classes share. The rules themselves are untouched in `auth/password_policy.py`, and the OpenAPI diff is additions only.

## Verification

Backend, from this worktree against Postgres on 5434:

- `uv run pytest tests/test_admin_password_reset.py -q` — 8 passed (new file)
- `uv run pytest tests/test_layering.py -q` — 47 passed
- `uv run pytest tests/test_audit_action_registry.py tests/test_password_policy.py tests/test_api_contract_openapi.py -q` — 42 passed
- `uv run pytest tests/test_admin_user_operations.py tests/test_admin_rate_limit.py tests/test_admin_create_user_failfast.py tests/test_phase_279_user_lifecycle.py -q` — 25 passed
- `uv run pytest tests/test_rule1_structural.py tests/test_rule2_structural.py tests/test_redirect_slashes.py tests/test_admin_invariants_regression.py tests/test_lifecycle.py -q` — 121 passed, 4 skipped
- `uv run pytest tests/test_auth.py tests/test_password_policy.py tests/test_phase_279_admin_polish.py tests/test_admin_identity_migration.py -q` — 61 passed
- `uv run pytest tests/test_sdks_round_trip.py -q` — 15 passed, 1 skipped
- `uv run ruff check .` and `uv run ruff format --check .` — clean
- `make openapi-check` and `make sdks-check` — both pass on the committed tree

Frontend:

- `npx vitest run src/components/admin/__tests__/UserResetPasswordDialog.test.tsx` — 4 passed
- `npx vitest run src/api/__tests__/admin.contract.test.ts` — 17 passed
- `npx vitest run` over the four adjacent admin suites (UserList sorting, CardHeaderPatterns, AuditLogViewer sorting, admin contract) — 47 passed
- `npm run typecheck`, `npm run lint`, `npm run test:i18n`, `npm run check:i18n:changed`, `npm run check:i18n:toast-strings`, `npm run check:i18n:api-errors` — all clean

No Playwright spec was added or run. Exercising this end to end needs a stack built from this worktree's backend, and the shared stack serves main.

## Commits two and three: the password floor in the copy

Three strings had said 8 characters since before SEC-S16 raised the policy floor to 12, so each promised a length the API refuses. Eight is only the schema's fast-fail floor.

- `userCreate.placeholders.password` (admin create dialog)
- `users.resetPasswordDialog.passwordPlaceholder` (the new reset dialog, added to match rather than stay silent)
- `auth.passwordRequirement` (the register form, where a self-signing-up user meets it first)

All three say 12 now, in all four locales. No length claim of 8 remains under `frontend/src/i18n/locales/`.

The client `minLength` stays at 8 on all three fields on purpose. `PASSWORD_MIN_LENGTH` is configurable (`core/config.py`, `Field(default=12, ge=8)`) and is not exposed to the frontend, so a hardcoded 12 in the HTML gate would block valid values on an instance that lowered it. The copy states the shipped default, which is what the OpenAPI field descriptions already say; the backend refusal names the actual configured rule. There is a comment to that effect in the reset dialog.

No test pinned any of the old strings.

## Review round 1

Four findings, all fixed in `b605bab7e`. Each one was measured before and after rather than reasoned about.

**P1, the write race between reset and change-password.** `change_password` verified the pre-reset hash and assigned its replacement, then blocked inside `revoke_all_tokens`' own `FOR UPDATE`. When the reset released the row, SQLAlchemy autoflush wrote the self-service password on top of the admin's, so whoever held the old password silently won the recovery and the admin saw a success. `change_password` now calls `db.refresh(current_user, with_for_update=True)` before it reads anything off the row, so it waits on the same lock and verifies against the committed hash, refusing the stale `current_password` with the usual 400.

Lock ordering is safe: the reset takes the admin-lifecycle advisory lock and then the row lock, while change-password only ever takes the row lock, so there is no cycle. `convert_saml_to_local` cannot interleave with a reset at all, because it requires `auth_provider == 'oauth'` and the reset requires `'local'`, and only the conversion flips it.

`test_reset_is_not_overwritten_by_a_racing_change_password` holds the row the way the reset does, starts a change-password underneath it, then commits the admin's hash before releasing. With the lock removed it fails with `assert [204] == [400]`, and the old-password holder's value is the one that works afterwards. That is the exploit, reproduced.

**Bcrypt's 72-byte input limit.** Every password entry point accepts up to 256 characters, but bcrypt hashes at most 72 bytes and pwdlib raises rather than truncating. The bound now lives in `validate_password_complexity`, not in the admin-only helper, because that is the function all five entry points already funnel through. It is checked before class diversity so a long multibyte value is told the real reason instead of a class message it may also trip, and it is deliberately not configurable: it is a property of the hash function, not a policy knob.

Two shipped endpoints had the same gap and are fixed by the same change, both confirmed by removing the check and re-running:

- `POST /admin/users/` returned **409 Conflict** with the bcrypt library's own text as the detail (`"password cannot be longer than 72 bytes, truncate manually if necessary..."`).
- `POST /auth/change-password/` let the `ValueError` escape the handler entirely. Nothing registers a `ValueError` handler, so that is a 500.

Both now return a 422 naming the real constraint. The five field descriptions state the byte limit, so the contract no longer advertises a length it refuses; that changed the schema text, so the four generated artifacts are regenerated and committed.

**Dismissing the dialog mid-request.** A close while the reset was in flight unmounted the dialog without stopping anything, so the backend went on to change the password and revoke the credentials with the UI showing no trace. Cancel, the X, Escape and the overlay are all blocked while pending, at both the `onOpenChange` guard and the primitive's own props. The request is deliberately not aborted: aborting the fetch would not undo a commit the backend may already have made, so it would swap a confusing outcome for an unknowable one.

**A self-reset left the dead session in place.** The backend revokes the acting admin's own token, but the store kept it until some later request happened to 401, contradicting the warning the dialog had just shown. The success path calls `logout()` when the target is the signed-in admin; `ProtectedRoute` redirects and `wireAuthCacheReset` clears the query cache off the identity change.

Both dialog tests are real positive controls: strip the guards and the close test fails, remove the `logout()` and the self-reset test fails.

## Verification for this round

Backend:

- `uv run pytest tests/test_admin_password_reset.py tests/test_password_policy.py tests/test_auth.py -q` — 70 passed
- `uv run pytest tests/test_layering.py tests/test_audit_action_registry.py tests/test_api_contract_openapi.py tests/test_sdks_round_trip.py -q` — 89 passed
- `uv run pytest tests/test_admin_user_operations.py tests/test_admin_rate_limit.py tests/test_lifecycle.py tests/test_phase_279_user_lifecycle.py tests/test_admin_identity_migration.py tests/test_admin_create_user_failfast.py -q` — 26 passed, 4 skipped
- `uv run ruff check .` / `uv run ruff format --check .` — clean
- `make openapi-check`, `make sdks-check` — pass on the committed tree

Frontend:

- `npx vitest run src/components/admin src/components/auth src/api/__tests__/admin.contract.test.ts` — 25 files, 246 passed
- `npm run typecheck`, `npm run lint`, `npm run test:i18n` — clean

No cap was raised. `auth/router.py` is 1294 lines against the 1500 router ceiling.

## The published example password

`UserCreate.password` in `auth/schemas.py` carried `json_schema_extra={"example": "securePass123!"}`, which Rule 3 forbids: it shipped into `backend/openapi.json`, the Python SDK's docstring and `api.generated.ts`, where it reads as a value to copy. Removed, with no replacement, since for a password no example beats one that looks usable. The sibling username and email examples stay; neither is a credential.

One hand-written line, three generated artifacts. A repo-wide grep for `securePass` (excluding `node_modules`, `.venv`, `.git`) now returns nothing.

## Review round 2

**P2, the self-reset bypassed the refresh abort.** The dialog called the store's `logout()` directly. That bumps `sessionEpoch`, which stops a late refresh writing rotated tokens back, but it does not stop the browser processing that response's `Set-Cookie`. A refresh still in flight during the reset could land after the admin signs in again and overwrite the new refresh cookie with the one the reset had just revoked, breaking the next refresh. Every other deliberate sign-out pairs the epoch bump with `abortInflightRefresh()`: `useAuth`'s logout, `notifySessionExpired`, and the cross-tab storage listener. The dialog was the one path that did not.

It now calls `useAuth().logout`, which aborts first and also handles the server-side revoke, the query-cache teardown and the redirect. No extraction was needed: `useAuth` works inside the dialog, and the test harness already wraps renders in a router.

`UserResetPasswordDialog.session.test.tsx` drives the real `useAuth` rather than a mock and records `sessionEpoch` at the moment the abort fires, pinning it to the pre-teardown value. Against the bare store logout the abort never fires at all, so the test fails.

## The red CI step

`Frontend Lint & Typecheck` failed on `964b11439` at the hardcoded-toast scanner, which reads test files too. The close-guard test called `toast.success` with a literal to stand in for the mutation's `onSuccess`; it goes through `i18n.t` now.

The cause was process, not the flag: `--strict` is already inside `npm run check:i18n:toast-strings`, and that gate simply was not re-run on the commit that introduced the line. Every step of that job ran locally before this push: `types:check`, `test:i18n`, `check:i18n:toast-strings`, `check:i18n:changed`, `lint`, the four ESLint security-rule pins, and `typecheck`.

Verification for this round: 27 files / 261 passed across `src/components/admin`, `src/components/auth`, `use-auth` and the admin contract test.

## Review round 4

**P1, the login path had the same race as change-password.** `LocalAuthProvider.authenticate()` looked the account up with an unlocked `SELECT`, so a login that read the row before the reset committed verified the stale hash and then minted from the post-reset row: an access JWT stamped with the new `token_version`, and a refresh row created after `sessions_revoked_at`. Both outlive the revocation, so the old-password holder came out of the recovery holding a live session.

The lookup takes `FOR SHARE` now. It conflicts only with the reset's `FOR UPDATE`, so concurrent logins for one account still run in parallel, and the lock is held to the end of the request transaction, which is what makes it cover the mint as well as the verify. Nothing between the lookup and the commit leaves the database (`enforce_email_domain_gate`, `PASSWORD_LOGIN_ENABLED.get_uncached`, `user_has_capability` are all DB reads). Ordering is unchanged and acyclic: the reset takes the advisory lock then the row; login and change-password take only the row.

Measured both ways. `test_reset_is_not_outrun_by_a_racing_password_login` holds the row the way the reset does, starts a login underneath it, and commits the admin's hash before releasing. With the lock the login is refused 401 and no refresh row appears; with the lock removed it returns 200 and leaves a refresh row behind. That is the exploit, reproduced.

**The SAML-to-local conversion gets the same lock.** A lost update was already unreachable there through `auth_provider` (that path demands `oauth`, the reset and change-password demand `local`, and only that path flips it), but it rotates credentials and the invariant is an argument a future edit can break silently, unlike a lock.

**Checked and deliberately left alone.** The refresh rotation already takes the owner-row `FOR UPDATE` and re-checks under it (#1446), and carries the use-time revocation horizon (#1455). A repo-wide grep for password verification finds exactly two consumers, login and change-password, so there is no API-key confirm or re-auth prompt to cover. The OAuth login path mints without this lock by design: it proves a current IdP assertion rather than the password the reset replaced, so its caller is not the old-password holder.

Verification: reset + auth + policy suites 71 passed; refresh/logout/horizon/api-key 93 passed; lifecycle, admin-lifecycle, seed-admin-concurrency, layering, audit-registry 59 passed + 4 skipped; rule1/rule2 structural, api contract, sdks round-trip 143 passed; ruff clean; `make openapi-check` clean (no contract change, so no regeneration). Also 106 passed under `-n 4` across the login-heavy suites, since a new row lock on the path every test authenticates through is exactly where contention would show.

## Review round 5

**P2, the login lock deadlocked concurrent logins.** The lock added in round 4 took `FOR SHARE`, and the login handler goes on to assign `user.last_login_at`. That UPDATE needs `FOR NO KEY UPDATE`, so two transactions holding a shared lock on one account both tried to upgrade, and Postgres broke the cycle by aborting one. Measured: an otherwise-valid sign-in comes back **503**, the `DBAPIError` handler's rendering of the deadlock.

The lookup takes `FOR NO KEY UPDATE` now (`with_for_update(key_share=True)`, verified to render exactly that). Two logins to one account serialize for the few milliseconds of verify-plus-mint rather than deadlocking, and the mode is still weaker than `FOR UPDATE` where it matters: it does not conflict with `FOR KEY SHARE`, so foreign-key checks from other tables, including this request's own `refresh_tokens` insert, are not blocked. It still conflicts with the reset's `FOR UPDATE`, so round 4's race stays closed and that test still yields 401.

**Why per-account serialization of logins is acceptable.** The window is one bcrypt verify plus the token mint, and it is scoped to a single account: logins to different accounts never touch the same row and are unaffected. Concurrent logins to one account are the rare case, not the hot path, since a person signs in once and the login endpoint is rate limited per IP. It is also the mode the transaction was always going to need, because the handler writes `last_login_at` before it commits; taking it at the SELECT only moves the acquisition earlier, and a transaction that acquires the strongest lock it will need up front is the standard remedy for a lock-upgrade deadlock. What it buys is that a simultaneous sign-in waits a few milliseconds instead of one of them failing outright.

**On the test.** Two real logins in one event loop cannot reproduce this: `verify_password` is synchronous, so it serializes them before their lock windows overlap, and the naive concurrent test passes on both lock modes. `test_login_does_not_deadlock_against_a_shared_lock_then_write` drives the other side from a second session instead, which makes the interleaving deterministic and fails with 503 on the old mode. The plain two-concurrent-logins test is kept as a cheap guard on the happy path, labelled as such rather than as a reproduction.

Verification: reset + auth + policy 73 passed; refresh cookies, refresh logout, revocation horizon, api-key, lifecycle, seed-admin-concurrency 95 passed + 4 skipped; layering, rule1/rule2 structural, api contract, sdks round-trip, audit registry 193 passed; 108 passed under `-n 4` across the login-heavy suites; ruff clean; `make openapi-check` clean, no contract change.

## CI failure on 09c41ac79

`tests/test_login_revocation_serialization.py` is #1460's test, not one of mine, and the failure was real, mine, and deterministic. Not a load flake: it fails on an idle host, in isolation, without `-n 4`, in 35 seconds every time.

| provider | test | result |
| --- | --- | --- |
| main, no lock | main | passes in 4.3s |
| mine, locked | main | fails in 34s, the CI failure |
| main, no lock | my first fix attempt | fails, so that attempt was wrong |
| mine, locked | fixed test | passes in 6s |

That test stalls a logout mid-transaction and releases a login underneath it, holding the revocation open until it has seen the login waiting on the owner-row lock. It signalled from `create_access_token`, because that call's `token_version` re-SELECT was the query that autoflushed the pending `last_login_at` write into the lock queue.

The login takes `SELECT ... FOR NO KEY UPDATE` in `authenticate()` now, so it blocks one step earlier and never reaches `create_access_token` while the stall holds the lock. The signal could only fire once the lock was released and the stall was the thing holding it, so the two waited on each other until the 30s deadline killed both. Confirmed by sampling `pg_stat_activity` during the hang: the only blocked statement is the login's locking SELECT of the account row.

The fix moves the signal to `authenticate`, the call that now blocks, still firing before it delegates so it cannot depend on the lock it is about to wait for. The `pg_stat_activity` observation is unchanged and still matches, because that locking SELECT lists every user column including `last_login_at`.

Test-only diff. The product behaviour is correct and is what the test exists to protect: the racing login queues behind the revocation and comes out a clean successor.

Two checks that the edit did not hollow the test out. It still has teeth: with the serialization removed outright it fails on the assertion that the racing login's own access token must authenticate, which is the #1459 zombie-session shape. And it is not coupled to this implementation: it passes unchanged against a provider with no lock at all, which is what the OAuth path still looks like, and the structural guards that protect that path are untouched.

Stability: 5 consecutive runs green, and 108 passed under `-n 4` across the login-heavy suites.


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1773, opened 2026-09-02
fix(manifest): add checksum field for change detection under stable URIs

ishiland:
## What changed and why

`manifest_dataset_fingerprint` hashes the manifest entry, not the bytes a source URI points at. A source whose file changed under a stable URI (`latest.gpkg`, or a path an ETL job overwrites in place) classified as `skip_complete` forever, with no way to force a re-import short of renaming the file every run (gh#1736, options 1 and 3).

- Added `ManifestSource.checksum: str | None`, validated as `sha256:<64 lowercase hex>`. It is declared, not verified: apply folds it into the existing entry fingerprint like any other field, and never fetches the source to check it against the declared digest. An absent checksum is `exclude_none`-dumped, so every existing fingerprint is unaffected; a changed checksum reclassifies the entry from `skip_complete` to `update`.
- Reworded the `skip_complete` message. It previously read "Manifest dataset is already up to date," which implies the source was checked. It now says the entry is unchanged and the source was not inspected, and names `checksum` as the escape hatch. Other skip reasons (in-flight, etc.) are unchanged.
- Documented `checksum` in the CLI README next to the existing entry-fingerprint paragraph.
- Mirrored the field in the CLI's packaged JSON Schema (`cli/geolens_cli/manifest/schemas/geolens-manifest-v1.schema.json`), which is hand-kept in step with the backend Pydantic model (same convention already used for `ManifestMetadata.attribution`). Without this, `geolens validate` / `geolens apply --dry-run` would reject `checksum` locally with an `additionalProperties` error even though the backend accepts it.

### Byte verification: not added

I grepped the ingest code for `sha256`/`md5` before deciding whether to verify staged bytes against the declared checksum (issue option 2, deliberately out of scope for this PR, but the "verify only if the staging path already computes a digest" condition needed checking). The result is mixed:

- The reupload/update path (`tasks_reupload.py`) already computes `file_hash = sha256_file(file_path)` on the staged file before ingest.
- The initial create path (`tasks_vector.py`) does not; a comment there says so explicitly ("no file_hash — it is computed on the re-upload path only").

Since the digest is not computed uniformly across both paths apply can take, and both task files are outside this lane's scope, I did not add verification. Follow-up: plumb `file_hash` into the create path too, then compare it to a declared `checksum` on both paths.

## Schema/API impact

Additive, optional field on `ManifestSource`. Regenerated `backend/openapi.json`, both SDKs (`sdks/python`, `sdks/typescript`), and `frontend/src/types/api.generated.ts`. No migration.

## Verification

- `run_pytest.sh tests/test_manifest_apply_service.py tests/test_manifest_apply_api.py tests/test_manifest_apply_roundtrip.py tests/test_manifest_apply_vrt.py -q` — 96 passed
- `run_pytest.sh tests/test_layering.py -q` — 47 passed (no LOC cap needed raising)
- `run_pytest.sh tests/test_sdks_round_trip.py -q` — 15 passed, 1 skipped (pre-existing skip)
- `make openapi-check` — clean after regen
- `cd cli && uv run --extra dev python -m pytest -q` — 337 passed
- `cd backend && uv run pytest tests/test_cli_round_trip.py -v` (with SDK/CLI editable installs, per `make cli-test`) — 8 passed, 2 skipped (pre-existing skips)
- `cd frontend && npm run typecheck && npm run lint && npm run test:i18n` — clean
- `cd frontend && npm run types:check` (regenerated `api.generated.ts`) — matches committed output

No LOC caps needed raising; no new `t()` keys (backend/CLI-only surface).


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

Issue #1556, opened 2026-08-17
Further job-lifecycle and audit-trail edges after the backfill queue move


ishiland:
Follow-up home for further job-lifecycle and audit-trail edges after #1550, so that PR can land rather than absorb an open-ended series.

## Where #1550 got to

Eight review rounds, every finding real. A full embedding regenerate used to run inline in the HTTP request and outgrow the 600 s edge timeout at roughly 59,000 records, after which the operator saw a 504 while the server kept working — and a retry started a second concurrent regenerate, which on the force path meant a second DELETE of every embedding.

The endpoint now returns a job id immediately and the run happens on the Procrastinate queue that already lives inside PostgreSQL. Along the way the review turned up nine distinct ways the job row and its durable audit trail could disagree:

- queue dispatch failing left the trail permanently at `requested`;
- a lost fence recorded a completion that never landed;
- `asyncio.CancelledError` inherits from `BaseException`, so cancellation bypassed every `except Exception` handler and held the unique backfill slot until the 60-minute stale sweep;
- a failure of the terminal write itself escaped the same way;
- a commit whose acknowledgement was lost was read as failure, overwriting a durable success;
- a hard kill left no in-process actor to close the trail at all;
- the sweeper added to close it audited unconditionally, so a job that was actually running or complete could be recorded as failed;
- a mixed-success run finalized as `complete` with no way for the operator to see the errors, losing a warning the synchronous UI used to show;
- and the concurrency guard itself was a check-then-insert race, replaced by a unique partial index (`uq_ingest_jobs_active_embedding_backfill`, migration `0050`) so PostgreSQL enforces one active run rather than the application.

## The principle the remaining work belongs to

**In-process state cannot describe what the database did.** Recovery decides outcomes by observing the row, not by reading a fence result or an in-process flag, because "my fenced update matched nothing" has two causes — someone else settled it, or I settled it and never heard back — that demand opposite outcomes.

**And terminal audit emission should be idempotent.** One operation gets one terminal entry, whoever writes it first. The worker, the status poll and the stale sweeper can all reach the closing path, and idempotency needs no coordination between them, whereas teaching each actor to check the others is a rule that has to hold in three places forever.

## Candidates

- Any remaining path where an awaited commit sits outside the recovery boundary at either end of the run.
- Whether other job types sharing `platform/jobs/router.py` inherit the same audit behaviour, since the sweeper is shared and only the embedding-backfill semantics were taught to it.
[...]

PR #1774, opened 2026-09-02
fix(jobs): settle abandoned uploads as cancelled, not failed

### LINKED ISSUE #1744: Abandoned direct uploads report as failed, so most of the admin Jobs failure list is noise
Found auditing the live demo's Jobs list: **4 of its 6 failed jobs are not failures.** They are direct uploads whose owner walked away at the preview step, and the sweeper reports each one as `failed` with `Stale: pending too long (never queued)`.

## Evidence from the demo (1.17.0, build 8dc529f17)

```
source_filename        | file_path                              | presigned | created_at           | completed_at
t1.geolibre.json       | /app/staging/e9bdac4b-..._t1.geolibre.json | (none) | 2026-08-31 16:59:48 | 2026-08-31 18:01:21
宾馆酒店_point.zip      | /app/staging/2f8968c0-...zip           | (none)    | 2026-08-28 13:13:56 | 2026-08-28 14:14:22
uls.csv                | /app/staging/0fbea6c5-..._uls.csv      | (none)    | 2026-08-18 20:21:29 | 2026-08-18 21:21:44
uls.csv                | /app/staging/6a2e5860-..._uls.csv      | (none)    | 2026-08-18 20:20:17 | 2026-08-18 21:21:44
```

Three distinct users, spanning two weeks, the most recent one today on 1.17.0. Every row has an absolute `/app/staging/...` path, no `presigned` marker, no `source_url`, and **no `catalog.procrastinate_jobs` row at all** — nothing was ever dispatched. Each settles at almost exactly `pending_job_timeout_seconds` after creation (the two `uls.csv` rows share a `completed_at` to the microsecond: one sweep transaction reaped both).

## Why they take the failure path

`POST /ingest/upload` says it in its own docstring:

> Validates the file extension, creates an ingest job, and saves the file to staging. **Does NOT auto-queue ingestion -- use preview then commit.**

So on the default upload path a `pending` row with no Procrastinate job is the *normal, expected* state between upload and commit. When the user never commits, the unbound half of the pending sweep reaps it.

#1556 already established the right treatment for exactly this behaviour — `cancelled` plus `Abandoned: upload was never completed`, on the grounds that "for a visitor who presigned an upload and walked away, nothing was ever attempted, and on the public demo those rows are indistinguishable from real ingest failures in the admin jobs list and the failed-jobs badge that counts it."

That reasoning applies verbatim here. But the carve-out is keyed to the presigned door:

```python
# backend/app/platform/jobs/sweep.py:233
return and_(
    func.coalesce(IngestJob.file_path, "") == "",
    func.coalesce(IngestJob.user_metadata["presigned"].astext, "false") == "true",
)
```

A direct upload binds an absolute staging path and stamps no `presigned` marker, so it matches neither predicate. The fix landed on the less-travelled door: the demo has zero presigned uploads and four abandoned direct ones.

## Why this isn't a one-line widening

`abandoned_presigned_upload`'s docstring is explicit that it is "deliberately narrower than 'falsy `file_path`'", and the absolute-path class it excludes is a real one: a commit whose `defer_with_orphan_guard` rollback did not land leaves a genuinely broken job `pending` with an absolute path, and that must keep reporting `failed` so `/jobs/{id}/retry` stays available. Reclassifying the whole class as `cancelled` would take a recoverable job's only recovery path away — the same trap #1235 avoided.

The two states are only indistinguishable because the row carries no record of whether a commit was ever attempted. A marker stamped at commit-dispatch time, in the shape `presigned` and `staged_at` already use, separates them:

- pending, unbound, **never reached commit** → `cancelled` / abandoned
- pending, unbound, **commit attempted** → `failed` / retryable, unchanged

Hence `needs-design` rather than a patch: the marker has to be written at every door that dispatches a staged job, and the poll path, the background sweep and the worker's startup recovery all have to read it, which is precisely the three-site consistency problem `stale_pending_unbound_values` exists to prevent anyone from solving one site at a time.

## Impact

The admin Jobs list and its failed-jobs badge are the operator's ingest health signal. On the demo that signal currently reads 6 failures where 2 are real, and both real ones are already-fixed GDAL cases (#1636, #1640) that no longer reproduce. A self-hosted operator with visitors or a team gets the same inflation, and the message they see — "never queued" — points at the queue, which is working correctly.

## Not in scope

The two genuine failures in the same list are already remediated and are recorded here only so the audit is complete:

- `Peshawar_City_LULC_2050.tif` (08-21) — `PREDICTOR=2 is only supported with 8/16/32/64 bit samples`, fixed by #1636 on 08-24. The `_predictor_supported` docstring cites this exact NBITS=4 LULC raster.
- `march.gpkg` (08-18) — a 1,305-character dump of GDAL's entire driver list as the user-facing error, fixed by #1640 on 08-24.

Both fixes are live on the demo. The rest of the queue is healthy: 149/157 jobs complete, no stuck `pending`/`running` rows, the single `fanned_out` row is a terminal parent with `completed_at` set, and the four `failed` Procrastinate rows map exactly to those two GDAL failures at two attempts each — the retry path worked.

ishiland:
Closes #1744.

## What changed

`POST /ingest/upload` stages a file and does not queue anything, so a `pending` row with no Procrastinate job is the normal state between upload and commit. When the owner walks away at the preview step, the pending sweep reaps that row as `failed` with "Stale: pending too long (never queued)". On the demo that put four abandoned direct uploads, from three users over two weeks, into the admin failed-jobs badge, with a message pointing at a queue that was working correctly.

#1556 fixed the same reasoning for the presign doors, but keyed the carve-out to an empty `file_path` plus a `presigned` marker. A direct upload binds an absolute staging path and stamps no marker, so it matched neither half.

The two states were only indistinguishable because the row carried no record of whether a dispatch had ever been attempted. Now it does:

- `defer_with_orphan_guard` stamps `user_metadata["commit_attempted_at"]` (ISO-8601 UTC) and commits it immediately before the defer.
- `abandoned_upload` in `jobs/sweep.py` reads its absence. Pending, unbound, aged, no live queue row, no stamp becomes `cancelled` with "Abandoned: upload was never completed". With the stamp it stays `failed` with the existing "never queued", so `/jobs/{id}/retry` remains reachable.
- `abandoned_presigned_upload` and its Python twin are gone. The new rule subsumes them: a presigned upload nobody bound bytes to was never dispatched either. The absolute-path carve-out the old docstring defended no longer needs a branch, because the shape of `file_path` was never what the two classes differed by.

All three settling sites still read the one predicate through `stale_pending_unbound_values`, untouched: the background sweep, the status poll and the worker's startup recovery.

## One stamping site, not one per door

`defer_with_orphan_guard` is the only route to the queue for an `IngestJob`: nothing calls `defer_async` with a `job_id` outside a closure the guard runs. The `job` argument is now required there, so a new dispatch site cannot be written past the stamp. The 17 call sites, one per dispatch branch:

| Module | Site |
| --- | --- |
| `processing/ingest/service.py` | `create_vrt_job` |
| `processing/ingest/service.py` | `create_fan_out_jobs` |
| `processing/ingest/service.py` | `queue_ingest_job`, service branch |
| `processing/ingest/service.py` | `queue_ingest_job`, raster branch |
| `processing/ingest/service.py` | `queue_ingest_job`, vector branch |
| `processing/ingest/router.py` | `add_vrt_source` |
| `processing/ingest/router.py` | `remove_vrt_source` |
| `processing/ingest/manifest_service.py` | `_queue_reupload_job` |
| `catalog/datasets/api/router_vrt.py` | `regenerate_vrt_endpoint` |
| `catalog/datasets/api/router_refresh.py` | `_dispatch_postgis_refresh` |
| `catalog/datasets/api/router_refresh.py` | `_dispatch_stac_refresh` |
| `catalog/datasets/api/router_refresh.py` | `refresh_dataset` |
| `catalog/datasets/api/router_reupload.py` | `_dispatch_reupload_task`, service branch |
| `catalog/datasets/api/router_reupload.py` | `_dispatch_reupload_task`, raster branch |
| `catalog/datasets/api/router_reupload.py` | `_dispatch_reupload_task`, file branch |
| `catalog/datasets/api/router_analysis.py` | `analysis_materialize_endpoint` |
| `modules/admin/router.py` | `trigger_backfill` |

`commit_import`, `/jobs/{id}/retry` and manifest apply reach the queue through `queue_ingest_job`, so they are covered by its three branches.

`tests/test_commit_attempted_marker_doors.py` walks the AST of `backend/app/` and pins this in both directions: no `defer_async_with_tenant(job_id=...)` call sits outside a function that runs the guard, every guard call names its row, the guard stamps before it defers, and the stamp commits.

## Accepted behaviour changes

Rows created before the stamp existed carry no stamp, so a pending unbound row that predates the deploy reclassifies as `cancelled`. That is the correct outcome for the abandoned uploads it will mostly be (the demo audit found four of those and zero broken commits) and the wrong one for a genuinely broken commit. Accepted rather than migrated: a migration would have to guess the same thing from the same missing fact.

## Noticed and recorded, not fixed here

- A door commits its row and then dispatches, so a process death in the one-statement gap between those leaves the row unstamped and it settles `cancelled` instead of `failed`. That window was unbounded before this change for the same class. `create_fan_out_jobs` closes it outright by putting the stamp in the metadata it commits, because it runs inside a worker task and its child cannot be recreated by repeating a user action: the layer selection lived in the fan-out request body and nowhere else (#1709 r8). Every other door creates its row inside a request the user can simply issue again.
- In S3 storage mode a direct upload binds `staging/{job}/{name}`, which is the BOUND half's business, so an abandoned S3 upload still settles `failed` after the 24h backstop with "Stale: upload completed but never committed". #1744's evidence and scope are the unbound half, and widening the bound half is a separate decision.

## Schema, API, config

None. `cancelled` is already in the `ingest_jobs` status CHECK constraint, in the admin `JobStatus` literal and in `openapi.json`. No migration, no new setting, nothing stored that was not stored before.

## Review round 1 (codex P2, session recovery)

A serialization failure or deadlock on the marker write leaves the `AsyncSession` in a failed transaction, and SQLAlchemy refuses every further statement on it until it is rolled back. The guard handed that session straight to the rollback closure, so `settle_ingest_job_failed` raised too and the job stayed `pending` with no marker: the exact combination the sweep reads as an abandoned upload. It would have cancelled a dispatch that really was attempted.

The stamp now has its own `try`. On failure the session is reset first, then the existing orphan settlement runs, then the usual `DeferFailed` goes out, so the caller's 503 path is unchanged. The settlement block is extracted into `_settle_after_failed_dispatch` so both failure points settle identically.

`create_fan_out_jobs` had the same ordering at its second stamping site: its per-layer commit now carries the child's marker as well as the row, and its exception handler returned without resetting the session, so one layer's transactional failure would fail every remaining layer and strand the parent `fanned_out` with no child importing. It resets too.

`test_a_failed_marker_write_resets_the_session_before_settling` drives a session double that refuses statements until rolled back (with a positive control proving the double is not vacuous) and asserts the reset precedes the settlement and the row ends `failed`. Counterfactual: deleting the `db.rollback()` makes that test fail with `PendingRollbackError`, verified locally.

## Review round 2 (codex P2 x2, rollback expiry)

`Session.rollback()` expires every instance that was in the failed transaction, and `expire_on_commit=False` does not change that: the flag only governs commit. The next synchronous read of an expired column attribute is a lazy load, and an `AsyncSession` has no greenlet to run one, so SQLAlchemy raises `MissingGreenlet`. Verified against SQLAlchemy 2.0.52 and a real session before writing either fix.

So the round 1 reset made the session usable and expired the state the settlement immediately needs. `settle_ingest_job_failed` reads `job.id` and `job.attempt_id` synchronously, and it is the one function every rollback closure in the codebase funnels through. The guard now snapshots those two before the marker write and puts them back after the reset through `set_committed_value`, which marks an attribute loaded with no query and without marking it dirty. A snapshot cannot fail; a reload is one more statement on a connection that has just misbehaved, so the snapshot is the guarantee and a best-effort `db.refresh(job)` covers whatever a caller-supplied closure reads beyond those two.

`create_fan_out_jobs` had the same problem from its own reset: the next layer reads `original_job.source_filename` and friends off that instance, and `restore_fan_out_parent_pending` reads `job.id`, so the reset meant to let the siblings continue would instead turn one layer's failure into a 500 and strand the parent `fanned_out`. It reloads the parent in the same breath. A reload rather than a snapshot there, because the attributes read downstream are spread across two functions and the response, and one SELECT restores all of them.

Tests: the session double now expires its instances on rollback and raises `MissingGreenlet` on a synchronous read until they are restored or reloaded, with a positive control proving the double is not vacuous. Three counterfactuals were run locally, each failing the intended test: dropping the restore and the reload together, dropping only the snapshot restore (caught by a DB-backed test that runs the real guard against a genuinely poisoned transaction with the reload disabled), and dropping the fan-out reload. A fourth test drives a real expired `IngestJob` and shows the restore making exactly the two identifiers readable again while everything else stays expired.

## Caps raised

- `backend/app/platform/jobs/sweep.py` 1827 to 1857. Almost all docstring: why the shape-based carve-out missed the door the demo uses, why the absolute-path class no longer needs its own branch, why pre-stamp rows reclassify rather than get a migration, and the two residual gaps above.
- `backend/app/processing/ingest/service.py` 1465 to 1482 to 1500 to 1513. One `job=` argument at each of five guard call sites, the fan-out child's early stamp and the comment saying why that door is the one worth closing, then the fan-out session reset from review round 1 and the parent reload from round 2.

## Verification

Run from the worktree with `.env.test` and Postgres on `localhost:5434`.

- `uv run pytest tests/test_commit_attempted_marker_doors.py tests/test_defer_orphan_guard.py tests/test_stale_pending_reaper.py tests/test_ingest_fan_out.py tests/test_job_cancel_fan_out.py tests/test_job_cancel_defer_rollback.py tests/test_job_cancel_reupload_commit.py tests/test_embedding_backfill_queue_1542.py tests/test_worker.py tests/test_jobs_router.py tests/test_vrt_creation_173.py tests/test_ingest.py -q -p no:randomly` -> 287 passed (review round 2; 284 at round 1)
- `uv run pytest tests/test_manifest_apply_api.py tests/test_manifest_apply_service.py tests/test_manifest_apply_roundtrip.py tests/test_refresh_gate_1269.py tests/test_service_refresh_1220.py tests/test_reupload.py tests/test_reupload_service.py tests/test_regenerate_vrt_integration.py tests/test_analysis_materialize.py tests/test_import_token_lease_1676.py tests/test_door_token_policy_1746.py tests/test_service_auth_contract_1746.py -q -p no:randomly` -> 389 passed (re-run at review round 2)
- `uv run pytest -q -p no:randomly -n 4` (full backend suite) -> 10668 passed, 100 skipped, 1 failed; the single failure was `test_module_loc_caps_have_no_headroom` against caps that had not yet been re-measured, and `uv run pytest tests/test_layering.py -q` is 47 passed after the raise above
- `uv run pytest tests/test_layering.py -q -p no:randomly` -> 47 passed
- `uv run ruff check .` and `uv run ruff format --check .` -> clean
- `PYTHONPATH=. uv run python scripts/dump_openapi.py --check` -> exit 0, no drift

### DIFF
--- a/backend/app/modules/admin/router.py
+++ b/backend/app/modules/admin/router.py
@@ -1121,6 +1121,7 @@ async def trigger_backfill(
                 job, message_prefix="Failed to queue embedding backfill"
             ),
             db=db,
+            job=job,
         )
     except DeferFailed as dispatch_exc:
         # fix(#1550 review P2, round 1): the orphan guard has already marked the
--- a/backend/app/modules/catalog/datasets/api/router_analysis.py
+++ b/backend/app/modules/catalog/datasets/api/router_analysis.py
@@ -684,6 +684,6 @@ async def analysis_materialize_endpoint(
             **extra_kwargs,
         )
 
-    await defer_with_orphan_guard(_defer, rollback=rollback, db=db)
+    await defer_with_orphan_guard(_defer, rollback=rollback, db=db, job=job)
 
     return AnalysisMaterializeResponse(job_id=job.id, status="pending")
--- a/backend/app/modules/catalog/datasets/api/router_refresh.py
+++ b/backend/app/modules/catalog/datasets/api/router_refresh.py
@@ -604,7 +604,7 @@ async def _dispatch_postgis_refresh(
             dataset_id=str(dataset_id),
         )
 
-    await defer_with_orphan_guard(_defer_refresh, rollback=rollback, db=db)
+    await defer_with_orphan_guard(_defer_refresh, rollback=rollback, db=db, job=job)
 
     return DatasetRefreshResponse(
         run_id=run_id,
@@ -857,7 +857,7 @@ async def _dispatch_stac_refresh(
             dataset_id=str(dataset_id),
         )
 
-    await defer_with_orphan_guard(_defer_refresh, rollback=rollback, db=db)
+    await defer_with_orphan_guard(_defer_refresh, rollback=rollback, db=db, job=job)
 
     return DatasetRefreshResponse(
         run_id=run_id,
@@ -1340,7 +1340,7 @@ async def refresh_dataset(
             credential_ref=credential_ref,
         )
 
-    await defer_with_orphan_guard(_defer_refresh, rollback=_rollback, db=db)
+    await defer_with_orphan_guard(_defer_refresh, rollback=_rollback, db=db, job=job)
 
     return DatasetRefreshResponse(
         run_id=run_id,
--- a/backend/app/modules/catalog/datasets/api/router_reupload.py
+++ b/backend/app/modules/catalog/datasets/api/router_reupload.py
@@ -638,7 +638,7 @@ async def _dispatch_reupload_task(
                 credential_ref=credential_ref,
             )
 
-        await defer_with_orphan_guard(_defer_service, rollback=rollback, db=db)
+        await defer_with_orphan_guard(_defer_service, rollback=rollback, db=db, job=job)
         return
 
     if job.file_path is None:
@@ -669,7 +669,7 @@ async def _dispatch_reupload_task(
                 user_id=str(user_id),
             )
 
-        await defer_with_orphan_guard(_defer_raster, rollback=rollback, db=db)
+        await defer_with_orphan_guard(_defer_raster, rollback=rollback, db=db, job=job)
         return
 
     # Route small files to priority queue
@@ -698,7 +698,7 @@ async def _dispatch_reupload_task(
             user_id=str(user_id),
         )
 
-    await defer_with_orphan_guard(_defer_file, rollback=rollback, db=db)
+    await defer_with_orphan_guard(_defer_file, rollback=rollback, db=db, job=job)
 
 
 @router.post(
--- a/backend/app/modules/catalog/datasets/api/router_vrt.py
+++ b/backend/app/modules/catalog/datasets/api/router_vrt.py
@@ -532,7 +532,7 @@ async def regenerate_vrt_endpoint(
         previous_status=previous_status,
         previous_generation_id=previous_generation_id,
     )
-    await defer_with_orphan_guard(_defer, rollback=rollback, db=db)
+    await defer_with_orphan_guard(_defer, rollback=rollback, db=db, job=job)
 
     return VrtMutationResponse(
         job_id=job.id,
--- a/backend/app/platform/jobs/defer_guard.py
+++ b/backend/app/platform/jobs/defer_guard.py
@@ -41,7 +41,11 @@ from fastapi import HTTPException, status
 from sqlalchemy import update
 from sqlalchemy.ext.asyncio import AsyncSession, async_object_session
 
-from app.platform.jobs.models import IngestJob
+from app.platform.jobs.models import (
+    COMMIT_ATTEMPTED_METADATA_KEY,
+    IngestJob,
+    commit_attempted_marker,
+)
 
 if TYPE_CHECKING:
     # Typing-only edge: `platform/` must not import `processing/` at module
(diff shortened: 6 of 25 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1776, opened 2026-09-02
fix(settings): settings key spelling and ISO option CHECK drift

ishiland:
Addresses two frontend/backend contract bugs from the codebase audit of 2026-08-30 (8dc529f17), tracked in #1778, and adds the drift tests that would have caught them.

### Settings key spelling

`updateBranding` in `frontend/src/api/settings.ts` built its PUT payload as `branding_show_badge`. The persistent_config registry key is `branding.show_badge` (dotted). The PUT handler in `settings/router.py` does no key normalization, so every click of the Admin > Settings > Appearance "Show Powered by GeoLens Footer Label" toggle returned 400 "Unknown setting key: branding_show_badge", in both editions. Fixed to send the registry key as-is.

### ISO metadata option lists vs. records CHECK constraints

`frontend/src/lib/iso-constants.ts` offered `fortnightly` in `UPDATE_FREQUENCY_OPTIONS` and `unclassified`/`secret`/`topSecret` in `SENSITIVITY_OPTIONS`. None of those are accepted by `chk_records_update_frequency` / `chk_records_sensitivity` on `Record`. Nothing validates a PATCH before it reaches the database, and the datasets PATCH path has no `IntegrityError` handler, so picking one of these raised an unhandled CheckViolation that the global DBAPIError handler re-raises as a 500. `savePendingDrafts` then discards the server detail and shows a generic "Failed to save pending edits." toast that names no field, so the whole batch of staged metadata edits is lost with no indication which one caused it.

Replaced the four unstorable values with the two legal sensitivity values the UI never offered (`public`, `internal`) and added translations to all four locales (en/es/fr/de).

### Drift tests

Both bugs are the kind this repo already guards against for capabilities, basemap config, and builder alias keys, but nothing existed yet for settings keys or ISO option vocabularies. Added:

- `backend/tests/test_settings_key_drift.py`: parses every settings-key literal the admin frontend sends or reads and asserts the set is a subset of the `persistent_config` registry. Covers both the `useSettingsForm`-mediated shapes (`key: '...'` field defs, `findSetting` lookups, `settingKey` attrs) and the direct-access shapes the one tab that bypasses that hook uses instead (`settings.find((s) => s.key === '...')`, `onSave({ ... })` payload keys, `onReset('...')`), plus the hand-built payload assignments in `api/settings.ts`.
- `backend/tests/test_iso_option_drift.py`: parses `UPDATE_FREQUENCY_OPTIONS` / `SENSITIVITY_OPTIONS` out of `iso-constants.ts` and asserts each is a subset of the matching CHECK constraint's parsed values.
- `frontend/src/api/__tests__/settings.contract.test.ts`: pins the exact PUT body key `updateBranding` sends.

## Review round 1

A codex pass flagged that the first version of `test_settings_key_drift.py` only recognized the `useSettingsForm` shapes, so a typo in `SettingsPermissionsTab.tsx` (the one tab that reads/saves/resets `role_permissions` directly instead of through that hook) would have kept the test green while still producing the same 400-on-write / silent-default-on-read failure the guard exists to catch. Extended the parser to the three direct-access call shapes, added a positive control asserting the parser finds `role_permissions` in that tab through all three shapes, and added a floor on the total parsed key count so a regex regression that silently shrinks the frontend set fails loudly instead of letting the subset check pass vacuously.

Also switched the in-source comments citing this audit finding from the commit-hash label to the repository's `fix(#NNNN)` inline-comment convention (AGENTS.md), now pointing at #1778.

## Schema/API/config impact

None. No migration, no OpenAPI change, no new registry key. Both fixes make the frontend match a backend contract that already existed.

## Verification

Backend (host, Postgres at localhost:5434):
- `uv run pytest tests/test_settings_key_drift.py tests/test_iso_option_drift.py -v`: 4 passed
- `uv run pytest tests/test_branding_settings.py tests/test_capability_drift.py tests/test_basemap_config_drift.py tests/test_builder_alias_keys_drift.py tests/test_dataset_source_freshness.py tests/test_layering.py -q`: 161 passed
- `uv run ruff check` and `uv run ruff format --check` on the two test files: clean

Counterfactual runs (reverted, then restored, for each of the three guards):
- Reverting `updateBranding` back to `settings.branding_show_badge = data.show_badge` makes `test_frontend_setting_keys_subset_of_backend_registry` fail, naming `branding_show_badge`.
- Reverting `iso-constants.ts` back to include `fortnightly` and `unclassified`/`secret`/`topSecret` makes both `test_iso_option_drift.py` tests fail, naming the offending values.
- Typo'ing the `onReset` argument in `SettingsPermissionsTab.tsx` from `'role_permissions'` to `'role_permission'` makes both `test_frontend_setting_keys_subset_of_backend_registry` and the new `test_parser_floor_and_permissions_tab_positive_control` fail, naming `role_permission`.

All three restored to green afterward.

Frontend:
- `npm run typecheck`: clean
- `npm run lint`: clean
- `npm run test:i18n`: 4 passed
- `npx vitest run src/api/__tests__ src/components/dataset src/components/admin/settings src/hooks/__tests__/use-settings.test.ts`: 570 passed across 59 files

## Left alone

Two lower-severity findings from the same audit rows are not addressed here, since the brief scoped this lane to the branding-key spelling and the ISO-option CHECK mismatch:

- `updateBranding` accepts `Partial<BrandingConfig>` (including `privacy_url`), but only ever forwards `show_badge`. A caller passing `privacy_url` alone would PUT an empty settings object and get a silent no-op "Saved" toast. Today `SettingsAppearanceTab` is the only caller and only ever passes `show_badge`, so this is latent, not live.
- The two legal `sensitivity_classification` values the UI never offered (`public`, `internal`) now render correctly since they're valid `SelectItem` options, but no dedicated test exercises the read-only branch of `SourceQualityTab`, since no test file for that component exists yet.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -202,6 +202,14 @@ jobs:
               - 'frontend/src/lib/__tests__/public-urls.test.ts'
               - 'frontend/src/lib/normalize-style-config.ts'
               - 'frontend/vite.config.ts'
+              # fix(#1778): test_settings_key_drift.py reads these by
+              # literal path (the settings-key subset guard: frontend_keys
+              # is a subset of backend registry keys) and
+              # test_iso_option_drift.py reads iso-constants.ts (the ISO
+              # dropdown-option subset guard).
+              - 'frontend/src/api/settings.ts'
+              - 'frontend/src/components/admin/settings/Settings*Tab.tsx'
+              - 'frontend/src/lib/iso-constants.ts'
               # test_embed_framing_csp.py::test_e2e_sec_audit_s08_skips_when_no_share_token
               # reads this spec by literal path to pin the S08 test block; the
               # 'e2e' filter alone does not trigger Backend Tests.
--- a/backend/tests/test_iso_option_drift.py
+++ b/backend/tests/test_iso_option_drift.py
@@ -1,5 +1,6 @@
-"""Static-analysis test: the frontend ISO metadata dropdown option lists must
-be a subset of the records CHECK constraints they feed.
+"""DB-backed drift test: the frontend ISO metadata dropdown option lists must
+be a subset of the records CHECK constraints they feed, as those constraints
+are actually enforced by the migrated database.
 
 fix(#1778): the ISO metadata dropdowns offer four values the records CHECK
 constraint rejects, so picking them 500s the whole pending-edit batch.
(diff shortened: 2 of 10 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

Issue #1778, opened 2026-09-02
Codebase audit 2026-08-30: follow-up tracking


ishiland:
Tracking issue for the read-only codebase audit run on 2026-08-30/31 against 8dc529f17 (v1.17.0). It exists so in-source comments and PR bodies from the follow-up work have a stable, lookup-able anchor per the inline review-comment convention in AGENTS.md. The audit itself is a private artifact; each PR below quotes the finding it addresses.

The audit's "do these first" list, with the PR taking each:

1. Settings key `branding_show_badge` versus registry key `branding.show_badge` (Appearance toggle 400s on every click since v1.17.0), plus a settings-key drift test; ISO dropdown values the records CHECK rejects, plus a drift test. PR #1776.
2. Two route-table gates (`test_api_contract_openapi.py`, `test_redirect_slashes.py`) that walk `app.routes` and inspect 1 of 265 and 0 of 168 operations after fastapi 0.140 stopped flattening `include_router`; the CI backend paths-filter missing frontend files backend gates read by literal path; PostgreSQL log rotation. Lane in progress.
3. The commit-ambiguity probe from `ingest/router.py` applied to the three raster and VRT commit tails (a dropped connection during COMMIT leads cleanup to delete the COG keys the committed row points at), the vector "NO unlink" sibling, and the vector failure notification. Lane in progress.
4. Feature writes persist shapely's normalized geometry rather than the client's raw dict (an unclosed ring passes validation and then 500s every bbox read). Lane in progress.
5. Global rate limit keyed on handler rather than URL path; the nginx raster-proxy bypass; unvalidated `cols=` in the tile cache key; the `?v=` cache-buster pinning a pool connection. Lane in progress.
6. Export handler releases its pooled connection before ogr2ogr and derives its deadline from the 600 s edge timeout. Lane in progress.
7. `geolens publish --wait` exits non-zero on failed, cancelled, timed-out and token-expired outcomes. PR #1777.
8. Release chain binds the smoked artifact to the tag ref; tag-ref guards on the CLI and SDK publish workflows; `gh release create --verify-tag`. Lane in progress.
9. Destructive button and badge contrast in dark mode; a dark-mode pass in the gating axe suite; the bulk-review mid-commit layer guard. Lane in progress.
10. (Folded into item 2.)

[...]

PR #1779, opened 2026-09-02
fix(release): bind the smoked artifact to the tag ref across the release chain

ishiland:
codebase audit 2026-08-30 (8dc529f17), tracked in #1778: "A workflow_dispatch smoke of `latest` from a tag ref releases that tag, the smoke gate tests an entirely different artifact" and "publish-cli.yml and publish-sdks.yml have no tag-ref guard, a dispatch from a branch publishes to PyPI/npm with zero version assertion".

## What changed

`prod-smoke.yml` resolved the image version to smoke from a `workflow_dispatch` input that defaults to `latest` on any non-push run. A manual re-smoke of a red tag, which is the natural recovery gesture, therefore booted and tested the previous GA's images instead of the tag's own. `release.yml` then trusted that smoke's success for the tag on `head_branch` alone, with no check that the smoke run actually came from a tag push, so a dispatched smoke of a tag could pass and hand `release.yml` a success for a tag nothing had really tested: publishing a GitHub Release, dispatching to the sibling repos, and scheduling the newsletter.

`publish-cli.yml` and `publish-sdks.yml` (both jobs) only ran their tag-matches-version check `if: github.event_name == 'push'`. A `workflow_dispatch` from a branch published to PyPI/npm with no ref-type check and no version assertion at all. `publish-mcp.yml` already carries a hard tag-ref guard for this (fix(#1623)); these two workflows did not.

- `prod-smoke.yml`: the "Assert semver release tag" step and the "Resolve image tag to smoke" step now gate on `github.ref_type == 'tag'` instead of `github.event_name == 'push'`. On any run from a tag ref (push or dispatch) the semver check runs and `VER` is taken from `github.ref_name`, ignoring the `version` input. A dispatch from a branch still falls back to the input, unchanged.
- `release.yml`: the "Assert semver-shaped tag" step now also requires `github.event.workflow_run.event == 'push'`, so only a smoke that itself ran from a tag push can create or update a release. Fixed the stale comment that claimed a dispatch smoke of a tag could not reach this job; it can, since `head_branch` carries the dispatched ref's name either way. Added `--verify-tag` to both `gh release create` invocations.
- `publish-cli.yml` and both jobs of `publish-sdks.yml`: added the same "require a tag ref" guard `publish-mcp.yml` already carries (a hard failure, not a skip, on `GITHUB_REF_TYPE != tag`), and changed the tag-matches-version assertion from `if: github.event_name == 'push'` to `if: ${{ !inputs.dry_run }}` so a dispatched publish is checked too.

No behavior change on the normal push-tag path in any of the four files.

## Before and after

push-tag is unchanged everywhere; it is included for reference.

| Gate | push-tag | workflow_dispatch on a tag ref, before | workflow_dispatch on a tag ref, after | workflow_dispatch on a branch (e.g. main), before | workflow_dispatch on a branch, after |
|---|---|---|---|---|---|
| `prod-smoke.yml`, Assert semver release tag | runs, passes | skipped (`event_name != push`) | runs, passes (`ref_type == tag`) | skipped (`ref_type != tag`) | skipped (`ref_type != tag`), unchanged |
| `prod-smoke.yml`, image version smoked | `github.ref_name` (the tag) | `INPUT_VERSION`, defaults to `latest` (the previous GA's images) | `github.ref_name` (the tag itself); the input is ignored | `INPUT_VERSION` | `INPUT_VERSION`, unchanged (no tag to bind to) |
| `release.yml`, Assert semver-shaped tag | runs, passes | runs, passes (only checked that `head_branch` was semver-shaped, which a dispatched tag ref is) | fails: `workflow_run.event` is `workflow_dispatch`, not `push`; refuses to release | job-level `if` skips this job (`head_branch` does not start with `v` for a normal branch name) | skipped, unchanged |
| `publish-cli.yml` / `publish-sdks.yml`, tag-ref and version guard | runs, passes | skipped entirely (`event_name != push`); publish proceeds with zero check | runs; fails hard if the ref is not a tag, otherwise checks the tag against pyproject/package.json | skipped entirely; publish proceeds with zero check | runs; fails hard on `GITHUB_REF_TYPE != tag`, refusing to publish |
| `release.yml`, `gh release create` | `--verify-tag` added; the tag already exists on the remote at this point, so no behavior change | reached, passed pre-fix | not reached post-fix (the job now dies at the semver-shaped-tag step above) | not reached, before or after (job-level `if` skips) | not reached |

## Verification

- `actionlint` (installed via Homebrew, `brew list actionlint`, 1.7.12) on all four touched files, plus `publish-mcp.yml` and `verify-published.yml` for comparison. Result: one pre-existing shellcheck info-level finding each, in `release.yml` (SC2016, on an unrelated line, confirmed present on `main` before this branch's changes) and in `verify-published.yml` (SC2086, an untouched file). No new findings from this change.
- `python3 -c "import yaml, sys; [yaml.safe_load(open(f)) for f in sys.argv[1:]]"` on all four touched files: parses clean.
- `backend/tests/test_release_dispatch_targets_1609.py` (parses `release.yml`, matches its dispatch step by name): 14 passed.
- `backend/tests/test_ci_job_timeouts.py` (covers `prod-smoke.yml`'s job timeout): 5 passed.
- `backend/tests/test_layering.py`: 47 passed. This PR touches no files under `backend/app/`; ran it anyway per the lane's standing instruction.

Commands run from the worktree:
```
cd backend && set -a && source ../.env.test && set +a && uv run pytest tests/test_release_dispatch_targets_1609.py tests/test_ci_job_timeouts.py tests/test_layering.py -q
```

## Left alone

- `verify-published.yml`'s two duplicated `--jq '.jobs[] | select(.name == "Create GitHub Release") | .conclusion'` selectors (currently at lines 274 and 503) still pick the upstream `release.yml` job by display-name string, as the register excerpt cites. Out of scope for this lane. A rename of that job would still make both verify jobs report a hollow success with no red signal.
- The `dep-audit.yml` (pip-audit coverage gap for `cli/`, `mcp/`, `sdks/python/`) and `cli/pyproject.toml` (CI Python-version floor drift) rows in the register excerpt are explicitly out of scope for this lane.

## Already fixed on main

None. All five in-scope findings (the four numbered items in the brief, plus the verify-published.yml note) were verified present in the four touched files before this PR's changes.

### DIFF
--- a/.github/workflows/prod-smoke.yml
+++ b/.github/workflows/prod-smoke.yml
@@ -60,8 +60,13 @@ jobs:
       # "Assert semver-shaped tag" guard (keep the two in sync). A red run on a
       # junk tag is correct and cheap: release.yml fires only on this
       # workflow's SUCCESS, so nothing downstream advances.
+      #
+      # fix(#1778): gate on ref_type, not event_name. A workflow_dispatch run
+      # from a tag ref (`gh workflow run prod-smoke.yml --ref vX.Y.Z`) must be
+      # held to the same semver check as a tag push, or a re-smoke of a junk
+      # tag would skip straight past this and into the pull-retry loop below.
       - name: Assert semver release tag
-        if: github.event_name == 'push'
+        if: github.ref_type == 'tag'
         env:
           TAG: ${{ github.ref_name }}
         run: |
@@ -80,9 +85,17 @@ jobs:
           # the script body, where it would be evaluated as shell.
           INPUT_VERSION: ${{ github.event.inputs.version }}
         run: |
-          # On a tag push, github.ref_name is the tag (vX.Y.Z). On manual
-          # dispatch, use the input (default latest).
-          if [ "${{ github.event_name }}" = "push" ]; then
+          # fix(#1778): bind the smoked artifact to the ref, not the event.
+          # On ANY run from a tag ref — a tag push, or a workflow_dispatch
+          # re-run from that tag — github.ref_name IS the tag (vX.Y.Z), so
+          # ignore INPUT_VERSION and smoke that tag's own images. A dispatch
+          # defaults INPUT_VERSION to "latest", which is the PREVIOUS GA's
+          # images; using it here would let a re-smoke of a red tag pass by
+          # smoking a different artifact than the one release.yml is about to
+          # release (the tag it fires on is bound to THIS run's ref, not to
+          # VER). Only a dispatch from a branch (no tag to bind to) falls back
+          # to the input.
+          if [ "${{ github.ref_type }}" = "tag" ]; then
             VER="${{ github.ref_name }}"
           else
             VER="$INPUT_VERSION"
--- a/.github/workflows/publish-cli.yml
+++ b/.github/workflows/publish-cli.yml
@@ -66,10 +66,21 @@ jobs:
         with:
           python-version: "3.14"
 
-      - name: Assert tag matches package version (tag push only)
-        if: ${{ github.event_name == 'push' }}
+      # fix(#1778): every publishing run must come from a tag ref, and this is
+      # a hard failure rather than a skipped check, same as publish-mcp.yml's
+      # #1623 guard: a manual dispatch defaults to the branch selected in the
+      # Actions UI, and the repo sits at the last released version between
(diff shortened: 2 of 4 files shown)


ishiland:
@codex review

PR #1780, opened 2026-09-02
fix(features): write shapely's normalized geometry, not the client's raw dict

ishiland:
Codebase audit 2026-08-30 (8dc529f17), tracked in #1778. Finding: "An unclosed polygon ring passes both write validators and poisons every bbox read of the dataset with a 500".

## What changed

`insert_feature`, `replace_feature` and `update_feature` in `backend/app/modules/catalog/features/service.py` validated geometry with shapely (`_validate_geometry_structure`) but then serialized the client's original dict (`json.dumps(geometry)`) into `ST_GeomFromGeoJSON`. Shapely auto-closes an unclosed polygon ring and reports it valid, but `ST_GeomFromGeoJSON` does not repair the equivalent GeoJSON, so the row was stored with an open ring. Every later bbox read whose envelope touched that feature raised an unhandled 500 from `ST_Intersects` ("Points of LinearRing do not form a closed linestring"), reaching anonymous viewers of public datasets, while vector tiles kept rendering fine because the tile query tolerates invalid geometry.

`_validate_geometry_structure` now returns the shapely geometry it already builds, and all three write sites bind `shapely.to_geojson(normalized_geom)` instead of re-serializing the raw request dict. This closes the whole class of shapely-repairs-it/PostGIS-does-not divergences, not only ring closure.

## Schema/API impact

None. No request or response schema changed; `make openapi-check` passes with no diff. The response shape for already-valid input is unchanged (verified with a byte-equal round-trip test).

## Tests added

`TestGeometryValidity` in `backend/tests/test_features_crud.py`:
- `test_closed_ring_round_trips_unchanged`: an already-closed ring produces byte-identical response GeoJSON.
- `test_insert_unclosed_ring_stored_closed_and_bbox_readable`: POST with an unclosed ring is stored closed, and a subsequent bbox GET returns 200 with the feature.
- `test_replace_unclosed_ring_stored_closed_and_bbox_readable`: same guard on PUT.
- `test_patch_geometry_unclosed_ring_stored_closed_and_bbox_readable`: same guard on PATCH (partial geometry update).

## Verification run

- `uv run pytest tests/test_features_crud.py -q` — 44 passed
- `uv run pytest tests/test_feature_geometry_limits.py tests/test_features_geojson_z.py tests/test_ogc_features.py tests/test_ogc_features_filter.py -q` — 117 passed
- `uv run pytest tests/test_layering.py -q` — 47 passed (no cap changes needed; `features/service.py` is not in `_MODULE_LOC_CAPS` and stayed under any other size gate)
- `uv run ruff check .` and `uv run ruff format --check .` — pass
- `make openapi-check` — pass, no diff

Counterfactual: stashed the `service.py` fix, kept the new tests, and reran `tests/test_features_crud.py -k TestGeometryValidity`. The three unclosed-ring tests failed as expected: the write response showed the ring stored open (`ring[0] != ring[-1]`), and hitting the bbox endpoint directly against the unfixed code reproduced the exact audit-cited crash: `sqlalchemy.exc.InternalError ... IllegalArgumentException: Points of LinearRing do not form a closed linestring` from the `ST_Intersects` query. Restored the fix and reran; all 8 pass.

## Left alone

Two other rows in the same audit excerpt are context, not in this lane's scope, and are unaddressed:
- `schemas.py:12` `MAX_POSITION_DIMENSIONS = 4` advertises XYZM positions the service unconditionally rejects with a 400 (low severity, low effort) — still present on main.
- The `chat_geojson.py` empty-geometry/NaN-bbox finding is explicitly out of scope for this lane.

The register's third row (`TestGeometryValidity`'s fixtures never probed ring closure) is addressed as a side effect of the tests added here.

### DIFF
--- a/backend/app/modules/catalog/features/service.py
+++ b/backend/app/modules/catalog/features/service.py
@@ -2,14 +2,15 @@
 
 from __future__ import annotations
 
-import json
 import math
 import re
 from collections.abc import Sequence
 from typing import TYPE_CHECKING
 
+from shapely import to_geojson
 from shapely.errors import GEOSException
 from shapely.geometry import shape as shapely_shape
+from shapely.geometry.base import BaseGeometry
 from shapely.validation import explain_validity
 from sqlalchemy import func, text
 from sqlalchemy.ext.asyncio import AsyncSession
@@ -506,7 +507,7 @@ async def effective_geometry_type(session: AsyncSession, dataset) -> str:
     return dataset.geometry_type
 
 
-def _validate_geometry_structure(geometry: dict) -> None:
+def _validate_geometry_structure(geometry: dict) -> BaseGeometry:
     """Reject degenerate or topologically invalid geometry before PostGIS.
 
     fix(#458 E-02): degenerate-but-schema-valid input (2-point polygon rings,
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

ishiland:
@codex review

PR #1781, opened 2026-09-02
fix(export): release the pooled connection and bound the export by the edge

ishiland:
Two findings from the codebase audit 2026-08-30 (8dc529f17), tracked in #1778:

- "The export handler holds an API DB-pool connection open for the entire ogr2ogr conversion (up to 1 hour)"
- "Synchronous export runs on a 3600s subprocess deadline behind a 600s edge read timeout, and uvicorn never cancels the orphaned handler"

Both were still present on current main. Nothing under `backend/app/processing/export/` had changed since the audit commit, and the two pieces of evidence reproduce exactly: `rg 'db\.(commit|rollback|close)\(\)' backend/app/processing/export/router.py` returned only the two commits the audit named, and `export/ogr.py` still imported the worker's `OGR2OGR_FILE_TIMEOUT_SECONDS`.

## The pooled connection

`get_db` yields one session for the whole request. `port.get_dataset` checked a connection out and nothing released it, so the handler kept one of the API process's `db_pool_size + db_max_overflow` connections across ogr2ogr, the file hash and the object-store upload. Export is reachable anonymously for a public and published dataset, so a small number of concurrent exports could starve every other request in the process on `pool_timeout`.

The handler now copies the five values the rest of it needs (title, table name, column info, whether the dataset has geometry, the spatial extent) into locals, drops the `dataset` name so a later ORM read is a NameError rather than a MissingGreenlet at runtime, and rolls the session back before the conversion. It rolls back a second time before the hash and the upload, which matters for the GeoParquet branch: that one streams its rows through the same session, so it checks a connection back out. The audit row at the end re-acquires. This is the treatment `processing/tiles/router.py` already had from fix(#1451 codex P1) for a query that holds a connection roughly a thousand times less long.

`_emit_export_audit` now takes a `user_id` instead of the `Identity`. The request's `user` is the concrete `User` ORM instance on the session being rolled back, so reading `user.id` after the release raised MissingGreenlet. The id is read once, while the loading transaction is still open.

## The deadline

`export/ogr.py` imported the offline worker's `OGR2OGR_FILE_TIMEOUT_SECONDS` (3600s) and used it for both the subprocess wall clock and the libpq `statement_timeout` of a request-scoped conversion. The only `proxy_read_timeout` in `frontend/nginx.conf` is 600s, so past ten minutes nginx severed the upstream read and answered 504 while the handler kept converting for up to another fifty minutes.

The route now stamps `time.monotonic() + EDGE_PROXY_READ_TIMEOUT_SECONDS` on entry and threads that deadline through `export_dataset` to `run_ogr2ogr_export`, which reads what is left of it at spawn time. The bound is therefore the time actually remaining, not an allowance: a fast pre-conversion phase hands its unused time to the conversion, and a slow one (an unindexed `_count_selected_features` scan, a `plan_parquet_export` over a wide table) takes it away.

The only thing subtracted is what is still outstanding when the subprocess starts: a 120s margin for the format-specific finish inside `export_dataset` (the shapefile ZIP, the GeoPackage timestamp normalization), the file hash, the object-store upload and the audit commit, plus one `db_pool_timeout` for the audit row's checkout, which is the single pool wait the elapsed clock cannot already contain because it has not happened yet. Every earlier checkout is behind us and is already priced into elapsed time. `EDGE_PROXY_READ_TIMEOUT_SECONDS` comes from `processing/ingest/url_fetch.py`, the constant the URL importer already derives its own synchronous budget from, so there is one edge number rather than two.

The clock is monotonic so an NTP step or a suspend cannot lengthen or shorten a running export. A remainder at or below zero clamps to a one second floor: the request is lost either way, and it should fail through the ordinary timeout path with the ordinary error rather than by handing `asyncio.wait_for` a negative number. A pool timeout large enough that the reserve alone exceeds the edge window is a configuration problem rather than a slow request, and logs once with the cause.

The interval before the route body runs (dependency resolution: auth, the permission lookup, and any pool wait either incurs) sits outside this clock. Nothing inside the handler can observe it; the reserve is what absorbs it, and it is written down at the function rather than left implied.

The comment in `export/ogr.py` that said `_communicate_with_timeout` also kills the child on a client disconnect is corrected. It does kill on cancellation, and the ingest caller gets that from Procrastinate's shutdown, but nothing cancels this task: uvicorn's `connection_lost` only marks the cycle disconnected and wakes `receive()`, and a GET handler never awaits `receive()`.

## Client disconnect

Not addressed, deliberately. The brief allowed `request.is_disconnected()` at the poll points the handler already has, and there are none: the conversion is a single await, so polling would need a concurrent watchdog task plus a response status for a GET that declares none. The deadline is what bounds the orphan now, from an hour down to what remains of the window the client was going to give up on anyway.

None. No request or response schema changed, no route signature changed, no new setting. `make openapi-check` is clean.

`export_dataset` and `run_ogr2ogr_export` each gained a keyword-only `deadline: float | None = None`. Seven test doubles that stand in for one of them with an explicit signature were updated to accept it; `None` keeps the direct-call tests and any future offline caller on the same arithmetic with an elapsed time of zero.

## Verification

Run from the worktree with `.env.test` sourced and Postgres on localhost:5434.

- `uv run pytest tests/test_export_request_budget.py tests/test_export.py tests/test_export_access.py tests/test_export_antimeridian.py tests/test_export_artifact_cache_1532.py tests/test_export_hardening.py tests/test_export_head_1513.py tests/test_export_parquet.py tests/test_export_parquet_writer.py tests/test_export_temp_cleanup.py tests/test_export_where_validator.py tests/test_tenant_ingest_export_routing.py tests/test_staging_runtime.py tests/test_layering.py tests/test_rule1_structural.py tests/test_rule2_structural.py tests/test_api_key_scope_875.py tests/test_ogr_communicate_cancellation.py tests/test_ogr_subprocess_env.py tests/test_worker_exports_sweep.py -q -p no:randomly`: 570 passed, 3 skipped
- `uv run ruff check .` and `uv run ruff format --check .`: clean
- `PYTHONPATH=. uv run python scripts/dump_openapi.py --check` (the body of `make openapi-check`, which needs the env the Makefile target does not source in a worktree): exit 0, no drift
- `pre-commit run --files` over the eleven changed files: all hooks pass
- After the codex round 1 rework, a focused re-run of `tests/test_export_request_budget.py tests/test_export_antimeridian.py tests/test_export_head_1513.py -q -p no:randomly`: 72 passed

## Caps

None raised, and none needed. Neither source file has an entry in `_MODULE_LOC_CAPS`: `export/router.py` is 1088 lines against the 1500 default in `test_router_orchestrator_modules_stay_within_loc_cap`, and `export/ogr.py` is 432, below the 1000-line `_RATCHET_INCLUSION_LOC` threshold that would pull it into the ratchet.

## Noticed and left alone

- The same false cancellation premise appears in two comments outside this lane's two files: `processing/export/service.py` ("BaseException, not Exception: a client disconnect cancels this task") and the `_communicate_with_timeout` docstring in `processing/ingest/ogr.py`. The ingest one is half true, since Procrastinate really does cancel worker jobs, and that file belongs to another active lane, so neither was touched.
- The GeoParquet branch has no wall-clock deadline of its own and never did; the 3600s constant only ever bounded the ogr2ogr subprocess. The request deadline is now available to give it one, but that is a behaviour change on a path this lane was not asked to bound.
- On a cache hit the handler holds its connection across two object-storage metadata reads before committing. That is milliseconds, and the response streams after the commit, so it is not the same hazard.
- The audit's KISS note about decomposing the 715-line handler is out of this lane's scope.

### DIFF
--- a/backend/app/api/middleware/logging.py
+++ b/backend/app/api/middleware/logging.py
@@ -31,7 +31,20 @@ class RequestLoggingMiddleware(BaseHTTPMiddleware):
         # ID without parsing client-supplied headers (RESILIENCE-9).
         request.state.request_id = request_id
 
-        start_time = time.perf_counter_ns()
+        # fix(#1778 codex r2): the moment the request entered the app, on the
+        # same clock a handler can compare against. A route that must answer
+        # inside the edge proxy's read timeout cannot start its own clock at
+        # its function body: FastAPI resolves that route's dependencies first,
+        # and one of those can block on a database pool checkout for as long
+        # as ``db_pool_timeout``. Stamped here, beside the request id and for
+        # the same reason, so the handler measures from where the proxy's own
+        # clock started rather than from where it got control.
+        #
+        # ``monotonic`` rather than ``perf_counter``: both are monotonic and
+        # both are ns-resolution here, and one read serves both the deadline
+        # below and the duration this middleware already logs.
+        started_at = time.monotonic()
+        request.state.started_at_monotonic = started_at
         response: Response | None = None
 
         try:
@@ -40,7 +53,7 @@ class RequestLoggingMiddleware(BaseHTTPMiddleware):
             structlog.stdlib.get_logger("api.error").exception("Unhandled exception")
             raise
         finally:
-            duration_ms = (time.perf_counter_ns() - start_time) / 1_000_000
+            duration_ms = (time.monotonic() - started_at) * 1000
             status_code = response.status_code if response is not None else 500
 
             access_logger.info(
--- a/backend/app/processing/export/ogr.py
+++ b/backend/app/processing/export/ogr.py
@@ -107,16 +107,23 @@ def export_subprocess_timeout_seconds(deadline: float | None) -> float:
 
     floored at ``EXPORT_BUDGET_FLOOR_SECONDS``.
 
-    ``deadline`` is a ``time.monotonic()`` stamp taken when the route was
-    entered, offset by ``EDGE_PROXY_READ_TIMEOUT_SECONDS``. ``None`` means the
-    caller is not serving a request (the direct-call tests, and any future
-    offline caller): the full edge window is assumed to start now, which is
-    the same arithmetic with an elapsed time of zero.
+    ``deadline`` is a ``time.monotonic()`` stamp taken when the request
+    entered the app (``RequestLoggingMiddleware``), offset by
+    ``EDGE_PROXY_READ_TIMEOUT_SECONDS``. ``None`` means the caller is not
+    serving a request (the direct-call tests, and any future offline caller):
+    the full edge window is assumed to start now, which is the same
+    arithmetic with an elapsed time of zero.
 
-    The interval before the route body runs (dependency resolution: auth, the
-    permission lookup, and any pool wait either incurs) sits outside this
-    clock and is the one residual it cannot see. Nothing inside the handler
-    can observe it, and the reserve above is what absorbs it.
+    Anchoring it at the middleware rather than at the route body is what puts
+    dependency resolution inside the clock. ``Depends(get_optional_user)``
+    runs before the body and can block on a pool checkout for as long as
+    ``db_pool_timeout``; a clock started in the body would not see that, and
+    an export that spent its whole calculated budget could still answer after
+    the proxy had given up. What remains outside is the handful of outer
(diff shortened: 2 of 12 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1782, opened 2026-09-02
fix(a11y): destructive contrast, dark-mode axe pass, bulk mid-commit guard

ishiland:
Fixes three findings from the codebase audit 2026-08-30 (8dc529f17), tracked in #1778: register excerpt AUD9-fe-a11y-bulk.json, audit "do first" item 9.

All three findings were still present on main; none had already been fixed.

## Destructive button/badge dark-mode contrast

button.tsx's and badge.tsx's destructive variant paired `dark:bg-destructive/60` with `text-destructive-foreground`. That foreground token is tuned dark-on-dark for the solid light-mode fill (`--destructive-foreground: oklch(0.985 0 0)` in light, `oklch(0.25 0.05 22)` in dark). Composited at 60% over `--card` in dark mode it measures 2.61:1, below the 4.5:1 WCAG 1.4.3 floor for normal text.

I reimplemented the OKLCH to sRGB to WCAG contrast math independently rather than trusting the audit's numbers and reproduced them exactly:

- 60% destructive over dark `--card`: rgb(159, 67, 70). Against dark `--destructive-foreground` rgb(54, 23, 23): 2.61:1. Against white: 6.24:1.
- Dropping `dark:bg-destructive/60` (solid fill) against the same dark foreground: 5.63:1, which would pass, but the translucent dark fill matches a pattern already used by other variants in the same file (`dark:bg-input/30`, `dark:hover:bg-accent/50`), so I kept it and changed the text color instead.
- Light-mode solid fill against plain white: 5.98:1 (was 5.73:1 against the near-white `--destructive-foreground` token).

Both variants now use `text-white` instead of `text-destructive-foreground`, matching upstream shadcn's own destructive-button recipe. `frontend/src/components/report/ReportProblemHost.tsx` also uses `text-destructive-foreground`, but on a solid `bg-destructive` fill with no `/60` modifier, which measures 5.63:1 in dark mode and passes; left untouched.

Added `frontend/src/components/ui/__tests__/destructive-contrast.test.tsx`, which renders both variants and pins that `text-destructive-foreground` is not paired with the translucent dark fill.

## Dark-mode axe pass

`playwright.config.ts` sets no `colorScheme`, so every Playwright browser context defaults to light, and the app's `ThemeProvider` (`defaultTheme="system"`) resolves off `prefers-color-scheme` with nothing stored in `localStorage`. The entire dark palette was structurally unreachable to the gating axe suite.

`e2e/accessibility.spec.ts` had one project-wide `beforeAll`/`afterAll` (dataset, collection, map, share) and about 19 axe scans hung off routes and dialogs reachable from those fixtures. Duplicating that setup per color scheme (via a second Playwright project) would double the seeded fixtures for no reason, so instead the existing scans are wrapped in a `for (const colorScheme of ['light', 'dark'])` loop with `test.describe` and `test.use({ colorScheme })` per pass, keeping the shared `beforeAll`/`afterAll` outside the loop. Test count goes from 19 to 38 (light and dark for each route); the shared fixtures are still seeded once.

Ran the full suite (both schemes) from the :5174 worktree recipe: 38/38 passed. No dark-mode-only violation surfaced beyond the button/badge contrast fix above, which this run now also exercises.

Left alone, not part of this lane's scope: the same wcagTags list also excludes axe's `target-size` rule (tag `wcag22aa`), which the audit's third finding calls out separately for several 22px builder row controls; that finding was not in this lane's deliverable and is untouched. `playwright.builder-hardening.config.ts` and `e2e/record-detail-ux-audit.spec.ts` have the same light-mode-only gap and were not named in this lane's brief; also untouched.

## Bulk upload mid-commit guard

The #1708 r24 "no layer change mid-commit" guard, applied to `UrlImportForm.tsx`'s `handleLayerChange`, was never applied to the bulk upload path. `BulkReviewList.tsx`'s layer/sheet `<Select>` had no `disabled` state, and `UploadForm.tsx`'s `handleSheetChange` drove the entry back to `'preview'` unconditionally on any change, including while a commit was in flight. That re-enabled the commit button (and the belt-and-braces UI wiring in `ReviewFormBlock`) against a request the backend had already accepted, and a second commit on the same job id would be rejected as a 400, misreporting a still-running import as failed.

Mirrored the same guard shape UrlImportForm.tsx already carries: `disabled={entry.status === 'committing'}` on the `<Select>`, and an early `if (entry.status === 'committing' || entry.status === 'tracking') return;` in `handleSheetChange` before it touches the entry.

Added two tests: `BulkReviewList.midCommitGuard.test.tsx` pins the Select's disabled state, and `UploadForm.midCommitGuard.test.tsx` drives a full drop-preview-commit flow with a commit held pending, then confirms a mid-commit `onSheetChange` call is a no-op (no second `previewFile` call, status stays `committing`). Both tests fail without the fix (verified locally before restoring it) and pass with it.

Left alone: the audit's verification notes that `handleCommitSingle` in `UploadForm.tsx` also has no internal in-flight guard, a distinct double-commit race (rapid double-click on the commit button itself, not the layer picker) that this lane's brief did not ask to fix. `FilterPanel.tsx`, `UrlImportForm.tsx`, and `StacImportForm.tsx` rows from the same register excerpt are out of scope for this lane; the latter two are owned by lane 1712.

## Impact

Frontend and e2e test changes only. No schema, API, or config changes.

## Verification

- `cd frontend && npx vitest run src/components/ui/__tests__/destructive-contrast.test.tsx src/components/import/__tests__/BulkReviewList.midCommitGuard.test.tsx src/components/import/__tests__/UploadForm.midCommitGuard.test.tsx`: 3 files, 5 tests passed
- `cd frontend && npx vitest run` (full suite): 428 files, 5682 tests passed
- `cd frontend && npm run typecheck`: passed
- `cd frontend && npm run lint`: passed
- `cd frontend && npm run test:i18n`: passed (4 tests, no new t() keys added)
- Vite on :5174 with `API_PROXY_TARGET=http://localhost:8001`, then `E2E_ALLOW_WORKTREE=1 E2E_BASE_URL=http://localhost:5174 npx playwright test e2e/accessibility.spec.ts --project=chromium`: 38/38 passed (19 routes, light and dark)

### DIFF
--- a/e2e/accessibility.spec.ts
+++ b/e2e/accessibility.spec.ts
@@ -146,238 +146,251 @@ test.describe('Accessibility - WCAG 2AA', () => {
     });
   });
 
-  test('public search page has no accessibility violations', async ({ page }) => {
-    await page.goto('/');
-    await page.waitForLoadState('networkidle');
+  // fix(#1778): every gating axe scan below runs in both color schemes.
+  // playwright.config.ts's `use` block sets no colorScheme, so Playwright
+  // defaults to light and the ThemeProvider's "system" default
+  // (main.tsx:106) resolves off that media query -- the dark palette was
+  // structurally unreachable to axe. beforeAll/afterAll above are shared
+  // across both passes so the fixture dataset/map/collection are seeded
+  // once, not once per scheme.
+  for (const colorScheme of ['light', 'dark'] as const) {
+    test.describe(`(${colorScheme} mode)`, () => {
+      test.use({ colorScheme });
 
-    const results = await new AxeBuilder({ page })
-      .withTags(wcagTags)
-      .analyze();
+      test('public search page has no accessibility violations', async ({ page }) => {
+        await page.goto('/');
+        await page.waitForLoadState('networkidle');
 
-    expect(results.violations, formatViolations(results.violations)).toEqual([]);
-  });
+        const results = await new AxeBuilder({ page })
+          .withTags(wcagTags)
+          .analyze();
 
-  test.describe('logged-out routes', () => {
-    test.use({ storageState: { cookies: [], origins: [] } });
-
-    test('login page has no accessibility violations', async ({ page }) => {
-      await page.goto('/login');
-      await page.waitForLoadState('networkidle');
-
-      const results = await new AxeBuilder({ page })
-        .withTags(wcagTags)
-        .analyze();
-
-      expect(results.violations, formatViolations(results.violations)).toEqual([]);
-    });
-
-    test('public saved-map output has no accessibility violations', async ({ page }) => {
-      await page.goto(`/m/${shareToken}`);
-      await expect(page.getByText(builderMapName)).toBeVisible({ timeout: 15_000 });
-      await page.getByRole('button', { name: 'Map data' }).click();
-      const dataDialog = page.getByRole('dialog', { name: 'Map data' });
-      await expect(dataDialog).toBeVisible();
-      await expect(dataDialog.getByText(datasetTitle).first()).toBeVisible();
-      await expect(
-        dataDialog.getByRole('region', { name: 'Map layer and feature data' }),
-      ).toBeVisible();
-      await expect(dataDialog.getByText('E2E Test Point', { exact: true })).toBeVisible({
-        timeout: 15_000,
-      });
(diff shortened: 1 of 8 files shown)


ishiland:
@codex review

PR #1784, opened 2026-09-02
fix(ingest): decide the publish reap by observation, not by the commit's ack

ishiland:
Three findings from the codebase audit of 2026-08-30 (8dc529f17), tracked in #1778. I re-checked each one against main at 3b79588e4 before writing code; all three were still present, and none of them had been fixed by the week of merges since the audit ran.

## 1. "A cancel or connection loss during `session.commit()` makes the raster-replace failure cleanup delete the live COG the committed row now points at"

A commit whose acknowledgement is lost may still have been applied by PostgreSQL. Every raster and VRT publish tail set its "published" flag on the line after `await session.commit()` and read that flag in the terminal cleanup, so a dropped connection, or the `asyncio.CancelledError` that #1709's `abort=True` delivers, left the flag false and the cleanup deleted the exact object keys the committed row had just been pointed at. The superseded objects are not restored, because the reaper of the old keys never ran, so the dataset names bytes that no longer exist and nothing in the product can put it back.

`publish_commit_landed` in `tasks_raster_common.py` applies the reasoning `_url_import_transition_landed` already carries for the URL-import path (fix #1708 codex r11): read the row back on a fresh session and decide from what you observe, standing down when the probe itself cannot read. One helper serves every tail because all of them stamp the job `complete` in the same transaction as the pointer swap, so `status == 'complete'` fenced on this attempt's id is an exact signal, and no other attempt can produce it (fresh attempt token, `retry=0`).

Four call sites, not the three the brief named. The audit's excerpt cites `tasks_raster_replace.py`, `tasks_raster.py` and `regenerate_vrt`, but `ingest_vrt` in the same file has the same tail with the same consequence, so it is fixed here too rather than left as the one that still deletes what it published.

A tail that observes its publish landed now STANDS DOWN: it returns instead of re-raising, so no failure write runs at all. Gating only the storage reap was not enough, and codex round 1 found why. The tail still entered its failure handler, and in `regenerate_vrt` two of that handler's three writes are fenced (the job on `running`, the asset on `current_generation_id`) while the third is not: it reloads the now `completed` `VrtGeneration` and stamps it `failed`. `get_vrt_status` derives `last_generation_at` from the latest completed generation, so the regeneration vanishes from the UI, and the stale-generation sweep reads the same row as evidence the asset was unhealthy.

Returning means swallowing the `CancelledError` a cancel delivers, so `absorb_cancellation` calls `Task.uncancel()` first. Suppressing a cancellation without saying so leaves `Task.cancelling()` above zero for a structured-concurrency parent to misread.

The stand-down cannot cover every route into the handler. `invalidate_catalog_cache`, `defer_embedding`, the prior-key reap and the billing event all run inside the same `try` as the publish, so a Valkey outage lands there after the dataset is live. Three tails now refuse to run their failure writes once the published flag is set. `ingest_raster`'s is the one an operator notices, because its handler emails them that an ingest they can see in the catalog failed. `reupload_raster` keeps no such guard and says why in place: its post-swap followups have their own fence, and every write in its handler is already fenced against a completed job. The generation update also gained a `status != 'completed'` fence, so the rule is stated at the write and not only at the caller, beside the two fences that were already there.

Round 2 found the stand-down was too broad by one step: `reupload_raster` returned before `_run_post_swap_followups`, which holds the only deletion of the superseded COG and quicklooks. The committed pointer already names the new keys, so each lost acknowledgement stranded three objects that no row references and no quota counts, for the life of the dataset. Standing down from the failure handler is not standing down from the success work, so both stand-downs now run their reclaiming followup first, through a shared wrapper (`run_post_swap_followups_best_effort`) and a named `_reap_superseded_generation_objects` that carry the fence both paths need. `ingest_vrt` and `ingest_raster` were checked for the same gap: a first build supersedes nothing, and what they skip is the cache purge, the embedding defer, the completion notification and the metering event, all recoverable and none of them holding unreferenced bytes. Their comments record that.

The reap keys off the probe's answer instead of `final_status`. That also closes the post-commit half of the same class in `tasks_raster.py`: its broad handler sets `final_status = "failed"` even when the failure happened after the swap was durable, which is what fix(#1290 review) corrected in the replace tail and never carried over. `final_status` keeps deciding the staged-original deletions, deliberately, because standing down there would turn a probe failure into a second and worse deletion.

`reupload_raster` sat one point under the McCabe ceiling, so the two branches that always did the same unlink are merged into one condition to pay for the stand-down branch.

## 2. "The upload-safety validation exit is copy-pasted into four ingest tasks; the fix(#1290) NO unlink correction reached only the two raster copies"

`tasks_vector.ingest_file`'s copy still unlinked, so on a local-storage install a worker-side validation failure deleted the user's only copy of the file it had just recorded as failed. The canonical trigger needs nobody hostile: lowering `UPLOAD_MAX_SIZE_MB` while a job sits queued fails it at worker pickup. The exit now leaves the decision to `_should_unlink_staging` in the terminal `finally`, which already distinguishes a durable local original from a downloaded scratch copy and runs on that return.

## 3. "Four ingest tasks hand-roll the failure-write block instead of calling `_cleanup_staging_on_failure`"

`ingest_file`, `ingest_service` and `ingest_vrt` route through the shared helper now. The three things their copies left out are the three that matter to somebody: the `redact_url_credentials` backstop on the persisted message, the `pending`-inclusive attempt fence from fix(#1274 review), and the `ingest_failed` notification. An operator with `notify_on_ingest_failed` on was told about raster imports and re-uploads and heard nothing when a vector file import, a service import or a VRT build failed.

The helper skips its `DROP` when handed an empty staging-table name, which is what lets the VRT tail use it: interpolating `""` raises inside the best-effort guard and would have logged a cleanup failure on every VRT build failure while saying nothing true.

Round 2 found that the helper ran its `DROP` before writing the failure row. PostgreSQL aborts the whole transaction on any statement error, so a drop that hit a lock or statement timeout left the session unusable and the failure `UPDATE` that followed raised `current transaction is aborted`. The job stayed `running` with no reason recorded until the stale sweep, and routing three more tails through this helper widened that from two callers to five. The failure row is written and committed first now, and the drop runs last in its own guarded block that rolls back its own wreckage. The drop still runs before the rowcount early return, because the table name is attempt-scoped and belongs to this attempt whether or not a newer one owns the job row.

`tasks_raster.ingest_raster` keeps its own copy. It is the one tail that already emits `ingest_failed`, from two hand-written blocks, and routing it through the helper would change its subject lines and move the emission on its early validation exit. Listed under "left alone" rather than silently skipped.

None. Every change is inside Procrastinate task bodies and one shared helper. `dump_openapi.py --check` passes with no drift, so no SDK or `api.generated.ts` regeneration is owed. No migration, no new setting, no user-facing copy, no new translation key.

## Verification

Run from this worktree with `.env.test` and Postgres at localhost:5434, on the committed tree at 1410c15d6.

- `uv run pytest tests/ -q -p no:randomly -k "vrt or raster or ingest or staging or reupload or job or event"`: 1957 passed, 5 skipped, 8824 deselected, 0 failed
- targeted set (`test_raster_replace_1221`, `test_ingest_vector_failure_tails`, `test_event_ingest`, `test_raster_ingest`, `test_vrt_ingest_tasks`, `test_regenerate_vrt_integration`, `test_vrt_creation_173`, `test_job_cancel_vrt`, `test_vrt_stale_sweep_gap002`, `test_vrt_lifecycle_188`, `test_service_token_redaction_1277`, `test_reupload_completion_contract_1207`): 311 passed, 0 failed
- `uv run pytest tests/test_raster_replace_1221.py -k "PublishCommitLandedProbe or AckLostCommit" -q`: 15 passed
- `uv run pytest tests/test_ingest_vector_failure_tails.py -q`: 6 passed
- `uv run pytest tests/test_layering.py -q`: 47 passed, 0 failed
- `uv run ruff check .` and `uv run ruff format --check .`: clean
- `PYTHONPATH=. uv run python scripts/dump_openapi.py --check`: exit 0, no drift

Counterfactuals, each run by restoring the pre-fix code and re-running the new tests:

- `tasks_vector.py` restored from before the first commit: the local-original test fails on the deleted upload, the notification test fails with "the operator was told nothing", and the structural gate fails. 3 failed, 1 passed.
- `tasks_raster.py` and `tasks_vrt.py` restored: the first-ingest, VRT-regeneration and VRT-creation ack-loss tests fail on a deleted artifact, and the structural gate fails. 4 failed, 3 passed.
- `tasks_raster_replace.py` restored: both parametrized replace tests fail, on the connection-loss and the cancellation shape. 2 failed.
- Round 1, stand-down removed: the VRT regeneration ack-loss test fails, because the task raises where it must now return.
- Round 1, handler guard removed: `test_a_post_publish_followup_failure_writes_no_failure_row` fails.
- Round 1, generation fence removed: `test_the_generation_failure_write_is_fenced_at_the_statement` fails.
- Round 2, `tasks_common.py` restored to the drop-first order: `test_the_job_still_records_the_original_failure` fails with `InFailedSQLTransactionError: current transaction is aborted`, and the order gate fails. 2 failed. Reordering alone was not enough to reproduce it, which is why both halves shipped: with the rollback still in place the job recovered, so the behavioural test pins the outcome and the AST gate pins the order.
- Round 2, both stand-downs returned to a bare `return`: the two replace ack-loss tests and the VRT regeneration test fail on a superseded object that survived the swap. 3 failed.

A permanent counterfactual ships with the suite: `test_a_probe_that_observes_no_commit_still_reaps` drives the same ack-loss path with the probe answering "not committed" and asserts the keys are reaped, so GAP-017's orphan cleanup is still pinned in the direction the fix must not change.

The ack-loss shape cannot be produced through a real session on demand, so the tests patch `AsyncSession.commit` to run the real commit, ask a fresh session whether the job became `complete`, and raise if it did. That fires on exactly one commit per task, the publishing one, without the test having to count commits or know which phase block it lives in. The aborted-transaction shape is produced the same way, at the server: a patched `_qtable` emits malformed DDL, because a double raising in Python would leave the session healthy and the reordering counterfactual would then pass for the wrong reason.

## Module size caps

Re-measured with `wc -l` and set exact, each with a comment saying what the lines bought:

- `tasks_common.py` 2380 to 2426
- `tasks_vrt.py` 1354 to 1480
- `tasks_vector.py` 1143 to 1161

`tasks_raster.py` (838), `tasks_raster_common.py` (494), `tasks_raster_replace.py` (854) and `tasks_raster_swap.py` (680) are under the 1000-line inclusion threshold and stay ungated.

## Noticed and left alone

- Codex round 1 raised the `VrtGeneration` overwrite that this PR had listed here as a history-row inaccuracy. It has readers, so it is fixed above rather than left; the entry is gone from this list.

- `tasks_reupload.py` carries the fourth copy of the validation exit and still unlinks at line 274, and it is where the audit's fourth excerpt row sits (orphaned attempt-scoped staging tables, and the stale "drop stale staging table from any prior failed attempt" comment at 287 to 292 that can only ever name its own attempt's table). Both verified still present. The file belongs to lane B2b on #1770, so nothing here touches it.
- `ingest_service` revalidates its URL for SSRF before `resolve_ingest_attempt_or_skip`, so that refusal escapes the task before any handler runs and leaves the job pending for the stale sweep. Unrelated to the failure-write copies, untouched.
- The audit recommends extracting the validation exit into a shared `validate_or_fail_attempt` helper. Not done: the defect is the unlink, the fourth call site is in a file this lane must not touch, and an extraction that covers only three of four copies leaves the same drift it was meant to end.
- Routing `ingest_file`'s failure through the shared helper means a superseded attempt's exception now logs nothing, because the helper returns before its log when the attempt fence matches no row. That matches `reupload_file` exactly and is the correct reading: an attempt that no longer owns the job is not reporting that job's outcome.

### DIFF
--- a/backend/app/processing/ingest/tasks_common.py
+++ b/backend/app/processing/ingest/tasks_common.py
@@ -1146,7 +1146,7 @@ async def _cleanup_staging_on_failure(
     task_name: str,
     attempt_id: uuid.UUID | None = None,
 ) -> None:
-    """Roll back, drop staging table, and mark job as failed.
+    """Mark the job failed, then drop the staging table, in that order.
 
     Shared by ``reupload_file`` and ``reupload_service`` which have
     structurally identical exception handlers, and — fix(#1778) — by the
@@ -1164,6 +1164,15 @@ async def _cleanup_staging_on_failure(
     name skips the DROP outright, because interpolating it raises inside the
     best-effort guard below, which would log a cleanup failure on every VRT
     build failure and say nothing true.
+
+    fix(#1778 codex r2): the ORDER is the contract. The failure row is
+    written and committed BEFORE the drop is attempted, because a statement
+    error aborts the whole PostgreSQL transaction and every later statement
+    on that session raises until it is rolled back. With the drop first, a
+    lock or statement timeout on it took the failure write down with it and
+    the job sat `running` with no reason recorded. Anything added here that
+    can fail belongs after the commit, in its own guarded block, with a
+    rollback of its own wreckage.
     """
     from sqlalchemy import text
     from sqlalchemy import update as sa_update
@@ -1182,21 +1191,6 @@ async def _cleanup_staging_on_failure(
     # that dispatch on its type or re-raise it are unaffected.
     error_message = redact_url_credentials(str(exc))
     await session.rollback()
-    try:
-        if staging_table:
-            await session.execute(
-                text(
-                    f"DROP TABLE IF EXISTS {_qtable(staging_table, schema=_current_tenant_schema())}"
-                )
-            )
-            await session.commit()
-    except Exception as cleanup_exc:  # broad: cleanup is best-effort after rollback; DB may be in bad state
-        structlog.get_logger().warning(
-            f"Staging-table cleanup failed during {task_name} failure",
-            staging_table=staging_table,
-            cleanup_error=str(cleanup_exc),
-            original_error=str(exc),
-        )
 
     failure_update = sa_update(type(job)).where(type(job).id == job_id)
     if attempt_id is not None:
@@ -1218,6 +1212,43 @@ async def _cleanup_staging_on_failure(
         )
     )
     await session.commit()
+
+    # fix(#1778 codex r2): the DROP runs AFTER the failure row is committed,
+    # not before it. PostgreSQL aborts the whole transaction on any statement
+    # error, so a drop that hit a lock or statement timeout left this session
+    # unusable and the failure UPDATE that followed it raised
+    # `current transaction is aborted` — the job stayed `running` until the
+    # stale sweep and the reason nobody recorded was the one the user needed.
+    # A best-effort cleanup must never be able to swallow the write it
+    # precedes, so it goes last and rolls back its own wreckage.
+    #
+    # Placed before the rowcount return so this attempt's table is dropped
+    # even when a newer attempt already owns the job row: the name is
+    # attempt-scoped, so it is ours to clean up either way.
+    if staging_table:
+        try:
+            await session.execute(
+                text(
+                    f"DROP TABLE IF EXISTS {_qtable(staging_table, schema=_current_tenant_schema())}"
+                )
+            )
+            await session.commit()
+        except Exception as cleanup_exc:  # broad: best-effort cleanup
+            structlog.get_logger().warning(
+                f"Staging-table cleanup failed during {task_name} failure",
+                staging_table=staging_table,
+                cleanup_error=str(cleanup_exc),
+                original_error=str(exc),
+            )
+            try:
+                await session.rollback()
+            except Exception:  # broad: a dead connection cannot be rolled back
+                structlog.get_logger().warning(
+                    "staging_cleanup_rollback_failed",
+                    staging_table=staging_table,
+                    task=task_name,
+                )
+
     if attempt_id is not None and not result.rowcount:
         return
     job.status = "failed"
--- a/backend/app/processing/ingest/tasks_common.py
+++ b/backend/app/processing/ingest/tasks_common.py
@@ -1149,7 +1149,21 @@ async def _cleanup_staging_on_failure(
     """Roll back, drop staging table, and mark job as failed.
 
     Shared by ``reupload_file`` and ``reupload_service`` which have
-    structurally identical exception handlers.
+    structurally identical exception handlers, and — fix(#1778) — by the
+    import tasks that used to paste a narrower copy of the terminal write.
+    What the copies were missing is what makes this the one place to fail a
+    job: the
+    ``redact_url_credentials`` backstop on the persisted message, the
+    ``pending``-inclusive attempt fence (fix #1274 review: a worker-time
+    refusal that raises before the claim must still finalize the job it owns,
+    rather than leave it pending until the stale sweep), and the
+    ``ingest_failed`` notification an operator has switched on.
+
+    fix(#1778): ``staging_table`` is "" for the paths that have none (the VRT
+    tail, whose artifacts are object keys its own ``finally`` reaps). An empty
+    name skips the DROP outright, because interpolating it raises inside the
+    best-effort guard below, which would log a cleanup failure on every VRT
+    build failure and say nothing true.
     """
     from sqlalchemy import text
     from sqlalchemy import update as sa_update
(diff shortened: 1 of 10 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

PR #1785, opened 2026-09-02
fix(perf): close the anonymous request-cost ladder (audit 2026-08-30)

ishiland:
Fixes four findings from the codebase audit 2026-08-30 (8dc529f17), tracked in #1778, that share one cause: the ladder of things bounding what an anonymous request can cost has a rung missing at every level. All four were re-verified as still present on current main before any code was written. None had been fixed since the audit.

## The "Global Rate Limit" is keyed per URL PATH, not per IP

slowapi's `Limiter` defaults to `key_style="url"`. `_check_request_limit` sets `_endpoint_key = request["path"]` and `__evaluate_limits` counts under `(key_func(request), _endpoint_key)`, so the admin-editable "Global Rate Limit (per second)" counted under (client IP, URL). On a path-parameterised route the caller picks the URL, so the caller picked how many budgets to have: one per dataset id, one per z/x/y, one per record id. `openapi.json` lists 46 anonymous parameterised operations, and the expensive ones are among them.

The limiter is now built with `key_style="endpoint"`, which counts under (client IP, `module.function`). The route table decides how many handlers exist, so the multiplier is fixed at deploy time.

Two things move with it, both recorded at the limiter rather than papered over. The trailing-slash aliases of a dual-shape route now share one bucket, which is stricter and is what the T-1238-03 guard in `test_admin_rate_limit.py` was always asking for; its docstring is corrected in the same commit. Two methods on one path no longer share theirs, which is looser by a number the route table fixes, unlike the number of URLs. Requests matching no route are unaffected either way, because `slowapi/middleware.py::_should_exempt` already exempts a request whose handler it cannot resolve, before key style is consulted.

I walked every `@limiter.limit` and `@limiter.exempt` site. No comment anywhere claimed URL keying except that one test docstring; every other one describes a per-IP or per-user limit, which endpoint keying preserves. `@limiter.exempt` is matched by function name in both styles and is unaffected.

### The ArcGIS sign-in limits (#1758)

Rebased onto main past #1758, which added the two per-route limits on `POST /services/arcgis/signin/`. They are worth calling out because the key style governs per-route limits too: a `@limiter.limit` decorator carries its own `key_func` but no `scope`, so `__evaluate_limits` falls back to `lim.scope or endpoint` and counts under `(key_func(request), _endpoint_key)`. The key functions are the first half; the key style decides the second.

They were not already handler-keyed, and nothing in them assumed URL keying either. `_signin_user_key` and `_signin_portal_key` (`sources/router.py:1211-1224`) both read the user id off `request.state` and never touch the path, and `signin_guard.py` counts committed `audit_logs` and `arcgis_signin_attempts` rows keyed on user id, account key and portal host, with no path or slowapi input anywhere in it. So no scope derivation needed changing.

What did change is that control 2 of the five controls now admits the number its own comment states. The route is dual-shape, so under URL keying `/signin` and `/signin/` were two buckets, and a caller alternating them put six requests per worker onto the database-backed control 4 rather than three. Measured with control 4 and the per-portal limit raised out of the way: six 200s under `key_style="url"`, and 200/200/200/429/429/429 under `key_style="endpoint"` with only three sign-in POSTs reaching the portal. A `fix(#1778)` note at the control list records it, and the regression test lives beside the other key-style tests rather than in the sign-in suite, to keep the hunk out of the lane working in `sources/**`.

`tests/test_arcgis_signin.py`, `tests/test_arcgis_signin_tenancy.py`, `tests/test_sources_safe_client_structural.py` and `tests/test_rule1_structural.py` all pass under endpoint keying (189 passed). No existing test hits two spellings expecting two buckets; every sign-in test uses `SIGNIN_URL = "/services/arcgis/signin/"`.

## /api/tiles/raster-proxy/ bypasses the only rate limit in the deployment

nginx declares one `limit_req_zone` and applied it in one place: the `/raster-tiles/` location, which rewrites to `/tiles/raster-proxy/` and proxies that to the api. The same handler answers `/api/tiles/raster-proxy/...` through `location /api/`, which carries no `limit_req`, `proxy_cache off`, and an `@limiter.exempt` handler behind it. The rate limit was opt-in by URL spelling, and only the product's own clients opted in.

A nested location under `/api/` now carries the same zone and burst, so both spellings draw on one per-IP bucket.

I picked that over dropping the handler's `@limiter.exempt` because it keeps cached tile serving unchanged and keeps legitimate map loads working. The app's global default is a general-purpose per-second cap; the nginx zone is sized for tiles (600r/m with burst 200, for the 50-100+ request cold start of a tilted multi-layer 3D view). Putting the app default in front of the handler would refuse the tail of exactly that fan-out, including the part nginx forwards on a `raster_cache` MISS. Cached serving is untouched under either choice, since a cache hit never reaches the api and the `/api/` block already runs with `proxy_cache off`.

Measured with the file rendered against nginx 1.29 (alpine), the upstream pointed at a closed port so a proxied request answers 502 and a refused one answers 503, and the zone temporarily narrowed to 1r/m with no burst:

- without the nested block: `/api/tiles/raster-proxy/...` answered 502, 502, 502, and a following `/raster-tiles/...` answered 502 as well. Not limited, and not even drawing on the other spelling's bucket.
- with it: 502 then 503; `/api/health` answered 502 twice, so the outer block stays unlimited; `/raster-tiles/...` answered 503 on its first request, because the tokens had already been spent under the other spelling.

`nginx -t` passes on the rendered file, and a 502 from the nested block carries a header set identical to one from `/api/health`, so the inherited security and proxy headers still apply.

## cols= is spliced into the tile cache key unvalidated

`parse_cols_param` returned the caller's string as the cache-key segment and deferred validation to `_select_tile_columns`, which drops an unrecognised name silently. So the key varied on input the projection ignores, and it did so in three ways that compound:

- an unknown name, dropped by the selector, leaving the unfiltered tile under a fresh key;
- **any** valid subset at or above the z<10 attribute budget, where the zoom default already projects every column. On a wide public table that is exponentially many keys for one set of bytes, not one bogus name at a time. A name already on an explicit `tile_columns` allowlist is the same case at any zoom;
- a name no query builder emits: `gid`/`geom`/`geom_4326` on either route, and the cluster query's own `point_count`/`cluster_id` family on the cluster route.

Each cost a full `ST_AsMVT` on an anonymous, `@limiter.exempt` route serving public datasets, and then wrote the result. The default production stack has no Valkey, so the writes land in an `LRUCache(maxsize=50_000)` and evict legitimate tiles.

The function now takes `z`, the allowlist and the route mode, resolves the projection through `_select_tile_columns`, mirrors each query builder's exclusion set, and keys on what the request **adds** to the projection the tile would have had anyway. `additional_columns` only ever unions into the base selection, so that difference identifies the projection uniquely for one dataset at one zoom, and an empty suffix keeps meaning what it always meant, so no existing cache entry changes shape. `z` is positional and required so a third tile endpoint cannot quietly go back to keying on the request.

Measured: at z=12, no cols, one column, two columns and an unknown name all read and write one entry. At z=2 two distinct subsets stay apart, `name,gid` keys as `name`, and `gid` alone keys as no cols. A name inside a `tile_columns` allowlist keys as no cols at both zooms while one outside it still keys separately. `point_count` collapses on the cluster route and stays distinct on the vector route, which is why `mode` reaches the key.

## The ?v= raster cache-buster pins a pool connection across the Titiler round trip

`_resolve_raster_meta` looks up under the request's `v` (#1329), and `_meta_cache_version_segment` accepts any ASCII digit run of length 1 to 10, so a caller varying it misses that cache on every request and pays the `db.execute`. `get_db` yields one session for the whole request and only rolls back on an exception, so that read's transaction was held until the response was written, while `raster_tile_proxy` awaited Titiler for up to three attempts at a 30s timeout plus 0.5s and 1.0s of backoff. The pool is `db_pool_size + db_max_overflow` per uvicorn worker; once it is parked on upstream fetches, every other request in that worker waits out `db_pool_timeout` and then errors.

`_resolve_raster_access` now rolls back before returning, the remedy fix(#1451) applied to the vector path for the same reason. It sits in the resolver rather than at either call site because a caller cannot release what it did not know was taken, and it covers `raster_auth_check` (a mounted route) and `raster_tile_proxy` (its in-process caller) without either having to remember. Everything past it is locals: `meta` is a plain snapshot and `storage_backend` a string, and the embed-token usage bump uses its own side session (fix(#1436)), so the rollback discards nothing. The #1329 note that priced a cache-busting `v` as "at most one extra indexed read" is corrected in place, since the read was never the expensive half.

No schema change. No request or response schema change: `dump_openapi.py --check` passes with `backend/openapi.json` untouched, so no SDK regeneration is needed. I did not add `Query(max_length=...)` to `cols`, which the finding itself calls secondary, because it would change the published operation description and parameter schema and churn every generated SDK for a bound the validation supersedes. For the same reason the handler docstring is unchanged and the note about the key lives at `parse_cols_param` and in an inline comment at the call site.

Config impact is `frontend/nginx.conf` only: one nested location, no new envsubst placeholder, no change to the zone, burst, cache or headers.

Operators should know the rate-limit change is a real tightening. An integration that today gets 60 requests a second per dataset id from one IP will get 60 a second across all of them. The setting is admin-editable at runtime, so raising it needs no deploy.

Module cap: `backend/app/processing/tiles/router.py` 2590 to 2639, re-measured after each change and still exact. `tiles/service.py` is 870, under the 1000 inclusion threshold, so it stays ungated. `sources/router.py` is 1373, under the 1500 default router cap and not in the allowlist, so the nine comment lines need no cap change.

## Verification

Run from this worktree against Postgres at localhost:5434.

- `uv run pytest tests/test_layering.py -q` -> 47 passed
- `uv run pytest tests/test_rate_limit_key_style_1778.py tests/test_nginx_raster_proxy_ratelimit_1778.py tests/test_tile_cols_cache_key_validation_1778.py tests/test_raster_proxy_releases_connection_1778.py tests/test_layering.py -q` -> 70 passed (post-rebase)
- after the projection-key change: the four new files plus `test_layering.py` and seven tile suites -> 152 passed, 3 skipped
- `uv run pytest tests/test_tile_cols_cache_key_validation_1778.py tests/test_tile_cols_endpoint.py tests/test_tile_cache_cols_key.py tests/test_tile_column_allowlist.py tests/test_tiles.py tests/test_tile_cache.py tests/test_tile_cache_scope_sec009.py tests/test_tile_cache_generation_key_1429.py tests/test_tile_cached_authz_liveness_1451.py -q` -> 156 passed
- `uv run pytest tests/test_arcgis_signin.py tests/test_arcgis_signin_tenancy.py tests/test_sources_safe_client_structural.py tests/test_rule1_structural.py -q` -> 189 passed (post-rebase)
- `uv run pytest tests/test_arcgis_signin.py tests/test_arcgis_signin_tenancy.py tests/test_sources_safe_client_structural.py tests/test_rule1_structural.py tests/test_rate_limits.py tests/test_admin_rate_limit.py tests/test_middleware.py -q` -> 215 passed (post-rebase, random order)
- `uv run pytest tests/test_layering.py tests/test_rate_limit_key_style_1778.py tests/test_nginx_raster_proxy_ratelimit_1778.py tests/test_tile_cols_cache_key_validation_1778.py tests/test_raster_proxy_releases_connection_1778.py tests/test_rate_limits.py tests/test_admin_rate_limit.py tests/test_middleware.py tests/test_tiles.py tests/test_tile_cols_endpoint.py tests/test_tile_cache_cols_key.py tests/test_raster_tiles.py tests/test_raster_colormap_proxy.py -q` -> 226 passed
- `uv run pytest tests/test_tiles.py tests/test_tile_cols_endpoint.py tests/test_tile_cache_cols_key.py tests/test_tile_column_allowlist.py tests/test_tile_cache.py tests/test_tile_cache_scope_sec009.py tests/test_tile_cache_generation_key_1429.py tests/test_tile_cached_authz_liveness_1451.py -q` -> 131 passed
- `uv run pytest tests/test_raster_tiles.py tests/test_raster_colormap_proxy.py tests/test_raster_tile_url_version_1372.py tests/test_vrt_catalog_175.py tests/test_titiler_url_helper.py tests/test_optional_auth_failure_mode_1518.py -q` -> 176 passed
- `uv run pytest tests/test_vector_tile_auth.py tests/test_tile_signing.py tests/test_seed_tiles.py tests/test_worker_tile_cache_1315.py tests/test_dataset_tile_columns_metadata.py tests/test_dataset_delete_tile_cache_purge.py tests/test_phase_274_tile_cache.py tests/test_tile_gzip_offload_perf005.py -q` -> 95 passed, 3 skipped
- `uv run pytest tests/test_auth.py tests/test_rule1_structural.py tests/test_rule2_structural.py -q` -> 142 passed
- `uv run pytest tests/test_api_contract_openapi.py tests/test_health.py tests/test_module_import_seams.py tests/test_register_verification_flow.py tests/test_sandbox_query_endpoint_565.py tests/test_search_cors_1596.py tests/test_data_serving_extension.py -q` -> 173 passed
- the six pre-existing `tests/test_nginx_*.py` suites -> 21 passed
- `uv run ruff check .` and `uv run ruff format --check .` -> clean
- `PYTHONPATH=. uv run python scripts/dump_openapi.py --check` -> exit 0

Each new test was run against the un-fixed code as a counterfactual. Keying the tile cache on the requested subset rather than the effective projection fails 13 of that file's 25 tests and leaves green the 12 that guard against over-collapsing. Removing the nested nginx block fails three of the four nginx assertions and leaves the control passing. Removing `await db.rollback()` fails all three raster assertions. Forcing `key_style = "url"` makes two different URLs on one handler both pass a 1/minute limit, and makes the alternating sign-in spellings return six 200s instead of three 200s and three 429s.

## Left alone

- Three neighbouring findings in the same audit register, out of this lane's scope: the `/raster-tiles/` location has no `proxy_read_timeout` so it runs on nginx's 60s default against a handler whose worst case is about 121s; `stretch=percentile&pmin=&pmax=` buys an anonymous caller a full-raster statistics computation against a 256-entry cache and a single Titiler worker; and no `limit_conn` exists at either layer, so a rate limit cannot bound in-flight anonymous work.
- `db/postgresql.conf` statement timeouts, excluded by the lane brief as operator-facing config.
- `Query(max_length=...)` on `cols`, for the SDK-churn reason above.
- The vector `/api/tiles/data.*` routes are still unthrottled at both layers. That belongs to the `limit_conn` finding rather than this one, and the nested nginx block deliberately covers only the raster proxy so a tile-shaped zone does not start governing MVT traffic it was not sized for.

### DIFF
--- a/backend/app/modules/catalog/sources/router.py
+++ b/backend/app/modules/catalog/sources/router.py
@@ -1135,6 +1135,15 @@ async def preview_service_layer(
 #    worker and `docker-compose.prod.yml:294` starts two, so on a stock
 #    install these three are three per worker. They are the cheap first layer
 #    that keeps a flood off the database.
+#
+#    fix(#1778): three per worker is now the real number. slowapi scopes a
+#    limit by the request PATH unless the limiter is built with
+#    key_style="endpoint", and this route is dual-shape, so before that a
+#    caller alternating `/signin` with `/signin/` drew on two buckets and put
+#    six requests per worker onto control 4 instead of three. Measured, with
+#    control 4 raised out of the way: six 200s under the old keying, three
+#    200s then three 429s under the new. The key functions below were never
+#    the leak; they carry the user id in both.
 # 3. A PostgreSQL advisory lock per (user, portal host), held across the
 #    mint, so the count below cannot be read by two workers at once.
 # 4. The same three attempts per fifteen minutes counted from the audit rows
--- a/backend/app/platform/ratelimit.py
+++ b/backend/app/platform/ratelimit.py
@@ -16,6 +16,25 @@ four other domains import this module.
 ``_global_rate_limit`` is a callable rather than a literal because the global
 limit is admin-editable at runtime; SlowAPI re-evaluates the default per
 request, so a settings change takes effect without a restart.
+
+fix(#1778): ``key_style="endpoint"`` keys the counter on (client IP, handler)
+instead of (client IP, request path), which is slowapi's default. Under the
+default every distinct URL got its own budget, so a path-parameterised route
+handed one IP a fresh "Global Rate Limit (per second)" allowance per dataset
+id and per z/x/y. The setting operators read as a cap on anonymous request
+cost bounded nothing the caller could not multiply for free by varying a path
+segment. Handlers are enumerated by the route table, so the remaining
+multiplier is fixed at deploy time rather than chosen by the caller.
+
+Two things move with it, both recorded rather than papered over. The
+trailing-slash aliases of a dual-shape route now share one bucket, which is
+strictly stricter and is what the guard in ``tests/test_admin_rate_limit.py``
+was always asking for. Two methods on one path no longer share theirs, which
+is looser by the number of methods the route table declares; that number is
+fixed, unlike the number of URLs. Requests matching no route are unaffected
+either way: slowapi's middleware exempts a request whose handler it cannot
+resolve before key style is ever consulted
+(``slowapi/middleware.py::_should_exempt``).
 """
 
 from fastapi import Request
@@ -29,4 +48,8 @@ def _global_rate_limit(_request: Request | None = None) -> str:
     return f"{get_cached_global_rate_limit()}/second"
 
 
-limiter = Limiter(key_func=get_remote_address, default_limits=[_global_rate_limit])
+limiter = Limiter(
+    key_func=get_remote_address,
+    default_limits=[_global_rate_limit],
+    key_style="endpoint",
+)
--- a/backend/app/processing/tiles/router.py
+++ b/backend/app/processing/tiles/router.py
@@ -2331,7 +2331,17 @@ async def cluster_tile_endpoint(
     # for.
     _ensure_clusterable_dataset(meta)
 
-    additional_columns, cols_cache_key = parse_cols_param(cols, meta.column_info)
+    # fix(#1778): keyed on the effective projection, so `z`, the allowlist and
+    # the mode all reach it. They decide what the request actually changes:
+    # the cluster query emits its own `point_count`/`cluster_id` names, so a
+    # `cols=` naming one of those changes nothing at any zoom.
+    additional_columns, cols_cache_key = parse_cols_param(
+        cols,
+        meta.column_info,
+        z,
+        tile_columns=meta.tile_columns,
+        mode="cluster",
+    )
 
     cache_ttl = meta.tile_cache_ttl or settings.tile_cache_ttl
 
@@ -2494,11 +2504,15 @@ async def tile_endpoint(
     # Get column info for attribute selection
     columns = meta.column_info
 
-    # fix(#1778): `columns` gates the cache key, not just the projection. The
-    # docstring above is the published operation description, so the mechanism
-    # is written down at `parse_cols_param` instead of churning every generated
-    # SDK to say it.
-    additional_columns, cols_cache_key = parse_cols_param(cols, columns)
+    # fix(#1778): the cache key comes from the EFFECTIVE projection, not from
+    # the request, so `z` and the allowlist have to reach it. At z >= 10 the
+    # zoom default already projects every column and every valid subset
+    # collapses onto one entry. The docstring above is the published operation
+    # description, so the mechanism is written down at `parse_cols_param`
+    # instead of churning every generated SDK to say it.
+    additional_columns, cols_cache_key = parse_cols_param(
+        cols, columns, z, tile_columns=meta.tile_columns
+    )
 
     # Use per-dataset cache TTL when set, else global default
     cache_ttl = meta.tile_cache_ttl or settings.tile_cache_ttl
--- a/backend/app/processing/tiles/router.py
+++ b/backend/app/processing/tiles/router.py
@@ -754,9 +754,17 @@ async def _resolve_raster_meta(
     #
     # The lookup is under the request's value, but the STORE is under the row's
     # (see the write site below) — the request never names the entry it writes.
-    # Memory stays bounded by the LRU, and a caller varying `v` to evict entries
-    # costs at most one extra indexed read per request — the same order as the
-    # uncached miss an unknown dataset id already produces.
+    # Memory stays bounded by the LRU, and a caller varying `v` costs one extra
+    # indexed read per request in the database.
+    #
+    # fix(#1778): that read is not the whole price, and this note used to say it
+    # was ("at most one extra indexed read per request, the same order as the
+    # uncached miss an unknown dataset id already produces"). The read opens a
+    # transaction, and `get_db` holds the connection it opened until the
+    # response is written, so the real cost of a miss on the raster path was an
+    # API-pool connection pinned across the caller's whole Titiler round trip.
+    # `_resolve_raster_access` now releases it before returning; the sentence
+    # above is true again because that call site makes it true, not on its own.
     version_segment = _meta_cache_version_segment(requested_version)
     cache_key = (
         f"{base_key}:v{version_segment}" if version_segment is not None else base_key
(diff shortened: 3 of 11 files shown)


ishiland:
@codex review

ishiland:
@codex review

Issue #1775, opened 2026-09-02
Reserve the ArcGIS sign-in attempt before the credential POST, and hold no pooled connection across network I/O


ishiland:
## What

Restructure the ArcGIS sign-in handler (`POST /api/services/arcgis/signin/`) so that a counted attempt is durably reserved *before* the credential POST, and no pooled database connection is held across external network I/O.

This is the follow-up to two findings raised in codex review of #1758 and deliberately deferred from that PR, which was capped at round 17. Both are **bounded, not open** in the shipped code; see Residual below.

## Why

Two defects share one fix.

**Pool saturation.** The request `AsyncSession` is checked out while `require_permission("create_layers")` resolves, and its transaction stays checked out across discovery, both advisory locks and the external mint, for up to the 45-second network budget. Production runs 10 pooled connections plus 3 overflow, so concurrent sign-ins for distinct scopes could occupy the whole pool and time out unrelated API requests.

**Lost attempt on cancellation.** If Uvicorn cancels the task during shutdown after the credential POST has begun, `CancelledError` bypasses both `PortalSignIn.mint()`'s `except Exception` and the handler's `except ArcGISSignInError`, so `_signin_audit()` never runs and session teardown rolls back without a ledger row. ArcGIS may count the failed password while GeoLens does not, so cancellation followed by retries can exceed the account budget. The per-phase timeout fix from round 11 only covers cancellations generated by those timeout contexts, not external task cancellation.

## Shape

Reserve, then settle:

1. **No session held.** Discovery runs first and resolves the token-service scope. It is credential-free and uncounted, so a cancellation here costs nothing.
2. **Short transaction.** Take both advisory locks, read both budgets, `INSERT` the ledger attempt row as a counted failure (the reservation), `COMMIT`. The connection returns to the pool.
3. **No session held.** Run the mint.
4. **Short transaction.** Settle: write the audit row with the real outcome. The reservation already counts, so success needs no ledger change and failure needs none either. On `CancelledError`, do nothing and re-raise — the reservation is already durable and conservative.

Nothing then holds a pooled connection across the network budget, and a cancelled request is already counted.

### Why the reservation cannot simply precede discovery

[...]

PR #1786, opened 2026-09-02
fix(#1758): two flaky assertions in the sign-in suite

ishiland:
Fixes the flaky `test_a_signin_holds_exactly_one_pooled_connection`, which went red in merge-queue run 33648625488 on an unrelated frontend-only PR. Test-only; no production code changes.

## What was wrong

The test summed the pool's `checkout` event and asserted 1. That conflates two different things:

- how many connections the request held **at once**, which is the property the test exists to protect
- how many times it **acquired** one, which is what `checkout` counts

Those differ whenever a connection is released and re-acquired in sequence, and that is legitimate on this path. `_signin_audit` commits, which returns the connection to the pool; its rollback-and-retry on a poisoned transaction then acquires another; and the test harness has contention retries of its own in `_RetryingAsyncEngine` and `_acquire_test_session_with_retry`, both added precisely because acquisition can fail transiently under `-n 4`.

I measured it rather than reasoning about it. A single session that commits and then runs one more statement scores **2 cumulative checkouts with a peak of 1 held**:

```
CUMULATIVE=2 PEAK_CONCURRENT=1
```

So the assertion could go red on a loaded runner with nothing whatsoever wrong in the code under test. The failing run shows no connection errors and the test failed at 100% completion after a twenty-minute suite, which fits a contention-driven re-acquire rather than a defect.

## What it does now

Tracks `checkout` against `checkin` and asserts the **peak concurrent** count is 1. That is the property the test's own docstring names: both advisory locks must ride the request's own session, and the defect it guards against is a dedicated lock session holding its connection *while* the request holds its own.

This is a sharper assertion, not a weaker one. Sequential re-acquisition cannot reach a peak of 2, so the only way to trip it is the real defect. Verified by counterfactual: reinstating a dedicated lock session in `_signin_locks` still turns it red, now with `a sign-in held 2 pooled connections at once`.

## Verification

- The single test 10 times under `-n 4`: 10/10 pass.
- `tests/test_arcgis_signin.py` under `-n 4`: 175 passed.
- Same file in default (random) order: 175 passed.
- With the tenancy and layering suites under `-n 4`: 225 passed.
- Counterfactual: a second session in `_signin_locks` reddens it with the new message.
- `ruff check` clean.

## Note

#1775's reserve-then-settle redesign will deliberately introduce a second *sequential* connection acquisition per sign-in, in two short transactions either side of the network phases. Under the old cumulative assertion that redesign would have looked like a regression; under peak-concurrency it reads correctly, since the two transactions never overlap. The test stays valid as one of that issue's acceptance criteria.

---

# Second commit: the discovery-failure lock assertion

`43844a2` fixes a second flake in the same file, `test_a_discovery_failure_takes_no_lock_and_writes_no_ledger_row`, red in merge-queue runs 33649640369 and 33649761185 on two more frontend-only PRs. Also test-only.

## What was wrong

The test proved "no lock was taken" by monkeypatching `app.core.db.async_session` and asserting it was never called. That proxy made sense when the sign-in opened a dedicated session for its advisory locks. Round 10 moved both locks onto the request's own transaction, and from that point the assertion observed nothing about locking at all.

What it did observe was a **process-global** factory that plenty of unrelated code calls, `get_db` among them, plus the ingest and embedding task paths. Anything in the worker that opened a session inside the patch window scored as a lock this sign-in took.

Measured, not assumed. With the sign-in taking no lock whatsoever, one unrelated `async_session()` call inside the same window scores:

```
OLD PROXY COUNTED: 1  (sign-in took no lock at all)
```

## What it does now

Wraps `_signin_locks`, the helper the handler actually calls, and asserts it was never entered. Same sentence, strictly more precise: nothing outside this request can reach that helper.

Verified by counterfactual. With discovery no longer short-circuiting, the handler reaches the lock and the test fails.

## Sweep of the rest of the file

Checked for other assertions reading state wider than their own test:

- **No `pg_locks` or `pg_stat_activity` read anywhere in the file.** The advisory-lock tests observe locks by trying to take them on their own session, not by querying the cluster.
- **Ledger reads are all scoped.** Two filter on `account_key`; three are before/after deltas within a single test.
- **Per-worker databases** (`geolens_test_gw0_<uuid>`) isolate both table rows and advisory locks, so cross-worker contamination is not reachable for either.
- **The pool listener in the first commit** does read a process-global engine, so I measured it rather than trusting it. During a sign-in exactly one checkout occurs, from `get_db`. `peak == 1` holds with nothing else competing.

The one other counter in the file patches `_signin_audit`, which only the sign-in path calls.

## Verification

- Target test alone: passes.
- `test_arcgis_signin.py` + `test_arcgis_signin_tenancy.py` under `-n 4`, 5 consecutive runs: 178 passed each time.
- Adding `test_layering.py`: 225 passed.
- Counterfactual on the new assertion: red.
- `ruff check` and `ruff format --check` clean; pre-commit hooks pass.

### DIFF
--- a/backend/tests/test_arcgis_signin.py
+++ b/backend/tests/test_arcgis_signin.py
@@ -2121,18 +2121,42 @@ async def test_a_signin_holds_exactly_one_pooled_connection(
     were holding, stalling unrelated traffic until the 30-second pool timeout.
     Both locks now ride the request's own transaction, so the count is one.
 
-    Counted with a pool `checkout` listener rather than at the session factory,
-    because the factory is not what the pool runs out of.
+    Counted at the pool rather than at the session factory, because the
+    factory is not what the pool runs out of.
+
+    fix(#1758): counted as PEAK CONCURRENT connections, not as cumulative
+    checkout events. The first version summed `checkout` and asserted 1, which
+    conflated "how many connections did this request hold at once" with "how
+    many times did it acquire one". Those differ whenever a connection is
+    released and re-acquired in sequence, which is legitimate and does happen
+    here: `_signin_audit` commits, its rollback-and-retry path on a poisoned
+    transaction acquires again, and the harness has contention retries of its
+    own in `_RetryingAsyncEngine` and `_acquire_test_session_with_retry`. A
+    single session that commits and then runs one more statement already
+    scores 2 cumulative while never holding more than one. That made the
+    assertion red on a loaded runner with nothing wrong, first seen in
+    merge-queue run 33648625488 on an unrelated frontend-only PR.
+
+    Peak concurrency is the property the paragraph above actually names, and
+    it is what a second lock session would break: that session held its
+    connection WHILE the request held its own. Sequential re-acquisition
+    cannot reach 2, so this is a sharper assertion rather than a weaker one.
     """
     from sqlalchemy import event
 
     import app.core.db as db_module
 
-    checkouts = 0
+    live = 0
+    peak = 0
 
-    def _count(*_args) -> None:
-        nonlocal checkouts
-        checkouts += 1
+    def _acquired(*_args) -> None:
+        nonlocal live, peak
+        live += 1
+        peak = max(peak, live)
+
+    def _released(*_args) -> None:
+        nonlocal live
+        live = max(0, live - 1)
 
     exchange = _Exchange(
         {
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review

ishiland:
@codex review

PR #1790, opened 2026-09-02
fix(import,settings,a11y): frontend tail items from the 2026-08-30 audit

ishiland:
Fixes four frontend tail items from the codebase audit 2026-08-30 (8dc529f17), tracked in #1778.

## Changes

"handleCommitSingle has no in-flight guard, so a double click commits twice." The commit button only disables once the entry's status has re-rendered as 'committing', so a click landing before that render reached the handler with no guard at all. Added the same guard shape used for the layer-picker bulk guard (`entry.status === 'committing' || entry.status === 'tracking'`) and a vitest pinning it. This is distinct from that layer-picker guard and from the in-flight upload session added earlier.

"updateBranding declares privacy_url in its request type and silently drops it." `updateBranding` accepts `Partial<BrandingConfig>`, so a `privacy_url`-only call typechecks, but the function only ever forwarded `show_badge`. A `privacy_url`-only call issued `PUT /settings/` with an empty settings object and still reported success. Now forwards every field `BrandingConfig` declares. Added a vitest pinning the request body for `privacy_url` alone and for both fields together. The separate, more severe finding in the same register entry, that `show_badge` itself was sent under the wrong key (`branding_show_badge` instead of the registry's `branding.show_badge`), landed in #1776 while this branch was in flight; this branch is rebased onto that fix and the new test asserts the dotted key.

"wcagTags omits wcag22aa, the 22px builder row controls would fail it." Did not add `wcag22aa` to the gating suite's tags, since it would fail immediately on the builder's 22px row controls. Instead added `e2e/accessibility-target-size.spec.ts`, a separate, non-gating scan of the map builder route with `wcag22aa` that only reports violations (console output plus a test attachment) and never fails. It is not listed in any `package.json` `e2e:smoke:*` script, so the per-PR smoke gates never pick it up (same precedent as `analysis.spec.ts`). A run against the current build found one violation to scope later: the basemap drag handle button fails axe's `target-size` rule.

"playwright.builder-hardening.config.ts and e2e/record-detail-ux-audit.spec.ts are light-mode only." Both configs set no `colorScheme`, so the dark palette was structurally unreachable to them, the same gap already closed for `accessibility.spec.ts` and `playwright.config.ts`. `record-detail-ux-audit.spec.ts`'s tests now loop over both color schemes, the same shape used for `accessibility.spec.ts`. `colorScheme` is folded into the per-run evidence file names (screenshots, console/network/axe/notes logs) so a dark pass no longer overwrites the light-mode files for the same record type and viewport. For `playwright.builder-hardening.config.ts`, added one `chromium-hardening-dark` project rather than doubling every browser project: `firefox-hardening` and `webkit-hardening` exist for cross-browser rendering parity, not theme parity, so tripling wall-clock there would buy nothing. `ci.yml`'s builder-hardening step now selects both `chromium-hardening` and `chromium-hardening-dark` in one invocation (still chromium only, no extra browser install), so this step's wall-clock roughly doubles rather than the whole hardening suite tripling.

A codex review on this PR found two gaps in the new dark project, both fixed in a follow-up commit: `builder-hardening.spec.ts`'s public-share test opens its own unauthenticated context with `browser.newContext(...)`, which does not inherit the project's `use.colorScheme` (that option only applies to the default context/page fixtures a test gets automatically), so the dark project was running that one flow in light mode regardless; and the CI step only ever passed `--project=chromium-hardening`, so `chromium-hardening-dark` was defined but never actually selected in CI. Fixed both: the manual context now reads `testInfo.project.use.colorScheme` and passes it through, with a `matchMedia('(prefers-color-scheme: dark)').matches` assertion pinning that it resolves correctly per project; and the CI step now selects both projects.

None. Frontend-only, no schema, API contract, or backend config changes.

## Left alone

The wider settings-key-drift finding in the same register excerpt as the branding key mismatch (fixed by #1776, this branch is rebased onto it) is out of this lane's scope.

A pre-existing failure surfaced while verifying the dark-mode loop: the mobile vector-dataset audit in `record-detail-ux-audit.spec.ts` already fails a header-overlap check ("mobile H1 and action area overlap horizontally") in both light and dark mode identically. It reproduces the same way with or without this PR's colorScheme loop, so it is not a regression from this change. Flagging it here because this is apparently the first time anyone has actually run this spec against a real target: it requires a hand-built `e2e/fixtures/qa-targets.json` that does not ship in the repo.

## Verification

`cd frontend && npm run typecheck`: passed.
`cd frontend && npm run lint`: passed.
`cd frontend && npm run build`: passed.
`cd frontend && npx vitest run src/components/import/ src/api/__tests__/ src/hooks/__tests__/use-settings.test.ts src/components/admin/settings/`: 46 files, 392 tests passed.
`cd frontend && npm run test:i18n`: passed (no copy changed).

E2E via the `:5174` worktree recipe against main's API (`API_PROXY_TARGET=http://localhost:8001`, `E2E_ALLOW_WORKTREE=1 E2E_BASE_URL=http://localhost:5174`):

`npx playwright test e2e/accessibility-target-size.spec.ts --project=chromium`: 3 passed. Reports the one target-size violation noted above and does not fail on it.
`npx playwright test e2e/upload.spec.ts --project=chromium`: 4 passed. The single-commit path is unaffected by the new guard.
`npx playwright test -c playwright.builder-hardening.config.ts --project=chromium-hardening`: 8 passed.
`npx playwright test -c playwright.builder-hardening.config.ts --project=chromium-hardening-dark`: 8 passed on the new project.
`npx playwright test e2e/record-detail-ux-audit.spec.ts --project=chromium` against a hand-built one-target fixture: 4 passed, 2 failed (the pre-existing, colorScheme-independent header-overlap issue noted above), 12 skipped (no fixture for the other record types). Confirms the light/dark loop runs symmetrically and writes separate evidence files per scheme.

This branch was rebased onto main after #1776 merged (it also touches `frontend/src/api/settings.ts`), resolving the conflict by keeping #1776's dotted `branding.show_badge` key and adding the `privacy_url` forwarding on top of it. After the rebase, also ran the backend drift test that reads this file statically: `cd backend && uv run pytest tests/test_settings_key_drift.py tests/test_layering.py -q`, 49 passed.

After the codex-round-1 fixes above, re-ran `npx playwright test -c playwright.builder-hardening.config.ts --project=chromium-hardening --project=chromium-hardening-dark --reporter=line` via the same :5174 recipe: 14 passed (both projects, all 6 spec tests each plus setup/cleanup), confirming the matchMedia assertion resolves true under chromium-hardening-dark and false under chromium-hardening.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2005,12 +2005,18 @@ jobs:
 
       # fix(#524 B-036): the builder-hardening suite previously ran in NO CI
       # trigger (this job runs only the default config), so a red hardening
-      # gate on main was invisible. Chromium project only — the runner installs
-      # chromium alone, and the firefox/webkit projects are a local reviewer
-      # concern. --reporter=line keeps it from clobbering the main suite's
-      # playwright-report/ artifact.
-      - name: Run builder-hardening E2E suite (chromium)
-        run: npx playwright test -c playwright.builder-hardening.config.ts --project=chromium-hardening --reporter=line
+      # gate on main was invisible. Chromium projects only: the runner
+      # installs chromium alone, and the firefox/webkit projects are a local
+      # reviewer concern. --reporter=line keeps it from clobbering the main
+      # suite's playwright-report/ artifact.
+      #
+      # fix(#1778): chromium-hardening-dark (light-mode gap closed for this
+      # config) was defined but never selected here, so it never actually
+      # ran in CI. Both chromium projects in one invocation; still no
+      # firefox/webkit install needed since both are chromium under the
+      # hood, just a different colorScheme.
+      - name: Run builder-hardening E2E suite (chromium, light + dark)
+        run: npx playwright test -c playwright.builder-hardening.config.ts --project=chromium-hardening --project=chromium-hardening-dark --reporter=line
 
       # fix(#896): extract the failing spec names from the json reporter
       # output (playwright.config.ts) so the nightly report can say WHICH
--- a/e2e/accessibility-target-size.spec.ts
+++ b/e2e/accessibility-target-size.spec.ts
@@ -1,0 +1,94 @@
+import { test } from '@playwright/test';
+import AxeBuilder from '@axe-core/playwright';
+import { getAuthToken, seedDataset, deleteDataset, type SeededDataset } from './helpers/catalog';
+
+const BASE_URL = process.env.E2E_BASE_URL ?? 'http://localhost:8080';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function formatViolations(violations: any[]): string {
+  return violations
+    .map(
+      (v) =>
+        `[${v.id}] ${v.description} (${v.impact})\n` +
+        v.nodes.map((n) => `  - ${n.html}`).join('\n'),
+    )
+    .join('\n\n');
+}
+
+/**
+ * fix(#1778): the gating suite's `wcagTags` (accessibility.spec.ts) omits
+ * `wcag22aa`, the only tag that reaches axe-core's `target-size` rule, so
+ * the map builder's 22px row controls (StackRow.tsx, FolderGroupRow.tsx,
+ * BasemapGroupRow.tsx, UnifiedStackPanel.tsx, EmptyStackState.tsx,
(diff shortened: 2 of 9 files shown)


ishiland:
@codex review

ishiland:
@codex review

PR #1815, opened 2026-09-03
fix(cli): keep the retained refresh token with the bearer on backend fallback

### LINKED ISSUE #1807: cli: refresh without a new refresh token leaves the retained token in keyring when the bearer falls back to the file
Follow-up from codex round 32 on #1802 (review 5106633328, cli/geolens_cli/auth.py:1551), deferred under the merge-on-no-P1 policy.

When a refresh response omits a replacement refresh token and `store_bearer_token()` falls back from the keyring to the credentials file, the no-new-refresh branch moves only the fingerprint to the file and leaves the retained refresh token in the keyring. On the next refresh `_detect_credential_backend()` selects the keyring, a successful bearer write there is shadowed by the expired file-backed bearer, and subsequent commands refresh repeatedly instead of converging.

Fix: persist the retained refresh token and its fingerprint through `store_refresh_token()` using the bearer's actual backend, and extend the refresh pairing matrix in `cli/tests/test_credential_resolution_matrix.py` with a backend-fallback row.

Part of the 2026-08-30 audit tracked in #1778.

ishiland:
## What

`try_refresh()` in `cli/geolens_cli/auth.py` now persists a **retained** refresh token (the case where the server renews the access token but issues no replacement refresh token) through `store_refresh_token()` — value and pairing fingerprint together — into the backend the bearer **actually** landed in, and drops the superseded keyring copy once the file write has committed.

## Why

Round 31 rewrote only the pairing FINGERPRINT in that branch. That is enough while the bearer lands where `_detect_credential_backend()` said it would, and wrong the moment `store_bearer_token()` falls back from the keyring to `credentials.toml` on its own (a locked or account-specific `KeyringError`): the fingerprint moved to the file, the refresh token stayed in the keyring.

The next refresh then read the two halves from different backends. `_detect_credential_backend()` decides by looking for a `refresh_token` in the FILE, found none, answered "keyring", and wrote the newly rotated bearer there — where the stale file-backed bearer written by the earlier fallback outranks it under `load_bearer_token()`'s file-over-keyring precedence. Every later command 401s, refreshes, and 401s again instead of converging.

The keyring deletion reuses the keyring half of `_discard_unpaired_refresh_token()`, factored out as `_delete_keyring_refresh_entries()` so the file side — which now holds the only copy — is deliberately left alone. It keeps the same never-raises contract and is enumerated in `test_best_effort_helpers_never_raise.py`. `_rewrite_refresh_fingerprint()` had exactly one call site and is gone.

Invariants left intact: `_CREDENTIAL_SET` is unchanged, the three-state strict reads and the pairing check are untouched, and the new write still satisfies both structural gates in `test_refresh_token_backend_derivation.py` (`no_keyring=(backend != "keyring")` derived from the store call's real outcome, plus `bearer_token=` for pairing).

Fixes #1807.

Part of the 2026-08-30 audit tracked in #1778 (deferred from codex round 32 on #1802 under the merge-on-no-P1 policy).

## Tests

A backend-fallback row on the refresh pairing matrix (`cli/tests/test_credential_resolution_matrix.py`) — a dimension the existing table's axes cannot express, since every other row holds the backend fixed. It drives the whole sequence through the real CLI (`whoami` against a fake `/auth/me` + `/auth/refresh` pair): a bearer whose keyring write fails once, a refresh response with no replacement refresh token, then a second expiry with the keyring healthy again.

Counterfactual (both rows run against the unfixed `auth.py`):

- `test_retained_refresh_token_follows_the_bearer_into_the_file` — `KeyError: 'refresh_token'`; the retained token never reaches the file.
- `test_the_next_refresh_converges_instead_of_looping` — `assert 3 == 2`; the third command refreshes again, which is the loop the issue describes.

Both pass with the fix.

## Verification

```
cd cli && uv run --extra dev python -m pytest -q          # 647 passed (643 before, +4)
cd cli && uv run --extra dev ruff check geolens_cli tests/test_credential_resolution_matrix.py tests/test_best_effort_helpers_never_raise.py
cd cli && uv run --extra dev ruff format --diff geolens_cli/auth.py tests/test_credential_resolution_matrix.py
```

`ruff format --check` is red across `cli/` on `main` already (32 files) — the repo's ruff gate runs with `working-directory: backend`. The `--diff` above confirms none of the reformat hunks touch a line this PR adds.

No schema, API, or config impact. CLI-only, no SDK regeneration needed.

---

## Round 2 (codex round 1 P2, `auth.py:1576`)

Valid finding. Round 1 rewrote the retained refresh token through `store_refresh_token()` on **every** no-new-refresh-token rotation, including the ordinary one where the bearer landed in the backend it was already in — asking a second keyring account for a write of a secret that did not change.

Regression that opens: a keyring whose refresh account is readable but rejects writes, while the bearer and fingerprint accounts accept them, with no usable `credentials.toml` behind it. The write fails, the file fallback fails, and `try_refresh()` reports a rotation the server already performed as a failure — leaving the new bearer stored against the OLD fingerprint, so the next attempt discards a still-valid refresh token as unpaired. The pre-#1807 fingerprint-only path completed that rotation.

The retained-token branch now turns on whether the bearer actually moved backends:

- **unchanged** (the common case, and every `--no-keyring` profile) → `_rewrite_refresh_fingerprint()`, restored, touching exactly the one member that went stale;
- **moved** (keyring → file fallback) → the full `store_refresh_token()` migration plus the keyring cleanup, which is what this PR is about.

One `backend_moved` predicate gates both the write and the cleanup, so they cannot disagree about whether a move happened.

New matrix row: `TestRefreshPairingRefreshAccountWriteRejectedRow` — unchanged backend, refresh account rejects writes, no file fallback. Asserts the refresh succeeds, the fingerprint is updated in place, the retained token is untouched, nothing reaches the file, and the token is still paired and spendable on the next expiry.

Counterfactual against `732fb1089` (the round 1 head): the new row fails with `assert 3 == 0` — `SystemExit(3)`, EXIT_AUTH, the lost rotation. The two backend-fallback rows pass either way, as required.

```
cd cli && uv run --extra dev python -m pytest -q          # 648 passed (+1)
cd cli && uv run --extra dev python -m pytest -q tests/test_best_effort_helpers_never_raise.py tests/test_refresh_token_backend_derivation.py tests/test_credential_set_structural.py tests/test_refresh.py   # 70 passed
cd cli && uv run --extra dev ruff check geolens_cli tests/test_credential_resolution_matrix.py tests/test_best_effort_helpers_never_raise.py
```

The never-raise gate and both backend-derivation structural gates stay green: the two remaining `store_refresh_token()` call sites still derive `no_keyring=(backend != "keyring")` and pass `bearer_token=`.

### DIFF
--- a/cli/geolens_cli/auth.py
+++ b/cli/geolens_cli/auth.py
@@ -475,6 +475,45 @@ def _load_refresh_fingerprint(instance: str) -> Optional[str]:
         raise KeyringCredentialUnreadable("refresh_fingerprint", str(exc)) from exc
 
 
+def _rewrite_refresh_fingerprint(
+    instance: str, bearer_token: str, *, no_keyring: bool
+) -> None:
+    """Rewrite ONLY the pairing fingerprint to match ``bearer_token``,
+    without touching the refresh token value itself.
+
+    fix(#1778 round 31): used by try_refresh() when a rotation renews
+    the access token but the server does NOT issue a new refresh
+    token -- the existing (unrotated) refresh token still needs its
+    fingerprint updated to the NEW bearer, or the next refresh attempt
+    would find it "mismatched" against the very rotation this call
+    just performed and discard a perfectly good, still-valid refresh
+    token as if it were stale.
+
+    fix(#1807 round 2): still the right call whenever the bearer stayed
+    in the backend it was already in, which is every rotation but the
+    fallback one. Rewriting the unchanged refresh SECRET there (what
+    round 1 did unconditionally) asks a second account for a write
+    nothing needs, and a keyring that accepts the bearer and
+    fingerprint accounts while refusing the refresh one -- with no
+    usable file fallback behind it -- then loses a rotation this helper
+    would have completed: try_refresh() reports failure with the new
+    bearer already stored against the OLD fingerprint, and the next
+    attempt discards a still-valid refresh token as unpaired. Only the
+    backend-MOVED case needs the whole pair rewritten; see
+    try_refresh().
+    """
+    fingerprint = _fingerprint_bearer(bearer_token)
+    if not no_keyring:
+        try:
+            keyring.set_password(
+                SERVICE, _keyring_account_refresh_fingerprint(instance), fingerprint
+            )
+            return
+        except KeyringError as exc:
+            log.warning("keyring_unavailable_falling_back_to_file", error=str(exc))
+    _set_credential_field(instance, "refresh_fingerprint", fingerprint)
+
+
 def _delete_keyring_refresh_entries(instance: str) -> None:
     """Best-effort delete of the refresh token AND its pairing
     fingerprint from the KEYRING only, leaving the file side alone.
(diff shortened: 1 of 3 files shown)


ishiland:
@codex review

ishiland:
@codex review

PR #1819, opened 2026-09-04
test(e2e): wait on the layer-add response, not a fresh map read, in the duplicate-renderings spec

### LINKED ISSUE #1810: flake(e2e): builder duplicate-renderings poll sees 0 layers after a 201 under host load
`e2e/builder.spec.ts:351` (duplicate dataset renderings) failed twice in seven runs during the 2026-09-03 post-#1804 smoke: the poll expected 3 layers and saw 0 after every map call returned 200 or 201 (no 4xx/5xx on `/maps/` routes in the trace). `--repeat-each 4` in isolation passed 4/4, so it is load sensitive. The spec predates this wave. Likely a `getMapDetails` read racing the layer-add commit. Investigate the ordering and make the poll wait on the write's own response rather than a fresh read.

Part of the 2026-08-30 audit follow-ups tracked in #1778.

ishiland:
Fixes #1810

## Cause

`duplicates dataset renderings from row overflow and Add Dataset modal` (`e2e/builder.spec.ts`) polled a fresh `GET /api/maps/{id}` right after each layer-add `POST` to (a) find the newly created layer and (b) confirm the total layer count. Under host load that poll intermittently observed a stale/short-lived snapshot (as low as 0 layers) instead of the just-committed write, even though the trace showed every `/maps/` call returning 200/201 (no 4xx/5xx) — i.e. the poll was reading a snapshot that predated the write it was supposed to be confirming, not failing outright.

This is a spec issue, not a product bug: the poll used raw `fetch()` calls from the Playwright test process, entirely bypassing the frontend/TanStack Query layer, so there's no client-side invalidation-ordering path to fix here.

## Fix

The `add_layer` `POST` response already carries the created layer (that response *is* the write's own confirmation). Instead of re-reading the map afterward, the spec now builds the "after" layer list from the known "before" state (fetched once, before any mutation) plus the layer returned directly in each `POST` response. The two racy `expect.poll(() => getMapLayers(...))` calls are removed entirely, since there's no longer a fresh read to poll. The existing `#stack-row-{id}` visibility assertions stay, confirming the page's own render of the write.

## Verification

`--repeat-each 10 --workers 1`, run against the worktree frontend per AGENTS.md's worktree recipe (`API_PROXY_TARGET=http://localhost:8001 npm run dev -- --port 5174`, then `E2E_ALLOW_WORKTREE=1 E2E_BASE_URL=http://localhost:5174 npx playwright test e2e/builder.spec.ts -g "duplicates dataset renderings" --project=chromium --repeat-each 10 --workers 1`):

- Before: 10/10 passed
- After: 10/10 passed

The race is load-sensitive and didn't reproduce locally even under synthetic CPU load (matches the issue's own note that isolated `--repeat-each 4` runs pass 4/4) — so this isn't a before-red/after-green repro. The fix instead removes the read-after-write race class structurally: the spec no longer performs a fresh read that can observe a pre-write snapshot at all.

`npx eslint e2e/builder.spec.ts` (via frontend's config) is clean. No frontend/backend product code touched, so `npm run lint`/`npm run typecheck` in `frontend/` don't apply.

### DIFF
--- a/e2e/builder.spec.ts
+++ b/e2e/builder.spec.ts
@@ -372,24 +372,27 @@ test.describe.serial('Map Builder', () => {
       (resp) => resp.url().includes(`/api/maps/${mapId}/layers`) && resp.request().method() === 'POST',
     );
     await page.getByRole('menuitem', { name: /^Duplicate$/i }).click();
-    expect([200, 201]).toContain((await rowDuplicateResponse).status());
+    const rowDuplicateRes = await rowDuplicateResponse;
+    expect([200, 201]).toContain(rowDuplicateRes.status());
 
-    let afterRowLayers: MapLayerListItem[] = [];
-    await expect.poll(async () => {
-      afterRowLayers = await getMapLayers(mapId, headers);
-      return afterRowLayers.length;
-    }).toBe(initialLayers.length + 1);
+    // fix(#1810): read the new layer straight off the write's own response
+    // body instead of re-fetching the map right after — a fresh GET here
+    // raced the layer-add commit under host load and intermittently saw a
+    // pre-write (sometimes empty) snapshot. The response IS the commit's
+    // confirmation, so build the "after" state from it plus the known
+    // "before" state rather than a racy re-read.
+    const rowDuplicate = await rowDuplicateRes.json() as MapLayerListItem;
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review

PR #1820, opened 2026-09-04
fix(sources): reserve the ArcGIS sign-in attempt before the mint and hold no connection across it

### LINKED ISSUE #1768: fix(reupload): commit should carry an expected-origin condition so a concurrent origin change cannot be severed
Found by review on #1767 (`geolens replace`). The CLI refuses to replace a dataset whose origin is a service, STAC or registered table, but that is a client-side precheck: it reads the dataset once, then uploads, previews and waits for confirmation. If a service or STAC re-upload of the same dataset commits in that window, the later file commit is accepted, and `tasks_reupload.py` (~426-440) rebinds every file swap to an upload origin, silently severing the remote binding that was just established. The web re-upload dialog has the same window.

`reupload_commit` performs no origin check today; `refresh_not_applicable` lives only in `router_refresh.py`.

Fix shape: the commit request accepts an optional `expected_origin_kind` (and optionally the origin ref hash) captured at upload time; the commit door compares it against the dataset's current origin under the dataset mutation lock and refuses with a coded 409 (`origin_changed`, the same vocabulary the refresh door already uses for a moved source) when they differ. The CLI and the dialog send the value they saw. Additive schema, OpenAPI and SDK regeneration, one door, one test per origin kind.

Cross-ref #1739, #1767, ADR-002 invariant 7 (one mutation at a time) which bounds but does not close this window.

ishiland:
## What

`POST /datasets/{id}/reupload/{job_id}/commit` now accepts an optional
`expected_origin_kind` — the dataset origin the client saw when it staged the
replacement. The door compares it against the dataset's current origin and
refuses a mismatch with `409 {"code": "origin_changed", ...}`, the same
vocabulary the refresh door in `router_refresh.py` already uses for a source
that moved.

## Why — the race

`geolens replace` (#1767) and the web re-upload dialog both refuse to replace
a dataset bound to a service, a STAC item, or a registered table. That refusal
is a **client-side precheck**, decided from a single read taken before the
upload. Between that read and the commit the user confirms sits an upload, a
preview, and a human. A service or STAC re-upload committing in that window is
invisible to the client, and the swap the file commit queues rebinds the
dataset to `upload` unconditionally (`_apply_reupload_swap`,
`tasks_reupload.py` ~426-440) — silently severing the remote binding that was
just established.

ADR-002 invariant 7 (one mutation at a time) bounds that window but does not
close it: the competing run has already *finished* by the time the file commit
lands.

## Where the check sits, and why

After `create_pending_run`, not before. That insert is the one-active-run
admission gate, so with the slot held no other run for this dataset can be in
flight — any origin change is already committed, and a READ COMMITTED re-read
sees it. Placed earlier, the re-read would be one more racing read. On a
mismatch the whole request rolls back, so the reservation is released rather
than leaked (a leaked run row would refuse every refresh of the dataset until
the stale-run sweep's cutoff).

It lives in `_refuse_if_origin_changed` rather than inline because two more
branches pushed `reupload_commit` past the McCabe gate — the same reason
`_dispatch_reupload_task` exists.

## Contract change — additive, old clients unaffected

`expected_origin_kind` is **optional**. A client that sends nothing gets
exactly the pre-#1768 behaviour: no comparison, no refusal. An older CLI, SDK,
or frontend build keeps working unchanged, and `test_an_omitted_expected_origin_keeps_the_old_contract`
pins that. The value is typed with the shared `OriginKind` literal from
`platform/dataset_origin.py`, so the closed set is published in OpenAPI and an
unknown kind is a 422 at the schema boundary rather than a silent no-op.

Only `expected_origin_kind` is carried, not an origin-ref hash. The issue
listed the hash as optional; the kind is what every client-side gate actually
decides from, and a hash would add a second comparison surface with no
refusal the kind does not already cover.

## Clients

- **CLI** (`geolens replace`): captures the origin at the same Stage 0 read
  its own gate is decided from and sends it with the commit. An origin the CLI
  cannot classify (redacted → SDK `UNSET`, or a kind a newer server added)
  asserts nothing rather than being forwarded, so a stale CLI degrades to the
  old behaviour instead of 422-ing. `origin_changed` gets an actionable
  refusal message.
- **Web dialog**: captures `dataset.origin` beside the job id at staging time.
  `dataset` is a live query result, so reading it at confirm time would send
  the very change the condition exists to catch — there is a test for exactly
  that.
- **i18n**: `errors.reuploadOriginChanged` added to en/es/fr/de. It is a
  separate key from `errors.refreshOriginChanged` even though the code is
  shared, because the two refusals describe different windows and ask for
  different actions.

## Regenerated artifacts

`backend/openapi.json`, both SDKs (`make sdks` — one new Python literal
module), and `frontend/src/types/api.generated.ts`. Hand-typed mirrors
(`frontend/src/types/api.ts`) updated to match.

Fixes #1768

## Verification

```bash
# backend (host, Postgres at :5434)
uv run pytest tests/test_reupload_expected_origin_1768.py \
  tests/test_reupload.py tests/test_reupload_service.py \
  tests/test_job_cancel_reupload_commit.py tests/test_reupload_idor.py \
  tests/test_reupload_record_type_guard.py tests/test_sdks_round_trip.py -q

# contract + SDKs
make openapi-check
make sdks-check

# CLI
cd cli && uv run --extra dev python -m pytest -q

# frontend
npx vitest run src/components/dataset src/api src/lib/__tests__/error-map.test.ts
```

Results: backend 100 passed / 1 skipped (the TS-SDK round-trip case, which
needs `sdks/typescript/dist` built) plus 153 passed on the structural gates;
CLI 652 passed; frontend 593 passed across the touched trees; `openapi-check`
and `sdks-check` both clean.

No migration, no config change. `test_layering.py` caps raised with comments
for `router_reupload.py` (1294 → 1355) and `schemas.py` (1499 → 1511).

---

# r1b — audit round

Four findings from the pre-trigger audit, plus the coverage that was missing for them.

## P1 — reading a pending task's exception (`signin_guard.py`)

At the drain deadline the finaliser called `settle.cancel()` and then read `settle.exception()`. `Task.cancel()` only *requests* cancellation, so on the very next line the task is still pending: `done()` is False, `cancelled()` is False, and `Future.exception()` raises `InvalidStateError`. That escaped `_signin_settle_cancelled`, which the route awaits inside `except asyncio.CancelledError:` before a bare `raise` — so on the one path the one-second ceiling exists for, the handler propagated `InvalidStateError` instead of re-raising the cancellation, and the warning saying the row was lost was never logged.

Confirmed with a plain asyncio repro before fixing:

```
done() after cancel(): False
cancelled() after cancel(): False
exception() raises: InvalidStateError
after a turn -> done(): True cancelled(): True
```

The outcome is now read in one place, `_settle_failure`, which answers **pending first**. A cancellation this module asked for reads as `CancelledError` whether or not the task has finished unwinding, because that is what stopped the write either way.

## P2 — the settle write ran on the request's session

On the deadline path the write was still going when `_signin_settle_cancelled` returned, the route re-raised, and FastAPI ran `get_db`'s teardown — closing the session, and its connection, under a live statement. `_write_cancelled_outcome` now opens and closes a session of its own, so an abandoned write is a task that finishes quietly rather than one that faults with an `InterfaceError` or a greenlet error on a shutting-down worker. Late-bound `async_session` import, per the #909 rule `test_layering.py` enforces.

## P3 — sweep deadlock

`_sweep_expired_signin_attempts` now selects `ORDER BY id ... FOR UPDATE SKIP LOCKED` and deletes by id. Two sign-ins for different scopes hold different advisory locks, so their sweeps run concurrently, and a bare `DELETE ... WHERE attempted_at < ...` let them take the same expired rows in whatever order the plan chose — a 40P01 on a housekeeping statement, failing a sign-in that had nothing wrong with it. The loser of a race now drops the row instead of queueing; it is already expired and the winner is deleting it.

## P3 — the AUDIT_CANCELLED comment was wrong

It said external cancellation is "a uvicorn shutdown rather than anything a client can trigger". `BaseHTTPMiddleware` also cancels the downstream task on client disconnect, so a caller who hangs up reaches it too. The comment now says so — and says why that is safe: the reservation, not this row, is what counts the attempt, so a client cannot make a credential POST free by disconnecting. The same correction is applied to `_signin_reserve`'s docstring and the route's inline comment.

## P3 accepted, not changed

The drain allocates one shield future per event-loop turn. That is the price of surviving anyio's re-arm and it is only paid *while* a re-arm is happening: with no cancel scope re-arming, the loop awaits once and blocks there until the write finishes, so it is one allocation. `asyncio.sleep(0)` would replace it with an unconditional spin, including on the paths that have no re-arm. The comment in the loop now states this.

## New coverage

- `test_a_settle_that_outruns_the_drain_still_leaves_cancellederror` — hangs `_signin_audit` past a monkeypatched 50 ms ceiling and asserts the warning is logged, that `CancelledError` is what leaves the handler, and that the ledger row still stands.
- `test_the_settle_outcome_reader_treats_pending_as_cancelled` — pins `_settle_failure` over all four future states, with no event loop in the way.

### Counterfactuals

With the P1 fix reverted (`if settle.cancelled():` instead of `if not settle.done() or settle.cancelled():`), both go red:

```
E       AssertionError: []
E       assert False
E        +  where False = any(<generator object ...>)
E       asyncio.exceptions.InvalidStateError: Exception is not set.
FAILED tests/test_arcgis_signin.py::test_a_settle_that_outruns_the_drain_still_leaves_cancellederror
FAILED tests/test_arcgis_signin.py::test_the_settle_outcome_reader_treats_pending_as_cancelled
```

The pool counterfactual is unchanged and re-run on this round: `_signin_reserve` ending in `flush()` instead of `commit()` still gives `5 pooled connections were held while 5 credential POSTs were in flight`.

### Counts on this round

`test_arcgis_signin.py` + tenancy + layering + rule1 + rule2: **333 passed**. `test_arcgis_signin.py` under `-n 4`: **177 passed**, three consecutive runs. `ruff check` / `ruff format --check` clean.

`_MODULE_LOC_CAPS` for `sources/arcgis_signin.py` goes 1293 → 1297 (exact) for the corrected comment; the entry says what the lines bought.

---

# r1c — audit round

## The disconnect claim from r1b was wrong; three comments retracted

r1b added the assertion that `BaseHTTPMiddleware` cancels the downstream task on client disconnect. That is false for the pinned Starlette 1.6.0, checked against the installed source rather than recalled:

- `BaseHTTPMiddleware` contains exactly one `cancel()` — `task_group.cancel_scope.cancel()` at `middleware/base.py:121` — and it lives inside `receive_or_disconnect`'s **local** task group, ending the sibling `response_sent.wait` race. It does not touch the downstream coroutine.
- A client hanging up arrives as an `http.disconnect` **message** on the receive channel, which a non-streaming route never reads. Uvicorn only sets `cycle.disconnected = True`.

So a worker shutdown is the only external cancellation that reaches the handler. All three comments (`arcgis_signin.py` at `AUDIT_CANCELLED`, the route's `except asyncio.CancelledError` clause, `_signin_reserve`'s docstring) now say that, cite the line they were checked against, and say what makes the path safe whatever the source turns out to be here or in a later Starlette: **the reservation is committed before the credential POST, so no cancellation makes one free.**

The re-arm explanation in `_signin_settle_cancelled` is a separate claim and it stands — it is now precise about its mechanism. The downstream app runs as a child of the anyio task group each middleware layer starts it in (`task_group.start_soon(coro)`, `middleware/base.py:148`), and an anyio cancel scope re-arms cancellation on every await a task makes inside it while cancelled. That is what a bare shielded await loses to on shutdown, and what the drain absorbs.

## The settle task had no strong reference

`asyncio` holds tasks weakly, and this one is deliberately left running when the drain gives up on it at the deadline — so nothing referenced it and the collector could take it mid-write. A module-level `_SETTLE_TASKS` set holds it, and `add_done_callback(_SETTLE_TASKS.discard)` releases it however it finishes.

## Counts on this round

`test_arcgis_signin.py` + `test_layering.py`: **226 passed**. `ruff check` / `ruff format --check` clean.

`_MODULE_LOC_CAPS` for `sources/arcgis_signin.py` re-pinned 1297 → **1301** (exact, measured) for the corrected comment.

### DIFF
--- a/backend/alembic/versions/0058_arcgis_signin_attempts_user_scope.py
+++ b/backend/alembic/versions/0058_arcgis_signin_attempts_user_scope.py
@@ -1,0 +1,64 @@
+"""Carry the per-caller ArcGIS sign-in budget on the cluster-global ledger.
+
+fix(#1775). ``POST /api/services/arcgis/signin/`` now reserves a counted
+attempt in its own short transaction BEFORE the credential POST and writes the
+audit row afterwards, so that a request cancelled mid-POST is already counted:
+ArcGIS may have counted that password, and GeoLens must not be the side that
+forgot it.
+
+That reordering moves the per-caller budget's source. It was counted from
+``catalog.audit_logs`` filtered on ``user_id`` and the ``token_service_host``
+detail, and a cancelled request writes no audit row at all, so exactly the
+attempts this issue exists to preserve would have gone uncounted. The budget
+therefore reads the ledger, and the ledger needs a column to key it on.
+
+``user_scope`` is the same kind of value as ``account_key``: an HMAC-SHA256
+digest under a key derived from the instance JWT secret, here over the caller's
+user id and the canonical token-service scope. It is NOT a plaintext user id.
+Migration 0056's rule that this table holds nothing tenant-identifying and
+nothing secret is what makes it safe to keep outside the RLS boundary, and a
+digest keeps that rule intact.
+
+Nullable, and not backfilled. A row written before this migration stands for an
+attempt whose caller cannot be recovered, and charging it to an arbitrary
+caller would be worse than not charging it at all. Every row is swept fifteen
+minutes after it is written, so the gap closes itself within one window.
+
+Revision ID: 0058_arcgis_signin_user_scope
+Revises: 0057_security_revocation_generation
+Create Date: 2026-09-03
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "0058_arcgis_signin_user_scope"
+down_revision: Union[str, None] = "0057_security_revocation_generation"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "arcgis_signin_attempts",
+        sa.Column("user_scope", sa.String(length=64), nullable=True),
+        schema="catalog",
+    )
+    # The same shape as the account index: the windowed count for one caller.
+    op.create_index(
+        "ix_catalog_arcgis_signin_attempts_user_time",
+        "arcgis_signin_attempts",
+        ["user_scope", "attempted_at"],
+        schema="catalog",
+    )
+
+
+def downgrade() -> None:
+    op.drop_index(
+        "ix_catalog_arcgis_signin_attempts_user_time",
+        table_name="arcgis_signin_attempts",
+        schema="catalog",
+    )
+    op.drop_column("arcgis_signin_attempts", "user_scope", schema="catalog")
--- a/backend/app/modules/catalog/sources/arcgis_signin.py
+++ b/backend/app/modules/catalog/sources/arcgis_signin.py
@@ -191,12 +191,16 @@ AUDIT_DISCOVERY_TIMEOUT = "discovery_timeout"
 # this instance would not follow.
 AUDIT_DISCOVERY_UNTRUSTED_DELEGATE = "discovery_untrusted_delegate"
 # fix(#1775): the outcome of an attempt whose credential POST was interrupted
-# by external task cancellation, which is a uvicorn shutdown rather than
-# anything a client can trigger. GeoLens does not know whether ArcGIS counted
-# it, so it is recorded as its own outcome rather than guessed at, and it is
-# absent from UNCOUNTED_SIGNIN_RESULTS below because the password was already
-# on the wire. The row is best effort; what makes the attempt count is the
-# reservation the route commits BEFORE the POST.
+# by external task cancellation. fix(#1775 audit): a worker shutdown is one
+# source and NOT the only one — Starlette's `BaseHTTPMiddleware` cancels the
+# downstream task when the client disconnects, so a caller who hangs up can
+# reach this too. That is why the reservation, not this row, is what counts
+# the attempt: a client cannot make a credential POST free by disconnecting.
+# GeoLens does not know whether ArcGIS counted it, so it is recorded as its
+# own outcome rather than guessed at, and it is absent from
+# UNCOUNTED_SIGNIN_RESULTS below because the password was already on the wire.
+# The row is best effort; what makes the attempt count is the reservation the
+# route commits BEFORE the POST.
 AUDIT_CANCELLED = "cancelled"
 
 # The outcomes above that are NOT an attempt against ArcGIS, because GeoLens
--- a/backend/app/modules/catalog/sources/arcgis_signin.py
+++ b/backend/app/modules/catalog/sources/arcgis_signin.py
@@ -190,6 +190,14 @@ AUDIT_DISCOVERY_TIMEOUT = "discovery_timeout"
 # what this records is that the portal tried to send the password somewhere
 # this instance would not follow.
 AUDIT_DISCOVERY_UNTRUSTED_DELEGATE = "discovery_untrusted_delegate"
+# fix(#1775): the outcome of an attempt whose credential POST was interrupted
+# by external task cancellation, which is a uvicorn shutdown rather than
+# anything a client can trigger. GeoLens does not know whether ArcGIS counted
+# it, so it is recorded as its own outcome rather than guessed at, and it is
+# absent from UNCOUNTED_SIGNIN_RESULTS below because the password was already
+# on the wire. The row is best effort; what makes the attempt count is the
+# reservation the route commits BEFORE the POST.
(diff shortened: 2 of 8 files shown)


ishiland:
@codex review

ishiland:
PR-open codex threads: the settle-drain finding was fixed in r1b (`_settle_failure` reads pending-first and the settle runs on its own session; verified by counterfactual). The post-mint cancellation window (audit row lost during settlement under worker shutdown; reservation stays counted) is filed as https://github.com/geolens-io/geolens/issues/1825. Codex round on 06be8b872 was CLEAN; merging.

PR #1821, opened 2026-09-04
fix(reupload): refuse a commit whose expected origin no longer matches

### LINKED ISSUE #1768: fix(reupload): commit should carry an expected-origin condition so a concurrent origin change cannot be severed

ishiland:
## What

`POST /datasets/{id}/reupload/{job_id}/commit` now accepts an optional
`expected_origin_kind` — the dataset origin the client saw when it staged the
replacement. The door compares it against the dataset's current origin and
refuses a mismatch with `409 {"code": "origin_changed", ...}`, the same
vocabulary the refresh door in `router_refresh.py` already uses for a source
that moved.

## Why — the race

`geolens replace` (#1767) and the web re-upload dialog both refuse to replace
a dataset bound to a service, a STAC item, or a registered table. That refusal
is a **client-side precheck**, decided from a single read taken before the
upload. Between that read and the commit the user confirms sits an upload, a
preview, and a human. A service or STAC re-upload committing in that window is
invisible to the client, and the swap the file commit queues rebinds the
dataset to `upload` unconditionally (`_apply_reupload_swap`,
`tasks_reupload.py` ~426-440) — silently severing the remote binding that was
just established.

ADR-002 invariant 7 (one mutation at a time) bounds that window but does not
close it: the competing run has already *finished* by the time the file commit
lands.

## Where the check sits, and why

After `create_pending_run`, not before. That insert is the one-active-run
admission gate, so with the slot held no other run for this dataset can be in
flight — any origin change is already committed, and a READ COMMITTED re-read
sees it. Placed earlier, the re-read would be one more racing read. On a
mismatch the whole request rolls back, so the reservation is released rather
than leaked (a leaked run row would refuse every refresh of the dataset until
the stale-run sweep's cutoff).

It lives in `_refuse_if_origin_changed` rather than inline because two more
branches pushed `reupload_commit` past the McCabe gate — the same reason
`_dispatch_reupload_task` exists.

## Contract change — additive, old clients unaffected

`expected_origin_kind` is **optional**. A client that sends nothing gets
exactly the pre-#1768 behaviour: no comparison, no refusal. An older CLI, SDK,
or frontend build keeps working unchanged, and `test_an_omitted_expected_origin_keeps_the_old_contract`
pins that. The value is typed with the shared `OriginKind` literal from
`platform/dataset_origin.py`, so the closed set is published in OpenAPI and an
unknown kind is a 422 at the schema boundary rather than a silent no-op.

Only `expected_origin_kind` is carried, not an origin-ref hash. The issue
listed the hash as optional; the kind is what every client-side gate actually
decides from, and a hash would add a second comparison surface with no
refusal the kind does not already cover.

## Clients

- **CLI** (`geolens replace`): captures the origin at the same Stage 0 read
  its own gate is decided from and sends it with the commit. An origin the CLI
  cannot classify (redacted → SDK `UNSET`, or a kind a newer server added)
  asserts nothing rather than being forwarded, so a stale CLI degrades to the
  old behaviour instead of 422-ing. `origin_changed` gets an actionable
  refusal message.
- **Web dialog**: captures `dataset.origin` beside the job id at staging time.
  `dataset` is a live query result, so reading it at confirm time would send
  the very change the condition exists to catch — there is a test for exactly
  that.
- **i18n**: `errors.reuploadOriginChanged` added to en/es/fr/de. It is a
  separate key from `errors.refreshOriginChanged` even though the code is
  shared, because the two refusals describe different windows and ask for
  different actions.

## Regenerated artifacts

`backend/openapi.json`, both SDKs (`make sdks` — one new Python literal
module), and `frontend/src/types/api.generated.ts`. Hand-typed mirrors
(`frontend/src/types/api.ts`) updated to match.

Fixes #1768

## Verification

```bash
# backend (host, Postgres at :5434)
uv run pytest tests/test_reupload_expected_origin_1768.py \
  tests/test_reupload.py tests/test_reupload_service.py \
  tests/test_job_cancel_reupload_commit.py tests/test_reupload_idor.py \
  tests/test_reupload_record_type_guard.py tests/test_sdks_round_trip.py -q

# contract + SDKs
make openapi-check
make sdks-check

# CLI
cd cli && uv run --extra dev python -m pytest -q

# frontend
npx vitest run src/components/dataset src/api src/lib/__tests__/error-map.test.ts
```

Results: backend 100 passed / 1 skipped (the TS-SDK round-trip case, which
needs `sdks/typescript/dist` built) plus 153 passed on the structural gates;
CLI 652 passed; frontend 593 passed across the touched trees; `openapi-check`
and `sdks-check` both clean.

No migration, no config change. `test_layering.py` caps raised with comments
for `router_reupload.py` (1294 → 1355) and `schemas.py` (1499 → 1511).

### DIFF
--- a/backend/app/modules/catalog/datasets/api/router_reupload.py
+++ b/backend/app/modules/catalog/datasets/api/router_reupload.py
@@ -60,6 +60,7 @@ from app.platform.refresh.service import (
     create_pending_run,
     make_refresh_run_failed_rollback,
 )
+from app.platform.dataset_origin import classify_origin
 from app.platform.extensions import get_catalog_port
 from app.core.persistent_config import UPLOAD_MAX_SIZE_MB, get_allowed_extensions_list
 from app.modules.quota.service import check_replacement_quota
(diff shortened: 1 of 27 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
Round 1 P2 fixed in 58db1357d (invalidate the dataset detail query on origin_changed). Round 2 P2 (await the refetch before enabling retry) filed as #1822; merging on no-P1.

PR #1823, opened 2026-09-04
fix(refresh): re-derive geom_4326 for registered tables so out-of-band writes render

### LINKED ISSUE #1738: Registered tables: `geom_4326` is populated once and never re-derived, so out-of-band writes never render
## The contract as written

`register_existing_table`'s docstring (`backend/app/processing/ingest/service.py:644-654`) says:

> registration copies no data and serves from the live table, so the owner keeps writing to it directly

## What the code supports

`add_4326_column` (`backend/app/processing/ingest/metadata_projection.py:46-51`) creates a plain column and populates it once:

```sql
ALTER TABLE {tref} ADD COLUMN IF NOT EXISTS geom_4326 geometry(Geometry, 4326)
```

followed by a one-time `UPDATE`. It is not a generated column, and nothing re-derives it afterwards. `refresh_postgis` cannot: its data-schema phase runs under `REPEATABLE READ` with `postgresql_readonly: True` (`tasks_postgis_refresh.py:518-522`), and the comment there says a write from that phase should fail loudly by design. Its only metadata call is `extract_metadata` at `:554`.

So any row the owner writes directly has `geom_4326` NULL. Every reader filters on it: vector tiles at `tiles/service.py:272` and `:460` (`WHERE t.geom_4326 && bounds.geom_4326`), feature reads, analysis. Those rows exist in the table and are invisible in the product, with no signal anywhere that they were dropped.

Bulk reload is worse. `ogr2ogr -overwrite` drops and recreates the table, taking `geom_4326` and the reader grant from `grant_reader_access` with it, and nothing restores either. `DELETE FROM` plus `-append` keeps the column and the grant but leaves every new row NULL. Both leave a dataset that looks registered and renders nothing.

I have not found a repair path through the API. Re-registering means deleting the dataset first, which loses its ID, permalinks and any saved maps built on it.

## Why a generated column does not just solve it

Worth recording, since it is the obvious first idea. The derivation is `ST_Force2D(ST_Transform(ST_CurveToLine(geom), 4326))`. `ST_Transform` is not `IMMUTABLE`, so a `STORED GENERATED` column is only possible when the source SRID is already 4326, where the expression collapses to `ST_Force2D(ST_CurveToLine(ST_SetSRID(geom, 4326)))`. The code already handles encountering a generated column (`metadata_projection.py:105-115` skips the UPDATE when `is_generated = 'ALWAYS'`), it just never creates one.

## What to decide

Either:

- **Narrow the promise.** Registration supports tables that GeoLens itself writes, or tables an owner replaces by re-registering. Say so in the docstring and in whatever operator-facing docs describe the Register tab.
- **Support the promise.** Give registered tables a way to re-derive `geom_4326` and re-apply the grant, as a separate maintenance action rather than folding a write into the read-only refresh phase, which exists for good reasons.

Marking `needs-design` because the choice is a product one. The current state is the one thing that should not persist: the docstring invites a workflow that silently drops data.

ishiland:
## The bug

`geom_4326` on a registered table is a plain column, written **once** — by `add_4326_column` at registration — and never re-derived (design §1). Registration copies no data and serves from the live relation, so the owner keeps writing to that table directly, and none of those writes touch the render column:

- `UPDATE ... SET geom = ...` moves the feature; it keeps drawing where it used to be.
- `DELETE` + `INSERT` (an ETL's ordinary reload) lands rows whose `geom_4326` is NULL.
- `ogr2ogr -overwrite` drops the table and recreates it with no `geom_4326` column, no GiST index and no reader grant at all.

Every reader filters on `geom_4326 && <envelope>`, and `NULL && anything` is NULL (§1), so affected rows go **silently invisible** — tiles, feature reads, extent, OGC/CQL2, exports, analysis — rather than visibly wrong. `refresh_postgis` could not fix it: its measurement phase runs `postgresql_readonly=True` on purpose. Latent since #1265.

Fixes #1738.

## Why option (d)

Design §2 prices five options and §3 recommends (d), a re-derive phase inside `refresh_postgis`:

- **(a) STORED generated column** — PostgreSQL rejects any non-DEFAULT write to a generated column, which breaks in-app feature editing on every registered dataset (`features/service.py`), and an out-of-band INSERT with a different SRID starts raising inside the owner's ETL. Plus a full rewrite under ACCESS EXCLUSIVE.
- **(b) BEFORE INSERT OR UPDATE trigger** — a per-row trigger on somebody else's table silently changes the semantics of their bulk loads. A larger claim on a table we promise not to own than adding a column is, and invisible to them.
- **(c) drop the column, read through a functional index** — `geom_4326` is a name in ~38 backend modules, in the analysis sandbox's managed-column model and in the NL→SQL prompt contract. Out on blast radius.
- **(e) detect and warn** — tells an operator their data is invisible without giving them a fix; re-registration is not one, since it costs the dataset id, its permalinks and its saved maps.
- **Nothing in-table survives `-overwrite`** (§2, closing paragraph): the table is dropped, taking (a), (b) and (c)'s index with it. Only an invariant GeoLens re-applies **from outside** comes back — which is (d).

Refresh is also where the machinery already is: an admission gate, a run ledger, a user-visible history, a tile-version bump and a cache purge.

## What this does

**`rederive_geom_4326`** (`metadata_projection.py`) re-applies the whole invariant with the same DDL registration uses: `ADD COLUMN IF NOT EXISTS`, the same derive expression (now factored into `_geom_4326_expr`, so the write path and the repair path cannot drift), and `ensure_geom_4326_gist_index`. It skips a STORED GENERATED column (same refusal `linearize_existing_4326` makes) and a table with no geometry (#1359 attribute tables).

**Phase 1.5 in `refresh_postgis`** (`_repair_geom_4326`) runs it in its own session and transaction, **before** the read-only measurement phase (§3.1), then re-issues `grant_reader_access`. Reading `content_version` after the repair (§3.3) is free, because phase 2 already reads it after this point.

`register_existing_table`'s docstring is corrected (§3.4): the owner's direct writes are picked up by Refresh, not continuously.

### `-overwrite` behaviour

The recreated table gets its column, its GiST index and its `geolens_reader` GRANT back, against the same dataset id — no delete-and-re-add. Pinned by `test_an_overwrite_restores_the_column_the_index_and_the_grant`, which asserts all three are absent first.

### Only rows that changed

`UPDATE ... WHERE ST_AsBinary(geom_4326) IS DISTINCT FROM ST_AsBinary(<expr>)` — one statement, not a ctid-batched loop, because the statement **can** be bounded (below) and because one statement keeps the whole repair atomic: a table is either fully re-derived or untouched, never half. Deliberately *not* `geom_4326 IS NULL OR ...`: a row whose `geom` is NULL already has the correct NULL render value, and that disjunct would rewrite every such row NULL-to-NULL on every pass, which is both a pointless write and a drift count that lies.

The tile-version bump and the cache purge are gated on rewritten **rows**, not on the column or the index: restoring an index changes how a tile is computed, not what it contains, so a looser gate would bust every cached tile of every registered dataset on the first refresh after this ships.

### The timeout bound

`install_api_statement_timeout` is installed by the API process only, so a worker statement has no deadline at all. The repair phase sets **two** through `set_config(..., is_local => true)` (static SQL, bound values):

- `_REPAIR_STATEMENT_TIMEOUT_MS = 300_000` — the work bound.
- `_REPAIR_LOCK_TIMEOUT_MS = 5_000` — the wait bound. Measurement forced this one: `ADD COLUMN` takes ACCESS EXCLUSIVE, a merely *queued* lock request already blocks every reader behind it, and the first version of this phase sat on that queue for the full five minutes in a test. Five seconds gives the queue position back instead.

Both are non-fatal and coded (`timed_out`, `blocked`, `failed`, `not_applicable`, `repaired`), logged on every run. A repair that cannot write leaves the dataset exactly as broken as it already was and the refresh still takes its measurement — the change adds no failure mode to a strategy that had none.

**None.** No Alembic migration (so merging does not take the dev API down), no API surface, no config, no OpenAPI or SDK regeneration. Registration-time behaviour is unchanged; the new work only runs when someone presses Refresh.

## Verification

```
  tests/test_registered_geom_4326_rederive_1738.py \
  tests/test_postgis_refresh_1265.py \
  tests/test_register_geom_column_name_1737.py \
  tests/test_registered_delete_detach_1452.py \
  tests/test_staging_pipeline.py tests/test_phase_2_commit_boundary.py \
  tests/test_layering.py tests/test_rule1_structural.py \
  tests/test_rule2_structural.py tests/test_codeql_qtable_suppressions.py -q
```

Full backend suite on this branch: **11721 passed, 100 skipped** (23:27, `-n 4`).

Measured: re-deriving 5,000 out-of-band rows took **0.19 s**; a refresh that finds no drift rewrites 0 rows.

### DIFF
--- a/backend/app/modules/catalog/datasets/domain/models.py
+++ b/backend/app/modules/catalog/datasets/domain/models.py
@@ -622,6 +622,10 @@ class Dataset(Base):
         content (feature edits, column DDL, tile_columns, reupload) — the
         post-commit Valkey purge cannot reach CDN/browser caches keyed on the
         tile URL. fix(#525 B-038)
+
+        Writes an ABSOLUTE value read off this instance, so it is correct only
+        while the caller holds this row; one that does not must use
+        ``bump_tile_cache_version_atomic`` (ingest/tasks_common). #1738 r1
         """
         self.tile_cache_version = (self.tile_cache_version or 1) + 1
 
--- a/backend/app/processing/ingest/metadata.py
+++ b/backend/app/processing/ingest/metadata.py
@@ -72,10 +72,15 @@ from app.processing.ingest.metadata_mercator import (
     clip_to_mercator_bounds,  # noqa: F401 -- re-exported, see __all__
 )
 from app.processing.ingest.metadata_projection import (
+    REPAIR_APPLIED,  # noqa: F401 -- re-exported, see __all__
+    REPAIR_GENERATED,  # noqa: F401 -- re-exported, see __all__
+    REPAIR_NO_GEOMETRY,  # noqa: F401 -- re-exported, see __all__
+    Geom4326Repair,  # noqa: F401 -- re-exported, see __all__
     add_4326_column,  # noqa: F401 -- re-exported, see __all__
     ensure_geom_4326_gist_index,  # noqa: F401 -- re-exported, see __all__
     grant_reader_access,  # noqa: F401 -- re-exported, see __all__
     linearize_existing_4326,  # noqa: F401 -- re-exported, see __all__
+    rederive_geom_4326,  # noqa: F401 -- re-exported, see __all__
 )
 from app.processing.ingest.metadata_quality import (
     _score_attribute_completeness,  # noqa: F401 -- re-exported, see __all__
@@ -92,6 +97,10 @@ from app.processing.ingest.metadata_sql import (
 )
 
 __all__ = [
+    "REPAIR_APPLIED",
+    "REPAIR_GENERATED",
+    "REPAIR_NO_GEOMETRY",
+    "Geom4326Repair",
     "_ABSTRACT_TO_CONCRETE_GEOMETRY_TYPE",
     "_BOX3D_RE",
     "_DEGREE_UNIT_SRTEXT_RE",
@@ -139,6 +148,7 @@ __all__ = [
     "grant_reader_access",
     "linearize_existing_4326",
     "promote_z_to_elev",
+    "rederive_geom_4326",
     "refresh_attribute_metadata",
     "rename_reserved_columns",
 ]
--- a/backend/app/processing/ingest/metadata.py
+++ b/backend/app/processing/ingest/metadata.py
@@ -76,10 +76,12 @@ from app.processing.ingest.metadata_projection import (
     REPAIR_GENERATED,  # noqa: F401 -- re-exported, see __all__
     REPAIR_NO_GEOMETRY,  # noqa: F401 -- re-exported, see __all__
     Geom4326Repair,  # noqa: F401 -- re-exported, see __all__
+    Geom4326State,  # noqa: F401 -- re-exported, see __all__
     add_4326_column,  # noqa: F401 -- re-exported, see __all__
     ensure_geom_4326_gist_index,  # noqa: F401 -- re-exported, see __all__
     grant_reader_access,  # noqa: F401 -- re-exported, see __all__
     linearize_existing_4326,  # noqa: F401 -- re-exported, see __all__
+    probe_geom_4326,  # noqa: F401 -- re-exported, see __all__
     rederive_geom_4326,  # noqa: F401 -- re-exported, see __all__
 )
 from app.processing.ingest.metadata_quality import (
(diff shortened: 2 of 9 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
Round 1 P2s fixed in bd03d7de5 (unconditional reader grant + probe split; atomic tile-version increment). Round 2: GiST index restore fixed in 4e394dcc3; the feature-router stale read-modify-write is pre-existing and filed as #1826. Round 3 CLEAN; merging.

PR #1830, opened 2026-09-04
fix(redaction): treat authkey and maxar_api_key as sensitive query params; refresh two service-auth docstrings

ishiland:
## Summary

Lane B3 + issue #1755 items 7 and 15 — a small, self-contained cleanup with no behavior change beyond the redaction addition.

- **Lane B3**: adds `authkey` (ArcGIS) and `maxar_api_key` (Maxar) to `SENSITIVE_QUERY_PARAMS` in `backend/app/core/url_redaction.py`. Both are named query parameters distinct from the generic `api_key`/`token` entries already there, so a stored service pointer or a GDAL-composed query carrying either was not being redacted.
- **#1755 item 7**: `origin_probe.py`'s module docstring described only the three-value health vocabulary and the generic 401/403 split. It now also states that `probe_arcgis_origin` reads a provider error body (`error.code` 498/499) rather than the status line — that branch was added by #1746 but the module docstring was never updated to say so.
- **#1755 item 15**: `_require_service_token_if_marked`'s docstring in `router_refresh.py` said the ArcGIS probe reads the layer's `?f=json`. Since #1754 round 6 it reads `<layer>/query` via `build_arcgis_count_query_url` — verified against the current `probe_arcgis_origin` implementation before editing. Docstring only, no code change.

`router_refresh.py` grew by 2 lines from the docstring fix, so the `_MODULE_LOC_CAPS` entry in `backend/tests/test_layering.py` is raised 1374 -> 1376 in the same commit, per repo convention.

Closes nothing; addresses #1755 items 7 and 15.

## Test plan

- [x] `cd backend && uv run ruff check . && uv run ruff format --check .`
- [x] `uv run pytest tests/test_url_redaction_sec.py tests/test_layering.py tests/test_service_auth_transport_1746.py tests/test_source_health_1222.py tests/test_refresh_gate_1269.py tests/test_refresh_observability_1268.py tests/test_service_refresh_1220.py tests/test_postgis_refresh_1265.py tests/test_stac_refresh_1266.py tests/test_services_wfs_pure.py tests/test_vrt_lifecycle_188.py -q` — 985 passed
- [x] New parametrised tests: one case per new sensitive name (`authkey`, `maxar_api_key`) plus a mixed/upper-case spelling of each (matching is case-insensitive), plus a positive control that an unrelated `keyword` param is untouched

### DIFF
--- a/backend/app/core/url_redaction.py
+++ b/backend/app/core/url_redaction.py
@@ -51,9 +51,17 @@ SENSITIVE_QUERY_PARAMS = frozenset(
         "api-key",
         "api_key",
         "apikey",
+        # fix(#1755 item 7 lane B3): ArcGIS's own query-parameter name for a
+        # token, distinct from the generic "api_key"/"token" already here —
+        # a stored ArcGIS pointer or a GDAL-composed query can carry it and
+        # it needs the same redaction the others already get.
+        "authkey",
         "client_secret",
         "code",
         "key",
+        # fix(#1755 item 7 lane B3): Maxar's named API key query parameter,
+        # for the same reason as "authkey" above.
+        "maxar_api_key",
         "password",
         "refresh_token",
         "sig",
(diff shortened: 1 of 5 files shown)


ishiland:
@codex review

PR #1833, opened 2026-09-04
fix(#1755): coded preview refusals, redacted defer cause, guarded finally cleanups

ishiland:
Addresses items 9, 10, and 11 from #1755 (follow-ups from the #1746 authenticated-service import wave).

## Item 9 — `preview_service_layer`'s broad `except Exception` could turn a typed refusal into a 500

`run_service_preview` and its callees (`preview.py`, `service_endpoints.py::assert_endpoints_stay_on_origin`, `service_items.py::materialise_oapif_items`) already convert every typed refusal into an `HTTPException` at their own raise sites today, but that safety depended entirely on those internal contracts holding. `preview_service_layer`'s catch-all had only one `except HTTPException: raise` clause (from #1752) standing between any OTHER exception class and a 500.

Both calls to `run_service_preview` (the initial attempt and the WFS-namespace retry) now go through a new `_run_service_preview_or_refuse` helper, which maps every typed refusal to its coded 4xx via `_preview_refusal_response` before the broad catch-all is ever reached — so a future edit that breaks one of those "raises only HTTPException" contracts degrades to a coded 4xx instead of a silent 500.

| Exception class | Status | Notes |
|---|---|---|
| `CrossOriginEndpointError` | 422 | `{code, message, field}` — `.policy` is a fixed per-class string (same shape `/probe` already returns) |
| `EndpointCheckFailedError` | 422 | same shape; also covers `ItemFetchFailedError` (subclasses it) |
| `SSRFError` | 400 | fixed policy phrase — `SSRFResolutionError`'s message can carry a redirect-chosen hostname, never echoed (mirrors the `/probe` fix from #1770 round 49 P3) |
| `HrefTooLongError` | 422 | fixed message; defence in depth — already pre-wrapped into the two classes above at every current raise site |
| `ValueError` (fixed-policy, e.g. `build_credential_header`) | 400 | fixed message |
| `IngestionError`, `HTTPException` | unchanged | pass through untouched — used for the namespace-retry signal and already-finished answers respectively |

One endpoint test per class in `test_services_endpoints.py::TestPreviewEndpoint`, each mocking `run_service_preview` directly (bypassing the internal wrapping the same way a regression would). Verified as a counterfactual: disabling the `CrossOriginEndpointError`/`EndpointCheckFailedError` branch locally turned those three tests back into 500s.

## Item 10 — `DeferFailed`'s 503 body discarded the cause

`defer_with_orphan_guard` raised `DeferFailed(rolled_back=...) from stamp_exc`/`from defer_exc`, but the 503 body was a fixed string with no way to tell a bug in the dispatching closure (stamp-write failure, or a `TypeError` inside the caller's `defer_call`) from a genuine Procrastinate/queue outage.

`DeferFailed` now takes a `cause: BaseException` and carries `cause_class = type(cause).__name__` — the exception's TYPE name only, never its message — both as an attribute and as a coded field in the 503 `detail`. The body stays free of raw text (no `str(cause)` anywhere), while `cause_class` distinguishes the two failure shapes.

Two new tests in `test_defer_orphan_guard.py` cover both shapes (a `defer_call` failure and a dispatch-marker-write failure), asserting the raw exception text never reaches the body while `cause_class` differs correctly between them.

## Item 11 — a `finally`-block cleanup failure could replace the original ingest exception

Both service tasks' `finally` blocks (`ingest_service` in `tasks_vector.py`, `reupload_service` in `tasks_reupload.py`) awaited `stop_ingest_job_heartbeat` and `_drop_attempt_staging_table` bare — an exception from either would silently replace whatever the task's own `try`/`except` was already propagating.

Both now route through a new `_finally_cleanup` helper (one copy per module, since the two task modules deliberately don't import each other) that logs a cleanup failure (redacted via `redact_exception_text`, with `exc_info`) instead of raising it. The #1753 token purge (`purge_token_on_failure`, the decorator wrapping both tasks) still runs in every path, since neither cleanup's `except` re-raises.

Two new DB-backed tests in `test_failed_job_token_purge_1746.py` drive the real tasks (via the fully-decorated `Task` object, with a `JobContext` so `purge_token_on_failure` actually fires) with a controlled ingest failure plus a controlled failure in BOTH cleanup calls, asserting the original exception type still propagates and the queued row's token is purged. Verified as a counterfactual: temporarily removing `_finally_cleanup`'s try/except made the `ingest_service` test fail with the cleanup's `ValueError` instead of the ingest `RuntimeError`.

## Schema/API impact

None. The `/services/preview` 4xx bodies for the newly-mapped classes are new response shapes for cases that previously either already answered correctly (via the internal wrapping) or would have been an undifferentiated 500. `DeferFailed`'s `detail` changes from a bare string to a small dict (adds `code`/`cause_class`, keeps `message`); no caller reads `.detail` today (checked `admin/router.py`'s only consumer, which reads `.rolled_back`).

## LOC caps

`_MODULE_LOC_CAPS` raised for the three touched capped files, each with a comment on what the lines bought: `sources/router.py` 1639→1780, `tasks_vector.py` 1356→1392, `tasks_reupload.py` 1297→1333.

## Verification

```
cd backend && uv run pytest tests/test_layering.py tests/test_rule1_structural.py tests/test_rule2_structural.py -q
cd backend && uv run pytest tests/test_services_endpoints.py tests/test_defer_orphan_guard.py tests/test_commit_attempted_marker_doors.py tests/test_failed_job_token_purge_1746.py -q
cd backend && uv run pytest tests/test_door_token_policy_1746.py tests/test_service_auth_contract_1746.py tests/test_service_auth_transport_1746.py tests/test_preview_token_sec021.py tests/test_service_refresh_1220.py tests/test_reupload_idor.py tests/test_reupload.py -q
```

All green (233 + 586 tests respectively across the two broader sweeps).

Refs #1755

### DIFF
--- a/backend/app/modules/catalog/sources/router.py
+++ b/backend/app/modules/catalog/sources/router.py
@@ -90,6 +90,7 @@ from app.platform.service_endpoints import (
     OGC_JSON_ACCEPT,
     CrossOriginEndpointError,
     EndpointCheckFailedError,
+    HrefTooLongError,
     assert_endpoints_stay_on_origin,
 )
 from app.platform.security import (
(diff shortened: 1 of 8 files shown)


ishiland:
@codex review

PR #1834, opened 2026-09-04
feat(import): expose basic and header-key credential methods on the service tab

### LINKED ISSUE #1712: fix(import): switching tabs mid-import strands the job on the Upload, Service and STAC tabs

### LINKED ISSUE #1832: Upload tab: batch intermittently stranded after a mid-upload tab switch (upload.spec flakes ~30%, screenshot shows empty dropzone)
Found during the 2026-09-04 post-wave smoke on the rebuilt dev stack (main at e99d1ef12).

`e2e/upload.spec.ts:51` "a tab switch mid-upload does not strand the batch" fails intermittently: 1 of 234 in the full run, then 1 of 2 and 1 of 4 when repeated alone (3 failures in 10 runs), while three single smoke runs the same hour passed. The failure screenshot shows the Upload tab back in its empty dropzone state after STAC → Upload, with no batch chip for `sample.geojson`; the assertion that times out is `getByRole('button', { name: /sample\.geojson/i })` (30 s).

So the stranding #1712 describes still happens on the Upload tab intermittently: the upload keeps running server-side while the tab is hidden, but on remount the resumed batch is not shown, and whatever the preview created is unreachable from the UI. This is a product race, not only a test flake: #1708's keep-alive for the URL tab was applied to the Upload tab, but there is a window (likely the upload/preview response landing between unmount and the store rehydration on remount) where the batch is dropped.

Repro on the host: `npx playwright test e2e/upload.spec.ts:51 --project=chromium --repeat-each=6` (stack running). Screenshot at `test-results/upload-Upload-Flow-a-tab-s-*-repeat1/test-failed-1.png`.

Scope: `frontend/src/pages/ImportPage.tsx` / the Upload form's unmount handling; the same mechanism as #1712 (Service tab), so whichever lane fixes #1712 should reuse the store-backed keep-alive here and add a race test that resolves the upload response after unmount.

ishiland:
## Summary

Lane B4 of the service-auth wave.

- **Method select (plan 3.1/3.8).** `ServiceUrlForm`'s Service tab and `SourceRefreshAction`'s refresh dialog offered only a bearer token for WFS/OGC API Features services, with help text claiming Basic auth and custom headers "aren't supported." Backend B2b (#1770, already merged) accepts all four `CredentialMethod` values, so that claim is now false. Both surfaces get a four-way select: none / bearer token / username+password / API key in a header. Switching methods clears the other branches' fields (mirrors the backend's `auth` object oneOf rule). ArcGIS-shaped URLs are untouched — ArcGIS stays bearer-only (the token is percent-encoded into the query; `build_credential_header` never composes a header for it) and keeps its existing three-way sign-in/token/none select.
- **New i18n keys** in all four locales (`en/es/fr/de`) for `import.json` (`serviceUrl.credentialMethod*`, `credentialToken*`, `headerName*`, `headerValue*`, `credentialHelpText`) and `dataset.json` (`sourcePanel.refresh.credential.service.*`). The header-name field is free text with examples in the help copy (`X-API-Key`, `Ocp-Apim-Subscription-Key`, `key`) rather than a single hardcoded placeholder, since providers differ. No persistence promise is restated (ADR-002 invariant 4); no example password anywhere (Rule 3); the verb is "refresh," never "sync."
- **`fix(#1712)`: Service-tab keep-alive.** The Service tab had none — a tab switch mid-preview or mid-commit unmounted `ServiceUrlForm` and stranded the `pending` IngestJob the server keeps creating regardless. Adds `frontend/src/api/service-url-session.ts`, mirroring the URL tab's existing `url-import-session.ts` (#1708): the preview and commit promises live at module scope, a mount adopts an in-flight or already-settled session instead of starting a duplicate, and the response is kept, not aborted.
- **`fix(#1832)`: Upload-tab race, found and fixed alongside #1712 (same mechanism, same lane).** `UploadForm`'s existing #1712 keep-alive had a gap: a drop made while the upload-config query was still fetching sat in a plain `useState` queue, and a tab switch that unmounted the form before that query settled discarded the queue before anything reached the module-scoped upload-session — so a remount found nothing to adopt. Moved that queue into `upload-session.ts` (`queuePendingUploadFiles`/`peekPendingUploadFiles`/`clearPendingUploadFiles`) so a remount rehydrates it. This reproduced the intermittent (~30%) failure on `e2e/upload.spec.ts:51` reported in #1832.

**Other Import tabs are unchanged.** STAC shares the same #1712 exposure class but needs its own session shape (not folded into this PR); Register commits synchronously, so losing its confirmation UI on a tab switch is a UX papercut, not a stranded job.

## Screenshot

Verified visually via `e2e/service-import-credential-methods.spec.ts` (Username/Password → Probe → layer select → preview reached with the Basic credential). Screenshot available on request — not attached here.

## Verification

```
cd frontend
npm run lint
npm run typecheck
npm run test:i18n
npx vitest run src/components/import/__tests__ src/components/dataset/__tests__ src/api
npx vitest run   # full suite, 5934 tests, all green
```

E2E, against Vite on the host at `:5174` with `API_PROXY_TARGET=http://localhost:8001` (the worktree-frontend recipe in AGENTS.md — the shared `:8080` stack serves `main`, not this branch):

```
E2E_ALLOW_WORKTREE=1 E2E_BASE_URL=http://localhost:5174 npx playwright test \
  e2e/service-import-credential-methods.spec.ts \
  e2e/dataset-refresh-credentials.spec.ts \
  e2e/service-import-arcgis-signin.spec.ts \
  e2e/upload.spec.ts \
  --project=chromium
# 9 + 8 passed across runs, no regressions
```

`#1832`'s specific repro, run against the real backend at `:8001` through the fix:

```
E2E_ALLOW_WORKTREE=1 E2E_BASE_URL=http://localhost:5174 npx playwright test \
  "e2e/upload.spec.ts:51" --project=chromium --repeat-each=8
# 8/8 passed (previously ~30% flake rate per #1832)
```

Fixes #1712
Fixes #1832


ishiland:
@codex review

ishiland:
@codex review

ishiland:
Round 1 P2s fixed in 1c79000be (session restores url/shape/auth; refresh credential retained until success or close). Round 2 P2 (restore the minted token's expiry with the session) filed as https://github.com/geolens-io/geolens/issues/1835; merging on no-P1.

PR #1838, opened 2026-09-04
fix(tests): freeze the router clock in the degraded-S3 URL-import tests (#1808)

### LINKED ISSUE #1808: flake(tests): test_url_import_1705 degraded-S3 cleanup case returns 502 instead of 413
Observed on PR #1796 CI run 33776039613 (2026-09-03): `tests/test_url_import_1705.py::TestUrlImportDegradedS3Failure::test_non_timeout_failure_bounds_its_cleanup` failed with `assert 502 == 413`, then passed on `gh run rerun --failed` with no code change. #1796 did not touch URL import. The degraded-S3 cleanup bound is racing something, most likely the staging copy failing before the size check runs under `-n 4`. Find the ordering that lets the 502 win, pin it deterministically, and remove the race.

Part of the 2026-08-30 audit follow-ups tracked in #1778.

ishiland:
## Root cause

Both tests in `TestUrlImportDegradedS3Failure` shrink the joint stage budget to 1-2s so their bounded waits resolve fast, but every budget check in `upload_from_url` measures a real `time.monotonic()` deadline set once at request start. Any stall between that line and the check a test means to exercise — a slow pool checkout, a loaded runner under `pytest -n 4` — spends the whole budget, and the handler answers a budget refusal instead: `_remaining_fetch_budget` raises 502 before the fetch, or `_stage_put_bounded` raises `_StagePutAbandoned` (also 502) without ever creating the put task. In CI that surfaced as `assert 502 == 413` on `test_non_timeout_failure_bounds_its_cleanup`; the sibling `test_abandoned_put_hands_cleanup_to_the_reaper` has the same exposure on its 1s budget, where the reaper it asserts on would never be attached.

## The fix

Test-only. A `_freeze_router_clock` helper pins the `time` name bound in `app.processing.ingest.router`'s own module namespace to one instant, applied in both tests before the budget is derived, so every budget check sees the same instant and the ordering is the one each test states. The real `time` module is untouched — asyncio's timers read it directly — so `asyncio.wait(..., timeout=...)` still waits real wall-clock time and the bounded waits still genuinely time out, which is the behaviour under test.

No product-code change. The handler's ordering is correct; at the production budget (~510s) the pre-fetch work cannot exhaust it, and the 502 is the right answer when it genuinely does.

## Reproduction

The flake did not reproduce on a 16-core host: `tests/test_url_import_1705.py -n 4` was 0/10 failures, and 0/3 at `-n 8` with 12 CPU hogs, 0/2 at `-n 12` with 48. So the mechanism was pinned instead, with a scratch copy of the test carrying a 2.5s stall injected ahead of the fetch (standing in for CI contention), parametrized over the freeze:

- before: 502, `"Not enough time remained in the request budget to download this file"` — the exact CI symptom, 1/1
- after (same stall, `_freeze_router_clock` applied): 413, 0/1

Post-fix: `tests/test_url_import_1705.py -n 4` 10/10 clean (81 passed each), `ruff check` / `ruff format --check` clean.

Closes #1808

### DIFF
--- a/backend/tests/test_url_import_1705.py
+++ b/backend/tests/test_url_import_1705.py
@@ -17,6 +17,7 @@ import inspect
 import time
 import uuid
 from pathlib import Path
+from types import SimpleNamespace
 from unittest.mock import AsyncMock
 
 import httpx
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review

PR #1841, opened 2026-09-04
fix(builder): explain the refused public toggle when a map holds private datasets (#1831)

### LINKED ISSUE #1831: Share panel: making a map public with a private layer fails silently with 400 while the confirm copy says it will work
Found during the 2026-09-04 post-wave browser smoke on the dev stack (main at e99d1ef12 + follow-ups). Pre-existing: the backend rule dates from the initial release and the dialog copy from #430.

**Steps**
1. Create a map, add a layer whose dataset is private (any freshly imported dataset defaults to private).
2. Builder → Share → choose "Anyone with the link".
3. The confirm dialog says "Make this map public? … Hidden from viewers: <layer>. These layers' datasets are not visible to this map's audience, so they will not appear for anyone but you." and offers "Make public".
4. Click "Make public".

**Observed**
`PUT /api/maps/{id}` returns `400 {"message": "Cannot set visibility to public: map contains non-public datasets", "datasets": "Large Lakes"}`. The radio snaps back to "Only you" and nothing is shown to the user; the only trace is a console error. The confirm copy promises a behaviour (public map, hidden private layers) that the backend refuses.

**Expected**
Either (a) the confirm dialog reads the same rule the backend enforces and says the map cannot be made public until those datasets are shared, disabling "Make public" or offering to change the datasets' visibility, or (b) the backend allows it and hides the layers as the copy says. Whichever is chosen, a refused PUT must surface as an inline error in `components/builder/SharePanel.tsx`, keyed off `detail.message` / `detail.datasets`, not vanish.

Related copy key: `builder.audienceHiddenWarning`. Backend rule: the visibility check in the maps update handler ("contains non-public datasets").

ishiland:
## Root cause

`PUT /api/maps/{id}` refuses to switch a map to public visibility while it still holds non-public dataset layers, returning `400 {"message": "Cannot set visibility to public: map contains non-public datasets", "datasets": "<names>"}` (`backend/app/modules/catalog/maps/router.py`, `validate_public_visibility`). `SharePanel.tsx`'s `handleVisibilityChange` tried to parse those dataset names out of `err.message`, but `ApiError.message` is already a *translated fallback string* (built by `translateApiErrorDetail`), not JSON — so `JSON.parse(err.message)` always failed and fell back to a `share.cannotPublish` toast whose `{{datasets}}` value was just that generic fallback text, not the actual dataset names. The toggle itself never "snapped back" (the mutation only applies the new visibility on success), so the only visible trace of the refusal was a console error.

## Fix

`handleVisibilityChange` now reads the structured detail off `err.body` (the raw object FastAPI sent, preserved by `apiFetch`) instead of `err.message`, and renders it as a persistent inline message under the visibility control — the confirm `AlertDialog` is already closed by the time the response arrives, so a toast alone isn't enough. It reuses the existing `share.cannotPublish` locale string (already present, correctly interpolated, in all four locales) rather than inventing a new key. Also clears the message on a fresh attempt and when the Share dialog is reopened, since the component stays mounted across open/close.

## Tests

Added `describe('#1831 publish blocked by non-public datasets', ...)` to `frontend/src/components/builder/__tests__/SharePanel.test.tsx`: asserts the inline message names the dataset on a 400 refusal, that it persists after the confirm dialog closes, and that it doesn't appear after a successful visibility change. The private-visibility radio stays `aria-checked` on failure.

## Gates

- `npm run lint` — pass
- `npm run typecheck` — pass
- `npm run test:i18n` — pass
- `npx vitest run src/components/builder` — 148 files / 2509 tests pass (includes `src/components/builder/__tests__/SharePanel.test.tsx`, 68 tests)

Playwright not run against the shared dev stack from this worktree (stack serves `main`, per AGENTS.md).

Closes #1831

### DIFF
--- a/frontend/src/api/maps.ts
+++ b/frontend/src/api/maps.ts
@@ -283,6 +283,34 @@ export async function publishMap(id: string, visibility: 'public' | 'private' | 
   });
 }
 
+/**
+ * fix(#1831 review): distinguishes the one 400 `publishMap` can return that
+ * has its own dedicated UI (SharePanel's inline "can't go public" message)
+ * from every other failure, which still gets the generic toast. Reads the
+ * raw `err.body` the backend sent (`{ message, datasets }` from
+ * `validate_public_visibility` in `maps/router.py`), not `err.message` —
+ * that's already a translated fallback string, not JSON, and was the actual
+ * bug in #1831 (dataset names were read off it and silently dropped).
+ *
+ * Shared by `usePublishMap` (to skip its own toast for this one case) and
+ * `SharePanel` (to render the inline message), so the two can't drift.
+ */
+export function nonPublicDatasetsFromError(err: unknown): string | undefined {
+  if (!(err instanceof ApiError) || err.status !== 400 || !err.body || typeof err.body !== 'object') {
+    return undefined;
+  }
+  const detail = err.body as { message?: unknown; datasets?: unknown };
+  if (
+    typeof detail.message === 'string' &&
+    detail.message.includes('non-public datasets') &&
+    typeof detail.datasets === 'string' &&
+    detail.datasets.length > 0
+  ) {
+    return detail.datasets;
+  }
+  return undefined;
+}
+
 export async function createShareToken(
   mapId: string,
   expiration: ShareExpirationInput = {},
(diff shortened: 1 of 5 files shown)


ishiland:
@codex review

Issue #1859, opened 2026-09-05
flake(ci): de-flake wall-clock races behind merge-queue evictions


ishiland:
CI health report 2026-09-04 found several backend tests whose assertions race a
hard-coded wall-clock deadline against a loaded CI runner, each confirmed
fail-on-attempt-1 / pass-on-attempt-2 on the same sha in the 14 days before the
report:

- tests/test_arcgis_signin.py::test_a_discovery_failure_takes_no_lock_and_writes_no_ledger_row
- tests/test_ingest_progress.py::test_service_worker_advances_ogr2ogr_progress_while_remote_import_is_running
- tests/test_export_artifact_cache_1532.py (both timing assertions)
- tests/test_url_import_1705.py::TestUrlImportDegradedS3Failure::test_non_timeout_failure_bounds_its_cleanup
- tests/test_auth_refresh_cookies_1302.py::TestGraceWindowCannotRatchetOpen::test_concurrent_rotations_do_not_extend_the_retirement_cutoff
- tests/test_embedding_backfill_queue_1542.py::test_enqueue_returns_promptly_instead_of_waiting_for_the_run
- tests/test_tile_cache.py: pytest-asyncio/anyio marker mix under anyio_mode=auto, plus an unclosed FakeRedis client

Also: .github/workflows/ci.yml downloads the gitleaks binary with a bare curl -sSL
and no retry, so a transient download failure evicts a merge-queue entry.

Each merge_group Backend Tests failure re-runs the ~27 minute required job and can
evict an unrelated PR from the queue. Tracked here so the fix commits have a stable
reference.

ishiland:
Also: the postgis/postgis:18-3.6 base-image pull inside the Start PostgreSQL with PostGIS + pgvector docker build step (cli-test, backend-test, ai-evals) hit a transient Docker Hub token-fetch reset and evicted #1865's own merge-queue run (33966364665), and the same failure hit #1858's Backend Tests at 11:47 UTC. Fixed in #1865 with a 5-attempt retry around the docker build call.

ishiland:
All items landed: #1862, #1863 and the last one (the probe classification timing bound) in #1876, merged 2026-09-05.

PR #1865, opened 2026-09-05
test(ci): remove wall-clock races behind the merge-queue evictions

ishiland:
De-flakes the tests behind the CI health report's merge-queue-eviction list (2026-09-04), tracked in #1859. Test-only, plus two fixes in the CI workflow. No production code changes, no schema/API/config impact.

For each test: read what it protects, then replace the wall-clock or sleep-based assertion with a deterministic one, or confirm the flake no longer reproduces because an earlier PR already fixed it.

## Fixed

`test_ingest_progress.py::test_service_worker_advances_ogr2ogr_progress_while_remote_import_is_running` polled the job row on a 1-second wall-clock deadline, opening a fresh `db_module.async_session()` every 10ms to match the heartbeat's own mocked interval. Under a loaded runner that session churn produced `ConnectionError: unexpected connection_lost() call`, which had nothing to do with the ingest logic. It now waits on an event the heartbeat tick sets when it completes, then reads the job row once. 20/20 on a repeat loop.

`test_export_artifact_cache_1532.py::test_the_artifact_is_stamped_with_the_snapshot_not_the_upload` asserted `artifact.built_at <= before + 1.0`. That 1.0s margin assumed request overhead ahead of the conversion (auth, dataset fetch, precondition checks) stays under a second, and a loaded runner can push past that without the property under test actually breaking. It now records the conversion double's own start time and compares `built_at` against that directly. 20/20 on a repeat loop.

`test_tile_cache.py` had 19 tests marked `@pytest.mark.asyncio`, which hands them to pytest-asyncio's own per-function event loop instead of the anyio-owned path the rest of the suite uses. `anyio_mode` is `"auto"` project-wide, and pyproject.toml's own comment reserves `asyncio_mode = "strict"` for an explicit asyncio-only slice, which this file isn't (`test_tile_cache_generation_key_1429.py` tests the same production module and uses no marker at all). Removed the markers. Also closed the `tile_cache` fixture's fakeredis client, which held a connection pool nothing ever released. Full file green (38/38). 19 of 20 planned `-n 4` repeat runs finished clean (722 passes, 0 failures) before the local time budget on this box ran out.

`.github/workflows/ci.yml`: the gitleaks download used a bare `curl -sSL` with no retry, so a transient GitHub Releases hiccup failed a required job in the merge_group critical path. Added `--retry 5 --retry-all-errors --retry-delay 2`, matching the retry already used for the mcp-publisher download in publish-mcp.yml.

`.github/workflows/ci.yml`: the Start PostgreSQL with PostGIS + pgvector step in cli-test, backend-test, and ai-evals builds a custom image `FROM postgis/postgis:18-3.6`. This PR's own merge-queue run (33966364665) was evicted when that pull hit `failed to fetch oauth token: ... connection reset by peer`, and the same failure hit #1858's Backend Tests the same day. Wrapped the `docker build` call in a 5-attempt retry with a 10s delay between attempts, in all three jobs. Verified the retry construct against the actual YAML-dedented script with `bash -n`, then a fake `docker` binary that fails twice and succeeds on the third call, then one that fails all five, both under `bash -eo pipefail` (the default shell for `run:` steps).

The backup-roundtrip job also starts `postgis/postgis:18-3.6`, but through a native GitHub Actions `services:` container, not a `docker build` step. Actions pulls service-container images during job setup, before any of our steps run, and the `services:` syntax has no retry option, so this fix doesn't reach it. Left alone; that job wasn't the one reported failing.

## Already fixed, left as is

Verified 20/20 on a repeat loop, no code change needed:

- `test_arcgis_signin.py::test_a_discovery_failure_takes_no_lock_and_writes_no_ledger_row`: fixed by #1786 (merged 2026-09-02), which switched from watching a process-global session factory to watching the lock helper the handler actually calls.
- `test_url_import_1705.py::TestUrlImportDegradedS3Failure::test_non_timeout_failure_bounds_its_cleanup`: fixed by #1838/#1808 (merged 2026-09-04), which freezes the router clock before the budget is derived.
- `test_auth_refresh_cookies_1302.py::TestGraceWindowCannotRatchetOpen::test_concurrent_rotations_do_not_extend_the_retirement_cutoff`: fixed by #1634, which derives the race tolerance from the stall it's meant to catch instead of a literal 1s.
- `test_embedding_backfill_queue_1542.py::test_enqueue_returns_promptly_instead_of_waiting_for_the_run`: fixed by #1778, which replaced an `elapsed < 1.0` check with an event set at the top of the mocked backfill.
- `test_export_artifact_cache_1532.py::test_two_gpkg_conversions_of_unchanged_data_are_byte_identical`: the underlying cause, SQLite's change-counter pair bumping by transaction count rather than content and so diverging between two builds of identical data under CI load, was fixed by #1641 on 2026-08-24, inside the report's own 14-day window. Also passed 3/3 under `-n 4` alongside its sibling counter tests.

Left out per the brief: `test_probe_classification.py`'s `elapsed_ms < 100` bound is also a flake, but lane SEC6 owns the probe files right now.

## Verification

From the worktree, Postgres at localhost:5434:

- `uv run pytest tests/test_ingest_progress.py tests/test_export_artifact_cache_1532.py tests/test_tile_cache.py -q`: 179 passed
- `uv run pytest tests/test_layering.py -q`: 50 passed
- `uv run ruff check . && uv run ruff format --check .`: clean across the full backend tree
- `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/ci.yml'))"`: parses
- Each test item above was individually repeated 20x (19/20 for `test_tile_cache.py` under `-n 4`, which ran out of local wall-clock budget rather than failing) per the brief's method.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -512,10 +512,13 @@ jobs:
         run: node frontend/scripts/check-api-error-boundary.mjs
 
       - name: Install gitleaks
+        # fix(#1859): retries a transient download failure instead of letting
+        # it evict the whole merge-queue entry on a required job.
         run: |
           GL=8.30.1
           GL_SHA256=551f6fc83ea457d62a0d98237cbad105af8d557003051f41f3e7ca7b3f2470eb
-          curl -sSL "https://github.com/gitleaks/gitleaks/releases/download/v${GL}/gitleaks_${GL}_linux_x64.tar.gz" \
+          curl -sSL --retry 5 --retry-all-errors --retry-delay 2 \
+            "https://github.com/gitleaks/gitleaks/releases/download/v${GL}/gitleaks_${GL}_linux_x64.tar.gz" \
             -o gitleaks.tar.gz
           echo "${GL_SHA256}  gitleaks.tar.gz" | sha256sum --check --strict
           tar -xz -f gitleaks.tar.gz gitleaks
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1197,12 +1197,23 @@ jobs:
         # apt steps, so this also gets a bound.
         timeout-minutes: 10
         run: |
-          docker build -t geolens-cli-test-db -f - . <<'DOCKERFILE'
+          # fix(#1859): the base-image pull inside `docker build` hit a
+          # transient Docker Hub token-fetch reset ("failed to fetch oauth
+          # token: ... connection reset by peer") in merge-group run
+          # 33966364665 and again on #1858's Backend Tests, evicting an
+          # otherwise-clean queue entry. Retry the whole build, since that is
+          # where the pull happens.
+          for attempt in 1 2 3 4 5; do
+          docker build -t geolens-cli-test-db -f - . <<'DOCKERFILE' && break
           FROM postgis/postgis:18-3.6
           RUN apt-get update \
               && apt-get install -y --no-install-recommends postgresql-18-pgvector \
               && rm -rf /var/lib/apt/lists/*
           DOCKERFILE
+          [ "$attempt" -eq 5 ] && exit 1
+          echo "docker build failed (attempt $attempt/5), retrying in 10s..."
+          sleep 10
+          done
           docker run -d --name cli-test-postgres \
             -e POSTGRES_DB=postgres \
             -e POSTGRES_USER=geolens \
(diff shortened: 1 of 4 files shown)


ishiland:
@codex review

ishiland:
@codex review

PR #1872, opened 2026-09-05
docs(agents): keep inline review comments to the anchor and the invariant

ishiland:
Amends the inline review-comment convention in AGENTS.md: a comment carries the anchor and the invariant in at most three lines; review history, rejected alternatives and narrative go in the PR body or the anchored issue; docstrings state the contract; no comment restates the next line; a comment you touch is trimmed to the rule.

Why now: measured on main at b7478cc7a, comments and docstrings are 35% of the backend's non-blank lines and about 47% of its characters, and 46% of inline comment lines sit in blocks of eight or more consecutive lines. A dozen files are more than half prose. That is paid on every file read by a reviewer, a lane or codex, and it was the substance of a public review of the codebase on 2026-09-05. This PR changes the rule so the volume stops growing; a measured cleanup of the largest files is a separate, later change so no behaviour-free diff sits in the queue this week.

No code change. Verification: the file renders; the example line is unchanged.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -66,11 +66,13 @@ Frontend code uses TypeScript, React, ESLint, React Hooks rules, and JSX accessi
 
 ### Inline review-comment convention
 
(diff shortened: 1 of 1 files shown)


ishiland:
@codex review

Issue #1870, opened 2026-09-05
fix(export): where filter rejects any string value containing a word as an unknown column


ishiland:
Reported by a reviewer on the launch thread (2026-09-05) and reproduced on main.

`validate_where_clause` (`backend/app/processing/export/service.py`) runs an identifier regex over the raw WHERE string after the AST gate, and the regex does not skip string literals. Any value containing a word that is not a column name or SQL keyword is rejected as an unknown column:

```
name = 'Main Street'                -> ValueError: Unknown column: Main
height > 10 AND name LIKE 'Water%'  -> ValueError: Unknown column: Water
name = 'O''Brien Park'              -> ValueError: Unknown column: O
```

Every export door with a `where` parameter is affected: `export/router.py` (two sites), `export/service.py`, and `export/parquet.py`, so the CLI and SDK `--where` filters fail the same way. Filtering by any text value is effectively unusable unless the value happens to be a column name or keyword.

Fix: drop the regex pass and take the column names from the AST that `validate_where_ast` already builds (`exp.Column` nodes), checking them against `column_info`; the pre-parse checks for `;`, `--` and `/* */` and the AST allowlist stay. Add tests for the three inputs above plus a real unknown column still refused, and keep the existing SEC-S09 tests green.

ishiland:
Fixed in #1878, merged 2026-09-05: string literals in a where filter are masked before the identifier walk (both quote kinds scanned; only single-quoted runs blanked).

PR #1878, opened 2026-09-05
fix(export): stop reading string values in a where filter as column names

ishiland:
Fixes the defect in #1870: `validate_where_clause` ran its identifier walk over the raw WHERE fragment with quoted literals included, so any string value containing a word was refused as an unknown column.

```
name = 'Main Street'                -> ValueError: Unknown column: Main
height > 10 AND name LIKE 'Water%'  -> ValueError: Unknown column: Water
name = 'O''Brien Park'              -> ValueError: Unknown column: O
```

Reproduced on `origin/main` at 95a80e88d before changing anything.

## What changed

`mask_quoted_literals` in `backend/app/processing/export/where_validator.py` blanks every single-quoted literal, replacing it with a run of spaces of the same width. `validate_where_clause` runs its walk over the masked clause instead of the raw one. The mask preserves length so two identifiers separated by a literal cannot fuse into a third: `na'x'me` becomes `na   me`, two unknown identifiers, not the column `name`.

The identifier walk stays. It is defense in depth against a sqlglot parser gap, and the issue's suggested alternative (take the column names from the AST) would make the walk depend on the same parse it exists to back up.

A quote inside a double-quoted identifier is not a literal, and the mask scans both quote kinds in one pass so it never treats one as opening a value. An unknown column before, after, or between literals is still refused, and the error now names the real unknown column rather than the first word inside a value. The pre-parse rejection of `;`, `--` and `/* */` still reads the raw clause, so those are refused wherever they appear, inside a literal included. The AST allowlist still runs first, before any of this.

## Sibling sites

All four export doors with a `where` parameter (`export/router.py` twice, `export/service.py`, `export/parquet.py`) call `validate_where_clause`, so the single helper covers the class, and the CLI and SDK `--where` flags with it.

The other two identifier walks in `backend/app/` validate a different grammar and are left alone:

- `core/db/tenant_session.py` already masks literals and comments through its own `_mask_sql_noncode` state machine before classifying tokens. Its input is generated SQL, not a caller's fragment, and it needs the quoted-identifier masking that this helper deliberately does not do.
- `platform/sandbox/validator.py` walks the rendered target of a cast against a blocklist of OID-alias type names. That string is a type expression, never a clause carrying values, and masking there would only weaken a blocklist.

`modules/catalog/datasets/domain/column_stats.py` shares the name `_IDENTIFIER_RE` but anchors it over a single column name, not a clause.

## Left alone

A `;` or `--` inside a string value is still refused. A double-quoted identifier containing a space is still refused too, because the walk reads its parts as separate identifiers; that predates this branch and is unchanged by it. Narrowing those three checks to code-only would need the mask to run before them, which changes what the string-level layer promises, and that is a separate change with its own review. No CHANGELOG entry: the release-time pass collects these.

## Schema, API, config

None. No request or response schema moved, so `backend/openapi.json` and the SDKs are untouched. No migration, no new config, no new translation key.

## Verification

Run from the worktree against Postgres at localhost:5434.

```
uv run pytest tests/test_export.py tests/test_export_hardening.py \
  tests/test_export_where_validator.py tests/test_export_access.py \
  tests/test_export_head_1513.py -q                                  165 passed, 3 skipped
uv run pytest tests/test_export_parquet.py tests/test_export_parquet_writer.py \
  tests/test_export_parquet_row_stream_budget_1778.py \
  tests/test_export_antimeridian.py tests/test_export_artifact_cache_1532.py -q
                                                                     176 passed
  tests/test_rule2_structural.py -q                                  159 passed
uv run ruff check .                                                  All checks passed
uv run ruff format --check .                                         1167 files already formatted
```

Counterfactual, run for real: reverting only `export/service.py` to `origin/main` and keeping the tests fails 8 of the new unit cases and all 4 of the new endpoint cases.

```
FAILED TestWhereClauseStringLiterals::test_literal_containing_a_space_accepted
FAILED TestWhereClauseStringLiterals::test_literal_in_like_pattern_accepted
FAILED TestWhereClauseStringLiterals::test_literal_equal_to_a_sql_keyword_accepted
FAILED TestWhereClauseStringLiterals::test_literal_naming_a_foreign_column_accepted
FAILED TestWhereClauseStringLiterals::test_escaped_quote_in_literal_accepted
FAILED TestWhereClauseStringLiterals::test_in_list_of_literals_accepted
FAILED TestWhereClauseStringLiterals::test_unknown_column_after_a_literal_still_refused
FAILED TestWhereClauseStringLiterals::test_unknown_column_between_two_literals_still_refused
FAILED TestExportWhereWordLiteral::test_word_literal_with_a_space_exports
FAILED TestExportWhereWordLiteral::test_escaped_quote_literal_exports
FAILED TestExportWhereWordLiteral::test_word_literal_selecting_over_cap_413
FAILED TestExportWhereWordLiteral::test_unknown_column_beside_a_literal_still_400
```

The two unknown-column cases fail on main for the right reason: main refuses them naming the literal, so the assertion that the message names `nope` is what breaks.

## Second commit

The first commit's mask scanned for single quotes only, which made it a regression as well as a fix: a quote inside a double-quoted identifier opened a mask run that Postgres opens no literal for, and an unknown column sitting after it was blanked out of the walk and accepted, where main refused it. The scan now covers both quote kinds and blanks a run only when it opens with a single quote, so a double-quoted identifier is consumed by the same pass and handed to the walk unchanged. Dollar-quoting and `E''` are rejected at the AST gate and are not modelled here.

Three tests cover it, two on `validate_where_clause` and one on the helper. Reverting only `export/where_validator.py` to the first commit and keeping the tests fails all three:

```
FAILED TestWhereClauseStringLiterals::test_quote_in_a_double_quoted_identifier_does_not_hide_a_column
FAILED TestWhereClauseStringLiterals::test_lone_quote_in_a_double_quoted_identifier_does_not_hide_a_column
FAILED TestMaskQuotedLiterals::test_quote_inside_a_double_quoted_identifier_opens_no_literal
```

Every test from the first commit keeps its exact expected output; none was edited.

The endpoint tests lower the export cap to 2 rows against a real 3-row table, so the bounded filtered COUNT executes. The 413 case proves the clause is applied rather than dropped, and the 400 case proves an unknown column beside a valid literal is still refused. No LOC cap raised; neither `export/service.py` nor `export/where_validator.py` is in `_MODULE_LOC_CAPS`.

### DIFF
--- a/backend/app/processing/export/service.py
+++ b/backend/app/processing/export/service.py
@@ -13,7 +13,10 @@ from app.core.async_io import run_in_thread_draining
 from app.core.config import settings
 from app.core.csv_safety import numeric_column_names
 from app.processing.export.ogr import FORMAT_MAP, run_ogr2ogr_export
-from app.processing.export.where_validator import validate_where_ast
+from app.processing.export.where_validator import (
+    mask_quoted_literals,
+    validate_where_ast,
+)
 from app.core.runtime.staging import ensure_staging_ready
 
 logger = structlog.stdlib.get_logger(__name__)
@@ -388,7 +391,10 @@ def validate_where_clause(where: str, column_info: list[dict] | None) -> str:
     # aren't in this dataset's column_info, even if they parse cleanly.
     valid_names = {col["name"].lower() for col in column_info}
 
-    identifiers = _IDENTIFIER_RE.findall(where)
+    # fix(#1870): the walk reads the code around the values, never the values.
+    # A quoted literal is never a column reference, so masking it here cannot
+    # admit an unknown column before, after or between literals.
+    identifiers = _IDENTIFIER_RE.findall(mask_quoted_literals(where))
     for ident in identifiers:
         upper = ident.upper()
         if upper in _SQL_KEYWORDS:
--- a/backend/app/processing/export/where_validator.py
+++ b/backend/app/processing/export/where_validator.py
@@ -23,6 +23,8 @@ Allowed WHERE expression types (deny-by-default — anything not in this tuple r
 
 from __future__ import annotations
 
+import re
+
 import sqlglot
 from sqlglot import exp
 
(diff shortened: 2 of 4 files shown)


ishiland:
@codex review

ishiland:
@codex review

PR #1883, opened 2026-09-05
fix(ingest): alembic logger flag and cleanup steps that mask the ingest exception

ishiland:
Two of the remaining items from the #1755 audit list.

## Item 8: migrations disabled every logger already registered

`backend/alembic/env.py` called `fileConfig()` with no `disable_existing_loggers`
argument, so it inherited the stdlib default of True. That sets `.disabled` on
every logger object that existed at the time and is not named in alembic.ini's
`[loggers]` section, and nothing puts it back. `setup_logging()` never touches
that flag, so neither `tests/_logging_state.configured_logging()` nor conftest's
logging guard snapshots it.

In the test suite the session migration fixture runs after collection has already
imported the modules under test, so httpx, httpcore and application module loggers
stayed disabled for the whole session. Three sites had grown a local re-enable
dance around that, and all three are removed here:

- the autouse `_restore_httpx_logger_disabled_flag` fixture in
  `test_1746_stdlib_log_redaction.py`
- both copies of the `target_logger.disabled = False` try/finally in
  `test_maps_style_json.py`

Counterfactual actually run: with the one-line fix reverted and those workarounds
gone, `test_1746_stdlib_log_redaction.py` and `test_maps_style_json.py` report
9 failed, 183 passed. With the fix, 0 failed.

The new `test_1755_alembic_logger_flag.py` has two tests. One reads env.py's AST
and requires the keyword to be the literal False. The other asserts the flag after
the real migration ran, and skips rather than passing vacuously when it did not:
it checks that alembic.ini's logging was applied at all by looking at the alembic
logger's level, and it imports httpx at module scope so the logger exists before
the session fixture runs. Both fail on main.

## Item 11: a cleanup failure replaced the ingest exception

Each ingest task ends in a `finally:` block that stops its heartbeat, drops its
attempt staging table, unlinks scratch files and reaps staging objects. Every call
was awaited bare. A failing cleanup step therefore replaced whatever the
`try`/`except` above was propagating, and skipped every remaining step in the same
block.

`cleanup_step` in `tasks_common.py` is the one helper all seven task modules now
route those steps through. Design notes:

- An async context manager rather than a callable taking a coroutine. One
  `async with` covers a step's guard expression and its call, which the coroutine
  form cannot, and every call site is shorter.
- Wrapped per step, not per `finally`, so a failing step cannot skip its
  successors.
- `except Exception`, not `BaseException`. The cleanup runs during worker shutdown
  too, and swallowing that `CancelledError` would strand the shutdown.
- Logged with `redact_exception_text`, since an ingest exception can carry a
  credentialed URL, under a fixed structlog event `ingest_cleanup_step_failed`
  with the step name and job id as fields. The two already-routed sites logged
  under an f-string event name; a fixed key is greppable.

It replaces two private copies of `_finally_cleanup` that had grown in
`tasks_vector.py` and `tasks_reupload.py`, whose bodies differed only in the task
name in their docstrings.

All 11 `finally:` blocks in the seven modules are covered, 36 steps in total:

| module | function | steps |
| --- | --- | --- |
| tasks_postgis_refresh.py | refresh_postgis | heartbeat |
| tasks_stac_refresh.py | refresh_stac | heartbeat |
| tasks_raster.py | ingest_raster | heartbeat, orphaned storage keys, temp dir, local file, presigned staging object, downloaded source |
| tasks_raster_replace.py | reupload_raster | heartbeat, orphaned storage keys, temp dir, local file, presigned staging object, downloaded source |
| tasks_vrt.py | ingest_vrt | heartbeat, temp dir, orphaned storage keys |
| tasks_vrt.py | regenerate_vrt | generation heartbeat, heartbeat, temp dir, orphaned storage keys |
| tasks_reupload.py | reupload_file | heartbeat, staging table, local file, downloaded source, presigned staging object |
| tasks_reupload.py | reupload_service | heartbeat, staging table (converted from the private copy) |
| tasks_vector.py | ingest_file | heartbeat, staging table, local file, downloaded source, presigned staging object |
| tasks_vector.py | ingest_service (nested) | progress heartbeat |
| tasks_vector.py | ingest_service | heartbeat, staging table (converted from the private copy) |

The nested block in `ingest_service` is one the `stop_ingest_job_heartbeat` grep
does not name. It cancels the service-import progress heartbeat and awaits it
under `suppress(asyncio.CancelledError)`, which lets through anything the heartbeat
body itself failed with, so it is the same class.

One site is deliberately left alone. `ingest_file`'s fan-out lookup inside its
`finally` keeps its own `except Exception`, because it has to assign a fail-safe
default (`is_fan_out_child = True`) that the helper cannot supply.

## Already landed, so not in scope

Verified with one grep each, so #1755 can be trimmed: items 4
(`service_token_required` in the error map, `SourceRefreshAction.tsx`), 5
(`autoComplete="new-password"` on the settings secret fields), 6 (the
`unauthorized` detail copy), 7 (`origin_probe` docstring), 9 (preview classifies
typed refusals, `sources/router.py`), 10 (`DeferFailed(cause=...)`), 14
(`enrich_arcgis_feature_counts` on the shared builder), 15 (`router_refresh`
docstring). Items 1, 2, 3, 12 and 13 are product or design decisions and are left
alone.

Item 11 was also partly landed: `reupload_service` and `ingest_service` already
routed their two calls each, which is why this PR consolidates rather than
introduces.

## Schema, API, config impact

None. No migration, no request or response schema change, no new setting.

## Verification

Run from the worktree against Postgres at localhost:5434.

```
uv run pytest tests/test_1755_alembic_logger_flag.py tests/test_1755_cleanup_step.py -q
  7 passed

uv run pytest tests/test_1746_stdlib_log_redaction.py tests/test_maps_style_json.py -q
  192 passed

  50 passed

uv run pytest tests/test_rule1_structural.py tests/test_rule2_structural.py \
  tests/test_failed_job_token_purge_1746.py -q
  120 passed

uv run pytest tests/ -q -k "ingest or vrt or raster or reupload or refresh or vector or staging or heartbeat"
  2464 passed, 5 skipped, 10223 deselected

uv run pytest tests/ -q -k "log or redact or warn or metric or audit or observab"
  1107 passed, 3 skipped, 11582 deselected

  All checks passed, 1169 files already formatted
```

Counterfactuals actually run, not inferred:

- Item 8, described above: 9 failures with the fix reverted.
- Item 11: with `raise` added back to `cleanup_step`'s handler,
  `tests/test_1755_cleanup_step.py` reports 3 failed, 2 passed.

## Caps

Every `_MODULE_LOC_CAPS` entry touched was re-measured with `wc -l` and set to the
exact figure, each with a comment saying what the lines bought.

| module | cap |
| --- | --- |
| tasks_common.py | 2551 to 2581 |
| tasks_postgis_refresh.py | 1128 to 1130 |
| tasks_vrt.py | 1628 to 1640 |
| tasks_reupload.py | 1333 to 1310 |
| tasks_vector.py | 1392 to 1368 |

`tasks_raster.py` (919), `tasks_raster_replace.py` (958) and
`tasks_stac_refresh.py` (801) are still under the 1000-line inclusion threshold,
so none of them enters the dict.

Refs #1755

### DIFF
--- a/backend/alembic/env.py
+++ b/backend/alembic/env.py
@@ -32,7 +32,10 @@ import app.modules.tenancy.models  # noqa: F401 -- register tenancy models (Phas
 
 config = context.config
 if config.config_file_name is not None:
-    fileConfig(config.config_file_name)
+    # fix(#1755 item 8): the stdlib default disable_existing_loggers=True sets
+    # .disabled on every logger registered before this call and not named in
+    # alembic.ini, and nothing in the app or the test suite restores that flag.
+    fileConfig(config.config_file_name, disable_existing_loggers=False)
 
 import pathlib  # noqa: E402
 from importlib.metadata import entry_points as iter_entry_points  # noqa: E402
--- a/backend/app/processing/ingest/tasks_common.py
+++ b/backend/app/processing/ingest/tasks_common.py
@@ -26,7 +26,7 @@ from app.platform.cache.tiles import invalidate_catalog_cache
 from app.platform.dataset_origin import classify_origin, set_dataset_origin
 from app.core.config import settings
 from app.core.service_tokens import reset_registered_credential_secrets
-from app.core.url_redaction import redact_url_credentials
+from app.core.url_redaction import redact_exception_text, redact_url_credentials
 from app.processing.embeddings.helpers import defer_embedding
 from app.processing.ingest.source_format import derive_source_format
 from app.platform.storage import get_storage
@@ -330,6 +330,36 @@ def purge_token_on_failure(fn):
             raise
 
     return _wrapper
+
+
+@asynccontextmanager
+async def cleanup_step(what: str, *, job_id: str) -> AsyncGenerator[None, None]:
+    """Run one terminal-cleanup step; log a failure in it rather than raise it.
+
+    fix(#1755 item 11): a `finally`-block cleanup step must never replace the
+    exception the block is already propagating, and must never skip the steps
+    after it in the same block. Wrap ONE step per `async with` -- the isolation
+    is per block, not per `finally`.
+
+    Catches ``Exception``, not ``BaseException``: the cleanup runs during
+    worker shutdown too, and swallowing that ``CancelledError`` would strand
+    the shutdown it is running under.
+
+    ``what`` names the step for the operator; it is logged verbatim, so keep
+    it a fixed string. The failure is logged with ``exc_info`` and a redacted
+    message, since an ingest exception can carry a credentialed URL.
+    """
+    try:
+        yield
+    except (
+        Exception
+    ) as exc:  # broad: cleanup must not replace the exception it runs past
+        structlog.get_logger().exception(
+            "ingest_cleanup_step_failed",
+            step=what,
+            job_id=job_id,
+            error=redact_exception_text(exc),
+        )
 
 
 # ArcGIS esriFieldType → column_info type mapping
--- a/backend/app/processing/ingest/tasks_postgis_refresh.py
+++ b/backend/app/processing/ingest/tasks_postgis_refresh.py
@@ -60,6 +60,7 @@ from app.platform.refresh.service import (
 )
 from app.processing.ingest.tasks_common import (
     _bind_task_log_context,
+    cleanup_step,
     _current_tenant_role,
     _current_tenant_schema,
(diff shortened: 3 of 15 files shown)


ishiland:
@codex review

Issue #1892, opened 2026-09-05
refactor(sandbox): bind tenant schemas without re-rendering SQL


ishiland:
From the 2026-09-05 tenant SQL routing investigation (source revision 4315df1bf; report and prototype in the maintainer's ignored `docs-internal/investigations/search-path-2026-09-05/`), with corrections from a second audit of it.

The multi-tenant executor (`backend/app/platform/sandbox/executor.py`, `_rewrite_logical_data_schema`) parses validated SQL with SQLGlot, rewrites `data.*` to the tenant's physical schema, and re-renders the whole expression tree. Re-rendering is why the cosine operator needs the `<=>` sentinel and `_swap_operator_tokens`, and it means the SQL that executes is not the SQL the caller wrote: today multi-tenant execution normalizes comment style, inserts `AS`, and re-cases `E'...'` literals, while single-tenant executes the text as written. Replacing only identifier spans aligns the two modes, so single-tenant test coverage starts to transfer.

The investigation measured two designs against PostgreSQL 18.6 / PostGIS 3.6.4 / pgvector 0.8.6 / SQLGlot 30.17 / SQLAlchemy 2.0.52:

- Unqualified names plus a transaction-local `search_path` works (20 alternating tenant queries on one pooled connection; role and path reset after commit, error and timeout), but it is a contract migration: the public `POST /api/query/` documents `data.*`, the AI prompt and DDL require it, spatial-analysis previews call `execute_safe` directly with `data` names, a qualified name bypasses the path, a CTE that shares a dataset's name changes meaning when the prefix is dropped, path order admits `pg_catalog`/`pg_temp`/`public` fallbacks, and a reused prepared statement across tenants whose tables differ in shape fails with 0A000 unless the SQLAlchemy prepared-statement cache is disabled. The app engine uses SQLAlchemy's default prepared-statement caching (`config.py` connect args), so the tested pool matches production. Deferred.
[...]

Issue #1891, opened 2026-09-05
sandbox: unquoted uppercase DATA schema validates but is not routed to the tenant schema


ishiland:
Found by the 2026-09-05 tenant SQL routing investigation (source revision 4315df1bf) and confirmed by a second audit of it.

`validate_sql` folds an unquoted schema identifier the way PostgreSQL does, so `SELECT * FROM DATA.roads` passes the allowlist as `data.roads`. The multi-tenant rewrite in `backend/app/platform/sandbox/executor.py` (`_rewrite_logical_data_schema`) then compares `table.db == "data"` and `column.db == "data"` without that folding, so any non-lowercase unquoted spelling (`DATA`, `Data`; reproduced with `Data.Roads`) is left in place. PostgreSQL resolves it to the shared `data` schema, and the tenant reader role refuses it with 42501: tenant readers hold USAGE on their own schema only (migrations 0019 and 0024), so a mis-routed query is denied, never served from the legacy shared schema. In a multi-tenant deployment the query fails; this is a correctness defect, not an isolation gap. Single-tenant mode does not rewrite and is unaffected.

Fix: fold the case of an unquoted `db` identifier before comparing (quoted `"DATA"` stays distinct and keeps being rejected by the validator), for both `Table` and `Column` references. Regression tests in `backend/tests/test_sandbox_tenant_schema.py`: unquoted `DATA.roads`, `Data.Roads` (mixed case on both parts) and a three-part column `DATA.roads.geom` all rewrite to the physical schema; quoted `"DATA".roads` does not rewrite and does not validate.

Known blind spot, parity rather than regression: the composite-reference form `(data.roads).marker` is invisible to the rewriter today and would be to a span rewrite too; it fails closed under the tenant role.

ishiland:
Fixed by #1893 and #1895 (both merged 2026-09-05). The executor folds an unquoted schema identifier the way PostgreSQL does before deciding it is the logical data schema, so every spelling the validator accepts is routed to the tenant schema, and the AI-chat geometry auto-append folds both parts of the table reference the same way. Quoted mixed-case spellings stay refused. The span-rewrite refactor remains #1892.

PR #1893, opened 2026-09-05
fix(sandbox): fold an unquoted DATA schema before the tenant rewrite (#1891)

ishiland:
Issue: #1891 (found by the 2026-09-05 tenant SQL routing investigation).

## Defect

`validate_sql` folds an unquoted schema identifier the way PostgreSQL does, so `SELECT * FROM DATA.roads` passes the allowlist as `data.roads`. The multi-tenant rewrite in `backend/app/platform/sandbox/executor.py` (`_rewrite_logical_data_schema`) then compared `table.db == "data"` and `column.db == "data"` with no folding, so any non-lowercase unquoted spelling (`DATA`, `Data`, `Data.Roads`) stayed in place and PostgreSQL resolved it to the shared `data` schema.

Severity: the tenant reader role holds USAGE on its own tenant schema only. The provisioning function from migrations 0019 and 0024 revokes each tenant schema from PUBLIC and grants USAGE to that tenant's reader alone, and no tenant reader is granted anything on the shared schema (the baseline's `GRANT USAGE ON SCHEMA data` goes to the single-tenant `geolens_readonly` role). A mis-routed query is therefore refused with 42501 and is never served from the shared schema. This is a correctness defect, not an isolation gap. Single-tenant mode does not rewrite and was not affected.

Reproduced on main at 56aa03abe before the change: `_rewrite_logical_data_schema("SELECT * FROM DATA.roads", schema)` returned its input unchanged, and so did `Data.roads` and `SELECT DATA.roads.geom FROM DATA.roads`, while lowercase `data.roads` was rewritten.

## Fix

Both loops read the identifier node (`table.args.get("db")`, `column.args.get("db")`) and pass it to a new `_is_logical_data_schema` helper in the executor. It folds the way PostgreSQL and the validator's `_folded_identifier` do: an unquoted name to lowercase, a quoted name verbatim, and it matches on `data`. A `db` argument that is not an identifier is not rewritten. Only the schema identifier is replaced; the table part stays as written and PostgreSQL folds it on its own. Quoted `"DATA"` is therefore still not rewritten and `check_table_access` still refuses it. The physical replacement stays `exp.to_identifier(physical_schema, quoted=True)` and the cosine sentinel is untouched.

## Tests

Five regression cases in `backend/tests/test_sandbox_tenant_schema.py`, next to the existing rewrite tests, assert on the rendered SQL the way that file already does. Unquoted `DATA.roads`, `Data.roads` and `Data.Roads` (parametrized; each spelling is first checked to validate as `("data", "roads")`, and `Roads` is left as written in the output) and the three-part column `DATA.roads.geom` rewrite to the physical schema. Quoted `"DATA".roads` is not rewritten; `validate_sql` extracts it as `("DATA", "roads")` and `check_table_access` refuses it with `table_not_accessible`.

Counterfactual: tests in place, executor at main, so only the folding is missing.

```
E   assert 'FROM "data_t_00000000_0000_0000_0000_000000000001".roads' in 'SELECT * FROM DATA.roads'
E   assert 'FROM "data_t_00000000_0000_0000_0000_000000000001".roads' in 'SELECT * FROM Data.roads'
E   assert 'FROM "data_t_00000000_0000_0000_0000_000000000001".Roads' in 'SELECT * FROM Data.Roads'
E   assert '"data_t_00000000_0000_0000_0000_000000000001".roads.geom' in 'SELECT DATA.roads.geom FROM DATA.roads'
FAILED tests/test_sandbox_tenant_schema.py::test_schema_rewrite_folds_unquoted_logical_schema[DATA-roads]
FAILED tests/test_sandbox_tenant_schema.py::test_schema_rewrite_folds_unquoted_logical_schema[Data-roads]
FAILED tests/test_sandbox_tenant_schema.py::test_schema_rewrite_folds_unquoted_logical_schema[Data-Roads]
FAILED tests/test_sandbox_tenant_schema.py::test_schema_rewrite_folds_unquoted_three_part_column
4 failed, 1 passed
```

The one pass is the quoted `"DATA"` test, which pins behaviour main already has.

## Impact

No request or response schema change, no migration, no new setting. `executor.py` is not in `_MODULE_LOC_CAPS`, so no cap moved.

## Verification

Run on the host from the worktree, Postgres at localhost:5434.

```
uv run pytest tests/test_sandbox_tenant_schema.py tests/test_sandbox.py -q      # 81 passed
uv run pytest tests/test_layering.py -q                                         # 50 passed
uv run pytest tests/test_rule1_structural.py tests/test_rule2_structural.py -q  # 115 passed
uv run ruff check . && uv run ruff format --check .                             # clean
```

## Left alone

The validator keeps its private `_folded_identifier`. Sharing it would mean a rename across the capped validator for no behaviour change, and the executor helper is four lines. The span-rewrite refactor and the trailing-semicolon handling are #1892 and are not started here.

### DIFF
--- a/backend/app/platform/sandbox/executor.py
+++ b/backend/app/platform/sandbox/executor.py
@@ -70,6 +70,15 @@ def _swap_operator_tokens(sql: str, token_type: TokenType, replacement: str) -> 
     return out
 
 
+def _is_logical_data_schema(identifier: exp.Expression | None) -> bool:
+    """True when ``identifier`` folds to the logical ``data`` schema: unquoted
+    folds to lowercase, quoted keeps its case, so quoted ``"DATA"`` is not it."""
+    if not isinstance(identifier, exp.Identifier):
+        return False
+    name = identifier.name if identifier.quoted else identifier.name.lower()
+    return name == "data"
+
+
 def _rewrite_logical_data_schema(sql: str, physical_schema: str) -> str:
     """Rewrite validated ``data.*`` references to one physical tenant schema.
 
@@ -96,12 +105,13 @@ def _rewrite_logical_data_schema(sql: str, physical_schema: str) -> str:
     # quoting form) tokenizes as a string and leaves `guarded` unchanged.
     protect_cosine = guarded != sql
 
+    # fix(#1891): fold like the validator, so unquoted DATA is rewritten too.
     for table in statement.find_all(exp.Table):
-        if table.db == "data":
+        if _is_logical_data_schema(table.args.get("db")):
             table.set("db", exp.to_identifier(physical_schema, quoted=True))
 
     for column in statement.find_all(exp.Column):
-        if column.db == "data":
+        if _is_logical_data_schema(column.args.get("db")):
             column.set("db", exp.to_identifier(physical_schema, quoted=True))
 
     rendered = statement.sql(dialect="postgres")
--- a/backend/tests/test_sandbox_tenant_schema.py
+++ b/backend/tests/test_sandbox_tenant_schema.py
@@ -141,19 +141,23 @@ def test_schema_rewrite_preserves_non_single_quoted_literals():
     assert out2.count("<=>") == 1
 
 
-@pytest.mark.parametrize("spelling", ["DATA", "Data"])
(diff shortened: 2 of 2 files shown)


ishiland:
@codex review

Issue #1848, opened 2026-09-05
fix(sources): probe, preview and re-upload doors hold a pooled DB connection across network I/O and GDAL


ishiland:
Found by the post-launch backend audit of 2026-09-04 (v1.18.0, 4e3600e37); anchor for the wave A2 lane.

Five handlers keep their pooled `AsyncSession` checked out while calling the remote service or GDAL: `probe_service_url` (`backend/app/modules/catalog/sources/router.py:894`), `preview_service_layer` (`:1172`), and the three re-upload doors under `datasets/api/router_reupload*.py`. The pool is 10 plus 3 overflow with a 30 s timeout, so thirteen slow previews wedge every other request. The wave that shipped authenticated service import fixed this class four times elsewhere (`router_health`, `router_refresh` twice, `arcgis_signin`) and once in export under #1778; these were missed.

Fix: read what the handler needs, release the session (or end the transaction) before the network or subprocess call, re-acquire for the write. Four of the sites are S; `reupload_dataset` is M. After #1846 lands, `run_service_preview` is the one site that still runs unclamped for layering reasons; note it in the same pass.

ishiland:
#1880 (merged 2026-09-05) releases the pooled connection before network I/O and GDAL in the five doors it names (service probe, service preview, the two re-upload previews, and re-upload commit-first), including the release above the SSRF resolver wait and a compare-and-set bind that stamps the staging clock. Still holding a connection across I/O and left for a follow-up: `request_presigned_reupload` and the ingest `upload_file` door, which have the same shape. Leaving this open for those two.

ishiland:
Fixed by #1880 (the direct re-upload door) and #1898 (the presigned re-upload request and the ingest upload door), both merged 2026-09-05. Every door now commits its job row before the storage call and binds the result through a guarded UPDATE keyed on the row's id and pending status, so a pooled connection is never held across network I/O or GDAL, and a row the sweep or a cancel settled in the window is refused with a 409 and its storage object cleaned up.

PR #1898, opened 2026-09-05
fix(ingest,reupload): release the pooled connection before storage in the last two upload doors (#1848)

ishiland:
Finishes the connection-holding class from the backend audit of 2026-09-04, tracked in #1848. #1880 released the pooled `AsyncSession` in five doors and named two more it left alone: `request_presigned_reupload` in `datasets/api/router_reupload.py` and `upload_file` in `processing/ingest/router.py`. Both still reproduced on `origin/main` at 56aa03abe: each door's source holds the session across its storage call, and the two release counterfactuals below are exactly that shape.

Both doors take the shape #1880 gave `reupload_dataset`. The row is committed before the handler talks to storage, which hands the connection back, and every later write is a compare-and-set that matches the row only while it is still `pending`. A zero rowcount is a 409 telling the caller to start again.

## The presigned re-upload door

`request_presigned_reupload` created the job, then initiated the multipart upload against S3 and signed one URL per part, all inside the request transaction. It now commits the row right after the lifetime gate, with the two markers the job-binding gate reads (`reupload` and `dataset_id`) already on it, then asks storage. The presigned facts (`presigned`, `s3_key`, `upload_id`, `multipart`, `expected_size`) land afterwards through the guarded bind. The guard carries `dataset_id` as well as `status`, for the reason #1880's third round found: the early commit releases the foreign-key lock, so a dataset deleted while storage signs nulls the binding.

Invariant: the request session is not in a transaction when `initiate_multipart_upload` or either signing method runs, and no presigned fact is ever written onto a row that is not `pending` and bound to the path's dataset.

The bind does not stamp `staged_at`. Nothing is staged when this door answers; the completion door binds the frozen object, and that key is what moves the row into the sweep's 24-hour class. Stamping here would equal `created_at`, which #1880 measured as a no-op.

On a refused multipart bind the upload id storage handed out is aborted, the same cleanup every other failure in that branch already runs. A refused single-part bind has nothing to clean: the PUT URL was minted but never returned.

The lifetime gate stays above the commit on purpose. It is a pure check on `created_at`, and a 409 from it still rolls the uncommitted row back exactly as before.

Committing first has the same durable-row consequence here as on the direct door. A storage 502 at initiation, a lifetime 409 inside the sign loop, or a client cancel during initiation now leaves a `pending` row carrying the reupload markers and no presigned marker, where before the PR that row rolled back with the request. It stays until `pending_job_timeout_seconds`, then the sweep marks it `cancelled` as abandoned. Nothing gates on it in the meantime: neither re-upload door has an active-job gate, `/jobs/by-dataset` lists completed jobs only, and the completion door answers 400 for a row with no presigned marker.

## The ingest upload door

FastAPI receives and spools the whole multipart body before any dependency runs, so `upload_file` never held the connection across the network stream. It held it across `save_upload_file`, which is the S3 put in S3 mode and the copy from the spool into staging in local mode, then across the validation download (S3 mode) and the content sniff, and bound `file_path` through an ORM flush at the end. It now commits the row before `save_upload_file` and binds through a guarded UPDATE that stamps `staged_at`, so a local upload slower than `pending_job_timeout_seconds` is not cancelled by the very next sweep. The raster stamp goes through `_raster_stamped_metadata`, the pure form `upload_from_url` already uses, because a dirtied ORM instance would flush a second, unguarded UPDATE on the commit and bypass the guard.

Invariant: the request session is not in a transaction while the spooled body is staged, downloaded back for validation, or sniffed, and `file_path` is only ever bound onto a row that is still `pending`.

The guard here is `id` and `status` only. A fresh upload has no dataset binding to protect.

Two behaviours change, and both are the two #1880 accepted for the direct re-upload door:

- A content refusal (422) used to roll the uncommitted row back. The row is durable now, so it is stamped `failed` with the refusal and `completed_at` through the same guard as the bind, the terminal-row shape every other terminal writer produces. That is what the presigned completion door already does for the same 422. The direct re-upload door's refusal stamp from #1880 gets the same `completed_at`.
- A provider failure, a 413 from the size check in `save_upload_file`, or a failed validation download leaves a `pending` row with nothing bound. That is the unbound half of `stale_pending_clauses`, and the sweep reaps it on the one-hour policy. One test in `test_ingest.py` asserted the old rollback by name and now asserts this.

## Shared helpers

The guard (`_pending_reupload_update`), the refusal (`_reupload_bind_refusal`) and the bind (`_bind_presigned_reupload`) are module helpers in `router_reupload.py`, and the two guarded writes #1880 put in `reupload_dataset` now go through the first two. The ingest router gets `_pending_upload_update` for its two writes. The two files cannot share one helper: `modules/catalog` may not import `app.processing`, and the ingest guard has no dataset clause.

## Tests

Eight cases added to `tests/test_pooled_connection_release_1848.py`, with the same `in_transaction()` recording as the twelve already there.

Presigned door: the release, with the multipart initiation and the part signature both recorded on a released session; a sweep that reclaims the row while storage answers (409, row stays `cancelled`, `abort_multipart_upload` called with the upload id); a dataset deleted while signing (409, row orphaned `pending` with no presigned fact).

Upload door: the release, with the row visible from a second session before staging starts; a sweep mid-upload (409, staged file cleaned); a slow local staging step that survives the next sweep because of `staged_at`; a content refusal that leaves a `failed` row with `completed_at`. One more case pins the same `completed_at` on the direct re-upload door's refusal stamp.

Three mock-mirror tests updated. `test_upload_commit_failure_cleans_owned_staged_file` fails the second commit rather than the first. `test_presigned_reupload_round_trip_uses_tenant_provider_key` reads the logical `s3_key` off the UPDATE statement the door ran instead of off the ORM instance. `test_s3_validation_download_failure_deletes_uncommitted_source` becomes `..._deletes_the_staged_source` and asserts the surviving pending row.

## Counterfactuals

Each run by reverting only that piece, keeping the tests, and restoring the file afterwards.

- Presigned release removed: `assert [True, True] == [False, False]`, held at the initiation and at the signature.
- Upload release removed: `assert [True] == [False]`, held across the staging step.
- Presigned guard reduced to `id` only: the swept test and the deleted-dataset test both get `201` with a full presigned response on a row that is `cancelled` or unbound.
- Upload guard reduced to `id` only: the mid-upload sweep test gets `201` with `{"status": "pending"}` on a row the sweep had already cancelled.
- `staged_at` removed from the upload bind: the slow-upload test finds the row `cancelled` right after a successful 201 (`assert 'cancelled' == 'pending'`).

## Verification

From the worktree against Postgres on localhost:5434.

```
tests/test_pooled_connection_release_1848.py      20 passed
tests/test_upload_size_limit.py                    8 passed
tests/test_tenant_staging_storage_keys.py         11 passed
tests/test_ingest.py                              45 passed
tests/test_reupload.py                            44 passed
tests/test_reupload_idor.py                        7 passed
tests/test_reupload_service.py                     5 passed
tests/test_reupload_expected_origin_1768.py       14 passed
tests/test_reupload_record_type_guard.py          13 passed
tests/test_reupload_completion_contract_1207.py   12 passed
tests/test_reupload_swap_lock_retry.py             8 passed
tests/test_job_cancel_reupload_commit.py           2 passed
tests/test_presigned_content_validation_1202.py   26 passed
tests/test_1184_body_limit_per_route.py           45 passed
tests/test_1224_ingest_quota_enforcement.py        7 passed
tests/test_api_contract_openapi.py                23 passed
tests/test_1183_s3_spool_upload.py                 4 passed
tests/test_layering.py + test_rule1_structural.py + test_rule2_structural.py   165 passed
ruff check .                                      All checks passed
ruff format --check .                             1176 files already formatted
```

None. No migration, no request or response schema change, no new setting. The 409 the two doors can now answer is the one `reupload_dataset` has answered since #1880.

## Noticed and left alone

- `request_presigned_upload` in `processing/ingest/router.py`, the ingest twin of the presigned re-upload door, initiates the multipart upload and signs inside its transaction too. It is on neither #1848's list nor #1880's follow-up note, and its window is one S3 round trip rather than a stream, so it is not touched here. The same fix would apply.
- The `staged_at` comment in `platform/jobs/sweep.py` still says `upload_from_url` is the only writer of that key. That stopped being true in #1880 and is less true now. Left for a pass that otherwise touches that file.

Caps re-measured with `wc -l` and set exact: `router_reupload.py` 1455 to 1506, `ingest/router.py` 2616 to 2666.

### DIFF
--- a/backend/app/modules/catalog/datasets/api/router_reupload.py
+++ b/backend/app/modules/catalog/datasets/api/router_reupload.py
@@ -209,7 +209,8 @@ async def _bind_presigned_reupload(
     """Write the presigned facts onto the job; False when no row matched.
 
     No `staged_at` is stamped: nothing is staged until the completion door
-    binds the frozen object, and that key restarts the pending window.
+    binds the frozen key, which moves the row into the sweep's 24-hour
+    completion-bound class, still measured from `created_at`.
     """
     bound = await db.execute(
         _pending_reupload_update(job_id, dataset_id).values(user_metadata=metadata)
@@ -382,7 +383,9 @@ async def reupload_dataset(
             # already reclaimed keeps its terminal status and message.
             await db.execute(
                 _pending_reupload_update(job.id, dataset_id).values(
-                    status="failed", error_message=str(exc)
+                    status="failed",
+                    error_message=str(exc),
+                    completed_at=datetime.now(timezone.utc),
                 )
             )
             await db.commit()
--- a/backend/app/modules/catalog/datasets/api/router_reupload.py
+++ b/backend/app/modules/catalog/datasets/api/router_reupload.py
@@ -173,6 +173,51 @@ async def _cleanup_uncommitted_reupload_source(
         )
 
 
+def _pending_reupload_update(job_id: uuid.UUID, dataset_id: uuid.UUID):
+    """An UPDATE that matches the job only while it is pending and bound here.
+
+    Every write after the row is committed goes through this guard, so a row
+    the sweep reclaimed, or one a dataset deletion unbound, is left as it was
+    and the caller learns that from a zero rowcount.
+    """
+    return update(IngestJob).where(
+        IngestJob.id == job_id,
+        IngestJob.dataset_id == dataset_id,
+        IngestJob.status == "pending",
+    )
+
+
+def _reupload_bind_refusal() -> HTTPException:
+    """The 409 for a guarded write that matched no row."""
+    return HTTPException(
+        status_code=status.HTTP_409_CONFLICT,
+        detail=(
+            "This upload could not be attached to the dataset. It may "
+            "have taken too long, or the dataset may no longer exist. "
+            "Start the re-upload again."
+        ),
+    )
+
+
+async def _bind_presigned_reupload(
+    db: AsyncSession,
+    *,
+    job_id: uuid.UUID,
+    dataset_id: uuid.UUID,
+    metadata: dict,
+) -> bool:
+    """Write the presigned facts onto the job; False when no row matched.
+
+    No `staged_at` is stamped: nothing is staged until the completion door
+    binds the frozen object, and that key restarts the pending window.
+    """
+    bound = await db.execute(
+        _pending_reupload_update(job_id, dataset_id).values(user_metadata=metadata)
+    )
+    await db.commit()
+    return bool(bound.rowcount)
+
+
 def _assert_compatible_record_type(
     dataset,
     filename: str | None,
@@ -336,13 +381,9 @@ async def reupload_dataset(
             # fix(#1848 audit): guarded like the bind below, so a row the sweep
             # already reclaimed keeps its terminal status and message.
             await db.execute(
-                update(IngestJob)
-                .where(
-                    IngestJob.id == job.id,
-                    IngestJob.dataset_id == dataset_id,
-                    IngestJob.status == "pending",
+                _pending_reupload_update(job.id, dataset_id).values(
+                    status="failed", error_message=str(exc)
                 )
-                .values(status="failed", error_message=str(exc))
             )
             await db.commit()
             raise HTTPException(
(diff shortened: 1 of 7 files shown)


ishiland:
@codex review

PR #1916, opened 2026-09-06
fix(ingest): bump tile_cache_version atomically once the worker swaps hold the row (#1911)

### LINKED ISSUE #1911: fix(ingest): the raster replace and reupload swaps write an absolute tile_cache_version read before their lock wait (#1826 class)
Follow-up to #1910 (fix for #1826 and #1902), which made every request-side `tile_cache_version` bump atomic. Two worker doors are the same class and were left out because both issues scoped to the request side:

- `_write_swapped_fields` in `backend/app/processing/ingest/tasks_raster_swap.py` (called from `backend/app/processing/ingest/tasks_raster_replace.py` ahead of that task's lock acquisition) assigns the absolute `Dataset.bump_tile_cache_version()` on an instance loaded before the wait.
- `_apply_reupload_swap` in `backend/app/processing/ingest/tasks_common.py`: the instance is loaded by the caller before the ACCESS EXCLUSIVE wait.

A request-side edit that commits while either swap waits is overwritten by the swap's `N + 1`, so tiles are served under the earlier version. Fix shape: route both through `bump_tile_cache_version_on` from `backend/app/platform/catalog_locks.py` after the lock is held. `tests/test_reupload_swap_lock_retry.py` mirrors the swap with a stub session and a stubbed bump, so the mirror has to follow the change.

ishiland:
## What changed

Follow-up to #1910. The two worker swap doors assigned an absolute `Dataset.bump_tile_cache_version()` on an instance loaded before their lock wait, so a request-side edit that committed while the swap waited was overwritten by the swap's N+1 and its tiles stayed served under the earlier URL. Both doors now roll the counter through `bump_tile_cache_version_on` after the catalog rows are locked, the same spelling #1910 gave every request-side writer.

- `tasks_raster_swap._write_swapped_fields` no longer bumps. Its caller in `tasks_raster_replace.reupload_raster` bumps atomically right after `lock_catalog_rows`, which is also after the last I/O the door does before the lock (`archive_lossy_original`).
- `tasks_common._apply_reupload_swap` bumps atomically at the site of the old absolute write, which was already below its `lock_catalog_rows`.
- No new 409 or retry mapping. Both doors are workers: `lock_catalog_rows(lock_timeout=None)` and `CatalogLockConflict` propagates and fails the job, as #1910 left it.
- `tasks_common.py` cap in `tests/test_layering.py` lowered 2559 -> 2558, exact.

## Remaining absolute `bump_tile_cache_version()` callers in `backend/app/`

Grepped `bump_tile_cache_version(` and `tile_cache_version`. Every remaining absolute caller loads its instance under a `FOR UPDATE` on the datasets row that it already holds, so the value it writes back is the row's current value:

| Site | Why it is outside this class |
|---|---|
| `tasks_postgis_refresh.py` phase 3 (`_apply_measurement` tail) | Takes `select(Dataset.tile_cache_version).with_for_update()` for its superseded guard, then loads the instance. |
| `tasks_stac_refresh.py` moved-asset refresh | Takes `select(Dataset.origin_uri, ...).with_for_update()` for its binding guard, then loads the instance. |
| `tasks_vrt.py` regenerate swap | `select(RasterAsset).join(Dataset).with_for_update()` locks the joined datasets row, then loads `vrt_dataset`. |
| `Dataset.bump_tile_cache_version` (models.py) | The method itself; its docstring already names the atomic spelling for a caller that loaded before waiting. |

`modules/catalog/` has no absolute callers; `test_feature_lock_order_1847.py::TestNoCatalogModuleWritesAnAbsoluteTileVersion` gates that, and the new structural test extends the scan to the three worker modules touched here.

Adjacent, not in scope: both doors also write `dataset.current_version = new_version` from the same pre-wait read. That is a different counter, only these two doors write it, and #1911 scopes to `tile_cache_version`.

## Tests

`tests/test_worker_swap_bump_after_lock_1911.py`, one concurrency test per door, shaped like #1910's: a holder takes the datasets row, the swap parks behind it, the holder publishes N+1 and commits, the swap must publish N+2.

- The holder takes `FOR NO KEY UPDATE`, not `FOR UPDATE`. The raster task's phase-1 job-row writes re-run their foreign-key check against the datasets row (the job row was already updated in that transaction), and that `KEY SHARE` queued behind a `FOR UPDATE` holder and released the barrier before the swap reached its own acquisition. The barrier now also asserts that the parked statement is the swap's `catalog.datasets ... FOR UPDATE`.
- Counterfactual: with the three `backend/app` files restored to `origin/main`, both concurrency tests fail with "the swap published 2 ... must publish 3". With the fix, they pass.

`tests/test_reupload_swap_lock_retry.py` is a mock mirror of the swap (SimpleNamespace dataset, real session) and was updated deliberately: the stub can no longer carry a `bump_tile_cache_version` method, because the door now issues `UPDATE catalog.datasets` against `type(dataset)`, which a SimpleNamespace cannot satisfy. A `_stub_atomic_bump(monkeypatch)` helper stands in for `bump_tile_cache_version_on` and rolls the stub's counter in place; the happy-path test asserts the counter rolled, and the write-ordering test still sees that write after the acquisition.

## Verification

```
# All checks passed! / 1181 files already formatted

uv run pytest tests/test_worker_swap_bump_after_lock_1911.py tests/test_reupload_swap_lock_retry.py tests/test_layering.py -q
# 62 passed

uv run pytest tests/test_feature_lock_order_1847.py tests/test_raster_replace_1221.py tests/test_dataset_source_state.py tests/test_provenance_attribution.py tests/test_job_cancel_no_swap.py -q -n 4
# 353 passed, 1 skipped
```

Closes #1911

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ and releases use semantic versioning.
 - `make dev` preflight now judges a `.env` line whose key is empty or holds a tab the way docker
   compose loads it: a value compose cannot resolve is refused before the build instead of at
   compose load, and a multiline value under such a key no longer fails the check. (#1909)
+- A raster replace or a dataset reupload that waited behind a feature edit, column change or
+  tile-columns edit wrote back the tile cache version it had read before waiting, so the edit
+  committed during the wait was served under a stale tile URL. Both worker swaps now roll the
+  version atomically once they hold the catalog row, as the request side has since #1910. (#1911)
 
 ## [1.18.1] - 2026-09-05
 
--- a/backend/app/processing/ingest/tasks_common.py
+++ b/backend/app/processing/ingest/tasks_common.py
@@ -2386,7 +2386,7 @@ async def _apply_reupload_swap(
     # fix(#1847): the catalog writes start here, and this function dirties both
     # rows. `lock_timeout=None` so SET LOCAL does not clamp an ingest swap to a
     # request's budget.
-    from app.platform.catalog_locks import lock_catalog_rows
+    from app.platform.catalog_locks import bump_tile_cache_version_on, lock_catalog_rows
     from app.platform.extensions import get_processing_port
 
     _port = get_processing_port()
@@ -2472,10 +2472,9 @@ async def _apply_reupload_swap(
     dataset.source_filename = source_filename
     dataset.original_srid = original_srid
     dataset.current_version = new_version
-    # fix(#525 B-038): tile_version now reads tile_cache_version (not
-    # current_version), so reupload must roll it too or the _v= tile-URL
-    # cache-buster stops changing on the largest content mutation of all.
-    dataset.bump_tile_cache_version()
+    # fix(#1911): evaluated at write time, under the lock, so the counter read
+    # into `dataset` before the wait is never written back over a peer's commit.
+    await bump_tile_cache_version_on(session, dataset)
     dataset.record.updated_by = actor_id
     if source_url is not None:
         dataset.source_url = source_url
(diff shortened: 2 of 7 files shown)


ishiland:
@codex review

ishiland:
@codex review

Issue #1927, opened 2026-09-06
SEC-OBSV-01/02: record the Titiler redirect-following constraints behind a resolvable anchor


ishiland:
Found while reviewing #1922.

`SEC-OBSV-01` and `SEC-OBSV-02` are used as cross-file identifiers in three places:

- `backend/app/processing/tiles/router.py` — the `_titiler_client` built with `follow_redirects=True`
- `backend/app/platform/storage/titiler_url.py` — a docstring naming both
- `backend/tests/test_titiler_url_helper.py` — a comment naming SEC-OBSV-01

They came in with the initial public release, so `git log -S` finds no PR and no issue behind either one. A reader who meets `SEC-OBSV-01` in one of those files has no way to learn what it refers to, which is the situation AGENTS.md's inline review-comment convention exists to prevent. Deleting the tokens is not right either, because they are doing real work as a cross-reference between the three sites.

This issue is that public anchor. Recording the constraints so the markers resolve to something:

**SEC-OBSV-01 — the Titiler client may follow redirects only while Titiler is unreachable from outside.** `_titiler_client` is constructed directly with `follow_redirects=True` rather than through `make_safe_client()`. That is sound only while both hold: the Titiler service publishes no `ports:` in `docker-compose.yml`, so it is reachable only on the internal network; and every URL handed to that client is server-derived, never caller-supplied. If Titiler is ever exposed, or any caller-influenced URL reaches this client, it must move to `make_safe_client()` so the per-hop redirect revalidation in `backend/app/platform/security.py` applies. See Rule 2 in AGENTS.md.

**SEC-OBSV-02 — the same constraint for the STAC COG info fetch**, `stac_router.py::_fetch_cog_info`, as named in `titiler_url.py`.

Two follow-ups worth considering, neither urgent:

- both sites depend on a compose-file property (no `ports:` on Titiler) that nothing tests. A structural check that fails if the Titiler service gains a published port would turn a written assumption into an enforced one;
- `test_titiler_url_helper.py` already strips comments before checking for literal Titiler hosts, so part of this is pinned; the redirect exception itself is not.

#1922 anchors the router comment to this issue while keeping the `SEC-OBSV-01` token so the cross-references still line up.

PR #1994, opened 2026-09-09
fix(catalog): anchor SEC-OBSV-02 marker in cog_info.py

### LINKED ISSUE #1945: SEC-OBSV-02's docstring in cog_info.py misnames its function and cites a line the file does not have
Found while landing #1927 in #1944, which anchors the `SEC-OBSV-02` token in this file but deliberately stops short of the surrounding prose.

`backend/app/modules/catalog/sources/cog_info.py` carries a `SEC-OBSV-02` docstring block describing the dual SSRF gate. Two of its statements no longer describe the tree:

- It says "EVERY caller of `_fetch_cog_info`". The function is `fetch_cog_info`; the leading underscore went when it moved out of the STAC router in #1266.
- It says "The import-flow call at line 454 satisfies this." The file is 289 lines, so that citation resolves to nothing.

Neither is a security defect. The dual gate itself is intact, and #1944 corrected the sibling pointer in `backend/app/platform/storage/titiler_url.py`, which had the same stale function name and still named `stac_router.py` as the home of the call. What is left is a block that tells a reader to go and check a caller it misnames, at a line that does not exist.

## Why it was not folded into #1944

That PR's #1927 commit pairs each `SEC-OBSV` token with a resolvable anchor, three lines per site. Rewriting a 19-line docstring block is a different edit against a different rule, and #1944 is otherwise a no-behaviour-change hygiene pass whose whole value is that it is easy to review.

## What a fix looks like

Correct the function name, replace the line citation with something that survives edits (the calling function's name, not its offset), and bring the block within the inline review-comment convention's three-line limit while keeping the statement of the dual gate, which is a real invariant a future caller needs.

`cog_info.py` is 289 lines and carries no `_MODULE_LOC_CAPS` entry, so no cap moves.

### LINKED ISSUE #1951: cog_info.py's SEC-OBSV-02 anchor sits seven lines from its token, so the marker gate cannot credit it
Found by #1934's lane while regenerating its debt map after rebasing onto #1944.

#1944 rewrote the `SEC-OBSV-02` docstring block in `backend/app/modules/catalog/sources/cog_info.py` to state the dual-gate contract and drop the audit narrative. It kept both the token and the `#1927` anchor, but seven lines apart: the token opens the paragraph at line 194, and "Recorded in #1927." closes it at line 201.

#1934's detector uses a two-line anchor window, so a reader landing on the token does not find the anchor beside it and the marker is not credited as anchored. `cog_info.py` therefore stays at 1 in `UNANCHORED_MARKER_DEBT` rather than dropping to 0, and the map total stays 1417 instead of 1416.

Nothing is broken and the gate is behaving as designed — the window is deliberate, and AGENTS.md caps an inline review comment at three lines for the same reason. But it is the first case where the rule lands on a *docstring paragraph*, where prose legitimately runs longer than a comment block may.

## The fix, and why not the other one

Move the anchor to the token rather than widening the window:

```
SEC-OBSV-02 (#1927): SSRF protection is a dual gate and both halves must hold
```

and drop the trailing "Recorded in #1927." That is one line, it pays the debt entry down honestly, and it leaves the detector's window alone.

Widening the window for docstrings is the alternative and should be rejected unless something else argues for it. The window is what makes an anchor findable by a reader who lands on the marker, which is the whole point of the convention. Making it looser so a paragraph can pass costs the guarantee to buy nothing.

## Sequencing

Not folded into #1934: that PR no longer touches `cog_info.py` after its rebase, and reintroducing the file would re-create the conflict class its rebase just resolved. Do this after #1934 lands, then lower the `cog_info.py` entry to 0 in the same commit — the gate is exact in both directions, so removing the marker without lowering the number fails, and lowering it without removing the marker fails too.

ishiland:
## Summary

Fixes the `SEC-OBSV-02` marker in `cog_info.py`: the `#1927` anchor sat
five lines past the token, outside the marker gate's two-line window, and
the docstring paragraph around it had drifted out of date.

## Changes

- Move the `#1927` anchor onto the `SEC-OBSV-02` token's own line and
  restate the dual-gate contract in three lines instead of five, dropping
  the stale prose the gate could not credit.
- Drop the now-zero `cog_info.py` entry from `UNANCHORED_MARKER_DEBT` in
  `backend/tests/finding_markers.py`.
- No behavior change; `backend/app/platform/storage/titiler_url.py`'s
  sibling pointer already names `fetch_cog_info` correctly and needs no
  edit.

## Area

## Testing

Ran `python3 backend/tests/finding_markers.py` (clean; confirmed the
pre-fix docstring left `SEC-OBSV-02` unanchored via a direct
`scan_module` check, and the fixed one anchors it), `backend/tests/test_layering.py`,
`ruff check` / `ruff format --check`, and the focused suites that cover
`cog_info.py`: `test_cog_info.py`, `test_sources_safe_client_structural.py`,
`test_titiler_url_helper.py`.

## Checklist

Closes #1951
Closes #1945

### DIFF
--- a/backend/app/modules/catalog/sources/cog_info.py
+++ b/backend/app/modules/catalog/sources/cog_info.py
@@ -131,11 +131,9 @@ async def fetch_cog_info(url: str) -> dict | None:
     any upstream fetch), so ``last_checked_at`` is stamped only on success;
     every failure leaves it NULL for the probe to settle.
 
-    SEC-OBSV-02: dual SSRF gate, both halves required for every caller.
-    Gate 1 (caller-side): ``validate_url_for_ssrf`` before calling this.
-    Gate 2 (Titiler-side): its own CPL_VSIL_CURL_ALLOWED_EXTENSIONS clamp
-    rejects a non-raster extension that slipped past Gate 1. Removing
-    either is an SSRF regression, not a refactor side-effect (#1927).
+    SEC-OBSV-02 (#1927): dual SSRF gate, both halves required. Gate 1
+    (caller-side) is ``validate_url_for_ssrf`` before calling this; Gate 2
+    (Titiler-side) is its own CPL_VSIL_CURL_ALLOWED_EXTENSIONS clamp.
     """
     try:
         async with httpx.AsyncClient(
--- a/backend/tests/finding_markers.py
+++ b/backend/tests/finding_markers.py
@@ -291,7 +291,6 @@ UNANCHORED_MARKER_DEBT: dict[str, int] = {
     "modules/catalog/search/service_records.py": 1,
     "modules/catalog/sources/adapters/stac.py": 3,
     "modules/catalog/sources/classify.py": 4,
-    "modules/catalog/sources/cog_info.py": 1,
     "modules/catalog/sources/origin_probe.py": 2,
     "modules/catalog/sources/preview.py": 3,
     "modules/catalog/sources/probe.py": 7,


ishiland:
@codex review

PR #1995, opened 2026-09-09
fix(db): disable psql's .psqlrc so init stays fail-closed

ishiland:
## Summary

Refs #1992. `scripts/init-db.sh` passes `-v ON_ERROR_STOP=1` to psql, but a `.psqlrc` in the invoking user's home directory can override that setting with `\set ON_ERROR_STOP off`. When that happens, a failed statement no longer aborts the script and it exits 0 as if bootstrap succeeded. This adds `-X` (`--no-psqlrc`) so the script ignores any host `.psqlrc` and stays fail-closed.

## Changes

- Add `-X` to the `psql_args` array in `scripts/init-db.sh`.
- Audited every other `psql` invocation under `scripts/` and `backend/scripts/`. `scripts/lib/configure-runtime-db-role.sh` already carries `-X`, so no change was needed there; added a structural test pinning it so it can't regress.
- Added a structural test asserting `init-db.sh`'s `psql_args` carries `-X` ahead of `ON_ERROR_STOP=1`.
- A few other host-invoked `psql` call sites elsewhere under `scripts/` still lack `-X` (a local dev-database helper and parts of the backup/restore roundtrip test). Those files are outside this PR's scope; flagging for a follow-up pass.

## Area

## Testing

Ran `backend/tests/test_runtime_db_role.py` and `backend/tests/test_layering.py` against the local test database; all pass. Confirmed the new assertion fails against the pre-fix script and passes with the fix. Also demonstrated the underlying psql behavior directly: with a temporary home directory containing `\set ON_ERROR_STOP off` and a deliberately failing statement, psql without `-X` continues past the error and exits 0, while psql with `-X` stops at the error and exits non-zero.

## Checklist

### DIFF
--- a/backend/tests/test_runtime_db_role.py
+++ b/backend/tests/test_runtime_db_role.py
@@ -98,6 +98,14 @@ def test_init_db_is_one_entrypoint_for_container_and_managed_postgres() -> None:
         assert var in source, f"init-db.sh no longer honours {var}"
     assert 'psql "${psql_args[@]}"' in source
 
+    # fix(#1992): a host .psqlrc can `\set ON_ERROR_STOP off`, so -v
+    # ON_ERROR_STOP=1 alone does not guarantee a failed statement aborts
+    # the run; -X must precede it in the same psql_args assignment.
+    args_line = next(
+        line for line in source.splitlines() if line.startswith("psql_args=")
+    )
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

PR #1996, opened 2026-09-09
fix(ci): derive dep-audit's node/python/nginx pins from the Dockerfile

### LINKED ISSUE #1983: The nginx image tag is pinned twice, and dependabot only updates one
`nginxinc/nginx-unprivileged` is pinned in two places:

- `Dockerfile:303` — the `FROM` for the `frontend` stage
- `.github/workflows/dep-audit.yml:119` — the Trivy scan matrix

Dependabot's docker ecosystem updates the `FROM` and leaves the workflow matrix alone, so `backend/tests/test_docker_audit_regressions.py::test_docker_audit_matrix_pins_match_dockerfile_from_tags` fails on the drift and the bump PR sits red until someone pushes a companion commit.

#1975 is the first nginx bump since that gate landed (#1798, 2026-09-03); it needed `0d1455421` by hand. Every future one will need the same.

Two ways out, either is fine:

1. Derive the matrix entry from the Dockerfile at workflow run time, so there is one pin and the test becomes a tautology it can then drop.
2. Keep both pins and teach dependabot about the second: a `github-actions`-adjacent entry will not match a quoted image string inside a matrix, so this likely means a small `update-dep-audit-pin` step or a scheduled job rather than dependabot config.

Option 1 is the smaller surface — the test exists precisely because the second pin can drift, and one pin cannot.

ishiland:
## Summary

docker-audit's Trivy scan matrix kept its own copy of the node/python/nginx image tags next to the Dockerfile's `FROM` lines. Dependabot's docker ecosystem only bumps the Dockerfile, so the matrix drifted behind it — #1975 needed a manual companion commit to catch up, and every future bump would need the same. A test could only catch the drift after it already happened.

Closes #1983.

## Changes

- Add a `resolve-docker-audit-matrix` job to `.github/workflows/dep-audit.yml` that greps the node/python/nginx tags straight out of `Dockerfile` and emits them as a job output; `docker-audit` now consumes that output via `fromJson(...)` instead of a literal `include:` list. `postgis/postgis` stays a static, report-only entry since it has no Dockerfile `FROM` line to derive from.
- Replace `test_docker_audit_matrix_pins_match_dockerfile_from_tags` in `backend/tests/test_docker_audit_regressions.py` with two tests: one asserting `docker-audit` consumes the resolved output rather than a literal list, and one that runs the real derivation step against the Dockerfile and checks its output.

## Area

## Testing

Ran `backend/tests/test_docker_audit_regressions.py` (13 passed), executed the workflow's derivation script directly against the repo's Dockerfile to confirm its output, validated the workflow with `python3 -c 'import yaml; yaml.safe_load(...)'` and `actionlint`, and ran two counterfactuals: reverting `docker-audit`'s matrix to a literal list, and breaking the nginx grep pattern — both fail the corresponding new test, then were reverted.

## Checklist

### DIFF
--- a/.github/workflows/dep-audit.yml
+++ b/.github/workflows/dep-audit.yml
@@ -93,8 +93,36 @@ jobs:
           # npx license-checker --production --onlyAllow \
           #   'MIT;ISC;BSD-2-Clause;BSD-3-Clause;Apache-2.0;CC0-1.0;Unlicense;Python-2.0;PSF-2.0;BlueOak-1.0.0;0BSD;OFL-1.1;MPL-2.0;CUSTOM'
 
+  resolve-docker-audit-matrix:
+    name: Resolve docker-audit image pins
+    runs-on: ubuntu-latest
+    outputs:
+      include: ${{ steps.matrix.outputs.include }}
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+      # fix(#1983): node/python/nginx are Dockerfile FROM tags Dependabot
+      # bumps; deriving them here (not a second literal copy) is the single
+      # source of truth, so this matrix can't drift behind the Dockerfile.
+      - name: Derive matrix from Dockerfile FROM tags
+        id: matrix
+        run: |
+          set -euo pipefail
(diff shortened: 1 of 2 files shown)


ishiland:
@codex review

PR #1998, opened 2026-09-09
fix(gates): give the ponytail: denylist hook a same-line anchor escape

### LINKED ISSUE #1960: The ponytail: denylist hook has no anchor escape hatch, and the tag comes from a tool that is active in this repo
Found while reviewing #1934, from the environment rather than the diff.

#1934 adds `no-agent-tag-markers`, a pre-commit hook that fails any `python/ts/tsx/javascript/shell` file containing the case-insensitive literal `ponytail:`. Its own description records that 14 such markers reached `main` before anyone noticed.

The tag is not a historical accident. The `ponytail` plugin is installed and active in agent sessions on this repo, its instructions tell an agent to *"mark deliberate simplifications with a `ponytail:` comment"*, and it ships a `ponytail-debt` skill whose stated purpose is to harvest those comments into a debt ledger. So a tool that is present and running treats these markers as its own durable output, and its workflow encourages leaving them in place rather than anchoring them.

That makes the hook the only thing standing between an active tool convention and `main`, which is worth stating rather than discovering.

## The part that needs a decision: the two halves disagree

For `backend/app/`, the AST detector in the same PR treats `ponytail` as a marker like any other, so **`# fix(#1234) ponytail: global lock, per-account locks if throughput matters` passes** — anchored, findable, and compatible with a debt ledger that carries an issue number.

For everything else, the denylist half greps the bare literal:

```
grep -nFi "ponytail:" "$f"
```

so the identical anchored comment **fails** in `frontend/`, `e2e/`, `cli/` and `mcp/`. There is no escape hatch outside `backend/app/`.

The result is that the same convention is anchorable in one half of the repo and forbidden outright in the other, and an agent told to leave a `ponytail:` marker cannot comply with both the plugin and the repo unless it happens to be editing backend code.

## Coverage gap

`types_or: [python, ts, tsx, javascript, shell]`. Markers in `.md`, `.sql`, `.yaml` or `.json` pass unchecked. Whether that matters depends on where the skill actually writes, which nobody has established.

## Options

1. Give the denylist half the same anchor escape hatch the AST half has, so `fix(#N) ponytail:` is legal everywhere and a bare tag is legal nowhere. Most consistent, and compatible with `ponytail-debt` provided ledger entries carry an issue number.
2. Leave the asymmetry and document it, on the grounds that non-backend code has no debt map to record an anchored marker in.
3. Widen `types_or`. Cheap, and independent of the choice above.

Not urgent. #1934 is correct as written and this is not a defect in it — the hook does exactly what it claims. What is missing is a decision about how the repo and an active agent tool are supposed to coexist.

ishiland:
Closes #1960.

## Problem

`no-agent-tag-markers` greps the bare `ponytail:` literal outside `backend/app`, with no way to scope it. `no-unscoped-finding-markers` (the AST detector, `backend/app` only) accepts `fix(#1234) ponytail: ...` as a legal anchored marker, so the identical comment passed in `backend/app` and failed everywhere else. An agent following the `ponytail-debt` convention could not satisfy both the plugin's own workflow and this repo outside `backend/app`.

## Fix

- `no-agent-tag-markers`'s entry now requires the same line to also carry a `fix(#<issue>)` anchor. A bare tag still fails everywhere; an anchored one now passes everywhere.
- Widened `types_or` to include `markdown`, `sql`, `yaml` and `json`, which were previously unchecked (issue's "coverage gap").
- Excluded `.pre-commit-config.yaml` itself from the hook. Widening to markdown/yaml made the hook self-referential: its own name/description/entry necessarily name the literal it forbids, and none of those mentions carry a real numeric anchor. That file defines the rule rather than being subject to it.
- One clarifying sentence in AGENTS.md's Contracts and Comments section, since the anchor escape hatch outside `backend/app/` wasn't documented there before.

## Tests

`backend/tests/test_agent_tag_marker_hook.py` runs the hook's actual `entry` command (shlex-split, same way pre-commit invokes it) against fixture files, rather than reimplementing the grep in Python. Covers: bare tag in a `.ts` file fails, the anchored form passes, a bare tag in a `.md` file now fails (the coverage gap), an anchor two lines away does not scope a bare tag on its own line, and one bare line fails even when a sibling line in the same file is anchored. Also pins that `types_or` carries the four new file types alongside the original five.

## Verification

```
All checks passed!

$ uv run pytest tests/test_agent_tag_marker_hook.py tests/test_unscoped_finding_markers.py tests/test_layering.py -q
87 passed

$ python3 backend/tests/finding_markers.py; echo $?
0

$ SKIP=ruff-check,ruff-format,gitleaks pre-commit run --all-files
AGENTS.md — a `ponytail:` tag needs a same-line fix(#N) anchor outside backend/app ... Passed
(all other hooks Passed)
```

Negative control, run against the tree before the fix (`git stash`) and reverted: the anchored form (`fix(#1960) ponytail: ...`) failed in a `.ts` fixture, and `markdown` was absent from `types_or`, reproducing the issue exactly. Both are fixed on this branch.

No CHANGELOG entry: internal contributor gate, no user-facing behavior.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -114,30 +114,34 @@ repos:
         files: '^backend/(app/.*|tests/finding_markers)\.py$'
 
       - id: no-agent-tag-markers
-        name: "AGENTS.md — no `ponytail:` literal anywhere in a python/ts/tsx/js/shell file"
+        name: "AGENTS.md — a `ponytail:` tag needs a same-line fix(#N) anchor outside backend/app"
         description: >
           The denylist half of the same rule, for the file types the AST
-          detector above does not parse (frontend/, e2e/, cli/, mcp/). One
-          case-insensitive literal, grepped over the whole file rather than
-          its comments — that is all this hook checks and all it claims. The
-          general class is enforced for backend/app, and only there, by
-          no-unscoped-finding-markers.
+          detector above does not parse (frontend/, e2e/, cli/, mcp/, plus
+          markdown/sql/yaml/json anywhere in the tree). A case-insensitive
+          literal grepped over the whole file rather than its comments — that
+          is all this hook checks and all it claims. A `ponytail:` tag is
+          legal only when the same line also carries a `fix(#<issue>)`
+          anchor; a bare tag fails everywhere. The general class is enforced
+          for backend/app, and only there, by no-unscoped-finding-markers.
         language: system
         entry: >-
           bash -c '
           rc=0;
           for f in "$@"; do
-            if grep -nFi "ponytail:" "$f"; then
+            if grep -nFi "ponytail:" "$f" | grep -viE "fix\(#[0-9]+\)"; then
               echo "FAIL: $f has an unscoped finding marker (see above).";
-              echo "  Use fix(#<issue>): <context>, or remove the tag.";
+              echo "  Use fix(#<issue>) ponytail: <context> on the same line, or remove the tag.";
               echo "  See AGENTS.md > Inline review-comment convention.";
               rc=1;
             fi;
           done;
           exit $rc;
           ' --
-        types_or: [python, ts, tsx, javascript, shell]
-        exclude: '^(sdks/|frontend/node_modules/)'
+        types_or: [python, ts, tsx, javascript, shell, markdown, sql, yaml, json]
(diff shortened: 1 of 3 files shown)


ishiland:
@codex review

PR #2003, opened 2026-09-09
refactor(sandbox): bind tenant schemas by replacing schema spans

### LINKED ISSUE #1892: refactor(sandbox): bind tenant schemas without re-rendering SQL

ishiland:
## Summary

Multi-tenant execution bound the logical `data` schema by re-serializing the parsed
statement, so what ran was not what the caller wrote. Comments, `AS`, `E'...'`
literals and casts were normalized, and the pgvector cosine operator needed a
sentinel swap to survive the round trip. This parses once, takes the source offsets
of the `Table` and `Column` schema identifiers, and replaces only those spans in the
original text. Single-tenant already executes the text as written, so the two modes
now agree and single-tenant coverage starts to transfer.

Design and evidence come from the 2026-09-05 tenant SQL routing investigation.

## Changes

- `_rewrite_logical_data_schema` replaces proven schema-identifier spans right to
  left instead of rendering the tree. `_swap_operator_tokens`, the `<=>` sentinel and
  the `sqlglot.tokens` import are gone.
- Each span is proven against the text before anything is spliced: the slice must
  spell the identifier sqlglot parsed, and a span overlapping one already replaced is
  refused. Both raise `SandboxError` before a connection is opened.
- The validator cuts the whole trailing run of `;` at the first of those tokens. The
  old re-render dropped every terminator; preserving one would leave it inside the
  executor's `SELECT * FROM (...) AS _q LIMIT n` wrapper, which is the syntax error
  single-tenant already gives. `;;` and `; ;` are ordinary paste damage and go the same
  way. A `;` inside a literal or a comment is left alone, a bare `;;` reduces to nothing,
  and a second statement is still rejected.
- The rewrite parses with `parse`, not `parse_one`. `parse_one` reports a second
  statement as one `Block`, so a direct caller's `a; b` passed through byte-identical
  with both halves rewritten. Anything but one statement is now refused. Not
  exploitable before this (the wrapper is a syntax error and asyncpg refuses multi
  statement), but `execute_safe` documents itself as fail-closed for direct callers.
- Unchanged: the public `data.*` contract, single-tenant behaviour, RBAC and the
  dataset allowlist, role binding, the read-only transaction, timeout, concurrency
  lock and row cap. The unquoted-`DATA` folding from #1891 stays and quoted `"DATA"`
  is still refused. No API, snapshot or SDK regeneration, since the contract did not
  move.

### A defect this removes

The sentinel restore swapped every `&&` back to `<=>` whenever a real cosine operator
was present, so a PostGIS bbox overlap beside one became a distance operator. It was
not reachable: the validator rejects `&&`, so no validated query carried both, and the
direct analysis callers that emit `&&` carry no cosine operator. It is now a
regression case with an exact-text expectation.

### Security notes

The isolation boundary is the per-tenant reader role, which holds USAGE on its own
schema only, inside a read-only transaction with a statement timeout and a row cap.
The rewrite routes; it does not authorize. A cross-tenant statement is refused by the
role, and that is asserted against the test database with a same-tenant positive
control beside it.

Executing the caller's own bytes rather than a re-render means a place where sqlglot
and PostgreSQL disagree is no longer papered over by serialization. That is the
property single-tenant has had all along, and the runtime floor above bounds it. The
classic differentials were checked against sqlglot 30.18 and the postgres dialect
agrees: backslashes are not escapes in ordinary strings, `E'...'` escapes are, block
comments nest, and `U&'...'` stays a literal.

Independent review probed this with 37 hand-built shapes plus a 14k-case fuzz over
comment styles, E-strings, dollar quoting, `U&` literals, CRLF, tabs and emoji, and
seven spellings of the schema reference, with zero failures. Both parses see the same
stripped string, and the allowlist folding matches the span folding.

Known and carried from #1892: sqlglot still rejects `<#>` before execution, and
`(data.roads).marker` is a blind spot for a span rewrite and for the old one alike,
failing closed under the tenant role.

## Area

- [x] Auth / Permissions

## Testing

Backend, from the worktree: `tests/test_sandbox_tenant_schema.py`, `tests/test_sandbox.py`,
`tests/test_analysis_preview.py`, `tests/test_dp02_read_path_binding.py`,
`tests/test_dp03_cross_tenant_privilege_gate.py`, `tests/test_dp_single_tenant_byte_identical.py`,
`tests/test_sandbox_query_endpoint_565.py`, `tests/test_layering.py`,
`tests/test_rule1_structural.py`, `tests/test_rule2_structural.py`, and the AI/SQL
suites that consume the sandbox. All green. `ruff check`, `ruff format --check` and
`python3 backend/tests/finding_markers.py` pass. `_MODULE_LOC_CAPS` for the validator
is re-measured to 1687, exact.

New coverage, each with the failure observed on `main`:

- Fifteen exact-text span cases: quoted and unquoted schema and table names, three-part
  columns, mixed-case quoted tables, single-quoted and dollar-quoted literals spelling
  `data.roads` and operators, an `E'...'` literal, Unicode ahead of the target with a
  comment inside the reference and a trailing one, cosine beside L2, a real
  `IS NOT DISTINCT FROM` beside cosine, a bbox overlap beside cosine, a CTE named after
  a dataset, a CTE named `data`, a self-join, and quoted `"DATA"`. Seven of the fifteen
  fail on `main`, on normalized comments, re-cased `E'...'`, inserted `AS`, rewritten
  casts and the `&&` corruption.
- Unusable source offsets (missing, shifted, out of range) raise `SandboxError`. On
  `main` the rewrite has no offset path, so all three pass through.
- The analysis preview statement and the bbox count statement, rewritten from logical
  refs, equal what the physical schema renders. Fails on `main`.
- A validated `;`, `;;` and `; ;` each run inside the LIMIT wrapper against the test
  database, with the raw statement as the control. On `main` the wrapper gets
  `SELECT 1 AS a;` and PostgreSQL answers "syntax error at or near ;".
- `;;`, `; ;`, `;;;` and `;; -- pasted` validate to the exact statement text, a `;` in a
  literal or comment is kept, and `;`, `;;` and a second statement are rejected.
- The rewrite refuses `SELECT 1; DROP TABLE data.alpha` and
  `SELECT id FROM data.alpha; SELECT * FROM data.beta` with no connection opened.
- Against the test database: a statement naming another tenant's schema is refused
  under tenant A and returns its row under tenant B, and the awkward shape above
  (unquoted `DATA`, an inline comment, a non-ASCII literal, a trailing line comment)
  executes and reads only the active tenant's row. Both hold on `main` as well; they
  are database-backed evidence for the preserved-text form, not regression guards.
- An unparseable statement raises `SandboxError` with no connection opened. Holds on
  `main` too; it pins acceptance criterion 2.

## Checklist

Closes #1892

### DIFF
--- a/backend/app/platform/sandbox/executor.py
+++ b/backend/app/platform/sandbox/executor.py
@@ -30,24 +30,19 @@ DEFAULT_TIMEOUT_MS = 10_000
 _SINGLE_TENANT_READER_ROLE = "geolens_reader"
 
 
-def _logical_data_span(
-    identifier: exp.Expression | None, sql: str
-) -> tuple[int, int] | None:
-    """Source offsets of one logical ``data`` schema qualifier in ``sql``.
+def _is_logical_data_schema(identifier: exp.Identifier) -> bool:
+    """True when ``identifier`` folds to the logical ``data`` schema: unquoted
+    folds to lowercase, quoted keeps its case, so quoted ``"DATA"`` is not it."""
+    name = identifier.name if identifier.quoted else identifier.name.lower()
+    return name == "data"
 
-    ``None`` when the qualifier names another schema. An unquoted identifier
-    folds to lowercase as PostgreSQL folds it, so every spelling of ``data``
-    is the logical schema while quoted ``"DATA"`` stays a different one
-    (fix(#1891)).
+
+def _logical_data_span(identifier: exp.Identifier, sql: str) -> tuple[int, int]:
+    """Source offsets of one logical ``data`` schema qualifier in ``sql``.
 
     fix(#1892): the slice must spell exactly what sqlglot parsed. An absent or
     shifted offset would move an unrelated span, so it fails closed instead.
     """
-    if not isinstance(identifier, exp.Identifier):
-        return None
-    if (identifier.name if identifier.quoted else identifier.name.lower()) != "data":
-        return None
-
     start = identifier.meta.get("start")
     end = identifier.meta.get("end")
     spelling = f'"{identifier.name}"' if identifier.quoted else identifier.name
@@ -57,7 +52,9 @@ def _logical_data_span(
         or not 0 <= start <= end < len(sql)
         or sql[start : end + 1] != spelling
     ):
-        logger.error("sandbox.schema_span_unusable", sql=sql)
+        logger.warning(
+            "sandbox.schema_span_unusable", start=start, end=end, length=len(sql)
+        )
         raise SandboxError("query_failed", "Query failed")
     return start, end
 
@@ -80,26 +77,36 @@ def _rewrite_logical_data_schema(sql: str, physical_schema: str) -> str:
     only a normalized UUID-derived identifier in multi-tenant mode.
     """
     try:
-        statement = sqlglot.parse_one(sql, dialect="postgres")
+        statements = sqlglot.parse(sql, dialect="postgres")
     except sqlglot.errors.SqlglotError as exc:
         # execute_safe receives validated SQL in normal operation.  Keep direct
         # callers fail-closed if that contract is accidentally violated (covers
         # both tokenize and parse failures).
         raise SandboxError("invalid_query", "Invalid SQL syntax") from exc
 
+    # fix(#1892): `parse`, not `parse_one`, which reports a second statement as
+    # one Block and would rewrite a caller-supplied `a; b` in both halves.
+    statements = [statement for statement in statements if statement is not None]
+    if len(statements) != 1:
+        logger.warning("sandbox.rewrite_multi_statement", count=len(statements))
+        raise SandboxError("invalid_query", "Only single statements are allowed")
+
     spans: set[tuple[int, int]] = set()
-    for node in statement.walk():
-        if isinstance(node, (exp.Table, exp.Column)):
-            span = _logical_data_span(node.args.get("db"), sql)
-            if span is not None:
-                spans.add(span)
+    for node in statements[0].walk():
+        if not isinstance(node, (exp.Table, exp.Column)):
+            continue
+        identifier = node.args.get("db")
+        if isinstance(identifier, exp.Identifier) and _is_logical_data_schema(
+            identifier
+        ):
+            spans.add(_logical_data_span(identifier, sql))
 
     replacement = '"' + physical_schema.replace('"', '""') + '"'
     tail = len(sql)
     for start, end in sorted(spans, reverse=True):
         if end >= tail:
             # Overlapping spans would splice into text already replaced.
-            logger.error("sandbox.schema_span_overlap", sql=sql)
+            logger.warning("sandbox.schema_span_overlap", start=start, end=end)
             raise SandboxError("query_failed", "Query failed")
         sql = sql[:start] + replacement + sql[end + 1 :]
         tail = start
--- a/backend/app/platform/sandbox/executor.py
+++ b/backend/app/platform/sandbox/executor.py
@@ -10,7 +10,6 @@ from __future__ import annotations
 import structlog
 import sqlglot
 from sqlglot import exp
-from sqlglot.tokens import TokenType
 from sqlalchemy import text
 from sqlalchemy.exc import DataError, InternalError
 from sqlalchemy.ext.asyncio import AsyncSession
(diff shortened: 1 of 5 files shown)


ishiland:
@codex review

PR #2009, opened 2026-09-09
fix: host psql -X gaps and Azure managed identity docs

ishiland:
## Summary

Two small wave leftovers: a host `.psqlrc` gap left over from #1995, and an
`.env.example` claim that Azure managed identity works when it does not.

## Changes

- `scripts/init-test-db.sh` and the admin/direct `psql` calls in
  `scripts/tests/test-backup-restore-roundtrip.sh` now pass `-X`, matching
  `init-db.sh` and the runtime-role reconciler. Without it, a host
  `.psqlrc` that sets `\set ON_ERROR_STOP off` can let a failed statement
  in either script pass silently.
- `.env.example`'s Azure Blob Storage block claimed setting
  `AZURE_STORAGE_ACCOUNT_URL` alone enables managed identity / Entra ID
  auth. It does not: `azure-identity` is deliberately not a backend
  dependency (`backend/app/platform/storage/azure.py`, fix(#836)), so that
  configuration authenticates anonymously and fails at the first object
  read instead of at boot. The block now documents the two forms the
  backend actually supports: a connection string, or an account URL paired
  with an account key. Checked `RUNBOOK.md`, `README.md` and
  `backend/app/core/config.py` for the same claim; none carried it.

## Area

- [x] Docs / Config

## Testing

- [x] No tests needed (docs, config, etc.), for the `.env.example` change

Added `test_init_test_db_disables_host_psqlrc` and
`test_roundtrip_every_host_psql_call_disables_host_psqlrc` to
`backend/tests/test_runtime_db_role.py`, extending the pattern
`test_role_script_disables_host_psqlrc` already uses. Both fail against the
pre-fix scripts and pass after.

```
cd backend && uv run pytest tests/test_runtime_db_role.py -q
44 passed
```

Also ran `uv run ruff check . && uv run ruff format --check .`,
`pytest tests/test_layering.py -q`, `python3 backend/tests/finding_markers.py`,
`git diff --check`, and `make openapi-check` (clean, since `.env.example`
is not published API surface), plus `shellcheck` on both changed scripts
(no new findings).

## Checklist

Refs #1992, Refs #836

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -653,10 +653,12 @@ FRONTEND_PORT=8080
 
 # --- Azure Blob Storage (cloud-dev or production) ---
 # [OPTIONAL] Required when STORAGE_PROVIDER=azure.
-# Use either AZURE_STORAGE_CONNECTION_STRING (for Azurite or a full connection string)
-# OR AZURE_STORAGE_ACCOUNT_URL + AZURE_STORAGE_ACCOUNT_KEY (for production key-auth).
-# Managed identity / Entra ID auth: set AZURE_STORAGE_ACCOUNT_URL and leave
-# AZURE_STORAGE_CONNECTION_STRING and AZURE_STORAGE_ACCOUNT_KEY unset.
+# Two supported forms only:
+#   1. AZURE_STORAGE_CONNECTION_STRING (Azurite dev or a full connection string)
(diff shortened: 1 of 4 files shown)


ishiland:
@codex review

Issue #1955, opened 2026-09-06
ADR-002 Decision 5b's single-mutation admission control does not cover VRT regeneration


ishiland:
Found while reconstructing ADR-002 in #1952.

ADR-002 Decision 5b admits one mutation per dataset at a time, rejecting a second with 409 `dataset_busy`, enforced by a partial unique index over the pending run rows.

VRT regeneration is not covered by it. It mutates a dataset and dates `last_refreshed_at` under Decision 5a, but it does not call `create_pending_run`, and `catalog.vrt_generations` carries no single-active partial unique index of its own.

So the admission control that Decision 5b describes as covering dataset mutation has a mutation path outside it.

## What this is and is not

This is not a report of a live corruption. #1938 (fixed in #1947) showed the VRT publish takes the dataset and record rows under `FOR UPDATE`, so two concurrent regenerations serialise on row locks rather than interleaving. The gap is that they serialise by waiting instead of one being refused, and that a caller gets no `dataset_busy` answer telling it to stop.

That matters more since #1947 bounded the wait: a second regeneration that would previously have waited now fails after the budget, with a lock-conflict error rather than the 409 the API contract describes for this situation.

## What to decide

Either bring VRT regeneration under Decision 5b's admission control — a pending run row and the partial unique index, so a second attempt is refused rather than queued — or record in ADR-002 that VRT regeneration is a deliberate exception and say what serialises it instead.

The first is more work and makes the decision true as written. The second is honest and cheap. Do not leave the document claiming coverage it does not have.

ADR-002 now lives at `.github/ADR-002.md` (#1952).

PR #2020, opened 2026-09-09
fix(ingest): let the fan-out restore reclaim a parent the sweep settled (#2016)

### LINKED ISSUE #2016: A fan-out parent is left permanently un-retryable when the sweep settles it mid-dispatch
Found while enumerating the shared stale sweeper's behaviour per job type for #1556. Not a defect in that work; filing separately.

## What happens

`commit_fan_out` flips the parent to `fanned_out` and commits it before the first child exists, deliberately, because that transition is the mutex for the whole dispatch (`backend/app/processing/ingest/router.py:1427`). The layer loop then runs, and if every layer failed to queue, `restore_fan_out_parent_pending` undoes the flip so the user can commit again without re-uploading (`backend/app/processing/ingest/router.py:1451`).

That restore is a fenced compare-and-swap on `status == "fanned_out"` plus the attempt id (`backend/app/processing/ingest/service.py:1131`). The stale sweeper reaches the same row first if the loop runs long enough:

- `commit_fan_out` sets `completed_at = now()` (`backend/app/processing/ingest/service.py:1106`).
- The childless-`fanned_out` sweep matches a `fanned_out` parent whose `completed_at` is older than `FAN_OUT_CHILDLESS_GRACE_SECONDS`, 300 seconds, with no child row referencing it (`backend/app/platform/jobs/sweep.py:1189` and `:53`).
- It writes `status = "failed"` and stamps `FAN_OUT_INTERRUPTED_METADATA_KEY` (`backend/app/platform/jobs/sweep.py:1216`).

## The sequence

1. A fan-out commit flips the parent to `fanned_out` and commits.
2. Every layer's dispatch fails, for example during a Procrastinate outage where each `defer_async` waits out a connection timeout. With enough layers the loop exceeds 300 seconds.
3. A sweeper pass, which runs every 5 minutes, sees a childless `fanned_out` parent past the grace and writes `failed` plus the interrupted marker.
4. The loop finishes. `restore_fan_out_parent_pending` compare-and-swaps on `fanned_out`, matches zero rows, and returns. It does not check `rowcount` and logs nothing.
5. The request still returns 200 with every layer reported failed.

## Why it matters

The parent is left `failed` carrying `FAN_OUT_INTERRUPTED_METADATA_KEY`, and `_retry_capability` refuses that marker outright (`backend/app/platform/jobs/router.py:452`). The user is told to re-upload the file and select its layers again, which is exactly the outcome `restore_fan_out_parent_pending` exists to prevent. The restore's silence is what makes it hard to see: nothing records that the undo was lost.

Note the sweep's own comment says it writes `failed` "so /jobs/{id}/retry offers it", while the marker it writes in the same statement makes retry refuse. That is #1709's deliberate design for a genuinely interrupted dispatch, but this row is not one: the dispatch ran to completion and simply failed everywhere.

## Fix shape

Two parts, both small:

1. Have `restore_fan_out_parent_pending` return whether it matched and log when it did not, so a lost undo stops being silent.
2. Let the restore reclaim the row the sweep settled. It is the same row and the same attempt, and the sweep only reached it because this dispatch was still running, so widening the compare-and-swap to also accept `status == "failed"` carrying the interrupted marker for this attempt, clearing the marker as it restores, is sound. Fencing on the attempt id is what keeps it from resurrecting anything else.

Raising the grace or heartbeating the parent through the loop would narrow the window rather than close it, since the loop has no upper bound.

## Reachability

Needs a dispatch loop that runs past 300 seconds with zero successful layers, so a queue outage with slow defer timeouts and a multi-layer file. Not reachable in normal operation.

ishiland:
## Summary

`commit_fan_out` flips the parent to `fanned_out` and commits before the first child exists, because that transition is the mutex for the whole dispatch. If every layer then fails to queue, `restore_fan_out_parent_pending` undoes the flip so the user can commit again without re-uploading.

A dispatch loop that runs past the 300 second childless-fanout grace loses that undo. The stale sweep reaches the same row first, writes `failed` and stamps the interrupted marker. The restore's compare-and-swap matched only `fanned_out`, so it matched zero rows, ignored the rowcount and logged nothing, and the request still returned 202. The marker it left behind is the one `_retry_capability` refuses on, so the user was told to re-upload the file and pick its layers again, which is the outcome the restore exists to prevent.

The restore now returns whether it matched, and the compare-and-swap also accepts the row the sweep settled for this same attempt. Fencing on the attempt id is what makes the wider predicate safe: a `failed` row belonging to another attempt is never reclaimed. Raising the grace or heartbeating the parent through the loop would only narrow the window, since the loop has no upper bound.

## Changes

- `restore_fan_out_parent_pending` returns its CAS rowcount, and the fan-out door logs `fan_out_parent_restore_missed` with the job and attempt id when the undo matched nothing.
- The CAS accepts `fanned_out`, or `failed` carrying the interrupted marker for the same attempt id. It clears that one key off the row's own metadata with the jsonb `-` operator, so the layer list and everything else the dispatch wrote survives, and it clears the sweep's `error_message` the way generic retry does.
- Corrected the sweep's own comment: it said the row settles `failed` "so /jobs/{id}/retry offers it", while the marker written in the same statement makes retry refuse.
- Four tests in `backend/tests/test_job_cancel_fan_out.py`, plus exact `_MODULE_LOC_CAPS` updates for the two modules that changed size.

## Area

- [x] Import / Ingestion

## Testing

Each new test was counterfactual-verified against the pre-fix code:

- Narrowing the CAS back to `fanned_out` only fails `test_swept_parent_is_reclaimed_on_the_same_attempt`. That test drives the real sweep over a claimed parent aged past the real grace constant, so the marker under test is the one the sweep actually writes.
- Dropping the attempt fence fails `test_a_failed_row_from_another_attempt_is_refused` and `test_a_lost_undo_is_logged_with_the_job_and_attempt`.
- Dropping the caller's warning fails `test_a_lost_undo_is_logged_with_the_job_and_attempt`, which drives the endpoint end to end and reads the record with `structlog.testing.capture_logs`.

Suites run on the host: `test_job_cancel_fan_out.py`, `test_ingest_fan_out.py`, `test_ingest_job_attempt_fencing.py`, `test_stale_pending_reaper.py`, `test_jobs_router.py`, `test_admin_jobs_retry_capability.py`, `test_commit_attempted_marker_doors.py`, `test_rule1_structural.py`, `test_vrt_stale_sweep_gap002.py`, `test_stalled_queue_sweep_624.py`, `test_terminal_job_retention_1778.py`, `test_failed_job_token_purge_1746.py`, `test_tenant_job_recovery.py`, `test_defer_orphan_guard.py`, `test_layering.py`. Also `ruff check`, `ruff format --check` and `backend/tests/finding_markers.py`, all re-run after the rebase across #1955.

## Checklist

- [x] i18n: new user-facing strings added to all 4 locales (en, fr, es, de)
- [x] Migration included (if DB schema changed)
- [x] Release/docs deploys: ran `make deployed-surface-check` after the marketing/docs deploy and recorded the result in Testing

No published API shape changed, so no SDK or OpenAPI regeneration. No new user-facing strings, no schema change, no frontend change.

Closes #2016

### DIFF
--- a/backend/app/platform/jobs/sweep.py
+++ b/backend/app/platform/jobs/sweep.py
@@ -1183,9 +1183,9 @@ async def fail_stale_jobs(
             error_code="never_started",
         )
 
-    # fix(#1709): a childless `fanned_out` parent past the grace, inside
-    # the retention horizon, is the signature of a dispatch interrupted
-    # pre-commit. `failed`, so /jobs/{id}/retry offers it.
+    # fix(#1709): a childless `fanned_out` parent past the grace, inside the
+    # retention horizon, is the signature of a dispatch interrupted pre-commit.
+    # fix(#2016): a dispatch still looping reclaims this row on its way out.
     childless_fanout_clauses = [
         IngestJob.status == "fanned_out",
         IngestJob.completed_at.is_not(None),
--- a/backend/app/processing/ingest/router.py
+++ b/backend/app/processing/ingest/router.py
@@ -1451,9 +1451,19 @@ async def commit_fan_out(
     # success keeps the parent `fanned_out`.
     queued_count = sum(1 for r in results if r.status == "queued")
     if queued_count == 0:
-        await restore_fan_out_parent_pending(
+        restored = await restore_fan_out_parent_pending(
             db, job, parent_attempt_id=parent_attempt_id
         )
+        if not restored:
+            # fix(#2016): the undo matched no row, so a third writer holds the
+            # parent terminal and this 202 hides it. Never silent again.
+            logger.warning(
+                "fan_out_parent_restore_missed",
+                job_id=str(job.id),
+                attempt_id=(
+                    str(parent_attempt_id) if parent_attempt_id is not None else None
+                ),
+            )
 
     return FanOutCommitResponse(fan_out_id=job.id, results=results)
 
--- a/backend/app/processing/ingest/service.py
+++ b/backend/app/processing/ingest/service.py
@@ -16,7 +16,7 @@ from typing import Any
 
 import structlog
 from fastapi import HTTPException, UploadFile, status
-from sqlalchemy import or_, select, text
+from sqlalchemy import String, and_, literal, or_, select, text
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from app.core.async_io import run_in_thread_draining
@@ -51,7 +51,11 @@ from app.platform.jobs.heartbeat import (
     ATTEMPT_STAGING_NAME_PATTERN,
     is_attempt_scoped_staging_table,
 )
-from app.platform.jobs.models import IngestJob, commit_attempted_marker
+from app.platform.jobs.models import (
+    FAN_OUT_INTERRUPTED_METADATA_KEY,
+    IngestJob,
+    commit_attempted_marker,
+)
 from app.platform.storage.titiler_url import resolve_current_storage_key
 
 logger = structlog.get_logger(__name__)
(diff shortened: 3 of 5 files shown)


ishiland:
@codex review

ishiland:
@codex review

PR #2021, opened 2026-09-09
fix(ingest): bound the manifest admit step inside its reservation lease (#2017)

### LINKED ISSUE #2017: A manifest reservation's admit step runs without a bound inside its own lease
Found while enumerating the shared stale sweeper's behaviour per job type for #1556. Not a defect in that work; filing separately.

## What happens

A manifest entry's reservation row is inserted as `running` with `started_at` set and no heartbeat (`backend/app/processing/ingest/manifest_service.py:713`). That is deliberate: a `pending` row with no file path and no queue task is reapable by four actors while its source is still downloading, and the fixed lease is not. Nothing renews the lease, so the row's whole life as `running` is judged against one timestamp.

The running sweep matches `status == "running" AND coalesce(heartbeat_at, started_at) < now - JOB_TIMEOUT_SECONDS`, where `JOB_TIMEOUT_SECONDS` is 3600 (`backend/app/platform/jobs/sweep.py:1139` and `:50`).

Only the download is bounded. `_stage_source_if_needed` wraps `_stage_source` in `asyncio.timeout(MANIFEST_STAGE_MAX_SECONDS)`, which is `JOB_TIMEOUT_SECONDS // 2`, so 1800 seconds (`backend/app/processing/ingest/manifest_service.py:322` and `:73`). The steps that follow it carry no deadline of their own and run while the row is still heartbeat-less `running`:

- `_admit_staged_source` (`backend/app/processing/ingest/manifest_service.py:913`), which does quota accounting and dataset admission.
- `bind_reservation_to_staged_source` (`backend/app/processing/ingest/manifest_service.py:923`), the statement that finally moves the row to `pending` and stamps `staged_at`.

So the residual margin between the staging deadline and the lease is 1800 seconds, and it is the only thing bounding the admit.

## Consequences

The failure is detected rather than silent. `bind_reservation_to_staged_source` compare-and-swaps on `status == "running"` and the downloading stage marker (`backend/app/processing/ingest/manifest_reservation.py:141`), so a swept row makes it return false, the caller raises `RESERVATION_LOST_MESSAGE`, and `_settle_staged_entry` closes the entry. The user loses a download that had already succeeded and sees a reservation-lost error instead of the real reason.

## Fix shape

Either give the reservation a heartbeat for the admit window, which makes the lease mean what it says for the whole time the row is `running`, or bound the admit explicitly and derive that bound from the lease the way the staging deadline already is, so the two cannot drift apart. The second is smaller and matches the existing pattern at `manifest_service.py:73`.

A cheap intermediate step, worth doing either way: when `bind_reservation_to_staged_source` returns false, log the observed status so an operator can tell a sweep from a cancel.

## Reachability

Needs an admit slower than the residual margin, so a very large dataset creation or a heavily contended quota path after a download that already used most of its own budget. Low severity, and the impact is one failed manifest entry rather than corruption.

ishiland:
## Summary

Bounds the manifest apply's admit-plus-bind step so it can't outlast the reservation's running lease. Only the download stage carried a deadline; the admit and bind steps that follow ran unbounded against a heartbeat-less `running` row. If the sweep reaped the row before the bind's compare-and-swap ran, a completed download surfaced as a generic reservation-lost error instead of the real cause.

Closes #2017

## Changes

- Add `MANIFEST_ADMIT_MAX_SECONDS`, derived from the same lease `MANIFEST_STAGE_MAX_SECONDS` already uses, and wrap the admit-plus-bind step in `asyncio.timeout` so `stage + admit` stays under the lease with margin.
- On timeout, settle the entry the same way the existing staging timeout does, through the same `ManifestSourceError` shape.
- Log the observed row status whenever `bind_reservation_to_staged_source`'s compare-and-swap misses, so an operator can tell a sweep from a cancel.

## Area

- [x] Import / Ingestion

## Testing

- [x] No tests needed (docs, config, etc.) — n/a, new tests added below

New tests in `backend/tests/test_manifest_reservation_1814.py`:
- `MANIFEST_STAGE_MAX_SECONDS + MANIFEST_ADMIT_MAX_SECONDS < JOB_TIMEOUT_SECONDS`
- An admit that exceeds its bound settles the entry failed with the same message shape a staging timeout produces (counterfactual: reverting the timeout wrap makes this test fail)
- A CAS miss on the bind logs the observed row status at warning level

Verification run: `ruff check`, `ruff format --check`, `tests/test_layering.py`, `finding_markers.py`, `tests/test_rule1_structural.py`, `tests/test_manifest_reservation_1814.py`, `tests/test_manifest_apply_service.py`, `tests/test_manifest_apply_api.py`, `tests/test_manifest_apply_roundtrip.py`, `tests/test_manifest_apply_vrt.py`, `tests/test_stalled_queue_sweep_624.py`, `tests/test_stale_pending_reaper.py` — all pass.

## Checklist

- [x] i18n: no new user-facing strings
- [x] TypeScript compiles without errors (no frontend changes)
- [x] Migration included (if DB schema changed) — n/a, no schema change

### DIFF
--- a/backend/app/processing/ingest/manifest_reservation.py
+++ b/backend/app/processing/ingest/manifest_reservation.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 import uuid
 from datetime import datetime, timedelta, timezone
 
+import structlog
 from sqlalchemy import desc, func, select, text, update
 from sqlalchemy.dialects.postgresql import JSONB
 from sqlalchemy.ext.asyncio import AsyncSession
@@ -16,6 +17,8 @@ from sqlalchemy.orm.attributes import set_committed_value
 
 from app.platform.jobs.models import IngestJob
 from app.platform.jobs.sweep import JOB_TIMEOUT_SECONDS
+
+log = structlog.get_logger()
 
 # fix(#1814): the pre-queue stage a manifest job is in. This module's exits
 # clear it; a row the running sweep or worker recovery settles keeps it, which
@@ -156,6 +159,16 @@ async def bind_reservation_to_staged_source(
         .execution_options(synchronize_session=False)
     )
     if not result.rowcount:
+        # fix(#2017): distinguishes a sweep reaping the row from a cancel,
+        # for the same job the CAS just missed on.
+        observed = (
+            await db.execute(select(IngestJob.status).where(IngestJob.id == job.id))
+        ).scalar_one_or_none()
+        log.warning(
+            "Manifest reservation bind missed its CAS",
+            job_id=str(job.id),
+            observed_status=observed,
+        )
         return False
     # fix(#1814): `set_committed_value`, not assignment — a dirty attribute
     # would have the caller's own commit flush a second, unfenced update.
--- a/backend/app/processing/ingest/manifest_service.py
+++ b/backend/app/processing/ingest/manifest_service.py
@@ -72,6 +72,10 @@ _WRITE_BUFFER_BYTES = 4 * 1024 * 1024
 # reservation is judged by, so the budget and its judge cannot drift apart.
 MANIFEST_STAGE_MAX_SECONDS = JOB_TIMEOUT_SECONDS // 2
 
+# fix(#2017): total wall clock for admitting an already-staged entry against
+# quota and binding its reservation. stage + admit < lease, with margin.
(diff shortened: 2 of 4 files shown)


ishiland:
@codex review

PR #2022, opened 2026-09-09
fix(jobs): purge a stalled queue row's token when the sweep fails it

ishiland:
## Summary

Item 12 of #1755. A raw service token rides in `procrastinate_jobs.args` on
the default install, where no credential store is configured. Two purges
already exist: the task strips its own row on the way out
(`purge_token_on_failure`), and `purge_terminal_job_tokens` sweeps terminal
rows once per API sweeper pass.

Neither covers a crashed worker. The task-side purge needs the process to be
alive to run, and the periodic backstop is gated on
`status NOT IN ('todo', 'doing')`, so it skips the row until something else
makes it terminal. That something is the stalled-queue sweep in the worker,
which is also the first moment the token is provably dead weight: both service
tasks are `retry=0`, so nothing re-runs from those args. It now purges the rows
it just failed.

## What the args can carry

| Field | Durable | Handled |
| --- | --- | --- |
| `token` | Yes, the raw service credential on a store-less install | This PR, plus the two existing purges |
| `credential_ref` | No. The secret lives in Valkey under a random reference with `CREDENTIAL_TTL_SECONDS`, claimed once by `GETDEL`. The reference names nothing once claimed or expired | Left in place, see below |
| `source_url` | Yes, and a submitted URL can carry credentials in userinfo or the query | Not touched, see below |

## Changes

- `platform/jobs/sweep.py`: `purge_queue_row_tokens`, one by-id statement.
- `platform/jobs/worker.py`: `fail_stalled_queue_jobs` collects the ids it
  transitions and purges them. A job skipped because its ingest row is still
  leasing keeps its args, since it may still be running.
- The purge is best-effort and logs on failure. The periodic backstop still
  covers those rows, so a purge that raises must not cost the transition the
  sweep just made.
- `processing/ingest/tasks_common.py`: the task-side purge calls the shared
  statement instead of keeping a second copy of the same SQL.
- LOC caps re-measured exactly: sweep.py 1472 to 1494, tasks_common.py 2421 to
  2413.

## Which window remains

Detection, not cleanup. A worker counts as dead after `stalled_worker_seconds()`
of heartbeat silence, floored at the 60 minute `JOB_TIMEOUT_SECONDS`, and the
worker sweeps on a 60 second interval. So the token now clears within about a
minute of the row being declared dead, where before it waited a further API
sweeper cadence on top of that.

That hour is not reducible here. The threshold is global and deliberately
generous because a silent worker does not imply dead work: a split-queue fleet's
raster pool can configure a longer graceful shutdown than this threshold, and
failing a job that is provably still running loses work. `_ingest_jobs_still_leasing`
already prefers a live lease over any timeout for the same reason.

## Deferred, with reasons

- **`credential_ref`** is not purged. It is a reference, not a secret: the value
  lives in Valkey under a short TTL and `GETDEL` makes it single use, so a
  residual reference on a dead row names a key that is already claimed or
  expired. Purging it would buy nothing durable.
- **`source_url`** is not purged. It is the row's only record of what was being
  imported, so stripping it would cost the post-mortem the retention window
  exists for, and it is not credential-shaped by design. If submitted URLs
  should be stored redacted, that is a decision about the whole ingest path
  rather than about the queue row, and belongs in its own issue.

## Area

- [x] Import / Ingestion
- [x] Auth / Permissions

## Testing

Counterfactuals, both run:

- Removing the sweep's purge call makes
  `test_the_sweep_purges_the_tokens_of_only_the_rows_it_failed` fail on
  `purge.await_args` being `None`.
- Adding `AND status NOT IN ('todo', 'doing')` to the by-id statement makes the
  DB-backed `test_a_named_row_loses_its_token_while_a_live_one_keeps_everything`
  fail with the token still present, which is what proves the by-id statement,
  not the status-gated backstop, is what strips a `doing` row.

Host pytest at `localhost:5434`:

```
tests/test_stalled_queue_sweep_624.py + 6 queue/token files       434 passed
tests/test_layering.py + rule1 + rule2                            168 passed
```

`ruff check`, `ruff format --check` and `backend/tests/finding_markers.py`
pass.

## Checklist

Refs #1755

### DIFF
--- a/backend/app/platform/jobs/sweep.py
+++ b/backend/app/platform/jobs/sweep.py
@@ -7,6 +7,7 @@ lifespan sweeper, worker startup recovery, and the admin cleanup endpoint in
 """
 
 import uuid
+from collections.abc import Sequence
 from dataclasses import dataclass, field, replace
 from datetime import datetime, timedelta, timezone
 from pathlib import Path
@@ -1019,6 +1020,27 @@ def publish_refresh_reconciliation(outcome: StaleCleanupOutcome) -> None:
     """
     if outcome._refresh_runs_reconciled:
         refresh_sweep_reconciled_total.inc(outcome._refresh_runs_reconciled)
+
+
+_PURGE_JOB_TOKENS_BY_ID_SQL = (
+    "UPDATE catalog.procrastinate_jobs SET args = args - 'token' "
+    "WHERE id = ANY(:job_ids)"
+)
+
+
+async def purge_queue_row_tokens(db: AsyncSession, job_ids: Sequence[int]) -> None:
+    """Drop the raw service token from the named queue rows.
+
+    fix(#1755 item 12): the one statement both immediate purge sites use —
+    a task that fails on its own (``purge_token_on_failure``) and the
+    stalled sweep, which is the first moment a crashed worker's token is
+    provably dead weight. Both service tasks are ``retry=0``, so nothing
+    re-runs from these args.
+    """
+    if not job_ids:
+        return
+    await db.execute(text(_PURGE_JOB_TOKENS_BY_ID_SQL), {"job_ids": list(job_ids)})
+    await db.commit()
 
 
 async def purge_terminal_job_tokens(db: AsyncSession) -> None:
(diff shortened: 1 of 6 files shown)


ishiland:
@codex review

PR #2042, opened 2026-09-09
fix(reupload): answer 422 for a file the preview cannot read

### LINKED ISSUE #2036: Reupload preview returns 500 for a malformed file where the import preview returns 422
Found in the 1.19.0 pre-tag smoke on the dev stack (main d39c1bbdc); the same handler shape exists in v1.18.1, so this is not a regression of this cycle.

`POST /api/datasets/{id}/reupload/{job}/preview` with an unreadable `.geojson` (valid UTF-8, broken JSON) returns `500 Internal server error`, reproducibly. The equivalent import route `POST /api/ingest/preview/{job}` returns `422 "Unable to preview file. The file may be malformed or unsupported."` for the identical file.

Cause: `backend/app/modules/catalog/datasets/api/router_reupload.py` (the `run_ogrinfo_preview` call in the reupload preview handler) catches only `UnsafeUploadError` (#1846), so the `IngestionError` from a failed `ogrinfo` escapes as a 500. `backend/app/processing/ingest/router.py` has the broad handler that maps it. The CLI (`geolens replace <dataset> <file> --wait`) hits the same 500.

Secondary, same path: the stored reason for an unopenable vector file is `ogrinfo failed (exit 1)` (path-free, but not the sentence #2028 intended) because `_OGR_UNABLE_TO_OPEN_RE` in `processing/ingest/ogr.py` matches ogr2ogr's "Unable to open datasource" wording and not this build's `ogrinfo failed - unable to open '<path>'`.

ishiland:
## Summary

`POST /api/datasets/{id}/reupload/{job}/preview` returned 500 for a file GDAL cannot read. The
handler caught only `UnsafeUploadError` around its ogrinfo call, so the `IngestionError` a
malformed file raises escaped unhandled. The import preview answers 422 for the identical file.
The CLI's `geolens replace <dataset> <file> --wait` hits the same door.

Secondary, same path: the reason stored for an unopenable vector file was `ogrinfo failed
(exit 1)`. The unable-to-open pattern matched ogr2ogr's driver enumeration but not the one-line
wording ogrinfo prints for the same failure, so the sentence #2028 intended was never reached.

## Changes

- The re-upload preview maps every non-`UnsafeUploadError` failure from the preview call to the
  import preview's 422 and detail text, with the same `logger.exception` record. The
  `UnsafeUploadError` branch from #1846 still answers first with its own message, and the
  staged-download unlink still runs on the way out.
- The unable-to-open pattern gains ogrinfo's `ogrinfo failed - unable to open '<path>'` wording,
  so both tools reach the sentence built from the source filename alone, through the same
  `core/failure_reason` door as before. Confirmed against ogrinfo on GDAL 3.13.
- The re-upload service preview was checked and left alone. It maps `IngestionError` to 502,
  which is the right answer for a remote service, and every typed refusal below it already
  arrives as a finished `HTTPException`.

## Area

- [x] Import / Ingestion

## Testing

Run from the worktree with `.env.test` sourced:

```
pytest tests/test_reupload.py tests/test_ingest_ogr_pure.py tests/test_ingest_open_failure_message.py   177 passed
pytest tests/test_gdal_driver_clamp_1846.py tests/test_rule1_structural.py tests/test_pooled_connection_release_1848.py   113 passed
pytest tests/test_rule2_structural.py tests/test_stored_failure_reason_1953.py tests/test_failure_reason_readers_2010.py   152 passed
pytest tests/test_reupload_idor.py tests/test_reupload_service.py   12 passed
pytest tests/test_layering.py   50 passed
ruff check . && ruff format --check .   clean
python3 backend/tests/finding_markers.py   clean
make public-surface-check   passed
```

Counterfactuals, each observed with the fix reverted:

- With the broad handler removed, the new preview test gets an unhandled `IngestionError` and a
  500 from the route.
- With the ogrinfo alternative removed from the pattern, both new pattern tests and the new
  real-binaries ogrinfo test report `ogrinfo failed (exit 1)` instead of the friendly sentence.

`_MODULE_LOC_CAPS` re-measured for both changed modules: `router_reupload.py` 1428 to 1436,
`ogr.py` 1341 to 1343, both exact.

## Checklist

No API schema change: no route docstring or field description was touched, so no SDK or type
regeneration is needed. No new user-facing string: the 422 detail is the one the import preview
already returns.

## Deferred

A `.parquet` re-upload that exceeds the ingest ceiling now answers the generic 422 rather than
the ceiling message the import preview keeps for that class. It answered 500 before, so nothing
regresses; matching the import door would need the budget error class on the catalog port.

Closes #2036

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -210,6 +210,10 @@ and releases use semantic versioning.
   An unrecognised local-file failure records the tool and its exit status, with the driver's full
   output kept in the structured logs. The CLI job views and the ingest-failure notification
   render the same sentence the web app shows, instead of the stored code. (#2010)
+- Replacing a dataset with a file the server cannot read now answers the same 422 and the
+  same message the import preview gives, instead of a generic server error. The reason
+  recorded for such a file names the file that was uploaded, rather than reporting only the
+  tool and its exit status. (#2036)
 
 ## [1.18.1] - 2026-09-05
 
--- a/backend/app/modules/catalog/datasets/api/router_reupload.py
+++ b/backend/app/modules/catalog/datasets/api/router_reupload.py
@@ -633,6 +633,14 @@ async def reupload_preview(
             status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
             detail=str(exc),
         ) from exc
+    except Exception as exc:  # broad: GDAL subprocess can raise various errors on unsupported/malformed files
+        # fix(#2036): the 422 the import preview answers for the same failure.
+        # A malformed file's IngestionError escaped this door as a 500.
+        logger.exception("ogrinfo_preview failed", job_id=str(job_id), error=str(exc))
+        raise HTTPException(
+            status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+            detail="Unable to preview file. The file may be malformed or unsupported.",
+        ) from exc
     finally:
         if downloaded_preview_path is not None:
             downloaded_preview_path.unlink(missing_ok=True)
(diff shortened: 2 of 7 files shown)


ishiland:
@codex review

PR #2051, opened 2026-09-09
fix(auth): stop a transient refresh failure signing out every session

### LINKED ISSUE #2038: A transient refresh failure in one client signs the user out of every session
Observed while running parallel browser sessions against one account on the dev stack (main d39c1bbdc); the same code paths exist in v1.18.1.

`frontend/src/api/client.ts` treats a 401 whose refresh also fails as a dead session and dispatches a best-effort `POST /auth/logout/` (#1446, so a still-valid refresh cookie is revoked). The server's logout calls `revoke_all_tokens` for the user, so every other session of that user (another device, another tab with its own store) starts getting 401s, fails its refresh, and fires its own logout. The #1446 comment itself names transient causes for the first failure (a 429, a 5xx, a dropped connection), which means a refresh rate-limit hit in one tab logs the user out everywhere.

Ask: decide whether logout should revoke only the calling session's refresh row (keeping the explicit "sign out everywhere" action for all sessions), or whether the dead-session path should skip the server-side revocation when the failure was transient. Either way, add a test that a 429 on refresh in one client does not invalidate a second client's access token.

ishiland:
## Summary

A transient refresh failure in one browser client signed the user out of every
session. `frontend/src/api/client.ts` treated "401, and the follow-up refresh
also failed" as session death regardless of *why* the refresh failed, and its
dead-session path dispatched a best-effort `POST /auth/logout/` (#1446). That
endpoint calls `revoke_all_tokens`, so it invalidates every refresh row and
bumps `token_version` for the whole user. Every other client of that user then
started 401ing, failed its own refresh, and dispatched its own revocation — the
cascade observed on 2026-09-09 as six logouts in three minutes with nobody
logging out.

The rule this PR establishes: a client asks the server to revoke only on
evidence the server rejected the credential — a 401 or 403 — because
`/auth/logout/` is user-wide, not session-scoped. Everything else (a 429, a
5xx, a dropped connection, an abort, a malformed body) clears local state and
sends the user to `/login` without touching anyone else's session.

Client-side only. `revoke_all_tokens` keeps its semantics, no endpoint was
added, and explicit user logout still signs out everywhere as before.

## Changes

- `notifySessionExpired` no longer dispatches `logoutSession()`. It still
  aborts any in-flight refresh (so a late `Set-Cookie` cannot land on a later
  session, per #1446) and still clears the store and raises the single
  signed-out prompt.
- `tryRefresh`'s classification is split out as `attemptRefresh(): RefreshOutcome`
  (`refreshed` / `rejected` / `transient`). `tryRefresh` stays a boolean wrapper,
  so its existing call sites and their tests are untouched. Both 401 doors — the
  shared fetch core and the XHR upload path in `api/ingest.ts` — end the session
  only for a `rejected` credential.
- The three catches that revoked on *any* failure now gate on the same rule:
  the OAuth callback's `getMe` catch, the login `getMe` catch in `use-auth.ts`,
  and login's body-parse catch in `api/auth.ts`. Without this, SSO completing on
  a new device while `/auth/me/` answered 500 revoked every session.
- One shared predicate, `isCredentialRejected`, is the single definition of
  "the server rejected this credential", used by both the refresh
  classification and the three catches.
- A 401 raised because the refresh failed transiently carries `unconfirmed: true`,
  and `isCredentialRejected` refuses a flagged error. Without it a transient failure
  inside `getMe` surfaced as a 401 that the sign-in catches read as a rejection and
  revoked every session for. One boolean on `ApiError`, set in the two 401 doors that can
  produce an unverified 401, read only by the predicate.
- The OAuth callback's malformed-fragment bail-out no longer revokes either. A fragment
  too short to finish sign-in says nothing about the credential.
- A `transient` outcome arms a 30s back-off before the next attempt. Without
  one, an auth outage or a shared egress IP over the refresh endpoint's per-IP
  limit turned every 401'd surface in every tab into an unbounded refresh loop.
  The back-off yields to a peer tab's rotation — the same live-session evidence
  the failure path already trusts — and clears on a deliberate session end.

## Area

- [x] Frontend (UI, components, hooks)
- [x] Auth / Permissions

## Testing

Twelve tests fail on `origin/main` and pass here, verified by restoring the
source files from `origin/main` and re-running:

- `client.session-expiry.test.ts` — a rate-limited refresh keeps the session and
  issues no revocation; so does a refresh the client cannot deliver at all
  (a `TypeError`, which is what `refreshAccessToken`'s bare `fetch` throws); a
  rejected refresh clears local state without a revocation; and 403 ends the
  session exactly as 401 does. The rate-limit case is the scenario the issue
  asks for: no `POST /auth/logout/` is issued, which is the only thing that
  could have invalidated another client's tokens.
- `ingest-session-expiry-1446.test.ts` — the upload door does the same, with a
  transient (503) case alongside the terminal one.
- `OAuthCallbackPage.test.tsx` and `use-auth.test.tsx` — a 500 from `/auth/me/`
  after a good sign-in keeps the user's other sessions; a 401 still revokes.
- `refresh-cookie-1302.test.ts` — a login response with an unparseable body no
  longer revokes every session.
- `client.test.ts` — after a transient failure the client stops asking until the
  back-off expires, then asks again (fake timers).
- The transient 401 carries `unconfirmed: true` and the rejected one does not; the OAuth
  callback keeps the user's other sessions when `getMe` answers an unconfirmed 401; an
  incomplete fragment no longer revokes.

Gates run from `frontend/`, all green: `npm run build`, `npm run lint`,
`npm run typecheck`, `npm run test:coverage` (460 files, 6116 tests).

No API schema, docstring or config change, so no `make sdks` / `types:generate`.
No new user-facing strings, so no locale change.

## Checklist

- [x] i18n: no new user-facing strings
- [x] Migration included (if DB schema changed) — n/a

## Notes for review

Two residuals are deliberate, both consequences of `/auth/logout/` being user-wide
with no per-session alternative.

1. When a refresh *succeeds* and the retry is still 401, the freshly rotated
   refresh row stays alive server-side while the client signs out locally.
   That case implies the server is rejecting a token it just minted, so the
   credential chain is invalid anyway.
2. At login's body-parse catch the rejection predicate is unreachable — the
   `try` only covers a 2xx, so the sole error it can see is a malformed body.
   That site therefore no longer revokes at all. A bad response body is no
   evidence of a rejected credential, and revoking on it ends the user's other
   devices.

Refresh rotation has no reuse detection, so with server-side revocation gone
from the dead-session path, a refresh credential someone else rotated stays
valid until an explicit logout or a password change. Closing that needs a
revoke-this-session endpoint and reuse detection on the backend, which is out
of scope here.

Closes #2038

### DIFF
--- a/frontend/src/api/__tests__/client.session-expiry.test.ts
+++ b/frontend/src/api/__tests__/client.session-expiry.test.ts
@@ -1,4 +1,4 @@
-import { apiFetch, ApiError, onSessionExpired } from '@/api/client';
+import { abortInflightRefresh, apiFetch, ApiError, onSessionExpired } from '@/api/client';
 import { useAuthStore } from '@/stores/auth-store';
 import { refreshAccessToken, logoutSession } from '@/api/auth';
 import type { TokenResponse } from '@/types/api';
@@ -65,6 +65,8 @@ describe('session-expiry notification (fix #628)', () => {
   afterEach(() => {
     unregister();
     useAuthStore.setState({ token: null, refreshToken: null, expiresAt: null, user: null });
+    // fix(#2038): the transient refresh back-off is module state; end signed out.
+    abortInflightRefresh();
   });
 
   // fix(#2038): a rate-limited refresh in one client used to revoke every
@@ -82,12 +84,13 @@ describe('session-expiry notification (fix #628)', () => {
     expect(useAuthStore.getState().token).toBe(live);
   });
 
-  // fix(#2038): a dropped connection is no evidence the credential is dead.
+  // fix(#2038): refreshAccessToken issues a bare fetch, so a real offline
+  // failure arrives as a TypeError, not an ApiError.
   it('keeps the session and revokes nothing when the refresh cannot reach the server', async () => {
     signIn();
     const live = useAuthStore.getState().token;
     mockFetch.mockResolvedValue(errorResponse(401));
-    vi.mocked(refreshAccessToken).mockRejectedValue(new ApiError('network error', 0));
+    vi.mocked(refreshAccessToken).mockRejectedValue(new TypeError('Failed to fetch'));
 
     await expect(apiFetch('/a/')).rejects.toMatchObject({ status: 401 });
 
@@ -101,6 +104,20 @@ describe('session-expiry notification (fix #628)', () => {
     signIn();
     mockFetch.mockResolvedValue(errorResponse(401));
     vi.mocked(refreshAccessToken).mockRejectedValue(new ApiError('unauthorized', 401));
+
+    await expect(apiFetch('/a/')).rejects.toMatchObject({ status: 401 });
+
+    expect(logoutSession).not.toHaveBeenCalled();
+    expect(handler).toHaveBeenCalledTimes(1);
+    expect(useAuthStore.getState().token).toBeNull();
+  });
+
+  // fix(#2038): 403 on refresh is a rejection too, and ends the session the
+  // same way 401 does.
+  it('clears local state and prompts when the refresh is answered 403', async () => {
+    signIn();
+    mockFetch.mockResolvedValue(errorResponse(401));
+    vi.mocked(refreshAccessToken).mockRejectedValue(new ApiError('forbidden', 403));
 
     await expect(apiFetch('/a/')).rejects.toMatchObject({ status: 401 });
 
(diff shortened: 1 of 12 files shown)


ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

ishiland:
@codex review

